PR #6, opened 2020-08-26
Add H5Dchunk_iter method for iterating over chunks

gauteh:
This method iterates over all chunks in dataset, calling a user-supplied
callback with the chunk information and optional user supplied data.

The iterator is stopped when ITER_STOP is returned by the user-supplied
callback or the iterator is exhausted.

Existing methods to get chunk_info performs an iteration each time, so
to get many or all chunks causes SUM(i) for i = 0 -> N operations for N
chunks, as opposed to N operations when using this iterator for this use case.

About 200x speedup for iterating over all chunks in a 1.4gb file (previously 25 seconds, now 140 ms).

---
This change is [Reviewable](https://reviewable.io/reviews/hdfgroup/hdf5/6)

### DIFF
--- a/src/H5D.c
+++ b/src/H5D.c
@@ -2199,7 +2199,7 @@ H5Dchunk_iter(hid_t dset_id, H5D_chunk_iter_op_t cb, void *op_data)
         HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "invalid dataset identifier")
 
     /* Call private function to get the chunk info given the chunk's index */
-    if(H5VL_dataset_optional(vol_obj, H5VL_NATIVE_DATASET_CHUNK_ITER, H5P_DATASET_XFER_DEFAULT, H5_REQUEST_NULL, cb, op_data) < 0)
+    if(H5VL_dataset_specific(vol_obj, H5VL_DATASET_CHUNK_ITER, H5P_DATASET_XFER_DEFAULT, H5_REQUEST_NULL, cb, op_data) < 0)
         HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "Can't iterate over chunks")
 
 done:
--- a/src/H5D.c
+++ b/src/H5D.c
@@ -2142,3 +2142,66 @@ H5Dget_chunk_info_by_coord(hid_t dset_id, const hsize_t *offset, unsigned *filte
 done:
     FUNC_LEAVE_API(ret_value)
 } /* end H5Dget_chunk_info_by_coord() */
+
+/*-------------------------------------------------------------------------
+ * Function:    H5Dchunk_iter
+ *
+ * Purpose:     Iterates over all chunks in dataset with given callback and user data.
+ *
+ * Parameters:
+ *              hid_t dset_id;          IN: Chunked dataset ID
+ *              H5D_chunk_iter_op_t cb  IN: User callback function, called for every chunk.
+ *              void *op_data           IN/OUT: Optional user data passed on to user callback.
+ *
+ * Callback information:
+ *      H5D_chunk_iter_op_t is defined as:
+ *
+ *        typedef int (*H5D_chunk_iter_op_t)(
+ *            const hsize_t *offset,
+ *            uint32_t filter_mask,
+ *            haddr_t addr,
+ *            uint32_t nbytes,
+ *            void *op_data);
+ *
+ *      H5D_chunk_iter_op_t parameters:
+ *          hsize_t *offset;        IN/OUT: Array of starting logical coordinates of chunk.
+ *          uint32_t filter_mask;   IN: Filter mask of chunk.
+ *          haddr_t addr;           IN: Offset in file of chunk data.
+ *          uint32_t nbytes;        IN: Size in number of bytes of chunk data in file.
+ *          void *op_data;          IN/OUT: Pointer to any user-defined data
+ *                                  associated with the operation.
+ *
+ *      The return values from an operator are:
+ *          Zero (H5_ITER_CONT) causes the iterator to continue, returning zero when all
+ *              elements have been processed.
+ *          Positive (H5_ITER_STOP) causes the iterator to immediately return that positive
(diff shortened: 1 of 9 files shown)


src/H5D.c
```diff
@@ -1135,3 +1135,66 @@ H5Dget_chunk_info_by_coord(hid_t dset_id, const hsize_t *offset, unsigned *filte
...
+    FUNC_ENTER_API(FAIL)
+
+    /* Check arguments */
+    if(NULL == (vol_obj = (H5VL_object_t *)H5I_object_verify(dset_id, H5I_DATASET)))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "invalid dataset identifier")
+
+    /* Call private function to get the chunk info given the chunk's index */
+    if(H5VL_dataset_optional(vol_obj, H5VL_NATIVE_DATASET_CHUNK_ITER, H5P_DATASET_XFER_DEFAULT, H5_REQUEST_NULL, cb, op_data) < 0)
+        HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "Can't iterate over chunks")
+
```
derobins, line 2206: Something like this should probably be done under specific, not optional.

src/H5D.c
```diff
@@ -1135,3 +1135,66 @@ H5Dget_chunk_info_by_coord(hid_t dset_id, const hsize_t *offset, unsigned *filte
...
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Dchunk_iter(hid_t dset_id, H5D_chunk_iter_op_t cb, void *op_data)
+{
+    H5VL_object_t  *vol_obj = NULL;     /* Dataset for this operation */
+    herr_t          ret_value = SUCCEED;
+
+    FUNC_ENTER_API(FAIL)
+
```
derobins, line 2197: Need to run bin/trace over this file to get the TRACE macros set up. The bin/trace script will also have to be modified to recognize your new type.

src/H5D.c
gauteh: Changed in c1bebb8

src/H5D.c
gauteh: Done in 197b2a8e51.

gauteh:
Thanks for your comments, I think the MSVC failure is unrelated.

derobins:
I think github might be out of sync with Bitbucket and is missing a splitter VFD fix on Windows. It is almost certainly not due to your changes.

Thank you for your contribution, but keep in mind that we don't make changes to the VFD interface lightly, so this is going to require some discussion before it goes anywhere.

gauteh:
I understand ☺️ thanks for considering! Speedup is considerable when going
through all the chunks in a file, for a 1.5 GB file it is 130ms with this
patch vs 26s previously.

tor. 27. aug. 2020, 10:16 skrev Dana Robinson <notifications@github.com>:

gauteh:
Rebased off current develop.

src/H5Dchunk.c
```diff
@@ -7453,3 +7459,88 @@ H5D__get_chunk_info_by_coord(const H5D_t *dset, const hsize_t *offset, unsigned
...
+    layout = &(dset->shared->layout);
+    rdcc = &(dset->shared->cache.chunk);
+    HDassert(layout);
+    HDassert(rdcc);
+    HDassert(H5D_CHUNKED == layout->type);
+
+    /* Search for cached chunks that haven't been written out */
+    for(ent = rdcc->head; ent; ent = ent->next)
+        /* Flush the chunk out to disk, to make certain the size is correct later */
+        if(H5D__chunk_flush_entry(dset, ent, FALSE) < 0)
```
fortnern: This seems like it could be surprising, that calling H5Dchunk_iter flushes all chunks. If this is the way we want it to work it should be well documented. It would also be possible to return the cache status of chunks.

src/H5Dchunk.c
```diff
@@ -7453,3 +7459,88 @@ H5D__get_chunk_info_by_coord(const H5D_t *dset, const hsize_t *offset, unsigned
...
+ *              August 2020
+ *
+ *-------------------------------------------------------------------------
+ */
+static int
+H5D__chunk_iter_cb(const H5D_chunk_rec_t *chunk_rec, void *udata)
+{
+    const H5D_chunk_iter_cb_data_t * data = (H5D_chunk_iter_cb_data_t *) udata;
+
+    return (data->cb)(chunk_rec->scaled, chunk_rec->filter_mask, chunk_rec->chunk_addr, chunk_rec->nbytes, data->op_data);
```
fortnern: It's strange not to use any FUNC_ENTER/FUNC_LEAVE macros here. Is there a reason they were left out? I understand the desire to maximize performance but my opinion is that should be done using macros that reduce to no-ops when compiled in production mode.

src/H5Dchunk.c
gauteh: I probably just carried this over from the other iterating functions. I don't know if it has been changed there?

fortnern: We can leave this how it is for now

src/H5Dchunk.c
fortnern: Can you add FUNC_ENTER_STATIC_NOERR/FUNC_LEAVE_NOAPI macros here? After this we can merge to develop.

gauteh: Ok, I'll do that.

gauteh:
Changed according to comments, added some documentation to fit with the new style. Not added a `\since ...` line. Rebased off develop.

fortnern (approved):

byrnHDF (approved):

lrknox (approved):

lrknox:
checks other than format passed in action #1022.  Running format_source on jelly with clang 10.0.1 didn't generate any changes, so I will merge.

src/H5D.c
qkoziol: Actually, all the other chunk operations are optional.   I'll retarget it in the incoming combo branch PR.

PR #7, opened 2020-09-09
Add cmake variable HDF5_LIB_INFIX

barcode:
**What**
This infix is added to all library names after 'hdf5'.
e.g. the infix '_openmpi' results in the library name 'libhdf5_openmpi.so'
This name is used in packages on debian based systems.
(see https://packages.debian.org/jessie/amd64/libhdf5-openmpi-8/filelist)

This option is useful when testing projects on debian based systems with
custom builds of hdf5 while trying to minimize differences between the
custom setup and the environment created by installing system packages.

**Impact**
If the parameter is not set to a value, it has no impact. Hence this patch 
should not break existing setups.

**Note**
I created a PR against this branch, since i am currently using it.
If this PR is accepted, i can create a PR for `develop` to apply the same 
patch there.


byrnHDF (approved):
I have thought about this for quite awhile, I understand the reason, worried about the name.
I can't come up with a better name.

epourmal:
I think we should accept this pull request. Looks good to me and makes sense.

epourmal (approved):

qkoziol (approved):

lrknox (approved):

PR #15, opened 2020-10-03
Correct entries related to --enable-build-mode and --enable-profiling in

### LINKED ISSUE #9: Typos in INSTALL section "Optimization versus symbolic debugging
There are two typos in Section 4.3.8:

1. `--enable-build-mode=production` should be `--enable-build-mode=debug` on [line 354](https://github.com/HDFGroup/hdf5/blob/cdc4b8a0fb55af4cf1a7bb4da5d8c4d454a0b8b6/release_docs/INSTALL#L354).
1. `--enable-production` should be `--enable-profiling` on [line 362](https://github.com/HDFGroup/hdf5/blob/cdc4b8a0fb55af4cf1a7bb4da5d8c4d454a0b8b6/release_docs/INSTALL#L362).

lrknox:
INSTALL file, and remove obsolete SZIP paragraph from COPYING file.
Fix #9

### DIFF
--- a/release_docs/INSTALL
+++ b/release_docs/INSTALL
@@ -351,15 +351,18 @@ CONTENTS
         so it can be debugged with gdb, dbx, ddd, etc., or it can be
         compiled with various optimizations. To compile for symbolic
         debugging (the default for snapshots), say
-        `--enable-build-mode=production'; to compile with optimizations
+        `--enable-build-mode=debug'; to compile with optimizations
         (the default for supported public releases),
-        say `--enable-build-mode=production'. On some systems the
+        say `--enable-build-mode=production'. For a 'clean slate' configuration
+        with optimization disabled and nothing turned on, 
+        say `--enable-build-mode=clean'. On some systems the
         library can also be compiled for profiling with gprof by saying
-        `--enable-production=profile'.
+        `--enable-profiling'.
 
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

ajelenak (approved):

release_docs/INSTALL
```diff
@@ -351,15 +351,18 @@ CONTENTS
...
-        `--enable-build-mode=production'; to compile with optimizations
+        `--enable-build-mode=debug'; to compile with optimizations
         (the default for supported public releases),
-        say `--enable-build-mode=production'. On some systems the
+        say `--enable-build-mode=production'. For a 'clean slate' configuration
+        with optimization disabled and nothing turned on, 
+        say `--enable-build-mode=clean'. On some systems the
         library can also be compiled for profiling with gprof by saying
-        `--enable-production=profile'.
+        `--enable-profiling'.
```
derobins, line 361: This is an improvement, but the build modes are really just collections of features. For example, you can turn debug symbols on/off using --enable-symbols=(yes|no|<custom>).

derobins (approved):

release_docs/INSTALL
lrknox: I'll merge this.  If INSTALL needs more detail we could create an issue for it.

PR #16, opened 2020-10-05
Fix memory leak in H5PL__find_plugin_in_path

michaelkuhn:
When encountering a directory, the current loop iteration was skipped without freeing the memory allocated for the path.

This also affects version 1.12.

### DIFF
--- a/src/H5PLpath.c
+++ b/src/H5PLpath.c
@@ -669,8 +669,10 @@ H5PL__find_plugin_in_path(const H5PL_search_params_t *search_params, hbool_t *fo
                             HDstrerror(errno))
 
             /* If it is a directory, skip it */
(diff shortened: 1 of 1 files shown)


qkoziol (approved):

jhendersonHDF (approved):

michaelkuhn:
Thanks for the quick merge! Should I also submit a backport to 1.12 or will you take care of it?

derobins:
If you can also set up a PR for 1.12, that would be helpful. Thanks!

lrknox:
Thank you for the fix!

I will take care of it after it runs through daily testing on the develop branch for a couple of days.  However, if you want to submit it you are welcome to do so.

Larry

lrknox:
Ok, I’ll do It now.

PR #23, opened 2020-10-06
Fix memory leak in H5PL__find_plugin_in_path

michaelkuhn:
Since I could not find anything on the 1.12 branch, here is a backport of #16.

When encountering a directory, the current loop iteration was skipped without freeing the memory allocated for the path.

### DIFF
--- a/src/H5PLpath.c
+++ b/src/H5PLpath.c
@@ -669,8 +669,10 @@ H5PL__find_plugin_in_path(const H5PL_search_params_t *search_params, hbool_t *fo
                             HDstrerror(errno))
 
             /* If it is a directory, skip it */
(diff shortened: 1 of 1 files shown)


lrknox:
The test on Windows fails due to another problem in the hdf5_1_12 branch.  The test won't pass until that is fixed.  It should be fixed in a few days or so.  We will add another comment at that time, and you can either update your fork with the changes or I can create a new PR..

lrknox:
Actually, the change in this PR is also in PR #13.  When hdf5_1_12 passes the code change will come in through that PR.

lrknox:
This change will be merged as part of PR #28, along with changes to make the test pass.

Thanks!

PR #34, opened 2020-10-14
Add posix_errno to H5E_error2_t struct

takluyver:
This is my patch from [HDFFV-10631](https://jira.hdfgroup.org/browse/HDFFV-10631). I'm interested to see if the CI passes. :-)

### DIFF
--- a/src/H5E.c
+++ b/src/H5E.c
@@ -1357,7 +1357,7 @@ H5Epush2(hid_t err_stack, const char *file, const char *func, unsigned line, hid
         HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "memory allocation failed")
 
     /* Push the error on the stack */
-    if (H5E__push_stack(estack, file, func, line, cls_id, maj_id, min_id, tmp) < 0)
+    if (H5E__push_stack(estack, file, func, line, cls_id, maj_id, min_id, tmp, 0) < 0)
         HGOTO_ERROR(H5E_ERROR, H5E_CANTSET, FAIL, "can't push error on stack")
 
 done:
--- a/src/H5Edeprec.c
+++ b/src/H5Edeprec.c
@@ -203,7 +203,7 @@ H5Epush1(const char *file, const char *func, unsigned line, H5E_major_t maj, H5E
     H5TRACE6("e", "*s*sIuii*s", file, func, line, maj, min, str);
 
     /* Push the error on the default error stack */
-    if (H5E__push_stack(NULL, file, func, line, H5E_ERR_CLS_g, maj, min, str) < 0)
+    if (H5E__push_stack(NULL, file, func, line, H5E_ERR_CLS_g, maj, min, str, 0) < 0)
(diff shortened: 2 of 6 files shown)


takluyver:
Just to reiterate what I said about the patch: I'm not much of a C programmer, so this may be unusable. My main aim is to illustrate why I think this is not too big a change.

I haven't worked yet on integrating this with h5py; I hope to look into that at some point when I get time.

qkoziol:
Hi Thomas!
This would be a modification to the public interface, so it needs a high degree of thought, and if accepted, would only be possible to deploy in a future "major" release (i.e. 1.14.0).    With that in mind, I looked at your code and think that expanded all the internal calls to push errors on an error stack with a '0' POSIX errno would not be a great direction to go.   What do you think about pushing _another_ error on the error stack, like the MPI errors, with the string from strerror() as the description?   That would allow the POSIX errno to be only captured and maintained for POSIX calls and would avoid modifying the public API also.

qkoziol (changes requested):
Switch to pushing another error on the error stack, using strerror(), instead of expanding the parameters for pushing errors.

takluyver:
I'm not familiar with the error handling for MPI - is there a particular bit of code you can point me to?

My interest in this is in having the errno itself available in a machine-readable way, to allow more consistent exceptions from h5py. I'm not sure there is a backwards-compatible way to do that, except by scraping it from the string description (`errno = %d`) - we've asked about that previously, and been told that we shouldn't rely on it.

qkoziol:
Sure - take a look at HMPI_ERROR and related macros.

For machine readability, you can use a “standardized” way to format the error description that you push on the error stack, so the idea you have could work reasonably well (although you still are scraping the errno value).

takluyver:
Thanks, I've found the macros you mentioned.

If string scraping is going to be the answer anyway, is there any chance of declaring that the `errno = %d` already in the description for these errors is reliable? It's been consistent since you wrote it 11 years ago (a44328438aa5bbe4c4927b584af18d7965e6794a), so we could instantly support it in existing versions of HDF5. The only reason we're not already doing that is that we were told the description can't be relied on - which I understand as a general policy, but it would be really convenient to rely on this specific detail. @tacaswell relayed that to me; I'm not sure who he asked.

qkoziol:
Well, true, but I think adding the output from strerror() is valuable also.   And that description string would be a standardized place that we could guarantee the errno scraping didn't change.

takluyver:
The strerror is already there as well :wink:

qkoziol:
Ah - missed that earlier.  :-)  Seems like you should be OK then - maybe just a comment in the HSYS*  error macros that h5py (and maybe other layers) is depending on the output format, so it doesn't accidentally get changed in the future?

tacaswell:
My memory of walking through this was in the context of https://github.com/h5py/h5py/pull/1099

takluyver:
OK, I've opened a new PR to add the comment: #37. We'll look at using this after releasing h5py 3.0. Thanks!

PR #35, opened 2020-10-14
Updates CMake files to work with VFD SWMR

derobins:
- CMake files were updated to build new files in src and test.

- As with legacy SWMR, the test programs that are run via shell
  scripts are built but not run.

- Updated whitespace in the links_env output file. It's unclear why
  this changed, but CMake does an exact diff on the file whereas
  the autotools do not.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,6 +12,7 @@ set (H5_SOURCES
     ${HDF5_SRC_DIR}/H5system.c
     ${HDF5_SRC_DIR}/H5timer.c
     ${HDF5_SRC_DIR}/H5trace.c
+    ${HDF5_SRC_DIR}/hlog.c
 )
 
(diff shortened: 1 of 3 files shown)


qkoziol:
@derobins - Did you want me to review?   I could start looking at the vfd_swmr branch, possibly in pieces, but is it ready for that?

jrmainzer (approved):
I'm not familiar with CMake, but your changes look reasonable to me.

derobins:
> @derobins - Did you want me to review? I could start looking at the vfd_swmr branch, possibly in pieces, but is it ready for that?

This is just a change to the CMake files in the VFD SWMR branch. It's not going into develop anytime soon. Feel free to review, but there's nothing interesting, library-wise, going on here :)

byrnHDF (approved):

PR #37, opened 2020-10-15
Note that the description for system errors may be used by h5py

takluyver:
Following discussion on #34, we decided that this could simply be a comment in the source code. The string format here has been stable for 11 years (since a44328438aa5bbe4c4927b584af18d7965e6794a), so hopefully there's no major pressure to change it.

We're just about to release h5py 3.0, so we'll look into getting the errno from the string after that.

### DIFF
--- a/src/H5Eprivate.h
+++ b/src/H5Eprivate.h
@@ -114,14 +114,18 @@ typedef struct H5E_t H5E_t;
 #define HSYS_DONE_ERROR(majorcode, minorcode, retcode, str)                                                  \
     {                                                                                                        \
         int myerrno = errno;                                                                                 \
-        /* h5py may rely on the description format to get the errno - please try to avoid breaking it */     \
+        /* Other projects may rely on the description format to get the errno and any changes should be      \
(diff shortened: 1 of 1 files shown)


byrnHDF:
I am not disagreeing with the idea, just the content of the message.
I think the specific use of one project, h5py, is the main issue. It should be more general, something like:
/* Other projects may rely on the description format to get the errno and any changes should be considered as an API change */

takluyver:
That makes sense; I've updated the comment.

qkoziol:
Nice!

qkoziol (approved):

derobins (approved):

derobins:
This is fine as long as people are just relying on the errno = %d part and not the error strings we tack on to the end. Those absolutely cannot be considered a part of the public API or even a trivial refactoring will be an "API change".

takluyver:
Thanks! I am planning to only rely on the errno part. Specifically, we're likely to match it with a regex like `errno = (\d+)`.

PR #52, opened 2020-10-20
Switch ID code to use a hash table instead of a skip list

derobins:
Replaces the skip lists in the ID code with hash tables via BSD-licensed uthash.

http://troydhanson.github.io/uthash/userguide.html

This was only done for the library ID types. uthash can't handle delete callbacks that delete other IDs, which are valid (and explicitly tested for in tid.c) in user-defined types. User-defined ID types still use the skip lists.

Note that I'm just using the naked macros in this code. This should probably be hidden behind a package interface, but at this point I'd rather get useful performance improvements into the library than polish apples.


src/H5I.c
```diff
@@ -556,8 +581,23 @@ H5I_clear_type(H5I_type_t type, hbool_t force, hbool_t app_ref)
...
+            if (FAIL == ret)
+                HGOTO_ERROR(H5E_ATOM, H5E_BADITER, FAIL, "iteration failed")
+            if (TRUE == ret)
+                HASH_DELETE(hh, udata.type_ptr->ids.hash_table, item);
+        }
+    }
+    else
+        /* Attempt to free all ids in the type */
+        if (H5SL_try_free_safe(udata.type_ptr->ids.skip_list, H5I__clear_type_cb, &udata) < 0)
         HGOTO_ERROR(H5E_ATOM, H5E_CANTDELETE, FAIL, "can't free ids in type")
```
derobins, line 601: Oops. Needs to be tabbed over one more.

gnuoyd:
When you say a callback, I think you mean an H5Iiterate() callback? What's stopping a callback from deleting an ID belonging to a library-defined type?

gnuoyd:
Do you have any performance measurements to share?

byrnHDF (approved):
I understand why uthash.h was added to the format exclusion, but since it is not a generated file, I wished we could have just added a format off/on comment in the file.

derobins:
> I understand why uthash.h was added to the format exclusion, but since it is not a generated file, I wished we could have just added a format off/on comment in the file.

I try to keep files from external sources unchanged. That makes it easier to drag in an updated version.

derobins:
> Do you have any performance measurements to share?

Nothing formal. Informally, in ID-heavy tests, I see maybe 5-ish%. @jhendersonHDF has seen significant improvement in some DAOS profiling - upwards of 25%. (DAOS makes very heavy use of dataspace IDs)

byrnHDF:
> > I understand why uthash.h was added to the format exclusion, but since it is not a generated file, I wished we could have just added a format off/on comment in the file.
> 
> I try to keep files from external sources unchanged. That makes it easier to drag in an updated version.

Because you did not add the file to CMakeLists.txt, CMake will not format it either. But don't forget the MANIFEST.

derobins:
> > > I understand why uthash.h was added to the format exclusion, but since it is not a generated file, I wished we could have just added a format off/on comment in the file.
> > 
> > 
> > I try to keep files from external sources unchanged. That makes it easier to drag in an updated version.
> 
> Because you did not add the file to CMakeLists.txt, CMake will not format it either. But don't forget the MANIFEST.

Dang. Forgot that. I'll add that today.

src/H5I.c
```diff
@@ -169,15 +179,25 @@ H5I_term_package(void)
...
+                    n++;
+            }
 
         /* If no types are used then clean up */
         if (0 == n) {
             for (type = 0; type < H5I_next_type; type++) {
                 type_ptr = H5I_id_type_list_g[type];
                 if (type_ptr) {
-                    HDassert(NULL == type_ptr->ids);
+                    if (H5I_IS_LIB_TYPE(type))
```
soumagne, line 196: You could add an ifdef debug to avoid the extra if

src/H5I.c
```diff
@@ -556,8 +581,23 @@ H5I_clear_type(H5I_type_t type, hbool_t force, hbool_t app_ref)
...
 
-    /* Attempt to free all ids in the type */
-    if (H5SL_try_free_safe(udata.type_ptr->ids, H5I__clear_type_cb, &udata) < 0)
+    if (H5I_IS_LIB_TYPE(type)) {
+        H5I_id_info_t *item = NULL;
+        H5I_id_info_t *tmp  = NULL;
+
+        /* This is a "delete-safe" iteration */
+        HASH_ITER(hh, udata.type_ptr->ids.hash_table, item, tmp)
+        {
```
soumagne, line 590: You need to add HASH_ITER to the .clang-format file so that it recognizes it's an iterator, add it to the ForEachMacros section

src/H5I.c
```diff
@@ -756,6 +802,7 @@ H5Iregister(H5I_type_t type, const void *object)
  *
  *-------------------------------------------------------------------------
  */
+H5_GCC_DIAG_OFF("implicit-fallthrough")
```
soumagne: can you explain why that needs to be silenced?

soumagne (approved):
Some minor fixes but looks good to me.

src/H5I.c
derobins: Will do

src/H5I.c
derobins: uthash is implemented as a bunch of macros in a header file and some of those macros use fall-through. To fix this, I have to turn off the warnings in the H5I calls that use uthash. I can't just drop the 'diagnostic off' warnings in the header file since the macros just get pasted in and older versions of gcc require diagnostic switches to be at the function level.

derobins: We do this in HDF5 where we use implicit fall-through in things like Duff's device.

src/H5I.c
```diff
@@ -2107,7 +2173,20 @@ H5I_iterate(H5I_type_t type, H5I_search_func_t func, void *udata, hbool_t app_re
...
 
         /* Iterate over IDs */
-        if ((iter_status = H5SL_iterate(type_ptr->ids, H5I__iterate_cb, &iter_udata)) < 0)
+        if (H5I_IS_LIB_TYPE(type)) {
+            H5I_id_info_t *item = NULL;
+            H5I_id_info_t *tmp  = NULL;
+
+            HASH_ITER(hh, type_ptr->ids.hash_table, item, tmp)
+            {
+                int ret = H5I__iterate_cb((void *)item, NULL, (void *)&iter_udata);
```
gnuoyd, line 2176: You can safely delete `item`: iteration will resume at `tmp`. But what if `__iterate_cb` deletes `tmp`?

src/H5I.c
```diff
@@ -43,6 +43,8 @@
 #include "H5Tpkg.h"      /* Datatypes                                */
 #include "H5VLprivate.h" /* Virtual Object Layer                     */
 
+#include "uthash.h" /* Hash table functionality                 */
```
qkoziol, line 46: It would be _really_ good to have a comment (here, or somewhere) that this is an external piece of code we're including.

src/H5I.c
```diff
@@ -169,15 +179,25 @@ H5I_term_package(void)
 
         /* How many types are still being used? */
         for (type = 0; type < H5I_next_type; type++)
-            if ((type_ptr = H5I_id_type_list_g[type]) && type_ptr->ids)
-                n++;
+            if (H5I_IS_LIB_TYPE(type)) {
```
qkoziol, line 182: Is there really a concern about the different type of pointers here?  Perhaps adding a void* to the union for when you want to operate on the data structure in a generic way?

qkoziol:
There's a _lot_ of if / else pairs here, which just makes it obvious that these are effectively "methods" on two sub-classes of the "stores IDs" abstract class.  It would be a lot easier to read and maintain if there was a little "class" struct with function pointers in it that was set up for each type of ID class.  Then the "hash table" subclass would only have methods that used the uthash macros and the "skip list" subclass would only have methods that made H5SL calls.   Then all the "if / else" pairs could be replaced with function pointer invocations.   As an example, you could refer to the memory vs. disk-based VL code in src/H5Tvlen.c.

src/H5I.c
gnuoyd: uthash.h seems to use /* FALLTHROUGH */ comments to mark purposeful fallthrough, why isn't GCC respecting those?

derobins: We set -Wimplicit-fallthrough=5, which ignores comments and attributes, so the only thing you can do is turn it off. -Wextra sets -Wimplicit-fallthrough=3, which is probably suitable for our purposes (the integer specifies which comments match in the regexp . I'm not going to do that here, though.

I'll set up a separate PR that removes -Wimplicit-fallthrough and the now redundant diagnostic on/off statements.

src/H5I.c
derobins: Will do

src/H5I.c
derobins: I thought about that, but was unconvinced a bunch of extra boilerplate was "better" than a half dozen if statements. It's not like we're going to expand this code to more ID data containers.

derobins:
> There's a _lot_ of if / else pairs here, which just makes it obvious that these are effectively "methods" on two sub-classes of the "stores IDs" abstract class. It would be a lot easier to read and maintain if there was a little "class" struct with function pointers in it that was set up for each type of ID class. Then the "hash table" subclass would only have methods that used the uthash macros and the "skip list" subclass would only have methods that made H5SL calls. Then all the "if / else" pairs could be replaced with function pointer invocations. As an example, you could refer to the memory vs. disk-based VL code in src/H5Tvlen.c.

I thought about that, but was unconvinced a bunch of extra boilerplate was "better" than a half dozen if statements. It's not like we're going to expand this code to more ID data containers.

src/uthash.h
```diff
@@ -0,0 +1,1150 @@
...
+typedef unsigned char uint8_t;
+#endif
+#elif defined(__GNUC__) && !defined(__VXWORKS__)
+#include <stdint.h>
+#else
+typedef unsigned int uint32_t;
+typedef unsigned char uint8_t;
+#endif
+
+#ifndef uthash_malloc
```
brtnfld, line 82: So what is the policy, do we have to "HDF5ied" the code, like using HDmalloc, HDfree, HDmemset, etc...

qkoziol:
> > There's a _lot_ of if / else pairs here, which just makes it obvious that these are effectively "methods" on two sub-classes of the "stores IDs" abstract class. It would be a lot easier to read and maintain if there was a little "class" struct with function pointers in it that was set up for each type of ID class. Then the "hash table" subclass would only have methods that used the uthash macros and the "skip list" subclass would only have methods that made H5SL calls. Then all the "if / else" pairs could be replaced with function pointer invocations. As an example, you could refer to the memory vs. disk-based VL code in src/H5Tvlen.c.
> 
> I thought about that, but was unconvinced a bunch of extra boilerplate was "better" than a half dozen if statements. It's not like we're going to expand this code to more ID data containers.

I don't think we're planning to add a third ID data container, but we might decide to rip out  the uthash implementation and use something else.   I still would advocate for "good object-oriented design" in favor of the current if / else implementation.

src/uthash.h
derobins: I'm not marking up external code any more than I have to.

src/H5I.c
derobins: This was fixed in develop and merged here. We no longer ignore fall-through comments in gcc.

src/H5I.c
derobins: The current implementation of H5Iiterate() does not support deleting IDs of that type while iterating.

jrmainzer:
> > > There's a _lot_ of if / else pairs here, which just makes it obvious that these are effectively "methods" on two sub-classes of the "stores IDs" abstract class. It would be a lot easier to read and maintain if there was a little "class" struct with function pointers in it that was set up for each type of ID class. Then the "hash table" subclass would only have methods that used the uthash macros and the "skip list" subclass would only have methods that made H5SL calls. Then all the "if / else" pairs could be replaced with function pointer invocations. As an example, you could refer to the memory vs. disk-based VL code in src/H5Tvlen.c.
> > 
> > 
> > I thought about that, but was unconvinced a bunch of extra boilerplate was "better" than a half dozen if statements. It's not like we're going to expand this code to more ID data containers.
> 
> I don't think we're planning to add a third ID data container, but we might decide to rip out the uthash implementation and use something else. I still would advocate for "good object-oriented design" in favor of the current if / else implementation.

Correct me if I'm wrong, but the  if / else pairs are there because uthash doesn't provide the mark and sweep functionality needed by some non-library applications of the ID code.  If it did, Dana would have removed the the skip list code completely.

This being the case, doesn't it make more sense to use the available resources to augment uthash to provide the necessary functionality and then remove the skip list code completely?  I don't see the rational behind adding the abstraction layer now unless we are sure we will have a use for it beyond keeping the skip list code and avoiding some if / else pairs.

Bottom line:  If resources permit, I suggest augmenting uthash and removing the if / else pairs.  If there are not resources for this, I don't see that there are resources for adding the abstraction layer. 

Some un-related observations:

I spent several hours reviewing uthash.h.  In its default configuration, it seems to be optimized more for space efficiency than for time efficiency.  In particular, the default target hash bucket size of 10 may not be suitable for our purposes, as I would expect each linked list entry traversal to cause a processor cache miss.  If time permits, it might be useful to experiment with smaller values for hash bucket size.

I also noticed that uthash does not move entries to the head of the hash bucket linked list when they are the target of a successful search.  Whether this is relevant depends on whether lookups for a given ID are likely to be clustered.

The code to re-size the hash table could take a while if the table gets large enough.  Do we have any idea how large the set of IDs for any given type can become?  Do we need to do some profiling to explore this?

Finally, I didn't see much in the way of stats collection.

Given the performance improvements that Dana reports, I don't see any of these issues as show stoppers -- although performance with large collections of ID's makes me a little nervous.  I mention them as points of be considered if further optimization is needed.

jrmainzer (approved):

gnuoyd:
A simple and fast implementation of an ID→object lookup table that replaces both the hash table and the skip list uses an expanding array of `H5I_id_info_t` indexed by some bits in the ID. A reserved value of `H5I_id_info_t` (`.obj_ptr == NULL`, perhaps) indicates an empty slot in the array. (For O(1) reuse of empty slots, all empties can be chained together, with the chain head stored in the `H5I_id_type_t`.) When an empty slot is needed, but none are left, the array can be realloc'd.

To iterate over the IDs in the lookup table, record the current size of the array, n, and scan array slots 0..n-1, visiting the non-empty slots. If any slot is emptied while an iteration is underway, that's ok, it will be skipped.

By my lights, the simple and fast implementation looks like the way to go.

lrknox (approved):

PR #56, opened 2020-10-22
Removes -Wimplicit-fallthrough=5 from the gcc warnings

derobins:
-Wimplicit-fallthrough=3 is added by -Wextra, which we already set.
Bumping the warning level only changes how fall-through comments are
parsed, with level 5 turning off fall-through comments entirely.
This is unnecessary and results in having to do extra work to squash
warnings when included external code uses fall-through.

This change also adds /* FALLTHROUGH */ comments where
H5_ATTR_FALLTHROUGH is used so compilers that don't use attributes
but do respect fall-through comments don't raise spurious warnings.

### DIFF
--- a/config/gnu-warnings/7
+++ b/config/gnu-warnings/7
@@ -3,5 +3,4 @@
 -Wduplicated-branches
 -Wformat-overflow=2
 -Wformat-truncation=1
--Wimplicit-fallthrough=5
 -Wrestrict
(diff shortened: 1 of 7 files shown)


byrnHDF (commented):
Should/can this warning be added to the other compiler warnings files?

derobins:
> Should/can this warning be added to the other compiler warnings files?

gcc includes -Wimplicit-fallthrough=3 in -Wextra (which we already set)
clang has -Wimplicit-fallthrough (which we set)
MSVC ignores implicit fallthrough and has no warning setting
We don't have very in-depth warnings set for other compilers

byrnHDF (approved):

qkoziol (approved):

PR #59, opened 2020-10-24
HD prefix updates in src/ and test/

derobins:
Adds missing HD prefixes to API calls in src and test.

Adds some extra processing to bin/checkposix to keep the noise
levels down when running the script (not comprehensive).

### DIFF
--- a/src/H5FDs3comms.c
+++ b/src/H5FDs3comms.c
@@ -1925,7 +1925,7 @@ H5FD__s3comms_load_aws_creds_from_file(FILE *file, const char *profile_name, cha
     };
     char *const setting_pointers[] = {
         aws_region,
-        key_id,
+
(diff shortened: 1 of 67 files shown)


src/H5FDs3comms.c
```diff
@@ -1925,7 +1925,7 @@ H5FD__s3comms_load_aws_creds_from_file(FILE *file, const char *profile_name, cha
     };
     char *const setting_pointers[] = {
         aws_region,
-        key_id,
+
```
derobins: This will be fixed in a bit.

derobins: Fixed

qkoziol (approved):

lrknox (approved):

PR #64, opened 2020-10-29
Fixes minor issues in various virtual file drivers

derobins:
- HDFS VFD files now ignored in the Autotools when the HDFS VFD is not
  being built.
- All VFD init code uses idiomatic H5I_INVALID_HID instead of -1
  or FAIL.
- The HDFS VFD now includes H5FDdrvr_module.h in the right place.
- The HDFS tests in test/hdfs.c no longer pass NULL to VERIFY macros, which
  raised warnings.
- Minor tweaks to fix const, unused variables, etc. warnings in
  several VFDs.

### DIFF
--- a/src/H5FDmpio.c
+++ b/src/H5FDmpio.c
@@ -1746,33 +1746,4 @@ H5FD__mpio_communicator(const H5FD_t *_file)
     FUNC_LEAVE_NOAPI(file->comm)
 } /* end H5FD__mpio_communicator() */
 
-/*-------------------------------------------------------------------------
- * Function:       H5FD__mpio_get_info
- *
- * Purpose:        Returns the file info of MPIO file driver.
- *
- * Returns:        Non-negative if succeed or negative if fails.
- *
- * Programmer:     John Mainzer
- *                 April 4, 2017
- *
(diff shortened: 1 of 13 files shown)


src/H5FDmpio.c
```diff
@@ -1746,33 +1746,4 @@ H5FD__mpio_communicator(const H5FD_t *_file)
...
-    if (!mpi_info)
-        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "mpi info not valid")
-
-    *mpi_info = &(file->info);
-
-done:
-    FUNC_LEAVE_NOAPI(ret_value)
-} /* H5FD__mpio_get_info() */
-
 #endif /* H5_HAVE_PARALLEL */
```
derobins, line 1749: This is dead code, leftover from when the MPI info was moved to a real property instead of MPI-IO VFD private info.

byrnHDF (commented):
One thing I noticed when doing the 1.8 merge is that the drivers basically have the same template, yet there were significant differences in common code/code order when comparing driver to driver.

lrknox (approved):

qkoziol (approved):

PR #67, opened 2020-10-30
CMake: Set specific INSTALL_DATA_DIR only on MSVC

xantares:
For MinGW we want INSTALL_DATA_DIR to be set to share instead of .
to properly install files in a standard package layout.


byrnHDF:
What platform are you using, because CMake distinguishes between win32 and mingw, so (NOT WIN32) should be correct?

xantares:
no, as WIN32 is also defined for MINGW
I see now "if (WIN32 OR MINGW)" is used often, but in fact MINGW implies WIN32, so "if (WIN32)" is enough

I'm using mingw from linux

byrnHDF:
> no, as WIN32 is also defined for MINGW
> I see now "if (WIN32 OR MINGW)" is used often, but in fact MINGW implies WIN32, so "if (WIN32)" is enough
> 
> I'm using mingw from linux

I believe Kitware changed that assumption with the release of the CMake 3 series. In fact, I remember that CMake configure generated a warning for (if WIN32 AND NOT MINGW).

xantares:
we are fine now, hdf5 requires 3.x, see #70

byrnHDF:
> no, as WIN32 is also defined for MINGW
> I see now "if (WIN32 OR MINGW)" is used often, but in fact MINGW implies WIN32, so "if (WIN32)" is enough
> 
> I'm using mingw from linux

Platform Variables
CMake sets certain variables to true depending on the current platform and toolchain in use. These always describe the target platform.  In older versions of CMake, these were the only way of detecting the current platform.


UNIX : is TRUE on all UNIX-like OS's, including Apple OS X and
CygWin


WIN32 : is TRUE on Windows. Prior to 2.8.4 this included CygWin


APPLE : is TRUE on Apple systems. Note this does not imply the
system is Mac OS X, only that APPLE is #defined in C/C++
header files.

MINGW : is TRUE when using the MinGW compiler in Windows

MSYS : is TRUE when using the MSYS developer environment in Windows

CYGWIN : is TRUE on Windows when using the CygWin version of cmake


MINGW is true also on linux if using the mingw compiler; however the end goal of using mingw is to produce windows programs, so I would think you always want the windows paths for the installation.

derobins (approved):

xantares:
@byrnHDF in fact windows users often dont even care about the installation, so at least it will make happy those who want to benefit from the standard paths for cleanly packaging the binaries.

lrknox (approved):

PR #68, opened 2020-10-30
CMake: Add an option to opt-out MSVC naming convention

xantares:
We might want to keep the default mingw import name, see:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-hdf5/hdf5-default-import-suffix.patch

### DIFF
--- a/config/cmake_ext_mod/HDFMacros.cmake
+++ b/config/cmake_ext_mod/HDFMacros.cmake
@@ -143,8 +143,8 @@ macro (HDF_SET_LIB_OPTIONS libtarget libname libtype)
     endif ()
   endif ()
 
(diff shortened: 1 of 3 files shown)


byrnHDF (changes requested):
This option will need to be added to the files in release_docs; 
1. INSTALL_CMake.txt in the HDF5 Advanced Options section likely around line 780.
2. RELEASE.txt in the New Features section at the top of the Configuration list, please follow pattern (short description, long description, initials and date)

xantares:
ok, done

byrnHDF (approved):

lrknox (approved):

PR #70, opened 2020-10-30
CMake: MINGW implies WIN32

xantares:
hdf5 requires cmake 3, so we can simplify these


byrnHDF (commented):
I'm not sure this assumption is valid, I will need to investigate as there are two use cases for mingw, 1. On windows compiling as linux and 2. On linux compiling for windows (on windows for windows would be included).

byrnHDF:
> I'm not sure this assumption is valid, I will need to investigate as there are two use cases for mingw, 1. On windows compiling as linux and 2. On linux compiling for windows (on windows for windows would be included).

Of course I am not sure about (1) as I have not tried that environment, and might be confusing it with cygwin.

byrnHDF:
I was confusing cygwin and mingw, however:
Platform Variables
CMake sets certain variables to true depending on the current platform and toolchain in use. These always describe the target platform.  In older versions of CMake, these were the only way of detecting the current platform.


UNIX : is TRUE on all UNIX-like OS's, including Apple OS X and
CygWin


WIN32 : is TRUE on Windows. Prior to 2.8.4 this included CygWin


APPLE : is TRUE on Apple systems. Note this does not imply the
system is Mac OS X, only that APPLE is #defined in C/C++
header files.

MINGW : is TRUE when using the MinGW compiler in Windows

MSYS : is TRUE when using the MSYS developer environment in Windows

CYGWIN : is TRUE on Windows when using the CygWin version of cmake


NOTE that there is not a mention of MINGW on linux, however, if you are using mingw on linux then WIN32 will not be set and that is the reason for the :if (WIN32 OR MINGW) uses.

xantares:
no, MINGW is also set when crosscompiling from linux from my experience
in the doc it is said to be set as soon as the toolchain is mingw:
https://cmake.org/cmake/help/v3.18/variable/MINGW.html?highlight=mingw

byrnHDF:
> no, MINGW is also set when crosscompiling from linux from my experience
> in the doc it is said to be set as soon as the toolchain is mingw:
> https://cmake.org/cmake/help/v3.18/variable/MINGW.html?highlight=mingw

Right, but on linux it will not set "WIN32" so you need to check for "MINGW" as well.

xantares:
no, as the target platform is also windows, WIN32 is also set:
https://cmake.org/cmake/help/latest/variable/WIN32.html

xantares:
maybe you're confusing host platform and target platform ?

byrnHDF:
> maybe you're confusing host platform and target platform ?

maybe I am. Is it possible to have WIN32 not defined, yet have MINGW defined. If not then I agree with these changes. We will need to test.

xantares:
afaik the mingw toolchain can only target the windows platform so WIN32 would always be set

byrnHDF:
> afaik the mingw toolchain can only target the windows platform so WIN32 would always be set

okay, "if(WIN32 OR MINGW") == "if(WIN32)" - but now the tricky part - we need to differentiate between what is target settings vs build time settings - when building on linux, paths will need to be unix style (runtest.cmake). In this case we will need 
"if(WIN32 AND NOT MINGW)" checks.

byrnHDF:
> > afaik the mingw toolchain can only target the windows platform so WIN32 would always be set
> 
> okay, "if(WIN32 OR MINGW") == "if(WIN32)" - but now the tricky part - we need to differentiate between what is target settings vs build time settings - when building on linux, paths will need to be unix style (runtest.cmake). In this case we will need
> "if(WIN32 AND NOT MINGW)" checks.

Wait that's not right either, "if(MINGW AND NOT {HOST is WINDOWS})"

xantares:
In general there's no need to tweak paths because cmake uses the unix convention internally, except maybe when calling native tools.
I did run the tests and only a few failed, so I guess we're fine here.

byrnHDF:
> In general there's no need to tweak paths because cmake uses the unix convention internally, except maybe when calling native tools.
> I did run the tests and only a few failed, so I guess we're fine here.

That's good.
BTW, the variable is "CMAKE_HOST_SYSTEM_NAME".

I may need to setup a MINGW on Windows test machine - I haven't had much luck previously.

CMakeInstallation.cmake
```diff
@@ -290,7 +290,7 @@ if (NOT HDF5_EXTERNALLY_CONFIGURED AND NOT HDF5_NO_PACKAGES)
   set (CPACK_PACKAGE_ICON "${HDF_RESOURCES_EXT_DIR}/hdf.bmp")
 
   set (CPACK_GENERATOR "TGZ")
-  if (WIN32 OR MINGW)
+  if (WIN32)
```
byrnHDF, line 293: I think the "ZIP" packager is available on linux - so no issue here.

CMakeLists.txt
```diff
@@ -573,7 +573,7 @@ endif ()
 # and we need to call the executable from inside the CMake configuration
 #-----------------------------------------------------------------------------
 set (EXE_EXT "")
-if (WIN32 OR MINGW)
+if (WIN32)
```
byrnHDF, line 575: This needs checked - "if(MSVC)" might be preferable.

config/cmake/jrunTest.cmake
```diff
@@ -49,7 +49,7 @@ endif ()
 
 message (STATUS "COMMAND: ${TEST_TESTER} -Xmx1024M -Dorg.slf4j.simpleLogger.defaultLog=${LOG_LEVEL} -Djava.library.path=\"${TEST_LIBRARY_DIRECTORY}\" -cp \"${TEST_CLASSPATH}\" ${TEST_ARGS} ${TEST_PROGRAM} ${ARGN}")
 
-if (WIN32 OR MINGW)
+if (WIN32)
```
byrnHDF, line 52: Another check needed - if this is on a linux host is the ENV{PATH} valid?

config/cmake/jrunTest.cmake
```diff
@@ -131,7 +131,7 @@ if (NOT TEST_SKIP_COMPARE)
     file (READ ${TEST_FOLDER}/${TEST_REFERENCE} TEST_STREAM)
     list (LENGTH TEST_STREAM test_len)
     if (test_len GREATER 0)
-      if (WIN32 OR MINGW)
+      if (WIN32)
```
byrnHDF, line 134: This and the next are workarounds for tests on a windows host and might not be valid for a linux host.

config/cmake/jvolTest.cmake
```diff
@@ -54,7 +54,7 @@ message (STATUS "USING ${TEST_VOL} ON COMMAND: ${TEST_TESTER} -Xmx1024M -Dorg.sl
 
 set (ENV{HDF5_VOL_CONNECTOR} "${TEST_VOL}")
 
-if (WIN32 OR MINGW)
+if (WIN32)
```
byrnHDF, line 57: same comments as jrunTest.cmake

config/cmake_ext_mod/grepTest.cmake
```diff
@@ -46,7 +46,7 @@ endif ()
 message (STATUS "COMMAND: ${TEST_EMULATOR} ${TEST_PROGRAM} ${TEST_ARGS}")
 
 if (TEST_LIBRARY_DIRECTORY)
-  if (WIN32 OR MINGW)
+  if (WIN32)
```
byrnHDF, line 49: Same comments as jrunTest.cmake

config/cmake_ext_mod/runTest.cmake
```diff
@@ -38,7 +38,7 @@ endif ()
 message (STATUS "COMMAND: ${TEST_EMULATOR} ${TEST_PROGRAM} ${TEST_ARGS}")
 
 if (TEST_LIBRARY_DIRECTORY)
-  if (WIN32 OR MINGW)
+  if (WIN32)
```
byrnHDF, line 41: Same comments as jrunTest.cmake

byrnHDF (commented):
Looks like java and plugin tests will need investigation because of use of path and classpath changes.
Otherwise everything looks good.

CMakeLists.txt
xantares: yes, also EXE_EXT can be dropped

config/cmake/jrunTest.cmake
xantares: yes, it will be used by wine

config/cmake/jrunTest.cmake
xantares: if the issue is crlf, you will get the with wine, so this is fine

config/cmake_ext_mod/runTest.cmake
xantares: ok

byrnHDF (approved):
These changes look good, if at some time in the future this gets tested on a windows machine with mingw and issues are found - a new PR can be discussed.

lrknox (approved):

derobins (approved):

PR #72, opened 2020-11-03
Change H5R__copy to not assume destination reference storage is initi…

jhendersonHDF:
…alized

### DIFF
--- a/src/H5Rint.c
+++ b/src/H5Rint.c
@@ -746,8 +746,10 @@ H5R__copy(const H5R_ref_priv_t *src_ref, H5R_ref_priv_t *dst_ref)
         dst_ref->info.obj.filename = NULL;
 
         /* Set location ID and hold reference to it */
-        if (H5R__set_loc_id(dst_ref, src_ref->loc_id, TRUE, TRUE) < 0)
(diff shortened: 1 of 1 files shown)


src/H5Rint.c
```diff
@@ -746,8 +746,10 @@ H5R__copy(const H5R_ref_priv_t *src_ref, H5R_ref_priv_t *dst_ref)
         dst_ref->info.obj.filename = NULL;
 
         /* Set location ID and hold reference to it */
-        if (H5R__set_loc_id(dst_ref, src_ref->loc_id, TRUE, TRUE) < 0)
```
jhendersonHDF, line 749: H5R__set_loc_id checks the destination reference's loc_id field and decrements it if it is set to a valid ID before it assigns a new loc_id to the destination reference. After discussing this with @soumagne, it seems we should not be checking anything about the destination reference since its storage is probably uninitialized.

soumagne (approved):

qkoziol (approved):
Sensible - any other locations in the code that call H5R__set_loc_id() and have a similar problem?

jhendersonHDF:
> Sensible - any other locations in the code that call H5R__set_loc_id() and have a similar problem?

I don't believe so. All of the other places I could find look like the routine is called after the destination reference has just been created, which should be fine.

PR #76, opened 2020-11-05
Fix CMake package install dir on Windows not being considered by the search procedure

### LINKED ISSUE #75: CMake variable HDF5_INSTALL_CMAKE_DIR should be lib/cmake on Windows
On Windows, it looks like `HDF5_INSTALL_CMAKE_DIR` is set to `cmake` in https://github.com/HDFGroup/hdf5/blob/5b5a1a81029eb7bdc975beff5f18c9c440f5de56/config/cmake_ext_mod/HDFMacros.cmake#L359

Shouldn't this have the default value ` ${CMAKE_INSTALL_LIBDIR}/cmake` (given that the [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) module is imported) or just `lib/cmake`?

The issue is that the CMake package files (e.g. hdf5-config.cmake) ends up being installed in `%PREFIX%\cmake\hdf5` which is NOT a path considered by the [search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure). Not convenient as downstream projects have to define `HDF5_DIR` to workaround this issue.

Related to conda-forge/hdf5-feedstock#109

sdebionne:
Fixes #75 

It would be nice if it could be backported to the 1.10 branch as well.

### DIFF
--- a/config/cmake_ext_mod/HDFMacros.cmake
+++ b/config/cmake_ext_mod/HDFMacros.cmake
@@ -356,7 +356,7 @@ macro (HDF_DIR_PATHS package_prefix)
       set (${package_prefix}_INSTALL_CMAKE_DIR share/cmake)
     else ()
       set (${package_prefix}_INSTALL_DATA_DIR ".")
-      set (${package_prefix}_INSTALL_CMAKE_DIR cmake)
+      set (${package_prefix}_INSTALL_CMAKE_DIR lib/cmake)
     endif ()
   endif ()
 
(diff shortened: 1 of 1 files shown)


byrnHDF:
Pull Requests should be first created on the develop branch.
The normal sequence should be:
1. Create issue
2. Create PR on develop branch
3. Create PR for other branches after all tests (typically 2-3 days) are completed. These tests include testing on internal and external platforms.
4. Close issue.

PR #81, opened 2020-11-11
Fix non-collective metadata read in t_mdset with collective metadata reads enabled

jhendersonHDF:
Discovered with the DAOS VOL, where a mismatched MPI_Ibcast was causing weird results. All HDF5's parallel tests pass with or without this change, which is perhaps a bit concerning.

### DIFF
--- a/testpar/t_mdset.c
+++ b/testpar/t_mdset.c
@@ -1456,10 +1456,8 @@ read_attribute(hid_t obj_id, int this_type, int num)
     if (this_type == is_group) {
         HDsprintf(attr_name, "Group Attribute %d", num);
         aid = H5Aopen(obj_id, attr_name, H5P_DEFAULT);
-        if (MAINPROCESS) {
-            H5Aread(aid, H5T_NATIVE_INT, &in_num);
-            vrfy_errors = dataset_vrfy(NULL, NULL, NULL, group_block, &in_num, &num);
-        }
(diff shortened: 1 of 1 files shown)


testpar/t_mdset.c
```diff
@@ -1456,21 +1456,17 @@ read_attribute(hid_t obj_id, int this_type, int num)
     if (this_type == is_group) {
         HDsprintf(attr_name, "Group Attribute %d", num);
         aid = H5Aopen(obj_id, attr_name, H5P_DEFAULT);
-        if (MAINPROCESS) {
```
jhendersonHDF, line 1459: This routine is only used in a single test (multiple_group_read) in t_mdset.c. That test sets up a FAPL that has collective metadata reads enabled, but this does the attribute read on rank 0 only.

soumagne: could be a cache side-effect for native?

soumagne (approved):

testpar/t_mdset.c
jhendersonHDF: Maybe. I'm not sure why this hasn't seemed to be a problem in HDF5.

qkoziol:
Good catch!   It was almost certainly only working for the native connector because the attribute was embedded in the object header and was already cached.

qkoziol (approved):

PR #98, opened 2020-11-17
(a) Add async APIs for H5O module as listed in jira issue ID-283.

vchoi-hdfgroup:
(b) Remove verification of name parameter in async related routines for H55A and H5L modules
    because it is checked in H5VL_setup* routine.
(c) Modify h5dump expected output due to the async changes.

### DIFF
--- a/src/H5O.c
+++ b/src/H5O.c
@@ -68,7 +68,7 @@ static herr_t H5O__get_info_by_name_api_common(hid_t loc_id, const char *name,
     H5O_info2_t *oinfo /*out*/, unsigned fields, hid_t lapl_id, void **token_ptr, 
     H5VL_object_t ** _vol_obj_ptr);
 
-static herr_t H5O__close_api_common(hid_t object_id);
+static htri_t H5O__close_check_common(hid_t object_id);
 
 /*********************/
 /* Package Variables */
@@ -1373,18 +1373,19 @@ done:
 } /* end H5Ovisit_by_name3() */
 
 /*-------------------------------------------------------------------------
- * Function:    H5O__close_api_common
+ * Function:    H5O__close_check_common
  *
- * Purpose:     This is the common function for closing an object.
+ * Purpose:     This is the common function to validate an object
+ *              when closing it.
  *
- * Return:      SUCCEED/FAIL
+ * Return:      TRUE/FALSE/FAIL
  *
  *-------------------------------------------------------------------------
  */
-static herr_t
-H5O__close_api_common(hid_t object_id)
+static htri_t
+H5O__close_check_common(hid_t object_id)
 {
-    herr_t              ret_value = SUCCEED;    /* Return value */
+    htri_t ret_value = TRUE;    /* Return value */
 
     FUNC_ENTER_STATIC
 
@@ -1414,7 +1415,7 @@ H5O__close_api_common(hid_t object_id)
         case H5I_EVENTSET:
         case H5I_NTYPES:
         default:
-            HGOTO_ERROR(H5E_ARGS, H5E_CANTRELEASE, FAIL,
+            HGOTO_ERROR(H5E_ARGS, H5E_CANTRELEASE, FALSE,
                         "not a valid file object ID (dataset, group, or datatype)")
             break;
     } /* end switch */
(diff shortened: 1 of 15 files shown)


qkoziol:
Reviewing now

src/H5O.c
```diff
@@ -93,47 +149,100 @@
...
+    if (H5ES_NONE != es_id)
+        token_ptr = &token;     /* Point at token for VOL connector to set up */
+
+    /* Open the object asynchronously */
+    if((ret_value = H5O__open_api_common(loc_id, name, lapl_id, token_ptr, &vol_obj)) < 0)
+        HGOTO_ERROR(H5E_OHDR, H5E_CANTOPENOBJ, H5I_INVALID_HID, "unable to asynchronously open object")
+
+    /* If a token was created, add the token to the event set */
+    if (NULL != token)
+        if (H5ES_insert(es_id, vol_obj->connector, token,
```
qkoziol, line 199: There's some extra cleanup that is necessary on failure, can you please look at the changes to e.g. H5Dcreate_async, for the H5I_dec_app_ref_always_close() lines?

src/H5O.c
```diff
@@ -162,50 +271,60 @@ hid_t
...
-        HGOTO_ERROR(H5E_OHDR, H5E_CANTOPENOBJ, H5I_INVALID_HID, "unable to open object")
+    /* Open the object asynchronously */
+    if((ret_value = H5O__open_by_idx_api_common(loc_id, group_name, idx_type, order, n, lapl_id, token_ptr, &vol_obj)) < 0)
+        HGOTO_ERROR(H5E_OHDR, H5E_CANTOPENOBJ, H5I_INVALID_HID, "unable to asynchronously open object")
 
-    if ((ret_value = H5VL_register(opened_type, opened_obj, vol_obj->connector, TRUE)) < 0)
-        HGOTO_ERROR(H5E_OHDR, H5E_CANTREGISTER, H5I_INVALID_HID, "unable to atomize object handle")
+    /* If a token was created, add the token to the event set */
+    if (NULL != token)
+        if (H5ES_insert(es_id, vol_obj->connector, token,
```
qkoziol, line 326: Same here, with H5I_dec_app_ref_always_close, on failure.

src/H5O.c
```diff
@@ -554,6 +673,47 @@ H5Oget_info3(hid_t loc_id, H5O_info2_t *oinfo /*out*/, unsigned fields)
...
+    /* Check args */
+    if (!oinfo)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "oinfo parameter cannot be NULL")
+    if (fields & ~H5O_INFO_ALL)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "invalid fields")
+
+    /* "name" is checked in H5VL_setup_name_args() */
+    /* Set up object access arguments */
+    if (H5VL_setup_name_args(loc_id, name, H5P_CLS_LACC, FALSE, lapl_id, vol_obj_ptr, &loc_params) < 0)
+        HGOTO_ERROR(H5E_OHDR, H5E_CANTSET, H5I_INVALID_HID, "can't set object access arguments")
```
qkoziol: Should be FAIL on error.

src/H5O.c
```diff
@@ -1253,6 +1426,102 @@ H5Oclose(hid_t object_id)
...
+    H5VL_t *connector = NULL;       /* VOL connector */
+    void *token = NULL;             /* Request token for async operation        */
+    void **token_ptr = H5_REQUEST_NULL; /* Pointer to request token for async operation        */
+    herr_t ret_value = SUCCEED;
+
+    FUNC_ENTER_API(FAIL)
+    H5TRACE5("e", "*s*sIuii", app_file, app_func, app_line, object_id, es_id);
+
+    /* Get the type of the object and close it in the correct way */
+    switch (H5I_get_type(object_id)) {
```
qkoziol: Could this switch statement with the "is it OK to close this ID" check be moved into common code, for both H5Oclose and H5Oclose_async to use?

qkoziol: The other code here (and in H5Oclose) would be clearer, after that.

qkoziol:
Looks really good, just a few very small suggestions to update things.

src/H5O.c
```diff
@@ -1365,29 +1373,20 @@ H5Ovisit_by_name3(hid_t loc_id, const char *obj_name, H5_index_t idx_type, H5_it
...
  *
- * Programmer:	James Laird
- *		July 14 2006
- *
  *-------------------------------------------------------------------------
  */
-herr_t
-H5Oclose(hid_t object_id)
+static herr_t
+H5O__close_api_common(hid_t object_id)
```
qkoziol: Definitely the right idea, but maybe another name would be better since it's pretty different from the other api_common routines?  How about H5O__close_check and have it return an htri_t for yes / no / fail?   Then the H5Oclose / H5Oclose_async calls could call it and fail when they receive a no or fail return value.

qkoziol:
One small change that I would suggest.

qkoziol (approved):
Looks great!

PR #104, opened 2020-11-19
Revert "Switch ID code to use a hash table instead of a skip list (#52)"

derobins:
This reverts commit a50d211755cb272b2e468144e7d892a4c90813c4.

There's a memory issue with the hash tables that was masked by the free lists. Pulling this while I work on a fix.


lrknox (approved):

derobins:
As a revert, I don't think this needs the same level of scrutiny as a real PR.

PR #114, opened 2020-11-20
Minor/id code cleanup

derobins:
Cleans up H5I code:

- Consistent and improved naming
- Splits code into multiple files like other packages
- Quiets const warnings
- Cleans up includes


src/H5Itest.c
```diff
@@ -11,10 +11,8 @@
  * help@hdfgroup.org.                                                        *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
-/* Programmer:  Quincey Koziol
- *              Tuesday, July 27, 2010
- *
- * Purpose:	ID testing functions.
+/* 
+ * H5Iint.c - ID testing functions
```
qkoziol: H5Itest.c

qkoziol (approved):
There's some package-scoped variables that would be nice to resolve with package-scoped routines, but I can totally understand the decisions here and they are very reasonable.

src/H5Itest.c
derobins: Oops. Will fix before merge.

derobins:
> There's some package-scoped variables that would be nice to resolve with package-scoped routines, but I can totally understand the decisions here and they are very reasonable.

Yeah, I thought about wrapping those up but figured I'd save that work for later. It would certainly get done in the thread-safety work, for example.

lrknox (approved):

PR #115, opened 2020-11-22
Basic alignment with async branch

qkoziol:
Trivial changes to reduce clutter in overall diff.  Whitespace, alphabetization, code alignment, etc.


src/H5CX.c
```diff
@@ -1023,6 +1025,8 @@ H5CX_restore_state(const H5CX_state_t *api_state)
 
     /* Restore the VOL wrapper context */
     (*head)->ctx.vol_wrap_ctx = api_state->vol_wrap_ctx;
+    if (NULL != (*head)->ctx.vol_wrap_ctx)
```
qkoziol, line 1028: This is an actual bug fix.   (Might be the only actual code change of significance in this entire PR. ;-) )

src/H5Gname.c
```diff
@@ -355,44 +352,6 @@ H5G_build_fullpath_refstr_str(H5RS_str_t *prefix_r, const char *name)
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5G_build_fullpath_refstr_str() */
 
-#ifdef NOT_YET
```
qkoziol, line 358: If we haven't found a use for this code in 15 years, it can come out. :-)

derobins:
@byrnHDF - I feel like we need to change the formatter rules to put one space after ifdef, etc. ALWAYS. The formatter botches this all the time, often trying to align the preprocessor comments with other stuff that lies above.

derobins:
Is the formatter going to undo all these changes?

src/H5C.c
```diff
@@ -1310,7 +1310,7 @@ H5C_insert_entry(H5F_t *f, const H5C_class_t *type, haddr_t addr, void *thing, u
     hbool_t     flush_last;
 #ifdef H5_HAVE_PARALLEL
     hbool_t coll_access = FALSE; /* whether access to the cache entry is done collectively */
-#endif                           /* H5_HAVE_PARALLEL */
+#endif /* H5_HAVE_PARALLEL */
```
derobins, line 1313: The formatter did this. It will be undone as soon as someone re-runs it.

byrnHDF: I don't know which switch will give us what we want. However there are some easy (not perfect) solutions.
A blank line resets the comment/equal sign alignments.
Put the previous comment by itself on the line before statement instead of at the end of the line.

qkoziol:
> @byrnHDF - I feel like we need to change the formatter rules to put one space after ifdef, etc. ALWAYS. The formatter botches this all the time, often trying to align the preprocessor comments with other stuff that lies above.

These changes are from a run through the formatter, so maybe it's behaving better.

src/H5C.c
```diff
@@ -7275,9 +7275,9 @@ H5C__load_entry(H5F_t *f,
                                 (mpi_code = MPI_Bcast(image + len, buf_size, MPI_BYTE, 0, comm)))
                                 HMPI_GOTO_ERROR(NULL, "MPI_Bcast failed", mpi_code)
                         } /* end if */
-#endif                    /* H5_HAVE_PARALLEL */
-                    }     /* end if */
-                }         /* end if (actual_len != len) */
+#endif /* H5_HAVE_PARALLEL */
+                    } /* end if */
+                }     /* end if (actual_len != len) */
```
derobins, line 7280: Comments at the end of braces shouldn't be lined up. We'll need to fix the formatter to stop doing this.

src/H5CX.c
derobins: I feel like real bugfixes should be separated from giant noise-reduction PRs.

src/H5C.c
qkoziol: > Comments at the end of braces shouldn't be lined up. We'll need to fix the formatter to stop doing this.

Yeah, not my favorite either.   Fixing the formatter to avoiding doing this would be nice.

src/H5CX.c
qkoziol: > I feel like real bugfixes should be separated from giant noise-reduction PRs.

Yes, I agree and try to avoid it also.  I can remove it, if you think it's enough of a distraction.   (It only affects code that's for async operations, so won't be an issue to port back to versions before 1.10, if the overall patch doesn't get pushed back)

src/H5trace.c
```diff
@@ -3722,5 +3722,5 @@ H5_trace(const double *returning, const char *func, const char *type, ...)
     if (H5_debug_g.ttimes)
         return function_times.elapsed;
     else
-        return 0.0F;
+        return (double)0.0F;
```
derobins, line 3725: These casts of suffixed floats (cruft leftover from vainly trying to deal with gcc's insistence that we should use their non-standard D suffix) need to go. return 0.0; should be acceptable.

derobins (approved):

PR #120, opened 2020-11-25
Enhance API tracing

qkoziol:
Enhance API tracing to handle more types, and to put tracing info in a string, allowing it to be used when reporting errors.  Also refactored ref-counted strings (H5RS) module to add capabilities needed for the tracing.  Refactored H5Gname.c routines to use new H5RS routines also.  Added /*out*/ tags to API routines that are returning information to the application.  Updated H5TRACE macros from running updated trace script over library code.  Added tests for new H5RS routines.

### DIFF
--- a/src/H5Adeprec.c
+++ b/src/H5Adeprec.c
@@ -373,13 +373,13 @@ done:
     Deprecated in favor of H5Aiterate2
 --------------------------------------------------------------------------*/
 herr_t
-H5Aiterate1(hid_t loc_id, unsigned *attr_num, H5A_operator1_t op, void *op_data)
+H5Aiterate1(hid_t loc_id, unsigned *attr_num /*in,out*/, H5A_operator1_t op, void *op_data)
 {
     H5VL_object_t *vol_obj = NULL; /* Object of loc_id */
     herr_t         ret_value;      /* Return value */
 
     FUNC_ENTER_API(H5_ITER_ERROR)
-    H5TRACE4("e", "i*Iux*x", loc_id, attr_num, op, op_data);
+    H5TRACE4("e", "i*IuAo*x", loc_id, attr_num, op, op_data);
 
     /* check arguments */
     if (H5I_ATTR == H5I_get_type(loc_id))
--- a/src/H5RS.c
+++ b/src/H5RS.c
@@ -349,7 +349,7 @@ done:
  *      format_templ in the code below, but early (4.4.7, at least) gcc only
  *      allows diagnostic pragmas to be toggled outside of functions.
  */
-H5_GCC_DIAG_OFF(format - nonliteral)
+H5_GCC_DIAG_OFF("format-nonliteral")
 H5_ATTR_FORMAT(printf, 2, 3)
 herr_t
 H5RS_asprintf_cat(H5RS_str_t *rs, const char *fmt, ...)
@@ -392,7 +392,7 @@ H5RS_asprintf_cat(H5RS_str_t *rs, const char *fmt, ...)
 done:
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5RS_asprintf_cat() */
-H5_GCC_DIAG_ON(format - nonliteral)
+H5_GCC_DIAG_ON("format-nonliteral")
 
 /*-------------------------------------------------------------------------
  * Function:    H5RS_acat
--- a/src/H5RS.c
+++ b/src/H5RS.c
@@ -15,8 +15,15 @@
  * Reference counted string algorithms.
  *
  * These are used for various internal strings which get copied multiple times.
+ * They also efficiently handle dynamically allocations and appends.
  *
  */
+
+/****************/
+/* Module Setup */
+/****************/
+
+#include "H5RSmodule.h" /* This source code file is part of the H5RS module */
 
 /***********/
 /* Headers */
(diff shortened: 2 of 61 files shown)


qkoziol:
The focus of the changes here was primarily to allow sending the argument info from an API routine to another routine within the library.   The [not-yet-merged] event set routines use this argument information to add more context for asynchronous operations, particularly error reporting when one fails.   This new capability is set up with the new H5ARG_TRACE macro (as opposed to the existing H5TRACE macro, which is only used for printing the arguments of a routine directly to stdout), with the new H5_trace_args() using the tracing info to create a formatted string for the arguments.   Creating a formatted string with the tracing information required sending printf-like output to a string, but in a way that understood all the tracing types, etc.   So, I extended the H5RS package (ref-counted strings) with H5RS_asprintf() and H5RS_aputc() routines that provide the capabilities needed by H5_trace_args().

The remainder of the changes are just spin-offs from this primary set of changes.

qkoziol:

soumagne:
Could you please tell me where I should look at? I don't necessarily want to go through all the auto-generated trace changes :)

src/H5Adeprec.c
```diff
@@ -373,13 +373,13 @@ H5Aget_num_attrs(hid_t loc_id)
...
 herr_t
-H5Aiterate1(hid_t loc_id, unsigned *attr_num, H5A_operator1_t op, void *op_data)
+H5Aiterate1(hid_t loc_id, unsigned *attr_num /*in,out*/, H5A_operator1_t op, void *op_data)
 {
     H5VL_object_t *vol_obj = NULL; /* Object of loc_id */
     herr_t         ret_value;      /* Return value */
 
     FUNC_ENTER_API(H5_ITER_ERROR)
-    H5TRACE4("e", "i*Iux*x", loc_id, attr_num, op, op_data);
+    H5TRACE4("e", "i*IuAo*x", loc_id, attr_num, op, op_data);
```
bmribler, line 382: no need to change H5TRACE4 to 5?

src/H5RS.c
```diff
@@ -85,26 +106,146 @@ H5FL_BLK_DEFINE(str_buf);
...
+        size_t len = HDstrlen(s);
+
+        /* Determine size of buffer to allocate */
+        rs->max = H5RS_ALLOC_SIZE;
+        while ((len + 1) > rs->max)
+            rs->max *= 2;
+
+        /* Allocate the underlying string */
+        if (NULL == (rs->s = (char *)H5FL_BLK_MALLOC(str_buf, rs->max)))
+            HGOTO_ERROR(H5E_RS, H5E_CANTALLOC, FAIL, "memory allocation failed")
```
bmribler, line 129: I'm not clear when is H5E_RESOURCE and when is not?  H5E_RS here.

qkoziol:
> Could you please tell me where I should look at? I don't necessarily want to go through all the auto-generated trace changes :)

I would suggest looking mostly at the H5_trace() and H5_trace_args() calls in src/H5trace.c  (top and bottom of the file).   Also the new H5RS routines.

src/H5RS.c
qkoziol: > I'm not clear when is H5E_RESOURCE and when is not? H5E_RS here.

I suggest using H5E_RESOURCE in the actual resource allocation routines (like H5MM) and the current package (H5E_RS, here) in the package that calls the resource allocation routine.

bmribler: Ah, OK, thanks!

qkoziol:
No, there’s still only 4 parameters.

bmribler:
Yea, I realized that and deleted the comment right after.  I guess that wasn't good enough to hide my mistake. 🙂


qkoziol:
Ah, sorry!  I couldn’t figure out what happened, so I replied with an email.   (Now that I know what that situation looks like, I’ll just use the github comments as canonical)

	Quincey

bmribler:
Ah, so you saw the email but the comment disappeared on you!  And, for me, no matter how fast you delete it, the email will still point you out. 😉

Binh-Minh

src/H5RS.c
```diff
@@ -175,54 +317,224 @@ H5RS_wrap(const char *s)
...
+ *
+ *-------------------------------------------------------------------------
+ */
+/* Disable warning for "format not a string literal" here -QAK */
+/*
+ *      This pragma only needs to surround the sprintf() calls with
+ *      format_templ in the code below, but early (4.4.7, at least) gcc only
+ *      allows diagnostic pragmas to be toggled outside of functions.
+ */
+H5_GCC_DIAG_OFF(format - nonliteral)
```
derobins: You have to quote "format-nonliteral" or the formatter will mess it up, as it did here.

src/H5RS.c
derobins: I think H5E_RESOURCE should go away. Packages should just use their own major error code.

I hope to eventually refactor the error macros so that the package bit happens automatically and then that would always be the case.

derobins (approved):
I think this is fine. Just need to fix the gcc macros in H5RS.c.

lrknox (approved):

src/H5RS.c
qkoziol: > I think H5E_RESOURCE should go away. Packages should just use their own major error code.
> 
> I hope to eventually refactor the error macros so that the package bit happens automatically and then that would always be the case.

Agree - I think all the "generic" errors should go away and we should shift to package-based major error codes.   (And I would to also have them happen automatically, which the module headers could help with)

src/H5RS.c
qkoziol: Oops, I'll fix and update the PR.

lrknox (approved):

PR #121, opened 2020-11-25
Replace H5E_ATOM major error category with H5E_ID

derobins:
Converts H5E_ATOM to H5E_ID in the C library and HDF5AtomException to HDF5IdException in the Java wrappers. In several recent PRs, we've purged the archaic term "atom" from the library. This completes that work.

NOTE:
This PR will need changes when merged to the 1.10 and 1.12 branches so we don't modify the public interface (H5E_ATOM/H5E_ID, etc. will co-exist in those branches, with the former marked as deprecated for removal in 1.14/2.0). I will take care of those merges.


qkoziol (approved):

lrknox (approved):

src/H5VL.c
```diff
@@ -686,7 +686,7 @@ H5VLobject(hid_t id)
 hid_t
 H5VLget_file_type(void *file_obj, hid_t connector_id, hid_t dtype_id)
 {
-    H5T_t *        dtype;               /* unatomized type         */
+    H5T_t *        dtype;               /* unregisterd type         */
```
bmribler: typo

src/H5T.c
```diff
@@ -2997,7 +2997,7 @@ herr_t
 H5Tconvert(hid_t src_id, hid_t dst_id, size_t nelmts, void *buf, void *background, hid_t dxpl_id)
 {
     H5T_path_t *tpath;               /* type conversion info    */
-    H5T_t *     src, *dst;           /* unatomized types        */
+    H5T_t *     src, *dst;           /* unregisterd types        */
```
bmribler: typo

bmribler (approved):
Typo: unregisterd

src/H5VL.c
derobins: Fixed. Thanks!

src/H5T.c
derobins: Fixed. Thanks!

lrknox (approved):

byrnHDF (approved):

PR #122, opened 2020-11-25
Refactor to avoid using fork to create initial file

qkoziol:
It's not necessary to fork another process to create a file with only one process of an MPI job.

### DIFF
--- a/testpar/t_cache_image.c
+++ b/testpar/t_cache_image.c
@@ -1778,7 +1778,7 @@ par_insert_cache_image(int file_name_idx, int mpi_rank, int mpi_size)
             if (!serial_insert_cache_image(file_name_idx, mpi_size)) {
                 HDfprintf(stderr, "\n\nCache image insertion failed.\n");
                 HDfprintf(stderr, "  failure mssg = \"%s\"\n", failure_mssg);
-                HDexit(1);
+                HDexit(EXIT_FAILURE);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


testpar/t_cache_image.c
```diff
@@ -3937,50 +3657,28 @@ main(int argc, char **argv)
...
-                HDfprintf(stdout, "setup process reports failure.\n");
+            if (pass) {
+                HDprintf("done.\n");
+                HDfflush(stdout);
             }
             else {
-
-                HDfprintf(stdout, "testfile construction complete -- proceeding with tests.\n");
+                HDprintf("failed.\n");
+                HDexit(1);
```
derobins: Better to use EXIT_SUCCESS and EXIT_FAILURE

qkoziol: > Better to use EXIT_SUCCESS and EXIT_FAILURE

True, will update.

derobins (approved):

qkoziol:

byrnHDF:
So if I understand correctly, the fork was only used to create test files. This change then creates them inline before using them.

qkoziol:
> So if I understand correctly, the fork was only used to create test files. This change then creates them inline before using them.

Yes, forking was unnecessary.

lrknox (approved):

jhendersonHDF (approved):

PR #123, opened 2020-11-25
Update H5TS code to reflect library's coding style

qkoziol:
Use portability wrappers around pthread routines, and add some more FUNC_ENTER/EXIT macros that are very lightweight.

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -1259,6 +1259,57 @@ typedef off_t       h5_stat_size_t;
 #ifndef HDprintf
 #define HDprintf printf
 #endif /* HDprintf */
+#ifndef HDpthread_attr_destroy
+#define HDpthread_attr_destroy(A) pthread_attr_destroy(A)
+#endif /* HDpthread_attr_destroy */
+#ifndef HDpthread_attr_init
+#define HDpthread_attr_init(A) pthread_attr_init(A)
+#endif /* HDpthread_attr_init */
+#ifndef HDpthread_attr_setscope
+#define HDpthread_attr_setscope(A, S) pthread_attr_setscope(A, S)
+#endif /* HDpthread_attr_setscope */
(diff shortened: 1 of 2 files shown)


derobins (approved):

qkoziol:

src/H5private.h
```diff
@@ -2251,6 +2302,23 @@ H5_DLL herr_t H5CX_pop(void);
...
+#define FUNC_ENTER_API_NAMECHECK_ONLY                                                                        \
+    {                                                                                                        \
+        {                                                                                                    \
+            {                                                                                                \
+                {                                                                                            \
+                    {                                                                                        \
+                        {                                                                                    \
+                            FUNC_ENTER_COMMON_NOERR(H5_IS_API(FUNC));                                        \
+                            {
+
```
soumagne, line 2321: That's not pretty :)

qkoziol: > That's not pretty :)

Yeah, sometimes the code reformatted makes some ugly decisions. :-/

vchoi-hdfgroup (approved):

src/H5private.h
soumagne: Well I'm also wondering why you need so many "{" in the first place...

qkoziol: > Well I'm also wondering why you need so many "{" in the first place...

Each of the FUNC_ENTER / EXIT pairs needs to have a different # of brackets, so that they only match correctly, not matching a differently named ENTER / EXIT.   (Not my favorite method for enforcing the matching, so I'm open to other suggestions)

soumagne: Yeah I'm not sure, I feel like this would need some deeper cut probably. I don't really have the bandwidth right now to look more into it.

soumagne (approved):

PR #127, opened 2020-11-25
Minor refactoring in tid.c

derobins:
Small cleanup as part of the hash table ID work

### DIFF
--- a/test/tid.c
+++ b/test/tid.c
@@ -569,197 +569,299 @@ out:
 /* There was a rare bug where, if an id free callback being called by
  * H5I_clear_type() removed another id in that type, a segfault could occur.
  * This test tests for that error (and freeing ids "out of order" within
- * H5Iclear_type() in general). */
-/* Macro definitions */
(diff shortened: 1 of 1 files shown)


test/tid.c
```diff
@@ -569,197 +569,299 @@ test_id_type_list(void)
...
+    if (NULL == obj_list.objects)
+        goto error;
+
+    /* Set a convenience pointer to the object array */
+    objects = obj_list.objects;
+
+    for (i = 0; i < RCT_NITER; i++) {
+
+        /* The number of members in the type, according to the HDF5 library */
+        hsize_t nmembers = 1234567;
```
qkoziol, line 718: Initializing nmembers to 1234567 here was pretty confusing for me, until I realized that it was just an invalid value for the later call to H5Inmembers().   I would suggest moving the initialization to just before calling H5Inmembers(), to help the next person.

qkoziol (approved):
Nice cleanup!

lrknox (approved):

PR #129, opened 2020-11-25
Ensure that H5TSpublic.h header gets installed

jhendersonHDF:
Compile errors in the VOL tests when including hdf5.h otherwise.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -845,6 +845,7 @@ set (H5_PUBLIC_HEADERS
     ${H5S_HDRS}
     ${H5SM_HDRS}
     ${H5T_HDRS}
+    ${H5TS_HDRS}
     ${H5VL_HDRS}
     ${H5Z_HDRS}
 )


derobins (approved):

qkoziol (approved):
There isn't a H5TSpublic.h header in the develop branch yet.  But, one's on the way, so I can't see much harm in setting this up.

jhendersonHDF:
> There isn't a H5TSpublic.h header in the develop branch yet. But, one's on the way, so I can't see much harm in setting this up.

Sure, that's why I just made this PR against the async branch since I'm now using it for testing.

qkoziol:
> > There isn't a H5TSpublic.h header in the develop branch yet. But, one's on the way, so I can't see much harm in setting this up.
> 
> Sure, that's why I just made this PR against the async branch since I'm now using it for testing.

Oops!  I missed that this was against the async branch!  Thanks for getting it ready. :-)

PR #133, opened 2020-11-26
Move API routines for VOL objects to main source file.

qkoziol:
Helps to keep things consistent.   (Sometime in the future, we may want to consider moving all the API routines to more standard locations)


qkoziol:
(Attributes, groups, and named datatypes are already set up this way)

src/H5Dscatgath.c
```diff
@@ -370,7 +370,7 @@ H5D__gather_mem(const void *_buf, H5S_sel_iter_t *iter, size_t nelmts, void *_tg
     size_t         vec_size;           /* Vector length */
     size_t         ret_value = nelmts; /* Number of elements gathered */
 
-    FUNC_ENTER_STATIC
+    FUNC_ENTER_PACKAGE
```
qkoziol, line 373: This routine was already used at the package scope (and is already in H5Dpkg.h), I just updated the macro to reflect that.

derobins (approved):

qkoziol:
This PR needs one more review, from either: @soumagne @fortnern @jhendersonHDF @vchoi-hdfgroup or @jrmainzer    Please take a look, so I can move forward with other PRs.

soumagne (approved):

PR #134, opened 2020-11-26
Add flag to H5CX_pop that allow setting the properties to return to application to be skipped

qkoziol:
The properties only need to be set for API routines, not restoring the library state, or in tests.

### DIFF
--- a/src/H5Atest.c
+++ b/src/H5Atest.c
@@ -136,7 +136,7 @@ H5A__get_shared_rc_test(hid_t attr_id, hsize_t *ref_count)
         HGOTO_ERROR(H5E_ATTR, H5E_CANTGET, FAIL, "can't retrieve shared message ref count")
 
 done:
-    if (api_ctx_pushed && H5CX_pop() < 0)
+    if (api_ctx_pushed && H5CX_pop(FALSE) < 0)
         HDONE_ERROR(H5E_ATTR, H5E_CANTRESET, FAIL, "can't reset API context")
 
     FUNC_LEAVE_NOAPI(ret_value)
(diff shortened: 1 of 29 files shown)


qkoziol:
(This was a potential performance issue for the async VOL connector, which retrieves & restores the library's state)

bmribler (approved):

soumagne:
I guess that's probably fine for now modulo the fact that we should really start thinking about a better mechanism in the future. That's just getting too complex and out of hand in my opinion. We should not even have to do any distinction between app context or not. It's the same as for IDs, IDs should have remained public and _never_ get used internally. There's just such a mix of layers that it's almost impossible to understand what will happen in certain conditions...

soumagne (approved):

lrknox (approved):

qkoziol:
> I guess that's probably fine for now modulo the fact that we should really start thinking about a better mechanism in the future. That's just getting too complex and out of hand in my opinion. We should not even have to do any distinction between app context or not. It's the same as for IDs, IDs should have remained public and _never_ get used internally. There's just such a mix of layers that it's almost impossible to understand what will happen in certain conditions...

Sure - happy to participate in a conversation about ways to improve the API context idea.

derobins (approved):

byrnHDF (approved):

lrknox (approved):

qkoziol:
This PR needs one more review, from either: @derobins @rawarren  @fortnern @jhendersonHDF @byrnHDF  @vchoi-hdfgroup or @jrmainzer    Please take a look, so I can move forward with other PRs.

PR #135, opened 2020-11-27
Expand ID dec_ref and close callbacks to allow for asynchronous close operations

qkoziol:
There are no async close routines in develop yet, but this change allows for them to come in.


src/H5A.c
```diff
@@ -1456,12 +1456,14 @@ H5Aclose(hid_t attr_id)
     H5TRACE1("e", "i", attr_id);
 
     /* Check arguments */
-    if (NULL == H5I_object_verify(attr_id, H5I_ATTR))
-        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not an attribute")
+    if (H5I_ATTR != H5I_get_type(attr_id))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a attribute ID")
```
derobins: "not an attribute ID"

src/H5Ipublic.h
```diff
@@ -80,7 +80,7 @@ typedef int64_t hid_t;
  * can be removed from the ID type. If the function returns negative
  * (failure) then the object will remain in the ID type.
  */
-typedef herr_t (*H5I_free_t)(void *);
+typedef herr_t (*H5I_free_t)(void *, void **);
```
derobins, line 83: This is going to have to be handled like H5L_iterate_t and compatibility with 1.12.0 is going to be an issue.

derobins (changes requested):
The callback changes in H5Ipublic.h need to be discussed and addressed before this can be merged.

src/H5A.c
qkoziol: Fixed, thanks!

qkoziol:
> The callback changes in H5Ipublic.h need to be discussed and addressed before this can be merged.

Some subset of the developers have discussed it and I suggested that this callback was probably never used by outside developers, and that compatibility with 1.12.0 was already broken (with the VOL framework changes), so we would go with it.   If we want to revisit that decision, we can discuss at the async meeting on Monday afternoon.

derobins:
> > The callback changes in H5Ipublic.h need to be discussed and addressed before this can be merged.
> 
> Some subset of the developers have discussed it and I suggested that this callback was probably never used by outside developers, and that compatibility with 1.12.0 was already broken (with the VOL framework changes), so we would go with it. If we want to revisit that decision, we can discuss at the async meeting on Monday afternoon.

The binary compatibility people are going to scream, just like they did when we removed the probably-never-used HDFS VFD stubs in 1.10. We should consider making the next release 1.14.0 (or 2.0.0).

qkoziol:
> > > The callback changes in H5Ipublic.h need to be discussed and addressed before this can be merged.
> > 
> > 
> > Some subset of the developers have discussed it and I suggested that this callback was probably never used by outside developers, and that compatibility with 1.12.0 was already broken (with the VOL framework changes), so we would go with it. If we want to revisit that decision, we can discuss at the async meeting on Monday afternoon.
> 
> The binary compatibility people are going to scream, just like they did when we removed the probably-never-used HDFS VFD stubs in 1.10. We should consider making the next release 1.14.0 (or 2.0.0).

Yeah, they won't be happy.   Tough decision though - releasing 1.14.0 (or 2.0.0) so soon after 1.12.0 will also be upsetting to a different crowd of folks.    If it was just this callback, I could write a shim, but there are a number of VOL framework tweaks that also need to be made.  Let's kick ideas around at the meeting on Monday.

epourmal:
Yes, I agree. Let’s discuss on Monday.

I personally think we should bite a bullet and have tagged releases from develop.

Where are guaranties that no more VOL related changes will not be needed? Until we have a set of stable tested VOL connectors things will be in flux.

Elena

Sent from my iPhone

On Nov 29, 2020, at 4:27 PM, Quincey Koziol <notifications@github.com> wrote:

﻿

The callback changes in H5Ipublic.h need to be discussed and addressed before this can be merged.

Some subset of the developers have discussed it and I suggested that this callback was probably never used by outside developers, and that compatibility with 1.12.0 was already broken (with the VOL framework changes), so we would go with it. If we want to revisit that decision, we can discuss at the async meeting on Monday afternoon.

The binary compatibility people are going to scream, just like they did when we removed the probably-never-used HDFS VFD stubs in 1.10. We should consider making the next release 1.14.0 (or 2.0.0).

Yeah, they won't be happy. Tough decision though - releasing 1.14.0 (or 2.0.0) so soon after 1.12.0 will also be upsetting to a different crowd of folks. If it was just this callback, I could write a shim, but there are a number of VOL framework tweaks that also need to be made. Let's kick ideas around at the meeting on Monday.

derobins (approved):

qkoziol:
This PR needs two more reviews, from either: @soumagne @fortnern @jhendersonHDF @vchoi-hdfgroup or @jrmainzer    Please take a look, so I can move forward with other PRs.

src/H5Iint.c
```diff
@@ -77,6 +77,9 @@ typedef struct {
 static void * H5I__unwrap(void *object, H5I_type_t type);
 static htri_t H5I__clear_type_cb(void *_id, void *key, void *udata);
 static void * H5I__remove_common(H5I_type_info_t *type_info, hid_t id);
+static int    H5I__dec_ref(hid_t id, void **token);
+static int    H5I__dec_app_ref(hid_t id, void **token);
+static int    H5I__dec_app_ref_always_close(hid_t id, void **token);
```
soumagne: why are these named with a token? that's a request right? or I'm missing something?

src/H5Iint.c
```diff
@@ -887,26 +890,32 @@ H5I_remove(hid_t id)
...
+ * Note:        Allows for asynchronous 'close' operation on object, with
+ *              token != H5_REQUEST_NULL.
+ *
  * Return:      Success:    New reference count
  *              Failure:    -1
  *
+ * Programmer:  Houjun Tang
+ *              Oct 21, 2019
+ *
  *-------------------------------------------------------------------------
```
soumagne, line 907: Also nothing against Tang but you should really stop adding more of these programmer comments, it really does not make any sense. I could understand if that were a whole new module or something but for a single function like that it just adds more confusion that anything else.

soumagne: It would be better in my opinion that you add him as a co-author of your PR rather than adding names in the middle of the code like that. You can do that by adding: `Co-authored-by: name <name@example.com>` in your commit message.

src/H5Iint.c
qkoziol: Ah, yes, request would be better.  Fixed.

src/H5Iint.c
qkoziol: Sure, fixed.   (Will include Tang as co-author)

lrknox:
Ubuntu Debug test failed because H5Iint.c has multiple undefined H5E_ATOMs.
Lines 987, 1067, 1140.

lrknox (approved):

PR #136, opened 2020-11-27
Expand 'opt_query' VOL callback to return a bitfield instead of a flag

qkoziol:
Allows querying more information from a VOL connector.

### DIFF
--- a/src/H5VLnative_introspect.c
+++ b/src/H5VLnative_introspect.c
@@ -99,7 +99,7 @@ H5VL__native_introspect_opt_query(void H5_ATTR_UNUSED *obj, H5VL_subclass_t subc
                     break;
 
                 default:
-                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional attribute operation")
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional dataset operation")
                     break;
             } /* end switch */
             break;
@@ -156,7 +156,7 @@ H5VL__native_introspect_opt_query(void H5_ATTR_UNUSED *obj, H5VL_subclass_t subc
                     break;
 
                 default:
-                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional attribute operation")
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional file operation")
                     break;
             } /* end switch */
             break;
@@ -175,7 +175,7 @@ H5VL__native_introspect_opt_query(void H5_ATTR_UNUSED *obj, H5VL_subclass_t subc
 #endif /* H5_NO_DEPRECATED_SYMBOLS */
 
                 default:
-                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional attribute operation")
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional group operation")
                     break;
             } /* end switch */
             break;
--- a/src/H5VLnative_introspect.c
+++ b/src/H5VLnative_introspect.c
@@ -35,16 +35,192 @@
  *---------------------------------------------------------------------------
  */
 herr_t
-H5VL__native_introspect_opt_query(void H5_ATTR_UNUSED *obj, H5VL_subclass_t H5_ATTR_UNUSED cls,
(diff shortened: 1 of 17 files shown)


soumagne (approved):

derobins (approved):

byrnHDF (commented):
H5Oget_comment is used in Java wrappers, but H5VLquery_optionalH5VLquery_optional is not implemented. Oversight when H5VLquery_optional was added but not to the wrappers?

qkoziol:
> H5Oget_comment is used in Java wrappers, but H5VLquery_optionalH5VLquery_optional is not implemented. Oversight when H5VLquery_optional was added but not to the wrappers?

Yes, that seems to have been missed with the original H5VLquery_optional checking.  I'll see if I can update all the language wrappers (FORTRAN, C++, and Java) to check with H5VLquery_optional before calling H5Oget_comment (same as the tools do).

src/H5VLnative_introspect.c
```diff
@@ -35,16 +35,192 @@
...
+                case H5VL_NATIVE_DATASET_CHUNK_READ:
+                    *flags |= H5VL_OPT_QUERY_READ_DATA;
+                    break;
+
+                case H5VL_NATIVE_DATASET_CHUNK_WRITE:
+                    *flags |= H5VL_OPT_QUERY_WRITE_DATA;
+                    break;
+
+                default:
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional attribute operation")
```
bmribler: Should this be dataset operation?

src/H5VLnative_introspect.c
```diff
@@ -35,16 +35,192 @@
...
+                case H5VL_NATIVE_FILE_SET_LIBVER_BOUNDS:
+                case H5VL_NATIVE_FILE_GET_MIN_DSET_OHDR_FLAG:
+                case H5VL_NATIVE_FILE_SET_MIN_DSET_OHDR_FLAG:
+                case H5VL_NATIVE_FILE_GET_MPI_ATOMICITY:
+                case H5VL_NATIVE_FILE_SET_MPI_ATOMICITY:
+                case H5VL_NATIVE_FILE_POST_OPEN:
+                    break;
+
+                default:
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional attribute operation")
```
bmribler: ...and this

src/H5VLnative_introspect.c
```diff
@@ -35,16 +35,192 @@
...
+                    *flags |= H5VL_OPT_QUERY_NO_ASYNC;
+                    break;
+
+                case H5VL_NATIVE_GROUP_GET_OBJINFO:
+                    *flags |= H5VL_OPT_QUERY_QUERY_METADATA;
+                    break;
+#endif /* H5_NO_DEPRECATED_SYMBOLS */
+
+                default:
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unknown optional attribute operation")
```
bmribler: ...and this

bmribler (approved):
Please see typo comments.

src/H5VLpublic.h
```diff
@@ -49,6 +49,25 @@
...
+/* Flags to return from H5VLquery_optional API and 'opt_query' callbacks */
+/* Note: Operations which access multiple objects' data or metadata in a
+ *      container should be registered as file-level optional operations.
+ *      (e.g. "H5Dwrite_multi" takes a list of datasets to write data to, so
+ *      a VOL connector that implemented it should register it as an optional
+ *      file operation, and pass-through VOL connectors that are stacked above
+ *      the connector that registered it should assume that dataset elements
+ *      for _any_ dataset in the file could be written to)
+ */
+#define H5VL_OPT_QUERY_SUPPORTED       0x0001 /* VOL connector supports this operation */
```
byrnHDF, line 61: When wrappers are created, these may need to be added - at least to Java constants.

byrnHDF (approved):
Expect an issue for wrappers to be created.

qkoziol:
> Please see typo comments.

Got 'em, thanks!

qkoziol:
> Expect an issue for wrappers to be created.

I gave the wrapper situation a little more thought and think that the language wrappers are working correctly.   The language wrappers should just invoke the optional operation and allow it to succeed or fail, same as if an application invoked it directly.  The tools are different because they are trying to display information and they are the application, in this case.   So, I don't think any further changes are needed for the language wrappers and don't plan to update them.

lrknox (approved):

PR #138, opened 2020-11-29
Avoid deadlock when determining chunked dataset ownership in parallel

qkoziol:
Break single loop with interleaved sends and receives into two separate loops, to avoid deadlocks.  Also, avoid copying chunk entries when determining ownership.


qkoziol:
This fixes the hang for the testphdf5 regression test on the Mac, and possibly other platforms.

@lrknox - Are there any other platforms that were hanging in the testphdf5 test?

qkoziol:
@lrknox @byrnHDF - Are any of the github CI checks running parallel builds & tests?

derobins (approved):

qkoziol:
This PR needs two more reviews, from either: @soumagne @rawarren @fortnern @jhendersonHDF @vchoi-hdfgroup or @jrmainzer    Please take a look, so I can move forward with other PRs.

lrknox (approved):

PR #139, opened 2020-11-29
Add H5atclose and H5is_library_terminating routines.

qkoziol:
Adds two new API routines that provide useful infrastructure for third-party packages that wrap or extend HDF5.

### DIFF
--- a/src/H5.c
+++ b/src/H5.c
@@ -44,6 +44,13 @@
 /* Package Typedefs */
 /********************/
 
+/* Node for list of 'atclose' routines to invoke at library shutdown */
+typedef struct H5_atclose_node_t {
+    H5_atclose_func_t         func; /* Function to invoke */
+    void *                    ctx;  /* Context to pass to function */
+    struct H5_atclose_node_t *next; /* Pointer to next node in list */
+} H5_atclose_node_t;
+
 /********************/
 /* Local Prototypes */
 /********************/
@@ -87,6 +94,12 @@ H5_debug_t     H5_debug_g; /* debugging info */
 /*******************/
 /* Local Variables */
 /*******************/
+
+/* Linked list of registered 'atclose' functions to invoke at library shutdown */
+static H5_atclose_node_t *H5_atclose_head = NULL;
+
+/* Declare a free list to manage the H5_atclose_node_t struct */
+H5FL_DEFINE_STATIC(H5_atclose_node_t);
 
 /*--------------------------------------------------------------------------
 NAME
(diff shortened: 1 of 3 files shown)


qkoziol:
RFC for these routines here: https://www.overleaf.com/read/jmmpfskybqdx

qkoziol:
Figured out how to attach the RFC PDF:

[RFC_H5atclose_and_H5is_library_terminating.pdf](https://github.com/HDFGroup/hdf5/files/5611698/RFC_H5atclose_and_H5is_library_terminating.pdf)

qkoziol:
Updated RFC with copy edits:

[RFC_H5atclose_and_H5is_library_terminating-2.pdf](https://github.com/HDFGroup/hdf5/files/5611709/RFC_H5atclose_and_H5is_library_terminating-2.pdf)

src/H5.c
```diff
@@ -1124,6 +1198,34 @@ H5is_library_threadsafe(hbool_t *is_ts)
...
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5is_library_terminating(hbool_t *is_terminating)
+{
+    FUNC_ENTER_API_NOINIT
+    H5TRACE1("e", "*b", is_terminating);
+
+    HDassert(is_terminating);
```
derobins, line 1226: Should use real error checking and not assert() in public API calls.

src/H5public.h
```diff
@@ -386,6 +386,9 @@ typedef struct H5_alloc_stats_t {
     size_t             peak_alloc_blocks_count;  /**< Peak # of blocks allocated */
 } H5_alloc_stats_t;
 
+/* Library shutdown callback */
+typedef void (*H5_atclose_func_t)(void *ctx);
```
derobins, line 392: Do typedefs get doxygen comments?

src/H5.c
```diff
@@ -88,6 +95,12 @@ H5_debug_t     H5_debug_g; /* debugging info */
 /* Local Variables */
 /*******************/
 
+/* Linked list of registered 'atclose' functions to invoke at library shutdown */
+static H5_atclose_node_t *H5_atclose_head = NULL;
+
+/* Declare a free list to manage the H5_atclose_node_t struct */
+H5FL_DEFINE_STATIC(H5_atclose_node_t);
```
derobins, line 102: I feel like using a free list here is overkill. I seriously doubt there's going to be a lot of memory churn on close callbacks.

derobins:
I would approve this, but I feel like the assert-based error check needs to be fixed first.

src/H5.c
qkoziol: That's not possible here, since the library may not be initialized.

src/H5public.h
qkoziol: Seems like it (based on H5Fpublic.h).   Added.

qkoziol:
Updated RFC with formatting and funding acknowledgement:

[RFC_H5atclose_and_H5is_library_terminating-3.pdf](https://github.com/HDFGroup/hdf5/files/5612364/RFC_H5atclose_and_H5is_library_terminating-3.pdf)

src/H5.c
derobins: But you can still check the parameter and jump to the end returning FAIL if the parameter is NULL.

qkoziol: OK, sure - I've changed the code to do that for both H5is_library_terminating as well as the existing H5is_library_threadsafe routine.

derobins (approved):

qkoziol:
This PR needs one more review, from either: @soumagne @fortnern @jhendersonHDF @vchoi-hdfgroup or @jrmainzer    Please take a look, so I can move forward with other PRs.

qkoziol:
This is the last of my outstanding PRs, can I please get another review, so I can clear my plate and start on the next set?

src/H5.c
soumagne: also tend to agree with Dana

soumagne (approved):

PR #142, opened 2020-11-29
FreeBSD autotools changes

derobins:
Updates the freebsd file in config/ to work on FreeBSD 10+ where clang is the default compiler.

### DIFF
--- a/config/freebsd
+++ b/config/freebsd
@@ -18,19 +18,26 @@
 #
 # See BlankForm in this directory for details.
 
-# The default compiler is `gcc'
+# The default compiler is 'cc'
 if test "X-" = "X-$CC"; then
-  CC=gcc
-  CC_BASENAME=gcc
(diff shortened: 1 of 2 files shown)


qkoziol (approved):

byrnHDF (approved):

config/freebsd
```diff
@@ -18,19 +18,26 @@
 #
 # See BlankForm in this directory for details.
 
-# The default compiler is `gcc'
+# The default compiler is 'cc'
 if test "X-" = "X-$CC"; then
-  CC=gcc
-  CC_BASENAME=gcc
+  CC=cc
+  CC_BASENAME=cc
```
gnuoyd, line 24: This just changes the command configure uses to detect the brand of compiler—gcc, clang, Intel—right? Looks fine to me.

PR #143, opened 2020-11-30
(fix) Segmentation fault when using a compound type.

jwsblokland:
In the case when a compounded attribute is written to dataset
followed by writing the data with a data transform function
to the dataset will result in a segmentation fault. It turns out
the data is classified as compounded while it is not. Now, the
state is always reset first to not compounded followed by the
existing check if the variable is compounded.

An example of the problem and to show it is solved by this pull request: [compound.txt](https://github.com/HDFGroup/hdf5/files/5614090/compound.txt)

This pull request solves the items SUPPORT-857 and HDFFV-11050

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -5218,8 +5218,6 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
 
     /* Set the flag to indicate both source and destination types are compound types
      * for the optimization of data reading (in H5Dio.c).
-     *
-     * Jan-Willem Blokland, 11 March 2020
      * Make sure that path->are_compounds is only TRUE for compound types.
      */
     path->are_compounds = FALSE;
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -3405,8 +3405,8 @@ done:
  * Note:      Common code for both H5T_copy and H5T_copy_reopen, as part of
  *            the const-correct datatype copying routines.
  *
- * Programmer:	David Young
- *	        January 18, 2020
+ * Programmer:  David Young
+ *              January 18, 2020
  *
  *-------------------------------------------------------------------------
  */
@@ -3453,8 +3453,8 @@ done:
  * Return:    Success:    Pointer to a new copy of the OLD_DT argument.
  *            Failure:    NULL
  *
- * Programmer:	David Young
- *	        January 18, 2020
+ * Programmer:  David Young
+ *              January 18, 2020
  *
  *-------------------------------------------------------------------------
  */
@@ -3481,8 +3481,8 @@ done:
  * Return:    Success:    Pointer to a new copy of the OLD_DT argument.
  *            Failure:    NULL
  *
- * Programmer:	David Young
- *	        January 18, 2020
+ * Programmer:  David Young
+ *              January 18, 2020
  *
  *-------------------------------------------------------------------------
  */
@@ -3512,8 +3512,8 @@ done:
  *
  * Note:      Common code for both H5T_copy and H5T_copy_reopen.
  *
- * Programmer:	David Young
- *	        January 18, 2020
+ * Programmer:  David Young
+ *              January 18, 2020
  *
  *-------------------------------------------------------------------------
  */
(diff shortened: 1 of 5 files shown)


src/H5T.c
```diff
@@ -5217,7 +5217,12 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
     } /* end else-if */
 
     /* Set the flag to indicate both source and destination types are compound types
-     * for the optimization of data reading (in H5Dio.c). */
+     * for the optimization of data reading (in H5Dio.c).
+     *
+     * Jan-Willem Blokland, 11 March 2020
```
derobins: Please remove this line. It's going to be really distracting if the code fills up with line-by-line credit for edits.

Also, feel free to remove Quincey's similar edit, above :)

derobins:
This should get a JIRA issue and needs a test demonstrating the problem and ensuring it's fixed. Do you have a minimal test program that fails? If you can send me a test program that confirms the bug, I'll create a JIRA issue. I can also point you to where you should add a test and explain our test harness a bit, but that will have to wait until tomorrow morning.

src/H5T.c
jwsblokland: Good point. I will remove it.

jwsblokland:
The included example compound.txt is an minimal test program. It is c source file but github does not allow me to upload a file with the extension .c. 

Yes if you can explain your test setup, I will add a test for this case.

Furthermore, for your convenience I added the existing JIRA issue numbers SUPPORT-857 and HDFFV-11050 in which the problem is described.

qkoziol:
> The included example compound.txt is an minimal test program. It is c source file but github does not allow me to upload a file with the extension .c.
> 
> Yes if you can explain your test setup, I will add a test for this case.
> 
> Furthermore, for your convenience I added the existing JIRA issue numbers SUPPORT-857 and HDFFV-11050 in which the problem is described.

Yes, a test would be needed for this.

jwsblokland:
Added a new test to the test suite to demonstrate the segmentation fault is resolved.

qkoziol (changes requested):
Looks good, but can you please add the new test to the autotools framework also?

test/cmpd_dtransform.c
```diff
@@ -0,0 +1,139 @@
...
+ *                                                                           *
+ * This file is part of HDF5.  The full HDF5 copyright notice, including     *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the files COPYING and Copyright.html.  COPYING can be found at the root   *
+ * of the source code distribution tree; Copyright.html can be found at the  *
+ * root level of an installed copy of the electronic HDF5 document set and   *
+ * is linked from the top-level documents page.  It can also be found at     *
+ * http://hdfgroup.org/HDF5/doc/Copyright.html.  If you do not have          *
+ * access to either file, you may request a copy from help@hdfgroup.org.     *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
```
lrknox, line 11: The HDF Group copyright header has changed, and the line "* Copyright by the Board of Trustees of the University of Illinois." applies only to files created before 2007.
The text in the paragraph has changed and the link that is now for the COPYING file also needs updating in all of the HDF5 source files, which will be happening within a few days, so it isn't necessary to make changes for this pull request, especiallly since pasting it here made format changes.  Once that update is done, this header will be replaced with the following:
/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *  
 * Copyright by The HDF Group.                                                                 *
 * All rights reserved.                                                                                   *
 
 * This file is part of HDF5.  The full HDF5 copyright notice, including       *
 * terms governing use, modification, and redistribution, is contained in  *
 * the COPYING file, which can be found at the root of the source code   *
 * distribution tree, or in https://www.hdfgroup.org/licenses.                    *
 * If you do not have access to either file, you may request a copy from   *
 * help@hdfgroup.org.                                                                                 *

jwsblokland: I have updated the source header to the latest version.

jwsblokland:
The new test has been added to the autotools.

qkoziol:
> The new test has been added to the autotools.

Super, thanks!

qkoziol (approved):

test/cmpd_dtransform.c
lrknox: Thanks!

lrknox (approved):

PR #145, opened 2020-11-30
Fixed HDFFV-10480 and HDFFV-11159

bmribler:
Description
    Checked against buffer size to prevent segfault, in case of data corruption.

    + HDFFV-11159 CVE-2018-14033 Buffer over-read in H5O_layout_decode
    + HDFFV-10480 CVE-2018-11206 Buffer over-read in H5O_fill_new[/old]_decode and
        A user's patch was applied to this previously, but it is redone
        for a more correct fix, that is the check now accounted for the
        previous advance of the buffer pointer.

Platforms tested:
    Linux/64 (jelly)

### DIFF
--- a/src/H5Ofill.c
+++ b/src/H5Ofill.c
@@ -222,7 +222,7 @@ H5O_fill_new_decode(H5F_t H5_ATTR_UNUSED *f, hid_t H5_ATTR_UNUSED dxpl_id, H5O_t
 
                 /* Ensure that fill size doesn't exceed buffer size, due to possible data corruption */
                 if (p + fill->size - 1 > p_end)
-                    HGOTO_ERROR(H5E_RESOURCE, H5E_OVERFLOW, NULL, "fill size exceeds buffer size")
+                    HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "fill size exceeds buffer size")
 
                 if (NULL == (fill->buf = H5MM_malloc((size_t)fill->size)))
                     HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL, "memory allocation failed for fill value")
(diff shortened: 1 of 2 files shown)


qkoziol:
Is there a test for this?

bmribler:
> Is there a test for this?

No, I don't have the files.  I could create a file, figure out where the value is and mess with it to cause the segfault, similar to what the CVE reporter did. Does that justify the time spent?

src/H5Ofill.c
```diff
@@ -218,6 +219,11 @@ H5O_fill_new_decode(H5F_t H5_ATTR_UNUSED *f, hid_t H5_ATTR_UNUSED dxpl_id, H5O_t
             INT32DECODE(p, fill->size);
             if (fill->size > 0) {
                 H5_CHECK_OVERFLOW(fill->size, ssize_t, size_t);
+
+                /* Ensure that fill size doesn't exceed buffer size, due to possible data corruption */
+                if (p + fill->size - 1 > p_end)
+                    HGOTO_ERROR(H5E_RESOURCE, H5E_OVERFLOW, NULL, "fill size exceeds buffer size")
```
derobins: These errors should be H5E_OHDR, not H5E_RESOURCE.

qkoziol:
> > Is there a test for this?
> 
> No, I don't have the files. I could create a file, figure out where the value is and mess with it to cause the segfault, similar to what the CVE reporter did. Does that justify the time spent?

I think it would, if we want to be certain that the bug is fixed and stays fixed. :-)

src/H5Ofill.c
bmribler: I fixed, pushed, but haven't seen it...

bmribler: Thanks.  Fixed.

bmribler:
> > > Is there a test for this?
> > 
> > 
> > No, I don't have the files. I could create a file, figure out where the value is and mess with it to cause the segfault, similar to what the CVE reporter did. Does that justify the time spent?
> 
> I think it would, if we want to be certain that the bug is fixed and stays fixed. :-)

OK

derobins (approved):

lrknox (approved):

PR #150, opened 2020-12-09
Added selective async APIs

bmribler:
Description:
    Added the following APIs:
        H5Ropen_attr_async
        H5Ropen_object_async
        H5Ropen_region_async

        H5Mcreate_async
        H5Mopen_async
        H5Mput_async
        H5Mget_async
        H5Mclose_async

        H5Tcommit_async
        H5Topen_async
        H5Tcopy_async
        H5Tclose_async

    - Updated an expected output file to include a new internal function
      in the error stack for the failure case.


byrnHDF (approved):

src/H5M.c
```diff
@@ -461,6 +612,61 @@ H5Mclose(hid_t map_id)
...
+     */
+    if (H5I_dec_app_ref_always_close_async(map_id, token_ptr) < 0)
+        HGOTO_ERROR(H5E_MAP, H5E_CANTDEC, FAIL, "can't decrement count on dataset ID")
+
+    /* If a token was created, add the token to the event set */
+    if (NULL != token)
+        if (H5ES_insert(es_id, vol_obj->connector, token, H5ARG_TRACE5(FUNC, "*s*sIuii", app_file, app_func, app_line, map_id, es_id)) < 0)
+            HGOTO_ERROR(H5E_MAP, H5E_CANTINSERT, FAIL, "can't insert token into event set")
+
+done:
```
qkoziol, line 669: Need the "decrement the connector refcount" codeblock here.  (See e.g. H5Dclose_async)

src/H5T.c
```diff
@@ -1949,18 +1946,103 @@ H5Tcopy(hid_t obj_id)
...
+    /* Copy the object's datatype synchronously */
+    if ((ret_value = H5T__copy_api_common(obj_id, NULL, NULL)) < 0)
+        HGOTO_ERROR(H5E_DATATYPE, H5E_CANTGET, H5I_INVALID_HID, "unable to copy datatype synchronously")
+
+done:
     FUNC_LEAVE_API(ret_value)
 } /* end H5Tcopy() */
 
+/*-------------------------------------------------------------------------
+ * Function:    H5Tcopy_async
```
qkoziol: I'm still not entirely convinced we need an async version of H5Tcopy, since it's not _really_ an operation that deals with the file.  I see that it _could_ trigger a VOL callback, but I would defer delivering this to users until later, since I think that adding an asynchronous version of an operation that isn't really about VOL object could be a slippery slope into "asynchronous operations for *all* API routines", which is a Pandora's Box that I'd rather open later. :-)   So, I would suggest reverting the changes to H5Tcopy and just let it execute synchronously for situations where it triggered the VOL callback.  (We can always add it later)

qkoziol (changes requested):
Two suggested changes, but otherwise looks great!

src/H5T.c
bmribler: OK, put it simply, take everything new related to H5Tcopy out and put original H5Tcopy back, right?

qkoziol: Yes :-)

src/H5M.c
bmribler: I have no idea what happened.  For every file, a couple of them were missing.  I'm very sure, I checked and double checked...  I added all now, and am compiling.  I'll commit them soon.  Thank you!

qkoziol: Sorry, I wasn't specific enough with my last comment - the H5I_dec_ref_always_close() that you just put in shouldn't be there, but this routine needs the:
   if (connector && H5VL_conn_dec_rc(connector) < 0)
         HDONE_ERROR(H5E_DATATYPE, H5E_CANTDEC, FAIL, "can't decrement ref count on connector")
block right after the 'done:' label.

qkoziol (approved):
Looks good, thanks!

PR #151, opened 2020-12-10
VOL framework version compatibility

qkoziol:
Enforce VOL framework version compatibility when registering connectors.  Also add a version for the connector itself, some refactoring on the register calls, and move the logic for matching/rejecting a VOL connector class from the plugin module to the VOL module.

### DIFF
--- a/src/H5VLconnector.h
+++ b/src/H5VLconnector.h
@@ -445,9 +445,10 @@ typedef struct H5VL_token_class_t {
 //! [H5VL_class_t_snip]
 typedef struct H5VL_class_t {
     /* Overall connector fields & callbacks */
-    unsigned int       version;          /**< VOL connector class struct version #     */
+    unsigned           version;          /**< VOL connector class struct version #     */
     H5VL_class_value_t value;            /**< Value to identify connector              */
     const char *       name;             /**< Connector name (MUST be unique!)         */
+    unsigned           conn_version;     /**< Version # of connector                   */
     unsigned           cap_flags;        /**< Capability flags for connector           */
     herr_t (*initialize)(hid_t vipl_id); /**< Connector initialization callback        */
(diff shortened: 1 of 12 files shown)


qkoziol:
I'll leave out the 'conn_version', and set the H5VL_VERSION to 0 when merging this to the 1.12 branch.   And set the H5VL_VERSION to 2 when merging to the async branch.

src/H5VLconnector.h
```diff
@@ -445,9 +445,10 @@ typedef struct H5VL_token_class_t {
 //! [H5VL_class_t_snip]
 typedef struct H5VL_class_t {
     /* Overall connector fields & callbacks */
-    unsigned int       version;          /**< VOL connector class struct version #     */
+    unsigned           version;          /**< VOL connector class struct version #     */
     H5VL_class_value_t value;            /**< Value to identify connector              */
     const char *       name;             /**< Connector name (MUST be unique!)         */
+    unsigned           conn_version;     /**< Version # of connector                   */
```
soumagne, line 451: Can you remind me why we need an extra field? I feel like you should add a little more explanation in the header then since I could see people would get confused.

soumagne: also if we have that conn_version field I guess it could be helpful to be able to query it / print it from the HDF5 public API. It would be an easy way of doing sanity checks to make sure the right version of a plugin is loaded.

soumagne (approved):

src/H5VLconnector.h
qkoziol: We chatted about adding the connector version field to the struct in the async channel this morning.   The library ignores it currently, so it's merely informative, but another API to get info about a connector would be good in the future, maybe even a small command-line tool.

fortnern (approved):

PR #158, opened 2020-12-11
Minor cleanup in async branch in preparation for merge to develop

derobins:


qkoziol (approved):
These all look good, but I've changed the H5ES* routines quite a bit, so if you could wait until after the review at 3pm today before merging them in, that would be very helpful.

derobins:
> These all look good, but I've changed the H5ES* routines quite a bit, so if you could wait until after the review at 3pm today before merging them in, that would be very helpful.

Let me know when you are ready and I'll pull whatever changes you've made into my branch and tidy up. Will you want to click the merge button or should I?

PR #161, opened 2020-12-11
Combo set of async and other changes

qkoziol:
Draft PR for all code to merge to develop, includes async, dynamic optional VOL operation registration, and H5S_BLOCK code.


qkoziol:
This is the "combo" version of the async changes, plus the dynamic optional VOL operations, etc.   It's ready to merge after the base async branch is merged to develop.

qkoziol:
I've re-synced this branch with develop, after the async branch was merged in.  This should be ready to merge also now, after develop is tagged for the DAOS team.

src/H5ESpublic.h
```diff
@@ -42,11 +42,26 @@
...
+    /* Operation info */
+    char *   api_name;      /* Name of HDF5 API routine called */
+    char *   api_args;      /* "Argument string" for arguments to HDF5 API routine called */
+    char *   app_file_name; /* Name of source file where the HDF5 API routine was called */
+    char *   app_func_name; /* Name of function where the HDF5 API routine was called */
+    unsigned app_line_num;  /* Line # of source file where the HDF5 API routine was called */
+    uint64_t op_ins_count;  /* Counter of operation's insertion into event set */
+    uint64_t op_ins_ts;     /* Timestamp for when the operation was inserted into the event set */
+    uint64_t op_exec_ts;    /* Timestamp for when the operation began execution */
+    uint64_t op_exec_time;  /* Execution time for operation */
```
gnuoyd: Units?

src/H5ESint.c
```diff
@@ -232,6 +249,78 @@ H5ES__create(void)
...
+ *
+ * Purpose:     Insert a request token into an event set
+ *
+ * Return:      SUCCEED / FAIL
+ *
+ * Programmer:	Quincey Koziol
+ *	        Friday, December 11, 2020
+ *
+ *-------------------------------------------------------------------------
+ */
```
gnuoyd, line 258: You have spent 10 lines "documenting" H5ES__insert without telling me the meaning of a single argument. I can infer the return values from the return type, `herr_t`.

src/H5ESint.c
```diff
@@ -232,6 +249,78 @@ H5ES__create(void)
...
+
+    /* Sanity check */
+    HDassert(es);
+
+    /* Create new event */
+    if (NULL == (ev = H5ES__event_new(connector, request_token)))
+        HGOTO_ERROR(H5E_EVENTSET, H5E_CANTCREATE, FAIL, "can't create event object")
+
+    /* Copy the app source information */
+    if (NULL == (ev->op_info.app_file_name = H5MM_xstrdup(app_file)))
```
gnuoyd: I was concerned about unnecessary overheads when I saw that you were passing the source filename, function name, and line number. Now I see that you are strdup'ing some of the information. This is an awful lot of overhead for an application to absorb when it's not being debugged. Let's talk on the mailing list about how to lower the costs and put this under some kind of runtime switch.

src/H5ES.c
```diff
@@ -135,6 +184,41 @@ H5ESget_count(hid_t es_id, size_t *count /*out*/)
...
+ *
+ * Return:      SUCCEED / FAIL
+ *
+ * Programmer:	Quincey Koziol
+ *              Thursday, December 10, 2020
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5ESget_time_estimate(hid_t es_id, uint64_t *time_est /*out*/)
```
gnuoyd: I don't think we should add `get_time_estimate` to the library API until there are some example implementations that provide it and some evidence that any application actually needs it.

src/H5VLcallback.c
```diff
@@ -3526,6 +4171,50 @@ H5VLfile_get(void *obj, hid_t connector_id, H5VL_file_get_t get_type, hid_t dxpl
...
+ * Note:        Same as H5VLfile_get, but uses varargs instead of a va_list
+ *
+ * Return:      Success:    Non-negative
+ *              Failure:    Negative
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5VLfile_get_vararg(void *obj, hid_t connector_id, H5VL_file_get_t get_type, hid_t dxpl_id,
+                    void **req /*out*/, ...)
```
gnuoyd: Just reading this is giving me tendonitis. :-)

It is not necessary to be so repetitive. Let's talk.

gnuoyd (changes requested):
Is there some design document for the optional operations?

There is too much going on in this one PR and it should be broken down.

qkoziol:
> Is there some design document for the optional operations?
> 
> There is too much going on in this one PR and it should be broken down.

I completely understand, it was not my original intention to submit it in this manner either, but that was what we agreed to do last Friday.  I will continue to produce documentation, tests, code updates, etc. after it's merged in.   Please release your "change request" hold and add items to the mind map here for prioritization and discussion:  https://coggle.it/diagram/X9e7HxlvmVpHJ7Rb/t/ecp-deliverables

src/H5ES.c
derobins: I am also not a fan of this API call. I think it's going to be nigh impossible to make it useful.

qkoziol: It's up to the VOL connector to figure it out, and it's a "quality of implementation" aspect of an asynchronous VOL connector.   Any connector that doesn't want to try to estimate operations can always return the "unknown estimate" value (UINT64_MAX) for a request and the event set will stop iterating over the requests in the event set and return the "unknown" value to the application.

I would have used the new "dynamic optional VOL operations" feature for it, but it requires the addition of a new VOL operation value for requests and there's no way to have the event set "know" about a dynamically registered operation.

LBNL is currently improving the async VOL connector to track the I/O statistics and return more meaningful time estimates, and we will work with applications to take advantage of this capability to balance their compute vs. I/O.  Having a time estimate for a set of I/O operations is a _big_ enabler for pushing applications to take advantage of async I/O - it's something they've never had the ability to predict before.   When an application retrieves the time estimate and it's a lot larger or smaller than their predicted compute time, they can adjust one or the other to better overlap and take more advantage of the system resources.

soumagne:
I see a lot of formatting changes in that PR, can someone explain to me how this can be possible? (@byrnHDF) I thought that once we moved to clang-format everything was done automatically and checked whenever a new PR is made?

byrnHDF:
That should be the case - we will need to investigate why.

Allen

byrnHDF:
> That should be the case - we will need to investigate why. Allen
I see 
`clang-format on`
comments that would not be in format checks until the next line.
I saw some other comments with "//" that might need special handling, maybe?

byrnHDF:
As for the others, they must have been checked in without running format?

qkoziol:
Hmm, I’ll re-run the formatter and push any change to the PR.

soumagne:
If the current develop needs the formatter to be run, it should be done separately from that PR.

qkoziol:
> If the current develop needs the formatter to be run, it should be done separately from that PR.

Yep, submitting one for develop now.

qkoziol:
In addition to the mind map, I’m working on a support document for this PR, here:   https://docs.google.com/document/d/1xNOq3MfLp7YhKlmxOKoZzag03q0YoHONNiz9svQ60Iw/edit?usp=sharing

src/H5ESpublic.h
```diff
@@ -120,13 +120,18 @@ extern "C" {
 #endif
 
 H5_DLL hid_t  H5EScreate(void);
+H5_DLL herr_t H5ESinsert_request(hid_t es_id, hid_t connector_id, void *request);
 H5_DLL herr_t H5ESwait(hid_t es_id, uint64_t timeout, size_t *num_in_progress, hbool_t *err_occurred);
+H5_DLL herr_t H5EScancel(hid_t es_id, size_t *num_not_canceled, hbool_t *err_occurred);
 H5_DLL herr_t H5ESget_count(hid_t es_id, size_t *count);
+H5_DLL herr_t H5ESget_time_estimate(hid_t es_id, uint64_t *time_estimate);
```
epourmal: This function still needs more discussions based on the review on 1/8/21

qkoziol:
> In addition to the mind map, I’m working on a support document for this PR, here: https://docs.google.com/document/d/1xNOq3MfLp7YhKlmxOKoZzag03q0YoHONNiz9svQ60Iw/edit?usp=sharing

I’ve finished updating this support document to cover all of the changes in the PR.

src/H5ES.c
markcmiller86: Just spit-balling here but might there be a way here or in the future for a record of time an `unknown estimate` operation wound up taking and a place to store/hash that value so that if the situation is encountered again, instead of `unknown estimate` it can return that value based on previous experience? Thats likely well beyon what is necessary to get this initial implementation into the main lib but wanted to toss out the idea to possibly inform thinking.

src/H5ESint.c
markcmiller86: @epourmal for some reason, in my mind's eye, I had the impression the new docs-like-code workflow THG started in the PSIP work and has been working towards would result in being able to document functions more like this...

```
static herr_t
 H5ES__insert(
    H5ES_t *es,   /<** the event set object identifier **/
    H5VL_t *connector, /**< the VOL connector identifier **/
    void *request_token, /**< the request token, 
    const char *app_file, /**< the name of the originating application file of the request **/
    const char *app_func, /**< the name of the originating application function of the request **/
    unsigned app_line, /**< the line of application source code of the originating request **/
    const char *caller, /**< the name of the calling function **/
    const char *api_args) /**< the calling function's arguments, separated by commas **/
```

and have that get slurrped into the browseable documentation using whatever automation doc tool workflow you use. Am I totally misunderstanding things?

Also, @gnuoyd, apologies for inserting myself into your conversation here but I am curious...I have read absolutely nothing and have discussed absolutely nothing with anyone about this API and still believe I know what (most of) the arguments to this function are (supposed to be) because of how they are named. I don't know what the difference is between `app_` args and `caller` and `api_args` and I would want further clarification on that. However, I wonder how much more useful the additional documentation (or a version of it that you would prefer) really leads to deeper/better understanding.

test/vol.c
```diff
@@ -162,6 +295,85 @@ static const H5VL_class_t fake_vol_g = {
     NULL /* optional     */
 };
 
+/*-------------------------------------------------------------------------
```
markcmiller86, line 432: Does this test code demonstrate/exercise the API as it might be used by in a real use case or is it really just unit-checking the API? If the latter, I would want to see a real and well-documented end-user example which exercises key (if not all) modalities of use (or separate tests for separate modalities) also with this code base to understand how the user-facing API is intended to be used.

markcmiller86:
I happened to be browsing around here and just had a few questions. Not urgent to answer so don't let any of my questions get in the way of the real work of completing this PR...

- GitHub provides a pretty useful wiki which supports a wide variety of document formats include markdown but also sphinx, wikimedia, etc. I wonder if it makes sense to handle supporting documentat like the google doc referenced from this issue in that form instead? Maybe non-essential but I find it helps to keep all the docs in one place
- This PR seems like it is mixing a few different things. Any reason it cannot be decomposed into separate PRs each for specific functionality needed? Turns out [I've been writing](https://github.com/betterscientificsoftware/bssw.io/blob/cc-mcm86-10dec20-large-prs/Articles/ShortArticles/PullRequestSizeConsiderations-wikized.md) on just this issue recently.
- The google doc refers to a use case of sorts. Is it possible to have that end-user use case as a test/example code? With that, as API changes, it too would get updated. In addition, it could be documented such that it kinda sorta serves as the key umbrella artifact to facilitate dialog about the change and its specific details and why they are needed -- or maybe alternatives to consider.
- Are 14 reviewers really needed to get this through? That seems like huge overkill and a big use of resource/time. On VisIt, (2.8 million lines of C++ code with some Python) we handle PRs that are probably 10-20x this size with one or two reviewers. We also try to complete PR reviews in < 2-3 days. I don't think I've seen any project that involves 14 reviewers for a PR. Just curious what the rationale is.
- Just curious but in your testing (which is way cool BTW!!), do you have an idea on the coverage yet? Would be a nice warm fuzzy if the tests here for this new functionality were known to cover most of the code.
- If one wanted to submit doc updates with these code changes (e.g. docs-like-code), is that possible here?

java/src/hdf/hdf5lib/HDFArray.java
```diff
@@ -169,10 +169,6 @@ else if (ArrayDescriptor.className
                                 .equals("java.lang.Integer")) {
                             therow = IntegerToByte((Integer[]) _theArray);
                         }
-                        else if (ArrayDescriptor.className
```
byrnHDF: Why was this removed?

java/src/hdf/hdf5lib/HDFArray.java
```diff
@@ -522,36 +518,34 @@ else if (ArrayDescriptor.className.equals("java.lang.Long")) {
             flattenedArray = (Object) _barray;
             break;
         case 'L':
-            switch (ArrayDescriptor.className) {
```
byrnHDF: Is this file maybe not up-to-date?

test/Makefile.am
```diff
@@ -176,7 +176,7 @@ CHECK_CLEANFILES+=accum.h5 cmpd_dset.h5 compact_dataset.h5 dataset.h5 dset_offse
     zero_chunk.h5 chunk_single.h5 swmr_non_latest.h5 \
     earray_hdr_fd.h5 farray_hdr_fd.h5 bt2_hdr_fd.h5 \
     storage_size.h5 dls_01_strings.h5 power2up.h5 version_bounds.h5 \
-    alloc_0sized.h5 \
+    alloc_0sized.h5 h5s_block.h5 \
```
byrnHDF: Should also be added to CMake file.

src/H5ESint.c
gnuoyd: @markcmiller86, I find the argument names meaningful, too. I object to the 10-line header comment precisely because it takes up so much space while telling us little more about the program than the code itself does. An unadorned imperative sentence will suffice, `/* Insert a request token into an event set. */`.

The current comment is a waste of display space that could hold program meaning, instead.

test/vol.c
qkoziol: This is a unit test and is only checking the API for correctness.   There’s a separate VOL connector testing suite under development, here:   https://github.com/HDFGroup/vol-tests

java/src/hdf/hdf5lib/HDFArray.java
qkoziol: The changes in this file were to correct Java compile failures on one of the THG machines (sorry, can’t remember which one, maybe @lrknox remembers), which were failing with a complaint about a switch using a string variable.

test/Makefile.am
qkoziol: Yes, will add, thanks!

java/src/hdf/hdf5lib/HDFArray.java
byrnHDF: I think this file and the changes need to be extracted to a separate issue for further investigation.

java/src/hdf/hdf5lib/HDFArray.java
byrnHDF: I can understand the changes for this switch, again this should be extracted to a separate issue.

qkoziol:
> I happened to be browsing around here and just had a few questions. Not urgent to answer so don't let any of my questions get in the way of the real work of completing this PR...
> 
> * GitHub provides a pretty useful wiki which supports a wide variety of document formats include markdown but also sphinx, wikimedia, etc. I wonder if it makes sense to handle supporting documentat like the google doc referenced from this issue in that form instead? Maybe non-essential but I find it helps to keep all the docs in one place

Sounds like a good idea, I’ll try the wiki for future situations like this.

> * This PR seems like it is mixing a few different things. Any reason it cannot be decomposed into separate PRs each for specific functionality needed? Turns out [I've been writing](https://github.com/betterscientificsoftware/bssw.io/blob/cc-mcm86-10dec20-large-prs/Articles/ShortArticles/PullRequestSizeConsiderations-wikized.md) on just this issue recently.

I agree, but there were complex issues around this PR, outside of the code proper, and it worked out that all of the changes should be put into one PR.

> * The google doc refers to a use case of sorts. Is it possible to have that end-user use case as a test/example code? With that, as API changes, it too would get updated. In addition, it could be documented such that it kinda sorta serves as the key umbrella artifact to facilitate dialog about the change and its specific details and why they are needed -- or maybe alternatives to consider.

Yes, I’ll see if I can work that into the VOL testing code (in https://github.com/HDFGroup/vol-tests)

> * Are 14 reviewers really needed to get this through? That seems like huge overkill and a big use of resource/time. On VisIt, (2.8 million lines of C++ code with some Python) we handle PRs that are probably 10-20x this size with one or two reviewers. We also try to complete PR reviews in < 2-3 days. I don't think I've seen any project that involves 14 reviewers for a PR. Just curious what the rationale is.

Agree.   Some of the reviewers are auto-assigned, though.

> * Just curious but in your testing (which is way cool BTW!!), do you have an idea on the coverage yet? Would be a nice warm fuzzy if the tests here for this new functionality were known to cover most of the code.

There’s no code coverage testing on HDF5 currently.   I agree with you, that would be a really good addition. :-)

> * If one wanted to submit doc updates with these code changes (e.g. docs-like-code), is that possible here?

To some degree - the Doxygen comments in the public headers could be updated in a PR on the HDF5 repo, but much of the other HDF5 documentation is in other repos, and I don’t think the main hdf5doc repo has been moved to github (yet?), it’s still on the THG bitbucket:  https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5doc/browse

src/H5ESint.c
qkoziol: > @epourmal for some reason, in my mind's eye, I had the impression the new docs-like-code workflow THG started in the PSIP work and has been working towards would result in being able to document functions more like this...
> 
> ```
> static herr_t
>  H5ES__insert(
>     H5ES_t *es,   /<** the event set object identifier **/
>     H5VL_t *connector, /**< the VOL connector identifier **/
>     void *request_token, /**< the request token, 
>     const char *app_file, /**< the name of the originating application file of the request **/
>     const char *app_func, /**< the name of the originating application function of the request **/
>     unsigned app_line, /**< the line of application source code of the originating request **/
>     const char *caller, /**< the name of the calling function **/
>     const char *api_args) /**< the calling function's arguments, separated by commas **/
> ```
> 
> and have that get slurped into the browseable documentation using whatever automation doc tool workflow you use. Am I totally misunderstanding things?

Currently, the Doxygen comments are only in the HDF5 public header files.   Adding them to the parameters for internal routines may be useful if the coding standard was updated to push in that direction.  (Although it would be best done incrementally, in the Boy Scout "Leave the code you change in better shape than you found it" way, since it would be a _huge_ job overall :-)

java/src/hdf/hdf5lib/HDFArray.java
byrnHDF: Created PR just for this file

src/H5ES.c
qkoziol: Good point - still up the VOL connector, but could be useful for that code.

epourmal:
Hi Quincey,

I preparation for the merge, could you please add descriptions of the changes to the RELEASE.txt file? Thank you!

epourmal:
Would it be possible to indicate which changes should go into 1_12, 1_10 and (maybe?) 1_8 branches? Thank you!

qkoziol:
> Hi Quincey,
> 
> I preparation for the merge, could you please add descriptions of the changes to the RELEASE.txt file? Thank you!

Yes, I will definitely do this.

qkoziol:
> Would it be possible to indicate which changes should go into 1_12, 1_10 and (maybe?) 1_8 branches? Thank you!

If I have some time, I will go back to annotate merges to prior release branches.

byrnHDF (approved):
The CMake, java and tools look fine.

src/H5ESevent.c
```diff
@@ -185,6 +185,14 @@ H5ES__event_completed(H5ES_event_t *ev, H5ES_event_list_t *el)
     /* Sanity check */
     HDassert(ev);
 
+    /*
+        HDfprintf(stderr,
+                  "%s: releasing event for '%s' (count: %llu, timestamp: %llu), with args '%s', in app source
+       " "file '%s', function '%s', and line %u\n", FUNC, ev->op_info.api_name, (unsigned long
+       long)ev->op_info.op_ins_count, (unsigned long long)ev->op_info.op_ins_ts, ev->op_info.api_args,
+       ev->op_info.app_file_name, ev->op_info.app_func_name, ev->op_info.app_line_num);
```
jhendersonHDF: Would it be possible to loop this into HDF5's debugging capabilities? Not much reason to leave a commented out fprintf statement in the library.

src/H5TSdevelop.h
```diff
@@ -11,14 +11,15 @@
...
 
-#ifndef H5TSpublic_H
-#define H5TSpublic_H
+#ifndef H5TSdevelop_H
+#define H5TSdevelop_H
 
-/* Public headers needed by this file */
-#include "H5public.h" /* Generic Functions                    */
+/* Include package's public header */
+/* #include "H5TSpublic.h" */
```
jhendersonHDF: Any need to keep this line around?

src/H5TSprivate.h
```diff
@@ -25,8 +25,9 @@
 #define H5TSprivate_H_
 
 #ifdef H5_HAVE_THREADSAFE
-/* Public headers needed by this file */
-#include "H5TSpublic.h" /* Public API prototypes */
+/* Include package's public headers */
+/* #include "H5TSpublic.h" */
```
jhendersonHDF: Same as in H5TSdevelop.h - any need to keep this line?

src/H5VLcallback.c
```diff
@@ -1437,6 +1486,50 @@ H5VLattr_get(void *obj, hid_t connector_id, H5VL_attr_get_t get_type, hid_t dxpl
...
+ * Note:        Same as H5VLattr_get, but uses varargs instead of a va_list
+ *
+ * Return:      Success:    Non-negative
+ *              Failure:    Negative
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5VLattr_get_vararg(void *obj, hid_t connector_id, H5VL_attr_get_t get_type, hid_t dxpl_id,
+                    void **req /*out*/, ...)
```
jhendersonHDF: I may very well have missed the discussions surrounding adding var_args versions of the VOL callbacks, but did we end up coming to a conclusion on these? While it looks like it did help clean up the passthrough connector a bit, it seems very inconvenient to have to add (and maintain) separate versions of each VOL call. Not to mention that it seems like a connector author may have problems determining when one is needed vs. the other.

src/H5A.c
```diff
@@ -1057,15 +1057,15 @@ H5Aread(hid_t attr_id, hid_t dtype_id, void *buf /*out*/)
...
-              void *buf, hid_t es_id)
+              void *buf /*out*/, hid_t es_id)
 {
     H5VL_object_t *vol_obj   = NULL;            /* Object for attr_id */
     void *         token     = NULL;            /* Request token for async operation        */
     void **        token_ptr = H5_REQUEST_NULL; /* Pointer to request token for async operation        */
     herr_t         ret_value = SUCCEED;         /* Return value */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE7("e", "*s*sIuii*xi", app_file, app_func, app_line, attr_id, dtype_id, buf, es_id);
```
jhendersonHDF, line 1068: Looks like the *x -> x in this line and the change below might be unintentional?

src/H5D.c
```diff
@@ -1032,13 +1036,13 @@ H5Dread_async(const char *app_file, const char *app_func, unsigned app_line, hid
...
  */
 herr_t
-H5Dread_chunk(hid_t dset_id, hid_t dxpl_id, const hsize_t *offset, uint32_t *filters, void *buf)
+H5Dread_chunk(hid_t dset_id, hid_t dxpl_id, const hsize_t *offset, uint32_t *filters, void *buf /*out*/)
 {
     H5VL_object_t *vol_obj   = NULL;
     herr_t         ret_value = SUCCEED; /* Return value */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE5("e", "ii*h*Iu*x", dset_id, dxpl_id, offset, filters, buf);
```
jhendersonHDF, line 1041: The same as for H5Aread_async - is the argument change here intentional?

src/H5ESevent.c
qkoziol: There's already similar info available in the error info for failed async tasks, I'll take this debugging printf out.

src/H5TSdevelop.h
qkoziol: There are similar placeholders in the library, so that's why I had left it in.   But, there's plenty of examples of how / where to include a public header file for an interface, so I'll take it out.

src/H5TSprivate.h
qkoziol: Ditto - will take out

src/H5VLcallback.c
qkoziol: Yes, we discussed adding them, a while ago, so they are in.   Having two versions is overkill to me also, but the va_list version was "legacy" at the time I added the vararg version.   If we're happy with it enough to justify [another] breaking change to the VOL connector callbacks, I can remove the va_list version.

src/H5A.c
qkoziol: No, that's intentional and the result of the bin/trace script updates that parse the parameters better.

src/H5D.c
qkoziol: Yes

src/H5VLcallback.c
jhendersonHDF: Ok, I think we should definitely discuss this with everyone then to see what the consensus is. IMO, it's still a good time to finalize any breaking changes, and, again IMO, having two versions will probably just end up being a nightmare for both maintenance and cognitive overhead in the long run.

src/H5A.c
jhendersonHDF: This is probably just from my ignorance of bin/trace in general, but the only changes to bin/trace I see are adding cases for H5ES_event_complete_func_t and H5ES_event_insert_func_t. How does that affect needing the asterisk for the void * buf parameter here?

test/vol.c
```diff
@@ -1176,6 +1538,510 @@ test_basic_datatype_operation(void)
...
+    /* Test query w/NULL for cap_flags parameter */
+    if (H5Pget_vol_cap_flags(fapl_id, NULL) < 0)
+        FAIL_STACK_ERROR;
+
+    /* Override possible environment variable & re-initialize default VOL connector */
+    conn_env_str = HDgetenv("HDF5_VOL_CONNECTOR");
+    if (conn_env_str) {
+        if (NULL == (conn_env_str = HDstrdup(conn_env_str)))
+            TEST_ERROR
+        if (HDunsetenv("HDF5_VOL_CONNECTOR") < 0)
```
jhendersonHDF, line 1982: Seems Windows is unhappy about this, since setenv/unsetenv aren't standard C. @derobins Is there a Windows equivalent for these? They're POSIX 2001, so I figure there probably are.

src/H5A.c
qkoziol: > This is probably just from my ignorance of bin/trace in general, but the only changes to bin/trace I see are adding cases for H5ES_event_complete_func_t and H5ES_event_insert_func_t. How does that affect needing the asterisk for the void * buf parameter here?

Adding the "/*out*/" comment to the parameters indicates to the bin/trace script that the buffer contents are not valid at function entry and are therefore an 'x' (i.e. totally opaque to tracing) and not a '*x' (which indicates a pointer to a buffer of bytes).

src/H5VLcallback.c
qkoziol: > Ok, I think we should definitely discuss this with everyone then to see what the consensus is. IMO, it's still a good time to finalize any breaking changes, and, again IMO, having two versions will probably just end up being a nightmare for both maintenance and cognitive overhead in the long run.

Agree - I'll add it to the agenda for next Monday's technical meeting

src/H5ES.c
```diff
@@ -102,11 +103,62 @@ H5EScreate(void)
...
+ *
+ * Return:      SUCCEED / FAIL
+ *
+ * Programmer:	Quincey Koziol
+ *              Friday, December 11, 2020
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5ESinsert_request(hid_t es_id, hid_t connector_id, void *request)
```
fortnern, line 122: Should this include debugging parameters? Either explicitly or through a macro like the _async APIs?

src/H5ESint.c
```diff
@@ -369,6 +553,91 @@ H5ES__handle_fail(H5ES_t *es, H5ES_event_t *ev)
...
+    herr_t ret_value    = SUCCEED;         /* Return value */
+
+    FUNC_ENTER_STATIC
+
+    /* Sanity check */
+    HDassert(es);
+    HDassert(ev);
+
+    /* Check for event that hasn't actually completed */
+    if (H5VL_REQUEST_STATUS_IN_PROGRESS == ev_status || H5VL_REQUEST_STATUS_CANT_CANCEL == ev_status)
```
fortnern: Shouldn't this be an assertion?

src/H5FDdevelop.h
```diff
@@ -0,0 +1,260 @@
...
+ * a fair amount of work to add a new kind of file memory and they are similar
+ * enough to object headers and probably too minor to deserve their own type.
+ *
+ * Map "fixed array" data blocks & pages to 'lheap' type file memory, since
+ * they are similar enough to local heap info.
+ *
+ */
+#define H5FD_MEM_FARRAY_HDR       H5FD_MEM_OHDR
+#define H5FD_MEM_FARRAY_DBLOCK    H5FD_MEM_LHEAP
+#define H5FD_MEM_FARRAY_DBLK_PAGE H5FD_MEM_LHEAP
```
fortnern, line 104: Hmm maybe instead of all these ad-hoc redefinitions we should rename the categories to be decriptive of the type of objects they contain (a more general description than "ohdr" for example)

fortnern: Or are these already part of the public API?

src/H5Spublic.h
```diff
@@ -23,6 +23,7 @@
 
 /* Define atomic datatypes */
 #define H5S_ALL       (hid_t)0
+#define H5S_BLOCK     (hid_t)1
```
fortnern: Is H5S_BLOCK fully implemented?

src/H5VL.c
```diff
@@ -911,3 +944,133 @@ H5VLquery_optional(hid_t obj_id, H5VL_subclass_t subcls, int opt_type, uint64_t
...
+ *              operation on that subclass.
+ *
+ *              For example, registering a 'prefetch' operation for the
+ *              caching VOL connector written at the ALCF at Argonne National
+ *              Laboratory could have a UTI of: "gov.anl.alcf.cache.prefetch",
+ *              and the "evict" operation for the same connector could have a
+ *              UTI of: "gov.anl.alcf.cache.evict".   Registering a "suspend
+ *              background threads" operation for the asynchronous VOL connector
+ *              written at NERSC at Lawrence Berkeley National Laboratory could
+ *              have a UTI of: "gov.lbnl.nersc.async.suspend_bkg_threads".
```
fortnern, line 967: Seems strange to use external (to THG) organizations as examples here

src/H5VLcallback.c
```diff
@@ -5324,12 +6509,18 @@ H5VLlink_optional(void *obj, hid_t connector_id, H5VL_link_optional_t opt_type,
...
         HGOTO_ERROR(H5E_VOL, H5E_CANTOPERATE, FAIL, "unable to execute link optional callback")
 
 done:
+    /* End access to the va_list, if we started it */
+    if (arg_started)
+        HDva_end(arguments);
+
     FUNC_LEAVE_API_NOINIT(ret_value)
-} /* end H5VLlink_optional() */
+} /* end H5VLlink_optional_vararg() */
```
fortnern, line 6523: No H5VLlink_optional_op()?

src/H5VLcallback.c
```diff
@@ -3091,6 +3598,51 @@ H5VL_datatype_optional(const H5VL_object_t *vol_obj, H5VL_datatype_optional_t op
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5VL_datatype_optional() */
 
+/*-------------------------------------------------------------------------
```
fortnern, line 3145: no datatype_optional_vararg? (I see the existing one uses varargs already, but shouldn't they all be made analogous?)

src/H5ESint.c
```diff
@@ -50,6 +50,11 @@
 /* Local Typedefs */
 /******************/
 
+/* Callback context for get time estimate (gte) operations */
+typedef struct H5ES_gte_ctx_t {
+    uint64_t *time_est; /* Pointer to accumulated time estimate */
```
jrmainzer: It would be good to mention units both here and in the timeout field of H5ES_wait_ctx_t below.

src/H5VLcallback.c
```diff
@@ -5875,6 +7158,50 @@ H5VLobject_optional(void *obj, hid_t connector_id, H5VL_object_optional_t opt_ty
...
+        HGOTO_ERROR(H5E_VOL, H5E_CANTOPERATE, FAIL, "unable to execute object optional callback")
+
+done:
+    /* End access to the va_list, if we started it */
+    if (arg_started)
+        HDva_end(arguments);
+
+    FUNC_LEAVE_API_NOINIT(ret_value)
+} /* end H5VLobject_optional_vararg() */
+
```
fortnern, line 6107: No H5VLobject_optional_op()?

src/H5ESint.c
```diff
@@ -369,6 +553,91 @@ H5ES__handle_fail(H5ES_t *es, H5ES_event_t *ev)
...
+        if (es->comp_func) {
+            H5ES_status_t op_status; /* Status for complete callback */
+
+            /* Set appropriate info for callback */
+            if (H5VL_REQUEST_STATUS_SUCCEED == ev_status) {
+                /* Translate status */
+                op_status = H5ES_STATUS_SUCCEED;
+
+                /* Retrieve the execution time info */
+                if (H5VL_request_specific(ev->request, H5VL_REQUEST_GET_EXEC_TIME, &ev->op_info.op_exec_ts,
```
fortnern: We're concerned that this code will add unnecessary overhead to every operation. It will also require all connectors to handle this specific value. Maybe we should query this capability and only call this code if it's implemented?

src/H5VLcallback.c
fortnern: Elena is open to changing the existing functions to use varargs

test/vol.c
qkoziol: @jhendersonHDF - I believe that I've fixed this, can you please re-test?

src/H5ES.c
qkoziol: I trend in the "no" direction, since those debugging parameters are designed for application operations, and this API routine will only be invoked within a VOL connector.

test/vol.c
jhendersonHDF: I was just paying attention to the CI checks. Since they're running again we should see soon.

src/H5ESint.c
qkoziol: Make sense, done.

src/H5Spublic.h
qkoziol: Yes, with tests, etc.

src/H5VLcallback.c
qkoziol: Yes, datatype_optional_vararg is here, just a little further down in the file.

src/H5ESint.c
qkoziol: Done

src/H5VLcallback.c
qkoziol: The dynamic operations are registered on a subclass type, so it didn't seem like a good fit to add the generic form (H5VLobject_optional_op).

src/H5ESint.c
qkoziol: So - convert it to an optional operation?

qkoziol:
I received a few emails with comments from @fortnern, but don't see them here.  Were they deleted?

src/H5ESint.c
soumagne: I thought we agreed in the previous discussions that time estimates were not part of that PR? I think I'm fine with optional. If not optional, I think that should just be removed from the code at this point of time and a separate PR should be made with the specific changes that add time estimates?

src/H5ES.c
```diff
@@ -117,24 +169,69 @@ H5EScreate(void)
...
+            HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "NULL time_est pointer")
+
+        /* Get estimate for completing operations */
+        if (H5ES__get_time_estimate(es, time_est) < 0)
+            HGOTO_ERROR(H5E_EVENTSET, H5E_CANTGET, FAIL, "can't get time estimate")
+    } /* end if */
+
+done:
+    FUNC_LEAVE_API(ret_value)
+} /* end H5ESget_time_estimate() */
```
soumagne: Same comment as the other one, why is that routine still in this PR? I thought we agreed we won't add that just yet until we have a better prototype for it?

soumagne (changes requested):
Except the time estimate routines and corresponding VOL flags that should be removed at this point of time, the rest looks good to me.

fortnern:
> I received a few emails with comments from @fortnern, but don't see them here. Were they deleted?

There were a few I deleted after I realized they didn't apply. It looks like you found the ones I didn't delete.

fortnern:
I've finished reviewing the changes. Looks good to me once the outstanding comment threads are resolved.

qkoziol:
> Except the time estimate routines and corresponding VOL flags that should be removed at this point of time, the rest looks good to me.

These have been removed.

src/H5Dpublic.h
```diff
@@ -677,7 +678,7 @@ H5_DLL herr_t H5Dget_chunk_info_by_coord(hid_t dset_id, const hsize_t *offset, u
  *                  indicating failure.
  *
  */
-H5_DLL herr_t H5Dchunk_iter(hid_t dset_id, H5D_chunk_iter_op_t cb, void *op_data);
+H5_DLL herr_t H5Dchunk_iter(hid_t dset_id, hid_t dxpl_id, H5D_chunk_iter_op_t cb, void *op_data);
```
soumagne, line 682: Is it not an API change? should not you do a v2?

src/H5Spublic.h
```diff
@@ -21,8 +21,12 @@
 #include "H5public.h"
 #include "H5Ipublic.h"
 
-/* Define atomic datatypes */
-#define H5S_ALL       0 /* (hid_t) */
+/* Define special dataspaces for dataset I/O operations */
+#define H5S_ALL   0 /* (hid_t) */
+#define H5S_BLOCK 1 /* (hid_t) */
+#define H5S_PLIST 2 /* (hid_t) */
+
```
soumagne, line 28: I feel like you should add what they mean instead of hid_t ?

soumagne (approved):
Looks good to me, just a few minor comments.

src/H5Spublic.h
qkoziol: I was just extending the list (which is like that in develop).

src/H5Dpublic.h
qkoziol: No, this was only merged to develop a week ago and there's no BW compatibility issues.

src/H5Spublic.h
soumagne: Are they documented somewhere? I have no idea what they are

soumagne: I think it would not hurt to add a one-line description of their effect / when to use them, it does not seem to be in H5Dpublic.h either.

qkoziol: Yes, definitely a good thing to do:  https://github.com/HDFGroup/hdf5/issues/688

derobins (approved):

PR #166, opened 2020-12-12
Bring async branch to develop

qkoziol:
This is the base set of async operation support.  I'll also submit a PR for the "combo" version, so it can be merged after this PR is merged.


derobins (approved):
I'm sure there will be future discussions about the async feature, but for the purposes we've discussed, this can be merged.

lrknox (approved):

PR #169, opened 2020-12-14
Hdf5 1 8

bmribler:
Sorry, I didn't realize I had another commit hung around (Nov 30) and it's combined here when I tried to create a PR for just the commit today.  They're all simple though.  Thank you!

### DIFF
--- a/src/H5Ocache.c
+++ b/src/H5Ocache.c
@@ -1432,7 +1432,7 @@ H5O_chunk_proxy_dest(H5O_chunk_proxy_t *chk_proxy)
     HDassert(chk_proxy);
 
     /* Decrement reference count of object header */
-    if (chk_proxy->oh && H5O_dec_rc(chk_proxy->oh) < 0)
+    if (H5O_dec_rc(chk_proxy->oh) < 0)
         HGOTO_ERROR(H5E_OHDR, H5E_CANTDEC, FAIL, "can't decrement reference count on object header")
 
     /* Release the chunk proxy object */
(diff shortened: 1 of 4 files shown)


src/H5Ocache.c
```diff
@@ -1432,7 +1432,7 @@ H5O_chunk_proxy_dest(H5O_chunk_proxy_t *chk_proxy)
     HDassert(chk_proxy);
 
     /* Decrement reference count of object header */
-    if (chk_proxy->oh && H5O_dec_rc(chk_proxy->oh) < 0)
+    if (H5O_dec_rc(chk_proxy->oh) < 0)
```
derobins, line 1435: Why remove the check for chk_proxy->oh?

bmribler: Because a check for null pointer is added in H5O_dec_rc().  Is it wrong?

derobins: Normally, I try to fail as early as possible and not hope that later code checks things. In general, when we use a pointer, we should assert on it first.

lrknox (approved):

src/H5Ocache.c
bmribler: There was an assert in H5O_dec_rc() originally and no check for null pointer before calling H5O_dec_rc() until the user's corrupted file issue came up.  That original fix only patched that one particular issue but there are more than one place where H5O_dec_rc() is called.  So, instead of putting in the check for every call to H5O_dec_rc() and hoping its future appearance will be accompanied by a proper check, this is a better approach, in my opinion.  Quincey agreed.

PR #177, opened 2020-12-15
Doxygen - add reference manual entries for H5A, H5S, and H5P

bljhdf:


byrnHDF:
Looks like MANIFEST had a problem?

bljhdf:
start over

PR #182, opened 2020-12-15
Enable H5Fopen to probe list of available VOL connectors when opening a file

jhendersonHDF:
This is the H5Fopen VOL auto-detect feature that was previously reviewed, but needed some fixup before merging. This PR adds the capability to iterate over all of the available plugins in the plugin path, optionally processing only filter plugins, only VOL plugins or all plugins, and perform some callback function on them. Adding this capability also allows us to look for any VOL connector plugins in the plugin path that can open a given file when the initial file open fails for whatever VOL connector is set on the FAPL used.

@lrknox I believe that this is the last piece we need in HDF5 before creating our final RC tag for the DAOS team to use.

### DIFF
--- a/src/H5PLpath.c
+++ b/src/H5PLpath.c
@@ -62,6 +62,8 @@ static herr_t H5PL__insert_at(const char *path, unsigned int idx);
 static herr_t H5PL__make_space_at(unsigned int idx);
 static herr_t H5PL__replace_at(const char *path, unsigned int idx);
 static herr_t H5PL__expand_path_table(void);
+static herr_t H5PL__path_table_iterate_process_path(const char *plugin_path, H5PL_iterate_type_t iter_type,
+                                                    H5PL_iterate_t iter_op, void *op_data);
 static herr_t H5PL__find_plugin_in_path(const H5PL_search_params_t *search_params, hbool_t *found,
                                         const char *dir, const void **plugin_info);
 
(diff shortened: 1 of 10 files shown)


src/H5PLpath.c
```diff
@@ -546,7 +548,220 @@ H5PL__get_path(unsigned int idx)
...
+ * Return:      H5_ITER_CONT if all plugins are processed successfully
+ *              H5_ITER_STOP if short-circuit success occurs while
+ *                  processing plugins
+ *              H5_ITER_ERROR if an error occurs while processing plugins
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef H5_HAVE_WIN32_API
+static herr_t
+H5PL__path_table_iterate_process_path(const char *plugin_path, H5PL_iterate_type_t iter_type,
```
jhendersonHDF, line 602: I'm not a big fan of needing two versions of this routine, but it would take too long for me to refactor the system-dependent code like we discussed last time this was reviewed and I don't want to risk breaking the existing plugin code by getting that wrong. In the interest of time, I'm leaving sleeping dogs lie here.

src/H5PLpkg.h
```diff
@@ -153,7 +154,8 @@ H5_DLL herr_t      H5PL__replace_path(const char *path, unsigned int index);
 H5_DLL herr_t      H5PL__insert_path(const char *path, unsigned int index);
 H5_DLL herr_t      H5PL__remove_path(unsigned int index);
 H5_DLL const char *H5PL__get_path(unsigned int index);
-H5_DLL herr_t      H5PL__find_plugin_in_path_table(const H5PL_search_params_t *search_params,
-                                                   hbool_t *found /*out*/, const void **plugin_info /*out*/);
+H5_DLL herr_t H5PL__path_table_iterate(H5PL_iterate_type_t iter_type, H5PL_iterate_t iter_op, void *op_data);
+H5_DLL herr_t H5PL__find_plugin_in_path_table(const H5PL_search_params_t *search_params,
```
jhendersonHDF, line 158: Clang seems to be confused about the formatting here, but this is what it wants, so it's what goes.

src/H5VL.c
```diff
@@ -87,27 +87,6 @@ H5VLregister_connector(const H5VL_class_t *cls, hid_t vipl_id)
...
-                    "VOL connector class name cannot be the empty string")
-    if (cls->info_cls.copy && !cls->info_cls.free)
-        HGOTO_ERROR(
-            H5E_VOL, H5E_CANTREGISTER, H5I_INVALID_HID,
-            "VOL connector must provide free callback for VOL info objects when a copy callback is provided")
-    if (cls->wrap_cls.get_wrap_ctx && !cls->wrap_cls.free_wrap_ctx)
-        HGOTO_ERROR(H5E_VOL, H5E_CANTREGISTER, H5I_INVALID_HID,
-                    "VOL connector must provide free callback for object wrapping contexts when a get "
-                    "callback is provided")
-
```
jhendersonHDF, line 110: Since I bypass H5VLregister_connector internally and go straight through H5VL__register_connector_by_class, I need these checks to be down there and not at the API level. Otherwise, someone could theoretically force a load of an invalid VOL connector (like one without a name field set) and cause a segfault later on.

src/H5VLcallback.c
```diff
@@ -3346,6 +3358,97 @@ H5VL__file_open(const H5VL_class_t *cls, const char *name, unsigned flags, hid_t
...
+    if (NULL == (fapl_plist = (H5P_genplist_t *)H5I_object_verify(udata->fapl_id, H5I_GENPROP_LST)))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, H5_ITER_ERROR, "not a property list")
+    if ((fapl_id = H5P_copy_plist(fapl_plist, TRUE)) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTCOPY, H5_ITER_ERROR, "can't copy fapl");
+    if (NULL == (fapl_plist_copy = (H5P_genplist_t *)H5I_object_verify(fapl_id, H5I_GENPROP_LST)))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, H5_ITER_ERROR, "not a property list")
+    if (H5P_set_vol(fapl_plist_copy, connector_id, NULL) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTSET, H5_ITER_ERROR, "can't set VOL connector on fapl")
+
+    /* Check if file is accessible with given VOL connector */
```
qkoziol, line 3419: I think it would be good to only check the is_accessible flag if H5VL_file_specific() returns 'SUCCEED'.   Something like:

```
     /* Check if file is accessible with given VOL connector */
     H5E_BEGIN_TRY
     {
        status =  H5VL_file_specific(NULL, H5VL_FILE_IS_ACCESSIBLE, H5P_DATASET_XFER_DEFAULT, H5_REQUEST_NULL, fapl_id,
                            udata->filename, &is_accessible);
     }
     H5E_END_TRY;

      /* If the file was accessible with the current VOL connector, return
      * the FAPL with that VOL connector set on it. Errors are ignored here
      * as some VOL connectors may not support H5Fis_accessible.
      */
     if (status > 0 && is_accessible > 0) {
         /* Modify 'connector_prop' to point to the VOL connector that
          * was actually used to open the file, rather than the original
          * VOL connector that was requested.
          */
         udata->connector_prop->connector_id   = connector_id;
         udata->connector_prop->connector_info = NULL;

          udata->fapl_id = fapl_id;
         ret_value      = H5_ITER_STOP;
     } /* end if */
```

src/H5VLcallback.c
```diff
@@ -3373,8 +3476,39 @@ H5VL_file_open(const H5VL_connector_prop_t *connector_prop, const char *name, un
         HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, NULL, "not a VOL connector ID")
 
     /* Call the corresponding internal VOL routine */
-    if (NULL == (ret_value = H5VL__file_open(cls, name, flags, fapl_id, dxpl_id, req)))
-        HGOTO_ERROR(H5E_VOL, H5E_CANTOPENOBJ, NULL, "open failed")
+    if (NULL == (ret_value = H5VL__file_open(cls, name, flags, fapl_id, dxpl_id, req))) {
```
qkoziol, line 3481: I'm 98% on board here, but am concerned about the application specifying a particular VOL connector to open the file with, then that file open fails, and we tear off around the file system trying out VOL connectors to see if one works out.   What do you think about only iterating through the plugin path if the VOL connector class for the failed file open was the default connector?  (i.e. the native connector)

src/H5PLpath.c
derobins: I can revisit this when I work on the VFD plugins early next year.

src/H5VLcallback.c
jhendersonHDF: I think that seems reasonable enough to me.

src/H5VLcallback.c
jhendersonHDF: Hmm, what if the application specifies a VOL connector to use and it's the same as the library's default connector? Should we still avoid iterating through the plugin path then? And should we differentiate that by only doing the iteration when H5P_DEFAULT is passed for the FAPL ID?

I think I wrote this up in the way you're describing for the tools. If a VOL connector ID/name is passed on the command-line, the tools should fail right away if they can't open the file with that connector. Otherwise, they try native first and then whatever is in the plugin path.

qkoziol: I'm fine with either way of determining "the user didn't specify a particular VOL connector".  Whichever is easier / more sensible to you.   (Glad the tools work that way, too :-) )

derobins: I think that the tools way is a good way to handle it. Failure with a specified VOL should be immediate, but calling H5Fopen without a VOL specified should trigger a VOL spam. I would include using the environment variable as "specifying a VOL connector", so we should only trigger the flurry of VOL checks if the default is the native VOL connector.

jhendersonHDF: The only difficult part I see is trying to differentiate if the user has actually set a VOL connector. They could pass in a FAPL that has some unrelated property set on it (for example, H5Pset_fapl_mpio might cause issues here in parallel applications), so just checking H5P_DEFAULT might be problematic here.

src/H5VLcallback.c
jhendersonHDF: Updated now.

src/H5VLcallback.c
jhendersonHDF: This is updated now to check for H5P_DEFAULT OR compare the given connector ID against the default. This means if someone passes in a FAPL that they specifically set the default VOL connector on, it will still trigger this feature, but I think that's a very unlikely case (at least currently).

derobins (approved):

qkoziol (approved):

PR #184, opened 2020-12-16
Turns off clang formatter for H5ARG_TRACE

derobins:
The H5ARG_TRACE macro was added during the async merge. These macros
are updated by the bin/trace script, but conflict with the formatter
so formatted code gets dirtied when autogen.sh is run.

This change enshrines the bin/trace output as canonical and adds
comments to turn clang formatting off for those lines containing
H5ARG_TRACE.

### DIFF
--- a/src/H5A.c
+++ b/src/H5A.c
@@ -274,9 +274,10 @@ H5Acreate_async(const char *app_file, const char *app_func, unsigned app_line, h
 
     /* If a token was created, add the token to the event set */
     if (NULL != token)
+        /* clang-format off */
         if (H5ES_insert(es_id, vol_obj->connector, token,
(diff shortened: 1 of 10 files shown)


lrknox (approved):

byrnHDF (approved):

qkoziol (approved):
Ugly (since it requires each occurrence of an H5ARG_TRACE macro to be tagged), but workable.

derobins:
> Ugly (since it requires each occurrence of an H5ARG_TRACE macro to be tagged), but workable.

Yes, but it beats the "formatter v. bin/trace" war. Dirtying the repo by just running autogen.sh is madness and will create endless conflicts. We can revisit for 1.13.1 (and I'll create a JIRA issue for it).

PR #194, opened 2020-12-17
Update version to 1.8.23-1 after creating release branch for 1.8.22.

lrknox:


derobins:
It looks like you missed some version things...

lrknox:
> It looks like you missed some version things...

the '-' was missing in the full version string in H5public.h, due to a bin/h5vers deficiency for unanticipated value '0' for SUBRELEASE.  I added the '-', but h5vers will need a slight modification.  Otherwise I don;t see anything missing.  The "-0"  is omitted from the ddl files by design, and the "snap" was intentionally dropped.

src/H5public.h
```diff
@@ -94,10 +94,10 @@ extern "C" {
 /* Version numbers */
 #define H5_VERS_MAJOR      1                               /* For major interface/format changes           */
 #define H5_VERS_MINOR      8                               /* For minor interface/format changes           */
-#define H5_VERS_RELEASE    22                              /* For tweaks, bug-fixes, or development     */
-#define H5_VERS_SUBRELEASE "snap12"                        /* For pre-releases like snap0       */
+#define H5_VERS_RELEASE    23                              /* For tweaks, bug-fixes, or development     */
+#define H5_VERS_SUBRELEASE "0"                        /* For pre-releases like snap0       */
```
byrnHDF: I believe this is the issue with '0', because H5_VERS_SUBRELEASE is evaluated as nothing.

lrknox:
So I guess we always need to start with version 1.x.x-1 then?

byrnHDF (approved):

PR #202, opened 2020-12-18
Brings changes to ttsafe_attr_vlen.c from develop

derobins:
Fix for exposed pthreads code that causes Windows builds to fail.

### DIFF
--- a/test/ttsafe_attr_vlen.c
+++ b/test/ttsafe_attr_vlen.c
@@ -6,7 +6,7 @@
  * This file is part of HDF5.  The full HDF5 copyright notice, including     *
  * terms governing use, modification, and redistribution, is contained in    *
  * the COPYING file, which can be found at the root of the source code       *
- * distribution tree, or in https://support.hdfgroup.org/ftp/HDF5/releases.  *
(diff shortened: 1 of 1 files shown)


test/ttsafe_attr_vlen.c
```diff
@@ -6,7 +6,7 @@
  * This file is part of HDF5.  The full HDF5 copyright notice, including     *
  * terms governing use, modification, and redistribution, is contained in    *
  * the COPYING file, which can be found at the root of the source code       *
- * distribution tree, or in https://www.hdfgroup.org/licenses.               *
+ * distribution tree, or in https://support.hdfgroup.org/ftp/HDF5/releases.  *
```
derobins: Should I change this back? I just copied the entire file over from develop.

lrknox: Yes, it should be the line ending in "licenses."  I'll be updating develop in the near future.

lrknox (approved):

test/ttsafe_attr_vlen.c
derobins: Fixed

PR #207, opened 2020-12-19
Fixes a small memory leak in the h5jam test

derobins:


lrknox (approved):

tools/test/h5jam/getub.c
```diff
@@ -138,14 +138,15 @@ main(int argc, const char *argv[])
...
     HDfree(buf);
     HDclose(fd);
 
     return EXIT_SUCCESS;
 
 error:
-    if (buf)
-        HDfree(buf);
+    HDfree(filename);
+    HDfree(buf);
```
qkoziol, line 149: HDfree doesn't protect against NULL pointers, so it would be best to leave those in for the error situations.

qkoziol (approved):

qkoziol:
(Such a minor deal that I'm going to approve, but not terribly important to fix)

tools/test/h5jam/getub.c
derobins: The C standard says that freeing a NULL pointer is fine so checking for NULL isn't necessary.

PR #208, opened 2020-12-19
Fixes a small memory leak in the chunk_cache test

derobins:


lrknox (approved):

qkoziol:
It would be good to protect the HDfree() calls here also.

qkoziol (approved):

derobins:
The C standard says that freeing a NULL pointer is fine so checking for NULL isn't necessary.

PR #209, opened 2020-12-19
Fixes a minor memory leak in hl/test/test_file_image.c

derobins:


lrknox (approved):

hl/test/test_file_image.c
```diff
@@ -506,6 +506,8 @@ test_file_image(size_t open_images, size_t nflags, unsigned *flags)
     } /* end for */
 
     /* release temporary working buffers */
+    for (i = 0; i < open_images; i++)
```
qkoziol, line 509: You may consider protecting all of these.

qkoziol (approved):

hl/test/test_file_image.c
derobins: The C standard says that freeing a NULL pointer is fine so checking for NULL isn't necessary.

PR #217, opened 2020-12-21
Add compiler flags for GCC 10, along with updating warnhist script to accommodate them

qkoziol:


config/gnu-warnings/developer-10
```diff
@@ -0,0 +1,6 @@
+# New warning
+-Warith-conversion
+
+# Enable static analysis of program flow
+-fanalyzer
+-fdiagnostics-path-format=none
```
derobins, line 6: This is fine here for now, but I plan to add gcc sanitizer support and we may want to move these there when that functionality is available.

derobins (approved):

lrknox (approved):

PR #231, opened 2020-12-26
Hdf5 1 12 - Fix unaligned access to reference buffer during datatype conversion on 64-bit Solaris 

lrknox:


lrknox:
Bring fix for fixing unaligned access to reference buffer during datatype conversion on 64-bit Solaris to hdf5_1_12.

qkoziol (approved):

PR #254, opened 2021-01-12
Reverts lock/unlock callback signature to 1.8.21 version

derobins:
This callback is unused in 1.8. The ros3 and hdfs VFDs are the only VFDs
that have the lock callback implemented and that is just as no-op stubs.
These stubs were removed so the callbacks are now NULL pointers, like
the other VFDs in 1.8.

### DIFF
--- a/src/H5FDpublic.h
+++ b/src/H5FDpublic.h
@@ -236,8 +236,8 @@ typedef struct H5FD_class_t {
     herr_t (*write)(H5FD_t *file, H5FD_mem_t type, hid_t dxpl, haddr_t addr, size_t size, const void *buffer);
     herr_t (*flush)(H5FD_t *file, hid_t dxpl_id, unsigned closing);
     herr_t (*truncate)(H5FD_t *file, hid_t dxpl_id, hbool_t closing);
-    herr_t (*lock)(H5FD_t *file, hbool_t last);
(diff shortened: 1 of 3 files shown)


src/H5FDpublic.h
```diff
@@ -236,8 +236,8 @@ typedef struct H5FD_class_t {
     herr_t (*write)(H5FD_t *file, H5FD_mem_t type, hid_t dxpl, haddr_t addr, size_t size, const void *buffer);
     herr_t (*flush)(H5FD_t *file, hid_t dxpl_id, unsigned closing);
     herr_t (*truncate)(H5FD_t *file, hid_t dxpl_id, hbool_t closing);
-    herr_t (*lock)(H5FD_t *file, hbool_t last);
-    herr_t (*unlock)(H5FD_t *file);
+    herr_t  (*lock)(H5FD_t *file, unsigned char *oid, unsigned lock_type, hbool_t last);
```
byrnHDF: Whitespace?

derobins: Sorry, didn't run it through the formatter. I just copied it in.

derobins: I'll update in a sec.

derobins: Fixed

soumagne (approved):

lrknox (approved):

PR #260, opened 2021-01-14
Fixes naked pthread usage in H5TS code added for async

derobins:


derobins:
The real fix is around line 615, where the async dump brought in naked pthread usage. The rest of the changes move the "done" target inside the POSIX ifdef to quiet warnings on Windows.

byrnHDF (approved):

qkoziol (approved):

PR #262, opened 2021-01-14
Fixes a wayward parenthesis in H5TS.c

derobins:


derobins:
H5TS.c gets dirtied every time I run autogen.sh. It's driving me NUTS.

qkoziol:
The code formatter will just put it back the other way... ;-)

derobins:
> The code formatter will just put it back the other way... ;-)

Then the formatter is wrong and should change. It's completely ridiculous to have autogen dirty the repo. This has happened often enough to where I'm ready to cobble a script together that makes sure autogen.sh + git diff is clean before allowing a commit.

derobins:
> The code formatter will just put it back the other way... ;-)

Also, nobody appears to be running the formatter. I just ran it over develop and 150 files changed. (This may be due to my version of LLVM, but version-specific formatting is _another_ problem that we could have)

lrknox (approved):

byrnHDF (approved):

qkoziol (approved):

PR #270, opened 2021-01-19
Improve performance of multiple calls to H5Sget_select_elem_pointlist

fortnern:
It was found that, when making repeated calls to H5Sget_select_elem_pointlist to retrieve points, the point list traversal was extremely slow. This should fix that problem for common use cases.

    Cache the pointer to the next point to process after the last call to
    H5S__get_select_elem_pointlist.  This allows the normal process of
    iterating over the points in batches to be much more efficient, as the
    library does not need to traverse the entirety of the preceding points
    every time the funciton is re-entered.


soumagne (approved):

qkoziol:
This is OK, although duplicative in the sense that H5S_point_iter_t struct also does this, for selection iterators.   It's probably fine for now, but could be refactored to combine the two ideas at some point in the future.

qkoziol (approved):

lrknox (approved):

PR #274, opened 2021-01-22
Fix problems with vlens and refs inside compound using H5VLget_file_type()

fortnern:
Modified library to properly ref count H5VL_object_t structs and only consider file vlen and reference types to be equal if their files are the same. Should prevent errors due to files being closed out from under cached types in the type conversion path table. There is still a potential error if the VOL connector closes the underlying file object before closing the datatype returned from H5VLget_file_type().


src/H5T.c
```diff
@@ -4473,6 +4492,31 @@ H5T_get_size(const H5T_t *dt)
...
+ */
+hbool_t
+H5T_get_force_conv(const H5T_t *dt)
+{
+    /* Use FUNC_ENTER_NOAPI_NOINIT_NOERR here to avoid performance issues */
+    FUNC_ENTER_NOAPI_NOINIT_NOERR
+
+    /* check args */
+    HDassert(dt);
+
```
derobins, line 4519: Might also want to add HDassert(dt->shared)

derobins (approved):

soumagne (approved):

src/H5T.c
fortnern: Thanks, done

src/H5T.c
```diff
@@ -3489,18 +3493,24 @@ H5T__initiate_copy(const H5T_t *old_dt)
     /* Copy shared information */
     *(new_dt->shared) = *(old_dt->shared);
 
-    /* Reset VOL fields */
+    /* Increment ref count on owned VOL object */
+    if (new_dt->shared->owned_vol_obj)
+        new_dt->shared->owned_vol_obj->rc++;
```
qkoziol: Having other parts of the library modify fields in the H5VL structs is going to be difficult to maintain.   Could you please add a "increment refcount" routine in the H5VL code and call that?

src/H5T.c
```diff
@@ -6295,6 +6349,7 @@ H5T_own_vol_obj(H5T_t *dt, H5VL_object_t *vol_obj)
 
     /* Take ownership */
     dt->shared->owned_vol_obj = vol_obj;
+    vol_obj->rc++;
```
qkoziol: (Another refcount increment for a H5VL routine)

src/H5Tcommit.c
```diff
@@ -148,6 +148,7 @@ H5T__commit_api_common(hid_t loc_id, const char *name, hid_t type_id, hid_t lcpl
     new_obj->connector = (*vol_obj_ptr)->connector;
     new_obj->connector->nrefs++;
     new_obj->data = data;
+    new_obj->rc = 1;
```
qkoziol: Same with this piece of code (which I admit wasn't originally written by you) - could you please add a small routine in the H5VL code that encapsulated this setup and then call it here (and down below in this routine)?

src/H5Tdeprec.c
```diff
@@ -139,6 +139,7 @@ H5Tcommit1(hid_t loc_id, const char *name, hid_t type_id)
     new_obj->connector = vol_obj->connector;
     new_obj->connector->nrefs++;
     new_obj->data = data;
+    new_obj->rc = 1;
```
qkoziol: (Another candidate for the "setup" H5VL setup routine)

src/H5VLprivate.h
```diff
@@ -37,6 +37,7 @@ typedef struct H5VL_t {
 typedef struct H5VL_object_t {
     void *  data;      /* Pointer to connector-managed data for this object    */
     H5VL_t *connector; /* Pointer to VOL connector struct                      */
+    hsize_t rc;        /* Reference count                                      */
```
qkoziol: This should be a 'size_t', since you are counting objects in memory.

qkoziol:
This is a very nice change and if you could spend 15 minutes with a little refactoring, would be improved further.  :-)

src/H5T.c
fortnern: I can do this, though looking through the code there are many other places that directly manipulate the H5VL_object_t struct outside the H5VL package

src/H5Tcommit.c
fortnern: I can replace this with H5VL_create_object, though again there are still other places in the library that peek into H5VL_object_t. In fact this code would still need to do that ((*vol_obj_ptr)->connector), so we would need to write another routine to retrieve that, then propagate it to everywhere that currently gets the connector directly, which includes many (all?) of the *_api_common() routines.

fortnern: Are you thinking we should move H5VL_object_t to H5VLpkg.h? This is kind of outside the scope of this change.

src/H5T.c
qkoziol: No doubt!   I would suggest that you just herd as many cats as you have time for now and we'll keep tracking them down over time.

src/H5Tcommit.c
qkoziol: Agree - I would volunteer to pick up more of the changes like this in another PR, based on the start you make here.

qkoziol: > Are you thinking we should move H5VL_object_t to H5VLpkg.h? This is kind of outside the scope of this change.

I don't think that's necessary (yet?) and agree, would be far outside of the scope for your PR.

fortnern:
@qkoziol I think I've addressed your concerns.

qkoziol (approved):
Yes - looks great, thanks!

lrknox (approved):

PR #289, opened 2021-01-27
[DO NOT MERGE]: WIP: added Fortran optional parameter for async

brtnfld:
H5Acreate example using an optional arg for async.

Does not work because it calls the varg version, *_wrap does not show up in library.


brtnfld:
Going with adding new APIs instead.

PR #290, opened 2021-01-27
[DO NOT MERGE} : WIP: added async Fortran  function example

brtnfld:
H5Acreate example using a new async API

Does not work because it calls the varg version, *_wrap does not show up in library. A lot of duplicate code.


epourmal (commented):
I miss Crucible! Maybe I don't know how to use this one... 
Can we create an INTERFACE to a real C function and not return the first three parameters at all, or return them in optional arguments for h5acreate_async_f?

fortran/src/H5Aff.F90
```diff
@@ -200,6 +200,82 @@ END FUNCTION H5Acreate2
...
+                                           ! 0 on success and -1 on failure
+!*****
+    INTEGER(HID_T), OPTIONAL, INTENT(IN) :: acpl_id ! Attribute creation property list identifier
+    INTEGER(HID_T), OPTIONAL, INTENT(IN) :: aapl_id ! Attribute access property list identifier
+
+    INTEGER(HID_T) :: acpl_id_default
+    INTEGER(HID_T) :: aapl_id_default
+    CHARACTER(LEN=LEN_TRIM(name)+1,KIND=C_CHAR) :: c_name
+    INTERFACE
+       INTEGER(HID_T) FUNCTION H5Acreate_async(loc_id, name, type_id, &
```
epourmal: Ok, it is my browser that is slow. Now it works and I can leave a comment for the line.

So the function in the interface block will have the first three parameters that real C function has?

brtnfld:
I'm not sure what you mean by a real C function. It looks to me that the only C function available has the vargs in the signature. We could have the Fortran wrapper call a C wrapper that calls the C varg function, and then does not return the 3 extra parameters to the Fortran wrapper.   The 3 extra parameters would be meaningless for this case since it would be the information from the C wrapper.  But then we are back to having to call C wrappers from Fortran. Maybe this is not a huge deal since we are still calling a lot of C wrappers since they have not yet be converted over.

We can't create an interface to a C function that has VARGs since it is erroneous to call C varargs function from Fortran. From the Fortran standard 2018, 18.3.6:

A Fortran procedure interface is interoperable with a C function prototype if...
...
(7)   the prototype does not have variable arguments as denoted by the ellipsis (...)

That leaves us only with the option of using C wrappers, unless I'm missing something vargs in C.

brtnfld:
I added the file, func, line parameters as optional in the Fortran APIs, and then call the C API directly using those arguments if they are present.

https://github.com/HDFGroup/hdf5/pull/290/files


If we want to get fancy (and the compiler supports it), we can allocate the default strings instead of specifying a length for the strings.  They are currently set for the max filename length permitted in Linux.

What does the C API do if file and func are NULL (meaning they just have a NULL terminator)? Do I need to pass a null pointer instead?

__FUNC__ is not supported in Fortran, so the user will have to pass it manually if needed.I added an example of what the call looks like from Fortran.

Scot

epourmal (commented):
Do I understand correctly  that sync version becomes a "wrapper" around async? I like it and cannot see any problems except that FORTRAN APIs will now depend on async C functions that ideally shouldn't be in the core library. But as Quincey mentioned, it would be hard to separate async functions into its own module that can be compiled in/out.
How much overhead we will created by calling another level of subroutines?

fortran/src/H5Aff.F90
brtnfld: I also had that concern about async, but I'm assuming they are here to stay and can not be compiled out.  If we don't think that is the case, and we want to future-proof it, then I will have to change it to use a common code like Quincey.

I don't think there is much overhead, I am more concerned about what the overhead is of calling C_async APIs instead of calling sync API directly. Maybe that is a better case for using a common subroutine instead?

Yes, we will need to always include the 3 extra parameters when calling the C APIs.

PR #291, opened 2021-01-28
Allow using virtual datasets in files opened with H5F_CLOSE_SEMI and H5F_CLOSE_STRONG

fortnern:
Note that this will still cause issues if the user tries to open a source file with a close degree that is not H5F_CLOSE_WEAK while it is simultaneously in use by the VDS.


qkoziol:
Why did you need to switch the test error macros?

qkoziol:
Also, why is there a tmp_nerrors variable?   Are some of the tests expected to fail?

fortnern:
> Why did you need to switch the test error macros?

I am using them to hide the majority of the output, so instead of listing every I/O test for every configuration, it just says "Testing virtual dataset I/O" (once for each libver_bounds setting). If a test fails it prints the failed config and test. If VDS_TEST_VERBOSE is defined you get the whole 1800 line output.

> Also, why is there a tmp_nerrors variable? Are some of the tests expected to fail?

It's used to ensure it only prints "PASSED" if all of the I/O tests pass.

fortnern:
Anyone have time to review this?

qkoziol:
> Anyone have time to review this?

Sorry, was pulled off to other urgent things.

Looks fine to me, although unfortunate about the revision to the macros. :-/

qkoziol (approved):

lrknox (approved):

PR #295, opened 2021-01-29
Fixes bin/trace so it doesn't dirty the repo when autogen runs

derobins:


bin/trace
```diff
@@ -207,7 +207,7 @@ $Source = "";
 # Maximum length of H5TRACE macro line
 # If the ColumnLimit in .clang-format is changed, this value will need to be updated
 #
-my $max_trace_macro_line_len = 110;
+my $max_trace_macro_line_len = 109;
```
derobins: This line length keeps bin/trace and bin/format_source from fighting over long TRACE macros

bin/trace
```diff
@@ -470,7 +470,7 @@ for $file (@ARGV) {
   $file_args = 0;
 
   # Ignore some files that do not need tracing macros
-  unless ($file eq "H5FDmulti.c" or $file eq "src/H5FDmulti.c" or $file eq "H5FDstdio.c" or $file eq "src/H5FDstdio.c") {
+  unless ($file eq "H5FDmulti.c" or $file eq "src/H5FDmulti.c" or $file eq "H5FDstdio.c" or $file eq "src/H5FDstdio.c" or $file eq "src/H5TS.c") {
```
derobins, line 473: Nothing in H5TS.c has a TRACE macro and this is easier than screwing with bin/trace to get it to stop messing with the file.

byrnHDF (approved):

derobins:
Why is this failing the format checks?

bin/trace
soumagne: I feel like there's a +1/-1 that's not respected somewhere then?

derobins:
ARGH. NOW IT'S FIGHTING OVER ANOTHER FILE.

bin/trace
derobins: Right, but I'm unexcited about digging through a rat's nest of Perl to figure out where it's going wrong. I'm just going to turn the formatter off for that one line.

lrknox (approved):

bin/trace
qkoziol: This is not a good solution.   Those other files are the "public" examples and H5TS.c is not.

qkoziol (changes requested):
I don't think this is the correct direction to go.   The bin/trace script needs to be updated, not hacks in other files.

derobins:
@qkoziol The bin/trace issues are due to your checkins, so if you want to do a better job at fixing the problems, have at it. I'm just tired of having to engineer around this nonsense. autogen.sh should NEVER dirty the repo on a fresh checkout. Allowing that is insanity.

derobins:
Between this and all the bad format decisions clang-format makes, I'm of a mind to get rid of the formatter altogether.

bin/trace
derobins: H5TS has no calls that require the TRACE macro and I have zero interest in debugging the trace script to fix its processing.

qkoziol:
> @qkoziol The bin/trace issues are due to your checkins, so if you want to do a better job at fixing the problems, have at it. I'm just tired of having to engineer around this nonsense. autogen.sh should NEVER dirty the repo on a fresh checkout. Allowing that is insanity.

Yeah, I can look at bin/trace and try to get it to stop arguing with clang-format.   Especially since it's the component that we have actual control over...

lrknox:
I'm going to merge this to stop continuing issues with other PRs. This PR makes just a few changes which can be reverted when a better solution is available.

PR #308, opened 2021-02-03
Unwrap file VOL object when destroying file

jhendersonHDF:
During a native file "post open" call, a file VOL object is created, wrapped through the VOL stack and then set on the H5F_t struct. When destroying the file, the VOL object was being freed without being unwrapped first, leading to memory leaks when passthrough connectors are stacked on top of the native connector.

### DIFF
--- a/src/H5Fint.c
+++ b/src/H5Fint.c
@@ -247,7 +247,8 @@ H5F__close_cb(H5VL_object_t *file_vol_obj, void **request)
     if (H5VL_file_close(file_vol_obj, H5P_DATASET_XFER_DEFAULT, request) < 0)
         HGOTO_ERROR(H5E_FILE, H5E_CANTCLOSEFILE, FAIL, "unable to close file")
 
-    /* Free the VOL object */
+    /* Free the VOL object; it is unnecessary to unwrap the VOL
+     * object before freeing it, as the object was not wrapped */
     if (H5VL_free_object(file_vol_obj) < 0)
         HGOTO_ERROR(H5E_FILE, H5E_CANTDEC, FAIL, "unable to free VOL object")
 
@@ -1320,6 +1321,8 @@ H5F__new(H5F_shared_t *shared, unsigned flags, hid_t fcpl_id, hid_t fapl_id, H5F
 
 done:
     if (!ret_value && f) {
+        HDassert(NULL == f->vol_obj);
+
         if (!shared) {
             /* Attempt to clean up some of the shared file structures */
             if (f->shared->efc)
@@ -1331,11 +1334,6 @@ done:
 
             f->shared = H5FL_FREE(H5F_shared_t, f->shared);
         }
-
-        /* Free VOL object */
-        if (f->vol_obj)
-            if (H5VL_free_object(f->vol_obj) < 0)
-                HDONE_ERROR(H5E_FILE, H5E_CANTDEC, NULL, "unable to free VOL object")
 
         f = H5FL_FREE(H5F_t, f);
     }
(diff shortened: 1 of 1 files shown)


src/H5Fint.c
```diff
@@ -1333,9 +1333,12 @@ H5F__new(H5F_shared_t *shared, unsigned flags, hid_t fcpl_id, hid_t fapl_id, H5F
         }
 
         /* Free VOL object */
-        if (f->vol_obj)
+        if (f->vol_obj) {
+            if (NULL == H5VL_object_unwrap(f->vol_obj))
```
jhendersonHDF: I believe that this is the right solution here. While the call to H5VL_object_unwrap returns a VOL object that just gets discarded here, it is still important to make sure the VOL object gets unwrapped. It seems this is a one-off case that requires this specifically. Putting the unwrap logic into H5VL_free_object seems inappropriate since every usage I can find in the library seems to already be dealing with an unwrapped object, so the unwrap logic would just be unnecessary overhead.

qkoziol: I don't think this is the correct fix - it doesn't seem that f->vol_obj would be non-NULL here.   The f->vol_obj value is set in the "post open" code, which is invoked after H5F_new.

jhendersonHDF: This code is already surrounded by an existing `if (f->vol_obj)` check though, so we know that it would be non-NULL if it reaches here. Like you mention though, f->vol_obj will probably never be non-NULL here, so it's probably unnecessary to do the VOL object unwrapping during H5F__new. I can revert that back to the existing code, even though that probably will never be reached either. It's definitely needed inside H5F__dest though.

qkoziol: OK, cool.   Maybe take this out and replace it with "HDassert(NULL == f->vol_obj);" ?

src/H5Fint.c
```diff
@@ -1603,8 +1606,12 @@ H5F__dest(H5F_t *f, hbool_t flush)
     /* Free the non-shared part of the file */
     f->open_name   = (char *)H5MM_xfree(f->open_name);
     f->actual_name = (char *)H5MM_xfree(f->actual_name);
-    if (f->vol_obj && H5VL_free_object(f->vol_obj) < 0)
-        HDONE_ERROR(H5E_FILE, H5E_CANTDEC, FAIL, "unable to free VOL object")
+    if (f->vol_obj) {
```
qkoziol, line 1604: Agree - this is necessary.   (Although moving the "f->vol_obj = NULL;" into the new 'if' block would be slightly cleaner.

qkoziol: (You may also want to leave a comment in H5F__close_cb() that unwrapping the object before calling H5VL_free_object is unnecessary since it wasn't wrapped)

src/H5Fint.c
jhendersonHDF: Done.

src/H5Fint.c
jhendersonHDF: Also done. Hopefully this is the only place where this weird one-off unwrapping is needed. It seems like everywhere else that H5VL_free_object is used does the correct thing, but there might be more unwrapping issues to be discovered when passthrough connectors do complicated things.

qkoziol (approved):
Looks great now!

lrknox (approved):

qkoziol:
This is definitely an amusing commit - it’s exposed a bug in the async VOL connector also.   Tang’s working on a fix this afternoon, and I’ll post here when it’s pushed to the hpc-io/vol-async repo.

PR #318, opened 2021-02-07
Remove prohibition against building Universal Binaries on macOS

seanm:


seanm:
@epourmal

byrnHDF (approved):

lrknox (approved):

seanm:
Ideally this would be backported to earlier branches too.  Personally, I use 1.10.x and would most like to see it there (and newer).

seanm:
@imikejackson

byrnHDF:
Should there be a RELEASE.txt entry?

byrnHDF (commented):
Also, is the user expected to know how to build a universal binary and when not to?

seanm:
> Also, is the user expected to know how to build a universal binary and when not to?

Depending on who you mean by "user", then yes, any Mac programmer building HDF5 would know when/if he wants to build universal.  Now that Apple is selling both Intel and ARM Macs, the 'when' is usually "always".

imikejackson:
For what it's worth I'm ok with these changes.

I also wouldn't mind seeing this back ported to the 1.10.x branch as that is what we are using in our projects.

byrnHDF (changes requested):
I think we do need to have a release note entry about this.
Just something simple that dates the change and why.

qkoziol (approved):

byrnHDF:
Once the change is in develop, I can make the merges to 1.12 and 1.10.
Allen

seanm:
OK, I'll amend this with a note in the release notes...

seanm:
Where should the note go?  In RELEASE.txt?

byrnHDF:
Correct, newest at top of section Configuration. I would use the "New Features"  Configuration section rather then the "Bug Fixes".

release_docs/RELEASE.txt
```diff
@@ -48,6 +48,11 @@ New Features
 
     Configuration:
     -------------
+    - On macOS, Universal Binaries can now be built, allowing native execution on
+      both Intel and Apple Silicon (ARM) based Macs.
+
+      To do so, set CMAKE_OSX_ARCHITECTURES="x86_64;arm64"
+
```
byrnHDF, line 55: One more line - initials and date, please.

byrnHDF (approved):

PR #324, opened 2021-02-08
Hdf5 1 10

bmribler:


src/H5Ochunk.c
```diff
@@ -412,12 +412,13 @@ H5O__chunk_dest(H5O_chunk_proxy_t *chk_proxy)
     HDassert(chk_proxy);
 
     /* Decrement reference count of object header */
-    if (chk_proxy->oh && H5O__dec_rc(chk_proxy->oh) < 0)
+    if (H5O__dec_rc(chk_proxy->oh) < 0){
```
lrknox: format check failed  1. for lack of space between ){ above, and 2. because closing } below should be  indented 4 spaces.  However, are those braces necessary at all?

lrknox: format check still fails because 2. closing '}' below needs to be indented 4 spaces.

bmribler: Weird, my local version doesn't have those...  I made another commit, we'll see how it shows up.

byrnHDF (approved):

src/H5Ochunk.c
bmribler: Note: The user (Mathworks) did not provide the corrupted file. I couldn't craft a file that would cause segfault on this.

PR #325, opened 2021-02-09
Check for wrap context before unwrapping file VOL object in H5F__dest

jhendersonHDF:
In the EFC test, H5F__efc_open is called directly, which has the side effect of not setting the VOL wrapping context and thus not wrapping the file VOL object. Since there doesn't seem to be a reasonable way to set the VOL wrap context (no VOL object is available) in H5F__efc_open, we simply check that a VOL wrapping context is available before trying to unwrap the file VOL object. There is still a failure with the passthrough connector and the file image tests, but that seems to have existed before the file VOL object unwrapping changes.


lrknox (approved):

qkoziol (commented):
Looks reasonable

PR #332, opened 2021-02-12
Update license url

lrknox:


lrknox:
Github wouldn't let me create a PR with this change in all 1702 files, so I've updated files in src and test, and I'll create a second PR to update the other 1000 files.

byrnHDF (approved):

Issue #311, opened 2021-02-04
Support 'Universal Binaries' on macOS


seanm:
As when Macs previously changed from PowerPC to Intel, Mac developers these days again need to support 'Universal Binaries', which are basically libraries/executables that contain both Intel and ARM code.

Back in the PowerPC+Intel days, building HDF5 as Universal Binaries worked fine, for me at least.

But now the current code in the develop branch, and presumably other branches, contains in `config/cmake_ext_mod/ConfigureChecks.cmake` this:

```
if (APPLE)
  list (LENGTH CMAKE_OSX_ARCHITECTURES ARCH_LENGTH)
  if (ARCH_LENGTH GREATER 1)
    set (CMAKE_OSX_ARCHITECTURES "" CACHE STRING "" FORCE)
    message (FATAL_ERROR "Building Universal Binaries on OS X is NOT supported by the HDF5 project. This is"
    "due to technical reasons. The best approach would be build each architecture in separate directories"
    "and use the 'lipo' tool to combine them into a single executable or library. The 'CMAKE_OSX_ARCHITECTURES'"
    "variable has been set to a blank value which will build the default architecture for this system.")
  endif ()
  set (${HDF_PREFIX}_AC_APPLE_UNIVERSAL_BUILD 0)
endif ()
```

It's here: https://github.com/HDFGroup/hdf5/blob/develop/config/cmake_ext_mod/ConfigureChecks.cmake#L28

This is deliberately thwarting Universal builds, for some 'technical reasons' according to the text... but what are the reasons? How can we overcome them to get universal builds working?

See also:
https://forum.hdfgroup.org/t/apple-silicon-binaries-universal-binaries/7848

Thanks.

byrnHDF:
Creating merges to 1.12 and 1.10 now. PR #344

byrnHDF:
Also in 1.8 PR.

byrnHDF:
Change in all branches.

PR #335, opened 2021-02-15
Detect correct long double size with universal binaires

seanm:
On macOS on x86_64, long double is 16 bytes; on macOS on arm64, it's 8 bytes.

Use the same existing technique used to properly detect other sizes to now also properly detect the size of long double.

For bug #311.


seanm:
@qkoziol @epourmal @lrknox

byrnHDF (approved):

lrknox (approved):

seanm:

PR #336, opened 2021-02-15
Fix cmake syntax error with regex escaping

seanm:
Fixes:

   Syntax error in cmake code at

     /some/path/hdf5/CMakeLists.txt:240

   when parsing string

     .*#define[ \t]+H5_VERS_SUBRELEASE[ \t]+\"([0-9A-Za-z._\-]*)\".*$

   Invalid escape sequence \-

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,7 +237,7 @@ string (REGEX REPLACE ".*#define[ \t]+H5_VERS_MINOR[ \t]+([0-9]*).*$"
     "\\1" H5_VERS_MINOR ${_h5public_h_contents})
 string (REGEX REPLACE ".*#define[ \t]+H5_VERS_RELEASE[ \t]+([0-9]*).*$"
     "\\1" H5_VERS_RELEASE ${_h5public_h_contents})
-string (REGEX REPLACE ".*#define[ \t]+H5_VERS_SUBRELEASE[ \t]+\"([0-9A-Za-z._\-]*)\".*$"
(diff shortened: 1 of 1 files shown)


seanm:
@qkoziol @epourmal @lrknox

CMakeLists.txt
```diff
@@ -237,7 +237,7 @@ string (REGEX REPLACE ".*#define[ \t]+H5_VERS_MINOR[ \t]+([0-9]*).*$"
     "\\1" H5_VERS_MINOR ${_h5public_h_contents})
 string (REGEX REPLACE ".*#define[ \t]+H5_VERS_RELEASE[ \t]+([0-9]*).*$"
     "\\1" H5_VERS_RELEASE ${_h5public_h_contents})
-string (REGEX REPLACE ".*#define[ \t]+H5_VERS_SUBRELEASE[ \t]+\"([0-9A-Za-z._\-]*)\".*$"
```
byrnHDF, line 240: I believe the slash was added to escape the dash because of a problem on a test machine. Since this is a cmake command, we will need to find that machine or find that version of cmake.

seanm: The version of cmake that gives the error is 3.19.4 BTW.

byrnHDF (approved):
I found where this was added, it seems that the escape "\" char was part of the original fix. So, based on research, I think we should go ahead and make the change.

lrknox (approved):

seanm:

PR #337, opened 2021-02-15
Fix clang unknown pragma warnings

seanm:
Although clang defines __GNUC__ and supports most gcc warnings, it doesn't support all, and will in fact even warn about an unknown pragma.

Guard appropriately.


seanm:
@qkoziol @epourmal @lrknox

byrnHDF (changes requested):
I believe this file is generated and the change needs to be in the ".l" file somewhere.

byrnHDF:
> I believe this file is generated and the change needs to be in the ".l" file somewhere.

Correction this group of code is in the bin/genparser file.

seanm:
Thanks @byrnHDF I think my new commit gets them all.

byrnHDF (approved):

lrknox (approved):

seanm:

PR #338, opened 2021-02-15
Fix undefined left shifting of negative numbers

seanm:
Undefined Bahavior Sanitizer errored here about left shifting negative numbers.

Since the numbers are random anyway, I figure they can just be negated.

### DIFF
--- a/test/dsets.c
+++ b/test/dsets.c
@@ -4041,7 +4041,7 @@ test_nbit_compound_2(hid_t file)
             power               = HDpow(2.0F, (double)(precision[1] - 1));
             orig_data[i][j].a.c = (char)(((long long)HDrandom() % (long long)power) << offset[1]);
             power               = HDpow(2.0F, (double)(precision[2] - 1));
-            orig_data[i][j].a.s = (short)(-((long long)HDrandom() % (long long)power) << offset[2]);
+            orig_data[i][j].a.s = (short)(-(((long long)HDrandom() % (long long)power) << offset[2]));
             orig_data[i][j].a.f = float_val[i][j];
 
             power             = HDpow(2.0F, (double)precision[3]);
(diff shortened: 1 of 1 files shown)


seanm:
@qkoziol @epourmal @lrknox I have no idea if this is the right fix...

It's the only problem identified by UBSan anymore, so that's nice.

test/dsets.c
```diff
@@ -4041,7 +4041,7 @@ test_nbit_compound_2(hid_t file)
             power               = HDpow(2.0F, (double)(precision[1] - 1));
             orig_data[i][j].a.c = (char)(((long long)HDrandom() % (long long)power) << offset[1]);
             power               = HDpow(2.0F, (double)(precision[2] - 1));
-            orig_data[i][j].a.s = (short)(-((long long)HDrandom() % (long long)power) << offset[2]);
+            orig_data[i][j].a.s = (short)(((long long)HDrandom() % (long long)power) << offset[2]);
```
qkoziol: Hard to tell from the current code what would be best.  :-/  I would suggest keeping the negative sign and putting an extra pair of parentheses around the value that's been left-shifted.   That would keep a negative value in the test and should clear the warning up.

test/dsets.c
```diff
@@ -4057,7 +4057,7 @@ test_nbit_compound_2(hid_t file)
                 for (n = 0; n < (size_t)array_dims[1]; n++) {
                     power = HDpow(2.0F, (double)(precision[0] - 1));
                     orig_data[i][j].d[m][n].i =
-                        (int)(-((long long)HDrandom() % (long long)power) << offset[0]);
+                        (int)(((long long)HDrandom() % (long long)power) << offset[0]);
```
qkoziol: [Same here]

qkoziol (changes requested):
A little difficult to discern the correct course of action, but I've suggested a change that should keep the spirit of the test and alleviate the warning.

test/dsets.c
seanm: I'm not sure I understand your suggestion. Extra parens where?

Unlike many of my other patches, this is not for a compilation warning, it's a runtime bug.  In C, `x << y` is (obnoxiously) undefined behaviour if `x < 0`. Decent explanation: https://stackoverflow.com/questions/36751538/ubsan-reports-875-7-as-undefined-behaviour

I suppose we could cast the negative value to unsigned, that would probably preserve the behaviour the author expected.

qkoziol: Basically: instead of (short)(- x << y), how about (short)(- (x << y))

seanm: Oh I see!  Will verify...

qkoziol (approved):

kyang2014:
The change is fine. At least on Linux, the random() should always return a positive number. So (((long long)HDrandom() % (long long)power) << offset[2]) is supposed to be positive, which avoids the undefined behavior issue.

kyang2014 (approved):

lrknox (approved):

PR #340, opened 2021-02-15
Fixed various clang -Wformat-security warnings

seanm:


tools/src/h5import/h5import.c
```diff
@@ -2536,7 +2536,7 @@ parsePathInfo(struct path_info *path, char *temp)
 
     token = HDstrtok(temp, delimiter);
     if (HDstrlen(token) >= MAX_PATH_NAME_LENGTH) {
-        (void)HDfprintf(stderr, err1);
+        (void)HDfprintf(stderr, "%s", err1);
```
byrnHDF, line 2539: The problem here is that there is missing data, the err1 is:
    const char *err1  = "Invalid number of arguments:  %d.\n";

byrnHDF: No the problem is that a global was redefined!
    const char *err1 = "Path string larger than MAX_PATH_NAME_LENGTH.\n";

Mayber rename the new err1 to err_max to not conflict.

byrnHDF (approved):

lrknox (approved):

qkoziol (approved):
Switching to HDfputs() would also have worked.

seanm:
Since I personally use 1.10, I'd appreciate if as many as possible of my patches could get backported there too...

lrknox:
Our plan is to create issues to track porting to 1.12 and 1.10 (possibly 1.8 also) when patches are merged to develop, and then create PRs for those branches after 2 or 3 days testing of develop.  If creation of PR to branch is straightforward it will proceed with just one approval.  If not, we will ask for help in porting to those branches.

PR #341, opened 2021-02-15
Fixed all -Wincompatible-pointer-types-discards-qualifiers warnings

seanm:


examples/h5_ref_compat.c
```diff
@@ -76,9 +76,9 @@ main(void)
 
     /* Access reference and read dataset data through new API */
     assert(H5Rget_type((const H5R_ref_t *)&new_ref_buf[0]) == H5R_OBJECT2);
-    H5Rget_obj_type3((const H5R_ref_t *)&new_ref_buf[0], H5P_DEFAULT, &obj_type);
+    H5Rget_obj_type3((H5R_ref_t *)&new_ref_buf[0], H5P_DEFAULT, &obj_type);
```
gnuoyd: Is the cast necessary? Looks like the type of &new_ref_buf[0] should be H5R_ref_t *.

examples/h5_ref_compat.c
```diff
@@ -76,9 +76,9 @@ main(void)
...
     assert(obj_type == H5O_TYPE_DATASET);
-    dset1 = H5Ropen_object((const H5R_ref_t *)&new_ref_buf[0], H5P_DEFAULT, H5P_DEFAULT);
+    dset1 = H5Ropen_object((H5R_ref_t *)&new_ref_buf[0], H5P_DEFAULT, H5P_DEFAULT);
```
gnuoyd: Is the cast necessary?

examples/h5_ref_compat.c
seanm: Indeed not; will fix.

examples/h5_ref_compat.c
seanm: Indeed not; will fix.

tools/lib/h5tools_utils.c
```diff
@@ -196,7 +196,7 @@ get_option(int argc, const char **argv, const char *opts, const struct long_opti
         /* long command line option */
         int        i;
         const char ch      = '=';
-        char *     arg     = &argv[opt_ind][2];
+        char *     arg     = (char *)&argv[opt_ind][2]; /* TODO: should not cast away const! */
```
gnuoyd: Looks like you can strdup at the top of the block, `char * arg = strdup(&argv[opt_ind][2]);`, `free(arg)` at the bottom, and then replace occurrences of `argv[opt_ind][2]` with `arg`.

seanm: Good idea, done.

seanm:
@gnuoyd this look good to you now?

src/H5Dint.c
```diff
@@ -2256,7 +2256,8 @@ H5D_nameof(const H5D_t *dataset)
     /* Use FUNC_ENTER_NOAPI_NOINIT_NOERR here to avoid performance issues */
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
-    FUNC_LEAVE_NOAPI(dataset ? &(dataset->path) : NULL)
+    /* TODO: should not cast away const! */
+    FUNC_LEAVE_NOAPI(dataset ? (H5G_name_t *)&(dataset->path) : NULL)
```
qkoziol: Please don't cast away const, it's hiding the underlying design issue, which should be fixed.

src/H5Gint.c
```diff
@@ -747,7 +747,8 @@ H5G_nameof(const H5G_t *grp)
     /* Use FUNC_ENTER_NOAPI_NOINIT_NOERR here to avoid performance issues */
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
-    FUNC_LEAVE_NOAPI(grp ? &(grp->path) : NULL)
+    /* TODO: should not cast away const! */
+    FUNC_LEAVE_NOAPI(grp ? (H5G_name_t *)&(grp->path) : NULL)
```
qkoziol: Please don't cast away const, it's hiding the underlying design issue, which should be fixed.

src/H5HFcache.c
```diff
@@ -742,7 +742,7 @@ H5HF__cache_hdr_serialize(const H5F_t *f, void *_image, size_t H5_ATTR_NDEBUG_UN
     HDassert(len == hdr->heap_size);
 
     /* Set the shared heap header's file context for this operation */
-    hdr->f = f;
+    hdr->f = (H5F_t *)f; /* TODO: should not cast away const! */
```
qkoziol: Please don't cast away const, it's hiding the underlying design issue, which should be fixed.

src/H5HFcache.c
```diff
@@ -1328,7 +1328,7 @@ H5HF__cache_iblock_serialize(const H5F_t *f, void *_image, size_t H5_ATTR_NDEBUG
     hdr = iblock->hdr;
 
     /* Set the shared heap header's file context for this operation */
-    hdr->f = f;
+    hdr->f = (H5F_t *)f; /* TODO: should not cast away const! */
```
qkoziol: Please don't cast away const, it's hiding the underlying design issue, which should be fixed.

src/H5Shyper.c
```diff
@@ -10297,7 +10297,7 @@ H5Sselect_hyperslab(hid_t space_id, H5S_seloper_t op, const hsize_t start[], con
  REVISION LOG
 --------------------------------------------------------------------------*/
 herr_t
-H5S_combine_hyperslab(H5S_t *old_space, H5S_seloper_t op, const hsize_t start[], const hsize_t *stride,
+H5S_combine_hyperslab(const H5S_t *old_space, H5S_seloper_t op, const hsize_t start[], const hsize_t *stride,
```
qkoziol, line 10300: It looks like the ref count reached by old_space is modified, so I don't think that this should be changed to be const.

src/H5T.c
```diff
@@ -5606,7 +5606,7 @@ H5T_nameof(const H5T_t *dt)
             HGOTO_ERROR(H5E_DATATYPE, H5E_CANTINIT, NULL, "not a named datatype")
         case H5T_STATE_NAMED:
         case H5T_STATE_OPEN:
-            ret_value = &(dt->path);
+            ret_value = (H5G_name_t *)&(dt->path); /* TODO: should not cast away const! */
```
qkoziol: Please don't cast away const, it's hiding the underlying design issue, which should be fixed.

tools/lib/h5tools_utils.c
```diff
@@ -196,7 +196,7 @@ get_option(int argc, const char **argv, const char *opts, const struct long_opti
         /* long command line option */
         int        i;
         const char ch      = '=';
-        char *     arg     = &argv[opt_ind][2];
+        char *     arg     = strdup(&argv[opt_ind][2]);
```
qkoziol: Please use the HDstrdup wrapper here.

tools/lib/h5tools_utils.c
```diff
@@ -253,6 +253,8 @@ get_option(int argc, const char **argv, const char *opts, const struct long_opti
 
         opt_ind++;
         sp = 1;
+
+        free(arg);
```
qkoziol: And the HDfree wrapper here.

qkoziol (changes requested):
Generally in a good direction, but would prefer that the underlying design problems get fixed instead of hidden by casting away const-ness.

src/H5Shyper.c
seanm: Where is the modification?  Wouldn't that have caused a compiler error after I made it `const`?

tools/lib/h5tools_utils.c
seanm: Fixed.

tools/lib/h5tools_utils.c
seanm: Fixed.

src/H5Dint.c
seanm: Certainly fixing the underlying design issue would be preferable.

The cons of this change:
- a useful compiler warning has been downgraded to a TODO comment

The pros of this change:
- there's no real change. an implicit cast is now explicit. codegen will be identical.
- the codebase is now clean of `-Wincompatible-pointer-types-discards-qualifiers` warnings and the flag could now be enabled in your CI, thus preventing any ***new*** such warnings being introduced.

For me, this latter point outweighs the cons.

qkoziol: Yes, I understand the pros & cons. :-)  But the cons are much more important, I believe, since the underlying problem has not been resolved and probably never will be, since it's completely invisible to developers.  I understand the benefit to the overall codebase by clearing the codebase of the type of warning, but this sort of fix just encourages other developers to "fix" the warning by just casting away const also, instead of actually fixing the underlying design problem.  It generates a false sense of "security" since there's no warning, but the underlying problem still remains active.

src/H5Shyper.c
qkoziol: Line 10397:
 old_space->select.sel_info.hslab->span_lst->count--;

I would have thought that the compiler would barf also, but possibly the multiple levels of indirection are at play.

src/H5Dint.c
seanm: OK, I'll drop this commit then.

qkoziol:
Aside from the possible issue with H5S_select_hyperslab, I think this is good.

src/H5Shyper.c
seanm: `span_lst` isn't `const`, which I believe explains it. Tagging `old_space` as const doesn't promise infinite recursion of constness. And code is already passing `const H5S_t`s to this function, hence the warnings. So I still think this change is right.

qkoziol: Yeah, realized that later in the evening. :-)

qkoziol (approved):

lrknox (approved):

byrnHDF (approved):

PR #342, opened 2021-02-15
Fix for HDFFV-11109 - Copy MPI comm and info object into output FAPL from H5F_get_access_plist

jhendersonHDF:


testpar/t_prop.c
```diff
@@ -441,3 +441,192 @@ test_plist_ed(void)
     ret = H5Pclose(acpl);
     VRFY((ret >= 0), "H5Pclose succeeded");
 }
+
+void
+external_links(void)
```
jhendersonHDF, line 446: Reproducer for the issue from @brtnfld. Aborts with an "invalid communicator" message when the MPI comm on the FAPL used internally for H5Gopen of an external link is invalid (MPI_COMM_NULL in this case since it's the current default comm for FAPLs).

byrnHDF (approved):

lrknox (approved):

brtnfld (approved):

qkoziol:
I'm late to this PR, but it looks good to me.  (Although the asymmetry in the comm & info retrieval is a little weird)

jhendersonHDF:
> I'm late to this PR, but it looks good to me. (Although the asymmetry in the comm & info retrieval is a little weird)

I agree. Looks like we never routed the retrieval of the MPI Info object through the VFD layer. The MPI Comm retrieval is routed through there however, so it's a bit strange.

PR #347, opened 2021-02-17
Made some publicly exported symbols static, prefixed others with H5_

seanm:
When examining with nm -g, almost all public symbols from the library are prefexed with H5. There were just a handful of exceptions, so I suspected they were accidently exported. Only a couple were, the others I added a prefix.

### DIFF
--- a/hl/src/H5LTanalyze.c
+++ b/hl/src/H5LTanalyze.c
@@ -898,7 +898,7 @@ char *H5LTyytext;
 #include <hdf5.h>
 #include "H5LTparse.h"
 
-int my_yyinput(char *, int);
+static int my_yyinput(char *, int);
 #undef YY_INPUT
 #define YY_INPUT(b, r, ms) (r=my_yyinput(b, ms))
 #define token(x)        (int)x
@@ -920,8 +920,8 @@ struct cmpd_info {
     hbool_t         is_field;
     hbool_t         first_memb;
 };
-extern struct cmpd_info cmpd_stack[STACK_SIZE];
-extern int csindex;
+extern struct cmpd_info H5_cmpd_stack[STACK_SIZE];
+extern int H5_csindex;
 
 /*variables for array type*/
 struct arr_info {
@@ -929,21 +929,21 @@ struct arr_info {
     int                 ndim;
     hbool_t             is_dim;
 };
-extern struct arr_info arr_stack[STACK_SIZE];
-extern int asindex;
+extern struct arr_info H5_arr_stack[STACK_SIZE];
+extern int H5_asindex;
 
 /*variables for enumerate type*/
-extern hbool_t     is_enum;
-extern hbool_t     is_enum_memb;
+extern hbool_t     H5_is_enum;
+extern hbool_t     H5_is_enum_memb;
 
 /*variables for string type*/
-extern hbool_t is_str_size;
+extern hbool_t H5_is_str_size;
 
 /*variables for opaque type*/
-extern hbool_t is_opq_size;
-extern hbool_t is_opq_tag;
+extern hbool_t H5_is_opq_size;
+extern hbool_t H5_is_opq_tag;
 
-hbool_t        first_quote = 1;
+static hbool_t first_quote = 1;
 
 
 #line 920 "hl/src/H5LTanalyze.c"
@@ -1491,9 +1491,9 @@ case 55:
 YY_RULE_SETUP
 #line 140 "hl/src/H5LTanalyze.l"
 {
-                        if( is_str_size || (is_enum && is_enum_memb) ||
-                            is_opq_size || (asindex>-1 && arr_stack[asindex].is_dim) ||
-                            (csindex>-1 && cmpd_stack[csindex].is_field) ) {
+                        if( H5_is_str_size || (H5_is_enum && H5_is_enum_memb) ||
+                            H5_is_opq_size || (H5_asindex>-1 && H5_arr_stack[H5_asindex].is_dim) ||
+                            (H5_csindex>-1 && H5_cmpd_stack[H5_csindex].is_field) ) {
                             H5LTyylval.ival = atoi(H5LTyytext);
                             return NUMBER;
                         } else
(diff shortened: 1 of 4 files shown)


byrnHDF:
I believe these files are generated (remember bin/genparser) and the changes would need to happen in maybe the ".l" files.

seanm:
Did you maybe not scroll all the way down? I did indeed change the .l and .y files too.

byrnHDF:
No, I miss read the l as an h!
Also I accidentally closed it!
Lack of sleep.

byrnHDF:
We need to add some library devs to the review.

seanm:
The motivation here BTW is that VTK (www.vtk.org) & ITK (www.itk.org) mangle all HDF5 symbols.  As I've just recently redone this, I noticed 99% of symbols started with an HDF or H5 prefix, which is really nice because it makes it clear no weird symbols are leaking out.

gnuoyd:
I have some concerns about this change.

The names of many internal symbols in HDF5 already are prefixed in this way (H5_), but it's a poor practice that I think has outlasted its justification.

What do VTK and ITK do to mangle HDF5 symbols? Can HDF5 do something similar to protect internal symbols?

Also: it's my understanding that usually there are compiler/linker options to modify the default visibility of library symbols. Are there concrete reasons HDF5 does not hide internal symbols using one of those options? For a taste of what is possible, see the numerous options and techniques mentioned at https://stackoverflow.com/questions/3276474/symbol-hiding-in-static-libraries-built-with-xcode . HDF5 already uses some visibility settings on Windows—see H5_DLL and H5_DLLVAR.

Zooming out: it looks like most of the variables shared between the parser and the lexer are involved in lexing decimal numbers ([0-9]+) and strings. At a glance, I don't think that it's necessary for the symbol returned by the lexer (NUMBER versus REJECT) to be conditional on which production the symbol appears in, although I can see why a developer may have struck on this structure by accident.

seanm:
>What do VTK and ITK do to mangle HDF5 symbols?

They do this:
https://gitlab.kitware.com/third-party/hdf5/-/blob/for/vtk/src/vtk_hdf5_mangle.h

for these reasons:
https://gitlab.kitware.com/third-party/repo-requests/-/wikis/mangling

>Can HDF5 do something similar to protect internal symbols?

I think that's a different problem.  But yeah, maybe there are symbols exported that shouldn't be.

>Are there concrete reasons HDF5 does not hide internal symbols using one of those options?

Beats me.  I know nothing about HDF5.  :)  For me, it's just a dependency of a dependency. :)

>HDF5 already uses some visibility settings on Windows—see H5_DLL and H5_DLLVAR.

Not only Windows I don't think: https://github.com/HDFGroup/hdf5/blob/2ea165efd095d8876639814dc615dd01a2ec34e6/src/H5api_adpt.h#L38

qkoziol (approved):

seanm:
@gnuoyd BTW, at first, I tried making all these symbols static instead, and although it built, a test started failing. So I concluded they indeed needed to be `extern`, or that a larger refactoring was needed.

seanm:
Actually, I'm even more convinced this change is a good idea, for `is_enum` especially, because VTK's mangling trick of:

```
#define is_enum vtkhdf5_hl_is_enum
```

Causes issues for other C++ code that uses [`std::is_enum`](https://en.cppreference.com/w/cpp/types/is_enum).

seanm:
@gnuoyd @qkoziol this is now blocking me from further progress getting VTK's HDF5 updated.  I could always do whatever in VTK, but I prefer to do the same as upstream (i.e. you), so would be good to know if you approve this solution, or if there's a better idea... Thanks.

qkoziol:
@bmribler @byrnHDF @derobins - Any chance to take a look at this?

byrnHDF:
Another dev, familiar with the HL code, should check the changes, do we need a release.txt entry.

seanm:
Although the symbols are publicly exported, they're not API really, are they?  I wouldn't think this needs calling out in release notes. 2¢.

qkoziol:
> Although the symbols are publicly exported, they're not API really, are they? I wouldn't think this needs calling out in release notes. 2¢.

Agree

gnuoyd:
On Fri, Feb 26, 2021 at 08:11:33AM -0800, Sean McBride wrote:
> @gnuoyd @qkoziol this is now blocking me from further progress getting VTK's HDF5 updated.  I could always do whatever in VTK, but I prefer to do the same as upstream (i.e. you), so would be good to know if you approve this solution, or if there's a better idea... Thanks.

Here are a handful of responses:

* It looks like we can and should change the default visibility of
  global symbols on non-Windows platforms to "hidden" using, for
  example, -fvisibility=hidden with GCC.  That gives us much greater
  flexibility in the names we use for symbols internal to the library.
  We can have high confidence that changing the visibility will not
  break anything because that's the default on Windows.

* The visibility change should allow VTK's name munging to work,
  provided that we do not purposefully export any symbols that do not
  carry an HDF5-specific namespace prefix.

* We can avoid using library prefix across the board for *internal*
  globals, and we should.  It's undesirable for people to have to type
  and to see H5_ on so many symbols.

* The lexer & parser share globals because they have something wrong
  with them.  I will file a bug report.  When the visibility is changed,
  those globals should not affect VTK any longer.

Dave

qkoziol:
> On Fri, Feb 26, 2021 at 08:11:33AM -0800, Sean McBride wrote: @gnuoyd @qkoziol this is now blocking me from further progress getting VTK's HDF5 updated. I could always do whatever in VTK, but I prefer to do the same as upstream (i.e. you), so would be good to know if you approve this solution, or if there's a better idea... Thanks.
> Here are a handful of responses: * It looks like we can and should change the default visibility of global symbols on non-Windows platforms to "hidden" using, for example, -fvisibility=hidden with GCC. That gives us much greater flexibility in the names we use for symbols internal to the library. We can have high confidence that changing the visibility will not break anything because that's the default on Windows. * The visibility change should allow VTK's name munging to work, provided that we do not purposefully export any symbols that do not carry an HDF5-specific namespace prefix. * We can avoid using library prefix across the board for *internal* globals, and we should. It's undesirable for people to have to type and to see H5_ on so many symbols. * The lexer & parser share globals because they have something wrong with them. I will file a bug report. When the visibility is changed, those globals should not affect VTK any longer. Dave

These ideas should be separated into separate issues (seems like 2-3 issues) and handled with other PRs, so they don't hold this PR up.

seanm:
>It looks like we can and should change the default visibility of global symbols on non-Windows platforms to "hidden" using, for example, -fvisibility=hidden with GCC

I fully agree.  But that may be a substantial undertaking. Even if I had the time & inclination, would it be something backported to 1.10.x, or is it too big a change?  For now, I'm only trying to get VTK from 1.10.6 to 1.10.7 and this new exported symbol `is_enum` (not there in 1.10.6) is clashing with `std::is_enum`.  The renaming in this PR IMHO is a conservative change to solve that.

gnuoyd:
It was easy to fix the parser and lexer, so I did so in #399 .

seanm:
Nice! Thanks!

PR #350, opened 2021-02-17
added documentation for H5P APIs

brtnfld:
add documentation for H5Pget_buffer,H5Pget_data_transform,H5Pget_edc_c…heck,H5Pget_hyper_vector_size,H5Pget_preserve,H5Pget_type_conv_cb,H5Pget_vlen_mem_manager,H5Pset_btree_ratios

### DIFF
--- a/src/H5Ppublic.h
+++ b/src/H5Ppublic.h
@@ -5743,7 +5743,7 @@ H5_DLL herr_t H5Pget_btree_ratios(hid_t plist_id, double *left /*out*/, double *
  * \param[out] tconv     Address of the pointer to application-allocated type conversion buffer
  * \param[out] bkg       Address of the pointer to application-allocated background buffer
  *
- * \return Returns buffer size, in bytes, if successful; otherwise \e 0 on failure.
+ * \return Returns buffer size, in bytes, if successful; otherwise 0 on failure.
  *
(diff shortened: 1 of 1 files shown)


src/H5Ppublic.h
```diff
@@ -5733,52 +5733,194 @@ H5_DLL herr_t H5Pset_virtual_view(hid_t dapl_id, H5D_vds_view_t view);
...
+/**
+ *
+ * \ingroup  DXPL
+ *
+ * \brief Retrieves number of I/O vectors to be read/written in hyperslab I/O
+ *
+ * \param[in]   fapl_id Dataset transfer property list identifier
+ * \param[out]  size    Number of I/O vectors to accumulate in memory for I/O operations
+ *
+ * \return Returns a non-negative value if successful; otherwise returns a negative value.
```
gheber: Can you replace all `herr_t` returns with `\return \herr_t`? That way we prevent repetition and have to change it only in one place should we ever change our mind. I mean instead of
```
* \return Returns a non-negative value if successful; otherwise returns a negative value.
```
just use
```
* \return \herr_t
```

gheber (approved):

src/H5Ppublic.h
brtnfld: done

PR #356, opened 2021-02-18
Fixed HDFFV-11150

bmribler:
Description
    Replaced an HDassert with a check for null pointer in H5O_dec_rc() to
    catch null pointer in corrupted data situation.
    As a result, removed the null check prior to H5O_dec_rc() calls.
Platforms tested:
    Linux/64 (jelly)

### DIFF
--- a/src/H5Ochunk.c
+++ b/src/H5Ochunk.c
@@ -412,12 +412,12 @@ H5O__chunk_dest(H5O_chunk_proxy_t *chk_proxy)
     HDassert(chk_proxy);
 
     /* Decrement reference count of object header */
-    if (chk_proxy->oh && H5O__dec_rc(chk_proxy->oh) < 0)
+    if (H5O__dec_rc(chk_proxy->oh) < 0)
         HGOTO_ERROR(H5E_OHDR, H5E_CANTDEC, FAIL, "can't decrement reference count on object header")
 
+done:
     /* Release the chunk proxy object */
     chk_proxy = H5FL_FREE(H5O_chunk_proxy_t, chk_proxy);
 
-done:
     FUNC_LEAVE_NOAPI(ret_value)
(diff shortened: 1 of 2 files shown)


bmribler:
Regarding test, the user (Mathworks) hasn't provided the corrupted file yet, and I wasn't able to generate one that would cause the stated segfault.

qkoziol:
I still don't understand how the assert could get triggered with the original code.   If chk_proxy->oh is NULL, H5O__dec_rc should never be called and therefore the assert won't be executed / checked.   If chk_proxy->oh is non-NULL, H5O__dec_rc will get called, oh will be non-NULL and the assert shouldn't be triggered.

I agree with your assessment that it is impossible to generate a file that will cause the segfault. ;-)

qkoziol (changes requested):
Doesn't look like the code as originally written could trigger the type of bug described, so there must be some other cause for the bug.

bmribler:
Actually, the segfault was before the check for chk_proxy->oh being NULL was added.  The assert caused a crash at another place.  So instead of adding another check to the other place and, I think, yet another place, I thought it would be better to change H5O__dec_rc() instead.  I discussed that with you via email in December.

qkoziol:
> Actually, the segfault was before the check for chk_proxy->oh being NULL was added. The assert caused a crash at another place. So instead of adding another check to the other place and, I think, yet another place, I thought it would be better to change H5O__dec_rc() instead. I discussed that with you via email in December.

Ah, OK, that explains it! :-)  I think it's harmless to commit then, but would really be better if there was a reproducer for triggering it.

qkoziol (approved):

bmribler:
I was messing around with a file for that but got distracted.  If I ever get it, I'll add the test.  Thank you!

lrknox (approved):

PR #357, opened 2021-02-18
emscripten

thewtex:
- Remove invalid Emscripten HDF5 link flags
- Declare H5O__fsinfo_set_version
- Add output redirection for H5Tinit.c, H5lib_settings.c generation

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1103,7 +1103,7 @@ if (NOT EXISTS "${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c")
         OUTPUT ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
                ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
         COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:H5detect>
-        ARGS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
+        ARGS > ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
         COMMAND    ${CMAKE_COMMAND}
         ARGS       -E touch ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
         DEPENDS H5detect
@@ -1173,7 +1173,7 @@ add_custom_command (
     OUTPUT ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c
            ${HDF5_SRC_BINARY_DIR}/gen_SRCS.stamp2
     COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:H5make_libsettings>
-    ARGS ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c
+    ARGS > ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c
     COMMAND    ${CMAKE_COMMAND}
     ARGS       -E touch ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp2
     DEPENDS H5make_libsettings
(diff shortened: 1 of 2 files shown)


src/H5Fsuper.c
```diff
@@ -49,6 +49,7 @@
 /********************/
 static herr_t H5F__super_ext_create(H5F_t *f, H5O_loc_t *ext_ptr);
 static herr_t H5F__update_super_ext_driver_msg(H5F_t *f);
+static herr_t H5O__fsinfo_set_version(H5F_t *f, H5O_fsinfo_t *fsinfo);
```
qkoziol: This looks like a bit of leftover code.  (Definitely doesn't belong here :-) )

qkoziol (changes requested):
I'm not certain about the CMake changes, but the line in src/H5Fsuper.c won't work.

src/H5Fsuper.c
thewtex: It looks like this has moved to H5Oprivate.h -- I will include the header instead.

qkoziol: Not _strictly_ necessary, since H5Oprivate.h was being included by H5SMprivate.h, but since this module makes H5O calls directly, it's probably for the best.

qkoziol (approved):

src/CMakeLists.txt
```diff
@@ -1104,7 +1103,7 @@ if (NOT EXISTS "${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c")
         OUTPUT ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
                ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
         COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:H5detect>
-        ARGS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
+        ARGS > ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
```
byrnHDF, line 1106: I didn't think CMake allowed this ">" char here?

byrnHDF:
Researching the custom command seems to indicate that the "ARGS" option is not used and the content in "ARGS" should be in the "COMMAND". Will need further investigation.

src/CMakeLists.txt
byrnHDF: Searching cmake refs on add_custom_command, the preferred usage seems to indicate the argument should go on the command line. This might be a change since this was first implemented.
I think we should consider following convention.

src/CMakeLists.txt
```diff
@@ -1175,7 +1173,7 @@ add_custom_command (
     OUTPUT ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c
            ${HDF5_SRC_BINARY_DIR}/gen_SRCS.stamp2
     COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:H5make_libsettings>
-    ARGS ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c
+    ARGS > ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c
```
byrnHDF, line 1176: This is incorrect because H5make_libsettings uses the argument as a output file and not as stdout.
```
main(int argc, char *argv[])
{
    char *fname = NULL;
    FILE *f; /* temporary holding place for the stream pointer
              * so that rawoutstream is changed only when succeeded
              */

    if (argc > 1)
        fname = argv[1];

    /* First check if filename is string "NULL" */
    if (fname != NULL) {
        /* binary output */
        if ((f = HDfopen(fname, "w")) != NULL)
            rawoutstream = f;
    }
    if (!rawoutstream)
        rawoutstream = stdout;

```

src/CMakeLists.txt
```diff
@@ -1161,7 +1160,6 @@ target_compile_definitions(H5make_libsettings PUBLIC ${HDF_EXTRA_C_FLAGS} ${HDF_
 TARGET_C_PROPERTIES (H5make_libsettings STATIC)
 target_link_libraries (H5make_libsettings
     PRIVATE "$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_LIBRARIES}>" $<$<OR:$<PLATFORM_ID:Windows>,$<PLATFORM_ID:MinGW>>:ws2_32.lib>
-    PRIVATE $<$<PLATFORM_ID:Emscripten>:"-O0">
```
byrnHDF, line 1164: This S.B. a target_compile_option. Correctly changed in HDFFV-11083

byrnHDF (commented):
See CMake changes to byproducts - ninja warnings #568 for compile define

byrnHDF:
The necessary changes have been incorporated into develop with PR #568

src/CMakeLists.txt
attilaolah: (I realise this is already fixed upstream but) the problem here I think is that with `CMAKE_CROSSCOMPILING_EMULATOR` set to Node, `argv[0]` is actually the NodeJS binary itself, so `argv[1]` is the binary being executed, and `argv[2]` would be the output filename.

PR #358, opened 2021-02-19
Patches from vtk

seanm:


src/H5Cimage.c
```diff
@@ -1158,7 +1158,8 @@ H5C_load_cache_image_on_next_protect(H5F_t *f, haddr_t addr, hsize_t len, hbool_
     HDassert(cache_ptr->magic == H5C__H5C_T_MAGIC);
 
     /* Set information needed to load cache image */
-    cache_ptr->image_addr = addr, cache_ptr->image_len = len;
+    cache_ptr->image_addr = addr;
+    cache_ptr->image_len = len;
```
qkoziol: Good catch!

qkoziol (commented):
Not certain about the C++ changes, but the changes to the hl, src, test, and tools subdirs look good to me.

src/H5Cimage.c
seanm: Yes, clang has many great warning flags. In case you're not aware, there is a flag -Weverything that enables them all, literally (unlike -Wall), which of course is ridiculous, but then you can start subtracting.  That's what I did here.  Turned off the useless or noisy ones, then fixed others.  Once we get through my patches, I'll share that list.

seanm:
@lrknox could you do your magic to auto-fix the formatting of 8193858?

qkoziol (approved):

lrknox:
Formatting passes clang-format check.  Assigning to @bmribler for review of c++ code changes

qkoziol (approved):

bmribler (approved):

PR #359, opened 2021-02-19
Fix crash in H5TS_win32_thread_exit

chrisvarns:
Fix crash caused on thread exits when the following events occur:

1. HDF5 library is loaded
2. A thread is created
3. No HDF5 API's are called (so H5TS_win32_process_enter is never called, and H5TS_errstk_key_g and H5TS_funcstk_key_g are uninitialized and happen to be zero)
4. Some other library erroneously calls TlsSetValue with an index of zero (the same as our uninitialized H5TS_errstk_key_g and H5TS_funcstk_key_g keys)
5. The thread exits, calling H5TS_win32_thread_exit, which calls TlsGetValue with the uninitialized (zero) keys, gets a non-zero value back because of the other library previously calling TlsSetValue with that key, calls LocalFree on it and explodes with a heap error.

The fix is simply to initialize the keys to some known invalid value (TLS_OUT_OF_INDEXES) and only cleanup if they were ever actually allocated with a call to TlsAlloc.

### DIFF
--- a/src/H5TS.c
+++ b/src/H5TS.c
@@ -31,12 +31,15 @@ typedef struct H5TS_cancel_struct {
 /* Global variable definitions */
 #ifdef H5_HAVE_WIN_THREADS
 H5TS_once_t H5TS_first_init_g;
+H5TS_key_t H5TS_errstk_key_g = TLS_OUT_OF_INDEXES;
+H5TS_key_t H5TS_funcstk_key_g = TLS_OUT_OF_INDEXES;
+H5TS_key_t H5TS_cancel_key_g = TLS_OUT_OF_INDEXES;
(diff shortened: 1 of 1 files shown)


qkoziol (approved):
Nice find!

chrisvarns:
Thanks I did not notice the conflict until after I opened the PR
unfortunately, i originally fixed this late last year. I'll take a look at
resolving it tomorrow!

EDIT: Worst case hopefully this serves as a reasonable bug report.

derobins (approved):

byrnHDF:
I will create a new PR for this soon, maybe next week, based on the latest develop and get this merged down to the other branches.
I will of course reference this PR.

byrnHDF:
Replaced by #396

PR #380, opened 2021-02-24
Use do-while trick to force H5Epush_ret() to require trailing semi

seanm:
Fixed ensuing compiler errors.

### DIFF
--- a/src/H5Epublic.h
+++ b/src/H5Epublic.h
@@ -127,7 +127,7 @@ H5_DLLVAR hid_t H5E_ERR_CLS_g;
     do {                                                                                                     \
         H5Epush2(H5E_DEFAULT, __FILE__, func, __LINE__, cls, maj, min, str);                                 \
         return (ret);                                                                                        \
-    } while(0)
+    } while (0)
 
 /* Use the Standard C __FILE__ & __LINE__ macros instead of typing them in
  * And goto a label after pushing error onto stack.
--- a/src/H5Epublic.h
+++ b/src/H5Epublic.h
@@ -124,10 +124,10 @@ H5_DLLVAR hid_t H5E_ERR_CLS_g;
 /* Use the Standard C __FILE__ & __LINE__ macros instead of typing them in */
 /*  And return after pushing error onto stack */
 #define H5Epush_ret(func, cls, maj, min, str, ret)                                                           \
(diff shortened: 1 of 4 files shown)


seanm:
@lrknox this will need your magic auto-reformat please

seanm:
BTW I found this using clang-tidy's `readability-misleading-indentation` check.

byrnHDF:
The format issues are because H5Epush_ret is now correctly identified as a statement.

seanm:
>The format issues are because H5Epush_ret is now correctly identified as a statement.

Not sure I understand... the wacky formatting is already there, i did not cause it. Likely clang-format will reformat correctly now that I added the semi-colons.

byrnHDF:
> > The format issues are because H5Epush_ret is now correctly identified as a statement.
> 
> Not sure I understand... the wacky formatting is already there, i did not cause it. Likely clang-format will reformat correctly now that I added the semi-colons.

sorry, that is what I meant, the change will require a source format because the format will now recognize it correctly.

seanm:
@byrnHDF ah ok yes indeed. @lrknox has been kind enough to do the formatting for me previously. I have clang 12, which I fear will format slightly differently.

lrknox:
I can run formatter, if that's all it needs.   I'll also try to run it on others whenever needed.

qkoziol (approved):

lrknox (approved):

src/H5Epublic.h
```diff
@@ -124,10 +124,10 @@ H5_DLLVAR hid_t H5E_ERR_CLS_g;
 /* Use the Standard C __FILE__ & __LINE__ macros instead of typing them in */
 /*  And return after pushing error onto stack */
 #define H5Epush_ret(func, cls, maj, min, str, ret)                                                           \
-    {                                                                                                        \
+    do {                                                                                                     \
         H5Epush2(H5E_DEFAULT, __FILE__, func, __LINE__, cls, maj, min, str);                                 \
         return (ret);                                                                                        \
-    }
+    } while (0)
```
derobins, line 131: Although a minor change, this should probably get a note in RELEASE.txt as it's in the public API.

seanm: Done.

PR #382, opened 2021-02-25
removed const from H5D_nameof

brtnfld:
addresses warning with xlc compiler

### DIFF
--- a/.autom4te.cfg
+++ b/.autom4te.cfg
@@ -5,7 +5,7 @@
 # This file is part of HDF5.  The full HDF5 copyright notice, including
 # terms governing use, modification, and redistribution, is contained in
 # the COPYING file, which can be found at the root of the source code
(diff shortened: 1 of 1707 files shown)


qkoziol (changes requested):
I'd like to clean up this warning too, it's been annoying for a long time.   I don't think this is quite the right fix, but may be close.  Can you leave the const on the H5G_t / H5D_t / H5T_t struct, but _add_ a const to the return value?   That's a much better direction to go - it says to the caller: "You shouldn't modify this".

seanm:
I had tried that, but it leads to a cascade of different const problems.

My impression is that only someone that knows the codebase well can fix the remaining const problems, the low hanging fruit seems to be gone.

qkoziol:
> I had tried that, but it leads to a cascade of different const problems.
> 
> My impression is that only someone that knows the codebase well can fix the remaining const problems, the low hanging fruit seems to be gone.

Very true - the ones that remain are definitely going to take some expertise & time.

PR #383, opened 2021-02-25
H5G/T_nameof const arg. removal

brtnfld:
removes warnings for xl compiler


qkoziol (changes requested):
I'd like to clean up this warning too, it's been annoying for a long time.   I don't think this is quite the right fix, but may be close.  Can you leave the const on the H5G_t / H5D_t / H5T_t struct, but _add_ a const to the return value?   That's a much better direction to go - it says to the caller: "You shouldn't modify this".

PR #388, opened 2021-02-26
Moves the BSD queue to a wrapper to keep it out of the public API

derobins:
Adds an internal wrapper struct for the dedup queue. This keeps the BSD queue bits out of the H5FD public header.

Also adds some fixes to allow the code to build on MacOS.

Issues:
    * The library always builds with -lpthread now, even though that is really only necessary for the tests


derobins:
The H5queue.h changes look more complicated than they really are. Aside from some comments and ifdefs, the big changes are:

* Yanked the circular queue macros.
* Moved the SIMPLEQ macros to the end of the file. MacOS doesn't have SIMPLEQ in sys/queue.h so we have to tack this on at the end. We need to use sys/queue.h on MacOS because sys/file.h uses the queue macros and we get symbol redefinitions if we try to use our own implementation.

jrmainzer (approved):
Looks reasonable to me.  

We will probably revisit the timer issue, but this should be fine for now.

src/H5queue.h
```diff
@@ -31,8 +31,8 @@
  *	@(#)queue.h	8.5 (Berkeley) 8/20/94
  */
 
-#ifndef	_SYS_QUEUE_H_
-#define	_SYS_QUEUE_H_
+#ifndef	_H5queue_H_
+#define	_H5queue_H_
```
seanm: This should be `H5queue_H`. Otherwise you'll add more `Wreserved-id-macro` warnings.

test/vfd_swmr_common.c
```diff
@@ -202,7 +254,37 @@ await_signal(hid_t fid)
 
     dbgf(1, "waiting for signal\n");
 
+#if defined(__APPLE__) && defined(__MACH__)
```
seanm: Why the `__MACH__`?

derobins: That's something I pulled from Stackoverflow, IIRC. Things like that are used to differentiate between MacOS and iOS. Eventually, I'll change this so its conditional on sigtimedwait (I think FreeBSD also lacks sigtimedwait, and this hack won't help on that platform), but right now I just want to get it to compile on Macs.

src/H5queue.h
derobins: Ah. Will fix.

raylu-hdf (approved):
This PR compiled on my Mac OS machine (10.13.6).  The test suite for VFD SWMR finished ok.

src/H5queue.h
derobins: I fixed this one as well as the other VFD SWMR include guards.

test/vfd_swmr_common.c
derobins: Now I test on the presence of sigtimedwait, so this is obsolete.

PR #393, opened 2021-02-26
Add the two VFD SWMR demo programs to the test directory so that they…

vchoi-hdfgroup:
… can be built as needed.


test/Makefile.am
```diff
@@ -120,10 +120,24 @@ endif
...
 BUILD_ALL_PROGS=gen_bad_ohdr gen_bogus gen_cross gen_deflate gen_filters gen_new_array \
     gen_new_fill gen_new_group gen_new_mtime gen_new_super gen_noencoder \
     gen_nullspace gen_udlinks space_overflow gen_filespace gen_specmetaread \
-    gen_sizes_lheap gen_file_image gen_plist gen_bad_offset gen_bounds
+    gen_sizes_lheap gen_file_image gen_plist gen_bad_offset gen_bounds \
+    vfd_swmr_gaussians vfd_swmr_credel
+
+# VFD SWMR demo: vfd_swmr_gaussians needs the curses library
+vfd_swmr_gaussians_LDFLAGS=$(AM_LDFLAGS) @CURSES_LIBS@
+vfd_swmr_gaussians=$(AM_CFLAGS) @CURSES_LIBS@
```
derobins: Shouldn't this be:

vfd_swmr_gaussians=$(AM_CFLAGS) @CURSES_CFLAGS@

?

You shouldn't need to add @CURSES_LIBS@ to both CFLAGS and LDFLAGS.

test/vfd_swmr_common.h
```diff
@@ -88,6 +94,9 @@ H5TEST_DLL void esnprintf(char *, size_t, const char *, ...)
 
 H5TEST_DLL int fetch_env_ulong(const char *, unsigned long, unsigned long *);
 
+/* For demo */
+H5TEST_DLL size_t strlcpy(char *, const char *, size_t);
```
derobins: OpenBSD puts strlcpy in string.h so this will cause problems on that platform. This should be renamed to h5_strlcpy or something like that.

https://man.openbsd.org/strlcat.3

test/vfd_swmr_credel.c
```diff
@@ -0,0 +1,483 @@
+/*
```
derobins: If this is going to be built as a part of the library, it needs to be brought up to THG code standards or it needs to go in the examples.

vchoi-hdfgroup:
Put the demo programs in examples instead.

PR #399, opened 2021-02-26
Simplified hl parsing

### LINKED ISSUE #398: H5LT lexer and parser share state with global variables when there is no need.
The lexer uses an unconventional strategy for parsing lexical categories NUMBER (decimal numbers) and STRING (double-quoted strings) that involves sharing the parse context with the lexer using global variables. There are a couple of problems with that. First, the lexer is too complicated for the simple tokenization it performs—it's hard to tell if it is correct. Second, as @seanm points out, the shared global variables spill into the namespace shared by other libraries and application programs—e.g., VTK.

gnuoyd:
Fixes #398 .


lrknox:
There were 5 variations of this error in the Windows Latest MSVC test: 
Building Custom Rule D:/a/hdf5/hdf5/hl/tools/gif2h5/CMakeLists.txt
  decompress.c
  gif2hdf.c
  gif2mem.c
  gifread.c
  writehdf.c
  Generating Code...
libhdf5_hl.lib(H5LTanalyze.obj) : error LNK2019: unresolved external symbol strndup referenced in function H5LTyylex [D:\a\hdf5\build\hl\tools\gif2h5\gif2h5.vcxproj]
D:\a\hdf5\build\bin\Release\gif2h5.exe : fatal error LNK1120: 1 unresolved externals [D:\a\hdf5\build\hl\tools\gif2h5\gif2h5.vcxproj]
  Building Custom Rule D:/a/hdf5/hdf5/hl/src/CMakeLists.txt

seanm:
lgtm. Once ready, can you please bring this to 1.10 branch too?

seanm:
@gnuoyd seems `strndup` wasn't in HDF5 before this patch. It's a POSIX function, so not in Windows I guess, and only officially added to ISO C in C2x.

I guess you can replace with malloc+strcpy.

byrnHDF:
> @gnuoyd seems `strndup` wasn't in HDF5 before this patch. It's a POSIX function, so not in Windows I guess, and only officially added to ISO C in C2x.
> 
> I guess you can replace with malloc+strcpy.

It is in H5private.h, about line 1695:
#define HDwrite(F, M, Z) write(F, M, Z)
#endif /* HDwrite */

/*
 * And now for a couple non-Posix functions...  Watch out for systems that
 * define these in terms of macros.
 */
#if !defined strdup && !defined H5_HAVE_STRDUP
extern char *                   strdup(const char *s);
#endif

#ifndef HDstrdup
#define HDstrdup(S) strdup(S)
#endif /* HDstrdup */

byrnHDF:
> @gnuoyd seems `strndup` wasn't in HDF5 before this patch. It's a POSIX function, so not in Windows I guess, and only officially added to ISO C in C2x.
> 
> I guess you can replace with malloc+strcpy.

However strdup is used in the library.

hl/src/H5LTanalyze.l
```diff
@@ -138,31 +103,12 @@ OPQ_SIZE         {return token(OPQ_SIZE_TOKEN);}
...
 
 [0-9]+           {
-                        if( is_str_size || (is_enum && is_enum_memb) ||
-                            is_opq_size || (asindex>-1 && arr_stack[asindex].is_dim) ||
-                            (csindex>-1 && cmpd_stack[csindex].is_field) ) {
-                            H5LTyylval.ival = atoi(yytext);
-                            return NUMBER;
-                        } else
-                            REJECT;
+                        H5LTyylval.ival = atoi(yytext);
```
qkoziol: Please use the HD* portability macros for all standard / system calls.  (atoi, strdup, etc)

qkoziol (changes requested):

seanm:
@gnuoyd any chance you might be able to update this soon?  It seems to be the last thing blocking me from updating VTK's HDF (specifically the `is_enum` vs `std::is_enum` conflict).

gnuoyd:
On Fri, Mar 05, 2021 at 08:15:24AM -0800, Sean McBride wrote:
> @gnuoyd any chance you might be able to update this soon?  It seems to be the last thing blocking me from updating VTK's HDF (specifically the `is_enum` vs `std::is_enum` conflict).

I can probably get us over the Windows hump, today.

Dave

derobins (changes requested):
Please follow THG coding conventions. In particular, using the HD prefix on standard library calls.

hl/src/H5LTanalyze.l
```diff
@@ -123,6 +124,25 @@ OPQ_TAG          {return token(OPQ_TAG_TOKEN);}
...
+ */
+static char *
+trim_quotes(const char *quoted)
+{
+    size_t len = strlen(quoted);
+    char *trimmed;
+
+    assert(quoted[0] == '"' && quoted[len - 1] == '"');
+
+    trimmed = strdup(quoted + 1);
```
seanm: Even MSVC 2015 supports enough C99 to do the declaration & initialization at once, as in: `char *trimmed = strdup(quoted + 1);`  Is there a reason you do this C89 style? (I've noticed most of HDF5 is the same.)

gnuoyd:
On Sat, Mar 06, 2021 at 11:04:09PM -0800, Sean McBride wrote:
> +/* Allocate a copy of `quoted` with the double quote character at
> + * the beginning and the one at the end both removed.  The caller is
> + * responsible for free()ing the copy.
> + */
> +static char *
> +trim_quotes(const char *quoted)
> +{
> +    size_t len = strlen(quoted);
> +    char *trimmed;
> +
> +    assert(quoted[0] == '"' && quoted[len - 1] == '"');
> +
> +    trimmed = strdup(quoted + 1);
> 
> Even MSVC 2015 supports enough C99 to do the declaration & initialization at once, as in: `char *trimmed = strdup(quoted + 1);`  Is there a reason you do this C89 style? (I've noticed most of HDF5 is the same.)

It's only in the last year that THG dropped support for older versions
of MSVC.  So it has not been long that declarations that appeared
elsewhere than the beginning of a block would compile on every supported
platform.

Looks like -Wdeclaration-after-statement is still enabled.  Probably
just inertia?

Sometimes I use a mid-block declaration, combined with a liberal
dose of `const`, to get the compiler's help to enforce properties
on my code---for example, the property that a variable is
assigned only once.

Dave

seanm:
@gnuoyd I see, thanks.  So I guess it's only in `develop` that mid-block declarations are allowed, since you presumably don't want old branches to stop working with old toolchains. And since I need this change of your in 1.10.x, I retract my suggestion/comment. :) But for `develop`, I made this quickie: https://github.com/HDFGroup/hdf5/pull/447

epourmal:
> @gnuoyd I see, thanks. So I guess it's only in `develop` that mid-block declarations are allowed, since you presumably don't want old branches to stop working with old toolchains. And since I need this change of your in 1.10.x, I retract my suggestion/comment. :) But for `develop`, I made this quickie: #447

Sorry, maybe I missing some context :-) 1_10 is not old and should comply with C99. I don't see why the change that is in develop would not be in 1_10 branch for upcoming 1.10.8 release.

qkoziol (approved):

seanm (approved):

derobins (approved):

seanm (approved):

seanm:
@gnuoyd so I'm just integrating this into VTK and it looks like it will fix the `is_enum` issue. Thanks!

There are still a *small* number of public symbols that don't have any H5 (or similar) prefix, namely:

In hl:

- indent 
- input_len 
- my_yyinput  -> looks like this could be `static`
- myinput 

in main library:

- Nflock 
- Pflock 
- use_locks_env_g 

Would be cool if they could be either prefixed or marked as non-exported. Thoughts?

PR #400, opened 2021-02-26
More warning fixes

seanm:


lrknox:
@seanm I've run formatter resulting in changes to 4 test files.  Shall I commit changes?

seanm:
Commit, sure. But might want to give reviewers more time for merge.

test/ros3.c
```diff
@@ -865,7 +865,7 @@ test_vfd_open(void)
         if (T.which_fapl == FAPL_FILE_ACCESS)
             _fapl_id = fapl_file_access;
         else if (T.which_fapl == FAPL_ROS3_ANON)
-            _fapl_id = fapl_id;
+            _(void) fapl_id;
```
qkoziol: ?

qkoziol (approved):
Aside from the one typo, this looks good.

test/ros3.c
seanm: whoops. good catch. A find-replace screwup I guess. Clearly I'm not compiling that file. Do you know what build option I'm missing?

qkoziol: > whoops. good catch. A find-replace screwup I guess. Clearly I'm not compiling that file. Do you know what build option I'm missing?

I believe that the "read-only s3" ("ros3") VFD needs to talk to an S3 server, but it's enabled with the --enable-ros3-vfd configure flag.

c++/test/tarray.cpp
```diff
@@ -182,7 +182,7 @@ test_array_compound_array()
         // Check the array dimensions
         for (ii = 0; ii < ndims; ii++)
             if (rdims1[ii] != tdims1[ii]) {
-                TestErrPrintf("Array dimension information doesn't match!, rdims1[%d]=%zd, tdims1[%d]=z%d\n",
+                TestErrPrintf("Array dimension information doesn't match!, rdims1[%d]=%lld, tdims1[%d]=%lld\n",
```
gnuoyd: I'm not sure about `%lld`.... There's definitely a disagreement in signedness. `"%" PRIuHSIZE` will always work to format an `hsize_t`.

c++/test/tarray.cpp
```diff
@@ -459,7 +459,7 @@ test_array_info()
             for (ii = 0; ii < ndims; ii++)
                 if (rdims1[ii] != tdims1[ii]) {
                     TestErrPrintf(
-                        "Array dimension information doesn't match!, rdims1[%d]=%zd, tdims1[%d]=z%d\n", ii,
+                        "Array dimension information doesn't match!, rdims1[%d]=%llu, tdims1[%d]=z%llu\n", ii,
```
gnuoyd, line 465: Looks like that z is a typo?

hl/src/H5IM.c
```diff
@@ -169,6 +169,11 @@ find_palette(H5_ATTR_UNUSED hid_t loc_id, const char *name, H5_ATTR_UNUSED const
     if (name == NULL)
         return -1;
 
+    /* Shut compiler up */
+    (void)loc_id;
+    (void)ainfo;
+    (void)op_data;
+
```
gnuoyd, line 176: The H5_ATTR_UNUSED didn't do the trick? Maybe it needs to go between the type and the parameter name? If so, funny if the use above doesn't produce an error.

c++/test/tarray.cpp
seanm: Indeed the signedness was and is wrong, will fix.

c++/test/tarray.cpp
seanm: Good catch. I fixed one, but apparently there's this one too.

lrknox:
I ran the clang formatter to make the format check pass.  There was a conflict for line 868 of test/ros3.c;  please check to see whether it is correct.

hl/src/H5IM.c
seanm: In fact it does. I removed the redundancy.

seanm:
@lrknox oh dear, I just pushed my fixes to what @gnuoyd pointed out, before seeing your comment.  I think it maybe overwrote your formatting changes... let's see if anyone has any other necessary changes, then could you reformat again?

lrknox:
Sure, no problem.

derobins (approved):

test/ros3.c
seanm: Fixed.

PR #425, opened 2021-03-04
Fixes various warnings noticed on Windows

derobins:
- Adds a prototype for our implementation of vasprintf
- Return type of H5_get_utf16_str() is now non-const
- Fixes possible uninitialized return type in Wremove_utf8
- Better isolation of fork() code in accum.c:test_swmr_write_big()
- Better isolation of non-zlib code in dsets.c:test_filter_delete()
- Removed unused variable in trefer.c:test_reference_cmpnd_obj()

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -675,11 +675,6 @@ typedef struct {
 #ifndef HDacos
 #define HDacos(X) acos(X)
 #endif /* HDacos */
-#ifndef HDvasprintf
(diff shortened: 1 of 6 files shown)


byrnHDF:
Maybe needs a release.txt entry?

derobins:
> Maybe needs a release.txt entry?

I don't think so. This is all internal stuff.

byrnHDF (approved):

qkoziol (approved):

PR #427, opened 2021-03-04
Fixed clang-tidy readability-misleading-indentation warnings

seanm:


seanm:
There are two other such warnings:

1. hdf5/src/H5MFaggr.c:369:9: warning: different indentation for 'if' and corresponding 'else' [readability-misleading-indentation]

That `REPLACE` macro seems to never be defined?

2. hdf5/hl/src/H5LTanalyze.c:1140:13: warning: misleading indentation: statement is indented too deeply [readability-misleading-indentation]

I can't figure out where in the .l file this comes from, or if it's a flex bug.

byrnHDF (approved):

seanm:
Please check src/H5Oalloc.c especially carefully, I just took a guess at what it was supposed to do!

qkoziol (approved):

PR #429, opened 2021-03-04
Applied clang-tidy readability-non-const-parameter warning fixes auto…

seanm:
…matically

### DIFF
--- a/hl/test/test_file_image.c
+++ b/hl/test/test_file_image.c
@@ -51,7 +51,7 @@
  *-------------------------------------------------------------------------
  */
 static int
-test_file_image(size_t open_images, size_t nflags, unsigned *flags)
+test_file_image(size_t open_images, size_t nflags, const unsigned *flags)
 {
     hid_t * file_id, *dset_id, file_space, plist; /* HDF5 ids */
     hsize_t dims1[RANK]    = {2, 3};              /* original dimension of datasets */
--- a/hl/test/test_ld.c
+++ b/hl/test/test_ld.c
@@ -157,7 +157,7 @@ static herr_t test_LD_elmts_invalid(const char *file);
 static herr_t test_LD_elmts_one(const char *file, const char *dname, const char *fields);
 static herr_t test_LD_elmts_two(const char *file, const char *dname, const char *fields);
 
-static herr_t verify_elmts_two(int type, hsize_t *ext_dims, hsize_t *prev_dims, void *_ldbuf, void *_buf);
+static herr_t verify_elmts_two(int type, const hsize_t *ext_dims, const hsize_t *prev_dims, void *_ldbuf, void *_buf);
 
 /* data structures for compound data type */
 typedef struct sub22_t {
@@ -1135,7 +1135,7 @@ error:
  **************************************************************************************
  */
 static herr_t
-verify_elmts_two(int type, hsize_t *ext_dims, hsize_t *prev_dims, void *_ldbuf, void *_buf)
+verify_elmts_two(int type, const hsize_t *ext_dims, const hsize_t *prev_dims, void *_ldbuf, void *_buf)
 {
     int k, m; /* Local index variable */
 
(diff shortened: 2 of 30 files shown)


seanm:
@qkoziol perhaps this is a step forward in the const battle.  This patch was computer generated by running:

`run-clang-tidy.py -checks=-*,readability-non-const-parameter -fix`

The tool seems smart enough to change only what results in something that compiles. :)

byrnHDF (approved):
Assuming source formatting fixed.

gnuoyd:
This looks like a step in the right direction.

test/direct_chunk.c
```diff
@@ -76,9 +76,9 @@
 
 /* Local prototypes for filter functions */
 static size_t filter_bogus1(unsigned int flags, size_t cd_nelmts, const unsigned int *cd_values,
-                            size_t nbytes, size_t *buf_size, void **buf);
+                            size_t nbytes, const size_t *buf_size, void **buf);
 static size_t filter_bogus2(unsigned int flags, size_t cd_nelmts, const unsigned int *cd_values,
-                            size_t nbytes, size_t *buf_size, void **buf);
+                            size_t nbytes, const size_t *buf_size, void **buf);
```
qkoziol: These 2 may not work, since they are supposed to match the signature for H5Z_func_t

test/dsets.c
```diff
@@ -307,7 +307,7 @@ static size_t filter_bogus(unsigned int flags, size_t cd_nelmts, const unsigned
 static htri_t can_apply_bogus(hid_t dcpl_id, hid_t type_id, hid_t space_id);
 static herr_t set_local_bogus2(hid_t dcpl_id, hid_t type_id, hid_t space_id);
 static size_t filter_bogus2(unsigned int flags, size_t cd_nelmts, const unsigned int *cd_values,
-                            size_t nbytes, size_t *buf_size, void **buf);
+                            size_t nbytes, const size_t *buf_size, void **buf);
```
qkoziol: Same here (H5Z_func_t)

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -150,7 +150,7 @@ const H5Z_class2_t H5Z_MYFILTER[1] = {{
 #define MULTIPLIER          3
 
 static size_t H5Z_filter_dynlibud(unsigned int flags, size_t cd_nelmts, const unsigned int *cd_values,
-                                  size_t nbytes, size_t *buf_size, void **buf);
+                                  size_t nbytes, const size_t *buf_size, void **buf);
```
qkoziol: Same here (H5Z_func_t)

qkoziol (approved):
Aside from the H5Z_func_t issues, this looks good.

seanm:
@qkoziol good catch. see 3rd commit for followup to automated changes.

qkoziol:
> @qkoziol good catch. see 3rd commit for followup to automated changes.

The compiler is OK with this?   The H5Z_class2_t's 'filter' function pointer will be off slightly, since the buf_size pointer is const for these functions, but the H5Z_func_t typedef is non-const.

seanm:
I changed the `H5Z_func_t` typedef to const too.

It compiles for me.

qkoziol:
> I changed the `H5Z_func_t` typedef to const too.
> 
> It compiles for me.

Ah, missed that.  Unfortunately, that typedef can't change, since it's part of the public API for I/O filters.

seanm:
I'm not looking to change this in 1.10 or other stable branches.  Can there never ever be public API changes ever?  :(  This is about as minor an API change as you can get too.

qkoziol:
> I'm not looking to change this in 1.10 or other stable branches. Can there never ever be public API changes ever? :( This is about as minor an API change as you can get too.

Sure the public API can be changed.  :-)   In this case though, we shouldn't change it, since the buf_size pointer is an in/out value.   The filter will update it with the new size of the buffer, after it's been compressed (or otherwise filtered).

seanm:
Oh!  Odd that it compiles then.

seanm:
@qkoziol ok, I think I fixed all the `H5Z_func_t` cases.

qkoziol (approved):

derobins (approved):

PR #434, opened 2021-03-05
Removed bad function pointer casts

seanm:
In one case fixed the actual function signature to be correct.


seanm:
@qkoziol how about let's get this in first, maybe clang-tidy will be smarter about where it tries to add consts.

qkoziol (approved):
Sure, seems like a step toward that goal.

derobins (approved):

PR #437, opened 2021-03-05
Fix example of hyperslab block coordinate printing 

grimbough:
No sure if you welcome community pull requests, but hopefully this is helpful.

I've notice a small bug in the `h5_vds.c` example, where it prints the coordinates of the hyperslab blocks.  This is the current code:

```c
for (l=0; l<nblocks; l++) {
    printf("(");
    for (k=0; k<RANK2-1; k++) 
        printf("%d,", (int)buf[k]);
    printf("%d) - (", (int)buf[k]);
    for (k=0; k<RANK2-1; k++) 
        printf("%d,", (int)buf[RANK2+k]);
     printf("%d)\n", (int)buf[RANK2+k]);
}
```

Although it loops over the number of blocks, it doesn't use the block counter when printing the information, and so will just show the coordinates of the first block over and over.  The patch is intended to enable printing the coordinates for all blocks.

This behaviour isn't obvious in the current example as none of the hyperslabs consist of more than one block. However, I ended up at the example from [H5S_GET_SELECT_HYPER_BLOCKLIST](https://portal.hdfgroup.org/display/HDF5/H5S_GET_SELECT_HYPER_BLOCKLIST) and was trying to use it to understand my own program, where this printing had me confused for a while.  Hopefully this might help anyone else copy/pasting the example into their own code in the future.

### DIFF
--- a/examples/h5_vds.c
+++ b/examples/h5_vds.c
@@ -60,7 +60,7 @@ main(void)
     hsize_t start_out[2], stride_out[2], count_out[2], block_out[2];
     int     wdata[DIM0],         /* Write buffer for source dataset */
         rdata[VDSDIM0][VDSDIM1], /* Read buffer for virtual dataset */
-        i, j, k, l;
+        i, j, k, l, block_inc;
     int          fill_value = -1; /* Fill value for VDS */
     H5D_layout_t layout;          /* Storage layout */
     size_t       num_map;         /* Number of mappings */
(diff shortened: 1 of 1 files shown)


epourmal:
Thank you for your contribution! Yes, we do encourage PRs from community. The patch looks good to me!

qkoziol (approved):

lrknox (approved):

PR #448, opened 2021-03-08
Various clang tidy warning fixes

seanm:


test/error_test.c
```diff
@@ -129,13 +129,8 @@ test_error(hid_t file)
         TEST_ERROR;
     if (old_data != NULL)
         TEST_ERROR;
-#ifdef H5_USE_16_API
```
qkoziol, line 132: Why is this out?

qkoziol (approved):
The floating-point issues may generate more warnings on other compilers (*sigh*), and @bmribler will have to look over the C++ changes, but otherwise looks good to me.

seanm:
Thanks for the speedy review!

> The floating-point issues may generate more warnings on other compilers (_sigh_)

Really? I'd be surprised. Which do you think? We could try on godbolt.

test/error_test.c
seanm: I tried to explain in the commit message. It's dead code. See line 23.

qkoziol:
> Thanks for the speedy review!
> 
> > The floating-point issues may generate more warnings on other compilers (_sigh_)
> 
> Really? I'd be surprised. Which do you think? We could try on godbolt.

I believe that the current MPI compiler wrapper (mpicc) installed with homebrew on MacOS is wrapping a version of GCC that still barks at these issues. :-/

test/error_test.c
qkoziol: Ah, yes, definitely. :-)

qkoziol: > I tried to explain in the commit message. It's dead code. See line 23.

BTW - is there a way I view the commit messages from the PR?

seanm: Yes, click "Commits" at the top of this page, then click the commit itself, should go to a link like https://github.com/HDFGroup/hdf5/pull/448/commits/0d8c542e9133004b8b868609470e2b571b208e04

seanm: I'm less certain of the `DOXYGEN_SHOULD_SKIP_THIS` change there BTW, was a bit hard to untangle, but the warning was *not* a false positive.

seanm:
> I believe that the current MPI compiler wrapper (mpicc) installed with homebrew on MacOS is wrapping a version of GCC that still barks at these issues. :-/

On passing a double to a function that wants a double?!

tools/test/h5diff/h5diffgentest.c
```diff
@@ -465,15 +465,15 @@ test_basic(const char *fname1, const char *fname2, const char *fname3)
         float data15[6];
         float data16[6];
 
-        data15[0] = (float)HDsqrt(-1.0F);
```
byrnHDF, line 468: Not arguing against, just that I thought someone had added those "F" to solve a problem for a certain compiler/platform.
Unfortunately, I don't remember when or by whom.
We will need to monitor the regression testing after this goes in.

byrnHDF (approved):

PR #449, opened 2021-03-08
Removed checks/workarounds for pre-C++89 compatibility

seanm:
After 30+ years, just assume that the following exist:
- extension-less includes
- namespaces
- std::
- static_cast
- bool


byrnHDF:
m4/aclocal_cxx.m4 is not our file, it is from the m4 site - maybe we need to update it from the m4 download, but otherwise shouldn't alter it.

seanm:
> m4/aclocal_cxx.m4 is not our file, it is from the m4 site

I've undone my changes to that file then.

derobins (approved):

bmribler (approved):

PR #450, opened 2021-03-09
1 10 Fix java data export functions

byrnHDF:
Reopened HDFFV-11212


byrnHDF:
Examples are just HDF5Constants.H5I_INVALID_HID replacing -1.

.github/workflows/clang-format-check.yml
```diff
@@ -13,5 +13,11 @@ jobs:
...
         extensions: 'c,h,cpp,hpp'
         clangFormatVersion: 10
+        inplace: True
         style: file
         exclude: './config ./hl/src/H5LTanalyze.c ./hl/src/H5LTparse.c ./hl/src/H5LTparse.h ./src/H5Epubgen.h ./src/H5Einit.h ./src/H5Eterm.h ./src/H5Edefin.h ./src/H5version.h ./src/H5overflow.h'
+    - uses: EndBug/add-and-commit@v7
+      with:
+          author_name: github-actions
+          author_email: 41898282+github-actions[bot]@users.noreply.github.com
+          message: 'Committing clang-format changes'
```
jhendersonHDF: Have these changes already made it into develop? This looks like it should be in a separate PR.

java/src/hdf/hdf5lib/H5.java
```diff
@@ -523,7 +523,28 @@ public synchronized static native int H5set_free_list_limits(int reg_global_lim,
...
+     * @param binary_order
+     *            99 - export data as text.
+     *            1 - export data as binary Native Order.
+     *            2 - export data as binary Little Endian.
+     *            3 - export data as binary Big Endian.
+     *
+     * @exception HDF5LibraryException
+     *                - Error from the HDF-5 Library.
+     **/
+    public synchronized static native void H5export_attribute(String file_export_name, long dataset_id,
```
jhendersonHDF, line 547: Did the export attribute make it back into develop? I remember you had said this one wasn't ready for that PR, so it got pulled there.

byrnHDF: not yet, I started with 1.10 first (completing the original merge) and since that has different region specs was able to add it correctly. Will merge back up to develop.

.github/workflows/clang-format-check.yml
byrnHDF: Working on it here - needs trial runs.

jhendersonHDF: If we're making the change in the 1.10 branch, ok. I just thought we were going to use develop instead to test it.

byrnHDF: I needed to test it - and this was an active development for me.

.github/workflows/clang-format-check.yml
```diff
@@ -1,5 +1,6 @@
 name: clang-format Check
-on: [workflow_dispatch, push, pull_request]
+on:
+  pull_request:
```
jhendersonHDF, line 3: Does this mean that clang formats will only happen when the pull request is initially opened and won't happen on changes that happen during a review?

byrnHDF: So I ended up splitting this like the test checks between push and pull-request.
push will run the commit changes, pull-request will only check.

jhendersonHDF: Ah ok, I see the same changes in the other clang file now.

byrnHDF: yeah, everything was fine on a push to my fork, but it threw an error when I created the pull-request;

------------------------------------------------------------------------------------------------------------
Committing clang-format changes
This commit does not belong to any branch on this repository.

jhendersonHDF:
I still think the clang formatting checks should go into a separate PR so that it's easier to revert if need be, but everything else here looks fine to me.

jhendersonHDF (approved):

byrnHDF:
> I still think the clang formatting checks should go into a separate PR so that it's easier to revert if need be, but everything else here looks fine to me.

I'll do that for develop and 1.12.

PR #457, opened 2021-03-10
Replaces checks for fork, etc. with checks for unistd.h

derobins:
Code previously checked for individual POSIX API calls using
H5_HAVE_FORK, etc. The calls we use have been standardized for
decades and are available via unistd.h.

Some test messages that were missing when tests are skipped
due to a lack of unistd.h were also added.

The configure checks for individual POSIX API calls will be
removed in a later commit.

### DIFF
--- a/test/accum.c
+++ b/test/accum.c
@@ -2092,7 +2092,7 @@ error:
 unsigned
 test_swmr_write_big(hbool_t newest_format)
 {
-#if defined(H5_HAVE_FORK) && defined(H5_HAVE_WAITPID) && defined(H5_HAVE_UNISTD_H)
+#ifdef H5_HAVE_UNISTD_H
     hid_t    fid  = -1;   /* File ID */
     hid_t    fapl = -1;   /* File access property list */
     H5F_t *  rf   = NULL; /* File pointer */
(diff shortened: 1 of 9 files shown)


fortnern:
Are we certain there are no systems people are running HDF5 on that have fork, etc. but not unistd.h?

derobins:
> Are we certain there are no systems people are running HDF5 on that have fork, etc. but not unistd.h?

I know of no systems that do this. fork, etc. are POSIX and unistd.h has been the standard place for these things since the 90s. @gnuoyd or @jrmainzer might have more insight.

qkoziol:
> > Are we certain there are no systems people are running HDF5 on that have fork, etc. but not unistd.h?
> 
> I know of no systems that do this. fork, etc. are POSIX and unistd.h has been the standard place for these things since the 90s. @gnuoyd or @jrmainzer might have more insight.

fork() is part of the POSIX standard and will always be in unistd.h on POSIX compliant systems.

qkoziol (approved):
Good cleanup

derobins:
Technically, waitpid() is in sys/wait.h, but what we're really checking for here is POSIX-ness (as opposed to Win32, which is what we're really trying to exclude) and a single check for unistd.h does that.

jhendersonHDF (approved):

Baljak:
This is wrong and should be reverted: Mingw does have `unistd.h` but does not have `sys/wait` nor `fork`. As such, HDF5 does not compile anymore when using Mingw standalone. It works with MSYS though, since MSYS provides the missing POSIX functions.

PR #466, opened 2021-03-11
Brings the thread-safety recursive writer locks to 1.12

derobins:
Needed for the 1.12.1 release. Will go to develop in the future, but that's less straightforward due to the thread-safety changes made for async.

This feature is only supported on POSIX systems with Pthreads.

This will get a RELEASE.txt note before it's merged in.


byrnHDF (approved):

jrmainzer (approved):

lrknox (changes requested):
Need release note

src/H5TS.c
```diff
@@ -729,4 +737,856 @@ H5TS_create_thread(void *(*func)(void *), H5TS_attr_t *attr, void *udata)
...
+
+        ret_value = FAIL;
+    }
+    else {
+
+        /* we are commited to the takedown at this point.  Set magic
+         * to an invalid value, and call the appropriate pthread
+         * destroy routines.  Call them all, even if one fails along
+         * the way.
+         */
```
gnuoyd, line 831: The comment says "Call them all," but the code stops on the first failure. What is the intended behavior here?

src/H5TS.c
```diff
@@ -729,4 +737,856 @@ H5TS_create_thread(void *(*func)(void *), H5TS_attr_t *attr, void *udata)
...
+{
+    hbool_t                    have_mutex = FALSE;
+    int                        result;
+    H5TS_pt_rec_entry_count_t *count_ptr;
+    herr_t                     ret_value = SUCCEED;
+
+    if ((rw_lock_ptr == NULL) || (rw_lock_ptr->magic != H5TS_PT_REC_RW_LOCK_MAGIC)) {
+
+        ret_value = FAIL;
+    }
```
gnuoyd, line 876: Disparate error conditions are lumped under FAIL. Can we make the error reporting fine-grained?

For example, if the pointer is NULL or the magic is incorrect, the program is off the rails. Some UNIX systems would return EINVAL. However, if the call would "merely" deadlock, the program could at least try to recover—EDEADLK. Elsewhere, _rdlock returns the same error number, FAIL, for pthreads failures, inconsistent state, or memory-allocation failure. Other systems assign discrete numbers to some of those errors, too.

If a caller is supposed to perform its own deadlock recovery, then it seems like at least three error independent error numbers are needed: success (lock acquired), failure (would deadlock), all other errors (recovery unlikely/impossible).

src/H5TS.c
```diff
@@ -729,4 +737,856 @@ H5TS_create_thread(void *(*func)(void *), H5TS_attr_t *attr, void *udata)
...
+                        rw_lock_ptr->waiting_readers_count++;
+
+                        result = pthread_cond_wait(&(rw_lock_ptr->readers_cv), &(rw_lock_ptr->mutex));
+
+                        rw_lock_ptr->waiting_readers_count--;
+
+                        if (result != 0) {
+
+                            ret_value = FAIL;
+                            break;
```
gnuoyd, line 929: A handful of `return` statements would shorten the 6-step staircase and clarify the program logic.

src/H5TS.c
```diff
@@ -729,4 +737,856 @@ H5TS_create_thread(void *(*func)(void *), H5TS_attr_t *attr, void *udata)
...
+        else {
+
+            have_mutex = TRUE;
+        }
+    }
+
+    /* If there is no specific data for this thread, this is an
+     * initial write lock request.
+     */
+    if (ret_value == SUCCEED) {
```
gnuoyd, line 1008: If you short-circuit all of the `ret_value` tests with `return` statements, then the routine collapses in size.

This expansive form is out of step with ordinary C programming practice, which ISTM conserves typing effort and time spent in the physical and mental acts of reading the code. How is deviation from ordinary practice justified for this code?

test/ttsafe_rec_rw_lock.c
```diff
@@ -0,0 +1,1310 @@
...
+            CHECK_I(result, "H5TS_pt_rec_rw_wrlock -- 1");
+
+            udata->write_locks_granted++;
+            udata->real_write_locks_granted++;
+            wr_locks_remaining--;
+            rec_lock_depth = 1;
+
+            while (rec_lock_depth > 0) {
+
+                if ((rec_lock_depth >= max_rec_lock_depth) || ((HDrand() % 2) == 0)) {
```
gnuoyd, line 539: HDrand() is implemented in terms of rand(3), which isn't guaranteed to have a lengthy period or equal randomness in all bits. Linux and *BSD manuals warn against using it for those reasons. On NetBSD, the  distribution of `HDrand() % 2 == 0` seems biased to `true`. The loop does not complete.

So far the loop always has completed when I check the condition `HDrand() > RAND_MAX / 2`, instead.

src/H5TS.c
```diff
@@ -729,4 +737,856 @@ H5TS_create_thread(void *(*func)(void *), H5TS_attr_t *attr, void *udata)
...
+
+    if ((rw_lock_ptr == NULL) || (rw_lock_ptr->magic != H5TS_PT_REC_RW_LOCK_MAGIC)) {
+
+        ret_value = FAIL;
+    }
+
+    /* obtain the mutex */
+    if (ret_value == SUCCEED) {
+
+        if (pthread_mutex_lock(&(rw_lock_ptr->mutex)) != 0) {
```
gnuoyd: The most frequently taken path through _wrlock is the one where the recursion count increases from 0 to 1—that is, when an application thread that is not already in the library enters the library for the first time. On that path, _wrlock acquires a mutex, allocates some memory, and increases a shared statistic counter. That's awfully expensive.

Have you compared the number of _wrlock/_unlock cycles you can perform in a second with the number for a plain mutex or for a plain reader/writer lock?

gnuoyd: A thread that already holds the lock should be able to see that by looking at its own thread-local storage—no mutex acquire/release needed. I was expecting to find a fast path that increased the thread-local hold count and increased a thread-local statistic.

src/H5TSprivate.h
```diff
@@ -73,6 +73,150 @@ H5_DLL herr_t        H5TS_win32_thread_exit(void);
...
+    HDassert((count_ptr)->rec_lock_count >= 0);                               \
+    HDassert((count_ptr)->write_lock);                                        \
+                                                                              \
+    (rw)->stats.write_locks_released++;                                       \
+                                                                              \
+    if ( (count_ptr)->rec_lock_count == 0) {                                  \
+                                                                              \
+        (rw)->stats.real_write_locks_released++;                              \
+    }                                                                         \
+} while ( FALSE ) /* end REC_RW_LOCK_STATS__UPDATE_FOR_WR_UNLOCK */
```
gnuoyd: Why is the statistics counting in macros? Consider that the macros can be made into static inline functions that get straight to the point. For example:
```
static inline void
update_stats_for_wr_unlock(..._t *rw, ..._t *count_ptr)
{
    rw->stats.write_locks_released++;

    if (count_ptr->rec_lock_count == 0)
        rw->stats.real_write_locks_released++;
}
```

src/H5TSprivate.h
```diff
@@ -90,6 +234,255 @@ typedef pthread_mutex_t H5TS_mutex_simple_t;
...
+ *              avoid changes while the statistics are being read.  Collecting
+ *              them into a structure facilitates this.
+ *
+ ******************************************************************************/
+
+#define H5TS_PT_REC_RW_LOCK_MAGIC 0XABCD
+
+#define H5TS__RW_LOCK_POLICY__FAVOR_WRITERS 0
+
+typedef struct H5TS_pt_rec_rw_lock_t {
```
gnuoyd: Do not forget that programmers need to read and to type these things. The names of types, functions, and macros in this patch are downright ponderous.

`_pt_`? POSIX Threads?

`H5TS_rec_rwlock_t`? `H5TS_rrwlock_t`?

test/ttsafe_rec_rw_lock.c
```diff
@@ -0,0 +1,1310 @@
...
+                                                   /* real_write_locks_released      = */ 0,
+                                                   /* max_write_locks                = */ 0,
+                                                   /* max_write_lock_recursion_depth = */ 0,
+                                                   /* write_locks_delayed            = */ 0,
+                                                   /* max_write_locks_pending        = */ 0};
+    struct H5TS_pt_rec_rw_lock_t rec_rw_lock;
+
+    /* allocate the udata on the heap to shut up the compiler. */
+    udata =
+        (struct rec_rw_lock_test_udata_t *)malloc(sizeof(struct rec_rw_lock_test_udata_t) * MAX_NUM_THREADS);
```
gnuoyd: This looks like a good place to use the well-known idiom, `udata = malloc(sizeof(*udata) * MAX_NUM_THREADS);`, which is brief and robust.

test/ttsafe_rec_rw_lock.c
```diff
@@ -0,0 +1,1310 @@
...
+
+    /* discard the udata if it exists */
+    if (udata) {
+
+        HDfree(udata);
+    }
+
+    return;
+
+} /* end tts_rec_rw_lock_smoke_check_4() */
```
gnuoyd, line 1296: The read/write/mixed mob tests look a lot alike. The duplicate code cannot be reduced by extracting some common code into a function or two?

gnuoyd (changes requested):
I have sprinkled comments through the code. Here are some general questions and remarks:

What problem is the recursive reader/writer lock intended to solve? How does it solve that problem?

POSIX provides a reader/writer lock. Is there any reason that this implementation starts from scratch instead of augmenting that lock?

The code is unusually lengthy and obscure for what it does. This lock will be slower to acquire/release than implementations that are readily imagined or discovered in a web search. Can we justify the use of a slow, obscure implementation for this lock?

src/H5TSprivate.h
```diff
@@ -332,24 +380,24 @@ typedef struct H5TS_pt_rec_rw_lock_stats_t {
...
 
 /******************************************************************************
  *
- * Structure H5TS_pt_rec_rw_lock_t
+ * Structure H5TS_rw_lock_t
  *
  * A read / write lock, is a lock that allows either an arbitrary number
  * of readers, or a single writer into a critical region.  A recurssive
  * lock is one that allows a thread that already has a lock (be it read or
  * write) to successfully request the lock again, only droping the lock
```
lrknox: Typo here - missing second p in droping  - could be fixed while changes are being made to the file.

src/H5TSprivate.h
```diff
@@ -360,8 +408,8 @@ typedef struct H5TS_pt_rec_rw_lock_stats_t {
  *                                           JRM  -- 8/28/20
  *
  * magic:       Unsigned 32 bit integer field used for sanity checking.  This
- *              fields must always be set to H5TS_PT_REC_RW_LOCK_MAGIC.
- *              If this structure is allocated dynamically, rememver to set
+ *              fields must always be set to H5TS_RW_LOCK_MAGIC.
```
lrknox, line 284: This field (no s).

lrknox (approved):

src/H5TSprivate.h
derobins: These are static inline functions now

src/H5TSprivate.h
derobins: The names are all significantly shorter now

test/ttsafe_rec_rw_lock.c
derobins: Done

src/H5TS.c
derobins: These are invoked independently now

src/H5TSprivate.h
derobins: Fixed

src/H5TSprivate.h
derobins: Fixed

PR #470, opened 2021-03-13
Assume C99 fixed sized ints exist, use them

seanm:


seanm:
@qkoziol @derobins is this an acceptable direction?  I can do the same for 8, 16, and 64.

qkoziol:
The standard seems to indicate that these are not _required_ by a compiler, which is why our previous coding provided workarounds when they weren't available.   It's possible that we'd like to require them for HDF5 to work correctly now, but I don't know if there's any situations where that might be a problem for any users.

derobins (changes requested):
Let's do the H5public.h changes in a separate PR so we can merge the internal changes to 1.12 and 1.10.

derobins:
We directly use C99 types like uint64_t, uint32_t, etc. in many places in HDF5. If even Microsoft supports it, I think it's safe to assume it's ubiquitous. I'm definitely in favor of reducing our ifdef madness by eliminating obsolete cruft.

seanm:
> Let's do the H5public.h changes in a separate PR so we can merge the internal changes to 1.12 and 1.10.

In this case though, [line 293](https://github.com/HDFGroup/hdf5/blob/24c83cf73ebb80a58ce92463dac02713d3a4dbf8/src/H5public.h#L293) is always true, is it not?  The whole chunk is dead code.

qkoziol:
> We directly use C99 types like uint64_t, uint32_t, etc. in many places in HDF5. If even Microsoft supports it, I think it's safe to assume it's ubiquitous. I'm definitely in favor of reducing our ifdef madness by eliminating obsolete cruft.

We use it because we typedef it when it's not defined by the compiler. 

I _do_ think it's OK to get rid of our backward compatibility shims at this point though. :-)

qkoziol (approved):
As long as this passes for all CI & nightly regression tests, should be good.

epourmal:
Yes, let’s do it!

Thank you, all!

On Mar 15, 2021, at 11:47 AM, Quincey Koziol ***@***.***> wrote:

﻿

We directly use C99 types like uint64_t, uint32_t, etc. in many places in HDF5. If even Microsoft supports it, I think it's safe to assume it's ubiquitous. I'm definitely in favor of reducing our ifdef madness by eliminating obsolete cruft.

We use it because we typedef it when it's not defined by the compiler.

I do think it's OK to get rid of our backward compatibility shims at this point though. :-)

seanm:
OK, I'll keep working on this PR then...

byrnHDF (changes requested):
This define, H5_SIZEOF_LONG_DOUBLE, cannot be removed anywhere because not all platforms support long double. Windows insists that long double is just double and this define will be zero on that platform.

seanm:
> This define, H5_SIZEOF_LONG_DOUBLE, cannot be removed anywhere because not all platforms support long double. Windows insists that long double is just double...

`long double` has been in C since C89. That's not to say that it needs to be any bigger than `double`, just as `long int` is not necessarily any bigger than `int`.

>and this define will be zero on that platform

Really?!  It's not 8, just like `H5_SIZEOF_DOUBLE`?  The comment above it says:

```
/* The size of `long double', as computed by sizeof. */
#cmakedefine H5_SIZEOF_LONG_DOUBLE @H5_SIZEOF_LONG_DOUBLE@
```

And [MS docs](https://docs.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp?view=msvc-160) say `sizeof(long double)` is 8.

byrnHDF:
> Really?! It's not 8, just like `H5_SIZEOF_DOUBLE`? The comment above it

It is 0 because the code it encloses would be used, in tools lib the else code of the same if define is used.

byrnHDF:
> > Really?! It's not 8, just like `H5_SIZEOF_DOUBLE`? The comment above it
> 
> It is 0 because the code it encloses would be used, in tools lib the else code of the same if define is used.

I just double-checked that code and realized I might be wrong, the code in the ifdef block might be included and the "if type equals" fail the result will be the same that I see.

byrnHDF (commented):
Need to check a windows build, to verify the define is not 0.

brtnfld (approved):
Fortran-related changes look good.

seanm:
> I just double-checked that code and realized I might be wrong, the code in the ifdef block might be included and the "if type equals" fail the result will be the same that I see.

Could you point me to the code to which you refer?

byrnHDF:
> > I just double-checked that code and realized I might be wrong, the code in the ifdef block might be included and the "if type equals" fail the result will be the same that I see.
> 
> Could you point me to the code to which you refer?

tools/lib/h5tools_dump.c change.

byrnHDF:
Note that tests have not passed, because we have the clang format commit action, you have to look at the results of the previous commit for the test results.
select checks, then click the drop down next to the git hash after "Committing clang-format changes ", scroll to the bottom minus one and select to get the test results.

seanm:
> Need to check a windows build, to verify the define is not 0.

I'm afraid I don't have Windows, would you be able to check this?

byrnHDF:
> > Need to check a windows build, to verify the define is not 0.
> 
> I'm afraid I don't have Windows, would you be able to check this?

The github commit actions should be good enough for now (which failed on the last code commit).

byrnHDF (approved):

byrnHDF:
Ubuntu TS GCC failed dtransform test:
57: Testing contiguous, no data type conversion (ldouble->ldouble)        *FAILED*
57: 
57: CMake Error at /home/runner/work/hdf5/hdf5/config/cmake_ext_mod/runTest.cmake:124 (message):
57:   Failed: Test program /home/runner/work/hdf5/build/bin/dtransform exited !=

dtransform lines 176-179:
        TESTING("contiguous, no data type conversion (" TEST_STR "->" TEST_STR ")")                          \
                                                                                                             \
        if (H5Dread(dset, HDF_TYPE, H5S_ALL, H5S_ALL, XFORM, array) < 0)                                     \
            TEST_ERROR;                                                                                      \

byrnHDF:
Looks like the problem is elsewhere because the test has been passing:
57: Testing contiguous, with type conversion (float->double)               PASSED
57: Testing contiguous, no data type conversion (ldouble->ldouble)         PASSED
57: Testing contiguous, byte order conversion (ldouble->ldouble)           PASSED
57: Testing contiguous, with type conversion (float->ldouble)              PASSED

byrnHDF:
> Looks like the problem is elsewhere because the test has been passing:
> 57: Testing contiguous, with type conversion (float->double) PASSED
> 57: Testing contiguous, no data type conversion (ldouble->ldouble) PASSED
> 57: Testing contiguous, byte order conversion (ldouble->ldouble) PASSED
> 57: Testing contiguous, with type conversion (float->ldouble) PASSED

Windows dtransform was passing with the same output.

qkoziol (approved):

src/H5Ztrans.c
```diff
@@ -178,8 +177,6 @@ static void       H5Z__xform_reduce_tree(H5Z_node *tree);
             H5Z_XFORM_DO_OP1((RESL), (RESR), float, OP, (SIZE))                                              \
         else if ((TYPE) == H5T_NATIVE_DOUBLE)                                                                \
             H5Z_XFORM_DO_OP1((RESL), (RESR), double, OP, (SIZE))                                             \
-        else if ((TYPE) == H5T_NATIVE_LDOUBLE)                                                               \
```
byrnHDF: Why remove?

src/H5Ztrans.c
```diff
@@ -208,69 +205,8 @@ static void       H5Z__xform_reduce_tree(H5Z_node *tree);
...
```
byrnHDF: Why remove?

seanm: Good catch. I removed the wrong half of that big #if. Fixed.

src/H5Ztrans.c
seanm: Good catch. I removed the wrong half of that big #if. Fixed.

lrknox:
This is failing a test in dtransform.c on Ubuntu with gcc 9.3.0, both before and after the latest commit:

Testing contiguous, no data type conversion (ldouble->ldouble) 

It seems the calculated value for the first number in the array is "nan" instead of "2.22222".

seanm:
@lrknox why the new commit f4a044b?  Is it to undo a799020c68f58e8deb50038e4803852f5953e507?  I separated a799020c68f58e8deb50038e4803852f5953e507 as @derobins requested, and can drop that commit when this is no longer WIP.

But I still don't see the need.  When can `#if H5_SIZEOF_INT64_T >= 8` be false? You realize all that code only does anything in such dead branches, right?

lrknox (approved):

lrknox:
> 
> 
> Let's do the H5public.h changes in a separate PR so we can merge the internal changes to 1.12 and 1.10.

@derobins  H5public.h changes have been removed from this PR.

lrknox:
> 
> 
> @lrknox why the new commit [f4a044b](https://github.com/HDFGroup/hdf5/commit/f4a044b8757b54ca4d7c229bae066d4d91520889)? Is it to undo [a799020](https://github.com/HDFGroup/hdf5/commit/a799020c68f58e8deb50038e4803852f5953e507)? I separated [a799020](https://github.com/HDFGroup/hdf5/commit/a799020c68f58e8deb50038e4803852f5953e507) as @derobins requested, and can drop that commit when this is no longer WIP.
> 
> But I still don't see the need. When can `#if H5_SIZEOF_INT64_T >= 8` be false? You realize all that code only does anything in such dead branches, right?

@seanm  Sorry, did not see this comment.   I was preparing to merge this when @derobins was satisfied, but is your WIP indicating you aren't ready for merge?

seanm:
>WIP indicating you aren't ready for merge?

Correct.  For one, there are apparently tests failing, which I have yet to look at. There are also other C99 types I intend to clean up. I really created this PR to have a place to discuss which kinds of changes y'all would or would not accept.

seanm:
@qkoziol @derobins @lrknox @byrnHDF @epourmal OK, I found what was making tests fail.  I think this is ready now.

byrnHDF (approved):

lrknox:
All checks passed after the clang-format change commit.  See https://github.com/HDFGroup/hdf5/actions/runs/1382094984.

derobins (approved):

PR #477, opened 2021-03-17
Suppresses the tcheck_version test's abort dialog on Windows

derobins:
Windows raises a modal abort/retry/ignore dialog box when CRT
calls abort(). This change installs a report hook that suppresses
the dialog so that the CMake tests don't time out waiting for a
nonexistent user to click a dialog box.

### DIFF
--- a/test/tcheck_version.c
+++ b/test/tcheck_version.c
@@ -118,8 +118,7 @@ abort_intercept(int H5_ATTR_UNUSED sig)
  * Returning TRUE here lets Windows know that we've handled the abort() and that there
  * is no need to alert the user with a modal dialog box.
  */
-int __cdecl
-handle_crt_abort(int reportType, char *message, int *returnValue)
+int __cdecl handle_crt_abort(int reportType, char *message, int *returnValue)
 {
     return TRUE;
 }
--- a/test/tcheck_version.c
+++ b/test/tcheck_version.c
@@ -29,6 +29,10 @@
 
 #include "h5test.h"
 
+#ifdef H5_HAVE_WIN32_API
+#include <crtdbg.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


test/tcheck_version.c
```diff
@@ -108,12 +112,31 @@ abort_intercept(int H5_ATTR_UNUSED sig)
...
 {
+#ifdef H5_HAVE_WIN32_API
+    (void)_CrtSetReportHook2(_CRT_RPTHOOK_INSTALL, handle_crt_abort);
+#endif
     parse(ac, av);
     HDsignal(SIGABRT, &abort_intercept);
     H5check_version(major, minor, release);
     HDsignal(SIGABRT, SIG_DFL);
+#ifdef H5_HAVE_WIN32_API
+    (void)_CrtSetReportHook2(_CRT_RPTHOOK_REMOVE, handle_crt_abort);
```
derobins, line 139: Uninstalling the handler is overkill in this simple program, but that's the pattern to use on the off chance this gets cut-n-pasted someplace else.

test/tcheck_version.c
```diff
@@ -108,12 +112,31 @@ abort_intercept(int H5_ATTR_UNUSED sig)
...
+    return TRUE;
+}
+#endif
+
 int
 main(int ac, char **av)
 {
+#ifdef H5_HAVE_WIN32_API
+    (void)_CrtSetReportHook2(_CRT_RPTHOOK_INSTALL, handle_crt_abort);
+#endif
```
derobins, line 133: No need to check the return value since the test will be correct (albeit with the modal dialog box popping up) even if this call fails for some reason.

qkoziol (approved):

byrnHDF (approved):

PR #480, opened 2021-03-19
File locks now work on Windows

derobins:
Uses LockFileEx() and UnlockFileEx(). Fixes HDFFV-10191 (partial).


qkoziol (approved):

derobins:
What is going on with that H5Tconv.c file?

byrnHDF (approved):

byrnHDF:
> What is going on with that H5Tconv.c file?

Aligning equals.

derobins:
The Java test is failing because H5Fis_accessible() opens a newly-created HDF5 file via a new file handle, which is not allowed under Windows' mandatory file locking scheme. I feel like I'm going to have to fundamentally rework that API call.

src/H5Fint.c
```diff
@@ -3718,10 +3729,17 @@ H5F__start_swmr_write(H5F_t *f)
...
 
     /* Place an advisory lock on the file */
-    if (H5F_USE_FILE_LOCKING(f))
+    if (H5F_USE_FILE_LOCKING(f)) {
+        /* Have to unlock on Windows as Win32 doesn't support changing the lock
+         * type (exclusive vs shared) with a second call.
+         */
+        if (H5FD_unlock(f->shared->lf) < 0) {
+            HGOTO_ERROR(H5E_FILE, H5E_CANTUNLOCKFILE, FAIL, "unable to unlock the file")
+        }
```
derobins, line 3738: The unlock is only required on Windows (currently), but I went ahead and unlocked all the time. I could hide this in an ifdef if anyone feels strongly about it.

test/tfile.c
```diff
@@ -1666,6 +1666,29 @@ test_file_is_accessible(const char *env_h5_drvr)
...
+    CHECK(fid, H5I_INVALID_HID, "H5Fcreate");
+
+    /* Verify that the file is an HDF5 file */
+    is_hdf5 = H5Fis_accessible(filename, fapl_id);
+    VERIFY(is_hdf5, TRUE, "H5Fis_accessible");
+
+    /* Close file */
+    ret = H5Fclose(fid);
+    CHECK(ret, FAIL, "H5Fclose");
+
```
derobins, line 1691: This issue was originally exposed through the Java tests.

lrknox (approved):

PR #484, opened 2021-03-19
Apply test filename composition also to t_bigio to account for HDF5_PARAPREFIX.

tjahns:
* Other parallel tests can be instructed (via the HDF5_PARAPREFIX
  environment variable) to write test files to another directory.
  This change adds a corresponding capability to testpar/t_bigio.

### DIFF
--- a/testpar/t_bigio.c
+++ b/testpar/t_bigio.c
@@ -5,6 +5,7 @@
 
 /* FILENAME and filenames must have the same number of names */
 const char *FILENAME[3] = {"bigio_test.h5", "single_rank_independent_io.h5", NULL};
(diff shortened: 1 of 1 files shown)


qkoziol (changes requested):
Definitely a change that should be made, but the rest of the tests create these temporary filenames locally, within each test, to accommodate different FAPLs that can be passed in.   Could you please refactor this change to not use a global array of strings and use a local one within each test routine?  For example, the test_file_delete() test in test/tfile.c is a good example to draw from.

derobins:
There's been no motion on refactoring this for some time. Closing.

PR #492, opened 2021-03-21
Updates HSYS_GOTO_ERROR to emit GetLastError() values on Win32

derobins:


byrnHDF (approved):

lrknox (approved):

src/H5Eprivate.h
```diff
@@ -129,6 +130,35 @@ typedef struct H5E_t H5E_t;
         HGOTO_ERROR(majorcode, minorcode, retcode, "%s, errno = %d, error message = '%s'", str, myerrno,     \
```
jhendersonHDF, line 130: It's interesting that this does a 'goto' when the DONE_ macros are usually used past a goto label specifically. The fact that the new windows implementation does not use a goto might be confusing or even problematic. I think neither of them should use a 'goto', or both should, as long as we are consistent.

src/H5Eprivate.h
```diff
@@ -129,6 +130,35 @@ typedef struct H5E_t H5E_t;
...
+        int   myerrno   = errno;                                                                             \
+        DWORD win_error = GetLastError();                                                                    \
+        /* Other projects may rely on the description format to get the errno and any changes should be      \
+         * considered as an API change                                                                       \
+         */                                                                                                  \
+        HDONE_ERROR(majorcode, minorcode, retcode,                                                           \
+                    "%s, errno = %d, error message = '%s', Win32 GetLastError() = %" PRIu32 "", str,         \
+                    myerrno, HDstrerror(myerrno), win_error);                                                \
+    }
+#define HSYS_GOTO_ERROR(majorcode, minorcode, retcode, str)                                                  \
```
jhendersonHDF, line 150: Should this get a non-windows implementation as well so that you don't have to bracket its usage with an #ifdef, or does it already have an implementation elsewhere?

src/H5Eprivate.h
derobins: I'm not sure what you mean here. The Win32 implementation is a copy of the POSIX one and in both cases there is a DONE <=> DONE and GOTO <=> GOTO correspondence.

derobins: Is there a typo I'm not seeing?

jhendersonHDF: Ah, my mistake, ignore this and the other comment. GitHub's shortening of the diff made it appear as though the non-windows implementation for HSYS_DONE_ERROR used the HGOTO_ERROR from HSYS_GOTO_ERROR.

jhendersonHDF (approved):

PR #497, opened 2021-03-22
Modernize H5public.h

derobins:
Pulls out most of the cruft from H5public.h

Caveats:
* Minimal testing on non-Linux platforms
* No CMake updates yet
* Definitely can't go to 1.12 or earlier


src/H5public.h
```diff
@@ -140,172 +106,41 @@ typedef int herr_t;
...
-#ifndef __cplusplus
-#if defined(H5_SIZEOF_BOOL) && (H5_SIZEOF_BOOL != 0)
-#define bool _Bool
-#else
-#define bool unsigned int
-#endif
-#define true 1
-#define false 0
-#endif /* __cplusplus */
-#endif /* H5_HAVE_STDBOOL_H */
```
derobins: We should also change internal use of hbool_t to bool and ditch TRUE/FALSE in lieu of true/false.

src/H5FDmulti.c
```diff
@@ -35,6 +35,25 @@
...
+#define H5_GCC_DIAG_DO_PRAGMA(x)  _Pragma(#x)
+#define H5_GCC_DIAG_PRAGMA(x)     H5_GCC_DIAG_DO_PRAGMA(GCC diagnostic x)
+
+#define H5_GCC_DIAG_OFF(x) H5_GCC_DIAG_PRAGMA(push) H5_GCC_DIAG_PRAGMA(ignored H5_GCC_DIAG_JOINSTR("-W", x))
+#define H5_GCC_DIAG_ON(x)  H5_GCC_DIAG_PRAGMA(pop)
+#else
+#define H5_GCC_DIAG_OFF(x)
+#define H5_GCC_DIAG_ON(x)
+#endif
+
```
derobins: This does not belong in the public header. The main reason for it previously being in H5public.h is its use in this file, so I copied it here for the time being and put the version used by the rest of the library in H5private.h. We can move it to a separate header file if we really care about the code duplication (maybe H5warning.h or something).

src/H5private.h
```diff
@@ -42,21 +47,18 @@
 #endif /* H5_HAVE_THREADSAFE */
 
 /*
- * Include ANSI-C header files.
+ * Include ANSI-C header files that aren't included in H5public.h
  */
-#ifdef H5_STDC_HEADERS
```
derobins: lol

src/H5private.h
```diff
@@ -89,25 +89,6 @@
...
- */
-#if defined(H5_TIME_WITH_SYS_TIME)
-#include <sys/time.h>
-#include <time.h>
-#elif defined(H5_HAVE_SYS_TIME_H)
-#include <sys/time.h>
-#else
-#include <time.h>
-#endif
-
```
derobins, line 98: This is no longer necessary.

src/H5private.h
byrnHDF: Have you done a search for the H5_STDC_HEADERS? I seem to recall tripping over it somewhere else.

byrnHDF (approved):

src/H5private.h
derobins: I'll look again, but I don't think this is used anywhere else.

derobins: Update: I yanked it from the perform code.

src/H5FDmulti.c
qkoziol: I agree with you - this shouldn't be in H5public.h.   I doubt H5warning.h would work though, since this VFD requires a public header and I don't think H5warning.h is great for a public header.   Probably duplicating the code is fine, although I would add a comment here and in H5private.h that the macros should be kept the same.

Also - I wouldn't use the plain "H5" prefix here, but probably "H5FD_MULTI" or something similar.

derobins: Good point. I'll change that the next time I make changes to this PR.

src/H5public.h
```diff
@@ -140,172 +106,41 @@ typedef int herr_t;
...
-#define UINT64_MAX ULLONG_MAX
-#undef H5_SIZEOF_UINT64_T
-#define H5_SIZEOF_UINT64_T H5_SIZEOF_LONG_LONG
-#else
-#error "nothing appropriate for uint64_t"
-#endif
-
 /*
- * The sizes of file objects have their own types defined here, use a minimum
- * 64-bit type.
```
qkoziol: The "minimum 64-bit type" is important - hsize_t and haddr_t are not explicitly 64-bit values - they are supposed to be the largest integer available on the platforms the HDF5 supports.  Although it's unlikely that we'll get to 128-bit file sizes very soon, it's definitely on the horizon and probably will happen within 10 years.

qkoziol: Instead of deleting the "minimum 64-bit type", it would be better to add more information to the comment to indicate this.

src/H5public.h
qkoziol: It's also entirely reasonable to retire hbool_t at this point.

qkoziol:
Aside from some very minor caveats, this looks great. :-)

gnuoyd (approved):
Definitely a step in the right direction.

soumagne (approved):

PR #498, opened 2021-03-22
Ubsan fixes

seanm:


seanm:
Of the 4 commits here:

- Don't treat 2d array as 1d array, fixing UBSan complaint in `CPP_test…
  - this I think is correct
- Fixed use of null pointer identified by UBSan
  - this I'm not sure is the right fix. Thoughts?
- WIP: investigation of undefined float -> unsigned char conversion in …
  - this I dunno how to fix. suggestions?
- Removed dead skip_overflow_tests_g global
  - this I think is correct

hl/test/test_image.c
```diff
@@ -665,10 +665,17 @@ test_generate(void)
     HL_TESTING2("make indexed image from sea data");
 
     for (i = 0; i < n_elements; i++) {
-        if (data[i] > 0)
+        if (data[i] > 0.0f)
             image_data[i] = 0;
-        else
-            image_data[i] = (unsigned char)((255 * (data[i] - xmin)) / xmin);
+        else {
+            assert(xmin < 0.0f); // never occurs
```
derobins: We slap HD in front of all standard library calls. e.g., HDassert().

seanm: Uh, I think you misunderstand, this is just test code illustrating the issue, not code to keep. :)

tools/src/h5repack/h5repack.c
```diff
@@ -242,7 +242,7 @@ copy_named_datatype(hid_t type_in, hid_t fidout, named_dt_t **named_dt_head_p, t
         while (dt && token_cmp) {
             dt = dt->next;
 
-            if (H5Otoken_cmp(type_in, &dt->obj_token, &oinfo.token, &token_cmp) < 0)
+            if (dt && H5Otoken_cmp(type_in, &dt->obj_token, &oinfo.token, &token_cmp) < 0)
                 H5TOOLS_GOTO_ERROR(H5I_INVALID_HID, "failed to compare object tokens");
         }
```
gnuoyd, line 242: I think your fix is correct but the logic here is a bit convoluted. Lines 238 through 247 appear to boil down to a linked-list search:

```
/* Search the stack for the datatype. */
for (dt = *named_dt_head_p; dt != NULL; dt = dt->next) {
    if (H5Otoken_cmp(type_in, &dt->obj_token, &oinfo.token, &token_cmp) < 0)
        H5TOOLS_GOTO_ERROR(H5I_INVALID_HID, "failed to compare object tokens");

    if (token_cmp == 0)
        break; // found it!
}
```

What do you think?

seanm: Indeed seems nicer, will make that change...

hl/test/test_image.c
```diff
@@ -665,10 +665,17 @@ test_generate(void)
...
+
+            float temp = (255.0f * (data[i] - xmin)) / xmin;
+
+            assert(temp > -0.5f && temp < 255.5); // always occurs
+
+            image_data[i] = (unsigned char)(temp); // maybe call fabsf()?
+        }
```
gnuoyd, line 678: What UB was found in this function?

BTW, the error checking on the fscanf(3) calls looks wrong. `valex`, `xmin`, and `xmax` could all be undefined.

seanm: As per the assert, `temp` is outside the range (-0.5, 255.5) thus converting it to uint8 is nonsensical.

c++/test/ttypes.cpp
```diff
@@ -726,8 +725,11 @@ test_named()
 
         // It should be possible to define an attribute for the named type
         Attribute attr1 = itype.createAttribute("attr1", PredType::NATIVE_UCHAR, space);
-        for (i = 0; i < ds_size[0] * ds_size[1]; i++)
-            attr_data[0][i] = (int)i; /*tricky*/
+        for (hsize_t i = 0; i < ds_size[0]; i++) {
+            for (hsize_t j = 0; j < ds_size[1]; j++) {
+                attr_data[i][j] = (unsigned)i;
```
qkoziol: Probably doesn't matter, but this should be:

attr_data[i][j] = ((unsigned)i * ds_size[1]) + j;

config/cmake/ConversionTests.c
```diff
@@ -11,7 +11,7 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
 #if defined(__has_attribute)
-#if __has_attribute(no_sanitize)
+#if 0 // __has_attribute(no_sanitize)
```
qkoziol: ?

test/dt_arith.c
```diff
@@ -5403,9 +5313,6 @@ main(void)
      * for user-defined integer types */
     nerrors += (unsigned long)test_derived_integer();
 
-    /* Does floating point overflow generate a SIGFPE? */
-    generates_sigfpe();
```
qkoziol, line 5410: I believe that this is important - are you thinking about taking it out?

hl/test/test_image.c
gnuoyd: The expression assigned to `temp` on line 673 looks wrong to me. Probably should be `float temp = `255.0 * (data[i] - xmin) / (xmax - xmin);` if the intent is to map [xmin, xmax] to [0, 255].

seanm: OK, test passes with that change, and UBSan is also happy.  Thanks!

c++/test/ttypes.cpp
seanm: Good catch, stupid me.  In fact it should be `* ds_size[0]` I believe...

test/dt_arith.c
seanm: I am. It seems to do nothing but set `skip_overflow_tests_g` which in turn is read by nothing.

config/cmake/ConversionTests.c
seanm: As per the commit subject "TEMP: kill HDF_NO_UBSAN".  When this is no longer WIP, I'll remove that commit.  Hopefully we can eventually get to never turning UBSan off.

test/dt_arith.c
qkoziol: > I am. It seems to do nothing but set `skip_overflow_tests_g` which in turn is read by nothing.

I traced through the history of this file (which was originally part of the test/dtypes.c test) and indeed - the generates_sigfpe() routine and the skip_overflow_tests_g flag should have been removed when the test was overhauled in 2005.  Fine to get rid of them now!  :-)

qkoziol (approved):

test/dt_arith.c
seanm: Which is the easiest way to eliminate undefined behaviour, just remove the code entirely. :)

seanm:
@qkoziol ok since you've approved this, I've removed that TEMP commit and rebased, so this is ready to go IMHO.

c++/test/ttypes.cpp
```diff
@@ -726,8 +725,11 @@ test_named()
 
         // It should be possible to define an attribute for the named type
         Attribute attr1 = itype.createAttribute("attr1", PredType::NATIVE_UCHAR, space);
-        for (i = 0; i < ds_size[0] * ds_size[1]; i++)
-            attr_data[0][i] = (int)i; /*tricky*/
+        for (hsize_t i = 0; i < ds_size[0]; i++) {
+            for (hsize_t j = 0; j < ds_size[1]; j++) {
+                                attr_data[i][j] = (unsigned)(i * ds_size[0]] + j);
```
qkoziol: Ah, sorry, I was wrong about this when I get you a "thumbs up" earlier - it should be ds_size[1] here.

qkoziol (approved):

c++/test/ttypes.cpp
seanm: Really? OK, done.

qkoziol: Yes - 'i' goes in steps of the row size (ds_size[1]) down the array.

seanm:
@gnuoyd @derobins care to do a final review?  I think this is mergeable now.

derobins (approved):

lrknox:
Passed checks in CI #947.

PR #502, opened 2021-03-23
Update h5LT files with latest flex and Bison available on jelly 

lrknox:
@gnuoyd Change to H5LTanalyze.c line 126 eliminated warning "rule could not be matched".  Is the change correct?


derobins (approved):

byrnHDF (approved):

hl/src/H5LTparse.c
```diff
@@ -1643,419 +1486,418 @@ yyparse ()
...
-#line 169 "hl/src//H5LTparse.y"
+#line 169 "hl/src/H5LTparse.y" /* yacc.c:1646  */
     {   
                             size_t origin_size, new_size;
                             hid_t dtype_id = cmpd_stack[csindex].id;
 
                             /*Adjust size and insert member, consider both member size and offset.*/
                             if(cmpd_stack[csindex].first_memb) { /*reclaim the size 1 temporarily set*/
-                                new_size = H5Tget_size((yyvsp[(1) - (5)].hid)) + (yyvsp[(4) - (5)].ival);
+                                new_size = H5Tget_size((yyvsp[-4].hid)) + (yyvsp[-1].ival);
```
byrnHDF, line 1734: This looks weird! However that is how it is in 1.12.

hl/src/H5LTparse.y
```diff
@@ -315,7 +315,7 @@ strsize         :       H5T_VARIABLE_TOKEN     {$<ival>$ = H5T_VARIABLE_TOKEN;}
                 |       NUMBER
                 ;
 strpad          :       H5T_STR_NULLTERM_TOKEN {$<ival>$ = H5T_STR_NULLTERM_TOKEN;}
-                |       H5T_STR_NULLPAD_TOKEN  {$<ival>$ = H5T_STR_NULLPAD_TOKEN;}
```
byrnHDF: This seems like an issue? Why remove it?

gnuoyd:
"rule could not be matched" crops up because line 126 matches newlines and a previous line matches newlines with higher priority. The thing to do is remove line 126 and make sure the tests still pass.

hl/src/H5LTanalyze.l
```diff
@@ -123,7 +123,7 @@ OPQ_TAG          {return token(OPQ_TAG_TOKEN);}
 ":"              {return token(':');}
 ";"              {return token(';');}
 [ \t\n]*         ;
-"\n"             { return 0; }
+"[\n]"             { return 0; }
```
gnuoyd: BTW, I am pretty sure this new rule matches a newline in literal square brackets.

Without the double quotes, the rule would match a literal newline, and `flex` should warn that the rule will never match, because line 125 matches newlines, also, with higher priority.

lrknox:
Thanks, I’ll try that.

We have a failing test with CMake:  hl_test_lite with this output

Testing         text for string types                                 ERROR: syntax error before "H5T_STR_NULLPAD".

And these errors:
HDF5-DIAG: Error detected in HDF5 (1.13.0) thread 0:
  #000: /mnt/wrk/lrknox/hdf5-1.13.0/src/H5T.c line 2334 in H5Tis_variable_str(): not a datatype
    major: Invalid arguments to routine
    minor: Inappropriate type
HDF5-DIAG: Error detected in HDF5 (1.13.0) thread 0:
  #000: /mnt/wrk/lrknox/hdf5-1.13.0/src/H5Tstrpad.c line 54 in H5Tget_strpad(): not a datatype
    major: Invalid arguments to routine
    minor: Inappropriate type

Trying to figure out that problem.  Fails similarly on both github’s ubuntu and jelly.

gnuoyd:
On Tue, Mar 23, 2021 at 03:44:24PM -0700, Larry Knox wrote:
> Thanks, I’ll try that.
> 
> We have a failing test with CMake:  hl_test_lite with this output
> 
> Testing         text for string types                                 ERROR: syntax error before "H5T_STR_NULLPAD".

Does it have something to do with the deleted line in the .y file that
Allen pointed out?  I wondered if it was just an accident.

Dave

PR #520, opened 2021-03-25
Clean up MPI-IO VFD tracing support

qkoziol:


derobins (approved):

src/H5FDmpio.c
```diff
@@ -133,23 +133,26 @@ static const H5FD_class_mpi_t H5FD_mpio_g = {
 };
 
 #ifdef H5FDmpio_DEBUG
-/* Flags to control debug actions in H5Fmpio.
- * Meant to be indexed by characters.
- *
- * 'c' show result of MPI_Get_count after read
```
jhendersonHDF, line 139: Was the 'c' flag never implemented? Seems like something that would actually be helpful when debugging certain issues like reading/writing > 2GB where diagnosing failure from overflow on the byte count.

qkoziol: I don't believe that 'c' was ever implemented - I don't recall ever seeing it.

jhendersonHDF: Ok - might be something worth pursuing in the future then.

jhendersonHDF (approved):

qkoziol:
@lrknox - Same with this one: the code's been reviewed & approved, but the checks aren't reporting back. Is it ready to merge?

lrknox:
> 
> 
> @lrknox - Same with this one: the code's been reviewed & approved, but the checks aren't reporting back. Is it ready to merge?

I don't know that we've proven it yet, but it may be that waiting 5 minutes after committing code to a fork before creating a pull request may avoid the failure to report back.  I can check the actions for test results on the fork, and then override the requirement that the tests pass, which I'm doing for these two.  It's a little clumsy, but I'm getting more comfortable with it.

qkoziol:
> > @lrknox - Same with this one: the code's been reviewed & approved, but the checks aren't reporting back. Is it ready to merge?
> 
> I don't know that we've proven it yet, but it may be that waiting 5 minutes after committing code to a fork before creating a pull request may avoid the failure to report back. I can check the actions for test results on the fork, and then override the requirement that the tests pass, which I'm doing for these two. It's a little clumsy, but I'm getting more comfortable with it.

OK, I'll try that in the future.   Could you merge this one?

qkoziol:
Thanks @lrknox !

PR #521, opened 2021-03-25
Make custom command byproducts explicit for Ninja generators when CMake >= 3.3

### LINKED ISSUE #505: CMake with Ninja generator warns that Ninja requires custom command byproducts to be explicit for H5Tinit.c
When `H5Tinit.c` exists in the generated output directory, CMake generates rules which do not generate `H5Tinit.c` causing CMake >= 3.20.0 to issue a warning about policy CMP0058.

The policy warning I get is:
```
  CMake Warning (dev):
  Policy CMP0058 is not set: Ninja requires custom command byproducts to be
  explicit.  Run "cmake --help-policy CMP0058" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  This project specifies custom command DEPENDS on files in the build tree
  that are not specified as the OUTPUT or BYPRODUCTS of any
  add_custom_command or add_custom_target:

   3rdparty\hdf5\H5Tinit.c

  For compatibility with versions of CMake that did not have the BYPRODUCTS
  option, CMake is generating phony rules for such files to convince 'ninja'
  to build.

  Project authors should add the missing BYPRODUCTS or OUTPUT options to the
  custom commands that produce these files.
```
One possible solution to silence the error that I have found is to add a BYPRODUCTS parameter to the `add_custom_command` in `hdf5/src/CMakeLists.txt` which touches the `gen_SRCS.stamp1` timestamp file. See below for example implementation:

```
+  # Make custom command byproducts explicit for Ninja generators when CMake >= 3.3
+  if( POLICY CMP0058 )
+    set(NinjaByproducts BYPRODUCTS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c)
+  endif()
   add_custom_command (
       OUTPUT ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
+      ${NinjaByproducts}
       COMMAND    ${CMAKE_COMMAND}
       ARGS       -E touch ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
       DEPENDS    ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c
       WORKING_DIRECTORY ${HDF5_GENERATED_SOURCE_DIR}
   )
```
Disclaimer: I am relatively new to CMake, so this may not be the optimal solution.

My system specifications:
```
Windows 10
CMake 3.17.5
```

Thanks in advance!

DragonAxe:
Closes #505

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1126,8 +1126,13 @@ if (NOT EXISTS "${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c")
   endif ()
   set_source_files_properties (${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c PROPERTIES GENERATED TRUE)
 else ()
+  # Make custom command byproducts explicit for Ninja generators when CMake >= 3.3
+  if( POLICY CMP0058 )
+    set(NinjaByproducts BYPRODUCTS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c)
+  endif()
   add_custom_command (
       OUTPUT ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
+      ${NinjaByproducts}
       COMMAND    ${CMAKE_COMMAND}
       ARGS       -E touch ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1
       DEPENDS    ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c


DragonAxe:
After further testing, I have discovered that I get `ninja: error: dependency cycle: src/H5Tinit.c -> src/H5Tinit.c` after the second CMake and Build (cmake; build; cmake; build -> fails).
This is due to `DEPENDS    ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c` on line 1138. I'm investigating an alternative solution now.

PR #524, opened 2021-03-26
Brings the native implementation of H5Fdelete() from Bitbucket

derobins:


src/H5FDpublic.h
```diff
@@ -299,6 +299,7 @@ typedef struct H5FD_class_t {
     herr_t (*truncate)(H5FD_t *file, hid_t dxpl_id, hbool_t closing);
     herr_t (*lock)(H5FD_t *file, hbool_t rw);
     herr_t (*unlock)(H5FD_t *file);
+    herr_t (*del)(const char *name, hid_t fapl);
```
derobins, line 302: We're using del instead of delete since the latter is a C++ keyword.

soumagne (approved):
looks good to me.

src/H5FDdirect.c
```diff
@@ -170,6 +171,7 @@ static const H5FD_class_t H5FD_direct_g = {
     H5FD__direct_truncate,      /* truncate             */
     H5FD__direct_lock,          /* lock                 */
     H5FD__direct_unlock,        /* unlock               */
+    H5FD__direct_del,           /* del                  */
```
qkoziol: H5FD_direct_del -> H5FD_direct_delete

src/H5FDcore.c
```diff
@@ -1708,3 +1710,28 @@ H5FD__core_unlock(H5FD_t *_file)
...
+static herr_t
+H5FD__core_delete(const char *filename, hid_t H5_ATTR_UNUSED fapl_id)
+{
+    herr_t ret_value = SUCCEED; /* Return value             */
+
+    FUNC_ENTER_STATIC
+
+    HDassert(filename);
+
+    if (HDremove(filename) < 0)
```
qkoziol: Seems like HDremove() should only be called when the 'backing store' flag is set in the FAPL.

src/H5FDdirect.c
derobins: Fixed

src/H5FDmulti.c
```diff
@@ -2015,6 +2017,52 @@ open_members(H5FD_multi_t *file)
...
+
+    /* Delete each member file using the underlying fapl */
+    UNIQUE_MEMBERS (fa->memb_map, mt) {
+        assert(fa->memb_name[mt]);
+        assert(fa->memb_fapl[mt] >= 0);
+
+        /* Note: This truncates the user's filename down to only sizeof(tmp)
+         *      characters. -QK & JK, 2013/01/17
+         */
+        sprintf(full_filename, fa->memb_name[mt], filename);
```
qkoziol: Seems like snprintf was intended here?

testpar/t_file.c
```diff
@@ -948,3 +948,58 @@ test_file_properties(void)
...
+    /* close the file */
+    ret = H5Fclose(fid);
+    VRFY((SUCCEED == ret), "H5Fclose");
+
+    /* Verify that the file is an HDF5 file */
+    is_hdf5 = H5Fis_accessible(filename, fapl_id);
+    VRFY((TRUE == is_hdf5), "H5Fis_accessible");
+
+    /* Delete the file */
+    ret = H5Fdelete(filename, fapl_id);
```
qkoziol, line 989: MPI_File_delete is an independent operation, so only one rank should call it, otherwise one rank will succeed and all the others fail.

You will need an MPI_Barrier before & after deleting the file, so that some ranks don't race ahead / lag behind and erroneously report errors from H5Fis_accessible before / after the H5Fdelete call.

qkoziol (changes requested):
A couple of minor issues that should be resolved, but generally good

byrnHDF (approved):

testpar/t_file.c
derobins: Fixed

epourmal:
@derobins , did you have a chance to look at and address Quincey's comments? Let's move on merges.

src/H5FDcore.c
derobins: Fixed, though I think the behavior is awkward regardless.

src/H5FDmpio.c
```diff
@@ -1745,6 +1747,61 @@ H5FD__mpio_truncate(H5FD_t *_file, hid_t H5_ATTR_UNUSED dxpl_id, hbool_t H5_ATTR
...
+
+    FUNC_ENTER_STATIC
+
+    HDassert(filename);
+
+    if (NULL == (plist = H5P_object_verify(fapl_id, H5P_FILE_ACCESS)))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a file access property list")
+
+    /* Get the MPI communicator and info from the fapl */
+    if (H5P_FILE_ACCESS_DEFAULT != fapl_id && H5FD_MPIO == H5P_peek_driver(plist)) {
```
qkoziol: This code is "fence-sitting" - Either we “know” that the FAPL is non-default (because we arrived in the callback) or we need to protect the callback for the situation where the FAPL doesn’t have the MPI-IO VFD as the value from peek_driver.  Right now, the code looks like you could get to the callback with the default FAPL and then (if that theoretically occurred) you could call MPI_Comm_rank with MPI_COMM_NULL and have a very weird value for the mpi_rank variable.  I would just take out the check for a non-default FAPL and assert that you have the H5FD_MPIO driver, then you would be “safe” to assume a valid MPI communicator and a valid mpi_rank.

src/H5FDmulti.c
qkoziol: Still concerned about this - it's basically a potential buffer overrun.

qkoziol:
There's a small issue in the MPI-IO VFD that I've messaged Dana about (and left a comment), and a potential buffer overflow in the multi VFD that should really be addressed.

qkoziol:
Good progress - just the MPI-IO VFD issue left?

src/H5FDmulti.c
derobins: Should be improved now

derobins:
> Good progress - just the MPI-IO VFD issue left?

Honestly, we should be replacing all instances of sprintf with snprintf and return value checks. I've added that to my long-term queue, but doing that globally (even in the multi VFD) is out of scope for this PR.

test/h5test.c
```diff
@@ -190,65 +190,20 @@ h5_clean_files(const char *base_name[], hid_t fapl)
...
-            HDsnprintf(sub_filename, sizeof(sub_filename), "%s-%c.h5", filename, multi_letters[mt]);
-            HDremove(sub_filename);
-        }
-    }
-    else {
-        HDremove(filename);
-    } /* end driver selection tree */
+    H5E_BEGIN_TRY {
+        H5Fdelete(filename, fapl);
+    } H5E_END_TRY;
```
derobins: This ensures that H5Fdelete() is at least tested for all serial VFDs that we support via make check-vfd.

src/H5FDmpio.c
derobins: I'll make this change.

qkoziol (approved):
Looks great now!

PR #526, opened 2021-03-27
Minor warning fixes in develop

derobins:


src/H5FDlog.c
```diff
@@ -502,6 +502,10 @@ H5FD__log_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxaddr)
     if (ADDR_OVERFLOW(maxaddr))
         HGOTO_ERROR(H5E_ARGS, H5E_OVERFLOW, NULL, "bogus maxaddr")
 
+    /* Initialize timers */
+    H5_timer_init(&open_timer);
+    H5_timer_init(&stat_timer);
+
```
derobins, line 508: Incidentally, this is one of those places where it would make sense to intermingle declarations and statements as we could declare the timer and then immediately initialize it.

derobins:
I also removed a lot of nearby trailing /* end if */, etc. statements as they are not only noise that obscures code, but now look ridiculous when they get misaligned by the formatter.

src/H5FDlog.c
```diff
@@ -484,8 +484,8 @@ H5FD__log_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxaddr)
 #ifdef H5_HAVE_WIN32_API
     struct _BY_HANDLE_FILE_INFORMATION fileinfo;
 #endif
-    H5_timer_t open_timer = {{0}, {0}, {0}, FALSE}; /* Timer for open() call */
-    H5_timer_t stat_timer = {{0}, {0}, {0}, FALSE}; /* Timer for stat() call */
+    H5_timer_t open_timer; /* Timer for open() call */
+    H5_timer_t stat_timer; /* Timer for stat() call */
```
derobins, line 488: The changes look noisy, but the important changes just get rid of the initialization and move the call to the init function up (the warning is that the {0}s in the initialization don't fully initialize the struct).

test/cache_common.c
```diff
@@ -2757,7 +2757,6 @@ flush_cache(H5F_t *file_ptr, hbool_t destroy_entries, hbool_t dump_stats, hbool_
 
     if (pass) {
         H5C_t *cache_ptr;
-        herr_t result = 0;
```
derobins, line 2760: Fixes an unused variable warning.

test/dtypes.c
```diff
@@ -1485,7 +1485,7 @@ test_compound_8(void)
     struct s2 {
         char c;
         s1   d;
-    } s2;
+    };
     hid_t   tid1, tid1_copy, tid2, tid2_copy, tid3, arr_tid;
```
derobins, line 1489: Fixes an unused variable warning.

qkoziol (approved):

lrknox (approved):

jhendersonHDF (approved):

byrnHDF (approved):

PR #527, opened 2021-03-27
Removes implementation of my_strdup() from the multi VFD

derobins:
strdup() has been in POSIX since 2001. I think it's safe to assume it's present at this point. The only thing that's needed is redefining it for Windows.


src/H5FDmulti.c
```diff
@@ -35,6 +35,15 @@
...
 
+/* Windows doesn't like some POSIX names and redefines them with an
+ * underscore
+ */
+#ifdef _WIN32
+#define my_strdup _strdup
+#else
+#define my_strdup strdup
+#endif
+
```
derobins, line 46: We can't just define strdup as _strdup as the former also exists on Windows. The problem with just using strdup directly is that it's deprecated and raises warnings.

qkoziol: strdup isn't deprecated, it wasn't part of the C Standard (until the upcoming C23:  https://en.cppreference.com/w/c/experimental/dynamic/strdup).  We can't rely on it being available on all platforms yet.

derobins: The non-underscore name is deprecated on Windows. From [MSDN](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=msvc-160):

The Microsoft-implemented POSIX function names strdup and wcsdup are deprecated aliases for the _strdup and _wcsdup functions. By default, they generate Compiler warning (level 3) C4996. The names are deprecated because they don't follow the Standard C rules for implementation-specific names. However, the functions are still supported.

We recommend you use _strdup and _wcsdup instead. Or, you can continue to use these function names, and disable the warning. For more information, see Turn off the warning and POSIX function names.

qkoziol: > The non-underscore name is deprecated on Windows. From [MSDN](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=msvc-160):
> 
> The Microsoft-implemented POSIX function names strdup and wcsdup are deprecated aliases for the _strdup and _wcsdup functions. By default, they generate Compiler warning (level 3) C4996. The names are deprecated because they don't follow the Standard C rules for implementation-specific names. However, the functions are still supported.
> 
> We recommend you use _strdup and _wcsdup instead. Or, you can continue to use these function names, and disable the warning. For more information, see Turn off the warning and POSIX function names.

Ah, sorry!   I misinterpreted your comment and thought it was about strdup itself.  My comment was toward that direction:  strdup isn't (yet) part of the C Standard, so this VFD may need the my_strdup() code for a while longer.

derobins: Also, the library doesn't really have an option if you don't have strdup. Either you have POSIX strdup or Windows _strdup or you are out of luck. In the "out of luck" case, we declare a strdup function exists, but we don't provide an implementation. The code in H5private.h says strdup is non-POSIX, but that's probably pre-2001 code.

jhendersonHDF (approved):

lrknox (approved):

PR #528, opened 2021-03-27
Removes dead H5ST package from the library

derobins:
This has been dead code for over a decade and has no obvious immediate use. There is no need for it to be in the library.


qkoziol:
I still don't think this is a good idea.  It's working code and if someone needs a capability like this, it's good to have a well-tested tool in the toolbox.  I believe that I implemented this TST code to replace skip lists for key lookups in property lists, but didn't have time to finish the conversion.

lrknox (approved):

byrnHDF (approved):

gnuoyd (approved):
LGTM.

soumagne (approved):

PR #529, opened 2021-03-30
Add H5FD_read_selection() and H5FD_write_selection()

fortnern:
Currently only translate to scalar calls.  Also fix const buf in H5FD_write_vector().

### DIFF
--- a/src/H5FD.c
+++ b/src/H5FD.c
@@ -1467,7 +1467,7 @@ H5FDread_vector(H5FD_t *file, hid_t dxpl_id, uint32_t count, H5FD_mem_t types[],
     herr_t ret_value = SUCCEED; /* Return value             */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE7("e", "*xiIu*Mt*a*zx", file, dxpl_id, count, types, addrs, sizes, bufs);
+    H5TRACE7("e", "*#iIu*Mt*a*zx", file, dxpl_id, count, types, addrs, sizes, bufs);
 
     /* Check arguments */
     if (!file)
@@ -1543,12 +1543,12 @@ done:
  */
 herr_t
 H5FDwrite_vector(H5FD_t *file, hid_t dxpl_id, uint32_t count, H5FD_mem_t types[], haddr_t addrs[],
-                 size_t sizes[], void *bufs[] /* in */)
+                 size_t sizes[], const void *bufs[] /* in */)
 {
     herr_t ret_value = SUCCEED; /* Return value             */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE7("e", "*xiIu*Mt*a*z**x", file, dxpl_id, count, types, addrs, sizes, bufs);
+    H5TRACE7("e", "*#iIu*Mt*a*z**x", file, dxpl_id, count, types, addrs, sizes, bufs);
 
     /* Check arguments */
     if (!file)
--- a/src/H5FDint.c
+++ b/src/H5FDint.c
@@ -40,6 +40,9 @@
 /* Local Macros */
 /****************/
 
+/* Length of sequence lists requested from dataspace selections */
+#define H5FD_SEQ_LIST_LEN 128
+
 /******************/
 /* Local Typedefs */
 /******************/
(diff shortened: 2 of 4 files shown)


fortnern:
I should add that this is currently untested.

src/H5FDint.c
```diff
@@ -599,7 +602,517 @@ H5FD_write_vector(H5FD_t *file, uint32_t count, H5FD_mem_t types[], haddr_t addr
...
+        size_t         file_len[H5FD_SEQ_LIST_LEN];
+        hsize_t        mem_off[H5FD_SEQ_LIST_LEN];
+        size_t         mem_len[H5FD_SEQ_LIST_LEN];
+        size_t         file_seq_i;
+        size_t         mem_seq_i;
+        size_t         file_nseq;
+        size_t         mem_nseq;
+        size_t         io_len;
+        size_t         dummy_nelem;
+        H5S_sel_iter_t file_iter;
```
fortnern, line 744: Getting warnings about stack size - should these be dynamically allocated?

jrmainzer: Allocating these variables on the stack seems perfectly reasonable to me -- the last time I ran into trouble with this was pre-MacOS X, where stack size was limited to 32 K by default.  That said, some current compilers seem to have a thing against this.

Given the extra overhead and potential for error in allocating these arrays dynamically, I would go with what you have for now.  We can always go back and change it later.

src/H5FDint.c
```diff
@@ -599,7 +602,517 @@ H5FD_write_vector(H5FD_t *file, uint32_t count, H5FD_mem_t types[], haddr_t addr
...
+        /* otherwise, implement the selection read as a sequence of regular
+         * read calls.
+         */
+        extend_sizes = FALSE;
+
+        for (i = 0; i < count; i++) {
+
+            /* we have already verified that sizes[0] != 0 and
+             * types[0] != H5FD_MEM_NOLIST
+             */
```
jrmainzer: I see you are extending the sizes[] / types[] optimization to selection I/O proper.

Given that selections are a much heavier weight data structure, is this optimization significant?  To put it another way, is it buying us enough in terms of reduced footprint / ease of use to make up for the extra overhead?

Your call on this, but if you decide to keep it, we need to update the RFC accordingly, and also update the above comment to refer to just the element_sizes[] array.

src/H5FD.c
```diff
@@ -1467,7 +1467,7 @@ H5FDread_vector(H5FD_t *file, hid_t dxpl_id, uint32_t count, H5FD_mem_t types[],
     herr_t ret_value = SUCCEED; /* Return value             */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE7("e", "*xiIu*Mt*a*zx", file, dxpl_id, count, types, addrs, sizes, bufs);
+    H5TRACE7("e", "*#iIu*Mt*a*zx", file, dxpl_id, count, types, addrs, sizes, bufs);
```
jrmainzer, line 1470: Thanks for the touch ups

jrmainzer:
Hi Neil,

No significant issues -- you seem to be going in the right direction.  Let me know if you want to chat.

src/H5FDint.c
fortnern: I view it as mostly a convenience item since in most cases the element sizes will be the same. I don't think it adds much overhead (it should be faster if all the element sizes are the same) and I think it makes sense to use a similar interface to that for vector I/O. There's currently no optimization for types since there's only one type in selection I/O, I will fix that comment. I am still unsure about not having a types array though (again, I think it should be similar to the vector I/O interface).

jrmainzer: The overhead I was referring to was added development complexity.  If nothing else, the optimization makes the vector sort routine a bit more painful.  Similarly it makes adding support for vector I/O a little harder in the VFD proper.  In the case of vector I/O representation of point selections, there can be a significant reduction in memory foot print which justifies the optimization in my mind -- I'm not so sure about selection I/O.

That said, you make a reasonable case.  If you are happy with it, I will add it to the RFC.

As to selection I/O not having a types vector -- if you can think of a plausible use case, I'll put it in the RFC.

PR #543, opened 2021-04-06
DO NOT MERGE: Example of using tagged unions for VOL

jhendersonHDF:
This PR is intended to spark discussion around using tagged unions instead of var. args for all non-optional VOL infrastructure. My reasoning is:

- All non-optional VOL callback functions that currently accept a va_list are already accompanied by a "tag" for a union, e.g. "link get" has H5VL_link_get_t
- All of the non-optional VOL callbacks currently use a fixed, small amount of known arguments, which are just as easy to pass with a union. They are also just as easy to unpack and can sometimes be easier.
- Using a union eliminates the need for passthrough connectors to deal with var. args completely, except for the case of "optional" operations, where I think it makes sense to keep them, as the arguments to be passed will be arbitrary and generally outside of THG's control. If passthrough connectors don't need to modify any arguments, they will be able to directly pass them through to the reissuing call. If they do need to modify them, they can simply take a copy of the arguments with a struct assignment to a local struct, modify them, and pass the local struct through to the same reissuing call.
- Many HDF5 operations can still be added to these unions without making them unreadably large; each operation's arguments can be put in a separate struct that gets included in the union.
- Might be more peformant than variadic functions - only loosely based on https://blog.nelhage.com/2010/10/amd64-and-va_arg/ and other research

### DIFF
--- a/src/H5VLcallback.c
+++ b/src/H5VLcallback.c
@@ -133,9 +133,9 @@ static herr_t H5VL__group_specific(void *obj, const H5VL_class_t *cls, H5VL_grou
 static herr_t H5VL__group_optional(void *obj, const H5VL_class_t *cls, H5VL_group_optional_t opt_type,
                                    hid_t dxpl_id, void **req, va_list arguments);
 static herr_t H5VL__group_close(void *obj, const H5VL_class_t *cls, hid_t dxpl_id, void **req);
-static herr_t H5VL__link_create(H5VL_link_create_type_t create_type, void *obj,
-                                const H5VL_loc_params_t *loc_params, const H5VL_class_t *cls, hid_t lcpl_id,
-                                hid_t lapl_id, hid_t dxpl_id, void **req, va_list arguments);
+static herr_t H5VL__link_create(H5VL_link_create_type_t create_type, const H5VL_link_create_args_t *create_args,
+                                void *obj, const H5VL_loc_params_t *loc_params, const H5VL_class_t *cls,
+                                hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req);
 static herr_t H5VL__link_copy(void *src_obj, const H5VL_loc_params_t *loc_params1, void *dst_obj,
                               const H5VL_loc_params_t *loc_params2, const H5VL_class_t *cls, hid_t lcpl_id,
                               hid_t lapl_id, hid_t dxpl_id, void **req);
@@ -4672,9 +4672,9 @@ done:
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5VL__link_create(H5VL_link_create_type_t create_type, void *obj, const H5VL_loc_params_t *loc_params,
-                  const H5VL_class_t *cls, hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req,
-                  va_list arguments)
+H5VL__link_create(H5VL_link_create_type_t create_type, const H5VL_link_create_args_t *create_args,
+                  void *obj, const H5VL_loc_params_t *loc_params, const H5VL_class_t *cls,
+                  hid_t lcpl_id, hid_t lapl_id, hid_t dxpl_id, void **req)
 {
     herr_t ret_value = SUCCEED; /* Return value */
 
@@ -4685,7 +4685,7 @@ H5VL__link_create(H5VL_link_create_type_t create_type, void *obj, const H5VL_loc
         HGOTO_ERROR(H5E_VOL, H5E_UNSUPPORTED, FAIL, "VOL connector has no 'link create' method")
 
     /* Call the corresponding VOL callback */
-    if ((cls->link_cls.create)(create_type, obj, loc_params, lcpl_id, lapl_id, dxpl_id, req, arguments) < 0)
+    if ((cls->link_cls.create)(create_type, create_args, obj, loc_params, lcpl_id, lapl_id, dxpl_id, req) < 0)
         HGOTO_ERROR(H5E_VOL, H5E_CANTCREATE, FAIL, "link create failed")
 
 done:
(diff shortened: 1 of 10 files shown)


src/H5VLconnector.h
```diff
@@ -46,6 +46,50 @@
...
+ */
+typedef struct H5VL_loc_params_t {
+    H5I_type_t      obj_type;
+    H5VL_loc_type_t type;
+    union {
+        H5VL_loc_by_token_t loc_by_token;
+        H5VL_loc_by_name_t  loc_by_name;
+        H5VL_loc_by_idx_t   loc_by_idx;
+    } loc_data;
+} H5VL_loc_params_t;
```
jhendersonHDF, line 91: Just noise - the structs above needed to be moved up before the definition of the below H5VL_link_create_args_t.

src/H5VLcallback.c
```diff
@@ -4703,30 +4704,20 @@ H5VL__link_create(H5VL_link_create_type_t create_type, void *obj, const H5VL_loc
...
-    /* Start the varargs, so they can be copied */
-    HDva_start(arguments, req);
-    arg_started = TRUE;
-
     /* Special case for hard links */
     if (H5VL_LINK_CREATE_HARD == create_type && NULL == vol_obj->data) {
-        va_list tmp_arguments; /* Copy of argument list passed in */
-
         /* Get the VOL data pointer from the varargs */
-        HDva_copy(tmp_arguments, arguments);
```
jhendersonHDF, line 4727: An example of a case where avoiding var args can make things easier to unpack, as we don't need to worry about situations like using va_copy to avoid interfering with the va_start above.

src/H5VLpassthru.c
```diff
@@ -2209,31 +2181,31 @@ H5VL_pass_through_link_create(H5VL_link_create_type_t create_type, void *obj,
...
 
         /* Retrieve the object & loc params for the link target */
-        cur_obj    = va_arg(arguments, void *);
-        cur_params = va_arg(arguments, H5VL_loc_params_t *);
+        cur_obj = create_args->hard.vol_obj_data;
 
         /* If it's a non-NULL pointer, find the 'under object' and re-set the property */
         if (cur_obj) {
+            /* Copy the current link creation arguments */
+            new_create_args = *create_args;
```
jhendersonHDF, line 2192: Since the VOL object needs to be modified here, we just take a copy of the const args and pass the copy down. The var arg logic is mostly just substituted with logic to determine which of the arguments we need to pass down (the original or the copy), but it removes the need for the _reissue function and makes the reissue operation look consistent with other callbacks.

qkoziol:
> This PR is intended to spark discussion around using tagged unions instead of var. args for all non-optional VOL infrastructure. My reasoning is:
> 
> * All non-optional VOL callback functions that currently accept a va_list are already accompanied by a "tag" for a union, e.g. "link get" has H5VL_link_get_t
> * All of the non-optional VOL callbacks currently use a fixed, small amount of known arguments, which are just as easy to pass with a union. They are also just as easy to unpack and can sometimes be easier.
> * Using a union eliminates the need for passthrough connectors to deal with var. args completely, except for the case of "optional" operations, where I think it makes sense to keep them, as the arguments to be passed will be arbitrary and generally outside of THG's control. If passthrough connectors don't need to modify any arguments, they will be able to directly pass them through to the reissuing call. If they do need to modify them, they can simply take a copy of the arguments with a struct assignment to a local struct, modify them, and pass the local struct through to the same reissuing call.
> * Many HDF5 operations can still be added to these unions without making them unreadably large; each operation's arguments can be put in a separate struct that gets included in the union.
> * Might be more peformant than variadic functions - only loosely based on https://blog.nelhage.com/2010/10/amd64-and-va_arg/ and other research

Yes - definitely like the direction!  :-)  Since the parameters are well-defined, just overloaded, having a "pseudo-variadic" technique like this would definitely give better type safety and possibly a tiny performance boost.

Two suggestions:  
- I would move to a struct->union->struct pattern, with the 'type' parameter as the first field in the outer struct. (Similar to the H5O_link_t struct and others within the library)   That will bind the type to the operation parameters and should help reduce the odds of the type being updated out of sync with the parameters.
- As you say, it would be very nice to use something like this for the 'optional' callbacks, but since the parameters aren't pre-defined, we would need a different technique.  I would suggest switching from a va_list / varargs ("...") parameter to a void* that pointed to a struct that contained the parameters for the optional operation.   It's still flexible, but avoids the difficulties around using variable arguments.   We can even pre-defined a struct->union for all the native optional operations to pass as the void*, within the library.

@jhendersonHDF - Thoughts about those two refinements?

jhendersonHDF:
> I would move to a struct->union->struct pattern, with the 'type' parameter as the first field in the outer struct. (Similar to the H5O_link_t struct and others within the library) That will bind the type to the operation parameters and should help reduce the odds of the type being updated out of sync with the parameters.

I agree. If we do switch to using tagged unions, then it definitely makes the most sense to follow a traditional pattern of having the tag actually inside the structure containing the union. For getting a quick example up, I just left it as part of the function parameters, but I'll update the PR soon with this change so that the difference can be seen.

> As you say, it would be very nice to use something like this for the 'optional' callbacks, but since the parameters aren't pre-defined, we would need a different technique. I would suggest switching from a va_list / varargs ("...") parameter to a void* that pointed to a struct that contained the parameters for the optional operation. It's still flexible, but avoids the difficulties around using variable arguments. We can even pre-defined a struct->union for all the native optional operations to pass as the void*, within the library.

I'm certainly not against using a simple void * to a struct, but I'm not sure I understand why var args wouldn't work just the same in this case; could you explain more on that? If I'm not mistaken, it seems to me that the public "optional" APIs, like H5VLdatatype_optional, can be declared as taking var args (...), with the internal functions continuing to take a va_list like they do now. Then, the public APIs would just be responsible for constructing the va_list and passing it down.

qkoziol:
> > I would move to a struct->union->struct pattern, with the 'type' parameter as the first field in the outer struct. (Similar to the H5O_link_t struct and others within the library) That will bind the type to the operation parameters and should help reduce the odds of the type being updated out of sync with the parameters.
> 
> I agree. If we do switch to using tagged unions, then it definitely makes the most sense to follow a traditional pattern of having the tag actually inside the structure containing the union. For getting a quick example up, I just left it as part of the function parameters, but I'll update the PR soon with this change so that the difference can be seen.

OK, cool, thanks, we can discuss this at next Monday's meeting.

qkoziol:
> > As you say, it would be very nice to use something like this for the 'optional' callbacks, but since the parameters aren't pre-defined, we would need a different technique. I would suggest switching from a va_list / varargs ("...") parameter to a void* that pointed to a struct that contained the parameters for the optional operation. It's still flexible, but avoids the difficulties around using variable arguments. We can even pre-defined a struct->union for all the native optional operations to pass as the void*, within the library.
> 
> I'm certainly not against using a simple void * to a struct, but I'm not sure I understand why var args wouldn't work just the same in this case; could you explain more on that? If I'm not mistaken, it seems to me that the public "optional" APIs, like H5VLdatatype_optional, can be declared as taking var args (...), with the internal functions continuing to take a va_list like they do now. Then, the public APIs would just be responsible for constructing the va_list and passing it down.

I'm thinking that it would be smart to just eliminate the variable arguments entirely.   Switching to a void* for the arguments lets us do that, but it also means that the 'type' parameter may be better if it stays separate.   We could go to a struct like this though, I suppose:
typedef struct {
    H5VL_link_optional_t operation_type;
    void *argument_struct;
} H5VL_link_optional_t;

I suppose that's not terrible, and is mostly in the style of the other change you've outlined for the non-optional callbacks.   Thoughts?

jhendersonHDF:
> I'm thinking that it would be smart to just eliminate the variable arguments entirely. Switching to a void* for the arguments lets us do that, but it also means that the 'type' parameter may be better if it stays separate. We could go to a struct like this though, I suppose:
> typedef struct {
> H5VL_link_optional_t operation_type;
> void *argument_struct;
> } H5VL_link_optional_t;
> 
> I suppose that's not terrible, and is mostly in the style of the other change you've outlined for the non-optional callbacks. Thoughts?

Sure, I think that's just as valid of a way to go as opposed to var args, especially for keeping the interface consistent. Var args just seemed to make sense to keep for the optional interface since they are an arbitrary interface to a connector.

PR #547, opened 2021-04-07
Add new test: vfd_swmr_attrdset_writer.c

vchoi-hdfgroup:


test/vfd_swmr_attrdset_writer.c
```diff
@@ -0,0 +1,2017 @@
...
+ *      -- Delete attribute
+ *      -- Modify attribute
+ *      -- Add variable length attribute
+ *      -- Delete variable length attribute
+ *      -- Modify variable length attribute
+ *      -- Add sufficient attributes to force creation of object header continuation block
+ *      -- Remove sufficient attributes to allow deletion of object header continuation block
+ *      -- Transition from compact to dense attribute storage
+ *      -- Transition from dense to compact attribute storage
+ */
```
raylu-hdf, line 40: I don't think the three headers below are necessary.

raylu-hdf (commented):
Dana used the `H5_HAVE_WIN32_API` to skip the VFD SWMR tests on Windows, e.g. in vfd_swmr_group_writer.c.  That's a recent change.  I think you may do the same for this new test.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: Removed as indicated.

vchoi-hdfgroup: [Dana used the H5_HAVE_WIN32_API to skip the VFD SWMR tests on Windows, e.g. in vfd_swmr_group_writer.c. That's a recent change. I think you may do the same for this new test.]
Add that in as indicated.

test/vfd_swmr_attrdset_writer.c
```diff
@@ -0,0 +1,2015 @@
...
+        TEST_ERROR;
+    }
+
+    if((is_vl = H5Tis_variable_str(atid))) {
+        if(action == ADD_ATTR)
+            HDsprintf(vl_which, "%u", which);
+        else
+            HDsprintf(vl_which, "%u %u %u %u %u", which, which+1, which+2, which+3, which+4);
+
+        read_vl_which = HDmalloc(sizeof("9999999999"));
```
raylu-hdf: This allocated memory needs to be freed.

test/vfd_swmr_attrdset_writer.c
```diff
@@ -0,0 +1,2015 @@
...
+
+/* Action for attribute handling */
+#define ADD_ATTR        1
+#define MODIFY_ATTR     2
+#define DELETE_ATTR     3
+
+/* This will be removed when this routine is moved to vfd_swmr_common.c */
+/* Sleep for `tenths` tenths of a second */
+static void
+decisleep(uint32_t tenths)
```
raylu-hdf: I put the definition of this function in `vfd_swmr_common.c` in my pull request for the zoo test.  I already merged my PR.

test/vfd_swmr_attrdset_writer.c
kyang2014: I will check the current vfd_swmr_group_writer.c. If it is not there, I will add it in.

test/vfd_swmr_attrdset_writer.c
```diff
@@ -0,0 +1,2015 @@
...
+		"-b:	           write data in big-endian byte order if no -v option\n"
+        "                   (default is H5T_NATIVE_UINT32)\n\n"
+        "Note:\n"
+        "1. Require to specify at least -p, -g or -k option\n"
+        "2. -c <csteps> option cannot exceed -a <nattrs> option\n"
+        "3. -d <dattrs> option cannot exceed -a <nattrs> option\n"
+		"\n",
+		progname);
+	exit(EXIT_FAILURE);
+} /* usgae() */
```
kyang2014: typo: should be usage()

test/vfd_swmr_attrdset_writer.c
```diff
@@ -0,0 +1,2015 @@
...
+            printf("H5Pcreate failed\n");
+            TEST_ERROR;
+        }
+
+        if(H5Pset_layout(dcpl, H5D_CONTIGUOUS) < 0) {
+            printf("H5Pset_layout failed\n");
+            TEST_ERROR;
+        }
+
+        if(H5Pset_alloc_time(dcpl, H5D_ALLOC_TIME_EARLY) < 0) {
```
kyang2014, line 462: I think the early allocation is chosen for the testing purpose. Should we have a test for the default case, which I believe is more common. Or it doesn't matter. Note this is a question rather than a request to add such a test.

test/vfd_swmr_attrdset_writer.c
```diff
@@ -0,0 +1,2015 @@
...
+    if(nerrors)
+        ret = false;
+
+    return (ret);
+
+} /* attr_dsets_action() */
+
+/*
+ * Perform the attribute action on the specified dataset.
+ */
```
kyang2014, line 971: Maybe add description for this function. The action, s, and did are straightforward. But the "step" may not be obvious.

test/vfd_swmr_attrdset_writer.c
kyang2014: Also, in the group test, to verify the vl str attribute,  a HDfree_memory() is called after the attribute value is read and  is compared. I copied it from static void test_write_vl_string_attribute(void) in tvlstr.c under /test.  However, the group test only handles one VL string which is like test_write_vl_string_attribute(). The VL in this test has several elements.

kyang2014 (approved):
May add more comments to explain the test procedure. Especially for the followings:
1)  -- Add sufficient attributes to force creation of object header continuation block
 *      -- Remove sufficient attributes to allow deletion of object header continuation block
How this is verified. Perhaps this is just to me.
2) If I understand correctly, the program doesn't limit the number of attributes to add with the parameter asteps. This is a good robust test to test the addition of many attributes for one dataset.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: My thought is: the zoo test already tested contiguous dataset with default allocation, so I use early allocation for the contiguous dataset here.

vchoi-hdfgroup: [Add sufficient attributes to force creation of object header continuation block]
[Remove sufficient attributes to allow deletion of object header continuation block]

For the verification of object header continuation block, please see verify_storage_cont_real() where ninfo.hdr.nchunks is 1 and > 1 as attributes are added and storage is moving between compact and dense.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: Corrected.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: Corrected.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: Corrected.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: Added.

kyang2014: OK, good and make sense.

raylu-hdf (approved):

PR #567, opened 2021-04-21
Selection io

jrmainzer:
First cut at vector read / write support for the MPIO VFD.  This should be good enough for Jordan to proceed with conversion of parallel compression to vector I/O.

Note that this implementation doesn't support large (> 2 GB) entries in the vectors.  Considering vector I/O only, the issue is probably best solved by splitting large entries into < 2 GB entries, which is easily done.  However, I'd like to see how selection I/O proper comes out before I choose this route.  

Also fixed a sanity checking error in the H5FD versions of vector read / write in passing -- was not handling the size[] and type[] vector optimization correctly.

### DIFF
--- a/src/H5FDint.c
+++ b/src/H5FDint.c
@@ -1540,7 +1540,7 @@ H5FD_sort_vector_io_req(hbool_t *vector_was_sorted, uint32_t count, H5FD_mem_t t
             (*s_types_ptr)[j] = types[MIN(i, fixed_type_index)];
             (*s_addrs_ptr)[j] = addrs[i];
             (*s_sizes_ptr)[j] = sizes[MIN(i, fixed_size_index)];
-            (*s_bufs_ptr)[j] = bufs[i];
+            (*s_bufs_ptr)[j]  = bufs[i];
         }
     }
 
--- a/src/H5FDint.c
+++ b/src/H5FDint.c
@@ -1487,7 +1487,7 @@ H5FD_sort_vector_io_req(hbool_t *vector_was_sorted, uint32_t count, H5FD_mem_t t
         /* verify no duplicate entries */
         i = 1;
 
-        while ((*vector_was_sorted) && (i < (int)(count - 1))) {
+        while (i < (int)(count - 1)) {
 
             HDassert(H5F_addr_lt(srt_tmp[i - 1].addr, srt_tmp[i].addr));
 
--- a/src/H5FDint.c
+++ b/src/H5FDint.c
@@ -44,6 +44,33 @@
 /* Local Typedefs */
 /******************/
 
+/*************************************************************************
+ *
+ * H5FD_vsrt_tmp_t
+ *
+ * Structure used to store vector I/O request addresses and the associated
+ * indexes in the addrs[] array for the purpose of determine the sorted
+ * order.
+ *
+ * This is done by allocating an array of H5FD_vsrt_tmp_t of length
+ * count, loading it with the contents of the addrs[] array and the
+ * associated indicies, and then sorting it.
+ *
+ * This sorted array of H5FD_vsrt_tmp_t is then used to populate sorted
+ * versions of the types[], addrs[], sizes[] and bufs[] vectors.
+ *
+ * addr:        haddr_t containing the value of addrs[i],
+ *
+ * index:       integer containing the value of i used to obtain the
+ *              value of the addr field from the addrs[] vector.
+ *
+ *************************************************************************/
+
+typedef struct H5FD_vsrt_tmp_t {
+    haddr_t addr;
+    int     index;
+} H5FD_vsrt_tmp_t;
+
 /********************/
 /* Package Typedefs */
 /********************/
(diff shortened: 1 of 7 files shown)


src/H5FDmpio.c
```diff
@@ -1520,6 +1524,834 @@ H5FD__mpio_write(H5FD_t *_file, H5FD_mem_t type, hid_t H5_ATTR_UNUSED dxpl_id, h
...
+ *
+ *              Finally, set the file view back to its default state.
+ *
+ *              In contrast, if io_xfer_mode is H5FD_MPIO_INDEPENDENT,
+ *              this call is independent, and thus we cannot use
+ *              MPI_File_set_view().
+ *
+ *              In this case, simply walk the vector, and issue an
+ *              independent read for each entry.
+ *
```
jrmainzer, line 1555: Is the above description of the io_xfer_mode and the mpio_coll_opt DXPL entries correct?

jhendersonHDF: Based on the documentation for the related H5P routines, this sounds accurate to me. Particularly due to this note from H5Pset_dxpl_mpio_collective_opt:

> This function should be used only when H5FD_MPIO_COLLECTIVE has been set through H5Pset_dxpl_mpio. In such situations, normal behavior would be to use low-level collective I/O functions, but the library will use low-level MPI independent I/O functions when H5FD_MPIO_INDIVIDUAL_IO is set.

So it seems like H5Pset_dxpl_mpio for the io_xfer_mode is for specifying collectivity at the application level, while H5Pset_dxpl_mpio_collective_opt is available to ask HDF5 to actually use independent I/O regardless of the application level collectivity setting.

src/H5FDint.c
```diff
@@ -1248,3 +1333,258 @@ H5FD_driver_query(const H5FD_class_t *driver, unsigned long *flags /*out*/)
...
+            srt_tmp[i].index = i;
+        }
+
+        /* sort the srt_tmp array */
+        HDqsort(srt_tmp, (size_t)count, sizeof(struct H5FD_vsrt_tmp_t), H5FD__vsrt_tmp_cmp);
+
+        /* verify no duplicate entries */
+        i = 1;
+
+        while ((*vector_was_sorted) && (i < (int)(count - 1))) {
```
jhendersonHDF: Won't *vector_was_sorted always be false here?

jrmainzer: You are correct --  mindless cut and paste.  Will fix.

jhendersonHDF: I also wonder if we need to be doing the work of scanning the array once again while looking for duplicates. I could be wrong, but I think MPI I/O doesn't care much about duplicates and you'd only get in trouble if you were trying to write different data to the same offset.

Of course, it may not quite be correct to just omit checking for duplicates, but we also might be erroring on somebody's I/O call that may not be problematic. Thoughts on that?

jrmainzer: My impression is that MPI does not like overlapping entries in the derived types passed to MPI_File_set_view -- at least I triggered a number of failures by accident that seemed to indicate this. 

Even if MPI is OK with overlapping entries, wouldn't it be an error if we write the same location twice in a single I/O request?  If not, your argument against this check may be sound -- but I can't think of an example.

testpar/t_vfd.c
```diff
@@ -0,0 +1,4057 @@
...
+#if 1
+            for (j = base_index; j < base_index + INTS_PER_RANK; j++) {
+
+                k = j - base_index;
+#else
+            for (k = 0; k < INTS_PER_RANK; k++) {
+
+                j = k + base_index;
+#endif
+
```
jrmainzer, line 1713: The else case above is a code around an odd warning that I got in production, but not debug , builds.  Either one works just fine, although I find the if case a bit easier to read.

Absent strong objections, I'm going to leave this code here until I get a chance to discuss the warning message with Larry.

src/H5FDint.c
jhendersonHDF: This very well may be the case, though MPI's "monotonically non-decreasing" requirement seems to hint that this may be a possibility. I vaguely recall running into this with parallel compression at one point, but that was long enough ago that I don't remember the details. I would think that yes, writing the same location twice wouldn't generally be correct, but as long as the data is the exact same I wouldn't think it would be problematic.

In any case, I was mostly looking at the overall read/write process to see if it's possible to cut down on any work. It seems that, on average, for each I/O request we're likely going to:

- scan through the whole addrs array to check if it's sorted
- allocate and populate the H5FD_vsrt_tmp_t array
- sort the H5FD_vsrt_tmp_t array and scan it for duplicates
- allocate copies of the types, addrs, sizes, bufs arrays
- scan the sizes and types vectors for optimizations

This is probably just a case of premature optimization, but it seemed like a lot of work was going on, so I just wondered if there are places we can cut corners or not.

src/H5FDint.c
```diff
@@ -1248,3 +1333,258 @@ H5FD_driver_query(const H5FD_class_t *driver, unsigned long *flags /*out*/)
...
+
+        HDassert(sizes[0] != 0);
+        HDassert(types[0] != H5FD_MEM_NOLIST);
+
+        /* scan the sizes and types vectors to determine if the fixed size / type
+         * optimization is in use, and if so, to determine the index of the last
+         * valid value on each vector.
+         */
+        i = 0;
+        while ((i < (int)count) && ((fixed_size_index == (int)count) || (fixed_type_index == (int)count))) {
```
fortnern, line 1517: Can we skip this section if the length of types or sizes is 1 (which it will be in most cases)?

fortnern: As in only operate on the arrays with length > 1

jrmainzer: Unless we can think of a use case for the shift to constant lengths or types after several different values, I think you have a point.  Will think on this, and change if I can't come up with a use case.

PR #600, opened 2021-04-26
Hash table replacement for skip lists in ID code

derobins:


src/H5Idbg.c
```diff
@@ -182,7 +185,17 @@ H5I_dump_ids_for_type(H5I_type_t type)
...
-            H5SL_iterate(type_info->ids, H5I__id_dump_cb, &type);
+            /* Normally we care about the callback's return value
+             * (H5I_ITER_CONT, etc.), but this is an iteration over all
+             * the IDs so we don't care.
+             *
+             * XXX: Update this to emit an error message on errors?
+             */
+            HDfprintf(stderr, "     (HASH TABLE)\n");
+            HASH_ITER(hh, type_info->hash_table, item, tmp)
+            {
```
soumagne, line 196: Make sure you add that to clang-format as an iterator, that will fix the { }

src/H5Iint.c
```diff
@@ -328,33 +327,52 @@ H5I_clear_type(H5I_type_t type, hbool_t force, hbool_t app_ref)
...
+    H5I_marking_g = FALSE;
+
+    /* Perform sweep */
+    HASH_ITER(hh, udata.type_info->hash_table, item, tmp)
+    {
+        if (item->marked) {
+            HASH_DELETE(hh, udata.type_info->hash_table, item);
+            item = H5FL_FREE(H5I_id_info_t, item);
+        }
+    }
```
soumagne, line 357: why two iterations here? that doesn't seem clear to me

derobins: One iteration marks, the other deletes. An ID's free callback (called in H5I__mark_mode) can delete other IDs, perturbing the data structure so we just mark items for deletion instead of actually deleting them in the first iteration. We do the actual delete in the second one.

lrknox (approved):

byrnHDF (approved):

PR #602, opened 2021-04-28
VFD plugins

jhendersonHDF:
This PR should be ready to review now. The only piece missing from this feature is the support for configuration strings. The API routines to set a VFD configuration string are included here, but parsing of the strings isn't implemented yet and is being developed in a separate branch.

The main piece that might be controversial here is all of the changes to testing code to deal with avoiding running certain tests with specific VFDs. We already do this in a subset of HDF5's tests, but reimplementing the HDF5_DRIVER environment variable to be interpreted by the library rather than the testing code means that more tests need to be aware of the environment variable. A lot of these checks can go away once HDF5's test suite can be re-engineered to support checking what features a VFD is compatible with, but there are currently a lot of edge cases. For example, the Direct VFD has alignment requirements, which can be problematic for tests that check the resulting file size.


src/H5E.c
```diff
@@ -1532,16 +1532,22 @@ H5Eget_auto2(hid_t estack_id, H5E_auto2_t *func /*out*/, void **client_data /*ou
     H5E_auto_op_t op;                  /* Error stack function */
     herr_t        ret_value = SUCCEED; /* Return value */
 
-    FUNC_ENTER_API(FAIL)
+    /* Don't clear the error stack! :-) */
+    FUNC_ENTER_API_NOCLEAR(FAIL)
```
jhendersonHDF: The changes in this file make it so that H5Eget_auto2/H5Eauto_is_v2 do not clear the error stack. H5Eset_auto2 already avoids clearing the error stack, so it seems logical to do the same for H5Eget_auto2. H5Eauto_is_v2 also seems to make sense since we just want to check the type of error stack.

This was noticed by observing that when an H5E_BEGIN/END_TRY block is used, it would trigger a clear of the error stack and lose anything on the stack before the block, even though the purpose of this block is usually to avoid error printing of something localized that's expected to fail (which we don't want to see).

Since these blocks don't actually prevent those error messages from being put on the error stack, these changes have the side effect that those failure messages will probably now show up if an error stack eventually gets printed out; they will just get printed out later on. However, I still think that this is the "correct" behavior. The only reason error messages weren't printed before was due to the hidden stack clear from H5Eget_auto2/H5Eauto_is_v2 that was also blowing away other important and potentially unrelated error messages on the stack.

If this is a major problem, we can re-engineer H5E_BEGIN/END_TRY to pop the expected error messages off the stack in the same way that I do in H5VLcallback.c in this PR.

src/H5FDint.c
```diff
@@ -425,3 +441,285 @@ H5FD_delete(const char *filename, hid_t fapl_id)
...
+
+        /* Check if plugin value matches VFD class value */
+        if (cls->value == key->vfd.u.value)
+            *success = TRUE;
+    } /* end else */
+#if 0
+    /* Driver is a match, but might not be a compatible version */
+    if (*success && cls->version != H5FD_VERSION)
+        *success = FALSE;
+#endif
```
jhendersonHDF: Currently left in, in lieu of discussion around potential VFD versioning

src/H5FDint.c
```diff
@@ -425,3 +441,285 @@ H5FD_delete(const char *filename, hid_t fapl_id)
...
+        if (NULL == (cls = (const H5FD_class_t *)H5PL_load(H5PL_TYPE_VFD, &key)))
+            HGOTO_ERROR(H5E_VFL, H5E_CANTINIT, H5I_INVALID_HID, "unable to load VFD")
+
+        /* Register the driver we loaded */
+        /* TODO: class size calculation will not be correct with drivers
+         * that wrap the H5FD_class_t structure in their own super-structure,
+         * such as the MPI I/O VFD.
+         */
+        if ((ret_value = H5FD_register(cls, sizeof(*cls), app_ref)) < 0)
+            HGOTO_ERROR(H5E_VFL, H5E_CANTREGISTER, H5I_INVALID_HID, "unable to register VFD ID")
```
jhendersonHDF, line 575: The MPI I/O VFD wraps the H5FD_class_t structure inside its own structure, so it may be difficult to support loading it as a VFD plugin since the size calculation here will only account for the H5FD_class_t structure itself.

src/H5PLplugin_cache.c
```diff
@@ -263,32 +263,89 @@ H5PL__find_plugin_in_cache(const H5PL_search_params_t *search_params, hbool_t *f
 
     /* Loop over all the plugins, looking for one that matches */
     for (u = 0; u < H5PL_num_plugins_g; u++) {
-
-        /* If the plugin type (filter, VOL connector, etc.) and ID match, query the plugin for its info */
-        if ((search_params->type == (H5PL_cache_g[u]).type) &&
-            (search_params->key->id == (H5PL_cache_g[u]).key.id)) {
```
jhendersonHDF: Plugin caching only took filter IDs into account and would probably cause an invalid read for VOL/VFD plugins. This should fix caching for both of those.

src/H5VLcallback.c
```diff
@@ -3424,11 +3429,24 @@ H5VL__file_open_find_connector_cb(H5PL_type_t plugin_type, const void *plugin_in
...
+    if (status < 0) {
+        ssize_t new_num_errors = 0;
+
+        /* Pop any errors generated by the above call */
+        if ((new_num_errors = H5Eget_num(H5E_DEFAULT)) < 0)
+            HGOTO_ERROR(H5E_ERROR, H5E_CANTGET, H5_ITER_ERROR, "can't get current error stack size")
+        if (new_num_errors > num_errors) {
+            new_num_errors -= num_errors;
+            if (H5Epop(H5E_DEFAULT, (size_t)new_num_errors) < 0)
+                HGOTO_ERROR(H5E_ERROR, H5E_CANTRELEASE, H5_ITER_ERROR, "can't sanitize error stack")
```
jhendersonHDF, line 3589: An example of using H5Epop to remove expected error messages generated from the H5E_BEGIN/END_TRY block above so they don't get included in the eventual error stack printing.

src/H5VLint.c
```diff
@@ -1355,12 +1351,12 @@ H5VL__register_connector_by_value(H5VL_class_value_t value, hbool_t app_ref, hid
...
-    op_data.kind     = H5VL_GET_CONNECTOR_BY_VALUE;
-    op_data.u.value  = value;
-    op_data.found_id = H5I_INVALID_HID;
+    op_data.key.kind    = H5VL_GET_CONNECTOR_BY_VALUE;
+    op_data.key.u.value = value;
+    op_data.found_id    = H5I_INVALID_HID;
 
     /* Check if connector is already registered */
-    if (H5I_iterate(H5I_VOL, H5VL__get_connector_cb, &op_data, TRUE) < 0)
+    if (H5I_iterate(H5I_VOL, H5VL__get_connector_cb, &op_data, app_ref) < 0)
```
jhendersonHDF, line 1414: @derobins I believe this is the correct change from looking at the similar H5VL__register_connector_by_name, but if the two are supposed to act differently from each other, I'll change this back.

tools/src/h5dump/h5dump.c
```diff
@@ -1456,40 +1488,10 @@ main(int argc, const char *argv[])
     /* Initialize indexing options */
     h5trav_set_index(sort_by, sort_order);
 
-    if (driver_name_g != NULL) {
```
jhendersonHDF, line 1381: Whole block is duplicated by the call to h5tools_get_fapl below.

tools/src/h5ls/h5ls.c
```diff
@@ -2747,9 +2757,6 @@ main(int argc, const char *argv[])
         else if (!HDstrcmp(argv[argno], "--string")) {
             string_g = TRUE;
         }
-        else if (!HDstrncmp(argv[argno], "--vfd=", (size_t)6)) {
-            preferred_driver = argv[argno] + 6;
-        }
```
jhendersonHDF, line 2752: Old "drivername" parameter is retained, but looped into the new way of loading VFDs in the tools below.

tools/src/h5ls/h5ls.c
```diff
@@ -2959,42 +2985,15 @@ main(int argc, const char *argv[])
...
+    if (custom_vol_fapl || custom_vfd_fapl) {
+        if ((fapl_id = h5tools_get_fapl(H5P_DEFAULT,
+                custom_vol_fapl ? &vol_info : NULL,
+                custom_vfd_fapl ? &vfd_info : NULL)) < 0) {
             error_msg("failed to setup file access property list (fapl) for file\n");
             leave(EXIT_FAILURE);
         }
     }
 
-    if (preferred_driver) {
```
jhendersonHDF, line 2969: Another block duplicated by h5tools_get_fapl.

src/H5PLplugin_cache.c
```diff
@@ -263,59 +263,99 @@ H5PL__find_plugin_in_cache(const H5PL_search_params_t *search_params, hbool_t *f
...
-        if (search_params->type == H5PL_cache_g[u].type) {
-            hbool_t matched = FALSE; /* Whether cached plugin info matches */
+        hbool_t matched = FALSE; /* Whether cached plugin info matches */
 
-            switch (search_params->type) {
-                case H5PL_TYPE_FILTER:
-                    if (search_params->key->id == H5PL_cache_g[u].key.id)
+        /* Determine if the plugin types match */
+        if (search_params->type != H5PL_cache_g[u].type)
+            continue;
```
jhendersonHDF, line 270: The changes in this file fix some caching issues with VOL/VFD plugins.

src/H5Pint.c
```diff
@@ -523,6 +553,28 @@ H5P__init_package(void)
     HDassert(tot_init == NELMTS(init_class));
 
 done:
+    if (ret_value < 0 && tot_init > 0) {
```
jhendersonHDF, line 556: Changes to correctly bring down the H5P interface if it failed to initialize correctly.

MANIFEST
```diff
@@ -1191,10 +1191,10 @@
 ./test/external_common.h
 ./test/external_env.c
 ./test/external_fname.h
-./test/family_v16_00000.h5
-./test/family_v16_00001.h5
-./test/family_v16_00002.h5
-./test/family_v16_00003.h5
```
jhendersonHDF, line 1205: Renamed these files to match the family VFD's new default printf config, which is to append a %06d suffix to the given filename.

test/h5test.c
```diff
@@ -853,150 +846,6 @@ h5_fileaccess_flags(unsigned flags)
...
- * Return:      Success:    0
- *              Failure:    -1
- *
- * Programmer:  Dana Robinson
- *              February 2016
- *
- *-------------------------------------------------------------------------
- */
-herr_t
-h5_get_vfd_fapl(hid_t fapl)
```
jhendersonHDF, line 871: This functionality is now part of the library itself rather than test code. Setting HDF5_DRIVER will replace the default VFD on the default FAPL, so h5_get_vfd_fapl isn't needed for tests.

qkoziol:
@jhendersonHDF - Some of these sound useful now, would you consider making some smaller PRs out of them?

jhendersonHDF:
> @jhendersonHDF - Some of these sound useful now, would you consider making some smaller PRs out of them?

I'm guessing you're referring to changes like the H5P package termination stuff? I can definitely make some small PRs out of changes like that.

qkoziol:
> > @jhendersonHDF - Some of these sound useful now, would you consider making some smaller PRs out of them?
> 
> I'm guessing you're referring to changes like the H5P package termination stuff? I can definitely make some small PRs out of changes like that.

Yes, super, thanks!

byrnHDF (approved):

c++/test/CMakeVFDTests.cmake
```diff
@@ -24,11 +24,29 @@ set (VFD_LIST
     split
     multi
     family
+    splitter
+    #log - log VFD currently has file space allocation bugs
```
jrmainzer, line 28: What are the issues with the log VFD?  Perhaps more to the point, is there are Jira issue on this?

jhendersonHDF: I don't believe there are any JIRA issues for the problems I saw; I should create those. Out of all of the VFDs, the log VFD seemed to be the one that was most problematic in terms of testing. The testhdf5, stab, farray, earray, btree2, dsets, links, unlink, freespace, mf and fheap tests all failed in ways that seemed related to file space allocation and/or checking of file size. The ohdr test failed because it failed to expunge a cache entry that was pinned. Note that these testing failures occurred whether the VFD's logging capabilities were actually enabled or not.

Since the log VFD appears to just try to emulate the sec2 VFD with the logging capabilities on top (but separate from the normal I/O), it seems like these are bugs in that the log VFD does not pass the same tests as the sec2 VFD.

src/H5FDint.c
jrmainzer: This will be changing -- I don't think you will have to worry about this.

src/H5FDdirect.c
```diff
@@ -365,6 +350,53 @@ H5Pget_fapl_direct(hid_t fapl_id, size_t *boundary /*out*/, size_t *block_size /
...
+    /* Set the default to be true for data alignment */
+    fa_out->must_align = TRUE;
+
+    /* Copy buffer size must be a multiple of file block size */
+    if (fa_out->cbsize % fa_out->fbsize != 0)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "copy buffer size must be a multiple of block size")
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5FD__direct_get_default_config() */
```
soumagne: function name

src/H5Pfapl.c
```diff
@@ -808,6 +831,208 @@ H5P__facc_reg_prop(H5P_genclass_t *pclass)
...
+ *              is matched, the ID for that driver is returned through
+ *              `driver_id`. Otherwise, `driver_id` is set to
+ *              H5I_INVALID_HID.
+ *
+ * Return:      Non-negative on sucess/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static inline herr_t
+H5P__facc_set_def_driver_check_predefined(const char *driver_name, hid_t *driver_id)
```
soumagne, line 943: seems a bit long to make it inline :)

src/H5FDcore.c
```diff
@@ -407,6 +414,32 @@ H5FD__core_write_to_bstore(H5FD_core_t *file, haddr_t addr, size_t size)
...
+ *
+ * Return:      Valid Core VFD configuration information pointer (can't
+ *                  fail)
+ *
+ *-------------------------------------------------------------------------
+ */
+static inline const H5FD_core_fapl_t *
+H5FD__core_get_default_config(void)
+{
+    char *driver = HDgetenv("HDF5_DRIVER");
```
soumagne: I feel like it would be better to use a macro instead of "HDF5_DRIVER" everywhere. Just in case that environment variable has to be renamed one day or someone messes up the environment variable name, this way the compiler would be able to detect it for us.

test/app_ref.c
```diff
@@ -92,6 +93,19 @@ main(void)
 
     TESTING("library shutdown with reference count > 1");
 
+    /* Get the VFD to use */
+    env_h5_drvr = HDgetenv("HDF5_DRIVER");
+    if (env_h5_drvr == NULL)
+        env_h5_drvr = "nomatch";
+
```
soumagne, line 100: same here and in the tests in general, I would avoid using strings for const names, e.g. "multi" etc

MANIFEST
derobins: Are five-digit filename extensions a problem by default now? If so, that change seems likely to anger existing family driver users. If not, we should keep some five-digit files around for backward compatibility testing.

c++/test/CMakeVFDTests.cmake
derobins: The log VFD should be a clone of the sec2 VFD, but with log messages. I could see it failing tests that check stdout (like in the tools, etc.), but it's weird that it's failing with respect to file space allocation.

c++/test/CMakeVFDTests.cmake
```diff
@@ -24,11 +24,29 @@ set (VFD_LIST
...
+endif ()
+if (H5_HAVE_ROS3_VFD)
+  set (VFD_LIST ${VFD_LIST} ros3)
+endif ()
+if (H5_HAVE_LIBHDFS)
+  set (VFD_LIST ${VFD_LIST} hdfs)
+endif ()
+if (H5_HAVE_WINDOWS)
+  set (VFD_LIST ${VFD_LIST} windows)
+endif ()
```
derobins, line 49: I don't think expanding this list is a great idea. It's for the CMake equivalent of check-vfd, IIRC and that is going to have all sorts of problems with the VFDs listed here. Have you run the CMake VFD tests? Do they pass? I would be really surprised if they did for, say, S3 since we often check for very "native file format" things that will not be true for object store files.

Also, the Windows VFD doesn't really exist, so that just runs the sec2 tests twice.

config/conclude.am
```diff
@@ -270,10 +270,27 @@ build-check-p: $(LIB) $(PROGS) $(chk_TESTS)
...
+endif
+if MIRROR_VFD_CONDITIONAL
+  VFD_LIST += mirror
+endif
+if ROS3_VFD_CONDITIONAL
+  VFD_LIST += ros3
+endif
+if HDFS_VFD_CONDITIONAL
+  VFD_LIST += hdfs
+endif
```
derobins, line 293: See the CMake comments above.

MANIFEST
jhendersonHDF: Not an issue, no. It's just that the old testing code was using a 5-digit filename extension and I use a 6-digit filename extension as a new default in the family driver to allow for a larger family size. The family driver never used to have a default, so this shouldn't affect user code in any way since users usually specify their own printf-style filename. Having this default just makes it easier to generate testing code for VFDs because the filename doesn't need to have a printf-style format conditionally added on. When it's missing, the family driver will tack on a "-%06d" extension so unique filenames get generated for each family member.

c++/test/CMakeVFDTests.cmake
jhendersonHDF: Yes I have a suspicion that there are some sec2-specific things in the library because the log VFD is indeed just a clone of the sec2 VFD with the logging capability added in, but separate from the regular I/O done. We definitely have a problem somewhere here. Nothing obvious, such as missing VFD feature flags, stuck out to me.

c++/test/CMakeVFDTests.cmake
jhendersonHDF: Part of this PR was to go through all of the tests run under "check vfd" and the cmake equivalent and work around native file-specific things by checking if a VFD is compatible with these types of tests. The last I remember, the only VFDs I didn't test with were the S3 and HDFS VFDs, since I didn't currently have a platform to test on. For the Windows VFD, I don't see any reason NOT to also test it for the sake of completeness. Yes, it might just be the sec2 VFD, but it's still worth making sure it works the way we expect it to.

src/H5FD.c
```diff
@@ -300,6 +300,62 @@ H5FD_register(const void *_cls, size_t size, hbool_t app_ref)
...
+    H5TRACE1("t", "DV", driver_value);
+
+    /* Check if driver with this value is registered */
+    if ((ret_value = H5FD_is_driver_registered_by_value(driver_value, NULL)) < 0)
+        HGOTO_ERROR(H5E_VFL, H5E_CANTGET, FAIL, "can't check if VFD is registered")
+
+done:
+    FUNC_LEAVE_API(ret_value)
+} /* end H5FDis_driver_registered_by_value() */
+
```
derobins, line 358: I really wish we'd set the VOL connector calls that these mirror up as herr_t calls that set an out parameter instead of returning htri_t...

src/H5FDdirect.c
jhendersonHDF: Fixed

src/H5Pfapl.c
jhendersonHDF: Fixed

src/H5FDcore.c
jhendersonHDF: Added a few macros to H5public.h for this and some other related environment variables.

derobins (approved):

lrknox (approved):

PR #610, opened 2021-04-29
Remove the comment with "wild pointer."

hyoklee:
* The term "wild" sounds weird. Simplify the long comment. 
* Initialize pointer.
* Add period at the end of comment.

### DIFF
--- a/src/H5FDlog.c
+++ b/src/H5FDlog.c
@@ -322,18 +322,16 @@ herr_t
 H5Pset_fapl_log(hid_t fapl_id, const char *logfile, unsigned long long flags, size_t buf_size)
 {
     H5FD_log_fapl_t fa;        /* File access property list information */
-    H5P_genplist_t *plist;     /* Property list pointer */
+    H5P_genplist_t *plist = NULL;     /* Property list pointer */
     herr_t          ret_value; /* Return value */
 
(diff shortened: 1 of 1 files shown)


src/H5FDlog.c
```diff
@@ -321,19 +321,17 @@ H5FD__log_term(void)
...
+    H5P_genplist_t *plist = NULL; /* Property list pointer */
+    herr_t          ret_value;    /* Return value */
 
     FUNC_ENTER_API(FAIL)
     H5TRACE4("e", "i*sULz", fapl_id, logfile, flags, buf_size);
 
-    /* Do this first, so that we don't try to free a wild pointer if
-     * H5P_object_verify() fails.
-     */
+    /* Initialize. */
```
derobins, line 331: It's important to do this before calling something that can jump to the error condition. If someone moves this down, the bug will resurface so the comment can't just be "initialize". The whole point of the bugfix was that the initialization was in the wrong place. Also, "wild pointer" is a known term and the pointer in question isn't the plist pointer, it's a char pointer in the H5FD_log_fapl_t struct.

https://en.wikipedia.org/wiki/Dangling_pointer

derobins: Also, we typically don't add periods at the end of one-line coments.

PR #611, opened 2021-04-30
Fixes crashes when size_hint > UINT32_MAX is passed to H5Gcreate1

derobins:


src/H5Oprivate.h
```diff
@@ -684,7 +684,7 @@ typedef struct H5O_bogus_t {
  */
 typedef struct H5O_ginfo_t {
     /* "Old" format group info (not stored) */
-    uint32_t lheap_size_hint; /* Local heap size hint              */
+    size_t lheap_size_hint; /* Local heap size hint              */
```
derobins: I probably need to go look at the file format to make sure this value makes sense.

byrnHDF (approved):

src/H5Gdeprec.c
```diff
@@ -197,7 +196,7 @@ H5Gcreate1(hid_t loc_id, const char *name, size_t size_hint)
             HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, H5I_INVALID_HID, "can't get group info")
 
         /* Set the non-default local heap size hint */
-        H5_CHECKED_ASSIGN(ginfo.lheap_size_hint, uint32_t, size_hint, size_t);
+        ginfo.lheap_size_hint = size_hint;
```
qkoziol: Why the change here?   size_t is probably larger than uint32_t and would be truncated if too large.

src/H5Oprivate.h
qkoziol: Yes, the file format specifies this as a 32-bit value, so allowing it to be a size_t is probably not a good direction.

qkoziol (changes requested):
The size_hint should be range-checked against the maximum 32-bit value and an error thrown when that's exceeded.  Changing the internal values, which are dependent on the file format's 32-bit limit, to size_t is probably not the correct direction.

derobins:
> The size_hint should be range-checked against the maximum 32-bit value and an error thrown when that's exceeded. Changing the internal values, which are dependent on the file format's 32-bit limit, to size_t is probably not the correct direction.

Right. I originally did that, then realized it was simpler to fix the underlying problem, then realized the underlying problem was due to the file format. I'm switching it back to using uint32_t internally.

src/H5Gdeprec.c
derobins: I switched this because I updated the struct field to use size_t, but that's being switched back.

src/H5Oprivate.h
derobins: Right. I'm switching it back.

qkoziol (approved):

PR #618, opened 2021-05-03
Made private my_yyinput function static

seanm:
This prevents it being exported as a public symbol.


seanm:
@gnuoyd

byrnHDF (approved):

derobins (approved):

bmribler:
Sorry for my ignorance.  Why can't it be made to be one version and used in both files?  Second question, what about the other repeated function, H5LTyyerror?

PR #624, opened 2021-05-04
Add selection I/O to chunk code

fortnern:
Used when: not using chunk cache, no datatype conversion, no I/O filters, no page buffer, not using collective
I/O.  Requires global variable H5_use_selection_io_g be set to TRUE.
Also implemented selection to vector I/O translation at the file driver
layer.

### DIFF
--- a/src/H5.c
+++ b/src/H5.c
@@ -83,6 +83,8 @@ hbool_t H5_libinit_g = FALSE; /* Library hasn't been initialized */
 hbool_t H5_libterm_g = FALSE; /* Library isn't being shutdown */
 #endif
 
+hbool_t H5_use_selection_io_g = FALSE;
+
 #ifdef H5_HAVE_MPE
 hbool_t H5_MPEinit_g = FALSE; /* MPE Library hasn't been initialized */
 #endif
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -59,6 +59,7 @@
 #include "H5Iprivate.h"  /* IDs                      */
 #include "H5MMprivate.h" /* Memory management            */
 #include "H5MFprivate.h" /* File memory management               */
+#include "H5PBprivate.h" /* Page Buffer	                         */
 #include "H5VMprivate.h" /* Vector and array functions        */
 
 /****************/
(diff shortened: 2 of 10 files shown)


src/H5.c
```diff
@@ -83,6 +83,8 @@ hbool_t H5_libinit_g = FALSE; /* Library hasn't been initialized */
 hbool_t H5_libterm_g = FALSE; /* Library isn't being shutdown */
 #endif
 
+hbool_t H5_use_selection_io_g = FALSE;
```
qkoziol, line 86: Is this flag intended for permanent use, or just during development?

src/H5Dchunk.c
```diff
@@ -59,6 +59,7 @@
 #include "H5Iprivate.h"  /* IDs                      */
 #include "H5MMprivate.h" /* Memory management            */
 #include "H5MFprivate.h" /* File memory management               */
+#include "H5PBprivate.h" /* Page Buffer	                         */
```
qkoziol, line 62: There seems to be a lot of duplicated code between the read & write operations (more so now), could some of it be extracted as common code?

src/H5private.h
```diff
@@ -2125,6 +2125,11 @@ extern hbool_t H5_libterm_g; /* Is the library being shutdown? */
 
 #endif /* H5_HAVE_THREADSAFE */
 
+/* Extern global to determine if we shoudl use selection I/O if available (this
```
qkoziol, line 2128: Ah, never mind my earlier comment. :-)

src/H5Dchunk.c
fortnern: Do you mean the array allocation/free code?

qkoziol: No, I just put the comment at the top of the file, since it applied to all of your changes.

When reviewing this code, there's large chunks of common code that would benefit from refactoring out into common routines, which would reduce the size of the "main" routines, etc.

fortnern: Hmm, that and the file driver code are the only sections of newly common code I can find

fortnern: I tried to do this (pulling out common code between the selection and non-selection branches of H5D__chunk_write) and it became a mess with the routine needing at least 15 parameters. Another option would be a macro that just uses the variable names as is (not as parameters), though that's ugly as well since changing the variable names breaks the macro. Thoughts?

PR #625, opened 2021-05-04
Fix H5Eget_auto2/H5Eauto_is_v2 to not clear error stack

jhendersonHDF:
Split this change out of VFD plugin PR. These changes fix an issue where information might be lost from the error stack under certain circumstances.

The fact that H5Eget_auto2/H5Eauto_is_v2 clear the error stack seems a little weird to me, especially considering that H5Eset_auto2 does not clear the error stack. It seems like things may have been engineered this way so that the H5Eset_auto2 call in H5E_END_TRY does not clear errors from what's inside the H5E_BEGIN/END_TRY block, but then you realize that any following H5E_BEGIN/END_TRY block will clear that anyway. Since H5Eget_auto2/H5Eauto_is_v2 are just "checking" things, it makes sense to me that they shouldn't clear the error stack.

### DIFF
--- a/src/H5E.c
+++ b/src/H5E.c
@@ -245,6 +245,20 @@ H5E_term_package(void)
         nstk = H5I_nmembers(H5I_ERROR_STACK);
 
         if ((ncls + nmsg + nstk) > 0) {
+            /* Clear the default error stack. Note that
+             * the following H5I_clear_type calls do not
+             * force the clears and will not be able to
+             * clear any error message IDs that are still
+             * in use by the default error stack unless we
+             * clear that stack manually.
+             *
+             * Error message IDs will typically still be
+             * in use by the default error stack when the
+             * application does H5E_BEGIN/END_TRY cleanup
+             * at the very end.
+             */
+            H5E_clear_stack(NULL);
+
             /* Clear any outstanding error stacks */
             if (nstk > 0)
                 (void)H5I_clear_type(H5I_ERROR_STACK, FALSE, FALSE);
@@ -1532,7 +1546,8 @@ H5Eget_auto2(hid_t estack_id, H5E_auto2_t *func /*out*/, void **client_data /*ou
     H5E_auto_op_t op;                  /* Error stack function */
     herr_t        ret_value = SUCCEED; /* Return value */
 
-    FUNC_ENTER_API(FAIL)
+    /* Don't clear the error stack! :-) */
+    FUNC_ENTER_API_NOCLEAR(FAIL)
     H5TRACE3("e", "ixx", estack_id, func, client_data);
 
     if (estack_id == H5E_DEFAULT) {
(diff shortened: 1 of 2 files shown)


src/H5E.c
```diff
@@ -245,6 +245,20 @@ H5E_term_package(void)
...
+             * clear any error message IDs that are still
+             * in use by the default error stack unless we
+             * clear that stack manually.
+             *
+             * Error message IDs will typically still be
+             * in use by the default error stack when the
+             * application does H5E_BEGIN/END_TRY cleanup
+             * at the very end.
+             */
+            H5E_clear_stack(NULL);
```
jhendersonHDF, line 260: When an application does the usual H5E_BEGIN/END_TRY cleanup at the very end, the _current_ behavior in HDF5 actually relies (probably unknowingly) on the fact that H5Eauto_is_v2 clears the error stack. When you do something like:

```
H5E_BEGIN_TRY {
    H5Fclose(fid);
} H5E_END_TRY;
```

at the end of an application, the error message from that failure when fid is invalid gets placed on the error stack. Then, when the library is closing down, the first package that gets terminated (currently H5F in my testing) makes a call to H5I_dec_type_ref -> H5I__destroy_type which contains the following:

```
H5E_BEGIN_TRY
{
    H5I_clear_type(type, TRUE, FALSE);
}
H5E_END_TRY /* don't care about errors */
```

This block has the side effect of clearing the error stack, which is why the error message from the application's cleanup code doesn't get printed on library shutdown.

Modifying H5Eauto_is_v2 to NOT clear the error stack means these error messages get left inside the default error stack and prevent the H5E shutdown code from closing the error stack since the below H5I_clear_type calls don't force the type clearing. Therefore, when the H5E package is terminating and we have error message/class IDs laying around, we now have to manually clear the default error stack. However, I think this is better than relying on one of the packages to accidentally clear the stack. I also think it shouldn't be problematic because anything that needs to be printed should have been at this point and the user has had plenty of time to print out the error stack if they wanted to (e.g. to check why their begin/end try block failed).

src/H5E.c
```diff
@@ -1532,16 +1546,22 @@ H5Eget_auto2(hid_t estack_id, H5E_auto2_t *func /*out*/, void **client_data /*ou
...
     } /* end if */
-    else if (NULL == (estack = (H5E_t *)H5I_object_verify(estack_id, H5I_ERROR_STACK)))
-        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a error stack ID")
+    else {
+        /* Only clear the error stack if it's not the default stack */
+        H5E_clear_stack(NULL);
+
+        if (NULL == (estack = (H5E_t *)H5I_object_verify(estack_id, H5I_ERROR_STACK)))
+            HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a error stack ID")
+    } /* end else */
```
jhendersonHDF, line 1564: In each routine changed, this pattern just follows every other H5E routine that doesn't clear the error stack; the default error stack gets cleared if the routine is operating on an error stack other than the default one.

qkoziol (approved):
Seems like a reasonable improvement

soumagne (approved):

PR #631, opened 2021-05-06
Add group operation tests and object header continuation block tests  to VFD SWMR

kyang2014:
The group operation tests are still in vfd_swmr_group_writer.c. Overall the subroutines are not that many and the testing frame work is easy to be adapted in this way. Test description is as follows:
1. For both old-style and new-style group creation, the following tests are added:
create_group, delete_group,move_group, insert_links, delete_links

2. For new-styled group creation, the link storage transit from compact to dense and from dense to compact are tested.
3. Also add object header continuation block tests for both old-style and new-style group creation. the tests are:
add_ohr_block, del_ohr_block


test/vfd_swmr_group_writer.c
```diff
@@ -2981,58 +3304,2076 @@ verify_group(state_t *s, unsigned int which)
...
+     * then wait for the reader to verify and send an acknowledgement message back.*/
+    if (s->use_named_pipes && s->grp_op_test == true) {
+        dbgf(2, "Writer: ready to send the message: %d\n", s->np_notify + 1);
+        if (np_wr_send_receive(s) == false) {
+            H5_FAILED();
+            AT();
+            /* Note: This is (mostly) because the verification failure message
+             *       from the reader. So don't send the error message back to
+             *       the reader. Just stop the test. */
+            goto error2;
```
vchoi-hdfgroup, line 3201: I don't quite understand the "error2" and "error" situation.

test/vfd_swmr_group_writer.c
```diff
@@ -3109,37 +5450,39 @@ main(int argc, char **argv)
...
-        s.attr_test = true;
-        if (s.use_named_pipes)
-            s.csteps = 1;
-    }
+    if (s.attr_test && s.use_named_pipes)
+        s.csteps = 1;
+
+    /* For group operation test, force the named pipe to communicate in every step. */
+    if (s.grp_op_test && s.use_named_pipes)
+        s.csteps = 1;
```
vchoi-hdfgroup, line 5068: I think s.attr_test and s.grp_op_test cover all the test cases.  When will s.csteps be not set to 1?

test/vfd_swmr_group_writer.c
```diff
@@ -3109,37 +5450,39 @@ main(int argc, char **argv)
...
 
-            wg_ret = write_group(&s, step);
+            wg_ret = group_operations(&s, step);
 
             if (wg_ret == false) {
                 H5_FAILED();
                 AT();
                 printf("write_group failed at step %d\n", step);
 
                 /* At communication interval, notifies the reader about the failture and quit */
```
vchoi-hdfgroup: Minor: spelling--"failture"--> should be failure

test/vfd_swmr_group_writer.c
```diff
@@ -1676,6 +1870,16 @@ write_group(state_t *s, unsigned int which)
         goto error;
     }
 
+    /* We need to create a dummy dataset for the object header contiuation block test. */
```
vchoi-hdfgroup: Minor: spelling "contiuation"-->continuation

test/vfd_swmr_group_writer.c
```diff
@@ -2981,58 +3304,2076 @@ verify_group(state_t *s, unsigned int which)
...
+    H5F_vfd_swmr_config_t config;
+    const char *          fifo_writer_to_reader = "./fifo_group_writer_to_reader";
+    const char *          fifo_reader_to_writer = "./fifo_group_reader_to_writer";
+    int                   fd_writer_to_reader = -1, fd_reader_to_writer = -1;
+    int                   notify = 0, verify = 0;
+    bool                  wg_ret = false;
+    bool                  vg_ret = false;
+
+    if (!state_init(&s, argc, argv)) {
+        H5_FAILED();
```
vchoi-hdfgroup: I suggest to replace all the occurrences of the following to "TEST_ERROR"
[ H5_FAILED(); AT(); goto error;]

TEST_ERROR is a macro defined in h5test.h and is generally used by the tests in the test directory.  (See other tests in the test directory.)

test/vfd_swmr_group_writer.c
kyang2014: The default group test is the case when "s.attr_test and s.grp_op_test both false". For that test, s.csteps is not set to 1.

test/vfd_swmr_group_writer.c
kyang2014: Thanks done.

test/vfd_swmr_group_writer.c
kyang2014: Good, I will replace "[ H5_FAILED(); AT(); goto error;]" with TEST_ERROR. The " goto error2" has to stay the same.

test/vfd_swmr_group_writer.c
kyang2014: For the test case that goes to the label "error",  the error occurs on the writer side, so before it exits from the testing,  the writer needs to send  an error message to the reader, signal the reader that  "the test on my side" is wrong, please stop your test. Otherwise, the reader may still wait for a message from the writer. The test on the reader side will hang.
For the test case  that goes to the label "error2",  the error actually occurs on the reader side, the writer receives an error message from the reader to let the test stop, so the writer doesn't need to send an error message to the reader again. It just exits out of the current test.

test/vfd_swmr_group_writer.c
kyang2014: Thanks done.

test/vfd_swmr_group_writer.c
kyang2014: Updated the code.

vchoi-hdfgroup (approved):

raylu-hdf (approved):
I checked it twice.  This PR looked fine to me.

PR #637, opened 2021-05-10
Purges UFAIL from the library

derobins:
Having a magic "unsigned fail" value that isn't correct for large unsigned types probably isn't the best idea. Developers should use constants like SIZE_MAX as "bad" values (or better yet, return herr_t and make the former return value an out parameter) instead of doing unnatural things like casting negative values to an unsigned type.


src/H5private.h
```diff
@@ -363,7 +363,6 @@
  */
 #define SUCCEED 0
 #define FAIL    (-1)
-#define UFAIL   (unsigned)(-1)
```
derobins, line 366: Picking 2^32 as the "bad" value isn't going to work so hot when assigned to size_t on a 64-bit platform.

src/H5VM.c
```diff
@@ -43,20 +43,20 @@ static void H5VM__stride_copy2(hsize_t nelmts, hsize_t elmt_size, unsigned dst_n
...
+ *              modified by this function.
  *
- * Return:	None
+ * Return:      void
  *
- * Programmer:	Robb Matzke
- *		Saturday, October 11, 1997
+ * Programmer:  Robb Matzke
+ *              Saturday, October 11, 1997
  *
```
derobins, line 60: Sorry, the misaligned comments were bugging me.

src/H5VM.c
```diff
@@ -895,57 +885,57 @@ H5VM_array_fill(void *_dst, const void *src, size_t size, size_t count)
...
- * Programmer:	Quincey Koziol
- *		Monday, April 28, 2003
+ * Programmer:  Quincey Koziol
+ *              Monday, April 28, 2003
  *
  *-------------------------------------------------------------------------
  */
-herr_t
+void
 H5VM_array_down(unsigned n, const hsize_t *total_size, hsize_t *down)
```
derobins, line 903: If a function can't fail, just make it void.

src/H5VM.c
```diff
@@ -992,38 +981,36 @@ H5VM_array_offset(unsigned n, const hsize_t *total_size, const hsize_t *offset)
     hsize_t acc_arr[H5VM_HYPER_NDIMS]; /* Accumulated size of down dimensions */
     hsize_t ret_value;                 /* Return value */
 
-    FUNC_ENTER_NOAPI((HDabort(), 0)) /*lint !e527 Don't worry about unreachable statement */
+    FUNC_ENTER_NOAPI_NOERR
```
derobins, line 984: This probably couldn't be tripped in practice, but let's not encourage crashing applications because HDF5 is having a bad day.

byrnHDF (approved):

src/H5HL.c
```diff
@@ -508,27 +508,33 @@ END_FUNC(STATIC) /* end H5HL__dirty() */
...
 
-    /* check arguments */
+    /* Check arguments */
     HDassert(f);
     HDassert(heap);
     HDassert(buf_size > 0);
     HDassert(buf);
+    HDassert(offset_out);
+
+    *offset_out = SIZE_MAX;
```
qkoziol: I would avoid touching *offset_out until there's no possibility for errors, so you don't modify the user's state when if an error does occur.  As much as possible, we should try to have "atomic" operations that only modify state on success.

qkoziol (changes requested):
Just the one small suggested change in H5HL_insert, but otherwise looks like a nice cleanup!

src/H5HL.c
derobins: I went ahead and changed this, but we should probably think more carefully about what policy we want. Immediately setting output values to "likely to fail" values is probably more robust when people pass in pointers to uninitialized state and then don't bother to check error codes.

qkoziol: Agree -pros & cons either way.   Seems better to be "atomic" to me and encourage people to check error codes rather than mutating their data when an error occurs, but worth a discussion some time.

qkoziol (approved):

PR #644, opened 2021-05-15
Reduces overly-pedantic casting in the public headers

derobins:
Removes unnecessary casting of integer types in the public headers. Casting small integer values to size_t, etc. is overly pedantic and raises old-style-cast warnings when the headers are included in C++ programs.


src/H5ESpublic.h
```diff
@@ -25,7 +25,7 @@
 /*****************/
 
 /* Default value for "no event set" / synchronous execution */
-#define H5ES_NONE (hid_t)0
+#define H5ES_NONE 0 /* (hid_t) */
```
derobins, line 29: I left the type in a comment so that the integer is less of a magic number.

src/H5Dpublic.h
```diff
@@ -28,9 +28,9 @@
 /*****************/
 
 /* Macros used to "unset" chunk cache configuration parameters */
-#define H5D_CHUNK_CACHE_NSLOTS_DEFAULT ((size_t)-1)
-#define H5D_CHUNK_CACHE_NBYTES_DEFAULT ((size_t)-1)
-#define H5D_CHUNK_CACHE_W0_DEFAULT     (-1.0f)
+#define H5D_CHUNK_CACHE_NSLOTS_DEFAULT SIZE_MAX
+#define H5D_CHUNK_CACHE_NBYTES_DEFAULT SIZE_MAX
+#define H5D_CHUNK_CACHE_W0_DEFAULT     (-1.0)
```
derobins, line 34: Casting -1 to an unsigned integral type is an unnatural subversion of the type system and unnecessary. It's more intentional to use SIZE_MAX, etc.

byrnHDF (approved):

qkoziol (approved):

PR #646, opened 2021-05-15
Merge develop PR#625 to hdf5_1_12

lrknox:


lrknox:
This was only supposed to change 2 files.  I'll close and try again.

PR #654, opened 2021-05-18
A batch of life-cycle examples for different modules

gheber:
Preview at [HDF5 Reference Manual](http://doxygen.hdf5.s3-website.us-east-2.amazonaws.com/_r_m.html)


doxygen/examples/H5PL_examples.c
```diff
@@ -0,0 +1,97 @@
...
+    //! <!-- [create] -->
+    {
+        __label__ fail_set;
+        // enable ONLY filter plugins
+        if (H5PLset_loading_state(H5PL_FILTER_PLUGIN) < 0) {
+            ret_val = EXIT_FAILURE;
+            goto fail_set;
+        }
+
+        // ensure that "/tmp" is a the front of the search path list
```
ChristopherHogan: "is a" -> "is at"?

ChristopherHogan (approved):

doxygen/examples/H5PL_examples.c
gheber: :+1:

configure.ac
```diff
@@ -1102,9 +1102,11 @@ if test "X$HDF5_DOXYGEN" = "Xyes"; then
   AC_SUBST([DOXYGEN_HTML_FOOTER])
   AC_SUBST([DOXYGEN_HTML_EXTRA_STYLESHEET])
   AC_SUBST([DOXYGEN_HTML_EXTRA_FILES])
+  AC_SUBST([DOXYGEN_TAG_FILE])
```
byrnHDF, line 1105: These same changes need to be made to the CMake locations.

byrnHDF (changes requested):
There probably needs to be a Release.txt entry.
And we need to stop creating these large PRs. There are at least three ways to divide this; the configuration file, the source files and the dox files.

byrnHDF:
Can all these changes be applied to 1.12 and 1.10 branches? Other then obvious version sections?

configure.ac
gheber: Done.

byrnHDF: The CMake files are in c++/src/CMakeLists.txt and src/CMakeLists.txt. I can do that afterwards if you want.

byrnHDF: Don't forget DOXYGEN_PREDEFINED

byrnHDF: I think the exact same changes can be made to c++/src/CMakeLists.txt.

byrnHDF: And DOXYGEN_EXAMPLES_DIRECTORY.

byrnHDF (approved):
I will rework CMake doxygen after this is merged.

lrknox (approved):

PR #655, opened 2021-05-18
Make H5Dpublic.h work for CentOS 5.

hyoklee:
This will work on Jam / Koala.  Otherwise, you will get 
```
make[2]: Entering directory `/tmp/hdf5/c++/test'
  CXX      dsets.o
dsets.cpp: In function 'herr_t test_chunk_cache(const H5::FileAccPropList&)':
dsets.cpp:1213: error: 'SIZE_MAX' was not declared in this scope
```


derobins:
I'm not excited about hacks for supporting ancient compilers. This is fundamentally a problem of the C++ wrappers, which are stuck on C++98, and not a problem in the C library, where C99 is required. We should be endeavoring to move them to C++11, not applying band-aids so that C++98 compilers can build the wrappers.

If you really want to fix this, you should probably look into defining __STDC_LIMIT_MACROS and/or __STDC_CONSTANT_MACROS when building on creaky old gcc 4.1, but honestly, I don't think there's any reason to support such ancient compilers, especially when it comes to C++, where we should be looking forwards, not backwards.

qkoziol (approved):
*sigh*

derobins (changes requested):
I would prefer that we not merge this.

PR #661, opened 2021-05-19
Use internal version of H5Eprint2 to avoid possible stack overflow

jhendersonHDF:
Looking for feedback on whether this is the right way to go or not. If H5Eprint2 happens to fail, the FUNC_LEAVE_API on the way out will try to print the error stack by calling H5E_dump_api_stack. This will call the error stack's printing function, which is likely to be H5Eprint2. Then, H5Eprint2 will likely fail in the same way again and we've suddenly created a stack overflow. By using an internal non-API version of H5Eprint2 for the library's default printing function, we can at least avoid this chain for most use cases.

The main issue I can see is that if a user happens to manually set H5Eprint2 as the library's default printing function again, the issue can crop up once again. Alternatively, maybe H5Eprint2 should just not use FUNC_ENTER/FUNC_LEAVE macros at all? It seems to serve the dual purpose of being a user API to print the current error stack, as well as serving as the library's default printing function. Perhaps we should have two separate functions, where the user API-level call uses FUNC_ENTER/FUNC_LEAVE and the library callback version does not (or at least does not use FUNC_ENTER_API)?

### DIFF
--- a/src/H5E.c
+++ b/src/H5E.c
@@ -85,6 +85,7 @@ static H5E_t *    H5E__get_current_stack(void);
 static herr_t     H5E__set_current_stack(H5E_t *estack);
 static herr_t     H5E__close_stack(H5E_t *err_stack, void **request);
 static ssize_t    H5E__get_num(const H5E_t *err_stack);
+static herr_t     H5E__print2(hid_t err_stack, FILE *stream);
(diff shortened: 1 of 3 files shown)


byrnHDF:
I like the idea of two functions - this would allow the correct function to be used for the correct purpose.

qkoziol (approved):
Looks good to me

soumagne (approved):

PR #675, opened 2021-05-21
Fixes a minor valgrind issue in the chunk_info test

derobins:
The chunk_info test incorrectly set the number of bytes in a compressed
buffer, which caused valgrind to complain about writing uninitialized
bytes.

### DIFF
--- a/test/chunk_info.c
+++ b/test/chunk_info.c
@@ -449,7 +449,7 @@ error:
  * Return:      Success:    SUCCEED
  *              Failure:    FAIL
  *
- * Note:        Note that the dataspace argument in these new functions are
+ * Note:        Note that the dataspace argument in these new functions is
  *              currently not used.  The functionality involved the dataspace
  *              will be implemented in the next version.
  *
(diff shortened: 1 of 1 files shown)


test/chunk_info.c
```diff
@@ -544,14 +544,15 @@ test_get_chunk_info_highest_v18(hid_t fapl)
     /* Allocate input (compressed) buffer */
     inbuf = HDmalloc(z_dst_nbytes);
 
-    /* Set chunk size to the compressed chunk size and the chunk point
-       to the compressed data chunk */
-    chunk_size = (hsize_t)z_dst_nbytes;
```
derobins, line 549: Setting chunk_size here sets it to the size of the allocated buffer and writes the entire buffer, including any space after the compressed region, which is uninitialized.

test/chunk_info.c
```diff
@@ -544,14 +544,15 @@ test_get_chunk_info_highest_v18(hid_t fapl)
...
+    /* zlib-friendly alias for the input buffer */
     z_dst      = (Bytef *)inbuf;
 
     /* Perform compression from the source to the destination buffer */
     ret = compress2(z_dst, &z_dst_nbytes, z_src, z_src_nbytes, aggression);
 
+    /* Set the chunk size to the compressed chunk size */
+    chunk_size = (hsize_t)z_dst_nbytes;
+
```
derobins, line 555: The correct place to set chunk_size is after the call to compress2, when z_dst_nbytes will be set to the size of the compressed data.

derobins:
This is going directly to 1.12 since we're working on a test failure in that branch. I'll also set up a PR for develop.

byrnHDF (approved):

jhendersonHDF (approved):

qkoziol (approved):

PR #676, opened 2021-05-22
Fixes for a couple of trivial warnings

derobins:


tools/test/h5jam/h5jamgentest.c
```diff
@@ -277,7 +277,7 @@ gent_ub(const char *filename, size_t ub_size, size_t ub_fill)
         for (u = 0; u < ub_fill; u++)
             *bp++ = pattern[u % 10];
 
-        HDwrite(fd, buf, ub_size);
+        (void)HDwrite(fd, buf, ub_size);
```
derobins, line 281: The generators generally ignore errors. They should probably be rewritten to not do that, but in the meantime it's okay to ignore the warning about not using the return value from write().

byrnHDF (approved):

lrknox (approved):

PR #677, opened 2021-05-22
Brings chunk_info valgrind fix from 1.12

derobins:
Fixes uninitialized bytes alert from valgrind caused by not updating the buffer size after running compress2.


lrknox (approved):

byrnHDF (approved):

test/chunk_info.c
```diff
@@ -544,14 +544,15 @@ test_get_chunk_info_highest_v18(hid_t fapl)
     /* Allocate input (compressed) buffer */
     inbuf = HDcalloc(1, z_dst_nbytes);
```
derobins, line 545: Using calloc here quiets valgrind's 'uninitialized bytes' warning (1.12 used malloc), but still writes the larger, incorrect number of bytes.

PR #678, opened 2021-05-23
Removes alternative function enter/leave macro scheme

derobins:
An alternative function enter/leave scheme was added in HDF5 1.10 for the H5EA, H5FA, and H5HL packages. This was never propagated throughout the library nor is there any plan to do so. These macros make the code look even less like standard C than our existing macro scheme while not really adding any useful functionality.

This PR converts the H5EA, H5FA, and H5HL packages to use the original macro scheme and removes the second set of macros from the library.


byrnHDF (approved):

src/H5EA.c
```diff
@@ -990,27 +1015,26 @@ BEGIN_FUNC(PRIV, ERR, int, H5_ITER_CONT, H5_ITER_ERROR,
...
         /* Make callback */
-        if ((cb_ret = (*op)(u, elmt, udata)) < 0) {
-            H5E_PRINTF(H5E_BADITER, "iterator function failed");
-            H5_LEAVE(cb_ret)
-        } /* end if */
-    }     /* end for */
-
-    CATCH
+        if ((ret_value = (*op)(u, elmt, udata)) < 0)
+            HGOTO_DONE(ret_value)
```
jhendersonHDF: Should this be an HGOTO_ERROR with the old "iterator function failed" message rather than a goto_done?

src/H5FA.c
```diff
@@ -671,44 +687,42 @@ END_FUNC(PRIV) /* end H5FA_delete() */
...
-        if ((cb_ret = (*op)(u, elmt, udata)) < 0) {
-            H5E_PRINTF(H5E_BADITER, "iterator function failed");
-            H5_LEAVE(cb_ret)
-        } /* end if */
-    }     /* end for */
-
-    CATCH
+        /* Invoke callback */
+        if ((ret_value = (*op)(u, elmt, udata)) < 0)
+            HGOTO_DONE(ret_value)
```
jhendersonHDF: Same comment about maybe using HGOTO_ERROR

src/H5EA.c
qkoziol: This should be an HERROR with a break out of the loop or HGOTO_DONE, similar to other iterators in the library.   For a good example, see the H5A__iterate_common / H5A__iterate routines in src/H5Aint.c

src/H5FA.c
qkoziol: Same here:  HERROR w/break out of loop is typical for the rest of the library.

qkoziol (approved):
The two iterators should be switched to use HERROR (which allows for pushing an error & string on the error stack), instead of HGOTO_DONE, but otherwise looks good.

src/H5EA.c
derobins: Fixed

src/H5FA.c
derobins: Fixed

PR #679, opened 2021-05-24
Various CMake changes

derobins:
* Removes unused HDF5_ENABLE_HSIZET
* Switches TEST_SHELL_SCRIPTS to ON
* Sets SH_PROGRAM to bash instead of sh


config/cmake/ConfigureChecks.cmake
```diff
@@ -86,14 +86,6 @@ else ()
...
 endif ()
 
-#-----------------------------------------------------------------------------
-#  Are we going to use HSIZE_T
-#-----------------------------------------------------------------------------
-option (HDF5_ENABLE_HSIZET "Enable datasets larger than memory" ON)
-if (HDF5_ENABLE_HSIZET)
-  set (${HDF_PREFIX}_HAVE_LARGE_HSIZET 1)
-endif ()
-
```
derobins, line 96: As far as I can tell, this is not used anywhere in the library or elsewhere in the build system.

config/cmake_ext_mod/HDFMacros.cmake
```diff
@@ -29,9 +29,9 @@ macro (SET_HDF_BUILD_TYPE)
   endif()
   if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.15.0")
-      message (VERBOSE "Setting build type to 'RelWithDebInfo' as none was specified.")
+      message (VERBOSE "Setting build type to 'Debug' as none was specified.")
     endif()
-    set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build." FORCE)
+    set(CMAKE_BUILD_TYPE Debug CACHE STRING "Choose the type of build." FORCE)
     # Set the possible values of build type for cmake-gui
```
derobins: This should be Debug by default for develop, not RelWithDebInfo, for parity with the Autotools.

test/CMakeTests.cmake
```diff
@@ -870,7 +870,7 @@ if (BUILD_SHARED_LIBS)
 ##############################################################################
 endif ()
 
-option (TEST_SHELL_SCRIPTS "Enable shell script tests" OFF)
+option (TEST_SHELL_SCRIPTS "Enable shell script tests" ON)
 if (TEST_SHELL_SCRIPTS)
```
derobins, line 874: We should be running these all the time.

test/ShellTests.cmake
```diff
@@ -17,7 +17,7 @@
 
 if (UNIX)
 
-  find_program (SH_PROGRAM sh)
+  find_program (SH_PROGRAM bash)
   if (SH_PROGRAM)
```
derobins, line 21: Some shell script tests use bash-isms.

byrnHDF (approved):
Except for the DEBUG default - everything else is good

derobins:
> Except for the DEBUG default - everything else is good

I'll switch Debug back, but we should have a discussion at some point about parity between the Autotools and CMake. I'd like for those to track each other.

lrknox (approved):

PR #691, opened 2021-05-27
Split H5L.c internal code to H5Lint.c

derobins:
This structures H5L like the other packages in the library and will be helpful if we split the native VOL connector code into its own subdirectory.


byrnHDF (approved):

jhendersonHDF (approved):

src/H5L.c
```diff
@@ -1614,11 +1432,10 @@ H5Lis_registered(H5L_type_t id)
...
     /* Is the link class already registered? */
-    for (i = 0; i < H5L_table_used_g; i++)
-        if (H5L_table_g[i].id == id) {
-            ret_value = TRUE;
-            break;
-        } /* end if */
+    if (H5L_is_registered(id, &is_registered) < 0)
+        HGOTO_ERROR(H5E_LINK, H5E_BADTYPE, FAIL, "could not determine registration status of UD link type")
+
+    ret_value = is_registered ? TRUE : FALSE;
```
qkoziol, line 1438: From the Dept. of Redundant Redundancy. ;-) ;-)     [Could just assign the boolean variable, instead of checking if it was true]

qkoziol (approved):
Excellent!  I've been meaning to break this apart too. :-)

soumagne:
@qkoziol is it not going to create conflicts with your combo PR though ?

qkoziol:
> @qkoziol is it not going to create conflicts with your combo PR though ?

Probably, but so have lots of other merges, and I've been accommodating them as they come in.

soumagne (approved):

src/H5L.c
jhendersonHDF: I tend to do this solely because htri_t and hbool_t are distinct types. In practice, that shouldn't (?) matter, but I try not to assign different types to each other.

qkoziol: > I tend to do this solely because htri_t and hbool_t are distinct types. In practice, that shouldn't (?) matter, but I try not to assign different types to each other.

That doesn't really apply here, since this code is still assigning a boolean to an htri_t. :-)

lrknox (approved):

PR #699, opened 2021-05-27
Fix H5FD_mpio_Debug="rw" can report MPI_count instead of bytes

brtnfld:
H5FD_mpio_Debug="rw" is meant to profile  the offset and bytes written/read, but currently bytes written/read can be MPI_count and not the actual bytes written/read. 

Changed to reporting bytes written/read.


src/H5FDmpio.c
```diff
@@ -1349,6 +1343,12 @@ H5FD__mpio_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_UNU
     if (bytes_read < 0 || bytes_read > io_size)
         HGOTO_ERROR(H5E_IO, H5E_READERROR, FAIL, "file read failed")
 
+#ifdef H5FDmpio_DEBUG
+    if (H5FD_mpio_debug_r_flag)
+        HDfprintf(stderr, "%s: (%d) mpi_off = %ld  size_i = %lld\n", FUNC, file->mpi_rank, (long)mpi_off,
```
qkoziol: Could you switch the printf string to say 'bytes_read' instead of 'size_t' now?

src/H5FDmpio.c
```diff
@@ -1566,6 +1560,12 @@ H5FD__mpio_write(H5FD_t *_file, H5FD_mem_t type, hid_t H5_ATTR_UNUSED dxpl_id, h
     if (bytes_written != io_size || bytes_written < 0)
         HGOTO_ERROR(H5E_IO, H5E_WRITEERROR, FAIL, "file write failed")
 
+#ifdef H5FDmpio_DEBUG
+    if (H5FD_mpio_debug_w_flag)
+        HDfprintf(stderr, "%s: (%d) mpi_off = %ld  size_i = %lld\n", FUNC, file->mpi_rank, (long)mpi_off,
```
qkoziol: Could you switch the printf string to say 'bytes_written' instead of 'size_t' now?

qkoziol (approved):

src/H5FDmpio.c
brtnfld: That would make my post-processing easier. Hopefully, not too many users depend on it being "size_i" for their scripts. Switched.

src/H5FDmpio.c
brtnfld: done

lrknox (approved):

PR #706, opened 2021-05-29
Fixes for C++ warnings and build issues

derobins:
This PR clears all C++ warnings under gcc 9.3. The Autotools and CMake C++ build infrastructure was also updated to not test for functionality that is now available in C++11.


.github/workflows/main.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             os: ubuntu-latest
             build_type: "Release"
             cpp: ON
-            fortran: OFF
+            fortran: ON
             java: ON
```
derobins, line 40: Unclear why this is here. This was previously brought in via an earlier PR.

PR #707, opened 2021-05-29
C++ warning and build fixes

derobins:
This PR fixes all C++ warnings (under gcc 9.3, at least), though I have turned off warnings for the "rule of five" issues until we're ready to tackle that. As a part of these changes, I also updated the build system (both CMake and Autotools) to not check for pre-C++11 features that have been standardized.


c++/src/H5Cpp.h
```diff
@@ -48,14 +48,4 @@
 #include "H5File.h"
 #include "H5Library.h"
 
-/* Some C++ compilers do not have offsetof macro; define to bypass the problem
-   - BMR- -EIP- 2007/08/01
-*/
```
derobins, line 53: This is no longer true. C++11 adds additional functionality on top of offset(), so it will always be present.

c++/test/tcompound.cpp
```diff
@@ -115,7 +115,7 @@ test_compound_2()
             s_ptr->d    = i * 8 + 6;
             s_ptr->e    = i * 8 + 7;
         }
-        memcpy(buf, orig, nelmts * sizeof(src_typ_t));
+        HDmemcpy(buf, orig, nelmts * sizeof(src_typ_t));
```
derobins, line 119: I plan to eventually remove everything private and then the HD prefixes will go away, but until then I'll keep the pedantry.

c++/test/th5s.cpp
```diff
@@ -200,6 +200,7 @@ test_h5s_basic()
         // CHECK_I(ret, "H5Fclose");  // leave this here, later, fake a failure
         // in the p_close see how this will handle it. - BMR
 
+        // When running in valgrind, this PASSED macro will be missed
         PASSED();
     } // end of try block
```
derobins, line 205: Valgrind appears to have issues with the nested try block.

config/cmake/HDFCXXCompilerFlags.cmake
```diff
@@ -115,106 +111,24 @@ if (NOT MSVC AND NOT MINGW)
...
-      else ()
-        ADD_H5_FLAGS (H5_CXXFLAGS "${HDF5_SOURCE_DIR}/config/gnu-warnings/no-developer-4.8")
-      endif ()
-    endif ()
-
-    # Append more extra warning flags that only gcc 4.9+ know about
+    # Append more extra warning flags that only gcc 4.9+ knows about
     if (NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.9)
-      # autotools always add the C flags with the CXX flags
-      ADD_H5_FLAGS (H5_CXXFLAGS "${HDF5_SOURCE_DIR}/config/gnu-warnings/4.9")
```
derobins, line 163: This is (a) not true and (b) triggers an enormous number of "not a C++ warning" warnings.

config/cmake_ext_mod/HDFCXXTests.cpp
```diff
@@ -1,46 +0,0 @@
...
-  {
-    unsigned char type;
-    unsigned char num;
-    unsigned int len;
-  };
-  typedef struct index_st index_t;
-  int x,y;
-  x = offsetof(struct index_st, len);
-  y = offsetof(index_t, num)
-
```
derobins, line 41: Testing for offset() is unnecessary. This file and the infrastructure that used it have been removed.

config/gnu-warnings/cxx-9
```diff
@@ -0,0 +1,2 @@
+# Turn this on when the C++ wrappers obey the Rule of Five
+-Wno-deprecated-copy
```
derobins, line 2: There are a LOT of these warnings, but adding things like move constructors seemed like a job for later.

configure.ac
```diff
@@ -776,20 +776,6 @@ if test "X$HDF_CXX" = "Xyes"; then
...
-  PAC_PROG_CXX_NAMESPACE
-
-  ## if C++ can handle static cast
-  PAC_PROG_CXX_STATIC_CAST
-
-  ## Checking if C++ has offsetof extension,
-  ## note: this test has to be the last of the C++ tests because it sets a definition
-  ## which would be used in the other tests, causing them to fail.
-  PAC_PROG_CXX_OFFSETOF
-
```
derobins, line 792: None of this requires checking.

config/cmake/HDFCXXCompilerFlags.cmake
```diff
@@ -211,10 +207,11 @@ if (NOT MSVC AND NOT MINGW)
       endif ()
     endif ()
 
-    # Append more extra warning flags that only gcc 9.x+ know about
+    # Append more extra warning flags that only gcc 9.x+ knows about
     if (NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
       # autotools always add the C flags with the CXX flags
```
byrnHDF, line 212: Why does autotools add the C flags, is it required? They might conflict?

byrnHDF (approved):

bmribler (approved):

lrknox (approved):

PR #709, opened 2021-05-30
Cleans up hsize_t and haddr_t size guessing

derobins:
Also removes hacks that paper over missing C99 integer types

### DIFF
--- a/src/H5public.h
+++ b/src/H5public.h
@@ -254,7 +254,7 @@ typedef uint64_t hsize_t;
  * \internal Defined as a (minimum) 64-bit integer type. Use of hssize_t
  * should be discouraged in new code.
  */
-typedef int64_t  hssize_t;
+typedef int64_t hssize_t;
 #define PRIdHSIZE          PRId64
 #define PRIiHSIZE          PRIi64
 #define PRIoHSIZE          PRIo64
--- a/src/H5public.h
+++ b/src/H5public.h
@@ -217,7 +217,14 @@ typedef int herr_t;
 typedef bool hbool_t;
 typedef int  htri_t;
 
-/* Define the ssize_t type if it not is defined */
+/* The signed version of size_t
+ *
+ * ssize_t is POSIX and not defined in any C standard. It's used in some
+ * public HDF5 API calls so this work-around will define it if it's not
+ * present.
+ *
+ * Use of ssize_t should be discouraged in new code.
+ */
 #if H5_SIZEOF_SSIZE_T == 0
 /* Undefine this size, we will re-define it in one of the sections below */
 #undef H5_SIZEOF_SSIZE_T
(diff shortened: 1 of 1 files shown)


derobins:
@gheber - Please check out the Doxygen comments I added

src/H5public.h
```diff
@@ -217,7 +217,14 @@ typedef int herr_t;
...
 
-/* Define the ssize_t type if it not is defined */
+/* The signed version of size_t
+ *
+ * ssize_t is POSIX and not defined in any C standard. It's used in some
+ * public HDF5 API calls so this work-around will define it if it's not
+ * present.
+ *
+ * Use of ssize_t should be discouraged in new code.
+ */
```
derobins, line 227: Intentionally not a Doxygen comment since this is a work-around vs a new HDF5 type. ssize_t is defined pretty much everywhere now, but since it's not defined in any C standard we should probably leave the work-around in for now.

Ideally, we'd purge "negative values = badness" behavior from the library and could then get rid of the types, but since ssize_t and hssize_t are used in the public API we're stuck with them for now.

src/H5public.h
```diff
@@ -235,137 +242,47 @@ typedef long long ssize_t;
 #endif
 #endif
 
-/* int64_t type is used for creation order field for links.  It may be
- * defined in Posix.1g, otherwise it is defined here.
- */
```
derobins, line 240: Obsolete as we require C99 types now. Ditto uint64_t and uint32_t, below.

src/H5public.h
```diff
@@ -235,137 +242,47 @@ typedef long long ssize_t;
...
+typedef int64_t hssize_t;
+#define PRIdHSIZE          PRId64
+#define PRIiHSIZE          PRIi64
+#define PRIoHSIZE          PRIo64
+#define PRIuHSIZE          PRIu64
+#define PRIxHSIZE          PRIx64
+#define PRIXHSIZE          PRIX64
+#define H5_SIZEOF_HSIZE_T  8
+#define H5_SIZEOF_HSSIZE_T 8
+#define HSIZE_UNDEF        UINT64_MAX
```
derobins, line 266: Since we require C99 types and format specifiers now, we can more simply define h(s)size_t and haddr_t.

byrnHDF (approved):

lrknox (approved):

src/H5public.h
gheber: Why not do something similar to `h[s]size_t`, i.e., have a description for users and then an `\internal` elaboration?

gheber (approved):

jhendersonHDF (approved):

src/H5public.h
derobins: Because ssize_t is a POSIX type and people can google for its purpose. We don't need to Doxygen up wrappers that paper over platform inconsistencies.

qkoziol (approved):

PR #710, opened 2021-06-01
Updates CMake + Fortran to correctly guess type sizes

derobins:
CMake was incorrectly determining the type sizes on CMake versions > 3.14. This fix reverts to using the pre-3.14 behavior while we rework the type guessing code.

### DIFF
--- a/config/cmake/HDF5UseFortran.cmake
+++ b/config/cmake/HDF5UseFortran.cmake
@@ -32,11 +32,9 @@ macro (READ_SOURCE SOURCE_START SOURCE_END RETURN_VAR)
 endmacro ()
 
 set (RUN_OUTPUT_PATH_DEFAULT ${CMAKE_BINARY_DIR})
-if (NOT CMAKE_VERSION VERSION_LESS "3.14.0")
-  if (HDF5_REQUIRED_LIBRARIES)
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

brtnfld (commented):
So this works now with cmake > 3.14 now, why was that condition added?

derobins:
> So this works now with cmake > 3.14 now, why was that condition added?

Allen says that it was part of an effort to run fewer programs on HPC systems. Also, check_fortran_source_runs is probably less hacky than our FORTRAN_RUN macro.

byrnHDF:
> > So this works now with cmake > 3.14 now, why was that condition added?
> 
> Allen says that it was part of an effort to run fewer programs on HPC systems. Also, check_fortran_source_runs is probably less hacky than our FORTRAN_RUN macro.

Right, the macro uses the TRY_RUN to compile and execute the feature check, in CMake 3.14, kitware added a command to do just that, which simplifies the requirements for doing a TRY_RUN. However, our feature test also wants to create output files for use later by a program. There seems to be a conflict with this secondary need and the the primary yes/no feature need.

Maybe we need to revisit this configure issue and find a better way forward.

brtnfld (approved):
looks fine to remove the use of check_fortran_source_runs.

byrnHDF (changes requested):
Possibly the source function is incorrect.

PR #715, opened 2021-06-02
Updates PGI C/C++ configurations

derobins:
Tweaks some settings and configures pgc++ to use C++11

### DIFF
--- a/config/pgi-cxxflags
+++ b/config/pgi-cxxflags
@@ -1,7 +1,6 @@
 #							-*- shell-script -*-
 #
 # Copyright by The HDF Group.
-# Copyright by the Board of Trustees of the University of Illinois.
 # All rights reserved.
 #
 # This file is part of HDF5.  The full HDF5 copyright notice, including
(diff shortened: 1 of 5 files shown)


config/pgi-cxxflags
```diff
@@ -0,0 +1,97 @@
+#							-*- shell-script -*-
+#
+# Copyright by The HDF Group.
+# Copyright by the Board of Trustees of the University of Illinois.
```
lrknox:
```suggestion
```
line 4 should be removed since no part of this file existed before 2006.

lrknox (changes requested):
Remove just the extra copyright line that doesn't apply

config/pgi-cxxflags
derobins: Done

lrknox (approved):

byrnHDF (commented):
What about CMake?

derobins:
> What about CMake?

CMake sets the C++ standard the same way for all compilers, which is the fundamental problem being addressed here. I'll normalize the flags across platforms when the daily tests stabilize.

PR #716, opened 2021-06-03
Adds an Intel C++ configuration file

derobins:


lrknox (approved):

byrnHDF (commented):
What about CMake?

derobins:
> What about CMake?

CMake sets the C++ standard the same way for all compilers, which is the fundamental problem being addressed here. I'll normalize the flags across platforms when the daily tests stabilize.

PR #720, opened 2021-06-03
Fix several warnings

jhendersonHDF:
Fixes a few issues uncovered from GCC's static analysis, along with several pesky warnings such as strncpy with a size equal to the source object's size, all of the unused H5FD package variable warnings, etc.

### DIFF
--- a/java/src/jni/h5util.c
+++ b/java/src/jni/h5util.c
@@ -870,11 +870,8 @@ h5str_sprintf(JNIEnv *env, h5str_t *out_str, hid_t container, hid_t tid, void *i
             }
             else {
                 if (typeSize > 0) {
-                    if (NULL == (this_str = (char *)HDmalloc(typeSize + 1)))
+                    if (NULL == (this_str = HDstrdup(tmp_str)))
                         H5_OUT_OF_MEMORY_ERROR(ENVONLY, "h5str_sprintf: failed to allocate string buffer");
-
-                    HDstrncpy(this_str, tmp_str, typeSize);
-                    this_str[typeSize] = '\0';
                 }
             }
 
@@ -3664,18 +3661,13 @@ obj_info_all(hid_t loc_id, const char *name, const H5L_info2_t *info, void *op_d
     info_all_t *datainfo = (info_all_t *)op_data;
     H5O_info2_t object_info;
     htri_t      object_exists;
-    size_t      str_len;
 
     datainfo->otype[datainfo->count]     = -1;
     datainfo->ltype[datainfo->count]     = -1;
     datainfo->obj_token[datainfo->count] = H5O_TOKEN_UNDEF;
 
-    str_len = HDstrlen(name);
-    if (NULL == (datainfo->objname[datainfo->count] = (char *)HDmalloc(str_len + 1)))
+    if (NULL == (datainfo->objname[datainfo->count] = HDstrdup(name)))
         goto done;
-
-    HDstrncpy(datainfo->objname[datainfo->count], name, str_len);
-    (datainfo->objname[datainfo->count])[str_len] = '\0';
 
     if ((object_exists = H5Oexists_by_name(loc_id, name, H5P_DEFAULT)) < 0)
         goto done;
@@ -3702,7 +3694,6 @@ obj_info_max(hid_t loc_id, const char *name, const H5L_info2_t *info, void *op_d
 {
     info_all_t *datainfo = (info_all_t *)op_data;
     H5O_info2_t object_info;
-    size_t      str_len;
 
     datainfo->otype[datainfo->count]     = -1;
     datainfo->ltype[datainfo->count]     = -1;
@@ -3710,12 +3701,8 @@ obj_info_max(hid_t loc_id, const char *name, const H5L_info2_t *info, void *op_d
     datainfo->obj_token[datainfo->count] = H5O_TOKEN_UNDEF;
 
     /* This will be freed by h5str_array_free(oName, n) */
-    str_len = HDstrlen(name);
-    if (NULL == (datainfo->objname[datainfo->count] = (char *)HDmalloc(str_len + 1)))
+    if (NULL == (datainfo->objname[datainfo->count] = HDstrdup(name)))
         goto done;
-
-    HDstrncpy(datainfo->objname[datainfo->count], name, str_len);
-    (datainfo->objname[datainfo->count])[str_len] = '\0';
 
     if (H5Oget_info3(loc_id, &object_info, H5O_INFO_ALL) < 0)
         goto done;
--- a/src/H5E.c
+++ b/src/H5E.c
@@ -639,7 +639,7 @@ H5E__get_class_name(const H5E_cls_t *cls, char *name, size_t size)
 
     /* Set the user's buffer, if provided */
     if (name) {
-        HDstrncpy(name, cls->cls_name, MIN((size_t)(len + 1), size));
+        HDstrncpy(name, cls->cls_name, size);
         if ((size_t)len >= size)
             name[size - 1] = '\0';
     } /* end if */
--- a/src/H5Eint.c
+++ b/src/H5Eint.c
@@ -129,7 +129,7 @@ H5E__get_msg(const H5E_msg_t *msg, H5E_type_t *type, char *msg_str, size_t size)
 
     /* Copy the message into the user's buffer, if given */
     if (msg_str) {
-        HDstrncpy(msg_str, msg->msg, MIN((size_t)(len + 1), size));
+        HDstrncpy(msg_str, msg->msg, size);
         if ((size_t)len >= size)
             msg_str[size - 1] = '\0';
     } /* end if */
--- a/src/H5FDmulti.c
+++ b/src/H5FDmulti.c
@@ -646,7 +646,7 @@ H5FD_multi_sb_encode(H5FD_t *_file, char *name /*out*/, unsigned char *buf /*out
     H5Eclear2(H5E_DEFAULT);
 
     /* Name and version number */
-    strncpy(name, "NCSAmulti", (size_t)8);
+    strncpy(name, "NCSAmult", (size_t)9);
     name[8] = '\0';
 
     assert(7 == H5FD_MEM_NTYPES);
@@ -682,7 +682,7 @@ H5FD_multi_sb_encode(H5FD_t *_file, char *name /*out*/, unsigned char *buf /*out
     p = buf + 8 + nseen * 2 * 8;
     UNIQUE_MEMBERS (file->fa.memb_map, mt) {
         size_t n = strlen(file->fa.memb_name[mt]) + 1;
-        strncpy((char *)p, file->fa.memb_name[mt], n);
+        strcpy((char *)p, file->fa.memb_name[mt]);
         p += n;
         for (i = n; i % 8; i++)
             *p++ = '\0';
--- a/src/H5FDsplitter.c
+++ b/src/H5FDsplitter.c
@@ -340,10 +340,12 @@ H5Pset_fapl_splitter(hid_t fapl_id, H5FD_splitter_vfd_config_t *vfd_config)
         HGOTO_ERROR(H5E_VFL, H5E_CANTALLOC, FAIL, "unable to allocate file access property list struct")
 
     info->ignore_wo_errs = vfd_config->ignore_wo_errs;
-    HDstrncpy(info->wo_path, vfd_config->wo_path, H5FD_SPLITTER_PATH_MAX);
-    HDstrncpy(info->log_file_path, vfd_config->log_file_path, H5FD_SPLITTER_PATH_MAX);
-    info->rw_fapl_id = H5P_FILE_ACCESS_DEFAULT; /* pre-set value */
-    info->wo_fapl_id = H5P_FILE_ACCESS_DEFAULT; /* pre-set value */
+    HDstrncpy(info->wo_path, vfd_config->wo_path, H5FD_SPLITTER_PATH_MAX + 1);
+    info->wo_path[H5FD_SPLITTER_PATH_MAX] = '\0';
+    HDstrncpy(info->log_file_path, vfd_config->log_file_path, H5FD_SPLITTER_PATH_MAX + 1);
+    info->log_file_path[H5FD_SPLITTER_PATH_MAX] = '\0';
+    info->rw_fapl_id                            = H5P_FILE_ACCESS_DEFAULT; /* pre-set value */
+    info->wo_fapl_id                            = H5P_FILE_ACCESS_DEFAULT; /* pre-set value */
 
     /* Set non-default channel FAPL IDs in splitter configuration info */
     if (H5P_DEFAULT != vfd_config->rw_fapl_id) {
@@ -412,8 +414,8 @@ H5Pget_fapl_splitter(hid_t fapl_id, H5FD_splitter_vfd_config_t *config /*out*/)
     if (NULL == (fapl_ptr = (const H5FD_splitter_fapl_t *)H5P_peek_driver_info(plist_ptr)))
         HGOTO_ERROR(H5E_PLIST, H5E_BADVALUE, FAIL, "unable to get specific-driver info")
 
-    HDstrncpy(config->wo_path, fapl_ptr->wo_path, H5FD_SPLITTER_PATH_MAX);
-    HDstrncpy(config->log_file_path, fapl_ptr->log_file_path, H5FD_SPLITTER_PATH_MAX);
+    HDstrncpy(config->wo_path, fapl_ptr->wo_path, H5FD_SPLITTER_PATH_MAX + 1);
+    HDstrncpy(config->log_file_path, fapl_ptr->log_file_path, H5FD_SPLITTER_PATH_MAX + 1);
     config->ignore_wo_errs = fapl_ptr->ignore_wo_errs;
 
     /* Copy R/W and W/O FAPLs */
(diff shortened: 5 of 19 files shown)


java/src/jni/h5util.c
```diff
@@ -873,7 +873,7 @@ h5str_sprintf(JNIEnv *env, h5str_t *out_str, hid_t container, hid_t tid, void *i
                     if (NULL == (this_str = (char *)HDmalloc(typeSize + 1)))
                         H5_OUT_OF_MEMORY_ERROR(ENVONLY, "h5str_sprintf: failed to allocate string buffer");
 
-                    HDstrncpy(this_str, tmp_str, typeSize);
+                    HDstrcpy(this_str, tmp_str);
```
jhendersonHDF: We already know the size of the buffer; don't use a strncpy with length equal to the buffer size.

java/src/jni/h5util.c
```diff
@@ -3674,7 +3674,7 @@ obj_info_all(hid_t loc_id, const char *name, const H5L_info2_t *info, void *op_d
     if (NULL == (datainfo->objname[datainfo->count] = (char *)HDmalloc(str_len + 1)))
         goto done;
 
-    HDstrncpy(datainfo->objname[datainfo->count], name, str_len);
+    HDstrcpy(datainfo->objname[datainfo->count], name);
```
jhendersonHDF: Same as above

java/src/jni/h5util.c
```diff
@@ -3714,7 +3714,7 @@ obj_info_max(hid_t loc_id, const char *name, const H5L_info2_t *info, void *op_d
...
```
jhendersonHDF: Same as above

src/H5.c
```diff
@@ -963,8 +963,12 @@ H5check_version(unsigned majnum, unsigned minnum, unsigned relnum)
         HDsnprintf(lib_str, sizeof(lib_str), "HDF5 library version: %d.%d.%d", H5_VERS_MAJOR, H5_VERS_MINOR,
                    H5_VERS_RELEASE);
         if (*substr) {
-            HDstrncat(lib_str, "-", (size_t)1);
```
jhendersonHDF: Don't use strncat with size equal to source.

src/H5CX.c
```diff
@@ -1768,15 +1768,19 @@ H5CX_get_ring(void)
...
+    H5CX_node_t **head         = NULL; /* Pointer to head of API context list */
+    hbool_t       coll_md_read = FALSE;
 
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
     /* Sanity check */
     head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
     HDassert(head && *head);
 
-    FUNC_LEAVE_NOAPI((*head)->ctx.coll_metadata_read)
```
jhendersonHDF: In this and the following changes, (*head) could be accessed while head is NULL if the FUNC_ENTER fails. Instead, set the return value only when FUNC_ENTER succeeds; return a default value otherwise.

src/H5Dchunk.c
```diff
@@ -7483,7 +7483,7 @@ H5D__chunk_iter_cb(const H5D_chunk_rec_t *chunk_rec, void *udata)
     /* Check for callback failure and pass along return value */
     if ((ret_value = (data->op)(chunk_rec->scaled, chunk_rec->filter_mask, chunk_rec->chunk_addr,
                                 chunk_rec->nbytes, data->op_data)) < 0)
-        HERROR(H5E_DATASET, H5E_CANTNEXT, "iteration operator failed");
+        HGOTO_ERROR(H5E_DATASET, H5E_CANTNEXT, ret_value, "iteration operator failed")
```
jhendersonHDF: eliminates two pesky warnings about the unused 'done:' label, along with an unused 'err_occurred' variable from the FUNC_ENTER macro.

src/H5E.c
```diff
@@ -639,7 +639,7 @@ H5E__get_class_name(const H5E_cls_t *cls, char *name, size_t size)
 
     /* Set the user's buffer, if provided */
     if (name) {
-        HDstrncpy(name, cls->cls_name, MIN((size_t)(len + 1), size));
```
jhendersonHDF, line 642: While using MIN here can make strncpy do less work in the case that len + 1 is smaller than size, GCC isn't intelligent enough to figure out what we're doing here, so stick to something sensible.

src/H5Eint.c
```diff
@@ -129,7 +129,7 @@ H5E__get_msg(const H5E_msg_t *msg, H5E_type_t *type, char *msg_str, size_t size)
 
     /* Copy the message into the user's buffer, if given */
     if (msg_str) {
-        HDstrncpy(msg_str, msg->msg, MIN((size_t)(len + 1), size));
```
jhendersonHDF, line 132: Same as above

src/H5FDhdfs.c
```diff
@@ -20,8 +20,10 @@
  *             File System (HDFS).
  */
 
+#ifdef H5_HAVE_LIBHDFS
 /* This source code file is part of the H5FD driver module */
 #include "H5FDdrvr_module.h"
+#endif
```
jhendersonHDF: Eliminates a giant warning about unused H5FD package variables.

src/H5FDmirror.c
```diff
@@ -15,12 +15,12 @@
...
  */
 
-#include "H5FDdrvr_module.h" /* This source code file is part of the H5FD driver module */
-
 #include "H5private.h" /* Generic Functions        */
 
 #ifdef H5_HAVE_MIRROR_VFD
 
+#include "H5FDdrvr_module.h" /* This source code file is part of the H5FD driver module */
+
```
jhendersonHDF: Same as above.

src/H5FDmulti.c
```diff
@@ -624,7 +624,7 @@ H5FD_multi_sb_encode(H5FD_t *_file, char *name /*out*/, unsigned char *buf /*out
     H5Eclear2(H5E_DEFAULT);
 
     /* Name and version number */
-    strncpy(name, "NCSAmulti", (size_t)8);
+    strncpy(name, "NCSAmult", (size_t)9);
```
jhendersonHDF, line 649: Previously was trying to copy 8 bytes out of a 10 byte (including NUL terminator) string?

src/H5FDmulti.c
```diff
@@ -660,7 +660,7 @@ H5FD_multi_sb_encode(H5FD_t *_file, char *name /*out*/, unsigned char *buf /*out
     p = buf + 8 + nseen * 2 * 8;
     UNIQUE_MEMBERS (file->fa.memb_map, mt) {
         size_t n = strlen(file->fa.memb_name[mt]) + 1;
-        strncpy((char *)p, file->fa.memb_name[mt], n);
+        strcpy((char *)p, file->fa.memb_name[mt]);
```
jhendersonHDF, line 685: Using strcpy here is no less safe than strncpy. We don't know the total size of the output buffer (the API doesn't give us that info), and using strncpy with a size based on the source length will throw a warning anyway.

src/H5FDros3.c
```diff
@@ -22,8 +22,10 @@
  *     Relies on "s3comms" utility layer to implement the AWS REST API.
  */
 
+#ifdef H5_HAVE_ROS3_VFD
 /* This source code file is part of the H5FD driver module */
 #include "H5FDdrvr_module.h"
+#endif
```
jhendersonHDF: Another eliminated H5FD package variable warning.

src/H5FDsplitter.c
```diff
@@ -340,10 +340,12 @@ H5Pset_fapl_splitter(hid_t fapl_id, H5FD_splitter_vfd_config_t *vfd_config)
         HGOTO_ERROR(H5E_VFL, H5E_CANTALLOC, FAIL, "unable to allocate file access property list struct")
 
     info->ignore_wo_errs = vfd_config->ignore_wo_errs;
-    HDstrncpy(info->wo_path, vfd_config->wo_path, H5FD_SPLITTER_PATH_MAX);
-    HDstrncpy(info->log_file_path, vfd_config->log_file_path, H5FD_SPLITTER_PATH_MAX);
-    info->rw_fapl_id = H5P_FILE_ACCESS_DEFAULT; /* pre-set value */
-    info->wo_fapl_id = H5P_FILE_ACCESS_DEFAULT; /* pre-set value */
+    HDstrncpy(info->wo_path, vfd_config->wo_path, H5FD_SPLITTER_PATH_MAX + 1);
+    info->wo_path[H5FD_SPLITTER_PATH_MAX] = '\0';
```
jhendersonHDF, line 344: Make sure to copy the whole path and ensure it's NUL-terminated, otherwise GCC complains about truncated output.

src/H5Fsuper.c
```diff
@@ -1043,7 +1043,7 @@ H5F__super_read(H5F_t *f, H5P_genplist_t *fa_plist, hbool_t initial_read)
                 HDONE_ERROR(H5E_FILE, H5E_CANTUNPIN, FAIL, "unable to unpin driver info")
 
             /* Evict the driver info block from the cache */
-            if (H5AC_expunge_entry(f, H5AC_DRVRINFO, sblock->driver_addr, H5AC__NO_FLAGS_SET) < 0)
+            if (sblock && H5AC_expunge_entry(f, H5AC_DRVRINFO, sblock->driver_addr, H5AC__NO_FLAGS_SET) < 0)
```
jhendersonHDF, line 1046: GCC complains about a potential dereference of a NULL sblock here.

src/H5PL.c
```diff
@@ -359,7 +359,7 @@ H5PLget(unsigned int idx, char *path_buf, size_t buf_size)
 
     /* If the path buffer is not NULL, copy the path to the buffer */
     if (path_buf) {
-        HDstrncpy(path_buf, path, MIN((size_t)(path_len + 1), buf_size));
```
jhendersonHDF, line 362: Same comment about GCC being confused about the usage of MIN here.

src/H5Pdapl.c
```diff
@@ -1453,7 +1453,7 @@ H5Pget_efile_prefix(hid_t plist_id, char *prefix /*out*/, size_t size)
         /* Copy to user's buffer, if given */
         len = HDstrlen(my_prefix);
         if (prefix) {
-            HDstrncpy(prefix, my_prefix, MIN(len + 1, size));
```
jhendersonHDF, line 1456: Same as above

src/H5Pdapl.c
```diff
@@ -1543,7 +1543,7 @@ H5Pget_virtual_prefix(hid_t plist_id, char *prefix /*out*/, size_t size)
...
```
jhendersonHDF, line 1546: Same as above

src/H5Pencdec.c
```diff
@@ -333,7 +333,7 @@ H5P__encode_cb(H5P_genprop_t *prop, void *_udata)
         /* Encode (or not, if the 'encode' flag is off) the property's name */
         prop_name_len = HDstrlen(prop->name) + 1;
         if (udata->encode) {
-            HDstrncpy((char *)*(udata->pp), prop->name, prop_name_len);
```
jhendersonHDF, line 336: Again, we don't know the size of the output buffer, so using strcpy here is no less safe than strncpy and strncpy gives a warning since we're just giving it the length of the source being copied.

testpar/t_span_tree.c
```diff
@@ -35,6 +35,8 @@
 #include "H5private.h"
 #include "testphdf5.h"
 
+#define LOWER_DIM_SIZE_COMP_TEST__RUN_TEST__DEBUG 0
```
jhendersonHDF: Moving this up in the file eliminates a warning about it being undefined and evaluating to 0

tools/src/h5dump/h5dump_ddl.c
```diff
@@ -1421,10 +1429,10 @@ lnk_search(const char *path, const H5L_info2_t *li, void *_op_data)
     else {
         if (k == 2) {
             HDstrcpy(search_name, "/");
-            HDstrncat(search_name, op_name, search_len + 1);
+            HDstrcat(search_name, op_name);
```
jhendersonHDF, line 1432: Already know the buffer size, use strcat to not generate a warning.

src/H5FDmulti.c
bmribler: Regardless which one, shouldn't it be HD...?

jhendersonHDF: Ah, sure, that's a good point. Let me fix all the usages that were never converted over.

test/objcopy.c
```diff
@@ -4498,6 +4498,10 @@ test_copy_dataset_compressed(hid_t fcpl_src, hid_t fcpl_dst, hid_t src_fapl, hid
     TESTING("H5Ocopy(): compressed dataset");
 
 #ifndef H5_HAVE_FILTER_DEFLATE
+    (void)fcpl_src;
+    (void)fcpl_dst;
+    (void)src_fapl;
+    (void)dst_fapl; /* Silence compiler */
```
derobins: One of these days I'm going to go through the library and replace all the H5_ATTR_UNUSED macro nonsense with void statements, which are more portable and less of a maintenance headache.

jhendersonHDF: I'm all for this, though we should probably set up a variadic macro to do any number of arguments in one go so we don't have this big stack of them.

src/H5FDmulti.c
jhendersonHDF: It turns out that the multi driver uses only public HDF5 headers, so it doesn't use the HD-prefixed versions of standard library/system calls.

java/src/jni/h5util.c
qkoziol: Could just use a strdup in this case.

java/src/jni/h5util.c
qkoziol: Ditto

java/src/jni/h5util.c
qkoziol: Ditto

src/H5.c
qkoziol: This might be shorter / simpler as:  
HDsnprintf(lib_str, sizeof(lib_str), "HDF5 library version: %d.%d.%d%s%s", H5_VERS_MAJOR, H5_VERS_MINOR, H5_VERS_RELEASE, (*substr ? "-" : ""), substr);

Instead of the previous or updated code.

src/H5CX.c
qkoziol: No, there's no way for this FUNC_ENTER macro to fail, the compiler is just confused.

No objection to the change if it quiets the warning though. :-)

src/H5Dchunk.c
qkoziol: This isn't the correct idiom for the library.  To fix this problem, the routine should be using FUNC_ENTER_STATIC_NOERR

java/src/jni/h5util.c
jhendersonHDF: Sure, I was just going for the minimal amount of changes necessary here.

src/H5CX.c
jhendersonHDF: Ah, well even if the macro can't fail, this will also be problematic if the pkg_init var is not TRUE, because head only gets set if the package has been initialized. GCC's not confused ;)

src/H5FDmulti.c
qkoziol: Nope  - there's only room for 8 bytes in the "Driver ID" field for the file format:  https://portal.hdfgroup.org/display/HDF5/File+Format+Specification

src/H5Dchunk.c
jhendersonHDF: This seems a little bit misled? Why use a NOERR macro when we're going to push an error onto the stack?

src/H5Fsuper.c
qkoziol: This is trickier than that - the sblock should be set to NULL after the H5AC_unprotect, about 15 lines earlier.  But, if that's the case, how's the expunge supposed to work?   There's a bigger algorithmic problem here.

src/H5FDmulti.c
jhendersonHDF: I understand that. However,

1. "NCSAmulti" is 10 bytes long. Let's not confuse the issue by copying from a string that's larger than what we're actually using. Just format the string according to what we're actually going to use, e.g. "NCSAmult".
2. The header comment specifically says a 9-byte buffer is used: 8 bytes for the file format piece + a NUL terminator. Therefore, I want to copy at most 9 bytes from the source. I realize that we immediately manually NUL terminate the string, but I prefer to use strncpy how it's intended.

src/H5.c
jhendersonHDF: Sure, it probably makes more sense to switch to using snprintf where we're able to anyway, rather than having to manually manage strncats.

src/H5Fsuper.c
jhendersonHDF: Sure, but the algorithm is beyond me and adding an sblock check here theoretically won't do more harm than good. If sblock actually were NULL here, we'd immediately crash rather than H5AC_expunge_entry doing something "bad". Now, we just skip the expunge anyway.

jhendersonHDF: As far as I can tell, the H5AC_unprotect call flushes the entry, but I don't see anywhere that the entry is freed/set to NULL, especially since H5AC_unprotect only accepts a single level of pointer indirection. Since the code following the expunge also checks to make sure sblock is valid before using it, this seems like a reasonable change, but maybe @jrmainzer would have a better idea?

java/src/jni/h5util.c
jhendersonHDF: Fixed

java/src/jni/h5util.c
jhendersonHDF: Fixed

java/src/jni/h5util.c
jhendersonHDF: Fixed

src/H5.c
jhendersonHDF: Fixed

src/H5Dchunk.c
jhendersonHDF: It still doesn't make much sense to me to use a NOERR FUNC_ENTER and then issue an error, but changed this back to HERROR with a FUNC_ENTER_STATIC_NOERR and removed the done label.

src/H5FDmulti.c
derobins: The stdio VFD is the same way. They were intended as "demo" VFDs to show people how to write them using the public API (multi = multi-file VFD example, stdio = single-file VFD example).

byrnHDF (approved):

lrknox (approved):

PR #724, opened 2021-06-04
Moves gcc warning macros from H5public.h to H5private.h

derobins:
These do not belong in the public API. They were originally put there for use in the multi VFD (which can only use public headers). Pending future multi VFD refactoring, the macros have been duplicated in H5FDmulti.c.

### DIFF
--- a/src/H5FDmulti.c
+++ b/src/H5FDmulti.c
@@ -45,6 +45,28 @@
 #define my_strdup strdup
 #endif
 
+/* Macros for enabling/disabling particular GCC warnings
+ *
+ * These are (renamed) duplicates of macros in H5private.h. If you make changes
+ * here, be sure to update those as well.
+ *
+ * (see the following web-sites for more info:
+ *      http://www.dbp-consulting.com/tutorials/SuppressingGCCWarnings.html
+ *      http://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html#Diagnostic-Pragmas
+ */
+/* These pragmas are only implemented usefully in gcc 4.6+ */
+#if ((__GNUC__ * 100) + __GNUC_MINOR__) >= 406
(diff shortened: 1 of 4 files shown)


src/H5public.h
```diff
@@ -368,6 +343,13 @@ typedef struct H5_alloc_stats_t {
  */
 typedef void (*H5_atclose_func_t)(void *ctx);
 
+/* API adapter header (defines H5_DLL, etc.) */
+#include "H5api_adpt.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
derobins, line 352: Not crucial for the gcc macros. Just moves the extern "C" and API adapter closer to where they are actually needed.

src/H5FDmulti.c
```diff
@@ -45,6 +45,29 @@
...
+#define H5_MULTI_GCC_DIAG_PRAGMA(x)     H5_MULTI_GCC_DIAG_DO_PRAGMA(GCC diagnostic x)
+
+#define H5_MULTI_GCC_DIAG_OFF(x)                                                                             \
+    H5_MULTI_GCC_DIAG_PRAGMA(push) H5_MULTI_GCC_DIAG_PRAGMA(ignored H5_MULTI_GCC_DIAG_JOINSTR("-W", x))
+#define H5_MULTI_GCC_DIAG_ON(x) H5_MULTI_GCC_DIAG_PRAGMA(pop)
+#else
+#define H5_MULTI_GCC_DIAG_OFF(x)
+#define H5_MULTI_GCC_DIAG_ON(x)
+#endif
+
```
derobins, line 70: Future refactoring will remove the need for these warning macros, which only suppress one type of warning in this VFD.

qkoziol (approved):

lrknox (approved):

byrnHDF (approved):

PR #728, opened 2021-06-04
Includes hdf5dev.h in hdf5.h

derobins:
Might want to ditch hdf5dev.h in the future and just directly include the H5Xdevelop.h files since hdf5dev.h is kind of a moot point with this change, but this PR at least allows the plugins to be built.


qkoziol (changes requested):
I don't think this is the correct direction to go.   Why are you taking the hdf5dev.h header out of these files?   The whole point was to _not_ include this information in the header file for application developers (hdf5.h).

derobins:
> I don't think this is the correct direction to go. Why are you taking the hdf5dev.h header out of these files? The whole point was to _not_ include this information in the header file for application developers (hdf5.h).

It's breaking all the daily tests that build external plugins. Semantically separating the information into separate header files is useful, but having every external thing navigate our new headers isn't a bridge we're going to cross at this time.

qkoziol:
> > I don't think this is the correct direction to go. Why are you taking the hdf5dev.h header out of these files? The whole point was to _not_ include this information in the header file for application developers (hdf5.h).
> 
> It's breaking all the daily tests that build external plugins. Semantically separating the information into separate header files is useful, but having every external thing navigate our new headers isn't a bridge we're going to cross at this time.

I can look into this more, maybe another solution would suggest itself.   Could you send me an example or two of logs that show the failure(s)?

src/hdf5.h
```diff
@@ -41,8 +41,20 @@
 #include "H5VLpublic.h" /* Virtual Object Layer                     */
 #include "H5Zpublic.h"  /* Data filters                             */
 
-/* "developer-oriented" headers (for VFD, VOL, etc. authors) */
-#include "hdf5dev.h"
+/* Application developer headers for various interfaces */
```
qkoziol: These are for plugin authors - definitely not for application developers. :-)

qkoziol:
Dana and I chatted about this a lot and this current iteration seems like a reasonable place to land.

qkoziol (approved):

src/hdf5.h
derobins: I copied that out of the hdf5dev.h header that you authored :)

Will change, though.

qkoziol: Thanks - dunno where I was at with that, or where I pasted it from... :-)

byrnHDF (approved):

lrknox (approved):

PR #742, opened 2021-06-10
Fix Direct VFD Testing in CMake

jhendersonHDF:


c++/test/CMakeVFDTests.cmake
```diff
@@ -26,7 +26,7 @@ set (VFD_LIST
     family
 )
 
-if (DIRECT_VFD)
```
jhendersonHDF, line 29: "DIRECT_VFD" is used by Autotools. CMake uses "H5_HAVE_DIRECT" instead.

byrnHDF (approved):

derobins (approved):

PR #744, opened 2021-06-10
Removed -fanalyzer from the gcc 10 developer options

derobins:
This is extremely slow and generates a LOT of false positives. It's also
not really a warning flag, but an analysis feature and thus belongs in
a separate configure option, not lumped in with the warnings.


qkoziol (approved):
Yes, driving me nuts, how slow it is.

lrknox (approved):

Issue #750, opened 2021-06-14
Configure process will not complete with oneapi C compiler due to use of unsupported flag


frankwillmore:
The configure process was observed to fail because of the unsupported warning flag -Wcheck:

```
CC=mpicc FC=mpifort MPICH_FC=ifx CXX=mpicxx ./configure --enable-parallel --enable-fortran
```

excerpt of config.log:
```
4480 configure:29054: checking how mpicc reports undeclared, standard C functions
4481 configure:29079: mpicc -c   -std=c99 -Wall -Wcheck     -D_GNU_SOURCE -D_POSIX_C_SOURCE=200809L      conftest.c >&5
4482 warning: unknown warning option '-Wcheck' [-Wunknown-warning-option]
4483 conftest.c:136:8: warning: implicitly declaring library function 'strchr' with type 'char *(const char *, int)' [-Wimplicit-     function-declaration]
4484 (void) strchr;
4485        ^
4486 conftest.c:136:8: note: include the header <string.h> or explicitly provide a declaration for 'strchr'
4487 2 warnings generated.
4488 configure:29079: $? = 0
4489 configure:29117: mpicc -c   -std=c99 -Wall -Wcheck     -D_GNU_SOURCE -D_POSIX_C_SOURCE=200809L      conftest.c >&5
4490 warning: unknown warning option '-Wcheck' [-Wunknown-warning-option]
4491 1 warning generated.
4492 configure:29117: $? = 0
4493 configure:29121: error: in `/home/willmore/hdf5-1.10.7':
4494 configure:29123: error: cannot detect from compiler exit status or warnings
CC=mpicc FC=mpifort MPICH_FC=ifx CXX=mpicxx ./configure --enable-parallel --enable-fortran
```

Removing the flag will allow the configuration, installation, and testing to complete, but does not address the underlying check in configure.

frankwillmore:
see https://github.com/HDFGroup/hdf5/pull/751

byrnHDF:
So we will need to do multiple things - remove the -Wcheck from general and put it in a "version 19 and under" file. Then add the appropriate checks and load in the intel-(cxx|f)flags file for autotools and the cmake/HDF5Compiler files.

epourmal:
@frankwillmore we will be looking into the issues. Thank you for reporting!

byrnHDF:
We have some fixes for this and some other flags issues in process.

byrnHDF:
[...]

PR #751, opened 2021-06-14
remove -Wcheck flag which isn't recognized by oneapi/icx

frankwillmore:
configuring with the oneapi C compiler (icx) will fail if the -Wcheck flag is present. This PR does not address the intent of the original check in the configure process. Removing the flag was observed to allow the install to complete and pass tests on a linux-opensuse_leap15-zen2 platform.

Addresses https://github.com/HDFGroup/hdf5/issues/750

### DIFF
--- a/config/intel-warnings/general
+++ b/config/intel-warnings/general
@@ -1,2 +1,1 @@
 -Wall
--Wcheck


config/intel-warnings/general
```diff
@@ -1,2 +1 @@
 -Wall
--Wcheck
```
derobins, line 2: Checking for the suitability of this flag should be done in config/intel-flags, etc. instead of being removed for everyone.

derobins (changes requested):
This needs to be handled further upstream in the platform configuration instead of removing the flag for everyone.

byrnHDF:
This has been addressed in develop and will be merged to other dev branches.
(NOTE: master branches are release specific branches and not development code.)

PR #752, opened 2021-06-14
Maximum jump in tick for VFD SWMR reader

vchoi-hdfgroup:


src/H5Fint.c
```diff
@@ -1573,6 +1573,11 @@ H5F__dest(H5F_t *f, hbool_t flush)
             /* Push error, but keep going*/
             HDONE_ERROR(H5E_IO, H5E_CANTFLUSH, FAIL, "vfd swmr prep for flush or close failed")
 
+        if ((f->shared->vfd_swmr) && (!f->shared->vfd_swmr_writer))
+            HDfprintf(stdout,
+                      "The maximum jump in ticks is %llu; The maximum expected lag configured is %u\n",
```
derobins: Use PRIu64 instead of llu and PRIu32 instead of u. Those should always be used with uint64_t, etc.

derobins: https://stackoverflow.com/questions/9225567/how-to-print-a-int64-t-type-in-c

derobins (changes requested):
Please change the format strings

derobins (approved):

PR #774, opened 2021-06-18
Removes C++ dependency on H5private.h

derobins:
Most C API calls have been removed, aside from a few uses of free, where we just dropped the 'HD'. A couple of H5_ATTR_UNUSED macros were also replaced with (void) statements.

Only applies to the code in src. The code in test uses a bunch of H5private.h functionality which is imported via h5test.h.

### DIFF
--- a/c++/src/H5Attribute.cpp
+++ b/c++/src/H5Attribute.cpp
@@ -14,7 +14,6 @@
 #include <iostream>
 #include <string>
 
-#include "H5private.h" // for HDfree
 #include "H5Include.h"
 #include "H5Exception.h"
 #include "H5IdComponent.h"
@@ -313,7 +312,7 @@ Attribute::getName() const
     H5std_string attr_name; // attribute name to return
 
     // Preliminary call to get the size of the attribute name
-    ssize_t name_size = H5Aget_name(id, static_cast<size_t>(0), NULL);
+    ssize_t name_size = H5Aget_name(id, 0, NULL);
 
     // If H5Aget_name failed, throw exception
     if (name_size < 0) {
@@ -324,8 +323,8 @@ Attribute::getName() const
     }
     // Attribute's name exists, retrieve it
     else if (name_size > 0) {
-        char *name_C = new char[name_size + 1]; // temporary C-string
-        HDmemset(name_C, 0, name_size + 1);     // clear buffer
+        // Create buffer for C string
+        char *name_C = new char[name_size + 1]();
 
         // Use overloaded function
         name_size = getName(name_C, name_size + 1);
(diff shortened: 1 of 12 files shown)


c++/src/H5DataSet.cpp
```diff
@@ -759,7 +758,7 @@ DataSet::p_read_variable_len(const hid_t mem_type_id, const hid_t mem_space_id,
 
     // Get string from the C char* and release resource allocated by C API
     strg = strg_C;
-    HDfree(strg_C);
+    free(strg_C);
```
bmribler, line 761: Should this be delete?

c++/src/H5Attribute.cpp
```diff
@@ -551,7 +551,7 @@ Attribute::p_read_variable_len(const DataType &mem_type, H5std_string &strg) con
 
     // Get string from the C char* and release resource allocated by C API
     strg = strg_C;
-    HDfree(strg_C);
+    free(strg_C);
```
bmribler, line 554: delete?

c++/src/H5DataSet.cpp
derobins: No. We get that memory from the HDF5 C library, so it wasn't allocated with new.

c++/src/H5Attribute.cpp
derobins: Same here. You can't mix malloc/free and new/delete.

bmribler: I know.  It was an accident, but you fixed it by changing HDfree to free, and I asked should it be delete?

c++/src/H5DataSet.cpp
bmribler: Actually, I remember why I used HDfree, because of the very reason you said here.  The right fix is to remove "new char[[data_size + 1]]();" above.  The library will allocate for strg_C.

derobins: That new char[data_size + 1] line is in a different function. In this function, strg_C is just defined as a char pointer and not initialized.

c++/src/H5Attribute.cpp
derobins: It shouldn't be delete[]. If you expand the code, you'll see that srgc_C is an unallocated char pointer. The memory is allocated by the library and should be freed using free().

bmribler:
Oh, ha..ha..., that is funny.  I take your word for it.  Sorry to waste your time.

Binh-Minh

byrnHDF (approved):

lrknox (approved):

PR #782, opened 2021-06-21
Removes checks for system(), which is C89/90

derobins:


tools/test/perform/iopipe.c
```diff
@@ -95,7 +95,6 @@ print_stats(const char *prefix,
 static void
 synchronize(void)
 {
-#ifdef H5_HAVE_SYSTEM
 #if defined(H5_HAVE_WIN32_API) && !defined(__CYGWIN__)
     _flushall();
```
derobins, line 99: As far as I'm aware, this only flushes FILE objects, so it's not the same as POSIX's sync(2) and is probably the wrong thing to do here. FlushFileBuffers() would be better, but that requires the file's HANDLE.

tools/test/perform/iopipe.c
```diff
@@ -107,7 +106,6 @@ synchronize(void)
     status = HDsystem("df >/dev/null");
     HDassert(status >= 0);
 #endif
```
derobins, line 108: Unclear why we're not using sync(2) here instead of system("sync"). Also unclear what the df hack is supposed to be doing.

byrnHDF (approved):

tools/test/perform/iopipe.c
qkoziol: Ancient testcode - not certain either, except that a system sync will flush data from all processes.

qkoziol (approved):

lrknox (approved):

PR #789, opened 2021-06-23
Further simplifies Autotools type size checks

derobins:
Also fixes an issue where clock_gettime and difftime are not detected
due to earlier simplifications of this code.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1351,16 +1351,6 @@ CFLAGS="$H5_CFLAGS $AM_CFLAGS $CFLAGS"
 ## Checkpoint the cache
 AC_CACHE_SAVE
 
-## Write the confdefs.h header for checking sizes
-cat >>confdefs.h <<\EOF
-#include <stdbool.h>
-#include <stdint.h>
-#ifdef HAVE_UNISTD_H
-#include <sys/types.h>
-#endif
-#include <time.h>
-EOF
-
 AC_CHECK_SIZEOF(        [int8_t])
 AC_CHECK_SIZEOF(       [uint8_t])
 AC_CHECK_SIZEOF(  [int_least8_t])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1351,16 +1351,6 @@ CFLAGS="$H5_CFLAGS $AM_CFLAGS $CFLAGS"
...
 AC_CACHE_SAVE
 
-## Write the confdefs.h header for checking sizes
-cat >>confdefs.h <<\EOF
-#include <stdbool.h>
-#include <stdint.h>
-#ifdef HAVE_UNISTD_H
-#include <sys/types.h>
-#endif
-#include <time.h>
```
derobins, line 1361: Putting time.h here messes up checks for clock_gettime and difftime. The rest of this processing is unnecessary since these headers are already included by default in AC_CHECK_SIZEOF processing.

configure.ac
```diff
@@ -1389,12 +1379,12 @@ AC_CHECK_SIZEOF([uint_least64_t])
...
 AC_CHECK_SIZEOF( [uint_fast64_t])
 
-AC_CHECK_SIZEOF([bool])
+AC_CHECK_SIZEOF([bool], [], [#include <stdbool.h>])
 AC_CHECK_SIZEOF([off_t])
 AC_CHECK_SIZEOF([ptrdiff_t])
 AC_CHECK_SIZEOF([size_t])
 AC_CHECK_SIZEOF([ssize_t])
-AC_CHECK_SIZEOF([time_t])
+AC_CHECK_SIZEOF([time_t], [], [#include <time.h>])
```
derobins, line 1387: stdbool.h and time.h are not in the list of standard includes, so we set them separately and only when they are needed.

byrnHDF (approved):

lrknox (approved):

derobins:
For posterity, the list of standard include files is here:

https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Default-Includes.html#Default-Includes

PR #790, opened 2021-06-23
Removes ancient Autotools cruft

derobins:


configure.ac
```diff
@@ -476,15 +476,9 @@ esac
...
 AC_TYPE_OFF_T
-AC_CHECK_TYPE([size_t], [],
-        [AC_DEFINE_UNQUOTED([size_t], [unsigned long],
-                [Define to `unsigned long' if <sys/types.h> does not define.])])
 AC_CHECK_TYPE([ssize_t], [],
         [AC_DEFINE_UNQUOTED([ssize_t], [long],
                 [Define to `long' if <sys/types.h> does not define.])])
-AC_CHECK_TYPE([ptrdiff_t], [],
-        [AC_DEFINE_UNQUOTED([ptrdiff_t], [long],
-                [Define to `long' if <sys/types.h> does not define.])])
```
derobins, line 487: These checks are useless because the types are decorated with the H5_ prefix and we don't use H5_size_t, etc. in the library. size_t and ptrdiff_t are in the C standard anyway, so they are doubly useless. I'm leaving ssize_t and off_t alone for now since those are POSIX and we may want to strip the H5_ and use this definition instead of H5public.h.

configure.ac
```diff
@@ -1229,8 +1223,6 @@ AC_CHECK_LIB([dl], [dlopen])
 ## ----------------------------------------------------------------------
 ## Check for system header files.
 ##
-AC_HEADER_STDC
-AC_HEADER_TIME
```
derobins, line 1233: The gnu docs say that these are archaic and are not needed in new code.

configure.ac
```diff
@@ -2035,8 +2027,6 @@ AC_CHECK_FUNCS([tmpfile asprintf vasprintf waitpid])
 ## ----------------------------------------------------------------------
 ## Check compiler characteristics
 ##
-AC_C_CONST
```
derobins, line 2038: const is now defined in the C standard, so this archaic holdover is no longer needed.

src/H5private.h
```diff
@@ -61,25 +63,6 @@
...
- */
-#if defined(H5_TIME_WITH_SYS_TIME)
-#include <sys/time.h>
-#include <time.h>
-#elif defined(H5_HAVE_SYS_TIME_H)
-#include <sys/time.h>
-#else
-#include <time.h>
-#endif
-
```
derobins, line 82: The gnu docs say that this hack is archaic and unnecessary in new code. I just moved time.h to the list of standard headers and sys/time.h to the list of POSIX headers.

lrknox (approved):

soumagne (approved):

PR #793, opened 2021-06-23
Reorganization of C and POSIX headers in H5public.h & H5private.h

derobins:
Consolidated and removed duplicates


src/H5private.h
```diff
@@ -43,7 +41,6 @@
 #ifdef H5_HAVE_UNISTD_H
 #include <pwd.h>
 #include <unistd.h>
-#include <sys/types.h>
 #include <sys/wait.h>
 #endif
```
derobins, line 49: These are all duplicates from H5public.h

src/H5public.h
```diff
@@ -35,22 +35,20 @@
...
+#include <limits.h>
+#include <stdarg.h>
+#include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
+
+/* POSIX headers */
+#ifdef H5_HAVE_UNISTD_H
+#include <sys/types.h>
 #endif
```
derobins, line 55: Nothing really new here - just consolidated the C and POSIX headers so they are easier to find.

soumagne (approved):

lrknox (approved):

PR #803, opened 2021-06-25
Brings the tools getopt(3) replacement into the main library

derobins:
We should have a version of getopt(3) that is available everywhere in the HDF5 source. I'm not super excited about bringing the tools version over, but it has the advantage of handling long options, definitely works on Windows, and is already in use in the tools code. It's a little weird to put this in the C library when it's intended for use elsewhere, but that's where we put most of our platform-independent hacks and it avoids saddling downstream libraries with awkward dependencies.

This isn't 100% needed in the current library right now, but a flavor of getopt(3) is needed for the VFD SWMR test programs and could be used to replace the ad-hoc command line processing in the existing SWMR tests (and possibly elsewhere).


byrnHDF:
Will this be merged down to the other branches?
Either way can we make the 1.13+ code only use long names and get rid of the HORRIBLE shortened long names?

derobins:
> Will this be merged down to the other branches?
> Either way can we make the 1.13+ code only use long names and get rid of the HORRIBLE shortened long names?

It can move down and I'd like to clean a lot of that stuff up.

byrnHDF:
I would merge this change in all the dev branches then come back and remove the terrible shortened long options for develop going forward.

byrnHDF (approved):

lrknox (approved):

PR #807, opened 2021-06-26
Removes type guesses when C99 types are missing

derobins:


byrnHDF (approved):

lrknox (approved):

qkoziol:
We are assuming that these are available now?

derobins:
We compile with standard set to C99, so yes. Also, the build system changes are probably staying in develop, so this shouldn't alienate users of 1.8 et al.

soumagne (approved):

qkoziol (approved):

PR #820, opened 2021-07-02
Add full API support for selection I/O.  Add tests for this.

fortnern:


src/H5FDint.c
```diff
@@ -736,211 +996,201 @@ H5FD_read_selection(H5FD_t *file, uint32_t count, H5FD_mem_t type, H5S_t *mem_sp
...
+    /* If the file is open for SWMR read access, allow access to data past
+     * the end of the allocated space (the 'eoa').  This is done because the
+     * eoa stored in the file's superblock might be out of sync with the
+     * objects being written within the file by the application performing
+     * SWMR write operations.
+     */
+    /* For now at least, only check that the offset is not past the eoa, since
+     * looking into the highest offset in the selection (different from the
+     * bounds) is potentially expensive.
+     */
```
jrmainzer, line 1146: Assuming we shift to VFD SWMR, the possibility of reading past EOA in the SWMR case will go away, as VFD SWMR should always provide the reader with the correct EOA.

We will need a good way of testing to see if a selection read request extends beyond EOA.

test/vfd.c
```diff
@@ -3985,7 +3985,7 @@ test_vector_io(const char *vfd_name)
     void *     f_read_bufs_1[VECTOR_LEN];  /* fixed read bufs vector       */
     void *     f_read_bufs_2[VECTOR_LEN];  /* fixed read bufs vector       */
 
-    sprintf(test_title, "vector I/O with %s VFD", vfd_name);
+    HDsnprintf(test_title, sizeof(test_title), "vector I/O with %s VFD", vfd_name);
```
jrmainzer, line 3988: Thanks for cleaning this up.

test/vfd.c
```diff
@@ -4352,6 +4352,984 @@ test_vector_io(const char *vfd_name)
...
+    int     i;
+    int     j;
+    hbool_t result = TRUE;
+
+    /* Update write buffer */
+    for (i = 0; i < (int)count; i++)
+        if (wbufs[i] && (i == 0 || wbufs[i] != wbufs[i - 1]))
+            for (j = 0; j < SEL_IO_DIM0 * SEL_IO_DIM1; j++)
+                wbufs[i][j] += 2 * SEL_IO_DIM0 * SEL_IO_DIM1;
+
```
jrmainzer, line 4389: Am I missing something, or are you always setting wbufs[I][j] = 160 here?

jrmainzer (approved):
How hard would it be to add statistics collection to your selection I/O to vector or scalar I/O routines?

test/vfd.c
fortnern: Every time we run a new test case we *increment* by 160 so that each value we write is unique and if something goes wrong and we get data from a previous write we'll know its wrong.

fortnern:
> How hard would it be to add statistics collection to your selection I/O to vector or scalar I/O routines?

Shouldn't be hard

PR #823, opened 2021-07-04
Test 1.12 master merge

lrknox:


lrknox:
There were too many conflicts when creating a PR.  I merged with "-s recursive -Xtheirs", copied files to address a couple dozen files incorrectly changed,  then committed and pushed  the new version to 1.12/master.

PR #826, opened 2021-07-07
H5fget_name_f len_trim fix

### LINKED ISSUE #825: Undefined/unexpected behavior when passing a freshly allocated buffer or whitespace to h5fget_name_f
h5fget_name_f calls len_trim() on its output buffer to obtain the length. This causes the output buffer to be read, resulting in the following problems:

- If the buffer is filled with whitespace, h5fget_name_f treats it as though the buffer had a length of zero. In my tests this resulted in an "Invalid write" error from Valgrind, so this is apparently triggering an array out of bounds error somehow as well.
- Possibly undefined behavior may result if a freshly-allocated buffer is passed in.

h5fget_name_f should instead call len() which obtains the total buffer length without reading the buffer contents.

jhaiduce:
`h5fget_name_f` was calling the intrinsic `len_trim` to determine the length of the passed buffer. This can cause problems if the passed buffer contains trailing whitespace (in which case the h5fget_name_f will limit itself to the length of the non-whitespace characters currently in the passed buffer, even if the buffer's capacity is larger), or if the passed buffer was freshly allocated (in which case the resulting buffer length is unpredictable).

The pull request replaces the call to `len_trim` with a call to `len`, which returns the total buffer length regardless of the passed contents of the buffer, and is safe to call on a freshly allocated (i.e. unassigned) buffer.

This fixes #825.

### DIFF
--- a/fortran/src/H5Fff.F90
+++ b/fortran/src/H5Fff.F90
@@ -844,7 +844,7 @@ CONTAINS
          CHARACTER(KIND=C_CHAR), DIMENSION(*), INTENT(OUT) :: buf
        END FUNCTION h5fget_name_c
     END INTERFACE
-    buflen = LEN_TRIM(buf)
+    buflen = LEN(buf)
     hdferr = h5fget_name_c(obj_id, size, buf, buflen)
   END SUBROUTINE h5fget_name_f
 !****s* H5F/h5fget_filesize_f
--- a/fortran/test/tH5F.F90
+++ b/fortran/test/tH5F.F90
@@ -545,9 +545,11 @@ CONTAINS
               CALL check("h5fget_name_f",error,total_error)
               IF(name_size /= 9) THEN
                  write(*,*) "file name obtained from the dataset id has wrong length"
+                 call check("h5fget_name_f",-1,total_error)
               END IF
-              IF(file_name(1:name_size) .NE. fix_filename(1:name_size)) THEN
+              IF(file_name(1:name_size) /= fix_filename(1:name_size)) THEN
                  write(*,*) "file name obtained from the dataset id is incorrect"
+                 call check("h5fget_name_f",-1,total_error)
               END IF
 
          !
--- a/fortran/test/tH5F.F90
+++ b/fortran/test/tH5F.F90
@@ -543,6 +543,9 @@ CONTAINS
          !
          CALL h5fget_name_f(dset_id, file_name, name_size, error)
               CALL check("h5fget_name_f",error,total_error)
+              IF(name_size /= 9) THEN
+                 write(*,*) "file name obtained from the dataset id has wrong length"
+              END IF
               IF(file_name(1:name_size) .NE. fix_filename(1:name_size)) THEN
                  write(*,*) "file name obtained from the dataset id is incorrect"
               END IF
--- a/fortran/test/tH5F.F90
+++ b/fortran/test/tH5F.F90
@@ -532,24 +532,13 @@ CONTAINS
          !
          CALL h5dopen_f(reopen_id, dsetname, dset_id, error)
               CALL check("h5dopen_f",error,total_error)
-
-         ! Populate filename buffer with whitespace
-         do i = 1,80
-            file_name(i:i) = " "
-         end do
-
          !
          !Get file name from the dataset identifier
          !
          CALL h5fget_name_f(dset_id, file_name, name_size, error)
               CALL check("h5fget_name_f",error,total_error)
-              IF(name_size /= 9) THEN
-                 write(*,*) "file name obtained from the dataset id has wrong length"
-                 call check("h5fget_name_f",-1,total_error)
-              END IF
-              IF(file_name(1:name_size) /= fix_filename(1:name_size)) THEN
+              IF(file_name(1:name_size) .NE. fix_filename(1:name_size)) THEN
                  write(*,*) "file name obtained from the dataset id is incorrect"
-                 call check("h5fget_name_f",-1,total_error)
               END IF
 
          !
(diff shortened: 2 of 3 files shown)


epourmal:
Please provide testing code that shows the problem before the fix and passes after the fix.
Thank you!

jhaiduce:
> Please provide testing code that shows the problem before the fix and passes after the fix.
> Thank you!

Here's some example code that fails on the current development branch and passes once this PR is applied:
```
program h5test
  use hdf5
  use, intrinsic :: iso_fortran_env, only: error_unit

  implicit none

  integer::ierr

  integer(HID_T)::h5file_h
  integer(SIZE_T)::size

  character(80)::filename
  character(*), parameter::expected_filename = "test.h5"

  integer::i

  call h5open_f(ierr)

  call h5fcreate_f(expected_filename, H5F_ACC_TRUNC_F, h5file_h, ierr)

  do i = 1,80
     filename(i:i) = ' '
  end do

  call h5fget_name_f(h5file_h, filename, size, ierr)

  if(size/=len(expected_filename)) then
     write(error_unit,*) 'Wrong filename length'
     error stop
  end if

  if(filename(1:size)/=expected_filename) then
     write(error_unit,*) 'Filename does not match'
     error stop
   end if

  call h5fclose_f(h5file_h, ierr)

end program h5test
```

I also added some changes to this PR that modify tH5F.F90 so that the test subroutine `reopentest` would fail with the previous implementation, but passes with the fix implemented in this PR.

brtnfld (changes requested):
The issue is the majority of Fortran users expect the trailing spaces to be trimmed if they are creating a character string that is larger than the actual filename, having a filename with trailing white space is not common. This also mimics Fortran standard OPEN, which also does not necessarily create a filename with trailing whitespace. 

A better solution would be to add an optional len parameter for the filename. But that is not needed because a user can use C_NULL_CHAR to preserve the trailing white space.

 CHARACTER(*), PARAMETER::expected_filename = "test.h5    "//C_NULL_CHAR

C has a different behavior because the null terminator determines the string size.

brtnfld:
I see the issue, I had read the issue as trailing white space in the filename. Let me look into it.

fortran/src/H5Fff.F90
```diff
@@ -844,7 +844,7 @@ INTEGER FUNCTION h5fget_name_c(obj_id, size, buf, buflen) &
          CHARACTER(KIND=C_CHAR), DIMENSION(*), INTENT(OUT) :: buf
        END FUNCTION h5fget_name_c
     END INTERFACE
-    buflen = LEN_TRIM(buf)
+    buflen = LEN(buf)
```
brtnfld, line 847: Also, change the declaration (line 827):

CHARACTER(LEN=*), INTENT(OUT) :: buf

fortran/test/tH5F.F90
```diff
@@ -532,13 +532,24 @@ SUBROUTINE reopentest(cleanup, total_error)
...
          CALL h5fget_name_f(dset_id, file_name, name_size, error)
               CALL check("h5fget_name_f",error,total_error)
-              IF(file_name(1:name_size) .NE. fix_filename(1:name_size)) THEN
+              IF(name_size /= 9) THEN
+                 write(*,*) "file name obtained from the dataset id has wrong length"
+                 call check("h5fget_name_f",-1,total_error)
+              END IF
+              IF(file_name(1:name_size) /= fix_filename(1:name_size)) THEN
                  write(*,*) "file name obtained from the dataset id is incorrect"
+                 call check("h5fget_name_f",-1,total_error)
```
brtnfld: I would leave in the old test and add a new test:

         CALL h5dopen_f(reopen_id, dsetname, dset_id, error)
              CALL check("h5dopen_f",error,total_error)
         !
         !Get file name from the dataset identifier
         !
         ! Use an uninitialized buffer
         CALL h5fget_name_f(dset_id, file_name, name_size, error)
         CALL check("h5fget_name_f",error,total_error)
         IF(name_size .NE. LEN_TRIM(fix_filename))THEN
            WRITE(*,*) "  file name size obtained from the dataset id is incorrect"
            total_error = total_error + 1
         ENDIF
         IF(file_name(1:name_size) .NE. TRIM(fix_filename)) THEN
            WRITE(*,*) "  file name obtained from the dataset id is incorrect"
            total_error = total_error + 1
         END IF
         ! Use an initialized buffer
         file_name(:) = " "
         CALL h5fget_name_f(dset_id, file_name, name_size, error)
         CALL check("h5fget_name_f",error,total_error)
         IF(name_size .NE. LEN_TRIM(fix_filename))THEN
            WRITE(*,*) "  file name size obtained from the dataset id is incorrect"
            total_error = total_error + 1
         ENDIF
         IF(file_name(1:name_size) .NE. TRIM(fix_filename)) THEN
            WRITE(*,*) "  file name obtained from the dataset id is incorrect"
            total_error = total_error + 1
         END IF

         !
         !Read the dataset.
         !

brtnfld: The size should always be correct, just the filename would be wrong since the buffer size given was zero.

brtnfld (changes requested):

fortran/test/tH5F.F90
jhaiduce: > I would leave in the old test and add a new test

I pushed some changes reverting my edits to reopentest and adding a new test focusing on h5fget_name_f.

fortran/test/tH5F.F90
```diff
@@ -580,6 +580,155 @@ SUBROUTINE reopentest(cleanup, total_error)
...
+          !
+          !Create data space for the dataset.
+          !
+          CALL h5screate_simple_f(RANK, dims, dataspace, error)
+               CALL check("h5screate_simple_f",error,total_error)
+
+          !
+          !Create dataset "/dset" inside the file .
+          !
+          CALL h5dcreate_f(file_id, dsetname, H5T_NATIVE_INTEGER, dataspace, &
```
brtnfld, line 708: I would suggest simplifying this by just opening the root group "/" and using the group id instead of going through the trouble of creating a dataset. 
call h5gopen_f(file_id,"/",g_id,ierr)

brtnfld (commented):
@jhaiduce, should I make the changes and issue a new PR?

jhaiduce:
> @jhaiduce, should I make the changes and issue a new PR?

Sounds good to me, thanks!

fortran/src/H5Fff.F90
brtnfld: fixed in new PR.

brtnfld:
Reworked PR, new PR  #972

PR #832, opened 2021-07-14
Add support for parallel filters to h5repack

jhendersonHDF:
This is related to the dynamic loading of VFDs feature, and is a nice feature to have in h5repack anyway. Simply sets collective dataset writes when filters are involved in the repack operation and the output file is using the MPIO file driver.

### DIFF
--- a/tools/src/h5repack/h5repack_copy.c
+++ b/tools/src/h5repack/h5repack_copy.c
@@ -626,6 +626,7 @@ do_copy_objects(hid_t fidin, hid_t fidout, trav_table_t *travt, pack_opt_t *opti
     hid_t              wtype_id      = H5I_INVALID_HID; /* read/write type ID */
     hid_t              ocpl_id       = H5I_INVALID_HID; /* property to pass copy options */
     hid_t              lcpl_id       = H5I_INVALID_HID; /* link creation property list */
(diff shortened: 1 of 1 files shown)


qkoziol:
Is h5repack still being run from only one MPI process?  (Or without using "mpirun ...")

byrnHDF (approved):

jhendersonHDF:
> Is h5repack still being run from only one MPI process? (Or without using "mpirun ...")

In general, yes. This is mostly just to help with the fact that the parallel filters feature doesn't currently acknowledge a difference between running with one MPI rank and multiple. It just always requires collective I/O. However, this also might be useful if we start making more of HDF5's tools parallel-capable. I believe h5repack has been specifically called out as desirable for that.

lrknox (approved):

qkoziol:
> > Is h5repack still being run from only one MPI process? (Or without using "mpirun ...")
> 
> In general, yes. This is mostly just to help with the fact that the parallel filters feature doesn't currently acknowledge a difference between running with one MPI rank and multiple. It just always requires collective I/O. However, this also might be useful if we start making more of HDF5's tools parallel-capable. I believe h5repack has been specifically called out as desirable for that.

OK, sounds reasonable.  (And already merged :-) )

PR #835, opened 2021-07-14
OESS-168:Remove clang warnings.

hyoklee:
I removed some clang warnings.

### DIFF
--- a/hl/src/H5LTanalyze.c
+++ b/hl/src/H5LTanalyze.c
@@ -1947,7 +1947,7 @@ static int yy_get_next_buffer (void)
 					yyrestart( yyin );
 
 					/*FALLTHROUGH*/
-
+                                        __attribute__((fallthrough)); 
 				case EOB_ACT_END_OF_FILE:
 					{
 					if ( yywrap(  ) )
@@ -1961,6 +1961,7 @@ static int yy_get_next_buffer (void)
 					return input();
 #endif
 					}
+
 
 				case EOB_ACT_CONTINUE_SCAN:
 					(yy_c_buf_p) = (yytext_ptr) + offset;
--- a/hl/src/H5LTanalyze.c
+++ b/hl/src/H5LTanalyze.c
@@ -1947,7 +1947,7 @@ static int yy_get_next_buffer (void)
 					yyrestart( yyin );
 
 					/*FALLTHROUGH*/
-                                        __attribute__((fallthrough)); 
+                                         
 				case EOB_ACT_END_OF_FILE:
 					{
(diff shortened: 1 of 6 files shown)


hl/src/H5LTanalyze.c
```diff
@@ -1947,7 +1947,7 @@ static int yy_get_next_buffer (void)
 					yyrestart( yyin );
 
 					/*FALLTHROUGH*/
-
+                                        __attribute__((fallthrough));
```
byrnHDF: This needs to be changed in the source file - this file is generated.

byrnHDF (changes requested):

hl/test/test_ds.c
```diff
@@ -5210,6 +5210,8 @@ test_attach_detach(void)
 
     HL_TESTING2("permutations of attaching and detaching");
 
+    gid = var1_id = var2_id = var3_id = -1;
```
byrnHDF: We should be encouraging that -1 should be H5I_INVALID_HID for hid_t variables.

byrnHDF (changes requested):

qkoziol (approved):

byrnHDF (approved):

hl/test/test_packet.c
```diff
@@ -45,10 +45,10 @@ typedef struct particle_t {
  * a static array of particles for writing and checking reads
  *-------------------------------------------------------------------------
  */
-static particle_t testPart[NRECORDS] = {{"zero", 0, 0, 0.0f, 0.0f},    {"one", 10, 10, 1.0f, 10.0f},
-                                        {"two", 20, 20, 2.0f, 20.0f},  {"three", 30, 30, 3.0f, 30.0f},
-                                        {"four", 40, 40, 4.0f, 40.0f}, {"five", 50, 50, 5.0f, 50.0f},
-                                        {"six", 60, 60, 6.0f, 60.0f},  {"seven", 70, 70, 7.0f, 70.0f}};
+static particle_t testPart[NRECORDS] = {{"zero", 0, 0, 0.0f, 0.0},    {"one", 10, 10, 1.0f, 10.0},
```
byrnHDF, line 48: Just noticed the "f" is still on 4th value in the arrays - shouldn't it match the change to the 5th value?

byrnHDF (commented):
Better - just one new question.

hl/test/test_packet.c
hyoklee: 4th value is for float testing. 5th value is for double testing. clang doesn't complain for 4th value.

PR #837, opened 2021-07-14
Update version to 1.10.8-2 after snapshot release of 1.10.8-1

lrknox:


lrknox:
Something is wrong here - shouldn't have changed more than 6 files.

PR #840, opened 2021-07-15
Allow parallel filters feature for comm size of 1

jhendersonHDF:


qkoziol (approved):
No objections, but there's a whole _host_ of places where a communicator of size 1 should/could behave differently from a communicator with >1 ranks.

jhendersonHDF:
> No objections, but there's a whole _host_ of places where a communicator of size 1 should/could behave differently from a communicator with >1 ranks.

Would this be cases like initializing with MPI_Init_thread()?

EDIT: There are two other places in the library where we check for a comm_size > 1 and then do something in parallel if we're just using a comm_size of 1. One is just an optimization in H5Fsuper.c for reading the superblock from only 1 rank, so I don't think that one really matters if the communicator behaves differently from a communicator with >1 ranks. The other is in the page buffer code where we DON'T bypass the page buffer in parallel if comm_size is 1. In that case, I think this also might have the same issue? However, that code is currently #ifdef'ed out because apparently that check stopped working at some point. I think the main takeaway is, if checking for comm_size of 1 isn't good enough here, we probably shouldn't merge this or should add extra checks here and maybe in the page buffer code as well.

lrknox (approved):

PR #848, opened 2021-07-21
Avoid popping API context when one wasn't pushed

jhendersonHDF:
This PR avoids popping an API context when one wasn't pushed, which is usually the case when library initialization fails. If library initialization fails, the API context push will be skipped and then when H5CX_pop is hit inside the FUNC_LEAVE_ macros, an assertion will be triggered because it expects a context to have been pushed.

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -2082,6 +2082,8 @@ H5_DLL herr_t H5CX_pop(hbool_t update_dxpl_props);
     /* Push the API context */                                                                               \
     if (H5CX_push() < 0)                                                                                     \
         HGOTO_ERROR(H5E_FUNC, H5E_CANTSET, err, "can't set API context")                                     \
+    else                                                                                                     \
+        api_ctx_pushed = TRUE;                                                                               \
                                                                                                              \
     BEGIN_MPE_LOG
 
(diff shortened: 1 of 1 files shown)


src/H5private.h
```diff
@@ -2082,13 +2082,17 @@ H5_DLL herr_t H5CX_pop(hbool_t update_dxpl_props);
...
+    else                                                                                                     \
+        api_ctx_pushed = TRUE;                                                                               \
                                                                                                              \
     BEGIN_MPE_LOG
 
 /* Use this macro for all "normal" API functions */
 #define FUNC_ENTER_API(err)                                                                                  \
     {                                                                                                        \
         {                                                                                                    \
+            hbool_t api_ctx_pushed = FALSE;                                                                  \
```
jhendersonHDF, line 2094: While I'd like to have this variable restricted to the scope of FUNC_ENTER_API_PUSH, GCC gives false positives about possible use of an uninitialized variable unless it's at this broader scope. I suspect a ton of macro refactoring would be needed to straighten GCC's situation out.

src/H5private.h
```diff
@@ -2369,14 +2375,17 @@ H5_DLL herr_t H5CX_pop(hbool_t update_dxpl_props);
     H5_API_SET_CANCEL
 
 #define FUNC_LEAVE_API_COMMON(ret_value)                                                                     \
-    ;                                                                                                        \
-    } /*end scope from end of FUNC_ENTER*/                                                                   \
```
jhendersonHDF, line 2373: This doesn't belong here, as FUNC_ENTER_API_COMMON doesn't leave the trailing opening brace, it's the individual FUNC_ENTER_ macros that do. Including the brace here can mess up the takedown ordering inside the FUNC_LEAVE_ macros, so the changes here move back to duplicating the closing brace in each FUNC_LEAVE_ macro rather than trying to keep it in a common place.

byrnHDF (approved):

derobins (approved):

PR #849, opened 2021-07-22
Fixes a couple of warnings

derobins:
- warn_unused_result warnings from h5jamgentest and h5stat_gentest
   that were triggered by write(2) calls where we didn't check the
   return value
 - A missing callback in the NULL VOL connector
 - A possibly uninitialized property list pointer (error case only)


derobins:
I updated the generators to check error codes, in general, so there's a lot more changes in those files than just the changes around the write(2) calls, though. The overall logic didn't change, though - just a lot of "if (badness) goto error" markup.

jhendersonHDF (approved):

byrnHDF (approved):

PR #866, opened 2021-07-29
Added int variable for comparison with H5_VERS_RELEASE in H5.c to avoid

lrknox:
warning that unsigned comparison < 0 is always false, which is known,
        but for later versions the comparison can possibly be true.

### DIFF
--- a/src/H5.c
+++ b/src/H5.c
@@ -912,6 +912,8 @@ H5check_version(unsigned majnum, unsigned minnum, unsigned relnum)
     static unsigned int disable_version_check    = 0; /* Set if the version check should be disabled */
     static const char * version_mismatch_warning = VERSION_MISMATCH_WARNING;
     herr_t              ret_value                = SUCCEED; /* Return value */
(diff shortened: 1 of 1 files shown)


src/H5.c
```diff
@@ -912,6 +912,8 @@ H5check_version(unsigned majnum, unsigned minnum, unsigned relnum)
     static unsigned int disable_version_check    = 0; /* Set if the version check should be disabled */
     static const char * version_mismatch_warning = VERSION_MISMATCH_WARNING;
     herr_t              ret_value                = SUCCEED; /* Return value */
+    int                 releasenum               = relnum;  /* To avoid warning for unsigned < 0
```
byrnHDF, line 915: This does not make sense to me. Basically your're setting an int from an unsigned int?
What is H5_VERS_RELEASE defined as? Would a cast work? H5_VERS_RELEASE > (int)relnum

derobins (approved):

jhendersonHDF (approved):

PR #872, opened 2021-08-03
Removes partial long options from the tools

derobins:
Some of the command-line tools would allow partial long options (e.g.,
--datas instead of --dataset). These were inconsistently implemented,
difficult to maintain, and occasionally blocked useful long options.
They have been removed from all the tools.

NOTE: This change should NOT be merged to 1.12 or earlier.


tools/src/h5stat/h5stat.c
```diff
@@ -172,102 +172,19 @@ struct handler_t {
 static const char *s_opts = "Aa:Ddm:EFfhGgl:sSTO:Vw:H:";
```
byrnHDF, line 172: I doubt more options will be added, but these longer lists should be in alpha order by short letter, makes it easier to know which short character was used. Long term it would be nice to change the option parsing function.

byrnHDF (approved):

lrknox (approved):

PR #873, opened 2021-08-03
Removes the MANIFEST and associated scripts

derobins:
The MANIFEST is a pain to maintain and basically useless. There are other ways to exclude files from a tarball and the utility of excluding a handful of files that are in version control is dubious.


derobins:
I've left this as a draft while we discuss.

derobins:
I removed bbrelease as it doesn't seem to work with the --revision option, which is its main difference from release.

bin/release
```diff
@@ -591,8 +542,7 @@ fi
 test "$verbose" && echo "   Running tar..." 1>&2
 ( \
     cd $tmpdir; \
-    tar cf $HDF5_VERS.tar $HDF5_VERS/Makefile \
-    `sed 's/^\.\//hdf5-'$VERS'\//' $MANIFEST` || exit 1 \
+    tar --exclude-vcs -cvvf $HDF5_VERS.tar $HDF5_VERS
```
derobins, line 545: It looks like --exclude-vcs is a gnu extension. I'll probably switch this to use the more portable --exclude option to avoid packing the .git directory.

lrknox:
> I removed bbrelease as it doesn't seem to work with the --revision option, which is its main difference from release.

The --revision option in bbrelease is used for buildbot nightly tests, for these reasons:
1. It records the revision number in the source [tar,zip]file and also in the version of the source tested.  This revision is available in buildbot logs and possibly also in CDash.
2. it gives a generic name to the top-level directory inside the tarfiles, simplifying builbot configuration.
3. Releasing a tarfile for testing at the beginning of the nightly test run insures that all configurations for the nightly test run are testing the same revision.

We need to keep or replace the --revision option, but it can be moved to the release script.

derobins:
> > I removed bbrelease as it doesn't seem to work with the --revision option, which is its main difference from release.
> 
> The --revision option in bbrelease is used for buildbot nightly tests, for these reasons:
> 
> 1. It records the revision number in the source [tar,zip]file and also in the version of the source tested.  This revision is available in buildbot logs and possibly also in CDash.
> 2. it gives a generic name to the top-level directory inside the tarfiles, simplifying builbot configuration.
> 3. Releasing a tarfile for testing at the beginning of the nightly test run insures that all configurations for the nightly test run are testing the same revision.
> 
> We need to keep or replace the --revision option, but it can be moved to the release script.

How does it work? Even from develop (before I've touched it), `bbrelease --revision` fails to create a tarball. I'd be happy to merge the two scripts, but that's going to be challenging if I can't run the script successfully.

lrknox:
Your system is Ubuntu?  Or something else?  It works on jelly for me and wherever buildbot is running it.  Maybe if I try running it somewhere that it fails I can find the problem.

Larry

derobins:
My usual system is Ubuntu 20.04 LTS

lrknox:
I tried on Ubuntu 20.04.2 LTS installed on windows10, with no problem.
In a separate chat, the output from running "bin/bbrelease -d .. --revision" on your Ubuntu 20.04 included "Releasing hdf5-canonical/develop)-5dabec4cf6 to ..".  The part of the string "hdf5-canonical/develop)-5dabec4cf6" between the '-'s comes from "branch=`git branch | grep '*' | awk '{print $NF}'`".  The output of "git branch" in my instance was:
* develop
  merge_839_hdf5_1_10

and the corresponding output was "Releasing hdf5-develop-725da3768b to ./hdf5_release".  What output do you have from "git branch"?

From experimentation, I assume your current branch is canonical/develop, shown as "* (HEAD detached at canonical/hdf5_1_12)"?

If so, the problem here, which we hadn't previously encountered, is the '/' in hdf5-canonical/develop)-5dabec4cf6, which goes into the name of the tar file hdf5-canonical/develop)-5dabec4cf6.tar, but then can't be found, I suspect because the '/' is interpreted as a directory separator instead of as a part of a filename.  Extending this line (279 or 280):

branch=`git branch | grep '*' | awk '{print $NF}'`

to

branch=`git branch | grep '*' | awk '{print $NF}' | sed 's?/?-?g'`

will replace any '/'s with '-' and tar (or the script) won't be confused.  We tried escaping, but \\\\/ was still treated as /.

derobins:
I'm closing this for now. I'll probably update it an resubmit when I have more time.

PR #887, opened 2021-08-10
Noflush feature vfd swmr

vchoi-hdfgroup:


test/vfd_swmr_attrdset_writer.c
```diff
@@ -223,12 +227,20 @@ state_init(state_t *s, int argc, char **argv)
...
+
+    if (H5_basename(argv[0], &tfile) < 0) {
+        HDprintf("H5_basename failed\n");
+        TEST_ERROR;
+    }
+
+    esnprintf(s->progname, sizeof(s->progname), "%s", tfile);
+
+    if (tfile)
+        HDfree(tfile);
```
kyang2014, line 243: To be safe, this block may be written like this:
if(tfile) {
   HDfree(tfile);
   tfile=NULL;
}
This may avoid memory free error when tfile is free and something bad happens in the following code, then the code goes to the error: , tfile is free again.

test/vfd_swmr_attrdset_writer.c
```diff
@@ -1835,137 +2029,18 @@ main(int argc, char **argv)
...
-
-                result = attr_dsets_action(MODIFY_ATTR, &s, &ds, step);
-
-                if (s.use_np && !np_writer(result, step, &s, &np, &config)) {
-                    HDprintf("np_writer() for modification failed\n");
-                    TEST_ERROR;
-                }
-            }
+         if(!perform_dsets_operations(&s, &ds, &config, &np)) {
+            printf("perform_dsets_operations() failed\n");
```
kyang2014: Should we use HDprintf here?

test/vfd_swmr_attrdset_writer.c
```diff
@@ -1835,137 +2029,18 @@ main(int argc, char **argv)
...
-
-                result = verify_attr_dsets_action(MODIFY_ATTR, &s, &ds, step);
-
-                if (s.use_np && !np_reader(result, step, &s, &np)) {
-                    HDprintf("np_reader() for verifying modification failed\n");
-                    TEST_ERROR;
-                }
-            }
+         if(!verify_dsets_operations(&s, &ds, &config, &np)) {
+            printf("verify_dsets_operations() failed\n");
```
kyang2014: HDprintf here?

test/vfd_swmr_dsetchks_writer.c
```diff
@@ -263,18 +265,22 @@ state_init(state_t *s, int argc, char **argv)
...
         TEST_ERROR
+    }
 
-    *s = ALL_HID_INITIALIZER;
-    esnprintf(tfile, sizeof(tfile), "%s", argv[0]);
-    esnprintf(s->progname, sizeof(s->progname), "%s", base);
+    esnprintf(s->progname, sizeof(s->progname), "%s", tfile);
 
-    while ((ch = getopt(argc, argv, "siferom:n:x:y:g:p:t:l:bqSNu:c:")) != -1) {
+    if (tfile)
```
kyang2014: Add tfile = NULL after HDfree

test/vfd_swmr_dsetchks_writer.c
```diff
@@ -1439,19 +1451,22 @@ verify_dsets_operations(state_t *s, dsets_state_t *ds, H5F_vfd_swmr_config_t *co
         if (s->pwrites || s->lwrites) {
             dbgf(2, "Verify single partial chunk write to dataset with single index; only verify 1 write\n");
 
-            if (s->use_np && !np_confirm_verify_notify(np->fd_writer_to_reader, 0, s, np)) {
+            if(s->use_np && !np_confirm_verify_notify(np->fd_writer_to_reader, 0, s, np)) {
                 printf("np_confirm_verify_notify() verify/notify not in sync failed\n");
```
kyang2014: May use HDprintf? The same for all the printf()

test/vfd_swmr_group_writer.c
```diff
@@ -4985,7 +4985,7 @@ main(int argc, char **argv)
     }
 
     /* config, tick_len, max_lag, writer, flush_raw_data, md_pages_reserved, md_file_path */
-    init_vfd_swmr_config(&config, 4, 7, writer, FALSE, 128, "./group-shadow");
+    init_vfd_swmr_config(&config, 4, 7, writer, TRUE, 128, "./group-shadow");
```
kyang2014, line 4988: A question, here we only test the option when the flush_raw_data is TRUE, what about when the flush_raw_data is FALSE? This option is not checked for this test. I think some other tests also haven't checked this option.

vchoi-hdfgroup: It's up to individual test to add an option to flush/not flush raw data and then input that to init_vfd_swmr_config().  I added "not flush" to two tests here for testing: dsetops and dsetchks

kyang2014 (approved):
My other comments are minor. I don't need to review again.

raylu-hdf (approved):

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: Fixed.

test/vfd_swmr_attrdset_writer.c
vchoi-hdfgroup: All printfs changed to HDprintfs.

PR #907, opened 2021-08-13
Merge3 839 hdf5 1 12

lrknox:


hl/test/test_packet.c
```diff
@@ -45,10 +45,10 @@ typedef struct particle_t {
  * a static array of particles for writing and checking reads
  *-------------------------------------------------------------------------
  */
-static particle_t testPart[NRECORDS] = {{"zero", 0, 0, 0.0f, 0.0f},    {"one", 10, 10, 1.0f, 10.0f},
-                                        {"two", 20, 20, 2.0f, 20.0f},  {"three", 30, 30, 3.0f, 30.0f},
-                                        {"four", 40, 40, 4.0f, 40.0f}, {"five", 50, 50, 5.0f, 50.0f},
-                                        {"six", 60, 60, 6.0f, 60.0f},  {"seven", 70, 70, 7.0f, 70.0f}};
+static particle_t testPart[NRECORDS] = {{"zero", 0, 0, 0.0f, 0.0},    {"one", 10, 10, 1.0f, 10.0},
```
byrnHDF, line 48: Just noticed another "f" in the 4th value, I think it should be removed like the 5th value?

byrnHDF (approved):

hl/test/test_packet.c
lrknox: No, the fourth element in the structure is a float, the fifth is a double, and the numbers with .0f for the fifth were single precision values implicitly converted to double precision.  Removing the f from the fifth value eliminated that warning;  removing it from the fourth would probably add the opposite warning  (for clang).

PR #908, opened 2021-08-13
More clang tidy

seanm:


hl/examples/ex_table_04.c
```diff
@@ -65,23 +65,23 @@ main(void)
     hid_t     string_type;
     hid_t     file_id;
     hsize_t   chunk_size   = 10;
-    Particle  fill_data[1] = {{"no data", -1, -1, -99.0f, -99.0}}; /* Fill value particle */
+    Particle  fill_data[1] = {{"no data", -1, -1, -99.0F, -99.0}}; /* Fill value particle */
```
byrnHDF, line 68: Didn't we just change all these to lower case "f" earlier this year?

seanm: I don't know. I don't follow HDF5 closely. The clang-tidy docs are [here](https://clang.llvm.org/extra/clang-tidy/checks/readability-uppercase-literal-suffix.html) but they don't really give a rationale.  I think it's mostly for `l` which can look a lot like `1`. I guess the idea is to consistently use uppercase. I can drop this commit, I don't have strong feelings about it for `f` vs `F`, but for `l` vs `L` I do think uppercase is better.

byrnHDF: I am not entirely sure myself, but I think it was a gcc warning issue.

seanm: I've just checked the `develop` branch, and there is a mix of `f` and `F` currently.

byrnHDF: Unsurprising.

seanm: So using `readability-uppercase-literal-suffix` as per this patch would have the advantage of making it uniform, even though `F` vs `f` is not as useful as `l` vs `L`.

byrnHDF: We did a quick check and it seems we have no standard for F vs f.  We will need to decide so we don't keep churning.

byrnHDF (approved):

lrknox (approved):

hl/examples/ex_table_04.c
derobins: We have no standard, but I'm fine with upper case F being the standard, especially given the l vs L confusion.

derobins (approved):

PR #922, opened 2021-08-18
Improvement/data transform

jwsblokland:
Hello,

Recently, I madea small improvement for H5Z_xform_noop() function. Now, the function will also return TRUE if the data transform function expression = "x". For this case, the HDF5 library behaves similar fashion as the case when no data transform function has been specified. More information can be found [this thread](https://forum.hdfgroup.org/t/phdf5-1-12-trivial-data-transform-function-expression-x-and-optimized-parallel-i-o-routines/8854) on the forum.

Furthermore, I improved the robustness of CMake ph5example and testswmr test when running ctest.

Best regards,
Jan-Willem


qkoziol (changes requested):
The change to the src/H5Ztrans.c is fine, but it seems like the other pieces are out of date?

qkoziol:
> The change to the src/H5Ztrans.c is fine, but it seems like the other pieces are out of date?

Or possibly, those are correct, but probably should be in a separate PR?

jwsblokland:
Splitting this pull request in two is actually a good idea. I will do this and close this pull request.

PR #933, opened 2021-08-19
H5Ztrans: (feature) Improved H5Z_xform_noop() and H5Z_xform_create() …

jwsblokland:
Improved H5Z_xform_noop() and H5Z_xform_create() function

- Made a small improvement for H5Z_xform_noop() function. Now,
  the function will also return TRUE if the data transform function
  expression = "x". For this case, the HDF5 library behaves in a
  similar fashion as the case when no data transform function has been
  specified.
- Improved the inline documentation of the function
  H5Z_xform_create() such it is more inline with the rest of the
  code.

This pull request has been requested in the abandoned pull request https://github.com/HDFGroup/hdf5/pull/922.

### DIFF
--- a/testpar/t_filters_parallel.c
+++ b/testpar/t_filters_parallel.c
@@ -1367,8 +1367,8 @@ test_write_transformed_filtered_dataset_no_overlap(void)
     /* Add test filter to the pipeline */
     VRFY((set_dcpl_filter(plist_id) >= 0), "Filter set");
 
-    dset_id = H5Dcreate2(file_id, WRITE_UNSHARED_TRANSFORMED_FILTERED_CHUNKS_DATASET_NAME, HDF5_DATATYPE_NAME, filespace,
-                         H5P_DEFAULT, plist_id, H5P_DEFAULT);
+    dset_id = H5Dcreate2(file_id, WRITE_UNSHARED_TRANSFORMED_FILTERED_CHUNKS_DATASET_NAME, HDF5_DATATYPE_NAME,
+                         filespace, H5P_DEFAULT, plist_id, H5P_DEFAULT);
     VRFY((dset_id >= 0), "Dataset creation succeeded");
 
     VRFY((H5Pclose(plist_id) >= 0), "DCPL close succeeded");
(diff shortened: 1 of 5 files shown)


qkoziol (changes requested):
This code looks fine, but it needs a test, to demonstrate that it actually works.  Adding it to both the serial testing for data transforms (in the 'test' directory) and the parallel testing (in the 'testpar' directory) is probably necessary, since I believe that use in parallel HDF5 is desired.

qkoziol:
> This code looks fine, but it needs a test, to demonstrate that it actually works. Adding it to both the serial testing for data transforms (in the 'test' directory) and the parallel testing (in the 'testpar' directory) is probably necessary, since I believe that use in parallel HDF5 is desired.

Also, with a test, you'll guarantee that it stays working. :-)

jwsblokland:
As requested I have added 1 serial test and 2 parallel tests. I must admit usually I add some tests to demostrates the feature or bug fix works as expected. Apparently, this case I did not, so thank you for reminding me to add these tests.

jhendersonHDF (approved):

lrknox (approved):

testpar/t_filters_parallel.c
```diff
@@ -1287,6 +1291,177 @@ test_write_filtered_dataset_interleaved_write(void)
...
+ * particular chunk in the operation. Normally, a data
+ * transform function will cause the parallel library to
+ * break to independent I/O and this isn't allowed when
+ * there are filters in the pipeline. However, in this
+ * case the parallel library recognizes that the used
+ * data transform function "x" is the same as not applying
+ * the transform function. Therefore it does not apply
+ * the transform function resulting in not breaking to
+ * independent I/O.
+ *
```
epourmal, line 1306: Please add the comment to the RELEASE.txt file (similar to what you have here) explaining the change. It should go under Bug Fixed since HDF5 ... / Library.
Thank you!

jwsblokland: @epourmal , I have added a comment about the implemented improvement to the `RELEASE.txt` file. Thanks for mentioning this.

epourmal: Thank you! We are currently testing on HPC systems now and will accept the change.

lrknox:
The tests added also passed on jelly, lassen quartz and mutrino.

PR #934, opened 2021-08-19
CMake bug fix for ph5example and testswmr test

jwsblokland:
Hello,

While running ctest using the parallel version of HDF5 library I encountered 2 little bugs. The bugs, described below, has been fixed.

ph5example test
Ensure that the used number of processes times an integer number
is equal to SPACE1_DIM1 and SPACE_DIM2. These variables are
equal to 24 and are defined in ph5example.c. If this is not the
case the ph5example test will fail.

testswmr test
Adjusted POST_BUILD command of various swmr_* targets. Instead of
copying the files to the directory ${HDF5_TEST_BINARY_DIR}/H5TEST/swmr_test
they are now copied to ${HDF5_TEST_BINARY_DIR}/H5TEST directory.
This is more inline with the associated script testswmr.sh and allows
to run the SWMR tests multiple times without rebuilding the HDF5 library.

This pull request has been requested in the abandoned pull request #922.

Best regards,
   Jan-Willem


jwsblokland:
I just noticed that byrnHDF in pull request has fixed the bug in pull request https://github.com/HDFGroup/hdf5/pull/928, therefore I will close this pull request and yet create another one in which I only fix the ph5example test.

PR #953, opened 2021-08-23
1.10 cmake: simplify try_run result comparisons (#950)

lrknox:


byrnHDF (approved):

derobins:
Should probably stop using personal develop, hdf5_1_10 branches as the imported change logs make navigating 'git log' difficult.

derobins (approved):

PR #967, opened 2021-08-26
unlock file before closing

dschwoerer:
Unlocking ensures that if a new thread has spawned and thus there are
still copies of the fd around, the file get's still unlocked and can be
read again.

### DIFF
--- a/src/H5FDsec2.c
+++ b/src/H5FDsec2.c
@@ -457,6 +457,9 @@ H5FD__sec2_close(H5FD_t *_file)
     /* Sanity check */
     HDassert(file);
 
+    /* ensure we don't keep a lock, even if the fd has been duplicated */
+    HDflock(file->fd, LOCK_UN | LOCK_NB);
+
     /* Close the underlying file */
     if(HDclose(file->fd) < 0)
         HSYS_GOTO_ERROR(H5E_IO, H5E_CANTCLOSEFILE, FAIL, "unable to close file")


derobins:
This needs a test. Probably something in test/swmr.c, where we already have a test_file_locking() test that could be expanded.

lrknox:
Please note also that the 1.10/master branch is used exclusively for HDF5 1.10 releases.  Pull requests should preferably go to the develop branch first, or for changes appropriate for only HDF5 1.10, to the hdf5_1_10 branch.

dschwoerer:
> This needs a test. Probably something in test/swmr.c, where we already have a test_file_locking() test that could be expanded.

Something like:
open file (with locking), fork, close in one thread, try to open again, ensure it is open-able, delete other thread?

dschwoerer:
> Please note also that the 1.10/master branch is used exclusively for HDF5 1.10 releases. Pull requests should preferably go to the develop branch first, or for changes appropriate for only HDF5 1.10, to the hdf5_1_10 branch.

Thanks for the info, I will rebase to develop ...

derobins:
> > This needs a test. Probably something in test/swmr.c, where we already have a test_file_locking() test that could be expanded.
> 
> Something like: open file (with locking), fork, close in one thread, try to open again, ensure it is open-able, delete other thread?

Something like that, but keep in mind that Windows doesn't have fork. I can map a test that does fork + exec to Windows concepts, but the "fork then do one thing if you are the parent and another if you are the child" pattern is not a thing on Windows.

PR #968, opened 2021-08-26
Fixed failure on BE with xlf

brtnfld:
Fixed segfault with xlf on BE with -qintsize=8.

This is a workaround to a compiler bug since the failing test does not get executed when szip is not enabled (which was the failing case). \Just referencing the data_out array anywhere in the code caused a segfault.


lrknox (approved):

lrknox:
passes all tests on echidna with this change.

byrnHDF:
Release.txt entry needed? Should this be applied to other branches?

epourmal (approved):

brtnfld:
This is a test fix, so there is no need for Release.txt; if the other branches fail, it needs to be applied to them.

PR #972, opened 2021-08-27
Rework of PR #826

brtnfld:
Made changes suggested in review:

H5fget_name_f len_trim fix #826


brtnfld:
@jhaiduce, let me know if I missed something. Thanks.

jhaiduce:
> @jhaiduce, let me know if I missed something. Thanks.

Looks good to me. It looks like the biggest difference from PR #826 is the change from using a dataset to a group in get_name_test. But that distinction probably doesn't matter for the problem identified in issue #825.

Not sure if this matters to anyone, but there is trailing whitespace on Line 24 of fortran/test/tH5F.F90.

byrnHDF:
Does this need a Release.txt entry?

byrnHDF:
Can this change be applied to 1.12, 1.10, or 1.8?

lrknox:
Configurations with xlf and -qintsize=8 fail the same test in 1.12, 1.10 and 1.8;  change is needed for all branches.

epourmal (approved):

lrknox (approved):

brtnfld:
> Can this change be applied to 1.12, 1.10, or 1.8?

Yes.

PR #981, opened 2021-08-30
VFD ctl feature

jhendersonHDF:
This PR introduces the new VFD "ctl" callback feature, as designed by @jrmainzer. The new callback allows VFDs to handle relatively arbitrary operations identified by an op code.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -464,17 +464,29 @@ New Features
     Library:
     --------
     - Adds new "ctl" callback to VFD H5FD_class_t structure
+      with the following prototype:
 
-      The newly-added "ctl" callback allows Virtual File
+        herr_t (*ctl)(H5FD_t *file, uint64_t op_code,
+                      uint64_t flags, const void *input,
+                      void **output);
+
+      This newly-added "ctl" callback allows Virtual File
       Drivers to intercept and handle arbitary operations
-      identified by an operation code. The callback accepts
-      a pointer to the file to be operated on (H5FD_t *),
-      an operation code to identify the operation to perform,
-      a set of flags to govern the behavior of the operation
-      and input/output pointers that the VFD may use to
-      receive and return operation parameters/data.
+      identified by an operation code. Its parameters are
+      as follows:
 
-        (JRM - 2021/08/30)
+        `file` [in] - A pointer to the file to be operated on
+        `op_code` [in] - The operation code identifying the
+                         operation to be performed
+        `flags` [in] - Flags governing the behavior of the
+                       operation performed (see H5FDpublic.h
+                       for a list of valid flags)
+        `input` [in] - A pointer to arguments passed to the
+                       VFD performing the operation
+        `output` [out] - A pointer for the receiving VFD to
+                         use for output from the operation
+
+        (JRM - 2021/08/16)
 
     - Change how the release part of version, in major.minor.release is checked
       for compatibility
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -463,6 +463,19 @@ New Features
 
     Library:
     --------
+    - Adds new "ctl" callback to VFD H5FD_class_t structure
+
+      The newly-added "ctl" callback allows Virtual File
+      Drivers to intercept and handle arbitary operations
+      identified by an operation code. The callback accepts
+      a pointer to the file to be operated on (H5FD_t *),
+      an operation code to identify the operation to perform,
+      a set of flags to govern the behavior of the operation
+      and input/output pointers that the VFD may use to
+      receive and return operation parameters/data.
+
+        (JRM - 2021/08/30)
+
     - Change how the release part of version, in major.minor.release is checked
       for compatibility
 
--- a/src/H5FD.c
+++ b/src/H5FD.c
@@ -1688,6 +1688,110 @@ done:
 } /* end H5FD_unlock() */
 
 /*-------------------------------------------------------------------------
+ * Function:    H5FDctl
+ *
+ * Purpose:     Perform a CTL operation.
+ *
+ *              The desired operation is specified by the op_code
+ *              parameter.
+ *
+ *              The flags parameter controls management of op_codes that
+ *              are unknown to the callback
+ *
+ *              The input and output parameters allow op_code specific
+ *              input and output
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  JRM -- 8/3/21
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5FDctl(H5FD_t *file, uint64_t op_code, uint64_t flags, const void *input, void **output)
+{
+    herr_t ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_API(FAIL)
+    H5TRACE5("e", "*#ULUL*x**x", file, op_code, flags, input, output);
+
+    /* Check arguments */
+    if (!file)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "file pointer cannot be NULL")
+
+    if (!file->cls)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "file class pointer cannot be NULL")
+
+    /* Don't attempt to validate the op code.  If appropriate, that will
+     * be done by the underlying VFD callback, along with the input and
+     * output parameters.
+     */
+
+    /* Call private function */
+    if (H5FD_ctl(file, op_code, flags, input, output) < 0)
+        HGOTO_ERROR(H5E_VFL, H5E_FCNTL, FAIL, "VFD ctl request failed")
+
+done:
+
+    FUNC_LEAVE_API(ret_value)
+
+} /* end H5FDctl() */
+
+/*-------------------------------------------------------------------------
+ * Function:    H5FD_ctl
+ *
+ * Purpose:     Private version of H5FDctl()
+ *
+ *              The desired operation is specified by the op_code
+ *              parameter.
+ *
+ *              The flags parameter controls management of op_codes that
+ *              are unknown to the callback
+ *
+ *              The input and output parameters allow op_code specific
+ *              input and output
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:  JRM -- 8/3/21
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5FD_ctl(H5FD_t *file, uint64_t op_code, uint64_t flags, const void *input, void **output)
+{
+    herr_t ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_NOAPI(FAIL)
+
+    /* Sanity checks */
+    HDassert(file);
+    HDassert(file->cls);
+
+    /* Dispatch to driver if the ctl function exists.
+     *
+     * If it doesn't, fail if the H5FD_CTL__FAIL_IF_UNKNOWN_FLAG is set.
+     *
+     * Otherwise, report success.
+     */
+    if (file->cls->ctl) {
+
+        if ((file->cls->ctl)(file, op_code, flags, input, output) < 0)
+
+            HGOTO_ERROR(H5E_VFL, H5E_FCNTL, FAIL, "VFD ctl request failed")
+    }
+    else if (flags & H5FD_CTL__FAIL_IF_UNKNOWN_FLAG) {
+
+        HGOTO_ERROR(H5E_VFL, H5E_FCNTL, FAIL, "VFD ctl request failed (no ctl and fail if unknown)")
+    }
+
+done:
+
+    FUNC_LEAVE_NOAPI(ret_value)
+
+} /* end H5FD_ctl() */
+
+/*-------------------------------------------------------------------------
  * Function:    H5FD_get_fileno
  *
  * Purpose:     Quick and dirty routine to retrieve the file's 'fileno' value
--- a/src/H5FD.c
+++ b/src/H5FD.c
@@ -1782,7 +1782,8 @@ H5FD_ctl(H5FD_t *file, uint64_t op_code, uint64_t flags, const void *input, void
     }
     else if (flags & H5FD_CTL__FAIL_IF_UNKNOWN_FLAG) {
 
-        HGOTO_ERROR(H5E_VFL, H5E_FCNTL, FAIL, "VFD ctl request failed (no ctl and fail if unknown)")
+        HGOTO_ERROR(H5E_VFL, H5E_FCNTL, FAIL,
+                    "VFD ctl request failed (no ctl callback and fail if unknown flag is set)")
     }
 
 done:
--- a/src/H5FDmpi.c
+++ b/src/H5FDmpi.c
@@ -41,25 +41,41 @@
  * Programmer:	Quincey Koziol
  *              Friday, January 30, 2004
  *
+ * Changes:     Reworked function to use the ctl callback so we can get
+ *              rid of H5FD_class_mpi_t.  Since there are no real limits
+ *              on what the ctl callback can do, its file parameter can't
+ *              be constant.  Thus, I had to remove the const qualifier
+ *              on this functions file parameter as well.  Note also the
+ *              circumlocution required to use the ctl callbacks output
+ *              parameter to pass back the rank without introducing
+ *              compiler warnings.
+ *                                             JRM -- 8/13/21
+ *
  *-------------------------------------------------------------------------
  */
 int
-H5FD_mpi_get_rank(const H5FD_t *file)
+H5FD_mpi_get_rank(H5FD_t *file)
 {
-    const H5FD_class_mpi_t *cls;
-
-    int ret_value;
+    const H5FD_class_t *cls;
+    uint64_t            flags    = H5FD_CTL__FAIL_IF_UNKNOWN_FLAG | H5FD_CTL__ROUTE_TO_TERMINAL_VFD_FLAG;
+    int                 rank     = -1;
+    void *              rank_ptr = (void *)(&rank);
+    int                 ret_value;
 
     FUNC_ENTER_NOAPI(FAIL)
 
     HDassert(file);
-    cls = (const H5FD_class_mpi_t *)(file->cls);
+    cls = (const H5FD_class_t *)(file->cls);
     HDassert(cls);
-    HDassert(cls->get_rank); /* All MPI drivers must implement this */
+    HDassert(cls->ctl); /* All MPI drivers must implement this */
 
     /* Dispatch to driver */
-    if ((ret_value = (cls->get_rank)(file)) < 0)
+    if ((cls->ctl)(file, H5FD_CTL__GET_MPI_RANK_OPCODE, flags, NULL, &rank_ptr) < 0)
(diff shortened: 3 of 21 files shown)


src/H5FDmulti.c
```diff
@@ -2068,6 +2071,64 @@ H5FD_multi_delete(const char *filename, hid_t fapl_id)
...
+ *
+ *              At present, this VFD supports no op codes of its own and
+ *              simply passes ctl calls on to each member VFD.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5FD_multi_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void *input, void **output)
```
jhendersonHDF, line 2095: @jrmainzer The terminology for what constitutes a terminal vs passthrough VFD seemed to get a little bit fuzzy for certain VFDs. At first, it seemed like the splitter VFD was really the only passthrough VFD that we currently have. However, the multi VFD also seems like it might fit the bill, so I went ahead and implemented the ctl function here as though the multi VFD were a passthrough VFD. If that's not the case, I'm happy to change it back/remove the ctl callback from the multi VFD.

src/H5FDpublic.h
```diff
@@ -138,6 +138,78 @@
...
+#define H5FD_CTL_OPC_RESERVED 512 /* Opcodes below this value are reserved for library use */
+#define H5FD_CTL_OPC_EXPER_MIN                                                                               \
+    H5FD_CTL_OPC_RESERVED /* Minimum opcode value available for experimental use                             \
+                           */
+#define H5FD_CTL_OPC_EXPER_MAX                                                                               \
+    (H5FD_CTL_OPC_RESERVED + 511) /* Maximum opcode value available for experimental use */
+
+/* ctl function op codes: */
+#define H5FD_CTL__INVALID_OPCODE              0
+#define H5FD_CTL__TEST_OPCODE                 1
```
jhendersonHDF, line 151: Here, I added a reserved op code for testing purposes so that we can have a valid op code to test using a serial test. Otherwise, the only valid op codes applied only to the MPI I/O VFD.

src/H5FDsec2.c
```diff
@@ -1068,3 +1071,52 @@ H5FD__sec2_delete(const char *filename, hid_t H5_ATTR_UNUSED fapl_id)
...
+ *
+ *              At present, no op codes are supported by this VFD.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5FD__sec2_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void H5_ATTR_UNUSED *input,
+               void H5_ATTR_UNUSED **output)
```
jhendersonHDF, line 1097: Though this callback only handles the H5FD_CTL__TEST_OPCODE op code, I figured it would be good to have a ctl callback implemented in the default VFD for testing purposes.

src/H5FDmulti.c
jrmainzer: Actually, using the definition of terminal vs. passthrough VFD in H5FDpublic.h:

Terminal VFD: Lowest VFD in the VFD stack through which all VFD calls pass.  Note that this definition is situational.  For example, the sec2 VFD is typically terminal.  However, in the context of the family file driver, it is not -- the family file driver is the bottom VFD through which all VFD calls pass, and thus it is terminal.

Pass through VFD:  Any VFD that relays all VFD calls (with the possible exception of some non-I/O related calls) to underlying VFD(s).

The multi-file driver is terminal, not pass through.  In contrast splitter is a pass through VFDs.  When and if we implement a proper network VFD (i.e. mirror with R/W instead of just W/O) it will be a pass through VFD as well.

None of this is to say that the above definitions are what we should go with in the long term.  They are very much a first cut, and we should change them if they become inconvenient.

src/H5FDpublic.h
```diff
@@ -138,6 +138,78 @@
...
+#define H5FD_CTL__GET_MPI_COMMUNICATOR_OPCODE 2
+#define H5FD_CTL__GET_MPI_RANK_OPCODE         3
+#define H5FD_CTL__GET_MPI_SIZE_OPCODE         4
+#define H5FD_CTL__NUM_OPCODES                 5 /* must be last */
```
jrmainzer: Should decide if we want to keep H5FD_CTL__NUM_OPCODES.  If memory serves, this #define  is not used at present.  I believe I was thinking of using it to validate op-codes, but I'm not sure the idea makes any sense at present.

src/H5FDmulti.c
jhendersonHDF: This is interesting; is there an example of what API calls the multi driver doesn't pass down? It seems as though it passes all VFD calls down to all of its underlying member VFDs.

jrmainzer: I guess I didn't express this well.  Yes, the multi-file driver will pass all VFD calls down, but none of its underlying VFDs will receive all calls.  Thus the multi-file driver is terminal because all calls pass through it, and no underlying VFD receives all calls. 

Thus for example, in the split file driver use case, metadata I/O related VFD calls are routed to one VFD, and raw data I/O calls to another.

jhendersonHDF: Hmm, based just on the definition from H5FDpublic.h, then yes I can see how we could consider the splitter VFD as being the only passthrough VFD. Intuitively though, it seems to me like VFDs which use the public versions of certain H5FD calls (H5FDopen/H5FDread/H5FDwrite, etc.) could be considered as passthrough VFDs. However, if our current definitions do not consider the multi VFD to be a passthrough VFD, then I should either modify its ctl callback to not handle the "route to terminal VFD" flag, or remove the callback altogether.

jrmainzer: That or change the definition to something more useful.  That said, my take would be to keep the multi file driver ctl function as a stub for later use.  

Note that we will probably develop a more nuanced selection of routing flags.  For example, if we were to use the split file driver, with the onion VFD on the metadata channel, we might want to be able to route a ctl call to the metadata channel to tell the onion VFD to start a new version -- thus giving us a way of rolling back to the last consistent view of metadata in the event of a crash.  I'm sure you can think of other variations.

The bottom line is that this ctl call and its associated flags and op-codes will likely evolve over time.  if so, we will probably find ourselves changing and/or extending our definitions of pass through vs. terminal VFDs.

gnuoyd:
On Wed, Sep 01, 2021 at 12:04:00AM +0000, jrmainzer wrote:
> @jrmainzer commented on this pull request.
> 
> 
> 
> > + *
> + *              The flags parameter controls management of op_codes that
> + *              are unknown to the callback
> + *
> + *              The input and output parameters allow op_code specific
> + *              input and output
> + *
> + *              At present, this VFD supports no op codes of its own and
> + *              simply passes ctl calls on to each member VFD.
> + *
> + * Return:      Non-negative on success/Negative on failure
> + *
> + *-------------------------------------------------------------------------
> + */
> +static herr_t
> +H5FD_multi_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void *input, void **output)
> 
> I guess I didn't express this well.  Yes, the multi-file driver will pass all VFD calls down, but none of its underlying VFDs will receive all calls.  Thus the multi-file driver is terminal because all calls pass through it, and no underlying VFD receives all calls.  Does this make sense?

No, this does not make sense.  It defies any simple definition of
terminal.

Dave

src/H5FDmulti.c
derobins: > On Wed, Sep 01, 2021 at 12:04:00AM +0000, jrmainzer wrote: @jrmainzer commented on this pull request. > + * + * The flags parameter controls management of op_codes that + * are unknown to the callback + * + * The input and output parameters allow op_code specific + * input and output + * + * At present, this VFD supports no op codes of its own and + * simply passes ctl calls on to each member VFD. + * + * Return: Non-negative on success/Negative on failure + * + *------------------------------------------------------------------------- + */ +static herr_t +H5FD_multi_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void *input, void **output) I guess I didn't express this well. Yes, the multi-file driver will pass all VFD calls down, but none of its underlying VFDs will receive all calls. Thus the multi-file driver is terminal because all calls pass through it, and no underlying VFD receives all calls. Does this make sense?
> No, this does not make sense. It defies any simple definition of terminal. Dave

I agree with Dave. That definition is really confusing. I view the multi VFD as pass-through. If anything, we could have pass-through VFDs that solely forward and "demultiplexing" VFDs that route VFD calls to an array of underlying VFDs.

release_docs/RELEASE.txt
```diff
@@ -463,6 +463,19 @@ New Features
...
+    - Adds new "ctl" callback to VFD H5FD_class_t structure
+
+      The newly-added "ctl" callback allows Virtual File
+      Drivers to intercept and handle arbitary operations
+      identified by an operation code. The callback accepts
+      a pointer to the file to be operated on (H5FD_t *),
+      an operation code to identify the operation to perform,
+      a set of flags to govern the behavior of the operation
+      and input/output pointers that the VFD may use to
+      receive and return operation parameters/data.
```
derobins: I would probably replace the function description and replace it with the prototype.

src/H5FD.c
```diff
@@ -1687,6 +1687,111 @@ H5FD_unlock(H5FD_t *file)
...
+    if (!file)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "file pointer cannot be NULL")
+
+    if (!file->cls)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "file class pointer cannot be NULL")
+
+    /* Don't attempt to validate the op code.  If appropriate, that will
+     * be done by the underlying VFD callback, along with the input and
+     * output parameters.
+     */
```
derobins, line 1728: "Don't forget to validate the op code" isn't particularly helpful since that's out of scope in this function. I'd change the comment to simply say that the op_code and input and output pointers will be validated by the VFD.

src/H5FDmpi.c
```diff
@@ -41,26 +41,42 @@
...
  *
+ * Changes:     Reworked function to use the ctl callback so we can get
+ *              rid of H5FD_class_mpi_t.  Since there are no real limits
+ *              on what the ctl callback can do, its file parameter can't
+ *              be constant.  Thus, I had to remove the const qualifier
+ *              on this functions file parameter as well.  Note also the
+ *              circumlocution required to use the ctl callbacks output
+ *              parameter to pass back the rank without introducing
+ *              compiler warnings.
+ *                                             JRM -- 8/13/21
```
derobins, line 52: While I'm okay with leaving the "changes" sections in the cache code, I'd like to have that not spread to the rest of the library. We've been removing the ones that are noise and changing the useful bits to NOTE sections.

src/H5FD.c
jhendersonHDF: Maybe misread as "Don't forget (attempt) to validate the op code" here? It seems appropriate since it's explaining why we don't validate the op code at this point.

src/H5FDmpi.c
```diff
@@ -41,26 +41,42 @@
...
-H5FD_mpi_get_rank(const H5FD_t *file)
+H5FD_mpi_get_rank(H5FD_t *file)
 {
-    const H5FD_class_mpi_t *cls;
-
-    int ret_value;
+    const H5FD_class_t *cls;
+    uint64_t            flags    = H5FD_CTL__FAIL_IF_UNKNOWN_FLAG | H5FD_CTL__ROUTE_TO_TERMINAL_VFD_FLAG;
+    int                 rank     = -1;
+    void *              rank_ptr = (void *)(&rank);
```
derobins, line 62: Does this avoid a compiler warning? It's just as easy to do the cast when the callback is invoked.

src/H5FDmpi.c
```diff
@@ -41,26 +41,42 @@
...
+    HDassert(cls->ctl); /* All MPI drivers must implement this */
 
     /* Dispatch to driver */
-    if ((ret_value = (cls->get_rank)(file)) < 0)
+    if ((cls->ctl)(file, H5FD_CTL__GET_MPI_RANK_OPCODE, flags, NULL, &rank_ptr) < 0)
         HGOTO_ERROR(H5E_VFL, H5E_CANTGET, FAIL, "driver get_rank request failed")
 
+    HDassert(rank >= 0);
+
+    ret_value = rank;
```
derobins, line 78: I feel like this should be an actual error check and not an assert.

src/H5FDmpi.c
```diff
@@ -77,25 +93,43 @@ H5FD_mpi_get_rank(const H5FD_t *file)
...
+ * Changes:     Reworked function to use the ctl callback so we can get
+ *              rid of H5FD_class_mpi_t.  Since there are no real limits
+ *              on what the ctl callback can do, its file parameter can't
+ *              be constant.  Thus, I had to remove the const qualifier
+ *              on this functions file parameter as well.  Note also the
+ *              circumlocution required to use the ctl callbacks output
+ *              parameter to pass back the rank without introducing
+ *              compiler warnings.
+ *                                             JRM -- 8/13/21
+ *
```
derobins, line 105: See above with regard to the "changes" sections.

src/H5FDmpi.c
```diff
@@ -112,25 +146,43 @@ H5FD_mpi_get_size(const H5FD_t *file)
...
     HDassert(file);
-    cls = (const H5FD_class_mpi_t *)(file->cls);
+    cls = (const H5FD_class_t *)(file->cls);
     HDassert(cls);
-    HDassert(cls->get_comm); /* All MPI drivers must implement this */
+    HDassert(cls->ctl); /* All MPI drivers must implement this */
 
     /* Dispatch to driver */
-    if ((ret_value = (cls->get_comm)(file)) == MPI_COMM_NULL)
+    if ((cls->ctl)(file, H5FD_CTL__GET_MPI_COMMUNICATOR_OPCODE, flags, NULL, &comm_ptr) < 0)
```
derobins, line 178: Just put the comment about having to use a pointer variable instead of a cast here. No need for that in the changes section.

src/H5FDpublic.h
derobins: I don't think it's necessary unless we are going to iterate over all possible opcodes.

src/H5FDsec2.c
derobins: I feel like this should go in a testing VFD, not in the default VFD.

jhendersonHDF: Moved this into a minimal testing VFD in vfd.c

src/H5FDpublic.h
jhendersonHDF: Removed this definition for now. If we need it later we can add it back, but for now there's no need to add it to our public definitions if it's not being used anywhere.

release_docs/RELEASE.txt
jhendersonHDF: Done

src/H5FDmpi.c
jhendersonHDF: In this case, H5FD_CTL__GET_MPI_RANK_OPCODE is an internally-reserved operation, so if rank is still < 0 when the callback returns, the operation isn't implemented correctly in one of the library's internal MPI-aware VFDs (currently just the MPI I/O VFD), so I think an assertion makes sense here.

src/H5FDmpi.c
```diff
@@ -112,25 +146,43 @@ H5FD_mpi_get_size(const H5FD_t *file)
...
  *-------------------------------------------------------------------------
  */
 MPI_Comm
-H5FD_mpi_get_comm(const H5FD_t *file)
+H5FD_mpi_get_comm(H5FD_t *file)
 {
-    const H5FD_class_mpi_t *cls;
-    MPI_Comm                ret_value;
+    const H5FD_class_t *cls;
+    uint64_t            flags    = H5FD_CTL__FAIL_IF_UNKNOWN_FLAG | H5FD_CTL__ROUTE_TO_TERMINAL_VFD_FLAG;
```
soumagne, line 165: I'm not a big fan of these two flags that need to be passed almost everywhere. Why not make it the default behavior? Also why the double __ after H5FD_CTL ?

src/H5FDmpi.c
soumagne: Also agree, that cast is weird here

src/H5FDmpi.c
```diff
@@ -77,25 +93,43 @@ H5FD_mpi_get_rank(const H5FD_t *file)
...
 int
-H5FD_mpi_get_size(const H5FD_t *file)
+H5FD_mpi_get_size(H5FD_t *file)
 {
-    const H5FD_class_mpi_t *cls;
-    int                     ret_value;
+    const H5FD_class_t *cls;
+    uint64_t            flags    = H5FD_CTL__FAIL_IF_UNKNOWN_FLAG | H5FD_CTL__ROUTE_TO_TERMINAL_VFD_FLAG;
+    int                 size     = 0;
+    void *              size_ptr = (void *)(&size);
```
soumagne, line 114: Same here

src/H5FDmpi.c
```diff
@@ -112,25 +146,43 @@ H5FD_mpi_get_size(const H5FD_t *file)
...
 MPI_Comm
-H5FD_mpi_get_comm(const H5FD_t *file)
+H5FD_mpi_get_comm(H5FD_t *file)
 {
-    const H5FD_class_mpi_t *cls;
-    MPI_Comm                ret_value;
+    const H5FD_class_t *cls;
+    uint64_t            flags    = H5FD_CTL__FAIL_IF_UNKNOWN_FLAG | H5FD_CTL__ROUTE_TO_TERMINAL_VFD_FLAG;
+    MPI_Comm            comm     = MPI_COMM_NULL;
+    void *              comm_ptr = (void *)(&comm);
```
soumagne, line 167: also here

src/H5FDmpi.c
jrmainzer: That is the way I have been viewing it. 
I use an error for anything that can be triggered by user input (either via API calls, or data read from file), and asserts for conditions that can only occur through errors in our code.  Objections to this policy?  If so why?

src/H5FDmpi.c
jrmainzer: We need a method of recording significant changes to functions and the reasons for them, so as to avoid re-inserting bugs and alert programmers who are familiar with the code in question to changes.
Can you point to a better solution to this problem?

src/H5FDmpi.c
jrmainzer: That was my first solution -- unfortunately, the compiler still complained.

src/H5FDmpi.c
jrmainzer: Actually, I have a use case I'll be working on shortly that doesn't use either of these flags.

As for the H5FD_CTL__, that is a convention I have been using for package specific flags since forever.  That said, I don't regard it as particularly important one way or the other.

src/H5FDmpi.c
jrmainzer: Either way works for me.

src/H5FDmulti.c
jrmainzer: Please suggest a more useful set of terminology to classify VFDs.

lrknox (approved):

derobins (approved):

byrnHDF (approved):

PR #985, opened 2021-09-03
Run bin/format_source

fortnern:


fortnern:
Seems bin/format_source doesn't sync with the format check on here. No time to deal with it, I'll ignore failed format checks for now.

byrnHDF:
> Seems bin/format_source doesn't sync with the format check on here. No time to deal with it, I'll ignore failed format checks for now.

Not sure what you meant? Remember we are at v10 for clang-format.

PR #1017, opened 2021-09-16
Fixed HDFFV-11052 (CVE-2020-10812)

bmribler:
Description
    The tool h5debug on a corrupted file produced a core dump.   The
    issue occurred during failure recovery when a null shared file pointer
    was being accessed.

    A permanent solution to handle corrupted file during failure recovery
    will require more careful consideration to avoid bad side-effects to
    other parts of the library.  In the meantime, a workaround is used to
    catch the attempt of accessing the null shared file pointer.  The tool
    will then return error instead of segfault.
Platforms tested:
    Linux/64 (jelly)

### DIFF
--- a/src/H5VLnative_file.c
+++ b/src/H5VLnative_file.c
@@ -740,20 +740,34 @@ done:
  * Return:      Success:    SUCCEED
  *              Failure:    FAIL (file will not be closed)
  *
+ * Modifications:
+ *
+ *              BMR -- 9/15/21
+ *              Temporary solution for CVE-2020-10812:
+ *                  Added a check for the returned value of H5F_NREFS to detect
+ *                  a file shared pointer being null possibly due to file
+ *                  corruption.  When that situation occurs, H5F_NREFS will return
+ *                  UINT_MAX.  This is to avoid core dump in production mode.
  *-------------------------------------------------------------------------
  */
 herr_t
 H5VL__native_file_close(void *file, hid_t H5_ATTR_UNUSED dxpl_id, void H5_ATTR_UNUSED **req)
 {
-    int    nref;
-    H5F_t *f         = (H5F_t *)file;
-    hid_t  file_id   = H5I_INVALID_HID;
-    herr_t ret_value = SUCCEED; /* Return value */
+    int      nref;
+    H5F_t   *f         = (H5F_t *)file;
+    hid_t    file_id   = H5I_INVALID_HID;
+    unsigned f_nrefs   = UINT_MAX;
+    herr_t   ret_value = SUCCEED; /* Return value */
 
     FUNC_ENTER_PACKAGE
 
     /* This routine should only be called when a file ID's ref count drops to zero */
     HDassert(H5F_ID_EXISTS(f));
+
+    /* If f_nrefs is UINT_MAX, return failure to avoid core dump */
+    f_nrefs = H5F_NREFS(f);
+    if (f_nrefs >= UINT_MAX)
+        HGOTO_ERROR(H5E_FILE, H5E_BADVALUE, FAIL, "null pointer encountered, possible data corruption.")
 
     /* Flush file if this is the last reference to this id and we have write
      * intent, unless it will be flushed by the "shared" file being closed.
--- a/src/H5VLnative_file.c
+++ b/src/H5VLnative_file.c
@@ -754,7 +754,7 @@ herr_t
 H5VL__native_file_close(void *file, hid_t H5_ATTR_UNUSED dxpl_id, void H5_ATTR_UNUSED **req)
 {
     int      nref;
-    H5F_t   *f         = (H5F_t *)file;
+    H5F_t *  f         = (H5F_t *)file;
     hid_t    file_id   = H5I_INVALID_HID;
     unsigned f_nrefs   = UINT_MAX;
     herr_t   ret_value = SUCCEED; /* Return value */
--- a/src/H5Fquery.c
+++ b/src/H5Fquery.c
@@ -575,7 +575,7 @@ H5F_get_nrefs(const H5F_t *f)
     HDassert(f);
     HDassert(f->shared);
 
-    if(f->shared != NULL)
+    if (f->shared != NULL)
         ret_value = f->shared->nrefs;
 
     FUNC_LEAVE_NOAPI(ret_value)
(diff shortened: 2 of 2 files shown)


src/H5VLnative_file.c
```diff
@@ -740,21 +740,35 @@ H5VL__native_file_optional(void *obj, H5VL_optional_args_t *args, hid_t H5_ATTR_
...
 {
-    int    nref;
-    H5F_t *f         = (H5F_t *)file;
-    hid_t  file_id   = H5I_INVALID_HID;
-    herr_t ret_value = SUCCEED; /* Return value */
+    int      nref;
+    H5F_t *  f         = (H5F_t *)file;
+    hid_t    file_id   = H5I_INVALID_HID;
+    unsigned f_nrefs   = UINT_MAX;
+    herr_t   ret_value = SUCCEED; /* Return value */
```
bmribler, line 761: I'm a bit confused about H5I_get_ref() and H5F_NREFS -> H5F_get_nrefs, please confirm that the use of these two functions here are correct.

bmribler:
After I applied the changes, the core dump was avoided but I got the infinite loop error:
HDF5: infinite loop closing library
L,T_top,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F,F

I created this draft PR so you can look over the changes and give me suggestions to eliminate the infinite loop error.  I think H5L_table_g was not 0 causing that but what can I do?

gnuoyd:
This change does not appear to fix the bug. How does `f->shared` unexpectedly become `NULL`?

bmribler:
> This change does not appear to fix the bug. How does `f->shared` unexpectedly become `NULL`?

The corrupted file caused wacky things to happen, Dave.  Dana helped me and I'm making changes to the ...changes.  As I mentioned in the Description, this is a temporary workaround that a few of us discussed and decided to present to other developers for more suggestions.
The idea is to stop all the fun stuff segfault/core dump from happening and the tool will fail gracefully when situations like this occur.

gnuoyd (changes requested):
Root cause analysis should be performed.

gnuoyd:
It looks like the library tries to read a cache image from the file at byte offset 65792 when `h5debug` calls H5Fclose. The file is only 2565 bytes long, so the read fails. Here is the back trace:

```
bt
#0  H5F_block_read (f=f@entry=0x64a550, type=type@entry=H5FD_MEM_SUPER, 
    addr=65792, size=268435456, buf=0x7fffe6cdd010)
    at ../../../manydsets/src/H5Fio.c:147
#1  0x00007ffff785c2de in H5C__read_cache_image (cache_ptr=0x7ffff7f5e010, 
    f=0x64a550) at ../../../manydsets/src/H5Cimage.c:1083
#2  H5C__load_cache_image (f=0x64a550)
    at ../../../manydsets/src/H5Cimage.c:1083
#3  0x00007ffff785da14 in H5C__prep_image_for_file_close (f=0x64a550, 
    image_generated=image_generated@entry=0x7fffffffc21f)
    at ../../../manydsets/src/H5Cimage.c:1295
#4  0x00007ffff785026e in H5C_prep_for_file_close (f=f@entry=0x64a550)
    at ../../../manydsets/src/H5C.c:745
#5  0x00007ffff7837689 in H5AC_prep_for_file_close (f=f@entry=0x64a550)
    at ../../../manydsets/src/H5AC.c:1185
#6  0x00007ffff78ca3ab in H5F__dest (f=f@entry=0x64a550, 
    flush=flush@entry=true) at ../../../manydsets/src/H5Fint.c:1381
#7  0x00007ffff78ccc9f in H5F_try_close (f=0x64a550, 
    was_closed=was_closed@entry=0x0) at ../../../manydsets/src/H5Fint.c:2513
#8  0x00007ffff78ccf8c in H5F__close (f=f@entry=0x64a550)
    at ../../../manydsets/src/H5Fint.c:2315
#9  0x00007ffff7abe3bf in H5VL__native_file_close (file=0x64a550, 
    dxpl_id=<optimized out>, req=<optimized out>)
    at ../../../manydsets/src/H5VLnative_file.c:777
#10 0x00007ffff7aace8e in H5VL__file_close (cls=<optimized out>, req=0x0, 
    dxpl_id=792633534417207304, obj=<optimized out>)
    at ../../../manydsets/src/H5VLcallback.c:4115
#11 H5VL_file_close (vol_obj=vol_obj@entry=0x64c480, 
    dxpl_id=792633534417207304, req=0x0)
    at ../../../manydsets/src/H5VLcallback.c:4146
#12 0x00007ffff78ca096 in H5F__close_cb (file_vol_obj=0x64c480, 
    request=<optimized out>) at ../../../manydsets/src/H5Fint.c:216
#13 0x00007ffff793cf6a in H5I__dec_ref (request=0x0, 
    id=id@entry=72057594037927936) at ../../../manydsets/src/H5Iint.c:1036
#14 H5I__dec_app_ref (request=0x0, id=id@entry=72057594037927936)
    at ../../../manydsets/src/H5Iint.c:1108
#15 H5I_dec_app_ref (id=id@entry=72057594037927936)
    at ../../../manydsets/src/H5Iint.c:1156
#16 0x00007ffff78c0d79 in H5Fclose (file_id=file_id@entry=72057594037927936)
    at ../../../manydsets/src/H5F.c:1060
#17 0x0000000000401e3e in main (argc=<optimized out>, argv=<optimized out>)
    at ../../../../../manydsets/tools/src/misc/h5debug.c:820
(gdb) 
```

(Line numbers probably do not match `develop`.)

The error percolates up the stack to H5F__dest, which notes the error but does not return until after it sets `f->shared` to NULL at this line:

```
            f->shared = (H5F_shared_t *)H5FL_FREE(H5F_shared_t, f->shared);
```

At that point there is an extant ID for `f` but `f` is in a partially deinitialized state.

derobins:
In general, close routines shouldn't assume that the structures passed to them are in a sane state since they will be invoked after failed opens, etc. I haven't personally investigated this bug, but reworking the close call so it's more tolerant of NULL pointers and so that it tries to shut down as much as possible even in the face of badness seems to be what's needed here.

bmribler:
Dana will fix this instead.

PR #1033, opened 2021-09-22
To reduce maintenance effort, delete the `noerror-` variants of the compiler flags files

gnuoyd:
To reduce maintenance effort, delete the `noerror-` variants of the
    compiler flags files, since they essentially duplicate the `error-`
    files modulo the replacement of `-Werror=` with `-W` and any changes in
    comments.  (I verified the duplication with a script.)
    
    For autoconf, reinstate the use of the `demote_errors` shell function to
    derive the `noerror-` content from the `error-` content.  `demote_errors`
    replaces `-Werror=` with `-W` when `WARNINGS_AS_ERRORS` is `no`.
    
    Slightly reorder `configure.ac` so that the setting of
    `WARNINGS_AS_ERRORS` takes effect before the `error-` files are sourced.

    Take a stab at updating the CMake files to match the changes I made to
    the autoconf files to remove `noerror-` files.  I'm not much of a CMake
    user so these changes are quite rough.
    
    Looks like the duplication can be reduced with the introduction of a new
    macro.


config/cmake/HDFCXXCompilerFlags.cmake
```diff
@@ -134,11 +134,11 @@ else ()
         AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 4.8)
       # add the general CXX flags for g++ compiler versions 4.8 and above.
       ADD_H5_FLAGS (HDF5_CMAKE_CXX_FLAGS "${HDF5_SOURCE_DIR}/config/gnu-warnings/cxx-general")
-      if (HDF5_ENABLE_WARNINGS_AS_ERRORS)
```
byrnHDF, line 137: Just modify the Add_H5_FLAGS macro to check the HDF5_ENABLE_WARNINGS_AS_ERRORS option.
This would clean up the if else tree here. (Maybe even move the CMAKE_[C,CXX]_COMPILER_ID if in there also?)

byrnHDF (changes requested):

byrnHDF (approved):

byrnHDF:
Release.txt should have a note added?

config/cmake/HDFCXXCompilerFlags.cmake
gnuoyd: I moved the check into the macro.

lrknox (approved):

PR #1034, opened 2021-09-22
added simple API to get subfiling fapl

brtnfld:
Mimics the other H5Pset_fapl_* APIs. We still need to remove the subfile private header dependency.


jhendersonHDF:
Is this routine just meant for setting up a default subfiling FAPL when no config is given? If so, it seems like it'd be much better to just bake that into the existing routine when the config pointer is NULL, rather than maintaining two separate API routines.

brtnfld:
@jhendersonHDF  Updated to accept NULL, is this what you mean?

jhendersonHDF:
> @jhendersonHDF Updated to accept NULL, is this what you mean?

Yep. That just seems easier rather than having two API routines for this.

src/H5FDsubfiling.c
```diff
@@ -433,18 +434,50 @@ H5FD__copy_plist(hid_t fapl_id, hid_t *id_out_ptr)
...
 H5Pset_fapl_subfiling(hid_t fapl_id, H5FD_subfiling_config_t *fa)
 {
-    H5P_genplist_t *plist     = NULL; /* Property list pointer */
-    herr_t          ret_value = FAIL;
+    H5P_genplist_t *  plist      = NULL; /* Property list pointer */
+    hid_t             ioc_fapl   = H5I_INVALID_HID;
+    H5FD_ioc_config_t ioc_config = {
+        0,
+    };
+    H5FD_subfiling_config_t subfiling_conf = {
```
brtnfld, line 442: need to move this to if cond.

src/H5FDsubfiling.c
```diff
@@ -433,18 +434,50 @@ H5FD__copy_plist(hid_t fapl_id, hid_t *id_out_ptr)
...
+        0,
+    };
```
jrmainzer: Any particular reason for these initializers?  If I read the code correctly, you are not using these  structures, or are initializing them elsewhere.

src/H5FDsubfiling.c
```diff
@@ -433,18 +434,50 @@ H5FD__copy_plist(hid_t fapl_id, hid_t *id_out_ptr)
...
 
-    plist = H5P_object_verify(fapl_id, H5P_FILE_ACCESS);
+    if (fa == NULL) {
+        /* Create IOC fapl */
+        ioc_fapl = H5Pcreate(H5P_FILE_ACCESS);
+        if (H5I_INVALID_HID == ioc_fapl)
+            HGOTO_ERROR(H5E_PLIST, H5E_CANTSET, FAIL, "can't create ioc fapl")
+
+        /* Get subfiling VFD defaults */
+        if (H5Pget_fapl_subfiling(fapl_id, &subfiling_conf) < 0)
```
jrmainzer: Why not use fapl__get_subfiling_defaults() here?  If I read H5Pget_fapl_subfiling() correctly, it only returns the default sub-filing FAPL entry if no sub-filing FAPL entry is defined in the supplied FAPL.  In most cases, this is a distinction without a difference, but I could see it resulting in perplexing bugs now and then.

src/H5FDsubfiling.c
```diff
@@ -433,18 +434,50 @@ H5FD__copy_plist(hid_t fapl_id, hid_t *id_out_ptr)
...
+            HGOTO_ERROR(H5E_PLIST, H5E_CANTSET, FAIL, "can't get subfiling fapl")
+
+        if (subfiling_conf.require_ioc) {
+            /* Get IOC VFD defaults */
+            if (H5Pget_fapl_ioc(ioc_fapl, &ioc_config) < 0)
+                HGOTO_ERROR(H5E_PLIST, H5E_CANTSET, FAIL, "can't get ioc fapl")
```
jrmainzer, line 505: Since you use created the doc_fapl, I don't think you need to worry about pre-existing IOC fapl entries in this case.  However, calling fapl_get_ioc_defaults() with be a little quicker.

src/H5FDsubfiling.c
brtnfld: They are initialized to get around the NULL asserts.

src/H5FDsubfiling.c
brtnfld: I would need to remove the HDassert(fa) in fapl__get_subfiling_defaults

brtnfld: Is not  H5Pget_fapl_subfiling calling fapl__get_subfiling_defaults. And fapl__get_subfiling_defaults expects fa to to be NULL

src/H5FDsubfiling.c
brtnfld: I'll just remove it. Why does fapl_get_ioc_defaults have only one "_" and  fapl__get_subfiling_defaults have two in the name?

src/H5FDsubfiling.c
jrmainzer: As you say, fapl__get_subfiling_defaults() is called by H5Pget_fapl_subfiling(), but only if the supplied fapl doesn't contain a sub-filing fapl entry.

If I read the assert in fapl__get_subfiling_defaults correctly, it wants the fa parameter to be non-NULL (HDassert(fa);).

src/H5FDsubfiling.c
jrmainzer: Darned if I know.  Quincey was trying to require package private functions to use the double underscore (i.e. H5FD__), so it might have something to do with that.

brtnfld: It does not work if I rm this.

brtnfld: fapl_get_ioc_defaults is a static  function, I gues that is why it is one "_". Should I not make it static so that I can use it?

src/H5FDsubfiling.c
brtnfld: Changed to use fapl__get_subfiling_defaults

src/H5FDsubfiling.c
jrmainzer: Sorry, I didn't realize that fapl_get_ioc_defaults() was static, and thus not available outside of H5FDioc.  It is a very minor issue, and I wouldn't be surprised if the code changes so as to make the point moot.  I suggest you leave this as is.

PR #1036, opened 2021-09-24
Implement parallel collective support for selection I/O.

fortnern:


jrmainzer:
Other than the need for a formatting run, no obvious issues.

jrmainzer (approved):

src/H5Dcontig.c
```diff
@@ -658,7 +658,7 @@ H5D__contig_read(H5D_io_info_t *io_info, const H5D_type_info_t *type_info, hsize
         /* Issue selection I/O call (we can skip the page buffer because we've
          * already verified it won't be used, and the metadata accumulator
          * because this is raw data) Only call funciton if nelmts > 0. */
```
jhendersonHDF: Does this header comment need updated in relation to now always calling the function below?

src/H5Dcontig.c
```diff
@@ -705,9 +705,9 @@ H5D__contig_write(H5D_io_info_t *io_info, const H5D_type_info_t *type_info, hsiz
         /* Issue selection I/O call (we can skip the page buffer because we've
          * already verified it won't be used, and the metadata accumulator
          * because this is raw data). Only call funciton if nelmts > 0. */
```
jhendersonHDF: Same here about updating the comment?

src/H5Dcontig.c
fortnern: Oops, thanks, I'll fix that when I run the formatter.

PR #1054, opened 2021-09-29
Update compact dataset I/O routines to handle driver-level memory copy

jhendersonHDF:
These changes are needed to be compatible with the GPU VFD. Since the VFD uses buffers that are allocated via cudaMalloc, which aren't compatible with memcpy from the standard library, HDF5 needs to be able to request that a driver handle a memory copy if necessary.

### DIFF
--- a/src/H5Dcompact.c
+++ b/src/H5Dcompact.c
@@ -47,6 +47,24 @@
 /* Local Typedefs */
 /******************/
 
+/* Callback info for readvv operation when file driver
+ * uses special memory management functions
+ */
+typedef struct H5D_compact_readvv_special_mem_ud_t {
+    H5F_shared_t *f_sh;   /* Shared file for dataset */
+    void *        rbuf;   /* Pointer to buffer to be read from */
+    void *        dstbuf; /* Pointer to buffer to be read into */
+} H5D_compact_readvv_special_mem_ud_t;
+
+/* Callback info for writevv operation when file driver
+ * uses special memory management functions
+ */
+typedef struct H5D_compact_writevv_special_mem_ud_t {
+    H5F_shared_t *f_sh;   /* Shared file for dataset */
+    const void *  wbuf;   /* Pointer to buffer to be written from */
+    void *        dstbuf; /* Pointer to buffer to be written into */
+} H5D_compact_writevv_special_mem_ud_t;
+
 /********************/
 /* Local Prototypes */
 /********************/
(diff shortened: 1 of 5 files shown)


src/H5Dcompact.c
```diff
@@ -47,6 +47,24 @@
...
+} H5D_compact_readvv_special_mem_ud_t;
+
+/* Callback info for writevv operation when file driver
+ * uses special memory management functions
+ */
+typedef struct H5D_compact_writevv_special_mem_ud_t {
+    H5F_shared_t *f_sh;   /* Shared file for dataset */
+    const void *  wbuf;   /* Pointer to buffer to be written from */
+    void *        dstbuf; /* Pointer to buffer to be written into */
+} H5D_compact_writevv_special_mem_ud_t;
```
jhendersonHDF: While annoying, use two different structures similar to the approach in H5Dcontig.c for const-correctness without warnings.

src/H5Dcompact.c
```diff
@@ -239,6 +259,48 @@ H5D__compact_io_init(const H5D_io_info_t *io_info, const H5D_type_info_t H5_ATTR
...
+    H5D_compact_readvv_special_mem_ud_t *udata = (H5D_compact_readvv_special_mem_ud_t *)_udata;
+    H5FD_ctl_memcpy_args_t               op_args;
+    uint64_t                             op_flags;
+    H5FD_t *                             file_handle = NULL;
+    herr_t                               ret_value   = SUCCEED;
+
+    FUNC_ENTER_STATIC
+
+    /* Retrieve lower-level file handle for I/O */
+    if (H5F_shared_get_file_handle(udata->f_sh, &file_handle) < 0)
```
jhendersonHDF: Here, we need a handle to the shared file structure's H5FD_t pointer. However, the H5F package hides this behind an incomplete typedef, so we can't just directly access it through udata->f_sh->lf. We have to create a routine in H5Fquery.c to give this back to us.

src/H5FDpublic.h
```diff
@@ -185,6 +193,9 @@
 #define H5FD_CTL__GET_MPI_COMMUNICATOR_OPCODE 2
 #define H5FD_CTL__GET_MPI_RANK_OPCODE         3
 #define H5FD_CTL__GET_MPI_SIZE_OPCODE         4
+#define H5FD_CTL__MEM_ALLOC                   5
+#define H5FD_CTL__MEM_FREE                    6
+#define H5FD_CTL__MEM_COPY                    7
```
jhendersonHDF, line 198: Here, I only reserved the op codes for driver-level memory allocations and frees. I haven't added any argument structs for them since there isn't a use case yet, but can do so if it's useful.

src/H5Fprivate.h
```diff
@@ -24,11 +24,9 @@ typedef struct H5F_t H5F_t;
 /* Include package's public header */
 #include "H5Fpublic.h"
 
-/* Public headers needed by this file */
-#include "H5FDpublic.h" /* File drivers                 */
-
 /* Private headers needed by this file */
 #include "H5MMprivate.h" /* Memory management            */
+#include "H5FDprivate.h" /* File drivers                 */
```
jhendersonHDF, line 29: Since H5F_shared_get_file_handles needs the definition for the H5FD_t type, and since H5FDprivate.h includes H5FDpublic.h, just include the private header rather than both here.

derobins (approved):

soumagne (approved):

byrnHDF:
Should I assume all these changes to develop have had someone add release.txt entries?

src/H5FDpublic.h
jrmainzer: Agree -- wait until we have use cases.

jhendersonHDF:
> Should I assume all these changes to develop have had someone add release.txt entries?

It looks like the VFD plugins PR didn't have a RELEASE.txt entry associated with it. Let me create that in a separate PR and then add a RELEASE.txt entry for this PR here.

jrmainzer (approved):
VFD code looks good.
No obvious issues with H5D code, but then I'm not as familiar with it as I would like.

qkoziol (approved):

lrknox (approved):

PR #1058, opened 2021-09-30
Fix for CVE-2020-10812

derobins:
Fixing this in 1.10 first so we can get 1.10.8 out the door


derobins:
This fails tests

bmribler:
Did you mean "make check" failed?

PR #1085, opened 2021-10-08
Update version to 1.10.8 for release

lrknox:


README.txt
```diff
@@ -1,4 +1,4 @@
-HDF5 version 1.10.8-3 released on 2021-10-06
+HDF5 version 1.10.8-4 currently under development
```
byrnHDF: "under development"?

lrknox: Yes.  It only changes to "released on <date>" when running bin/release.  I don't expect that we will release 1.10.8-4.  Probably the next release will be 1.10.8,

byrnHDF (approved):

PR #1106, opened 2021-10-15
Add release note for HDFFV-11150 fix.

lrknox:


byrnHDF (approved):

byrnHDF (approved):

release_docs/RELEASE.txt
```diff
@@ -58,7 +58,13 @@ New Features
...
 
       Add configure options (autotools - CMake):
-            enable-hltools       HDF5_BUILD_HL_TOOLS
+            --enable-hltools       HDF5_BUILD_HL_TOOLS
+
+      Disabling this option prevents building the gif tool which
+      contains the following CVEs:
+          HDFFV-10592 CVE-2018-17433
+          HDFFV-10593 CVE-2018-17436
+          HDFFV-11048 CVE-2020-10809
```
bmribler, line 68: So, how to disable this option?

byrnHDF: It is explained in the known problems. But should be obvious to users of either autotools or cmake because it would be the opposite of the default?

bmribler: So, --enable-hltools was added but --disable-hltools is automatic?  Sorry, I'm clueless...

lrknox: In configure, adding option --enable-hltools also adds --disable-hltools and --enable-hltools=no to disable the option.

bmribler (approved):

PR #1109, opened 2021-10-18
Move test utilities to utils/test folder

byrnHDF:
Cleanup utils checks


lrknox (approved):

tools/test/h5dump/CMakeVFDTests.cmake
```diff
@@ -79,7 +66,7 @@ endmacro ()
 ##############################################################################
 
 # Run test with different Virtual File Driver
-foreach (vfd ${VFD_H5DUMP_LIST})
+foreach (vfd ${VFDP_LIST})
```
jhendersonHDF: VFDP -> VFD

test/CMakeLists.txt
```diff
@@ -378,6 +378,9 @@ set (H5_TESTS
     cmpd_dtransform
     event_set
 )
+if (HDF5_BUILD_UTILS)
+  set (H5_TESTS ${H5_TESTS} mirror_vfd)
+endif ()
```
jhendersonHDF, line 383: Why is this test only built based on the HDF5_BUILD_UTILS condition now? From a quick glance it didn't look like it was using anything from the utils dir.

test/CMakeLists.txt
```diff
@@ -552,30 +555,25 @@ if (HDF5_ENABLE_FORMATTERS)
...
+    TARGET_C_PROPERTIES (mirror_vfd SHARED)
+    target_link_libraries (mirror_vfd PRIVATE ${HDF5_TEST_LIBSH_TARGET})
+  endif ()
+  set_target_properties (mirror_vfd PROPERTIES FOLDER test)
 
-#-----------------------------------------------------------------------------
-# Add Target to clang-format
-#-----------------------------------------------------------------------------
-if (HDF5_ENABLE_FORMATTERS)
-  clang_format (HDF5_TEST_event_set_FORMAT event_set)
```
jhendersonHDF, line 578: Looks like this line for the separate event_set test got accidentally removed

byrnHDF: Not accident it is included in the make macro as event_set is in the list

test/CMakeLists.txt
byrnHDF: mirror_vfd needs the mirror_server in utils right?

jhendersonHDF: Ah ok, yes that's probably the case.

jhendersonHDF (approved):

PR #1126, opened 2021-10-21
Removes the "try free" behavior from the skip lists

derobins:
This was only used in the ID code when iterating and a callback
could delete IDs. It is not used anywhere else in the library and
is now pointless overhead.

Also quiets the const warnings when returning stored elements. They
only need to be const with respect to the skip list code, which should
never modify them. The library can do whatever it wants with the elements
it stored.


bsawicki (approved):
LGTM

lrknox (approved):

fortnern (approved):

PR #1127, opened 2021-10-21

hyoklee:
This will remove ``` warning: implicit conversion turns floating-point number into integer: 'float' to 'size_t' (aka 'unsigned long') [-Wfloat-conversion]``` warnings.

### DIFF
--- a/src/H5FDcore.c
+++ b/src/H5FDcore.c
@@ -190,8 +190,8 @@ static const H5FD_class_t H5FD_core_g = {
 };
 
 /* Default configurations, if none provided */
-static const H5FD_core_fapl_t H5FD_core_default_config_g       = {(size_t)H5_MB, TRUE, H5FD_CORE_WRITE_TRACKING_FLAG,
-                                                            H5FD_CORE_WRITE_TRACKING_PAGE_SIZE};
+static const H5FD_core_fapl_t H5FD_core_default_config_g = {
+    (size_t)H5_MB, TRUE, H5FD_CORE_WRITE_TRACKING_FLAG, H5FD_CORE_WRITE_TRACKING_PAGE_SIZE};
 static const H5FD_core_fapl_t H5FD_core_default_paged_config_g = {(size_t)H5_MB, TRUE, TRUE, (size_t)4096};
 
 /* Define a free list to manage the region type */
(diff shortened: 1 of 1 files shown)


derobins (approved):

src/H5FDcore.c
```diff
@@ -190,9 +190,9 @@ static const H5FD_class_t H5FD_core_g = {
 };
 
 /* Default configurations, if none provided */
-static const H5FD_core_fapl_t H5FD_core_default_config_g       = {H5_MB, TRUE, H5FD_CORE_WRITE_TRACKING_FLAG,
+static const H5FD_core_fapl_t H5FD_core_default_config_g       = {(size_t)H5_MB, TRUE, H5FD_CORE_WRITE_TRACKING_FLAG,
                                                             H5FD_CORE_WRITE_TRACKING_PAGE_SIZE};
-static const H5FD_core_fapl_t H5FD_core_default_paged_config_g = {H5_MB, TRUE, TRUE, (size_t)4096};
+static const H5FD_core_fapl_t H5FD_core_default_paged_config_g = {(size_t)H5_MB, TRUE, TRUE, (size_t)4096};
```
derobins, line 195: This is fine, I guess, but I would probably create a const int H5_CORE_DEFAULT_ALLOC_SIZE or something like that and make it 1024 * 1024 instead. That's more in the spirit of what's intended and doesn't torture the floating-point H5_MB into a new use.

Also, the size_t cast for 4096 is probably superfluous.

hyoklee: I tried that already and it generated more warnings.

derobins: Compilers are weird

byrnHDF (approved):

lrknox:
clang-format check failed, identifying the changes below for lines 193 and 955.  Usually it commits the changes, but this time did not.  If the changes look reasonable, maybe change them manually, commit the changes and see what happens?

--- ./src/H5FDcore.c	(original)
+++ ./src/H5FDcore.c	(reformatted)
@@ -190,8 +190,8 @@
 };
 
 /* Default configurations, if none provided */
-static const H5FD_core_fapl_t H5FD_core_default_config_g       = {(size_t)H5_MB, TRUE, H5FD_CORE_WRITE_TRACKING_FLAG,
-                                                            H5FD_CORE_WRITE_TRACKING_PAGE_SIZE};
+static const H5FD_core_fapl_t H5FD_core_default_config_g = {
+    (size_t)H5_MB, TRUE, H5FD_CORE_WRITE_TRACKING_FLAG, H5FD_CORE_WRITE_TRACKING_PAGE_SIZE};
 static const H5FD_core_fapl_t H5FD_core_default_paged_config_g = {(size_t)H5_MB, TRUE, TRUE, (size_t)4096};
 
 /* Define a free list to manage the region type */
@@ -952,8 +952,8 @@
                             "file read failed: time = %s, filename = '%s', file descriptor = %d, errno = %d, "
                             "error message = '%s', file->mem = %p, total read size = %llu, bytes this "
                             "sub-read = %llu, bytes actually read = %llu, offset = %llu",
-                            HDctime(&mytime), file->name, file->fd, myerrno, HDstrerror(myerrno), (void *)file->mem,
-                            (unsigned long long)size, (unsigned long long)bytes_in,
+                            HDctime(&mytime), file->name, file->fd, myerrno, HDstrerror(myerrno),
+                            (void *)file->mem, (unsigned long long)size, (unsigned long long)bytes_in,
                             (unsigned long long)bytes_read, (unsigned long long)offset);
                     } /* end if */

jhendersonHDF (approved):

PR #1129, opened 2021-10-21

hyoklee:
This will fix the following warning:

```h5dumpgentest.c:10258:5: warning: cast from 'long double **' to 'long double *'  increases required alignment from 8 to 16 [-Wcast-align]```

### DIFF
--- a/test/h5test.h
+++ b/test/h5test.h
@@ -210,7 +210,7 @@ H5TEST_DLLVAR MPI_Info h5_io_info_g; /* MPI INFO object for IO */
                                                                                                              \
         ARR = (TYPE **)HDmalloc(h5taa_pointers_size + h5taa_data_size);                                      \
                                                                                                              \
-        ARR[0] = (TYPE *)(ARR + (DIMS_I));                                                                   \
+        ARR[0] = (void *)(ARR + (DIMS_I));                                                                   \
                                                                                                              \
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by byrnHDF — APPROVED** [review]

File: test/h5test.h [hunk 1]
```diff
@@ -210,7 +210,7 @@ H5TEST_DLLVAR MPI_Info h5_io_info_g; /* MPI INFO object for IO */
                                                                                                              \
         ARR = (TYPE **)HDmalloc(h5taa_pointers_size + h5taa_data_size);                                      \
                                                                                                              \
-        ARR[0] = (TYPE *)(ARR + (DIMS_I));                                                                   \
+        ARR[0] = (void *)(ARR + (DIMS_I));                                                                   \
```
gnuoyd, line 213: This is a place where the warning is easy to suppress but the alignment may still be wrong.

I believe `DIMS_I` and `DIMS_J` are constant. If so, then you can malloc a `struct { TYPE *rows[DIMS_I]; TYPE values[DIMS_I * DIMS_J]; } *prototype;`, initialize all `DIMS_I` pointers in `rows`, and then set `ARR` to `&prototype->rows[0]`.

derobins: As in #1130, I agree with Dave. Casting to a void pointer to make the compiler quiet is probably not the best fix here.

gnuoyd:
I believe #1169 fixed this?

lrknox:
Fixed by #1169

PR #1130, opened 2021-10-21

hyoklee:
This will remove ``` warning: cast from 'unsigned char *' to 'H5R_ref_t *' increases  required alignment from 1 to 8 [-Wcast-align]```.

### DIFF
--- a/tools/lib/h5tools.c
+++ b/tools/lib/h5tools.c
@@ -1893,11 +1893,10 @@ render_bin_output(FILE *stream, hid_t container, hid_t tid, void *_mem, hsize_t 
 
                     for (block_index = 0; block_index < block_nelmts; block_index++) {
                         mem = ((unsigned char *)_mem) + block_index * size;
-                        if ((region_id = H5Ropen_object((H5R_ref_t *)mem, H5P_DEFAULT, H5P_DEFAULT)) < 0)
+                        if ((region_id = H5Ropen_object((void *)mem, H5P_DEFAULT, H5P_DEFAULT)) < 0)
                             H5TOOLS_INFO("H5Ropen_object H5T_STD_REF failed");
                         else {
-                            if ((region_space = H5Ropen_region((H5R_ref_t *)mem, H5P_DEFAULT, H5P_DEFAULT)) >=
-                                0) {
+                            if ((region_space = H5Ropen_region((void *)mem, H5P_DEFAULT, H5P_DEFAULT)) >= 0) {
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

tools/lib/h5tools.c
```diff
@@ -1893,11 +1893,10 @@ render_bin_output(FILE *stream, hid_t container, hid_t tid, void *_mem, hsize_t
 
                     for (block_index = 0; block_index < block_nelmts; block_index++) {
                         mem = ((unsigned char *)_mem) + block_index * size;
-                        if ((region_id = H5Ropen_object((H5R_ref_t *)mem, H5P_DEFAULT, H5P_DEFAULT)) < 0)
+                        if ((region_id = H5Ropen_object((void *)mem, H5P_DEFAULT, H5P_DEFAULT)) < 0)
                             H5TOOLS_INFO("H5Ropen_object H5T_STD_REF failed");
                         else {
-                            if ((region_space = H5Ropen_region((H5R_ref_t *)mem, H5P_DEFAULT, H5P_DEFAULT)) >=
-                                0) {
+                            if ((region_space = H5Ropen_region((void *)mem, H5P_DEFAULT, H5P_DEFAULT)) >= 0) {
```
gnuoyd, line 1899: Is `mem` guaranteed to be 8-byte aligned? If it is not, then the behavior of accesses through the `H5R_ref_t` pointer in calls under `H5Ropen_object((void *)mem, ...)` may be undefined.

If the `H5R_ref_t` in the `size` bytes at `mem` is known to be well-formed, then one way to quiet the alignment warning is to copy the bytes to an `H5R_ref_t`-aligned local variable, and then pass a pointer to the local variable instead of to `mem`. I'm not sure if that is the right way, here, though, because it looks like `H5Ropen_object` may modify the `H5R_ref_t`.

derobins: I feel like this needs much more careful consideration than simply trying to make the compiler quiet.

hyoklee: @derobins and @gnuoyd, can you guys come up with a simple test C program, add it into our test suite, and fails due to this change?

derobins: @hyoklee The problem is that the behavior is going to be implementation-specific. We may not even have access to a platform that fails.

gnuoyd:
#1171 resolved this.

gnuoyd:
@hyoklee, this is resolved by #1171, correct?

hyoklee:
@gnuoyd yes

PR #1136, opened 2021-10-22

hyoklee:
This will remove ```warning: variable 'fid' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]```.

### DIFF
--- a/test/swmr.c
+++ b/test/swmr.c
@@ -5220,7 +5220,7 @@ test_file_lock_swmr_concur(hid_t H5_ATTR_UNUSED in_fapl)
 static int
 test_file_lock_swmr_concur(hid_t in_fapl)
 {
-    hid_t fid;                     /* File ID */
+    hid_t fid = -1;                /* File ID */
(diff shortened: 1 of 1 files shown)


test/swmr.c
```diff
@@ -5220,7 +5220,7 @@ test_file_lock_swmr_concur(hid_t H5_ATTR_UNUSED in_fapl)
 static int
 test_file_lock_swmr_concur(hid_t in_fapl)
 {
-    hid_t fid;                     /* File ID */
+    hid_t fid = -1;                /* File ID */
```
byrnHDF: Instead of -1 use H5INVALID_ID

derobins: See my comments in #1137

derobins: Also, as a rule, I initialize all IDs to H5I_INVALID_HID as it makes cleanup easier.

lrknox (approved):

byrnHDF (approved):

PR #1137, opened 2021-10-23

hyoklee:
This removes ```warning: variable 'X' is used uninitialized whenever 'if' condition is true [-Wsometimes-uninitialized]``` where X occurs in many places.

### DIFF
--- a/test/cork.c
+++ b/test/cork.c
@@ -304,8 +304,8 @@ verify_obj_dset_cork(hbool_t swmr)
     hid_t       aid  = -1;            /* Attribute ID */
     hid_t       sid = -1, sid2 = -1;  /* Dataspace IDs */
     hid_t       did = -1, did2 = -1;  /* Dataset IDs */
-    hid_t       oid = -1;             /* Object ID */
-    hid_t       dcpl2;                /* Dataset creation property list */
+    hid_t       oid          = -1;    /* Object ID */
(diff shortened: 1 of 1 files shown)


test/cork.c
```diff
@@ -304,8 +304,8 @@ verify_obj_dset_cork(hbool_t swmr)
     hid_t       aid  = -1;            /* Attribute ID */
     hid_t       sid = -1, sid2 = -1;  /* Dataspace IDs */
     hid_t       did = -1, did2 = -1;  /* Dataset IDs */
-    hid_t       oid = -1;             /* Object ID */
-    hid_t       dcpl2;                /* Dataset creation property list */
+    hid_t       oid          = -1;    /* Object ID */
```
byrnHDF: Instead of -1 use the H5I_INVALID_HID

derobins: The library is inconsistent about using -1 vs H5I_INVALID_HID, but the latter is more idiomatic and preferred. I'm not a huge fan of C's type sloppiness.

derobins: I've been trying to switch code over as I work on tests (the library uses H5I_INVALID_HID almost exclusively now). I'd appreciate the change here but that doesn't need to hold the merge.

lrknox (approved):

byrnHDF (approved):
Thanks for changing the "-1" elements.
Going forward we should ask that the ids are declared vertically and individually for better visual aesthetics.

PR #1164, opened 2021-11-04
Post open fix

qkoziol:
Updates to allow cache VOL connector to work properly


byrnHDF (approved):

brtnfld (approved):
Fortran looks fine.

src/H5ESint.c
```diff
@@ -817,14 +820,21 @@ H5ES__get_err_info_cb(H5ES_event_t *ev, void *_ctx)
     /* The 'app_func_name', 'app_file_name', and 'api_name' strings are statically allocated (by the compiler)
      * so there's no need to duplicate them internally, but they are duplicated
      * here, when they are given back to the user.
+     *
+     * They are also allowed to be NULL, so only check for NULL values from
+     * the strdup when the original string is non-NULL.
      */
-    if (NULL == (ctx->curr_err_info->api_name = H5MM_strdup(ev->op_info.api_name)))
+    ctx->curr_err_info->api_name = H5MM_strdup(ev->op_info.api_name);
+    if (ev->op_info.api_name && NULL == ctx->curr_err_info->api_name)
```
bmribler, line 828: I'm just curious why using mixed styles, why not:
if (ev->op_info.api_name && !(ctx->curr_err_info->api_name))
or
if (NULL != ev->op_info.api_name && NULL == ctx->curr_err_info->api_name)?

soumagne (changes requested):
Sorry but except the renaming and clean up that I'm ok with, all the VOL/context changes are just awful. It's just doubling down on the existing context/library issues by introducing another layer of indirection to the VOL interface that not only is doubling the amount of code but I just have no idea what extra overhead it introduces. All these wrappers should be inline routines that call into the plugin callbacks and I don't think adding another func_enter macro when we already have so many different flavors of it is something that should be done. Also letting external VOLs getting / setting the API context is not something that we should do. Those changes are just here to "make things work", that's absolutely awful engineering work. There is just no design in the APIs that are exposed to the user / VOL developers.

src/H5CX.c
```diff
@@ -849,6 +849,52 @@ H5CX_push_special(void)
...
+ *
+ * Return:      <none>
+ *
+ * Programmer:  Quincey Koziol
+ *              June 9, 2021
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5CX_test_and_push(hbool_t *pushed)
```
soumagne, line 865: I understand why you have to do things like that but I really don't think that's the right direction and it's just adding more to the complexity of global states

gnuoyd:
I was trying to set up to profile the performance of the branch
hpc-io/hdf5:post_open_fix, but it doesn't look ready.  For example, the
compiler flags a glaring problem:

```
  CC       H5VL.lo
../../../manydsets/src/H5VL.c: In function ‘H5VLobject_is_native’:
../../../manydsets/src/H5VL.c:682:40: error: incompatible type for argument 2 of ‘H5VL_object_is_native’
     if (H5VL_object_is_native(vol_obj, &ret_value) < 0)
                                        ^~~~~~~~~~
In file included from ../../../manydsets/src/H5Fprivate.h:34,
                 from ../../../manydsets/src/H5Cprivate.h:33,
                 from ../../../manydsets/src/H5ACprivate.h:33,
                 from ../../../manydsets/src/H5CXprivate.h:26,
                 from ../../../manydsets/src/H5VL.c:31:
../../../manydsets/src/H5VLprivate.h:105:83: note: expected ‘H5VL_get_conn_lvl_’ {aka ‘enum H5VL_get_conn_lvl_t’} but argument is of type ‘hbool_t *’ {aka ‘_Bool *’}
 H5VL_object_is_native(const H5VL_object_t *obj, H5VL_get_conn_lvl_t lvl, hbool_t *is_native);
                                                 ~~~~~~~~~~~~~~~~~~~~^~~

../../../manydsets/src/H5VL.c:682:9: error: too few arguments to function ‘H5VL_object_is_native’
     if (H5VL_object_is_native(vol_obj, &ret_value) < 0)
         ^~~~~~~~~~~~~~~~~~~~~

```

I configure with `--enable-warnings-as-errors
--enable-build-mode=production` so that these problems cannot scroll by
unnoticed.  (I use production mode because GCC only finds some problems
with the deep analysis entailed by -O3.)

I may be able to try again next week, if someone will fix the branch.

Dave

src/H5Fpkg.h
```diff
@@ -413,6 +412,7 @@ H5_DLL herr_t H5F__get_info(H5F_t *f, H5F_info2_t *finfo);
 H5_DLL herr_t H5F__format_convert(H5F_t *f);
 H5_DLL herr_t H5F__start_swmr_write(H5F_t *f);
 H5_DLL herr_t H5F__close(H5F_t *f);
+H5_DLL herr_t H5F__close_obj(H5VL_object_t *file_vol_obj, void **request);
```
soumagne, line 415: That routine seems out of place wrt the overall API

src/H5VLint.c
```diff
@@ -1095,7 +1147,7 @@ H5VL_free_object(H5VL_object_t *vol_obj)
  *-------------------------------------------------------------------------
  */
 herr_t
-H5VL_object_is_native(const H5VL_object_t *obj, hbool_t *is_native)
+H5VL_object_is_native(const H5VL_object_t *obj, H5VL_get_conn_lvl_t lvl, hbool_t *is_native)
 {
```
soumagne, line 1151: I don't really understand why `H5VL_object_is_native()` is taking the `lvl` now, it should always be terminal? how can that depend on the level? I really think that adds complexity.

soumagne: I've also not seen anything other than LVL_TERM being used.

src/H5VLprivate.h
```diff
@@ -206,12 +208,12 @@ H5_DLL herr_t H5VL_datatype_optional_op(H5VL_object_t *vol_obj, H5VL_optional_ar
...
 
 /* File functions */
-H5_DLL void * H5VL_file_create(const H5VL_connector_prop_t *connector_prop, const char *name, unsigned flags,
-                               hid_t fcpl_id, hid_t fapl_id, hid_t dxpl_id, void **req);
-H5_DLL void * H5VL_file_open(H5VL_connector_prop_t *connector_prop, const char *name, unsigned flags,
-                             hid_t fapl_id, hid_t dxpl_id, void **req);
-H5_DLL herr_t H5VL_file_get(const H5VL_object_t *vol_obj, H5VL_file_get_args_t *args, hid_t dxpl_id,
-                            void **req);
+H5_DLL hid_t H5VL_file_create(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id, hid_t dxpl_id,
+                              void **req);
```
soumagne, line 212: I'm not too keen on having different return values for `file_create` and `file_open`. The rule was always that private versions would return private objects (even if we bypassed that rule several times). And now file_open returns both at the same time and it's not looking great.

src/H5Pdevelop.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * \dxpl_id{plist_id}
+ * \param[out]  new_api_ctx  Flag indicating API wrapper should create new API context
+ * \return \herr_t
+ *
+ * \details Retrieve "new API context" flag for plugin wrapper API calls.
+ *
+ */
+H5_DLL herr_t H5Pget_plugin_new_api_context(hid_t plist_id, hbool_t *new_api_ctx);
+
```
soumagne, line 74: I am really not enthusiastic about the idea of exposing internal defects to the VOL developers, we should not do that even if that's for the sake of getting something to work. I understand it seems harmless because only a handful of VOL developers would ever use it but we just keep adding more and more of these and it is not a good direction.

src/H5VLcallback.c
```diff
@@ -60,6 +62,541 @@ typedef struct H5VL_file_open_find_connector_t {
...
+typedef struct H5VL_object_optional_ctx_t {
+    /* DOWN: API routine parameters */
+    void *                   obj;
+    const H5VL_loc_params_t *loc_params;
+    H5VL_class_t *           cls;
+    H5VL_optional_args_t *   args;
+    hid_t                    dxpl_id;
+    void **                  req;
+} H5VL_object_optional_ctx_t;
+
```
soumagne, line 599: Defining all these additional structs is a lot of code duplication in my opinion

src/H5VLcallback.c
```diff
@@ -70,47 +607,69 @@ typedef herr_t (*H5VL_reg_opt_oper_t)(void *obj, const H5VL_class_t *cls, H5VL_o
 static void * H5VL__attr_create(void *obj, const H5VL_loc_params_t *loc_params, const H5VL_class_t *cls,
                                 const char *name, hid_t type_id, hid_t space_id, hid_t acpl_id, hid_t aapl_id,
                                 hid_t dxpl_id, void **req);
+static herr_t H5VL__attr_create_api_op(void *ctx);
```
soumagne, line 610: It really does not fit well in the overall API. I understand here it's private APIs but that does not mean there should not be an overall consistency to it.

src/H5VLcallback.c
```diff
@@ -371,6 +951,68 @@ H5VL__common_optional_op(hid_t id, H5I_type_t id_type, H5VL_reg_opt_oper_t reg_o
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5VL__common_optional_op() */
 
+/*-------------------------------------------------------------------------
+ * Function:    H5VL__common_api_op
+ *
+ * Purpose:     Provide common wrapping for VOL callback API routines.
+ *
```
soumagne, line 958: That should explain much more about what it's really doing and why it needs to do that. That routine looks very much like a hack to me and it might be cleaning up some of the previous code, it is still as bad in my opinion. That's a lot of effort I feel to do something that is still not good in the end, even if it does make external wrappers "work".

src/H5VLcallback.c
```diff
@@ -1039,6 +1681,37 @@ H5VL_attr_create(const H5VL_object_t *vol_obj, const H5VL_loc_params_t *loc_para
...
+
+    /* Call the corresponding internal VOL routine */
+    if (NULL == (ctx->ret_value =
+                     H5VL__attr_create(ctx->obj, ctx->loc_params, ctx->cls, ctx->name, ctx->type_id,
+                                       ctx->space_id, ctx->acpl_id, ctx->aapl_id, ctx->dxpl_id, ctx->req)))
+        HGOTO_ERROR(H5E_VOL, H5E_CANTCREATE, FAIL, "unable to create attribute")
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5VL__attr_create_api_op() */
```
soumagne, line 1713: It is a lot of code and repetitions for something that should just be inline / one line.

src/H5VLcallback.c
```diff
@@ -1067,13 +1741,28 @@ H5VLattr_create(void *obj, const H5VL_loc_params_t *loc_params, hid_t connector_
...
+    ctx.acpl_id    = acpl_id;
+    ctx.aapl_id    = aapl_id;
+    ctx.dxpl_id    = dxpl_id;
+    ctx.req        = req;
+    ctx.ret_value  = NULL;
+
+    /* Invoke common wrapper routine */
+    if (H5VL__common_api_op(dxpl_id, H5VL__attr_create_api_op, &ctx) < 0)
+        HGOTO_ERROR(H5E_VOL, H5E_CANTOPERATE, NULL, "unable to execute common wrapper operation")
+
```
soumagne, line 1760: It really adds here to the extra indirection. I don't feel like that's a good direction.

soumagne (commented):
I added more detailed comments. I also didn't mean to be rude in my previous message. I still think this PR, which was certainly a lot of effort, is not going in the right direction and is adding a lot of code to palliate wrappers defects (which should not exist in the first place) and passthrough VOL design. Looking more closely into it, the changes should not affect terminal VOL connectors though (unless I missed something since there's a lot to it). As I mentioned in previous comments, the main issue that I have with this PR is that it continues to build on defects and makes the overall code more and more complex, while also exposing some of these defects to external VOL developers.

fortran/test/t.c
```diff
@@ -155,3 +156,25 @@ nh5_env_nocleanup_c(int_f *status)
...
+ * Name:        h5vl_fapl_is_native_c
+ * Purpose:     Invokes the H5VL_fapl_is_native() operation
+ * Inputs:      fapl - file access property list
+ *              flag - flag indicating whether VOL connector stack in FAPL is the native connector
+ * Returns:     0 on success, -1 on failure
+ * Programmer:  Quincey Koziol
+ *              June 24, 2021
+ *---------------------------------------------------------------------------*/
+int_f
+h5vl_fapl_is_native_c(hid_t_f *fapl_id, int_f *flag)
```
epourmal, line 170: Since C API is H5VLfapl_is_native, why do we introduce "_" after "vl"? Typo? C wrappers usually use public C APIs.

gnuoyd:
I compared the user time before/after applying this change to develop, using the benchmark programs from LLNL (`testhdf5`) and Mathworks (`many_dsets`), and it's kind of a wash. `testhdf5` runs about 1% faster, `many_dsets` about 1% slower.

gnuoyd (changes requested):
I concur that this change is pushing the library code further in the wrong direction. The context (H5CX) is a creeping vine that needs to be dug up at its root. VOL appears to be in trouble.

PR #1165, opened 2021-11-04
Convert integer<->pointer conversions to errors

gnuoyd:
Make it an error if the library implicitly converts from integer to  pointer or from pointer to integer (-Werror=int-conversion).  Also, make it an error if the library explicitly converts to pointer from an integer of a different size (-Werror=int-to-pointer-cast).

### DIFF
--- a/config/gnu-warnings/error-general
+++ b/config/gnu-warnings/error-general
@@ -11,6 +11,8 @@
 -Werror=packed
 -Werror=pointer-sign
 -Werror=pointer-to-int-cast
+-Werror=int-to-pointer-cast
+-Werror=int-conversion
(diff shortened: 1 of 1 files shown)


byrnHDF (commented):
What about the compilers?

gnuoyd:
> What about the compilers?

What do you mean?

byrnHDF:
> > What about the compilers?
> 
> What do you mean?

clang, intel error files?

derobins (approved):

derobins:
This is fine but we should probably try to extrapolate this to clang, which probably has similar warnings.

lrknox (approved):

PR #1169, opened 2021-11-08
Create 2D arrays on the heap in a different way

gnuoyd:
Create 2D arrays on the heap by malloc'ing `struct { TYPE arr[ROWS][COLS]; }`.  This avoids the double-indirection through pointers and the additional memory of H5TEST_ALLOCATE_2D_ARRAY().

This change will safely quiet the cast warning that PR #1129 was intended to fix.


byrnHDF (approved):

lrknox:
Checks passed after format changes were committed:  https://github.com/HDFGroup/hdf5/actions/runs/1435686942.

lrknox (approved):

PR #1171, opened 2021-11-08
Avoid calling H5Ropen_object with a misaligned H5R_ref_t: copy the

gnuoyd:
raw H5R_ref_t bytes to a heap buffer that's known to have the right
alignment.

This should safely quiet the warning that #1130 was trying to quiet.


gnuoyd:
It would have been a waste of time to analyze this issue without producing a candidate fix, so here is the candidate fix.

tools/lib/h5tools.c
```diff
@@ -1890,15 +1890,20 @@ render_bin_output(FILE *stream, hid_t container, hid_t tid, void *_mem, hsize_t
                     hid_t        region_id    = H5I_INVALID_HID;
                     hid_t        region_space = H5I_INVALID_HID;
                     H5S_sel_type region_type;
+                    H5R_ref_t *  tref;
+                    size_t       trefsize = (sizeof(*tref) > size) ? sizeof(*tref) : size;
+
+                    if ((tref = HDmalloc(trefsize)) == NULL)
```
jhendersonHDF: Why can't we just use an H5R_ref_t stack variable here instead? Is there any particular reason why it would need to be malloc'ed? I'd prefer to avoid an unnecessary malloc if possible so that there's no potential for someone to accidentally introduce a memory leak at some point in the future.

gnuoyd: Using a stack variable is my preference.

H5R_ref_t is the size of a "default reference buffer," according to comments on H5R_REF_BUF_SIZE. I did not know if the size was variable, or if the default size was the maximum size, so I made the conservative change here.

jhendersonHDF: I'm not sure on the specifics either, but it seems like sizeof(H5R_ref_t) would generally be around sizeof(uint8_t[64]) since it just has a single union field. In which case, it seems like allocating `size` bytes for the type and then coercing it into an H5R_ref_t * might do bad things? Unless I'm reading the code wrong, that is.

jhendersonHDF: It appears that H5R_REF_BUF_SIZE should be a maximum size, so using a stack variable should be ok here, though it may be worth memsetting it to ensure that any garbage bytes are cleared out when size < sizeof(H5R_ref_t).

derobins: I would also like to avoid the malloc/free. I'm not crazy about the overhead of having to copy bytes, but I'm even less crazy about subtle bugs.

derobins (approved):

jhendersonHDF (approved):

hyoklee:
@gnuoyd , can you remove this warning as well?
```
h5tools.c:557:67: warning: cast from 'const void *' to 'struct H5FD_ros3_fapl_t *' \
drops const qualifier [-Wcast-qual]
                if (H5Pset_fapl_ros3(fapl_id, (H5FD_ros3_fapl_t *)vfd_info->info) <\
 0)
             
```

PR #1176, opened 2021-11-09
Improve efficiency of sorting by sorting just once for all writes that use a particular compound data type.

tonyhorrobin:
Hi, this Tony Horrobin from the University of Leeds.
- Add an index array for sorting by name.
- Use it instead of local arrays in H5T_cmp().
- Remember to deep copy it to avoid a SEGV due to using freed memory.

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -3326,6 +3326,7 @@ H5T__create(H5T_class_t type, size_t size)
             if (type == H5T_COMPOUND) {
                 dt->shared->u.compnd.packed    = FALSE; /* Start out unpacked */
                 dt->shared->u.compnd.memb_size = 0;
+                dt->shared->u.compnd.idx_name=NULL;
             } /* end if */
             else if (type == H5T_OPAQUE)
                 /* Initialize the tag in case it's not set later.  A null tag will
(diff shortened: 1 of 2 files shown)


tonyhorrobin:
Sorry for my lack of github knowledge.
How do I accept the formatting changes proposed by the clang-format check?
I tried pasting them into a patch file but it is appended to one line.

Cheers,

-Tony

gnuoyd:
Thank you for the patch, Tony.

The H5T_cmp code is in need of a clean-up. As you can see, bubblesort is open-coded in there four separate times!  I have applied your change on a branch where I have performed the clean-up. I will share the branch with you when it is ready for review.

BTW, I found a bug by running our test suite with `make check`. Here and there I added code to invalidate the cached member-name order, `idx_name`. That made the tests pass.

Do you think you can share a small program that demonstrates the speed-up you achieved? It will help us to catch regressions.

tonyhorrobin:
Hi,

I'll have a go at a performance regression test soon.  Does it have to be in C or can I use C++ too?

Cheers,

Tony

gnuoyd:
Tests should be in C.

PR #1387 contains my H5T_cmp refactorization and your change to cache member-name sorted indices.

PR #1183, opened 2021-11-10
Parallel rank0 deadlock fixes

### LINKED ISSUE #118: Deadlock in H5FD__mpio_open due to bad MPI_File_get_size
While debugging another issue unrelated to HDF5, I think I tripped up a deadlock condition.  In my test, it seems the MPI_File_get_size call returns an error here:
https://github.com/HDFGroup/hdf5/blob/c56464fc36a78167c22b84e4cfef0e0c2aafce80/src/H5FDmpio.c#L828-L832

That causes rank 0 to jump to the bottom of the function to close the file, which is a collective:
https://github.com/HDFGroup/hdf5/blob/c56464fc36a78167c22b84e4cfef0e0c2aafce80/src/H5FDmpio.c#L858-L861

While other ranks are stuck in a different collective here:
https://github.com/HDFGroup/hdf5/blob/c56464fc36a78167c22b84e4cfef0e0c2aafce80/src/H5FDmpio.c#L835

Also, I noticed that there may be other patterns like this while grepping the file to look for this location, like maybe:
https://github.com/HDFGroup/hdf5/blob/c56464fc36a78167c22b84e4cfef0e0c2aafce80/src/H5FDmpio.c#L1223-L1229

jhendersonHDF:
This PR rewrites several places where rank 0 could skip past collective MPI operations on failure, leading to deadlocks. There are a few cases after this rewrite where rank 0 participating in, e.g. an MPI_Bcast after failure may lead to library crash from segfaults or similar, but this seems to me like a better alternative than hanging.


src/H5ACmpio.c
```diff
@@ -428,18 +430,23 @@ H5AC__broadcast_clean_list(H5AC_t *cache_ptr)
...
-        udata.addr_buf_ptr = addr_buf_ptr;
-        udata.u            = 0;
-
-        /* Free all the clean list entries, building the address list in the callback */
-        /* (Callback also removes the matching entries from the dirtied list) */
-        if (H5SL_free(aux_ptr->c_slist_ptr, H5AC__broadcast_clean_list_cb, &udata) < 0)
-            HGOTO_ERROR(H5E_CACHE, H5E_CANTFREE, FAIL, "Can't build address list for clean entries")
+        if (NULL == (addr_buf_ptr = (haddr_t *)H5MM_malloc(buf_size))) {
+            /* Push an error, but still participate in following MPI_Bcast */
+            HDONE_ERROR(H5E_CACHE, H5E_CANTALLOC, FAIL, "memory allocation failed for addr buffer")
```
jhendersonHDF, line 435: This is a case where the MPI implementation may crash when MPI_Bcast is called with a NULL addr_buf_ptr. At least, this seems to be the case for OpenMPI. However, if memory allocation fails here, there isn't a whole lot else we can do, and rank 0 still needs to participate in the BCast in a similar manner as the other ranks.

src/H5CX.c
```diff
@@ -1397,9 +1397,7 @@ H5CX_set_apl(hid_t *acspl_id, const H5P_libclass_t *libclass,
 
         /* If parallel is enabled and the file driver used is the MPI-IO
          * VFD, issue an MPI barrier for easier debugging if the API function
-         * calling this is supposed to be called collectively. Note that this
-         * happens only when the environment variable H5_COLL_BARRIER is set
-         * to non 0.
```
jhendersonHDF, line 1402: Just a minor documentation issue here I noticed while fixing these other issues: The H5_COLL_BARRIER environment variable doesn't exist

src/H5Cimage.c
```diff
@@ -997,6 +997,9 @@ H5C__read_cache_image(H5F_t *f, H5C_t *cache_ptr)
 #endif /* H5_HAVE_PARALLEL */
 
             /* Read the buffer (if serial access, or rank 0 of parallel access) */
+            /* NOTE: if this block read is being performed on rank 0 only, throwing
+             * an error here will cause other ranks to hang in the following MPI_Bcast.
+             */
```
jhendersonHDF, line 1002: For now, I'm just leaving a note here as properly fixing the potential hang is a little tricky and very messy with #ifdefs

derobins (approved):

jhendersonHDF:
@lrknox please give @jrmainzer a chance to review this before merging, since I believe most of the changes occur in cache code.

src/H5AC.c
```diff
@@ -1636,9 +1636,14 @@ H5AC_unprotect(H5F_t *f, const H5AC_class_t *type, haddr_t addr, void *thing, un
...
 
-        if (deleted && aux_ptr->mpi_rank == 0)
-            if (H5AC__log_deleted_entry((H5AC_info_t *)thing) < 0)
-                HGOTO_ERROR(H5E_CACHE, H5E_CANTUNPROTECT, FAIL, "H5AC__log_deleted_entry() failed")
+        if (deleted && aux_ptr->mpi_rank == 0) {
+            if (H5AC__log_deleted_entry((H5AC_info_t *)thing) < 0) {
+                /* If we fail to log the deleted entry, push an error but still
+                 * participate in a possible sync point ahead
+                 */
+                HDONE_ERROR(H5E_CACHE, H5E_CANTUNPROTECT, FAIL, "H5AC__log_deleted_entry() failed")
```
jrmainzer, line 1644: I was of the impression that HDONE_ERROR is only used after the done: tag.  Is this this correct?

src/H5ACmpio.c
```diff
@@ -304,8 +304,10 @@ H5AC__broadcast_candidate_list(H5AC_t *cache_ptr, unsigned *num_entries_ptr, had
          * are used to receiving from process 0, and also load it
          * into a buffer for transmission.
          */
-        if (H5AC__copy_candidate_list_to_buffer(cache_ptr, &chk_num_entries, &haddr_buf_ptr) < 0)
-            HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "Can't construct candidate buffer.")
+        if (H5AC__copy_candidate_list_to_buffer(cache_ptr, &chk_num_entries, &haddr_buf_ptr) < 0) {
+            /* Push an error, but still participate in following MPI_Bcast */
+            HDONE_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "Can't construct candidate buffer.")
```
jrmainzer, line 309: call to HDONE_ERROR before done: flag?

src/H5AC.c
jhendersonHDF: In general yes. However the main upside to the macro is just that it doesn't do the typical goto. It still pushes an error message to the stack, sets the err_occurred variable and sets the return value. The HERROR macro only pushes an error message to the stack, without setting those needed variables. The HCOMMON_ERROR macro might be a good alternative, but the header comment for it says that it shouldn't need to be used outside H5Eprivate.h.

jrmainzer: If I understand what is going on here, you are trying to flag an error but not jump to done so as to participate in the upcoming sync point.

Using HDONE_ERROR doesn't seem the right way to do this.

For starters, we have comments expressly stating that HDONE_ERROR is only to be used after the done: flag.  Breaking this rule without explanation is bound to cause confusion.

I note that we do this already in H5Fint.c in H5F_flush_phase1, H5F_flush_phase2, and H5F_dest.  However, we add the comment "/* Push error, but keep going */".  At a minimum, we should have this.  It would be good to explain why.  

One could also argue that we should only be using HDONE_ERROR in this context in parallel builds.  I'm not sure that this is a good idea, but it might be worth thinking about.

src/H5ACmpio.c
jrmainzer: Also, if we can't copy the candidate list to buffer, I expect we will crash and burn regardless.  Thus I don't see much point in participating in the Bcast.

src/H5AC.c
jhendersonHDF: In general, I tried to add a comment to all these places to the same effect, mentioning that we push an error but keep participating in the collective operation. If any of those comments aren't too clear on that, please let me know so I can clarify. I do agree that HDONE_ERROR isn't necessarily meant for this, but I believe an alternative macro would be implemented exactly the same, just with a different name.

Perhaps "DONE_ERROR" was not the best name, because there should be no harm from calling it outside the done: tag. Seems to me like the reasoning was just the fact that calling HGOTO_ERROR after the done: tag would of course result in an infinite loop.

src/H5ACmpio.c
jhendersonHDF: Yes, it will almost certainly crash. On the other hand, it will probably crash in a useful way from a debugging perspective. The alternative would be a hang, which seemed the worse of the two options, but I could be convinced either way.

src/H5ACmpio.c
jrmainzer: Per comment above, I don't see much point in worrying about the Bcast if rank 0 can't create the buffer.

If you want to recover, the better solution would be to allocate the buffer before we Bcast the number of entries in it, and pretend entry_count is zero if the malloc fails.  That said, if rank 0 can't allocate a smallish buffer, the computation is hosed regardless.

soumagne (approved):

lrknox (approved):

PR #1188, opened 2021-11-11
Quiets most const warnings in the fractal heap code

derobins:
The biggest problem here is the 'generic operation' scheme that combines a const void pointer to arbitrary data + a callback that can sometimes modify said arbitrary data, thus undermining the const-ness in certain cases. Since nobody is going to rearchitect this scheme anytime soon, I've at least quieted the warnings that we know are not a problem since the "sometimes const" pointers won't be modified when they point to const data given the existing code.

In case anyone is wondering, making the operation callbacks simply take a non-const pointer is a bit like pulling on the loose yarn in a sweater. I started going down this road, but I think this is a more minimalist and cleaner solution for now.

I left the file pointer const warnings in the cache client code alone. That's a can of worms I'm unwilling to touch at this time and I also don't want to just glibly turn off the cast warnings.

### DIFF
--- a/src/H5HFman.c
+++ b/src/H5HFman.c
@@ -487,10 +487,16 @@ H5HF__man_write(H5HF_hdr_t *hdr, const uint8_t *id, const void *obj)
     HDassert(id);
     HDassert(obj);
 
-    /* Call the internal 'op' routine routine */
-    /* (Casting away const OK - QAK) */
+    /* Call the internal 'op' routine routine
+     *
+     * In this case, the callback operation needs to modify the obj buffer that
+     * was passed in as const. We quiet the warning here because an obj pointer
+     * that was originally const should *never* arrive here.
+     */
+    H5_GCC_CLANG_DIAG_OFF("cast-qual")
     if (H5HF__man_op_real(hdr, id, H5HF__op_write, (void *)obj, H5HF_OP_MODIFY) < 0)
         HGOTO_ERROR(H5E_HEAP, H5E_CANTOPERATE, FAIL, "unable to operate on heap object")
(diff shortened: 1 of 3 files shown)


byrnHDF (approved):

lrknox (approved):

src/H5HFman.c
```diff
@@ -487,10 +487,16 @@ H5HF__man_write(H5HF_hdr_t *hdr, const uint8_t *id, const void *obj)
...
+    /* Call the internal 'op' routine routine
+     *
+     * In this case, the callback operation needs to modify the obj buffer that
+     * was passed in as const. We quiet the warning here because an obj pointer
+     * that was originally const should *never* arrive here.
+     */
+    H5_GCC_CLANG_DIAG_OFF("cast-qual")
     if (H5HF__man_op_real(hdr, id, H5HF__op_write, (void *)obj, H5HF_OP_MODIFY) < 0)
         HGOTO_ERROR(H5E_HEAP, H5E_CANTOPERATE, FAIL, "unable to operate on heap object")
+    H5_GCC_CLANG_DIAG_ON("cast-qual")
```
gnuoyd, line 499: This is quite a wordy way to de-const a pointer.

Other C projects de-const pointers without making the compiler warn by using a macro or function that casts through uintptr_t. Something like this should do the trick,

```
static inline void *
deconst(const void *p)
{
    return (void *)(uintptr_t)p;
}
```

The H5HF__man_op_real() call would turn to `H5HF__man_op_real(hdr, id, H5HF__op_write, deconst(obj), H5HF_OP_MODIFY)`.

PR #1189, opened 2021-11-12
Fix MPI Comm and Info leak in H5FD__mpio_delete

jhendersonHDF:


soumagne (approved):

soumagne:
I feel like you're missing an if comm not equal to comm world or comm null though.

jhendersonHDF:
> I feel like you're missing an if comm not equal to comm world or comm null though.

We updated the two routines a while back to check against MPI_COMM_WORLD, MPI_COMM_NULL and MPI_INFO_NULL before freeing the objects, so this should be ok.

derobins (approved):

PR #1195, opened 2021-11-15
Open bsd fixes

seanm:


seanm:
Hmmm, not sure why I'm seeing 8 commits here, I expect just 2.  Any git experts know?

Anyway, not sure about the implications of defining `_BSD_SOURCE` for other OSes, but see below for how things go on OpenBSD:

```
builder3$ cat test.c
#include <fcntl.h>

int main (void)
{
  flock(0,0);
  return 0;
}

builder3$ cc -fsyntax-only test.c 
builder3$ cc -fsyntax-only -D_BSD_SOURCE test.c
builder3$ cc -fsyntax-only -D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L test.c
builder3$ cc -fsyntax-only -D_POSIX_C_SOURCE=200809L test.c               
test.c:5:3: warning: implicit declaration of function 'flock' is invalid in C99 [-Wimplicit-function-declaration]
  flock(0,0);
  ^
1 warning generated.
```

i.e. the definition of `_POSIX_C_SOURCE=200809L` makes `flock` unavailable unless `_BSD_SOURCE` is also defined.

derobins:
Instead of globally defining _BSD_SOURCE, we should have a fix that only affects OpenBSD.

configure.ac
```diff
@@ -1410,7 +1410,7 @@ case "$host_cpu-$host_vendor-$host_os" in
     ##
     ## POSIX feature information can be found in the gcc manual at:
     ## http://www.gnu.org/s/libc/manual/html_node/Feature-Test-Macros.html
-    H5_CPPFLAGS="-D_POSIX_C_SOURCE=200809L $H5_CPPFLAGS"
+    H5_CPPFLAGS="-D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L $H5_CPPFLAGS"
```
derobins: Did you test this? The change here is being applied to Linux.

seanm: I tested it on OpenBSD, but not other OSes.  The CI here on github seems to be green.  But as I said, I'm not sure if having this defined could be detrimental on other OSes or not...

byrnHDF: > I tested it on OpenBSD, but not other OSes. The CI here on github seems to be green. But as I said, I'm not sure if having this defined could be detrimental on other OSes or not...

I think this was removed because it was flagged as a warning in newer gcc builds? Anyway it should probably be just defined for openBSD.

seanm: I wouldn't be surprised if it was also needed for the other BSDs... do you do any testing with them?  I could also try on FreeBSD...

byrnHDF: So gcc 12 outputs the following for this:
# warning "_BSD_SOURCE  is deprecated, use _DEFAULT_SOURCE"

seanm: Some quick googling suggests this warning is from glibc.

Alas, using `_DEFAULT_SOURCE` instead of `_BSD_SOURCE` does not work for OpenBSD:

```
builder3$ cc -fsyntax-only -D_DEFAULT_SOURCE -D_POSIX_C_SOURCE=200809L test.c
test.c:5:3: warning: implicit declaration of function 'flock' is invalid in C99 [-Wimplicit-function-declaration]
  flock(0,0);
  ^
```

From what I can tell, `flock` is *not* part of POSIX, but rather is from BSD. See: https://gavv.github.io/articles/file-locks/#bsd-locks-flock

I suspect OpenBSD interprets `_POSIX_C_SOURCE=200809L` to mean to include *only* things in POSIX, and thus `flock` becomes hidden.  OTOH, glibc seems to interpret `_POSIX_C_SOURCE=200809L` differently.

derobins: Including -D_BSD_SOURCE on OpenBSD is fine, I'd just like to limit it to that OS. Also, we shouldn't be applying that flag to any untested operating systems. i.e., unless we've tested on NetBSD or FreeBSD, we shouldn't be blindly applying the flag there.

seanm: I don't know much about cmake/autotools, how can this be done?

byrnHDF: CMake has this line in CMakeDetermineSystem.cmake:
    elseif(CMAKE_HOST_SYSTEM_NAME MATCHES "OpenBSD")

I think it sets this from a uname call so, CMake could check this var and add the the define.

byrnHDF: I think autotools uses a "$host_os" var - at least in setting up flags. I think this var is set in configure.ac.

seanm: OK, please see my new change.

hyoklee (approved):
This patch is really good.

seanm:
> This patch is really good.

Kind of you to say. Thanks for reviewing.

hyoklee:
@seanm By the way, were you able to make all tests pass on OpenBSD 7.0? I get a segmentation fault error when I use autotools. I get SWMR test timeout error when I use CTest.

config/cmake_ext_mod/ConfigureChecks.cmake
```diff
@@ -223,65 +223,63 @@ set (LINUX_LFS 0)
...
-      if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.15.0")
-          message (VERBOSE "${msg}... no")
-      endif ()
-      file (APPEND ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log
-          "Test TEST_LFS_WORKS Compile failed\n"
-      )
-    endif ()
-  endif ()
-  set (CMAKE_REQUIRED_DEFINITIONS ${CMAKE_REQUIRED_DEFINITIONS} ${HDF_EXTRA_FLAGS})
+  if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")
```
byrnHDF: What about the MINGW option? MINGW could be hosted on linux or windows.

seanm: You're saying it should be re-added to this `if` statement then?

seanm:
> @seanm By the way, were you able to make all tests pass on OpenBSD 7.0? I get a segmentation fault error when I use autotools. I get SWMR test timeout error when I use CTest.

On OpenBSD 7.0 on PowerPC (I didn't try Intel) my results are:

```
The following tests FAILED:
	  5 - H5TEST-testhdf5-select (Timeout)
	345 - H5LS-tattrreg_be (Failed) -> https://github.com/HDFGroup/hdf5/issues/1214
	348 - H5LS-tdataregbe (Failed)
```

I've only tried with cmake.

config/cmake_ext_mod/ConfigureChecks.cmake
byrnHDF: I'm thinking that it should be there.

seanm: OTOH, in https://github.com/HDFGroup/hdf5/blob/8b3c09f780e31ba96b9f5c370a0c19f8a0dcfe6c/configure.ac#L1424 the definition of `_POSIX_C_SOURCE` seems strictly limited to linux.

byrnHDF: Which is why mingw needs to be added back - mingw trys to be both linux and windows. linux tools to create windows binaries.

seanm: I didn't explicitly remove the `MINGW` part.  I changed:

`  if (NOT ${HDF_PREFIX}_HAVE_SOLARIS AND NOT ${HDF_PREFIX}_HAVE_DARWIN)`

to

`  if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")`

I guess MINGW won't have gone in the old `if`. But will it go in the new `if`?  The [CMAKE_HOST_SYSTEM_NAME](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_NAME.html) docs aren't too helpful.

Do you have MINGW to test?

byrnHDF: According to CMake docs, they state that MINGW is a compiler type - so probably okay.

seanm: OK great. So is this mergeable now?

byrnHDF: One more thing to think about - using CMAKE_HOST_SYSTEM_NAME would that cause an issue with cross-compiling?

seanm: Hmm, indeed, I think using `CMAKE_SYSTEM_NAME` may be better.

byrnHDF:
@seanm  have you seen this SO item: https://stackoverflow.com/questions/29201515/what-does-d-default-source-do

byrnHDF:
I believe we removed the BSD_SOURCE because of: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"

seanm:
> I believe we removed the BSD_SOURCE because of: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"

Have you looked at my newer commits though?  I'm no longer #defining _BSD_SOURCE or anything special on BSD.  By changing:

` if (NOT ${HDF_PREFIX}_HAVE_SOLARIS AND NOT ${HDF_PREFIX}_HAVE_DARWIN)`

to

` if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")`

the BSDs no longer go into this `if` and don't do any of that stuff than really should be linux-only.  So on linux, nothing changes, and on BSDs none of that stuff is done at all, which solves the whole `flock` issue.

byrnHDF:
> > I believe we removed the BSD_SOURCE because of: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"
> 
> Have you looked at my newer commits though? I'm no longer #defining _BSD_SOURCE or anything special on BSD. By changing:
> 
> ` if (NOT ${HDF_PREFIX}_HAVE_SOLARIS AND NOT ${HDF_PREFIX}_HAVE_DARWIN)`
> 
> to
> 
> ` if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")`
> 
> the BSDs no longer go into this `if` and don't do any of that stuff than really should be linux-only. So on linux, nothing changes, and on BSDs none of that stuff is done at all, which solves the whole `flock` issue.

OK, I guess I was looking at old source or something. Apologies. And I agree CMAKE_SYSTEM_NAME might be better.

seanm:
>And I agree CMAKE_SYSTEM_NAME might be better.

Done.

byrnHDF (approved):

lrknox (approved):

PR #1206, opened 2021-11-17
Unify handling of collective metadata reads status

jhendersonHDF:
This PR unifies the handling of the collective metadata reads status, so that one doesn't need to check both the file-global flag and the API context. By doing this, it also fixes file-global collective metadata reads that became broken in 1.10.5.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -1083,6 +1083,27 @@ Bug Fixes since HDF5-1.12.0 release
 
       (DER - 2021/11/23, HDFFV-11286)
 
+    - Unified handling of collective metadata reads to correctly fix old bugs
+
+      Due to MPI-related issues occurring in HDF5 from mismanagement of the
+      status of collective metadata reads, they were forced to be disabled
+      during chunked dataset raw data I/O in the HDF5 1.10.5 release. This
+      wouldn't generally have affected application performance because HDF5
+      already disables collective metadata reads during chunk lookup, since
+      it is generally unlikely that the same chunks will be read by all MPI
+      ranks in the I/O operation. However, this was only a partial solution
+      that wasn't granular enough.
+
+      This change now unifies the handling of the file-global flag and the
+      API context-level flag for collective metadata reads in order to
+      simplify querying of the true status of collective metadata reads. Thus,
+      collective metadata reads are once again enabled for chunked dataset
+      raw data I/O, but manually controlled at places where some processing
+      occurs on MPI rank 0 only and would cause issues when collective
+      metadata reads are enabled.
+
+      (JTH - 2021/11/16, HDFFV-10501/HDFFV-10562)
+
     - Fixed several potential MPI deadlocks in library failure conditions
 
       In the parallel library, there were several places where MPI rank 0
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -1518,17 +1518,26 @@ H5C_insert_entry(H5F_t *f, const H5C_class_t *type, haddr_t addr, void *thing, u
 
 #ifdef H5_HAVE_PARALLEL
     if (H5F_HAS_FEATURE(f, H5FD_FEAT_HAS_MPI))
-        coll_access = H5CX_get_coll_metadata_read();
+        coll_access = H5F_get_coll_metadata_reads(f);
 
     entry_ptr->coll_access = coll_access;
     if (coll_access) {
         H5C__INSERT_IN_COLL_LIST(cache_ptr, entry_ptr, FAIL)
 
         /* Make sure the size of the collective entries in the cache remain in check */
-        if (cache_ptr->max_cache_size * 80 < cache_ptr->coll_list_size * 100)
-            if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
-                HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "can't clear collective metadata entries")
-    } /* end if */
+        if (H5P_USER_TRUE == H5F_COLL_MD_READ(f)) {
+            if (cache_ptr->max_cache_size * 80 < cache_ptr->coll_list_size * 100) {
+                if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
+                    HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "can't clear collective metadata entries")
+            } /* end if */
+        }     /* end if */
+        else {
+            if (cache_ptr->max_cache_size * 40 < cache_ptr->coll_list_size * 100) {
+                if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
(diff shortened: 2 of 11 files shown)


src/H5C.c
```diff
@@ -1518,17 +1518,26 @@ H5C_insert_entry(H5F_t *f, const H5C_class_t *type, haddr_t addr, void *thing, u
...
+            if (cache_ptr->max_cache_size * 80 < cache_ptr->coll_list_size * 100) {
+                if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
+                    HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "can't clear collective metadata entries")
+            } /* end if */
+        }     /* end if */
+        else {
+            if (cache_ptr->max_cache_size * 40 < cache_ptr->coll_list_size * 100) {
+                if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
+                    HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "can't clear collective metadata entries")
+            } /* end if */
```
jhendersonHDF, line 1538: The above code was here in 1.10.5, but I (probably mistakenly) removed it at the time that I made changes to collective metadata reads after the introduction of API contexts. I'm just restoring it now that collective metadata reads should be functional again.

src/H5Dmpio.c
```diff
@@ -808,10 +808,6 @@ H5D__chunk_collective_io(H5D_io_info_t *io_info, const H5D_type_info_t *type_inf
     HDassert(type_info);
     HDassert(fm);
 
-    /* Disable collective metadata reads for chunked dataset I/O operations
-     * in order to prevent potential hangs */
-    H5CX_set_coll_metadata_read(FALSE);
```
jhendersonHDF, line 813: This should no longer be necessary after these changes. The disabling of collective metadata reads is on a more granular scale rather than taking a hammer to it here.

src/H5Fmpi.c
```diff
@@ -402,4 +403,125 @@ H5F_mpi_retrieve_comm(hid_t loc_id, hid_t acspl_id, MPI_Comm *mpi_comm)
...
+ *              these settings are restored with another call to this
+ *              routine once the bracketed operation is complete.
+ * !! WARNING !!
+ *
+ * Return:      Nothing
+ *
+ *-------------------------------------------------------------------------
+ */
+void
+H5F_set_coll_metadata_reads(H5F_t *file, H5P_coll_md_read_flag_t *file_flag, hbool_t *context_flag)
```
jhendersonHDF, line 499: This API may be a little clunky, but I intentionally tried to make it more difficult to use because modifying the collective metadata reads status internally should be something that's done rarely and with great care. Hopefully the warning in the header comment is enough to make the caller think carefully beforehand.

src/H5Z.c
```diff
@@ -621,13 +616,8 @@ H5Z__flush_file_cb(void *obj_ptr, hid_t H5_ATTR_UNUSED obj_id, void H5_ATTR_PARA
                 /* Set the "sanity checked" flag */
                 object->sanity_checked = TRUE;
             } /* end if */
-
-            /* Check whether to use the collective metadata read DXPL */
-            coll_md_read = H5F_COLL_MD_READ(f);
-            if (H5P_USER_TRUE == coll_md_read)
-                H5CX_set_coll_metadata_read(TRUE);
```
jhendersonHDF, line 628: This code used to see if the user had set the file-global flag for metadata reads, since H5Zunregister doesn't take an Access Property List. It would then set collective metadata reads on the context, so that the flush below takes them into account. However, with everything now using the new H5F_get_coll_metadata_reads() routine, the flushing below should be directly aware of if the user set the file-global flag.

testpar/t_coll_md_read.c
```diff
@@ -34,10 +34,9 @@
```
jhendersonHDF, line 34: All the changes in this file were just to make one test more reproducible. It used to be that the test would only reliably reproduce the issue on exactly 2 or 3 MPI ranks (see Condition 3 from the old code below), but I refactored the test a bit so it should be reproducible for any number of MPI ranks.

release_docs/RELEASE.txt
```diff
@@ -1039,6 +1039,27 @@ Bug Fixes since HDF5-1.12.0 release
...
+
+      This change now unifies the handling of the file-global flag and the
+      API context-level flag for collective metadata reads in order to
+      simplify querying of the true status of collective metadata reads. Thus,
+      collective metadata reads are once again enabled for chunked dataset
+      raw data I/O, but manually controlled at places where some processing
+      occurs on MPI rank 0 only and would cause issues when collective
+      metadata reads are enabled.
+
+      (JTH - 2021/11/16)
```
lrknox: Is there a list of the jira IDs for the old bugs that were fixed?

jhendersonHDF: This would be HDFFV-10501 and HDFFV-10562. They were "fixed" previously and probably put in the RELEASE.txt that time, but this fixes them in a more correct way.

byrnHDF (approved):

release_docs/RELEASE.txt
lrknox: The jira issues should be included in the release note.

jhendersonHDF: Added

src/H5C.c
jrmainzer: This is one of the places where the collective read code can get into trouble with the adaptive cache resize code.  Briefly, each metadata cache on each rank can expand and contract independently depending on metadata access patterns on the rank in question.  Thus there is no guarantee that cache_ptr->max_cache_size is uniform across ranks.

src/H5C.c
```diff
@@ -2595,11 +2604,19 @@ H5C_protect(H5F_t *f, const H5C_class_t *type, haddr_t addr, void *udata, unsign
...
+                if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
+                    HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, NULL, "can't clear collective metadata entries")
+        } /* end if */
+        else {
+            if (cache_ptr->max_cache_size * 40 < cache_ptr->coll_list_size * 100)
+                if (H5C_clear_coll_entries(cache_ptr, TRUE) < 0)
+                    HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, NULL, "can't clear collective metadata entries")
+        } /* end else */
+    }     /* end if */
+#endif    /* H5_HAVE_PARALLEL */
```
jrmainzer, line 2624: Again, cache_ptr->max_cache_size need not be uniform across all ranks.

soumagne (approved):

soumagne:
Looks good, I like the clarification that these new routines make.

gnuoyd (approved):
I skimmed the changes. Nothing stands out.

jhendersonHDF:
Fixed conflicts with the develop branch

PR #1209, opened 2021-11-18
Stop lying about H5S_t const-ness

derobins:
Hyperslabs can be reworked inside several H5S callbacks, making H5S_t
non-const in some places where it is marked const. This change switches
these incorrectly const H5S_t pointer parameters and variables to
non-const where appropriate.

### DIFF
--- a/src/H5Pdxpl.c
+++ b/src/H5Pdxpl.c
@@ -2213,10 +2213,18 @@ H5P__dxfr_dset_io_hyp_sel_cmp(const void *_space1, const void *_space2, size_t H
         if (TRUE != H5S_extent_equal(*space1, *space2))
             HGOTO_DONE(-1);
 
-        /* Compare the selection "shape" of the dataspaces */
-        /* (Error & not-equal count the same) */
-        if (TRUE != H5S_select_shape_same(*space1, *space2))
+        /* Compare the selection "shape" of the dataspaces
+         * (Error & not-equal count the same)
+         *
+         * Since H5S_select_shape_same() can result in the dataspaces being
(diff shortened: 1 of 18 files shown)


derobins:
This fixes the majority of the pernicious const warnings in H5S and H5D. I left a few stragglers so that this PR focused only on a single const case.

src/H5Pdxpl.c
```diff
@@ -2213,10 +2213,18 @@ H5P__dxfr_dset_io_hyp_sel_cmp(const void *_space1, const void *_space2, size_t H
...
+         * Since H5S_select_shape_same() can result in the dataspaces being
+         * rebuilt, the parameters are not const which makes it impossible
+         * to match the cmp prototype. Since we need to compare them,
+         * we quiet the const warning.
+         */
+        H5_GCC_CLANG_DIAG_OFF("cast-qual")
+        if (TRUE != H5S_select_shape_same((H5S_t *)*space1, (H5S_t *)*space2))
             HGOTO_DONE(-1);
+        H5_GCC_CLANG_DIAG_ON("cast-qual")
     } /* end if */
```
derobins, line 2228: This is the one case where there wasn't much I could do. Being honest about const here prevents us from using this function as a cmp callback so I added the comment and silenced the cast.

hyoklee (approved):

derobins:
There's no real code changes here, btw - this is all about removing 'const' where it's lying.

byrnHDF (approved):

lrknox (approved):

gnuoyd (approved):
LGTM.

PR #1211, opened 2021-11-18
Fixes const warnings in H5ES package

derobins:


derobins:
IIRC, these were changed at some point to not be duplicated and instead point directly to internal const strings.

jhendersonHDF (approved):

byrnHDF (approved):

PR #1213, opened 2021-11-21
Only check HDF5_DRIVER in MPI I/O VFD if VFD isn't initialized

jhendersonHDF:


src/H5FDmpio.c
```diff
@@ -214,21 +214,22 @@ H5FD_mpio_init(void)
...
-        /* Initialize MPI if not already initialized */
-        if (MPI_SUCCESS != MPI_Initialized(&mpi_initialized))
-            HGOTO_ERROR(H5E_VFL, H5E_UNINITIALIZED, H5I_INVALID_HID, "can't check if MPI is initialized")
-        if (!mpi_initialized) {
-            if (MPI_SUCCESS != MPI_Init(NULL, NULL))
-                HGOTO_ERROR(H5E_VFL, H5E_CANTINIT, H5I_INVALID_HID, "can't initialize MPI")
-            H5FD_mpi_self_initialized = TRUE;
+        /* Check if MPI driver has been loaded dynamically */
+        env = HDgetenv(HDF5_DRIVER);
+        if (env && !HDstrcmp(env, "mpio")) {
```
jhendersonHDF, line 222: Some random profiling for parallel compression showed me that over the course of 44,674 calls to H5FD__mpio_get_eoa, 98% of the time was spent in `getenv` due to the fact that essentially every function in the MPI I/O VFD has a line like

`HDassert(H5FD_MPIO == file->pub.driver_id);`

which will call H5FDperform_init and hence the VFD's init routine H5FD_mpio_init. The MPI I/O VFD is the _only_ VFD that has these asserts, outside of the other two changes in this PR for the sec2 and splitter VFDs, where I added a similar assert but am now removing it. The other drivers occasionally make use of their respective macro, e.g. `H5FD_SEC2`, to check for the correct driver, but usually only in relatively infrequently called places.

To not disturb things too much, I'm leaving all the asserts in the MPI I/O VFD, but moving the above getenv call inside the block that checks whether the VFD has been registered or not. We really only need to check the environment once on first registration. Other VFDs like the sec2 VFD also call getenv in their init call to check for things like file locking status, so we should keep an eye out for other cases like this.

lrknox (approved):

derobins (approved):

src/H5FDmpio.c
derobins: I feel like the extra MPI I/O VFD asserts are overkill.

PR #1216, opened 2021-11-22
Fix plugin interface initialization ordering

jhendersonHDF:
Make sure plugin interface is initialized before property list interface phase 2 so that VFD plugins can be loaded correctly


soumagne (approved):

gnuoyd (approved):
LGTM.

jya-kmu (approved):

PR #1250, opened 2021-12-02
Subfile name fixes

brtnfld:
Fixed filenames in subfile's .config and added leading 0s in IOC id number.


rawarren (approved):

src/H5FDsubfile_mpi.c
```diff
@@ -31,8 +31,8 @@ static double sf_queue_delay_time = 0.0;
  * intend to use the user defined HDF5 filename for a
  * zeroth subfile as well as for all metadata.
  */
-#define SF_FILENAME_TEMPLATE ".subfile_%ld_%d_of_%d"
-
+//#define SF_FILENAME_TEMP_TEMPLATE ".subfile_%%ld_%%0%dd_of_%%d"
```
derobins: Probably don't want to check in commented-out code

src/H5FDsubfile_mpi.c
```diff
@@ -2440,7 +2467,8 @@ get__subfile_name(subfiling_context_t *sf_context, int64_t h5_file_id, int subfi
      *   ABC.h5.subfile_<file-number>_1_of_2, and
      *   ABC.h5.subfile_<file-number>.config
      */
-    sprintf(filepath, "%s/%s" SF_FILENAME_TEMPLATE, subfile_dir, base, h5_file_id, subfile_rank,
+    int numD = numDigits(n_io_concentrators);
+    sprintf(filepath, "%s/%s" SF_FILENAME_TEMPLATE, subfile_dir, base, h5_file_id, numD, subfile_rank,
             n_io_concentrators);
     if (prefix)
         HDfree(prefix);
```
derobins, line 2473: Probably best to be consistent with HD prefixes or not. Also, if this is intended to be a part of the library and not a plugin, it's better to use H5MM instead of HD(c|m|re)alloc/free. We caught a bunch of memory issues in the onion VFD and VFD SWMR with the H5MM sanity checks.

src/H5FDsubfile_mpi.c
brtnfld: fixed

src/H5FDsubfile_mpi.c
brtnfld: I'm not sure what you mean by consistent with prefixes, do you mean having a full path or not? I did not change the prefix with this PR.

brtnfld: Seems my comment disappeared.

brtnfld: What do you mean by HD prefixed consistentancy? Do you mean including the full path? This PR does not change the prefix.

PR #1258, opened 2021-12-02
Make default to build high-level tools the same as default for

lrknox:
high-level library.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -849,7 +849,7 @@ if test "X-$BUILD_MODE" = "X-clean" ; then
   HDF5_HL_TOOLS=no
 else
   HDF5_HL=yes
-  HDF5_HL_TOOLS=no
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -846,8 +846,10 @@ AC_SUBST([HDF5_HL_TOOLS])
 ## The high-level library is enabled unless the build mode is clean.
 if test "X-$BUILD_MODE" = "X-clean" ; then
   HDF5_HL=no
+  HDF5_HL_TOOLS=no
 else
   HDF5_HL=yes
+  HDF5_HL_TOOLS=no
```
byrnHDF: no or yes?

lrknox: yes

byrnHDF (approved):

derobins (approved):

PR #1271, opened 2021-12-09
Async Fixes

jhendersonHDF:
A post-release fix needed for async operations.

@lrknox I wasn't sure about procedure for this, so this PR is against the HDF5 1.13.0 branch for now. Please advise on how to handle this and any RELEASE.txt updating.

### DIFF
--- a/src/H5ESlist.c
+++ b/src/H5ESlist.c
@@ -88,6 +88,8 @@ H5ES__list_append(H5ES_event_list_t *el, H5ES_event_t *ev)
     HDassert(el);
     HDassert(ev);
 
+    ev->next = NULL;
+
     /* Append event onto the event list */
     if (NULL == el->tail)
         el->head = el->tail = ev;
--- a/src/H5VLcallback.c
+++ b/src/H5VLcallback.c
@@ -6452,18 +6452,28 @@ done:
 herr_t
 H5VL_request_wait(const H5VL_object_t *vol_obj, uint64_t timeout, H5VL_request_status_t *status)
 {
-    herr_t ret_value = SUCCEED; /* Return value */
+    hbool_t vol_wrapper_set = FALSE;
+    herr_t  ret_value       = SUCCEED; /* Return value */
 
     FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity checks */
     HDassert(vol_obj);
 
+    /* Set wrapper info in API context */
+    if (H5VL_set_vol_wrapper(vol_obj) < 0)
+        HGOTO_ERROR(H5E_VOL, H5E_CANTSET, FAIL, "can't set VOL wrapper info")
+    vol_wrapper_set = TRUE;
+
     /* Call the corresponding internal VOL routine */
     if (H5VL__request_wait(vol_obj->data, vol_obj->connector->cls, timeout, status) < 0)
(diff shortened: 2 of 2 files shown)


src/H5ESlist.c
```diff
@@ -88,6 +88,8 @@ H5ES__list_append(H5ES_event_list_t *el, H5ES_event_t *ev)
     HDassert(el);
     HDassert(ev);
 
+    ev->next = NULL;
```
jhendersonHDF, line 91: When we append an event to an event set, make sure to set the 'next' pointer to NULL, or it could be pointing to garbage, an event that's been freed, etc. This was problematic when moving events from the event set's active list to its failed list, where list iteration on failed events would previously try to follow into event nodes that were on the active list or were removed from the event set after completion, but a failed event node still pointed to them.

src/H5VLcallback.c
```diff
@@ -6452,18 +6452,28 @@ H5VL__request_wait(void *req, const H5VL_class_t *cls, uint64_t timeout, H5VL_re
...
+    herr_t  ret_value       = SUCCEED; /* Return value */
 
     FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity checks */
     HDassert(vol_obj);
 
+    /* Set wrapper info in API context */
+    if (H5VL_set_vol_wrapper(vol_obj) < 0)
+        HGOTO_ERROR(H5E_VOL, H5E_CANTSET, FAIL, "can't set VOL wrapper info")
```
jhendersonHDF, line 6465: Every 'request' callback needs to have a VOL wrapping context setup for it, since an async task may need to call something that uses the wrapping context. Notably, the DAOS VOL needs to call H5VLwrap_register.

lrknox:
> A post-release fix needed for async operations.
> 
> @lrknox I wasn't sure about procedure for this, so this PR is against the HDF5 1.13.0 branch for now. Please advise on how to handle this and any RELEASE.txt updating.

It should really go against the develop branch.  As 1.13.0 has been released, the hdf5_1_13_0 branch is destined for deletion, and a future hdf5_1_13_1 branch to prepare for a 1.13.1 release will be created from develop.  In the event we would decide to release 1.13.0-patch1 we should begin from the hdf5-1_13_0 tag, creating a short lived release preparation branch from there.

For corrections and additions to RELEASE.txt that need to be published before the next release we have in the past added notes to the download page, in addition to RELEASE.txt in develop and/or the support branch.

src/H5ESlist.c
soumagne: we should really stop manipulating lists like that anyway, i.e. and setting head and tails manually. I'll talk to you offline about it.

src/H5VLcallback.c
```diff
@@ -6958,18 +7008,28 @@ H5VL__request_free(void *req, const H5VL_class_t *cls)
...
 
     FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity check */
     HDassert(vol_obj);
 
+    /* Set wrapper info in API context */
+    if (H5VL_set_vol_wrapper(vol_obj) < 0)
+        HGOTO_ERROR(H5E_VOL, H5E_CANTSET, FAIL, "can't set VOL wrapper info")
+    vol_wrapper_set = TRUE;
```
soumagne, line 7022: instead of setting those intermediate state variables, we should just have HGOTO_ERROR macros return to a different error label.

soumagne (approved):

PR #1272, opened 2021-12-09
Fixes for async operations

jhendersonHDF:
A post-release fix needed for async operations.

Setup VOL wrapping context for each "request" callback

Set H5ES_event_t "next" pointer to NULL when appending to event list

### DIFF
--- a/src/H5ESlist.c
+++ b/src/H5ESlist.c
@@ -88,6 +88,8 @@ H5ES__list_append(H5ES_event_list_t *el, H5ES_event_t *ev)
     HDassert(el);
     HDassert(ev);
 
+    ev->next = NULL;
+
     /* Append event onto the event list */
     if (NULL == el->tail)
         el->head = el->tail = ev;
--- a/src/H5VLcallback.c
+++ b/src/H5VLcallback.c
@@ -6452,18 +6452,28 @@ done:
 herr_t
 H5VL_request_wait(const H5VL_object_t *vol_obj, uint64_t timeout, H5VL_request_status_t *status)
 {
-    herr_t ret_value = SUCCEED; /* Return value */
+    hbool_t vol_wrapper_set = FALSE;
+    herr_t  ret_value       = SUCCEED; /* Return value */
 
     FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity checks */
     HDassert(vol_obj);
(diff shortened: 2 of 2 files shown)


src/H5ESlist.c
```diff
@@ -88,6 +88,8 @@ H5ES__list_append(H5ES_event_list_t *el, H5ES_event_t *ev)
     HDassert(el);
     HDassert(ev);
 
+    ev->next = NULL;
```
jhendersonHDF, line 91: Moved comment from old PR:

> When we append an event to an event set, make sure to set the 'next' pointer to NULL, or it could be pointing to garbage, an event that's been freed, etc. This was problematic when moving events from the event set's active list to its failed list, where list iteration on failed events would previously try to follow into event nodes that were on the active list or were removed from the event set after completion, but a failed event node still pointed to them.

jhendersonHDF: >  soumagne
we should really stop manipulating lists like that anyway, i.e. and setting head and tails manually. I'll talk to you offline about it.

src/H5VLcallback.c
```diff
@@ -6452,18 +6452,28 @@ H5VL__request_wait(void *req, const H5VL_class_t *cls, uint64_t timeout, H5VL_re
...
+    herr_t  ret_value       = SUCCEED; /* Return value */
 
     FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity checks */
     HDassert(vol_obj);
 
+    /* Set wrapper info in API context */
+    if (H5VL_set_vol_wrapper(vol_obj) < 0)
+        HGOTO_ERROR(H5E_VOL, H5E_CANTSET, FAIL, "can't set VOL wrapper info")
```
jhendersonHDF, line 6465: Moved comments from old PR:

> Every 'request' callback needs to have a VOL wrapping context setup for it, since an async task may need to call something that uses the wrapping context. Notably, the DAOS VOL needs to call H5VLwrap_register.

jhendersonHDF: > soumagne
instead of setting those intermediate state variables, we should just have HGOTO_ERROR macros return to a different error label.

soumagne (approved):

lrknox (approved):

PR #1282, opened 2021-12-10

hyoklee:
```
t_file.c:109:43: warning: cast from 'const char *' to 'char *' drops const qualifier [-Wcast-qual]
            mrc = MPI_File_delete((char *)filename, info);
```

### DIFF
--- a/testpar/t_file.c
+++ b/testpar/t_file.c
@@ -106,9 +106,7 @@ test_split_comm_access(void)
 
         /* delete the test file */
         if (sub_mpi_rank == 0) {
-            char fname[NAME_MAX];
-            strncpy(fname, filename, strlen(filename));
-            mrc = MPI_File_delete(fname, info);
+            mrc = MPI_File_delete(filename, info);
             /*VRFY((mrc==MPI_SUCCESS), ""); */
         }
     }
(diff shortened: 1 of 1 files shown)


testpar/t_file.c
```diff
@@ -106,7 +106,9 @@ test_split_comm_access(void)
 
         /* delete the test file */
         if (sub_mpi_rank == 0) {
-            mrc = MPI_File_delete((char *)filename, info);
+            char fname[NAME_MAX];
+            strncpy(fname, filename, strlen(filename));
+            mrc = MPI_File_delete(fname, info);
```
gnuoyd: Looks like you can delete the cast to `char *` ?

jhendersonHDF: Agreed. MPI_File_delete takes a const char *, so no need to do a strncpy here, just drop the char * cast.

jhendersonHDF (changes requested):
> Looks like you can delete the cast to char * ?

Agreed. MPI_File_delete takes a const char *, so no need to do a strncpy here, just drop the char * cast.

jhendersonHDF (approved):

lrknox (approved):

PR #1287, opened 2021-12-13
Call FUNC_ENTER immediately in H5_init_library

jhendersonHDF:


src/H5.c
```diff
@@ -146,6 +146,8 @@ H5_init_library(void)
     size_t i;
     herr_t ret_value = SUCCEED;
 
+    FUNC_ENTER_NOAPI(FAIL)
```
jhendersonHDF, line 149: Since FUNC_ENTER... can do things like pushing function names where FUNC_LEAVE... expects symmetrical behavior, FUNC_ENTER... always needs to be called before any code that might cause a goto.

derobins (approved):

lrknox (approved):

PR #1288, opened 2021-12-13
Fix free list tracking and cleanup cast alignment warnings

jhendersonHDF:


src/H5FL.c
```diff
@@ -1001,27 +1001,32 @@ H5FL_blk_free(H5FL_blk_head_t *head, void *block)
 
 #ifdef H5FL_TRACK
     {
-        H5FL_track_t *trk = block = ((unsigned char *)block) - sizeof(H5FL_track_t);
```
jhendersonHDF, line 1004: This previous assignment to 'block' was corrupting memory because later code subtracted the size of the H5FL tracking info after it had already been subtracted here.

src/H5FL.c
```diff
@@ -1001,27 +1001,32 @@ H5FL_blk_free(H5FL_blk_head_t *head, void *block)
...
+        unsigned char *block_ptr = ((unsigned char *)block) - sizeof(H5FL_track_t);
+        H5FL_track_t   trk;
+
+        HDmemcpy(&trk, block_ptr, sizeof(H5FL_track_t));
```
jhendersonHDF, line 1007: Copy data in and out of a local H5FL_track_t struct instance so we don't have potential alignment issues when casting from unsigned char * to H5FL_track_t *

derobins (approved):

lrknox (approved):

PR #1299, opened 2021-12-15

hyoklee:
```
H5Lint.c:599:32: warning: cast from 'const char *' to 'char *' drops const qualifie\
r [-Wcast-qual]
    udata->lnk->name = (char *)name;
                               ^
H5Lint.c:1379:32: warning: cast from 'const char *' to 'char *' drops const qualifi\
er [-Wcast-qual]
    udata->lnk->name = (char *)name;
```

### DIFF
--- a/src/H5Lint.c
+++ b/src/H5Lint.c
@@ -596,8 +596,10 @@ H5L__link_cb(H5G_loc_t *grp_loc /*in*/, const char *name, const H5O_link_t H5_AT
 
     /* Set the link's name correctly */
     /* Casting away const OK -QAK */
-    udata->lnk->name = (char *)name;
-
+    udata->lnk->name = strdup(name);
+    if (NULL == udata->lnk->name) {
+        HGOTO_ERROR(H5E_LINK, H5E_CANTINIT, FAIL, "out of memory")
+    }
     /* Insert link into group */
     if (H5G_obj_insert(grp_loc->oloc, name, udata->lnk, TRUE,
                        udata->ocrt_info ? udata->ocrt_info->obj_type : H5O_TYPE_UNKNOWN,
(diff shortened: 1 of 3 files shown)


src/H5Lint.c
```diff
@@ -596,8 +596,10 @@ H5L__link_cb(H5G_loc_t *grp_loc /*in*/, const char *name, const H5O_link_t H5_AT
 
     /* Set the link's name correctly */
     /* Casting away const OK -QAK */
-    udata->lnk->name = (char *)name;
-
+    udata->lnk->name = strdup(name);
+    if (NULL == udata->lnk->name) {
+        HGOTO_ERROR(H5E_LINK, H5E_CANTINIT, FAIL, "out of memory")
+    }
```
jhendersonHDF, line 602: Since these changes are in callbacks, it sort of seems like doing a strdup each time probably wouldn't be optimal. This seems like the kind of thing that needs to be fixed somewhere higher up where we correctly differentiate between when the link name should be const and non-const.

If we do go with this change though, the strdup calls should be changed to HDstrdup, or even better, HDstrndup.

hyoklee: HDstrndup() is not available in H5private.h. 
strndup() is not available for MS VS C compiler. 
Thus, I replaced strdup() with HDstrdup().

qkoziol: > Since these changes are in callbacks, it sort of seems like doing a strdup each time probably wouldn't be optimal. This seems like the kind of thing that needs to be fixed somewhere higher up where we correctly differentiate between when the link name should be const and non-const.

Agree

derobins:
The strdup is probably too much overhead for simply squashing a warning.

PR #14, opened 2020-10-03
Correct entries related to --enable-build-mode and --enable-profiling in

lrknox:
INSTALL file, update COPYING file, also removing obsolete SZIP paragraph
from it.
Fix #9

PR #1302, opened 2021-12-16
Parallel Compression improvements

jhendersonHDF:


src/H5Dio.c
```diff
@@ -810,98 +810,35 @@ H5D__ioinfo_adjust(H5D_io_info_t *io_info, const H5D_t *dset, const H5S_t *file_
...
+                /* Retrieve size of MPI communicator used for file */
+                if ((comm_size = H5F_shared_mpi_get_size(io_info->f_sh)) < 0)
+                    HGOTO_ERROR(H5E_FILE, H5E_CANTGET, FAIL, "can't get MPI communicator size")
+
+                if (comm_size > 1) {
+                    char local_no_coll_cause_string[512];
+                    char global_no_coll_cause_string[512];
+
+                    if (H5D__mpio_get_no_coll_cause_strings(local_no_coll_cause_string, 512,
+                                                            global_no_coll_cause_string, 512) < 0)
```
jhendersonHDF, line 841: Moved most of this code into a new function to get strings for the reasons why collective I/O was broken for code reuse.

src/H5Dselect.c
```diff
@@ -256,6 +256,187 @@ H5D__select_io(const H5D_io_info_t *io_info, size_t elmt_size, size_t nelmts, co
...
+ *              with the only difference being that the readvv/writevv
+ *              calls are exchanged for H5VM_memcpyvv calls. Changes should
+ *              be made to both routines.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5D_select_io_mem(void *dst_buf, const H5S_t *dst_space, const void *src_buf, const H5S_t *src_space,
```
jhendersonHDF, line 277: A new routine that is very similar to H5D__select_io(), but rather than copying between application memory and the file, copies between two memory buffers according to the selection in the dst and src dataspaces.

src/H5FDmpio.c
```diff
@@ -188,6 +188,41 @@ H5FD__mpio_parse_debug_str(const char *s)
...
+ *
+ * Purpose:     Returns a string representing the enum value in an H5FD_mem_t
+ *              enum
+ *
+ * Returns:     H5FD_mem_t enum value string
+ *
+ *---------------------------------------------------------------------------
+ */
+static const char *
+H5FD__mem_t_to_str(H5FD_mem_t mem_type)
```
jhendersonHDF, line 203: The changes in this file allow one to see what type of I/O the MPI I/O file driver is doing. Previously one would only see the offset and length of the I/O. This now also shows you whether it's a superblock area, raw data, object header, etc.

src/H5mpi.c
```diff
@@ -549,4 +549,238 @@ H5_mpio_create_large_type(hsize_t num_elements, MPI_Aint stride_bytes, MPI_Datat
...
+ *                                      `recv_type`.
+ *
+ * Notes:       This routine is collective across `comm`.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5_mpio_gatherv_alloc(void *send_buf, int send_count, MPI_Datatype send_type, const int recv_counts[],
```
jhendersonHDF, line 625: The two new functions here are simply wrappers around MPI_(All)gatherv that hide a bit of boilerplate code. Both allocate the receive buffer for the caller. The only difference between the two is that the "simple" function calculates the recv_counts and displacements arrays for the caller before making the MPI_(All)gatherv call.

src/H5Dmpio.c
```diff
@@ -273,6 +373,185 @@ static int H5D__cmp_filtered_collective_io_info_entry_owner(const void *filtered
 /* Local Variables */
 /*******************/
 
+#ifdef H5Dmpio_DEBUG
```
jhendersonHDF, line 401: The below code adds debugging to H5Dmpio similar to that in the MPI I/O file driver.

src/H5Dmpio.c
```diff
@@ -499,145 +892,6 @@ H5D__mpio_select_write(const H5D_io_info_t *io_info, const H5D_type_info_t H5_AT
...
- *
- * Return:      Non-negative on success/Negative on failure
- *
- * Programmer:  Jordan Henderson
- *              Sunday, April 9th, 2017
- *
- *-------------------------------------------------------------------------
- */
-static herr_t
-H5D__mpio_array_gatherv(void *local_array, size_t local_array_num_entries, size_t array_entry_size,
```
jhendersonHDF, line 531: This whole routine was rewritten and move to H5mpi.c

src/H5Dmpio.c
```diff
@@ -808,6 +1068,36 @@ H5D__chunk_collective_io(H5D_io_info_t *io_info, const H5D_type_info_t *type_inf
     HDassert(type_info);
     HDassert(fm);
 
+    /* Obtain the current rank of the process and the number of ranks */
+    if ((mpi_rank = H5F_mpi_get_rank(io_info->dset->oloc.file)) < 0)
+        HGOTO_ERROR(H5E_IO, H5E_MPI, FAIL, "unable to obtain MPI rank")
+    if ((mpi_size = H5F_mpi_get_size(io_info->dset->oloc.file)) < 0)
+        HGOTO_ERROR(H5E_IO, H5E_MPI, FAIL, "unable to obtain MPI size")
```
jhendersonHDF, line 1125: Rather than retrieving the MPI rank and size multiple times in this file, do it once in H5D__chunk_collective_io which tends to be the main entrypoint in this file. Then, just hand those down to functions as needed.

src/H5Dmpio.c
```diff
@@ -1376,124 +1666,126 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_in
      */
     H5CX_set_mpio_actual_io_mode(H5D_MPIO_CHUNK_COLLECTIVE);
 
+    /* Enable collective metadata reads if not already enabled. The collective
+     * re-insertion of modified chunks into the chunk index does not scale well
+     * without reading metadata collectively and it should be safe to enable
+     * them here because it is required that these dataset writes be collective.
+     */
+    H5CX_set_coll_metadata_read(TRUE);
```
jhendersonHDF: The changes here currently turn on collective metadata reads for the feature since it will not scale well without them and we require the writes to be collective, but this may or may not be the best idea depending on the number of chunks in the operation.

src/H5Dmpio.c
```diff
@@ -1376,124 +1666,126 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_in
...
-                if (H5D__filtered_collective_chunk_entry_io(&chunk_list[i], io_info, type_info, fm) < 0)
-                    HGOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "couldn't process chunk entry")
+        H5D_MPIO_INIT_CHUNK_IDX_INFO(index_info, io_info);
 
-        /* Gather the new chunk sizes to all processes for a collective reallocation
-         * of the chunks in the file.
+        /* Redistribute any shared chunks being written to and
+         * send around chunk modification data as necessary
          */
-        if (H5D__mpio_array_gatherv(chunk_list, chunk_list_num_entries,
```
jhendersonHDF, line 1408: Rather than gathering everybody's list of chunks into a collective array, the feature has been revised in most places to construct different MPI derived types to only send as much data as needed, greatly reducing the feature's memory usage.

src/H5Dmpio.c
```diff
@@ -2954,464 +3400,1539 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
  *
  * Return:      Non-negative on success/Negative on failure
  *
- * Programmer:  Jordan Henderson
- *              Wednesday, January 18, 2017
- *
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5D__filtered_collective_chunk_entry_io(H5D_filtered_collective_io_info_t *chunk_entry,
```
jhendersonHDF, line 3234: This routine used to work on either reading an individual chunk (for dataset reads) or reading and writing an individual chunk (for dataset writes). However, any chunk reads here used to be independent which is a scalability problem for the feature. The new H5D__mpio_collective_filtered_chunk_read, H5D__mpio_collective_filtered_chunk_update and H5D__mpio_collective_filtered_chunk_common_io routines now perform the duties of this routine, but in a manner that allows chunk reads to be done collectively. This should generally scale much better and still allows the user the option of specifying independent chunk reads when desired.

src/H5Dmpio.c
```diff
@@ -2954,464 +3400,1539 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+#ifdef H5Dmpio_DEBUG
+    H5D_MPIO_TIME_STOP(mpi_rank);
+    H5D_MPIO_TRACE_EXIT(mpi_rank);
+#endif
+
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5D__mpio_collective_filtered_chunk_reinsert() */
+
+/*-------------------------------------------------------------------------
+ * Function:    H5D__mpio_get_chunk_redistribute_info_types
```
jhendersonHDF, line 5273: The 3 functions below here create different MPI derived datatypes to extract certain portions of information from the overall per-chunk H5D_filtered_collective_io_info_t structure. Usually, a particular operation (shared chunk redistribution, chunk reallocation, chunk reinsertion) only needs a few fields out of that structure and this information is gathered to all ranks, so sending just the few fields necessary can drastically save on memory usage at the expense of a bit of MPI overhead.

src/H5Dmpio.c
```diff
@@ -2954,464 +3400,1539 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+ *              The datatype contains the chunk offsets and lengths in the
+ *              file and the locations of the chunk data buffers to read
+ *              into/write from.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5D__mpio_collective_filtered_io_type(H5D_filtered_collective_io_info_t *chunk_list, size_t num_entries,
```
jhendersonHDF, line 5698: This routine was just revised a little bit to create slightly more efficient MPI derived types for performing I/O on filtered chunks.

src/H5Dmpio.c
```diff
@@ -1815,202 +2104,173 @@ H5D__multi_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_i
...
+            /* Free up resources in anticipation of following collective operation */
+            if (have_chunk_to_process && chunk_list[i].buf) {
+                H5MM_free(chunk_list[i].buf);
+                chunk_list[i].buf = NULL;
+            }
+
             /* Participate in the collective re-insertion of all chunks modified
              * in this iteration into the chunk index
              */
-            for (j = 0; j < collective_chunk_list_num_entries; j++) {
```
jhendersonHDF, line 1977: Chunk index reinsertion logic here moved into H5D__mpio_collective_filtered_chunk_reinsert, which more efficiently handles memory usage as well as chunk reinsertion itself.

src/H5Dmpio.c
```diff
@@ -1376,124 +1666,126 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_in
...
+                H5MM_free(chunk_list[i].buf);
+                chunk_list[i].buf = NULL;
+            }
+        }
+
         /* Participate in the collective re-insertion of all chunks modified
-         * in this iteration into the chunk index
+         * into the chunk index
          */
-        for (i = 0; i < collective_chunk_list_num_entries; i++) {
```
jhendersonHDF, line 1481: Chunk index reinsertion logic here moved into H5D__mpio_collective_filtered_chunk_reinsert, which more efficiently handles memory usage as well as chunk reinsertion itself.

src/H5Dmpio.c
```diff
@@ -1815,202 +2104,173 @@ H5D__multi_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_i
...
+                HGOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL,
+                            "couldn't collectively re-allocate file space for chunks")
+
+            /*
+             * If this rank has a chunk to work on, create a MPI type
+             * for writing out the chunk. Otherwise, the rank will
+             * use MPI_BYTE for the file and memory type and specify
+             * a count of 0.
              */
-            for (j = 0; j < collective_chunk_list_num_entries; j++) {
```
jhendersonHDF, line 1902: Chunk file space reallocation logic moved into H5D__mpio_collective_filtered_chunk_reallocate, which more efficiently handles memory usage.

src/H5Dmpio.c
```diff
@@ -1376,124 +1666,126 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_in
...
+         * send around chunk modification data as necessary
          */
-        if (H5D__mpio_array_gatherv(chunk_list, chunk_list_num_entries,
-                                    sizeof(H5D_filtered_collective_io_info_t),
-                                    (void **)&collective_chunk_list, &collective_chunk_list_num_entries, true,
-                                    0, io_info->comm, NULL) < 0)
-            HGOTO_ERROR(H5E_DATASET, H5E_CANTGATHER, FAIL, "couldn't gather new chunk sizes")
-
-        /* Collectively re-allocate the modified chunks (from each process) in the file */
-        for (i = 0; i < collective_chunk_list_num_entries; i++) {
```
jhendersonHDF, line 1415: Chunk file space reallocation logic moved into H5D__mpio_collective_filtered_chunk_reallocate, which more efficiently handles memory usage.

src/H5Dmpio.c
```diff
@@ -1815,202 +2104,173 @@ H5D__multi_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_i
...
          */
         for (i = 0; i < max_num_chunks; i++) {
-            /* Check if this process has a chunk to work on for this iteration */
-            hbool_t have_chunk_to_process =
-                (i < chunk_list_num_entries) && (mpi_rank == chunk_list[i].owners.new_owner);
+            /* Check if this rank has a chunk to work on for this iteration */
+            have_chunk_to_process = (i < chunk_list_num_entries) && (mpi_rank == chunk_list[i].new_owner);
 
-            if (have_chunk_to_process)
-                if (H5D__filtered_collective_chunk_entry_io(&chunk_list[i], io_info, type_info, fm) < 0)
```
jhendersonHDF, line 1887: Duties now performed by H5D__mpio_collective_filtered_chunk_update instead.

src/H5Dmpio.c
```diff
@@ -1376,124 +1666,126 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_in
...
-        udata.common.storage = index_info.storage;
-        udata.filter_mask    = 0;
-
-        /* Iterate through all the chunks in the collective write operation,
-         * updating each chunk with the data modifications from other processes,
-         * then re-filtering the chunk.
-         */
-        for (i = 0; i < chunk_list_num_entries; i++)
-            if (mpi_rank == chunk_list[i].owners.new_owner)
-                if (H5D__filtered_collective_chunk_entry_io(&chunk_list[i], io_info, type_info, fm) < 0)
```
jhendersonHDF, line 1402: Duties now performed by H5D__mpio_collective_filtered_chunk_update instead.

src/H5Dmpio.c
```diff
@@ -1815,202 +2104,173 @@ H5D__multi_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_i
...
     /* Initialize temporary contiguous storage info */
     ctg_store.contig.dset_size = (hsize_t)io_info->dset->shared->layout.u.chunk.size;
-    ctg_store.contig.dset_addr = 0;
-
-    /* Set dataset storage for I/O info */
-    io_info->store = &store;
 
     if (io_info->op_type == H5D_IO_OP_READ) { /* Filtered collective read */
-        for (i = 0; i < chunk_list_num_entries; i++)
-            if (H5D__filtered_collective_chunk_entry_io(&chunk_list[i], io_info, type_info, fm) < 0)
```
jhendersonHDF, line 1833: Duties now performed by H5D__mpio_collective_filtered_chunk_read instead.

lrknox:
These tests failed with module  OpenMPI/4.0.2-GCC-7.2.0-2.29 on jelly:
t_bigio
testphdf5 -- compressed dataset collective read (cmpdsetr)
Paralle Filters test - Testing write to unshared filtered chunks
t_2Gio -- compressed dataset collective read (cmpdsetr)

Error messages:
t_bigio - 

Single Rank Independent I/O
HDF5-DIAG: Error detected in HDF5 (1.13.0-7) MPI-process 0:
  #000: ../../src/H5D.c line 1222 in H5Dwrite(): can't synchronously write data
    major: Dataset
    minor: Write failed
  #001: ../../src/H5D.c line 1169 in H5D__write_api_common(): can't write data
    major: Dataset
    minor: Write failed
  #002: ../../src/H5VLcallback.c line 2181 in H5VL_dataset_write(): dataset write failed
    major: Virtual Object Layer
    minor: Write failed
  #003: ../../src/H5VLcallback.c line 2148 in H5VL__dataset_write(): dataset write failed
    major: Virtual Object Layer
    minor: Write failed
  #004: ../../src/H5VLnative_dataset.c line 345 in H5VL__native_dataset_write(): can't write data
    major: Dataset
    minor: Write failed
  #005: ../../src/H5Dio.c line 475 in H5D__write(): can't write data
    major: Dataset
    minor: Write failed
  #006: ../../src/H5Dcontig.c line 628 in H5D__contig_write(): contiguous write failed
    major: Dataset
    minor: Write failed
  #007: ../../src/H5Dselect.c line 492 in H5D__select_write(): write error
    major: Dataspace
    minor: Write failed
  #008: ../../src/H5Dselect.c line 224 in H5D__select_io(): write error
    major: Dataspace
    minor: Write failed
  #009: ../../src/H5Dcontig.c line 1249 in H5D__contig_writevv(): can't perform vectorized read
    major: Dataset
    minor: Can't operate on object
  #010: ../../src/H5VM.c line 1401 in H5VM_opvv(): can't perform operation
    major: Internal error (too specific to document in detail)
    minor: Can't operate on object
  #011: ../../src/H5Dcontig.c line 1179 in H5D__contig_writevv_cb(): block write failed
    major: Dataset
    minor: Write failed
  #012: ../../src/H5Fio.c line 191 in H5F_shared_block_write(): write through page buffer failed
    major: Low-level I/O
    minor: Write failed
  #013: ../../src/H5PB.c line 1018 in H5PB_write(): write through metadata accumulator failed
    major: Page Buffering
    minor: Write failed
  #014: ../../src/H5Faccum.c line 831 in H5F__accum_write(): file write failed
    major: Low-level I/O
    minor: Write failed
  #015: ../../src/H5FDint.c line 256 in H5FD_write(): driver write request failed
    major: Virtual File Layer
    minor: Write failed
  #016: ../../src/H5FDmpio.c line 1597 in H5FD__mpio_write(): file write failed
    major: Low-level I/O
    minor: Write failed

Test log for testphdf5
...
Testing  -- compressed dataset collective read (cmpdsetr)
[jelly:08172] *** Process received signal ***
[jelly:08172] Signal: Segmentation fault (11)
[jelly:08172] Signal code: Address not mapped (1)
[jelly:08172] Failing at address: 0x7ffd6aa548d0
[jelly:08172] [ 0] /lib64/libc.so.6(+0x36450)[0x7fc9f800a450]
[jelly:08172] [ 1] /lib64/libc.so.6(+0x154d0b)[0x7fc9f8128d0b]
...

Test log for t_filters_parallel
...
Testing write to unshared filtered chunks
[jelly:42035] *** Process received signal ***
[jelly:42035] Signal: Segmentation fault (11)
[jelly:42035] Signal code: Address not mapped (1)
[jelly:42035] Failing at address: 0x7ffe4e575000

Test log for t_2Gio
...
Testing  -- compressed dataset collective read (cmpdsetr)
[jelly:42526] *** Process received signal ***
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpiexec noticed that process rank 0 with PID 0 on node jelly exited on signal 11 (Segmentation fault).

ph5diff was also did hang, but I think that may be unrelated to this PR.

The complete output is in/mnt/wrk/lrknox/jhenderson-hdf5/atbuild/makecheckp.out for now.

lrknox (approved):

lrknox:
Tests passed with MPICH 3.3 and with OpenMPI 4.1.

examples/ph5_filtered_writes.c
```diff
@@ -0,0 +1,478 @@
...
+    /*
+     * Check if 'deflate' filter is available
+     */
+    filter_avail = H5Zfilter_avail(H5Z_FILTER_DEFLATE);
+    if (filter_avail) {
+        /*
+         * Set 'deflate' filter with reasonable
+         * compression level on DCPL
+         */
+        H5Pset_deflate(dcpl_id, 6);
```
derobins, line 68: If H5Zfilter_avail() fails, the FAIL (-1) value will be interpreted as truthy and this code will be inappropriately called. The other example also exhibits this behavior.

src/H5public.h
```diff
@@ -324,6 +324,23 @@ typedef uint64_t haddr_t;
...
+#elif HSIZE_UNDEF == UINT_MAX
+#define H5_MPI_HSIZE_T MPI_UNSIGNED
+#elif HSIZE_UNDEF == ULONG_MAX
+#define H5_MPI_HSIZE_T MPI_UNSIGNED_LONG
+#elif HSIZE_UNDEF == ULLONG_MAX
+#define H5_MPI_HSIZE_T MPI_UNSIGNED_LONG_LONG
+#else
+#error "no suitable MPI type found for hsize_t"
+#endif
+
```
derobins: I don't think we need this level of complexity. All 64-bit data models define long long as 64 bits, which is what we've hard-coded hsize_t to be.

I would:
- Define the type to MPI_UNSIGNED_LONG_LONG
- Either name the hsize_t type HSIZE_AS_MPI_TYPE or rename the haddr_t type H5_MPI_HADDR_T so the types have name parity
- Move the values closer to the definitions of hsize_t and haddr_t, respectively so they are less likely to get out of sync

examples/ph5_filtered_writes.c
jhendersonHDF: good catch, will fix

src/H5public.h
jhendersonHDF: I agree with most all of this. The only thing I would mention is that having the catch for not finding a suitable type to use for the definition is a bit more future proof than just defining the type as MPI_UNSIGNED_LONG_LONG. Though, whether or not haddr_t/hsize_t will ever be larger than 64 bits remains to be seen. The comparisons against sizes < 64 bits are probably unnecessary here however.

testpar/t_filters_parallel.c
```diff
@@ -5554,16 +5379,23 @@ test_read_cmpd_filtered_dataset_type_conversion_unshared(void)
     hsize_t              block[READ_COMPOUND_FILTERED_CHUNKS_TYPE_CONVERSION_UNSHARED_DATASET_DIMS];
     hsize_t              flat_dims[1];
     size_t               i, read_buf_size, correct_buf_size;
-    hid_t                file_id = -1, dset_id = -1, plist_id = -1, filetype = -1, memtype = -1;
-    hid_t                filespace = -1, memspace = -1;
+    hid_t                file_id = H5I_INVALID_HID, dset_id = H5I_INVALID_HID, plist_id = H5I_INVALID_HID;
+    hid_t                filetype = H5I_INVALID_HID, memtype = H5I_INVALID_HID;
+    hid_t                group_id  = H5I_INVALID_HID;
+    hid_t                filespace = H5I_INVALID_HID, memspace = H5I_INVALID_HID;
```
derobins: I usually put these one per line. I find them hard to parse when glommed together with commas.

testpar/t_filters_parallel.c
```diff
@@ -774,15 +808,17 @@ test_write_filtered_dataset_single_no_selection(void)
     data_size        = sel_dims[0] * sel_dims[1] * sizeof(*data);
     correct_buf_size = dataset_dims[0] * dataset_dims[1] * sizeof(*correct_buf);
 
-    data = (C_DATATYPE *)HDcalloc(1, data_size);
-    VRFY((NULL != data), "HDcalloc succeeded");
+    if (mpi_rank != WRITE_SINGLE_NO_SELECTION_FILTERED_CHUNKS_NO_SELECT_PROC) {
+        data = (C_DATATYPE *)HDcalloc(1, data_size);
```
derobins: Not a huge deal, but there's no need to cast the results of allocations.

src/H5mpi.c
```diff
@@ -549,4 +549,238 @@ H5_mpio_create_large_type(hsize_t num_elements, MPI_Aint stride_bytes, MPI_Datat
...
+    }
+
+    if (allgather || (mpi_rank == root)) {
+        *out_buf             = recv_buf;
+        *out_buf_num_entries = recv_buf_num_entries;
+    }
+
+done:
+    if (ret_value < 0) {
+        if (recv_buf)
```
derobins, line 698: Freeing NULL pointers is accepted by the standard so the check is superfluous.

src/H5Dselect.c
```diff
@@ -256,6 +256,187 @@ H5D__select_io(const H5D_io_info_t *io_info, size_t elmt_size, size_t nelmts, co
...
+    hsize_t *       src_off           = NULL;  /* Pointer to sequence offsets in source buffer */
+    size_t *        dst_len           = NULL;  /* Pointer to sequence lengths in destination buffer */
+    size_t *        src_len           = NULL;  /* Pointer to sequence lengths in source buffer */
+    size_t          curr_dst_seq;              /* Current destination buffer sequence to operate on */
+    size_t          curr_src_seq;              /* Current source buffer sequence to operate on */
+    size_t          dst_nseq;                  /* Number of sequences generated for destination buffer */
+    size_t          src_nseq;                  /* Number of sequences generated for source buffer */
+    size_t          dxpl_vec_size;             /* Vector length from API context's DXPL */
+    size_t          vec_size;                  /* Vector length */
+    ssize_t         bytes_copied;
```
derobins, line 294: Not critical for this PR, but we should consider refactoring H5VM_memcpyvv to return herr_t and return the number of bytes copied via an out parameter.

src/H5mpi.c
jhendersonHDF: Indeed. Switching to just freeing the pointer is something I've only started doing sort of recently but we should probably be doing that everywhere in the library.

src/H5Dselect.c
jhendersonHDF: Agreed. A memcpy implementation is certainly one where I could see the range of ssize_t being a problem one day. I'll make a note on changing this.

testpar/t_filters_parallel.c
jhendersonHDF: Very true. Just copy and pasted from old code that was written when I was less familiar with C conventions.

testpar/t_filters_parallel.c
jhendersonHDF: I mostly tried to just put related fields on the same line so that the function's declarations didn't get excessively long vertically, but I can always change this if it's hard to parse. With the introduction of H5S_BLOCK, I eliminated some old cruft variables, but didn't go over everything.

src/H5Dselect.c
```diff
@@ -256,6 +256,187 @@ H5D__select_io(const H5D_io_info_t *io_info, size_t elmt_size, size_t nelmts, co
...
+                if (H5S_SELECT_ITER_GET_SEQ_LIST(src_sel_iter, vec_size, nelmts, &src_nseq, &src_nelem,
+                                                 src_off, src_len) < 0)
+                    HGOTO_ERROR(H5E_DATASPACE, H5E_CANTGET, FAIL, "sequence length generation failed")
+
+                /* Start at the beginning of the sequences again */
+                curr_src_seq = 0;
+            } /* end if */
+
+            /* Perform vectorized memcpy from src_buf to dst_buf */
+            if ((bytes_copied = H5VM_memcpyvv(dst_buf, dst_nseq, &curr_dst_seq, dst_len, dst_off, src_buf,
```
fortnern, line 404: Are we sue this algorithm correctly handles the case where src and dst have different numbers of elements? This could happen get_seq_list is terminated due to vec_size being too small.

src/H5mpi.c
```diff
@@ -549,4 +549,238 @@ H5_mpio_create_large_type(hsize_t num_elements, MPI_Aint stride_bytes, MPI_Datat
...
+    /*
+     * Calculate the total size of the buffer being
+     * returned and allocate it
+     */
+    if (allgather || (mpi_rank == root)) {
+        size_t i;
+        size_t buf_size;
+
+#if MPI_VERSION >= 3
+        H5_CHECK_OVERFLOW(type_extent, MPI_Count, size_t);
```
fortnern: There should probably be a comment explaining why is this overflow check only done for MPI version > 3. Is it because lower versions will never have MPI_Aint larger than size_t?

Also: consider changing int to MPI_Aint for MPI version 2 routines where appropriate

src/H5Dmpio.c
```diff
@@ -436,6 +714,125 @@ H5D__mpio_opt_possible(const H5D_io_info_t *io_info, const H5S_t *file_space, co
...
+    size_t                         idx;
+    size_t                         num_cause_strings;
+    herr_t                         ret_value = SUCCEED;
+
+    FUNC_ENTER_PACKAGE
+
+    HDassert((local_cause && local_cause_len > 0) || (global_cause && global_cause_len > 0));
+
+    /* Ensure things are in order for our cause string table */
+    HDcompile_assert(H5D_MPIO_SET_INDEPENDENT == (H5D_mpio_no_collective_cause_t)1);
```
fortnern: Minor nitpick: I like to use hexadecimal for literal bitmap values like this

src/H5Dmpio.c
```diff
@@ -1377,123 +1671,118 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_in
...
 
-            /* Override the write buffer to point to the address of the first
-             * chunk data buffer
+        /* Setup contig storage info for I/O operation */
+        if (chunk_list_num_entries) {
+            /*
+             * Override the write buffer to point to the first
+             * chunk's data buffer
              */
             io_info->u.wbuf = chunk_list[0].buf;
```
fortnern, line 1802: Interesting - so this works even if the buffers aren't in the same order as the chunks? So some buffers have a negative offset (before factoring in this base address)?

src/H5Dmpio.c
```diff
@@ -2675,276 +2975,420 @@ H5D__construct_filtered_io_info_list(const H5D_io_info_t *io_info, const H5D_typ
...
-                shared_chunks_info_array[set_begin_index].owners.new_owner = new_chunk_owner;
-                shared_chunks_info_array[set_begin_index].num_writers      = num_writers;
-            } /* end for */
+                    /* Update the number of writers to this particular chunk */
+                    num_writers++;
 
-            num_assigned_chunks_array[new_chunk_owner]++;
-        } /* end for */
+                    chunk_entry++;
+                } while (++i < all_chunks_array_num_entries &&
```
fortnern: Should probably add a comment here or at the for loop about how the "i" variable is advanced both by the for loop and this loop, to improve readability

src/H5Dmpio.c
```diff
@@ -2954,464 +3398,1569 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+
+    if (MPI_SUCCESS != (mpi_code = MPI_Type_commit(struct_type)))
+        HMPI_GOTO_ERROR(FAIL, "MPI_Type_commit failed", mpi_code)
+
+    /* Created hindexed block type with an offset equal to the `fspace_info`
+     * structure's offset in the encompassing H5D_filtered_collective_io_info_t
+     * structure. Then, resize the type with an extent equal to the size of the
+     * encompassing structure to make it suitable for sending arrays of them.
+     */
+    {
```
soumagne: was there a reason for making a separate block?

jhendersonHDF: In this case, it was just to have a smaller scope for the 'extent' variable that we throw away after the call to MPI_Type_get_extent. To create the resized type, I only needed the lower bound ('lb') for the original type and not the extent, but I believe MPI_Type_get_extent requires you to provide variables for both 'lb' and 'extent'.

src/H5Dmpio.c
```diff
@@ -2954,464 +3398,1569 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
-                     * from the receive queue in the process and allocating that much memory
-                     * for the asynchronous receive
+                    /* Probe for a particular message for this chunk from any rank, removing
+                     * that message from the receive queue for the rank and allocating that
+                     * much memory for the non-blocking receive
                      */
-                    if (MPI_SUCCESS != (mpi_code = MPI_Mprobe(MPI_ANY_SOURCE, (int)chunk_entry->index,
-                                                              io_info->comm, &message, &status)))
+                    if (MPI_SUCCESS !=
+                        (mpi_code = MPI_Mprobe(MPI_ANY_SOURCE, (int)chunk_entry->index_info.chunk_idx,
```
fortnern: Seems strange to use a blocking call here then a non-blocking irecv, though I guess it simplifies the code. Do you think there would be performance to be gained by using MPI_Improbe (and adding another pass over the chunks)?

src/H5Dmpio.c
```diff
@@ -2954,464 +3398,1569 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+                                               io_info->comm, &message, &status)))
                         HMPI_GOTO_ERROR(FAIL, "MPI_Mprobe failed", mpi_code)
 
                     if (MPI_SUCCESS != (mpi_code = MPI_Get_count(&status, MPI_BYTE, &count)))
                         HMPI_GOTO_ERROR(FAIL, "MPI_Get_count failed", mpi_code)
 
                     HDassert(count >= 0);
```
fortnern: I would prefer a check and error here instead of an assertion, since assertions are supposed to only be used to guarantee a code's internal consistency, and this depends on MPI and the network. You could make an argument otherwise though

src/H5Dmpio.c
```diff
@@ -1816,201 +2102,165 @@ H5D__multi_chunk_filtered_collective_io(H5D_io_info_t *io_info, const H5D_type_i
...
+            redistribute_func = H5D__mpio_redistribute_shared_chunks_p0;
+
+            /* Perform redistribution operation */
+            result = (*redistribute_func)(chunk_list, chunk_list_num_entries, io_info, type_info, fm,
+                                          mpi_rank, mpi_size, NULL);
+            if (result < 0)
+                HGOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "unable to redistribute shared chunks")
+
+            /* Send chunk modification data */
+            if (H5D__mpio_share_chunk_modification_data(chunk_list, &chunk_list_num_entries, io_info,
```
fortnern, line 2235: Doesn't sharing all modification data at once use (potentially) lots of memory? If the primary purpose of multi chunk is to save memory, should modification data sharing be done in the loop, once per chunk? Or would the increased communication outweigh the memory savings?

src/H5Dmpio.c
```diff
@@ -2954,464 +3398,1569 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+                    local_entry = &chunk_list[num_processed].fspace_info;
+                    coll_entry  = &collective_list[i];
+
+                    if (H5D__chunk_file_alloc(idx_info, &coll_entry->chunk_current, &coll_entry->chunk_new,
+                                              &chunk_list[num_processed].index_info.need_insert, NULL) < 0)
+                        HGOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "unable to allocate chunk")
+                    *local_entry = *coll_entry;
+                } while ((++num_processed < chunk_list_num_entries) && (++i < collective_num_entries));
+
+                continue;
```
fortnern: Couldn't you just use an "else" instead of this continue?

src/H5Dmpio.c
```diff
@@ -2954,464 +3398,1569 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+                                                  &send_type_derived) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL,
+                        "can't create derived datatypes for chunk re-insertion info")
+
+        /*
+         * Gather information to all ranks for a collective re-insertion
+         * of the modified chunks into the chunk index
+         */
+        H5_CHECK_OVERFLOW(chunk_list_num_entries, size_t, int);
+        if (H5_mpio_gatherv_alloc_simple(chunk_list, (int)chunk_list_num_entries, send_type, recv_type, TRUE,
```
fortnern, line 5174: Would it be possible to avoid this allgather if we kept the needed data around for all chunks? I read the comment above saying it wasn't necessary to keep need_insert, but is there another reason we need to do the allgather?

src/H5Dmpio.c
```diff
@@ -2954,464 +3398,1569 @@ H5D__chunk_redistribute_shared_chunks(const H5D_io_info_t *io_info, const H5D_ty
...
+                    HDassert(file_offset_array[chunk_count] > file_offset_array[chunk_count - 1]);
+
+                /* Set the current chunk entry's size for the I/O operation */
+                if (op_type == H5D_IO_OP_READ) {
+                    H5_CHECK_OVERFLOW(chunk_list[i].fspace_info.chunk_current.length, hsize_t, int);
+                    length_array[chunk_count] = (int)chunk_list[i].fspace_info.chunk_current.length;
+                }
+                else {
+                    H5_CHECK_OVERFLOW(chunk_list[i].fspace_info.chunk_new.length, hsize_t, int);
+                    length_array[chunk_count] = (int)chunk_list[i].fspace_info.chunk_new.length;
```
fortnern: It might be worth setting up a convenience pointer to chunk_current/chunk_new to avoid these read/write branches and make the code a little more readable.

src/H5Dmpio.c
jhendersonHDF: Yes, the type created for the chunk memory buffers tends to have scattered buffers and the negative offsets for these are perfectly fine with MPI. We just have to make sure the chunks are sorted by file address.

src/H5Dmpio.c
jhendersonHDF: Here I think the changes are mostly just formatting and struct field renaming, so this blocking probe approach was the old first attempt at getting this correct. This has since been re-written into a non-blocking probing approach (not in this current PR), but I think in general it can be hard to guess about performance since we might be entering MPI multiple times in a non-blocking probing loop as opposed to blocking here. Hard to say, but I think initial profiling with the newly-rewritten non-blocking way will give some ideas.

src/H5Dmpio.c
jhendersonHDF: So, I agree with this point and did change it to an error check instead of an assert in my current set of changes. That said, I think the only danger here is if the MPI implementation is buggy/broken, in which case we probably have bigger problems. But yes, an assert probably wasn't best to use here, so I switched it (final changes to come in another PR rather than making this one larger).

src/H5Dmpio.c
jhendersonHDF: Yes, this is generally where a lot of memory usage will probably be encountered. However, for multi-chunk things get a little bit complicated because a rank can't finish updating a chunk until it's received all messages for that chunk. So we could end up in a situation where ranks are sending messages to other ranks for chunks that those ranks aren't processing yet. Things could be refactored a bit so that everyone keeps a collective list of all the chunks sorted in the same order and then all ranks process in order, either processing the chunk if the rank owns it, or sending a message out if not. That would probably increase memory usage a decent bit though (though probably not as much as the modification message overhead), and would probably have to use blocking communications since chunks are done one at a time. Not the end of the world or anything, but there's certainly room for overlapping non-blocking communications with some parallel compression work.

src/H5Dmpio.c
jhendersonHDF: Yes, the `continue` was more to just make this slightly odd section of code stand out, since it might be non-obvious what it's doing. If you'd prefer a simpler `else` for readability, I can change that.

src/H5Dmpio.c
jhendersonHDF: Since the struct fields used for the two operations (chunk file space allocation and chunk index re-insertion) differ from each other slightly, two allgathers are used mostly to save on memory. E.g., allgather only the info needed for re-allocation -> re-allocate -> free collective array, then allgather only the info needed for re-insertion -> re-insert -> free collective array. The `chunk_idx` and `chunk_new` fields are shared between the two operations, but the `chunk_current`, `filter_mask` and `need_insert` fields aren't.

It might seem like a bit of a micro-optimization, but would use around 24-32 bytes extra per chunk, multiplied by the number of ranks that have a copy of these collective arrays, so it can add up. I'm hoping that the memory saving is more useful than the overhead of the collective operations, but if two allgathers is prohibitive, the collective fields needed could be combined again.

src/H5Dmpio.c
jhendersonHDF: I think that's a good idea, especially for readability.

src/H5mpi.c
jhendersonHDF: Removed in latest changes as I now check for a negative extent first and casting should be safe for positive values.

src/H5Dmpio.c
jhendersonHDF: Updated this routine to be a bit cleaner about the enum/bit value handling.

src/H5Dmpio.c
jhendersonHDF: Added comments for this.

src/H5Dmpio.c
jhendersonHDF: Made much cleaner in latest changes.

src/H5Dmpio.c
jhendersonHDF: Fixed.

src/H5Dmpio.c
```diff
@@ -2997,6 +3003,7 @@ H5D__mpio_collective_filtered_chunk_io_setup(const H5D_io_info_t *io_info, const
                                              size_t *num_entries, int mpi_rank)
 {
     H5D_filtered_collective_io_info_t *local_info_array = NULL;
+    hbool_t                            filter_partial_edge_chunks;
```
fortnern, line 3117: Seems this is the inverse of how the option is usually presented. Not a big deal, but could cause some confusion.

src/H5Dint.c
```diff
@@ -384,37 +384,60 @@ H5D__get_space_status(const H5D_t *dset, H5D_space_status_t *allocation)
...
+
+            /* Compute the maximum size of the dataset in bytes */
+            full_size = nelmts * dt_size;
+
+            /* Check for overflow during multiplication */
+            if (nelmts != (full_size / dt_size))
+                HGOTO_ERROR(H5E_DATASET, H5E_OVERFLOW, FAIL, "size of dataset's storage overflowed")
+
+            /* Difficult to error check, since the error value is 0 and 0 is a valid value... :-/ */
+            if (H5D__get_storage_size(dset, &space_allocated) < 0)
```
fortnern: Is this better than H5D__get_num_chunks() for the unfiltered case or could we just always use H5D__get_num_chunks()?

soumagne (approved):
Approving assuming performance is validated. Should not be merged before.

src/H5Dmpio.c
```diff
@@ -3135,96 +3172,213 @@ H5D__mpio_collective_filtered_chunk_io_setup(const H5D_io_info_t *io_info, const
...
+    /*
+     * Allocate an array for each rank to keep track of the number of
+     * chunks assigned to any other rank in order to cut down on future
+     * MPI communication.
+     */
+    if (NULL == (num_chunks_map = H5MM_malloc((size_t)mpi_size * sizeof(*num_chunks_map))))
+        HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "couldn't allocate assigned chunks array")
+
+    /* Perform initial Allgather to determine the collective chunk list size */
+    if (MPI_SUCCESS != (mpi_code = MPI_Allgather(&chunk_list_num_entries, 1, H5_SIZE_T_AS_MPI_TYPE,
```
fortnern, line 3363: As discussed in the meeting, I'm not convinced there's a substantial reason to use anything other than p0. Is it really likely that choosing a different algorithm here could save more time than this allgather necessary to make the choice adds?

src/H5Dmpio.c
```diff
@@ -3321,18 +3489,25 @@ H5D__mpio_redistribute_shared_chunks_p0(H5D_filtered_collective_io_info_t *local
...
+    /* Scatter the segments of the collective chunk list back to each rank */
     if (MPI_SUCCESS !=
-        (mpi_code = MPI_Scatterv(all_chunks_array, send_counts_ptr, displacements_ptr, packed_type,
-                                 local_chunk_array, scatter_recvcount_int, struct_type, 0, io_info->comm)))
+        (mpi_code = MPI_Scatterv(coll_chunk_list, send_counts_ptr, displacements_ptr, packed_type,
+                                 chunk_list, num_chunks_int, struct_type, 0, io_info->comm)))
         HMPI_GOTO_ERROR(FAIL, "unable to scatter shared chunks info buffer", mpi_code)
 
+    /* Distribute rank number -> assigned chunks mapping back to all ranks */
+    if (MPI_SUCCESS != (mpi_code = MPI_Bcast(num_chunks_assigned_map, mpi_size, H5_SIZE_T_AS_MPI_TYPE,
```
fortnern: How did this work without the bcast before?

src/H5Dmpio.c
```diff
@@ -3683,105 +3820,204 @@ H5D__mpio_share_chunk_modification_data(H5D_filtered_collective_io_info_t *chunk
...
-
-            num_send_requests++;
         }
     }
 
-    /* Perform all the recvs on the chunks this rank DOES own */
-    for (i = 0, last_assigned_idx = 0; i < *chunk_list_num_entries; i++) {
-        H5D_filtered_collective_io_info_t *chunk_entry = &chunk_list[i];
+    /* Check if the number of send or receive requests will overflow an int (MPI requirement) */
+    H5_CHECK_OVERFLOW(num_send_requests, size_t, int)
```
fortnern, line 3995: Hmm is it possible to print a more informative error message here? Maybe unlikely that it'll be reasonable to post this many sends in the near future but it'd at least give app developers a hint that their I/O setup is bad

src/H5Dmpio.c
```diff
@@ -4228,135 +4462,154 @@ H5D__mpio_collective_filtered_chunk_update(H5D_filtered_collective_io_info_t *ch
...
-            chunk_list[i].chunk_buf_size =
-                MAX(chunk_list[i].chunk_current.length, file_chunk_size);
+        HDassert(mpi_rank == chunk_list[i].new_owner);
 
+        chunk_list[i].chunk_buf_size = MAX(chunk_list[i].chunk_current.length, file_chunk_size);
+
+        /*
+         * If this chunk hasn't been allocated yet and we aren't writing
+         * out fill values to it, make sure to 0-fill its memory buffer
+         * so we don't use uninitialized memory.
```
fortnern, line 4625: This behavior is controverial, see recent discussions with Suren. Though in the case of compression using malloc would likely do more harm than good.

src/H5Dmpio.c
```diff
@@ -3620,55 +3717,95 @@ H5D__mpio_share_chunk_modification_data(H5D_filtered_collective_io_info_t *chunk
...
+             * Since, at large scale, a chunk's index value may be larger than
+             * the maximum value that can be stored in an int, we cannot rely
+             * on using a chunk's index value as the tag for the MPI messages
+             * sent/received for a chunk. Therefore, add this chunk to a hash
+             * table with the chunk's index as a key so that we can quickly find
+             * the chunk when processing chunk messages that were received. The
+             * message itself will contain the chunk's index so we can update
+             * the correct chunk with the received data.
+             */
+            HASH_ADD(hh, chunk_table, index_info.chunk_idx, sizeof(hsize_t), &chunk_list[last_assigned_idx]);
```
fortnern, line 3896: I'm having trouble understanding this. It seems like you're using an hsize_t as the index into the hash table, which might be larger than an int.

src/H5Dmpio.c
fortnern: I noticed you changed the code in realloc to always save need_insert, but it seems these allgathers are still here?

src/H5Dselect.c
```diff
@@ -256,6 +256,187 @@ H5D__select_io(const H5D_io_info_t *io_info, size_t elmt_size, size_t nelmts, H5
...
+                                              src_nseq, &curr_src_seq, src_len, src_off)) < 0)
+                HGOTO_ERROR(H5E_IO, H5E_WRITEERROR, FAIL, "vectorized memcpy failed")
+
+            /* Decrement number of elements left to process */
+            HDassert(((size_t)bytes_copied % elmt_size) == 0);
+            if (elmt_size == 0)
```
fortnern: Shouldn't this just be an assert at the top of the function?

src/H5Dselect.c
fortnern: Nevermind, I think this algorithm is fine.

src/H5Dmpio.c
jhendersonHDF: Just seemed more natural to think of it this way for me. Perfectly happy to change/rename it if you think it'll be confusing.

src/H5Dmpio.c
jhendersonHDF: Note that the rank -> assigned chunks mapping is newly-added with the recent round of commits, so the MPI_Bcast goes along with that.

src/H5Dmpio.c
jhendersonHDF: I agree on the controversial part. However, I was previously using malloc and of course seeing lots of warnings about conditional jumps based on uninitialized values inside different filters' logic. While likely harmless there, 0-filling here hopefully shouldn't be as expensive as actually filling the memory buffer with a fill value and cleans up false positives. So, to be consistent with H5D__chunk_lock, I just chose to do the same thing as there and 0-fill the buffer (Calloc here, memset there). Users may expect to get junk, but it seems good to play it safe for the filters if the users don't care in the first place.

src/H5Dmpio.c
jhendersonHDF: Yes, uthash appears to use an unsigned for the hash value, so there are going to be hash collisions here. It's maybe not ideal, but since the chunk index can't be used for the MPI message tag and messages can come in any order, I needed a different way to quickly find a particular chunk structure by chunk index without iterating through the local array each time.

If we were to convert the library's chunk info structure to also include stuff needed for parallel compression, we could probably just do a skip list search to find a chunk. That may or may not be better here on average depending on the number of hash collisions and how well collisions are resolved, but there's a decent bit more work to do there. Open to suggestions.

src/H5Dselect.c
jhendersonHDF: Hmm, this was copied from H5D__select_io, but yes, I think in both cases the error check should be moved to the top of the functions since elmt_size shouldn't change.

jhendersonHDF:
> I noticed you changed the code in realloc to always save need_insert, but it seems these allgathers are still here?

The function is still fundamentally the same, I just cleaned it up a bit to not need the inner loop anymore by having one path for the alloc call and then deciding whether the rank needs to update a local chunk. The need_insert status still gets thrown away by ranks that aren't updating a local chunk. I didn't do any work on the optimization we discussed for saving need_insert, so both allgathers are still taking place.

src/H5Dint.c
jhendersonHDF: It looks like we could probably just use H5D__get_num_chunks here; I'll need to test that to make sure.

src/H5Dmpio.c
jhendersonHDF: Keep in mind that an initial (all)gather is necessary anyway because someone or everyone needs to know how big of a collective array to allocate for redistribution, so it's not really adding anything in terms of overhead. The allgather here also creates the rank -> num chunks assigned map which saves us two allgather operations later on in the reallocate and reinsert operations. Those used to each do 2 operations - first an allgather to see how many chunks everyone's contributing to the collective array (all of them for linked-chunk I/O, or possibly just 0 or 1 per rank for multi-chunk I/O) and then an allgatherv to actually gather the chunks. The rank -> num chunks assigned map lets us skip the initial allgather in each and just do the allgatherv, so we knock off 2 collective operations.

And remember that this is more about memory usage than performance, since memory usage multiplies by the number of ranks that each have to keep a copy of these distributed arrays. In any case, I feel like this is something that just needs to be tested at scale to see if this optimization is relevant or not. I think the next step is really to try using MPI-3's one-sided RDMA-like operations rather than dealing with all these collective arrays.

src/H5Dmpio.c
jhendersonHDF: I think that might be useful. Certainly if parallel compression is used at exascale (or maybe even smaller) we could probably start sending too many requests out and will have to rethink a little bit about how to break down the problem size. Let me convert this to something more helpful.

fortnern (approved):

brtnfld:
Performance testing in progress. Is there an issue I can attach results to?

brtnfld:
Using minIO--unstructured with gzip, on Cori, the memory issue seems to be much improved:

  _srun -n 1024 ./unstruct --pointspertask 16384 --tsteps 2 --hdf5 --hdf5_chunk 1024 961 9610  --hdf5_compress gzip,6_

HDF5-develop: Crashes 
HDF5-parallel-comp: Completes

jhendersonHDF:
> Using minIO--unstructured with gzip, on Cori, the memory issue seems to be much improved:
> 
> _srun -n 1024 ./unstruct --pointspertask 16384 --tsteps 2 --hdf5 --hdf5_chunk 1024 961 9610 --hdf5_compress gzip,6_
> 
> HDF5-develop: Crashes HDF5-parallel-comp: Completes

Good to hear! In just my testing on a local machine at small scale, writes appear to perform better and reads perform a little worse. I figured out why though and am deciding what to do there. Based on my profiling though, I imagine at scale the case I looked at won't have the same worse read performance characteristics.

jhendersonHDF:
> Performance testing in progress. Is there an issue I can attach results to?

Let's use TRILAB-283 for this.

brtnfld:
miniIO just does writing, I would need to use CGNS for both writing and reading.

jhendersonHDF:
Synced with develop for performance testing and added optimization to delay initialization of fill value buffer until first allocated chunk is found.

PR #1310, opened 2021-12-18

hyoklee:
```
mirror_writer.c:225:55: warning: passing 'const char [4]' to parameter of type 'cha\
r *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
    session->loginfo = mirror_log_init(opts->logpath, "W- ", MIRROR_LOG_DEFAULT_VER\
BOSITY);
                                                      ^~~~~
./mirror_remote.h:45:46: note: passing argument to parameter 'prefix' here
loginfo_t *mirror_log_init(char *path, char *prefix, unsigned int verbosity);
                                             ^
mirror_writer.c:766:41: warning: implicit conversion turns floating-point number in\
to integer: 'float' to 'unsigned long' [-Wfloat-conversion]
    buf = (char *)HDmalloc(sizeof(char) * H5FD_MIRROR_DATA_BUFFER_MAX);
                  ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/H5private.h:1053:28: note: expanded from macro 'HDmalloc'
#define HDmalloc(Z) malloc(Z)
                    ~~~~~~ ^
mirror_writer.c:791:57: warning: implicit conversion turns floating-point number in\
to integer: 'float' to 'size_t' (aka 'unsigned long') [-Wfloat-conversion]
        nbytes_in_packet = HDread(session->sockfd, buf, H5FD_MIRROR_DATA_BUFFER_MAX\
);

                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~\
~
../../src/H5FDmirror_priv.h:34:37: note: expanded from macro 'H5FD_MIRROR_DATA_BUFF\
ER_MAX'
#define H5FD_MIRROR_DATA_BUFFER_MAX H5_GB /* 1 Gigabyte */
                                    ^~~~~
../../src/H5private.h:443:34: note: expanded from macro 'H5_GB'
#define H5_GB (1024.0F * 1024.0F * 1024.0F)
               ~~~~~~~~~~~~~~~~~~^~~~~~~~~
../../src/H5private.h:1243:36: note: expanded from macro 'HDread'
#define HDread(F, M, Z) read(F, M, Z)
                        ~~~~       ^
mirror_writer.c:801:55: warning: implicit conversion changes signedness: 'ssize_t' \
(aka 'long') to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
            mirror_log_bytes(session->loginfo, V_ALL, nbytes_in_packet, (const unsi\
gned char *)buf);
            ~~~~~~~~~~~~~~~~                          ^~~~~~~~~~~~~~~~
mirror_server.c:341:56: warning: passing 'const char [4]' to parameter of type 'cha\
r *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
    run->loginfo = mirror_log_init(run->opts.log_path, "s- ", run->opts.verbosity);
                                                       ^~~~~
./mirror_remote.h:45:46: note: passing argument to parameter 'prefix' here
loginfo_t *mirror_log_init(char *path, char *prefix, unsigned int verbosity);
                                             ^
mirror_server.c:341:73: warning: implicit conversion changes signedness: 'int' to '\
unsigned int' [-Wsign-conversion]
    run->loginfo = mirror_log_init(run->opts.log_path, "s- ", run->opts.verbosity);
                   ~~~~~~~~~~~~~~~                            ~~~~~~~~~~^~~~~~~~~
mirror_server.c:459:20: warning: unused parameter 'sig' [-Wunused-parameter]
wait_for_child(int sig)
                   ^
mirror_server.c:531:47: warning: implicit conversion changes signedness: 'int' to '\
size_t' (aka 'unsigned long') [-Wsign-conversion]
        mirror_log_bytes(run->loginfo, V_ALL, ret, (const unsigned char *)mybuf);
        ~~~~~~~~~~~~~~~~                      ^~~
```

### DIFF
--- a/utils/mirror_vfd/mirror_remote.c
+++ b/utils/mirror_vfd/mirror_remote.c
@@ -78,7 +78,7 @@ mirror_log(struct mirror_log_info *info, unsigned int level, const char *format,
  * ----------------------------------------------------------------------------
  */
 void
-mirror_log_bytes(struct mirror_log_info *info, unsigned int level, size_t n_bytes, const unsigned char *buf)
+mirror_log_bytes(struct mirror_log_info *info, unsigned int level, ssize_t n_bytes, const unsigned char *buf)
 {
     FILE *       stream    = MIRROR_LOG_DEFAULT_STREAM;
     unsigned int verbosity = MIRROR_LOG_DEFAULT_VERBOSITY;
@@ -93,7 +93,7 @@ mirror_log_bytes(struct mirror_log_info *info, unsigned int level, size_t n_byte
     }
 
     if (level <= verbosity) {
-        size_t               bytes_written = 0;
+        ssize_t              bytes_written = 0;
         const unsigned char *b             = NULL;
 
         /* print whole lines */
@@ -147,7 +147,7 @@ mirror_log_bytes(struct mirror_log_info *info, unsigned int level, size_t n_byte
  * ----------------------------------------------------------------------------
  */
 loginfo_t *
-mirror_log_init(char *path, char *prefix, unsigned int verbosity)
+mirror_log_init(char *path, const char *prefix, unsigned int verbosity)
 {
     loginfo_t *info = NULL;
 
--- a/utils/mirror_vfd/mirror_writer.c
+++ b/utils/mirror_vfd/mirror_writer.c
@@ -763,7 +763,7 @@ do_write(struct mirror_session *session, const unsigned char *xmit_buf)
 
     /* Allocate the buffer once -- re-use between loops.
      */
-    buf = (char *)HDmalloc(sizeof(char) * (unsigned long)H5FD_MIRROR_DATA_BUFFER_MAX);
+    buf = (char *)HDmalloc(sizeof(char) * H5FD_MIRROR_DATA_BUFFER_MAX);
     if (NULL == buf) {
         mirror_log(session->loginfo, V_ERR, "can't allocate databuffer");
         reply_error(session, "can't allocate buffer for receiving data");
@@ -788,7 +788,7 @@ do_write(struct mirror_session *session, const unsigned char *xmit_buf)
      */
     sum_bytes_written = 0;
     do {
-        nbytes_in_packet = HDread(session->sockfd, buf, (unsigned long)H5FD_MIRROR_DATA_BUFFER_MAX);
+        nbytes_in_packet = HDread(session->sockfd, buf, H5FD_MIRROR_DATA_BUFFER_MAX);
         if (-1 == nbytes_in_packet) {
             mirror_log(session->loginfo, V_ERR, "can't read into databuffer");
             reply_error(session, "can't read data buffer");
--- a/utils/mirror_vfd/mirror_writer.c
+++ b/utils/mirror_vfd/mirror_writer.c
@@ -763,7 +763,7 @@ do_write(struct mirror_session *session, const unsigned char *xmit_buf)
 
     /* Allocate the buffer once -- re-use between loops.
      */
-    buf = (char *)HDmalloc(sizeof(char) * H5FD_MIRROR_DATA_BUFFER_MAX);
+    buf = (char *)HDmalloc(sizeof(char) * (unsigned long)H5FD_MIRROR_DATA_BUFFER_MAX);
     if (NULL == buf) {
         mirror_log(session->loginfo, V_ERR, "can't allocate databuffer");
         reply_error(session, "can't allocate buffer for receiving data");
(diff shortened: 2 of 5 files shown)


utils/mirror_vfd/mirror_writer.c
```diff
@@ -763,7 +763,7 @@ do_write(struct mirror_session *session, const unsigned char *xmit_buf)
 
     /* Allocate the buffer once -- re-use between loops.
      */
-    buf = (char *)HDmalloc(sizeof(char) * H5FD_MIRROR_DATA_BUFFER_MAX);
+    buf = (char *)HDmalloc(sizeof(char) * (unsigned long)H5FD_MIRROR_DATA_BUFFER_MAX);
```
lrknox: The casts in mirror_writer.c line 766 and line 871 are needed to eliminate warnings that a floating point number is implicitly converted to size_t (unsigned long).  While the cast may eliminate the warning, is it better to #define H5FD_MIRROR_DATA_BUFFER_MAX to H5_GB in H5FDmirror_priv.h (from H5private.h line 446: #define H5_GB (1024.0F * 1024.0F * 1024.0F)) than to #define H5FD_MIRROR_DATA_BUFFER_MAX (1024 * 1024 * 1024), casting it there if necessary?  Just curious.

hyoklee: @lrknox, I have difficulty in understanding your comment. 

1. Do you mean line 791 instead of line 871?
2. Do you mean `#define H5FD_MIRROR_DATA_BUFFER_MAX (1024 * 1024 * 1024)`  is better than the current `#define 5FD_MIRROR_DATA_BUFFER_MAX H5GB` in H5FDmirror_priv.h to remove casting?
3. What do you mean "there" in "casting it there if necessary"?

lrknox: 1. yes, 791 is the line I meant.

2. Not necessarily better, but no casts would be required where it is used.  And it is unlikely that 1 GB would be changed to some value other than 1024 * 1024 * 1024.

3. H5FD_MIRROR_DATA_BUFFER_MAX is used in 2 places, lines 766 and 791, both of which require casts to avoid the warning.  By "there" I meant in the #define in src/H5FDmirror_priv.h line 34.  If there is a need to use H5_GB to define H5FD_MIRROR_DATA_BUFFER_MAX, can src/H5FDmirror_priv.h line 34 be #define H5FD_MIRROR_DATA_BUFFER_MAX (unsigned long)H5_GB /* 1 Gigabyte */?  Or if it isn't necessary to define H5FD_MIRROR_DATA_BUFFER_MAX in terms of H5_GB, then perhaps #define H5FD_MIRROR_DATA_BUFFER_MAX (1024 * 1024 * 1024).

Searching the code for other uses of H5_GB, there are 3:  1 in src/H5Dmpio.c in an expression that is cast to hsize_t for the size of a dataset, and 2 in src/H5timer.c that are cast to double in calculating bandwidth.  It appears from the comment in H5private.h that H5_GB was defined for this purpose and later used in the other files because the value was correct, if not the type.

My comment was intended more as a question than as a suggestion.

gnuoyd: `#define H5FD_MIRROR_DATA_BUFFER_MAX (1024 * 1024 * 1024)` looks like the right thing to do.

lrknox (approved):

utils/mirror_vfd/mirror_remote.c
```diff
@@ -78,7 +78,7 @@ mirror_log(struct mirror_log_info *info, unsigned int level, const char *format,
  * ----------------------------------------------------------------------------
  */
 void
-mirror_log_bytes(struct mirror_log_info *info, unsigned int level, size_t n_bytes, const unsigned char *buf)
+mirror_log_bytes(struct mirror_log_info *info, unsigned int level, ssize_t n_bytes, const unsigned char *buf)
```
derobins, line 81: ssize_t is not in the C standard (it's POSIX). We should avoid using it in new code. If you want to indicate errors, this function should return herr_t, not set an output parameter to a special 'bad' value.

derobins (changes requested):
ssize_t should not be used in new code

utils/mirror_vfd/mirror_remote.c
hyoklee: reference please?

lrknox: HDread(), # defined to (presumably linux) read() in H5private.h:1243, returns ssize_t and is used in mirror_server.c:524, mirror_wwrite.c:791 and mirror)writer.c:892.  HDread() returns ssize_t and outside of mirror_vfd is used in code for several file drivers and a number of test files.  The warnings in the mirror_vfd generally occurred because the third parameter in calls to mirror_log_bytes, size_t nbytes, were sometimes ssize_t values from calls to HDread.  These followed a check that the ssize_t value was not -1 (other uses of HDread are usually checked directly to not be < 0), after which I suppose they can be presumed non-negative and cast to size_t?  I was able to eliminate the clang warnings in mirror_vfd code using some casts to size_t without adding any changes to ssize_t.

hyoklee: @lrknox , feel free to close this PR and apply your patch that worked for you.

lrknox:
Warnings fixed by PR #1444

PR #1322, opened 2021-12-21
Group tests issues

vchoi-hdfgroup:


kyang2014 (approved):
I've tested the pull request at Mac and CentOS 7. The fix works fine. By changing the number of groups to other numbers or by changing the page buffer size or page space size to odd numbers like(page buffer size = 4703, space size = 2301), #2 and #4 issues in the design-fail document may appear. I communicated with Vailin. These messages seem expected.

kyang2014 (commented):
I've tested Vailin's new change. The only error encountered is the meaningful error
"Reader's API time exceeds max_lag ticks, suggest to increase the value of max_lag.".

After I increase the max_lag(sometimes the number should be very big), all the tests get passed.

PR #1330, opened 2021-12-28
Changed to correct return datatype for H5Lexists

brtnfld:
Changed to correct return datatype, htri_t, for H5Lexists


byrnHDF (approved):
It would be nice to use a "clean" branch instead of pulling in all commits from develop.

lrknox (approved):

PR #1331, opened 2022-01-01
Fix H5Tinit.c dependency

hjmjohnson:
Simplify configuration time and build time dependency chains.

Using the latest CMake and ninja build tools identified potential missing dependancies
at build time.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1054,9 +1054,9 @@ if (HDF5_BATCH_H5DETECT)
   )
   add_custom_command (
       OUTPUT     gen_SRCS.stamp1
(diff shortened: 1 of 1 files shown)


hjmjohnson:
FYI:  I found similar simplification by the excellent VTK team:

https://gitlab.kitware.com/third-party/hdf5/-/commit/757aa42ebff8c9819c054f5e00559143c60f653d

byrnHDF (changes requested):
Please use develop for a new pull request unless the changes only apply to a development branch. Do not make pull-requests against a release version. If a patch is necessary for a release then a patch release should be requested.

byrnHDF:
See "cmake custom command fix" #1356 for PR against develop.

PR #1334, opened 2022-01-04

hyoklee:
```
pio_perf.c:1308:54: warning: cast from 'char **' to 'const char **' must have all i\
ntermediate pointers const qualified to be safe [-Wcast-qual]
    while ((opt = H5_get_option(argc, (const char **)argv, s_opts, l_opts)) != EOF)\
 {
                                                     ^
1 warning generated.
```

### DIFF
--- a/tools/src/h5perf/pio_perf.c
+++ b/tools/src/h5perf/pio_perf.c
@@ -188,7 +188,7 @@ typedef struct _minmax {
 
 /* local functions */
 static off_t           parse_size_directive(const char *size);
-static struct options *parse_command_line(int argc, char *argv[]);
+static struct options *parse_command_line(int argc, const char *argv[]);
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

tools/src/h5perf/pio_perf.c
```diff
@@ -214,21 +214,32 @@ static off_t squareo(off_t);
  * Modifications:
  */
 int
-main(int argc, char *argv[])
+main(int argc, const char *argv[])
```
jhendersonHDF, line 217: main should accept a char * array, not a const char * array. To fix this warning, the only change that should be needed is to cast argv to (const char * const *) when passing it to H5_get_option.

derobins: Agree. See my comments in the related PR.

PR #1341, opened 2022-01-07

hyoklee:
```
ph5diff_main.c  :65:31: warning: cast from 'const char *' to 'char *' drops const q\
ualifier [-Wcast-qual]
    MPI_Init(&argc, (char ***)&argv);
                              ^
```

### DIFF
--- a/tools/src/h5diff/ph5diff_main.c
+++ b/tools/src/h5diff/ph5diff_main.c
@@ -46,7 +46,9 @@ static void ph5diff_worker(int);
 int
 main(int argc, const char *argv[])
 {
+    int         i        = 0;
     int         nID      = 0;
+    char **     args     = NULL;
     const char *fname1   = NULL;
     const char *fname2   = NULL;
     const char *objname1 = NULL;
@@ -62,7 +64,16 @@ main(int argc, const char *argv[])
     outBuffOffset = 0;
     g_Parallel    = 1;
 
-    MPI_Init(&argc, (char ***)&argv);
+    if ((args = (char **)HDcalloc((size_t)argc, sizeof(char *))) == NULL) {
+        HDfprintf(stderr, "%s: HDcalloc call failed\n", PROGRAMNAME);
+        return EXIT_FAILURE;
+    }
+    else {
+        for (i = 0; i < argc; i++) {
+            args[i] = HDstrdup(argv[i]);
+        }
+    }
+    MPI_Init(&argc, &args);
 
     MPI_Comm_rank(MPI_COMM_WORLD, &nID);
     MPI_Comm_size(MPI_COMM_WORLD, &g_nTasks);
@@ -102,7 +113,13 @@ main(int argc, const char *argv[])
     } /* end else */
 
     MPI_Finalize();
-
+    if (args) {
+        for (i = 0; i < argc; i++) {
+            if (args[i])
+                HDfree(args[i]);
+        }
+        HDfree(args);
+    }
     return 0;
 }
 


tools/src/h5diff/ph5diff_main.c
```diff
@@ -62,7 +64,16 @@ main(int argc, const char *argv[])
     outBuffOffset = 0;
     g_Parallel    = 1;
 
-    MPI_Init(&argc, (char ***)&argv);
+    if ((args = (char **)HDcalloc((size_t)argc, sizeof(char *))) == NULL) {
```
jhendersonHDF, line 67: There's no need to copy the args to fix this warning. The signature for main should simply take a char * array rather than a const char * array. Just dropping the const in the signature should be enough.

hyoklee: Then, ```parse_command_line(int argc, const char *argv[], const char **fname1, const char **fname2,``` will trigger warnings.

jhendersonHDF: In that case, things have to be fixed further down the stack as well. main should be changed here to simply take a char * array, which will fix the warning from MPI_Init. parse_command_line will probably need some changing as well.

hyoklee: I disagree. MPI_Init should use const. Until then, we can add another layer like HDMPI_Init() since many HDF5 CLI tools use const in main() function, which is right thing to do.

jhendersonHDF: It is a mistake for the tools to use const for argv. This is not the prototype for main and it's very clearly the wrong thing to do because we act like argv is const and then immediately cast that away when passing it to MPI_Init. Note that MPI_Init might modify argc/argv so it should NOT accept const parameters. _All_ of HDF5's tools should be updated to not use a const qualifier for argv and it was likely done just to paper over the warnings caused by trying to pass argv directly to H5_get_option which takes a const argv.

hyoklee: From https://www.open-mpi.org/doc/current/man3/MPI_Init.3.php
```
All MPI programs must contain a call to MPI_Init or MPI_Init_thread. Open MPI accepts the C/C++ argc and argv arguments to main, but neither modifies, interprets, nor distributes them:
```
I don't know who (@qkoziol  or @derobins  ?) started adding the const in argv[] for main() and H5_get_option() in src/H5systemc. but I think it's the right thing to do for many reasons.

jhendersonHDF: See also: https://stackoverflow.com/questions/9185792/passing-arguments-via-command-line-with-mpi.

MPI will not consume what the caller has passed to the program itself but may consume any arguments to mpirun so that it effectively looks like those arguments were never passed in the first place. I think it was in good spirit to add const to H5_get_option, but it is definitely NOT the right thing to do for main. Regardless of whether the MPI implementation modifies argc/argv or not, MPI_Init's prototype does not use const for its arguments and simply casting the const away from the const qualified argv isn't good. If the tools really want to use a const qualified argv, then they should either copy the arguments into a new array or just cast argv appropriately. But NONE of the tools should use a const argv, since this doesn't match main's prototype as given by the C standard.

hyoklee: The C standard is for legacy when **const** was not available in compiler. Why do you think the modern compiler Objective-C / clang added **const**?

byrnHDF: What does C99 standard indicate -  we have moved to that minimum for develop.

jhendersonHDF: > What does C99 standard indicate - we have moved to that minimum for develop.

The C standard only mandates that the implementation should require either void for main or the `int argc, char * argv[]` combo. In practice, pretty much all implementations will allow const char * argv[], but that doesn't mean we should be using that. Clearly this is a case where using a const argv is a mistake. In theory, just changing H5_get_option's signature to accept a `const char * const *` should fix these warnings, but may add more inside H5_get_option depending on what it's doing with argv.

byrnHDF: Agreed.

gnuoyd:
On Fri, Jan 07, 2022 at 09:11:03AM -0800, H. Joe Lee wrote:
> @hyoklee commented on this pull request.
> 
> 
> 
> > @@ -62,7 +64,16 @@ main(int argc, const char *argv[])
>      outBuffOffset = 0;
>      g_Parallel    = 1;
>  
> -    MPI_Init(&argc, (char ***)&argv);
> +    if ((args = (char **)HDcalloc((size_t)argc, sizeof(char *))) == NULL) {
> 
> The C standard is for legacy when **const** was not available in compiler. Why do you think the modern compiler Objective-C / clang added **const**?

The standard added const a long time ago.  The compilers follow the
standard.

Dave

hyoklee:
> The standard added const a long time ago. The compilers follow the standard. Dave

@gnuoyd , yes, I'm old. I did not see **const** in 1991 when I wrote C program. The **const** still appears modern to me. :smile: Anyway, what's your final verdict? Use or don't use const in main(argv). If not, who will replace the main() codes under tools/src?

gnuoyd:
On Fri, Jan 07, 2022 at 10:48:44AM -0800, H. Joe Lee wrote:
> > The standard added const a long time ago. The compilers follow the standard. Dave
> 
> @gnuoyd , yes, I'm old. I did not see **const** in 1991 when I wrote C program. The **const** still appears modern to me. :smile: Anyway, what's your final verdict? Use or don't use const in main(argv). If not, who will replace the main() codes under tools/src?

I think that if MPI_Init() is called by main() or by one of main()'s
callees, then the definition should be `int main(int argc, char **argv)`
instead of `int main(int argc, const char **argv)`.

I cannot find anything in the C standard that explicitly rules *in* `int
main(int argc, const char **argv)`, but I cannot find anything that
rules it *out*. :-) Compilers are very lenient about the `const`.  So I
don't think it's urgent to change unless it helps with your project.

I feel that if you're not going to write through a pointer, then that
pointer should be const unless there is some extenuating circumstance.
So it would be good if main() passed `argv` to other functions as `const
char **` or `const char * const *` instead of `char **`.

(Extenuating circumstances: declarations get really long or really
strange, or a lot of awkward casts become necessary.)

As for the other main() routines in tools/src/, it depends on the
context.  Do they also call MPI_init()?

Dave

hyoklee:
@gnuoyd , Yes.

```
nene-[10:18pm]~/src/hdf5.HDFGroup/tools/src>grep -r MPI_Init */*.c
h5diff/ph5diff_main.c:    MPI_Init(&argc, (char ***)&argv);
h5perf/perf.c:    MPI_Init(&argc, &argv);
h5perf/perf.c:    MPI_Initialized(&mpi_initialized);
h5perf/pio_perf.c:    MPI_Init(&argc, &argv);
```

Here's the `argv` mess.

```
h5copy/h5copy.c:main(int argc, const char *argv[])
h5diff/h5diff_main.c:main(int argc, const char *argv[])
h5diff/ph5diff_main.c:main(int argc, const char *argv[])
h5dump/h5dump.c:main(int argc, const char *argv[])
h5format_convert/h5format_convert.c:main(int argc, const char *argv[])
h5import/h5import.c:main(int argc, char *argv[])
h5jam/h5jam.c:main(int argc, const char *argv[])
h5jam/h5unjam.c:main(int argc, const char *argv[])
h5ls/h5ls.c:main(int argc, const char *argv[])
h5perf/perf.c:main(int argc, char **argv)
h5perf/perf.c:main(int H5_ATTR_UNUSED argc, char H5_ATTR_UNUSED **argv)
h5perf/pio_perf.c:main(int argc, char *argv[])
h5perf/sio_perf.c:main(int argc, const char *argv[])
h5repack/h5repack_main.c:main(int argc, const char **argv)
h5stat/h5stat.c:main(int argc, const char *argv[])
misc/h5clear.c:main(int argc, const char *argv[])
misc/h5debug.c:main(int argc, char *argv[])
misc/h5delete.c:main(int argc, const char *argv[])
misc/h5mkgrp.c:main(int argc, const char *argv[])
misc/h5repart.c:main(int argc, char *argv[])
```

jhendersonHDF:
Agreed with Dave on

> So I
don't think it's urgent to change unless it helps with your project.

The tools _shouldn't_ be using `const char *argv[]` but it isn't really urgent that we change that since essentially all implementations will compile that fine. Anything that makes use of MPI_Init though should use non-const argv.

tools/src/h5diff/ph5diff_main.c
derobins: The C11 standard requires either int main(void) or int main(int argc, char* argv[]). Equivalents are allowed, like char **argv, but the const should be removed.

derobins: Also: "The parameters argc and argv and the strings pointed to by the argv array shall be modifiable by the program, and retain their last-stored values between program startup and program termination."

lrknox:
General opinion is that  main(int argc, const char *argv[]) should be  main(int argc, char *argv[]).
On the other hand, Apple xcode does it this way.
Does Apple have MPI application?  What do they do?

byrnHDF:
As the defacto tools maintainer and reviewing all the comments, I will remove the const from the signature and see what type of warnings appear. I will report my results and make a recommendation to move forward.

hyoklee:
> As the defacto tools maintainer and reviewing all the comments, I will remove the const from the signature and see what type of warnings appear. I will report my results and make a recommendation to move forward.

Please first add this coding standard about main() function under: https://github.com/HDFGroup/hdf5/blob/develop/doc/ 

I still believe that Apple engineers are right.

byrnHDF:
Changing the main signatures to remove the "const" from the signature produced no warnings in those files. "const" warnings decreased from 327 to 263.
I will create a PR for my changes.

derobins:
This no longer emits warnings with clang 14.0.0 w/ OpenMPI 4.1.2

PR #1345, opened 2022-01-08
fixed off-by-one error in h5fget_name_f, HDFFV-11290

brtnfld:
If an exact size buffer is passed to h5fget_name_f, then the returned filename will be truncated by one due to an error in the C wrapper. This fixes the issue and adds a test.


epourmal (approved):
Please fix the comment I missed. Thank you!

brtnfld:
Fixed typo.

lrknox (approved):

PR #1355, opened 2022-01-11
Implement H5ESget requests function to retrieve requests from an event set

fortnern:


epourmal (approved):
The new function needs Doxygen section. We will need to discuss how to present developers' functions in a separate section of RM.

src/H5ES.c
```diff
@@ -235,6 +235,61 @@ H5ESget_op_counter(hid_t es_id, uint64_t *op_counter /*out*/)
...
+
+    /* Call internal routine */
+    if (count && *count > 0 && (requests || connector_ids))
+        if (H5ES__get_requests(es, order, connector_ids, requests, *count) < 0)
+            HGOTO_ERROR(H5E_EVENTSET, H5E_CANTGET, FAIL, "can't get requests")
+
+    /* Retrieve the count, if non-NULL */
+    if (count)
+        *count = H5ES__list_count(&es->active);
+
```
soumagne, line 288: Not sure if you should allow for count to be NULL if the routine does nothing if count is NULL :)

soumagne (approved):

src/H5ES.c
fortnern: Well free() will happily do nothing if you pass NULL ;)  Though I don't have a strong feeling one way or the other.

In general though, I should mention that I thought this coding pattern (passing a pointer to the length of a buffer that is updated during the routine) was common throughout the library, but now I can't find any examples of it in the public API. Perhaps it is only used for internal interfaces. I'm open to changing the interface, though having the function return both the data and the length of the data in a single call may prevent some race conditions in a multithreaded application (with threadsafe on). Certainly we should always pass the length of the user's buffer.

fortnern:
> 

Is the doxygen section the comments stuff I see in the public headers above the prototypes? There isn't one for the other function in H5ESdevelop.h, should I add that? Some of the other develop headers are also missing them. Are developer functions supposed to have doxygen entries?

src/H5ES.c
fortnern: API revised according to discussion on Friday. There are now separate input and output parameters array_len and count. The function now does something even if array_len is 0 (it returns count). I don't know if it's worth throwing an error if array_len is 0 and count is NULL.

fortnern:
Github is apparently broken. The code checks passed, and the format check should pass as well since the auto check-in, but github won't run it. Let me know if it looks good and Larry can merge it manually.

src/H5ES.c
```diff
@@ -235,6 +235,61 @@ H5ESget_op_counter(hid_t es_id, uint64_t *op_counter /*out*/)
...
+ * Return:      SUCCEED / FAIL
+ *
+ * Programmer:  Neil Fortner
+ *              Tuesday, November 23, 2021
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5ESget_requests(hid_t es_id, H5_iter_order_t order, hid_t *connector_ids, void **requests, size_t array_len,
+                 size_t *count /*out*/)
```
soumagne, line 266: Looks fine, I think otherwise it's usually better to just name it `max_count`, similar to what `epoll_wait()` does: https://man7.org/linux/man-pages/man2/epoll_wait.2.html
That just makes it clearer that *count <= array_len

soumagne: OK forget my last comment, reading more I see it's kind of decoupled anyway and *count can be greater than array_len.

PR #1359, opened 2022-01-12
Used clang-tidy to change all floating point f suffixes to F

seanm:


byrnHDF (approved):

gnuoyd:
On Wed, Jan 12, 2022 at 07:15:42AM -0800, Sean McBride wrote:
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/HDFGroup/hdf5/pull/1359
> 
> -- Commit Summary --
> 
>   * Used clang-tidy to change all floating point f suffixes to F

Generally lowercase letters should be preferred to uppercase for
readability and ease of entry.

1 lowercase letter shapes are more distinguishable one from another
  and from digits
2 lowercase letters form more distinguished word shapes
3 it's not necessary to use Shift/Caps Lock to enter lowercase letters

@byrnHDF, what is the argument for using uppercase instead of lowercase?

Dave

seanm:
> @byrnHDF, what is the argument for using uppercase instead of lowercase?

There are some here: https://clang.llvm.org/extra/clang-tidy/checks/readability-uppercase-literal-suffix.html

It's convincing for `l`. From there, I guess the argument is consistency.

byrnHDF:
Because that was what authors were using in other commits - so we chose uppercase.

gnuoyd:
On Wed, Jan 12, 2022 at 09:13:00AM -0800, Allen Byrne wrote:
> Because that was what authors were using in other commits - so we chose uppercase.

Sure, I understand.  I am looking for the reason *at bottom* to
replicate this practice clear across the code base.

Do the authors have some reason?

If this is only the habit of authors, then is it a good habit or a bad
habit?

Dave

byrnHDF:
Not sure, just saw changes that were changing case to Uppercase - maybe the clang warnings PRs?

bmribler (approved):

PR #1383, opened 2022-01-19
Replaced several uses of sprintf with safer snprintf

seanm:


seanm:
There are a lot more `sprintf` that could be changed, but I want to start small to see if this will be accepted...

byrnHDF (approved):

soumagne:
It's a good change but we should also check the return value of `snprintf()`

seanm:
>It's a good change but we should also check the return value of snprintf()

That's much harder as then you have to figure out what to do upon failure.  But if you want to, go for it. :)

epourmal (commented):
Can we make our mind and use snprintf everywhere and eliminate HDsnprintf?

seanm:
> Can we make our mind and use snprintf everywhere and eliminate HDsnprintf?

That's an easy find-replace that could happen in a subsequent PR.

gnuoyd:
On Fri, Jan 28, 2022 at 07:01:20AM -0800, Sean McBride wrote:
> > Can we make our mind and use snprintf everywhere and eliminate HDsnprintf?
> 
> That's an easy find-replace that could happen in a subsequent PR.

Why don't we proceed as follows?

1) Accept Sean's PR as-is.  Clearly enforcing (correct) bounds on the
   formatted strings is better than what came before.

   Checking and handling snprintf errors is a project in and of itself.
   It will bulk up the code.  I doubt the bulk will pay for itself.

2) In a subsequent PR, remove the abominable HD from snprintf.

   (Why not remove it from all of the other C99 formatted print
   functions, too?)

   It should *not* be necessary for *every single standard C99 function*
   to be "blessed" with an HD.  Rather, the library should either poison
   or implement any C99 function that is, say, unavailable on Windows.

   When I say poison, I mean add to some HDF5 header file,

```
// poison snprintf
#ifdef snprintf
#undef snprintf
#endif
#define snprintf h5_library_snprintf_poison
```

Dave

seanm:
I agree with @gnuoyd on all points.

There is also [#pragma GCC poison](https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html) BTW.

I've since made even more sprint -> snprintf changes.  Shall I integrate them into this PR or make an additional one?  As @byrnHDF has already reviewed this one, I'm inclined to make another, but it's up to y'all.

epourmal:
Agree! Go ahead and approve so Larry can merge it.

epourmal:
> I agree with @gnuoyd on all points.
> 
> There is also [#pragma GCC poison](https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html) BTW.
> 
> I've since made even more sprint -> snprintf changes. Shall I integrate them into this PR or make an additional one? As @byrnHDF has already reviewed this one, I'm inclined to make another, but it's up to y'all.

Sure, just create a new one and it can be approved quickly. Just refer to this one.

byrnHDF:
Agree, same template of replace can be approved quickly.

gnuoyd (approved):
LGTM.

lrknox (approved):

PR #1384, opened 2022-01-20
Added another missing override keyword on a dtor

seanm:


seanm:
I thought I'd got all these months ago, but I either missed one or this class was added/changed, or...

byrnHDF (approved):

lrknox (approved):

PR #1387, opened 2022-01-24
refactor H5T_cmp and reintroduce Tony Horrobin's cache of the member-name sort order

gnuoyd:
I have extracted a couple of subroutines from H5T_cmp and simplified a lot, mainly by using temporary variables to replace repeated expressions. The code is more readable this way.

There are a lot of changes in this PR. It consists of 33 individual commits with descriptions. I recommend reading the individual commits.

I'm marking this WIP because I have some doubts about (re-)sorting compound-type members "on demand" in H5T_cmp.

It seems that applications probably can afford the time and space cost of the library keeping members sorted by both name and index, always. I don't think applications will change any type's members often enough to make an important performance difference. I don't think applications use types in numbers great enough that space will be a problem. Not so?

If the members are always stored in `offset` order, and there is always a `name`-ordered array of member indices alongside, then H5T_cmp does not have to allocate any memory. Right now, when an allocation fails, H5T_cmp returns 0, indicating arbitrarily that its H5T_t arguments are "equal". That sets up the library or application to fail mysteriously somewhere down the road.

Enum types, too, should keep their cases continually sorted, to avoid on-demand heap allocation.

### DIFF
--- a/src/H5Tcompound.c
+++ b/src/H5Tcompound.c
@@ -467,9 +467,6 @@ H5T__insert(H5T_t *parent, const char *name, size_t offset, const H5T_t *member)
         if (memb == NULL)
             HGOTO_ERROR(H5E_DATATYPE, H5E_CANTALLOC, FAIL, "memory allocation failed")
 
-        if (pcmpd->idx_name != NULL)
-            pcmpd->idx_name = H5MM_xfree(pcmpd->idx_name);
-
         pcmpd->nalloc = na;
         pcmpd->memb   = memb;
     } /* end if */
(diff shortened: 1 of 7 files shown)


src/H5Tcompound.c
```diff
@@ -426,77 +426,81 @@ H5Tpack(hid_t type_id)
...
 {
-    unsigned idx; /* Index of member to insert */
-    size_t   total_size;
-    unsigned i;                   /* Local index variable */
-    herr_t   ret_value = SUCCEED; /* Return value */
+    unsigned            idx; /* Index of member to insert */
+    size_t              total_size;
+    unsigned            i;                   /* Local index variable */
+    herr_t              ret_value = SUCCEED; /* Return value */
+    H5T_shared_t *const msh = member->shared, *const psh = parent->shared;
```
byrnHDF, line 433: I like the readability of the use of temp vars, however personally I'd prefer declarations on individual lines.

hyoklee:
https://github.com/HDFGroup/hdf5/pull/1176

derobins:
Closing due to age and lack of movement. Issue #4528 has been created as a reminder to go back over this.

Issue #1388, opened 2022-01-25
Renamed Attribute in dense storage can't be deleted


kmuehlbauer:
If I add 

```C
    /* Delete last attribute */
    ret = H5Adelete(dataset, attrname);
    CHECK(ret, FAIL, "H5Adelete");
```

before Dataset close in this test file, the Attribute can't be deleted. This could be another manifestation of https://forum.hdfgroup.org/t/bug-h5arename-fails-unexpectedly/4881

https://github.com/HDFGroup/hdf5/blob/4d45d05078fb1f67cc449fd5a6826c6dcfe55bfe/test/tattr.c#L2807-L2810

More context:

- creation order tracking enabled
- attributes which are created in dense storage can be renamed since this fix https://github.com/HDFGroup/hdf5/commit/4e05249f98d9f7247032cfb6250da654e813f14a
- renamed attributes can't be deleted anymore
- attributes which are not renamed, can be deleted

Conclusion:

Something is going wrong in `H5A__dense_rename`. The above fix was introduced by @vchoi-hdfgroup, so maybe @vchoi-hdfgroup can give some insight here?

 https://github.com/HDFGroup/hdf5/blob/4d45d05078fb1f67cc449fd5a6826c6dcfe55bfe/src/H5Adense.c#L858-L1027

kmuehlbauer:
I've added a possible fix to #1389. The testsuite runs smoothly, and downstream works nicely too.

epourmal:
Thank you for the proposed fix. We will be reviewing it shortly.

kmuehlbauer:
@epourmal Thanks for the heads-up.

kmuehlbauer:
#1389 doesn't solve the problem and I'm not able to fix this. I'd very much appreciate if HDF5 group could prioritize here to find some working solution for this annoying issue. At least a time line would be good.

# HDFFV-10579
- https://forum.hdfgroup.org/t/bug-h5arename-fails-unexpectedly/4881/5

Downstream we constantly get bug reports which can be tracked to this very bug:

# h5netcdf
- https://github.com/h5netcdf/h5netcdf/issues/207
- https://github.com/h5netcdf/h5netcdf/issues/202
- https://github.com/h5netcdf/h5netcdf/issues/179
- https://github.com/h5netcdf/h5netcdf/issues/136
- https://github.com/h5netcdf/h5netcdf/issues/128
- 
# h5py
- https://github.com/h5py/h5py/issues/1641
[...]

PR #1389, opened 2022-01-25
remove old attribute before inserting renamed attribute

kmuehlbauer:
Proof of #1388 

Update: added possible fix.

This changes the order of processing. The old attribute is removed from dense storage before the renamed attribute is inserted. With the former behavior, deleting a renamed attribute wasn't possible any more.

Questions:

Why is removing/inserting from/in dense storage needed anyway? Wouldn't just updating the indexes with the new name be sufficient? I'm sure I'm missing background information on the hdf5 internals, so any pointers are welcome.

### DIFF
--- a/src/H5Adense.c
+++ b/src/H5Adense.c
@@ -942,35 +942,10 @@ H5A__dense_rename(H5F_t *f, const H5O_ainfo_t *ainfo, const char *old_name, cons
     if (H5A__set_version(f, attr_copy) < 0)
         HGOTO_ERROR(H5E_ATTR, H5E_CANTSET, FAIL, "unable to update attribute version")
 
-    /* Need to remove the attribute from the creation order index v2 B-tree */
-    if (ainfo->index_corder) {
-        hbool_t corder_attr_exists; /* Attribute exists in v2 B-tree */
-
-        /* Open the creation order index v2 B-tree */
-        HDassert(H5F_addr_defined(ainfo->corder_bt2_addr));
-        if (NULL == (bt2_corder = H5B2_open(f, ainfo->corder_bt2_addr, NULL)))
-            HGOTO_ERROR(H5E_ATTR, H5E_CANTOPENOBJ, FAIL, "unable to open v2 B-tree for creation index")
-
-        /* Set up the creation order to search for */
-        udata.corder = attr_copy->shared->crt_idx;
-
-        corder_attr_exists = FALSE;
-        if (H5B2_find(bt2_corder, &udata, &corder_attr_exists, NULL, NULL) < 0)
-            HGOTO_ERROR(H5E_ATTR, H5E_NOTFOUND, FAIL, "can't search for attribute in name index")
-
-        if (corder_attr_exists) {
-            H5A_bt2_ud_rm_t rm_udata;
-
-            /* Set up the creation order in user data for the v2 B-tree 'record remove' callback */
-            rm_udata.common.corder = attr_copy->shared->crt_idx;
-
-            /* Remove the record from the creation order index v2 B-tree */
-            if (H5B2_remove(bt2_corder, &rm_udata, NULL, NULL) < 0)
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@byrnHDF Would you mind approving the workflow? I just want to see if and how things fail here. At least we would have some traceback to better understand what might be the cause. Thanks.

kmuehlbauer:
I've added a fix which solves the renaming/delete issue. At least it works locally on my machine. Please consider approving the workflow.

takluyver:
(I think HDF group are in Illinois, so it will probably be a few more hours before they're online to approve the Github workflow ;-)

kmuehlbauer:
> 



> (I think HDF group are in Illinois, so it will probably be a few more hours before they're online to approve the Github workflow ;-)

Thanks @takluyver, I'll wait patiently now. :grin:

kmuehlbauer:
Thanks for letting the workflow run. I've fixed the linter errors (hopefully). Should now be ready another try and for review. Please let me know, if and what I can do to finalize this. Thanks!

epourmal:
>

> Thanks for letting the workflow run. I've fixed the linter errors (hopefully). Should now be ready another try and for review. Please let me know, if and what I can do to finalize this. Thanks!

We need to check what the library behavior should be when creation order of attributes is tracked and then an attribute is deleted. There may be a deeper issue here and the change like this may affect netCDF-4 apps. Stay tuned.

kmuehlbauer:
@epourmal No worries, I'm interested to solve this in the best possible way.

Actually I'm maintaining `h5netcdf`, a Python library which writes netcdf4 files via `h5py`. So anything which works for `netcdf-c` should work for `h5py`/`h5netcdf` too.

And even if my proposed fix is not the correct tool, we might find the correct one in the process.

vchoi-hdfgroup:
>

vchoi-hdfgroup:
The suggested fix to swap the order of processing in H5A__dense_rename() to:
--delete old attribute first from dense storage via H5A__dense_remove()
--then insert renamed attribute into dense storage via H5A__dense_insert()
 
I think we need to keep the existing order of processing (insert first then delete), see comments for H5O__attr_link() which is called between the above two routines:
               -- /* Increment reference count on attribute components */
               -- /* (so that they aren't deleted when the attribute is removed shortly) */
 
The actual problem for not able to delete a renamed attribute is because:
--The attribute’s creation order got removed from the creation order index b-tree in H5A__dense_remove()

Need to work up a better fix.  
My input is:
Since the ‘rename’ operation just changes the name index b-tree and the creation order index btree for that attribute remains intact, my input for the suggested fix is:
--Don’t insert/remove the attribute’s creation order from the creation order index b-tree for the ‘rename’ operation in H5A__dense_insert()/H5A__dense_remove()


>

vchoi-hdfgroup:
Sorry I closed the PR by accident.

For the question: "Why is removing/inserting from/in dense storage needed anyway? Wouldn't just updating the indexes with the new name be sufficient? "
‘Rename’ is changing the name of the object, therefore changing its position in the name index b-tree.

kmuehlbauer:
@vchoi-hdfgroup Thanks for the details and suggestions.

So would it be possible to not use `H5A__dense_insert()/H5A__dense_remove()` at all and just copy the necessary lines of code into `H5A__dense_rename()` at the appropriate places?

My C skills are quite rusty. I can try to come up with a solution, but if anyone from inside hdf5 group immediately sees a solution, I'd very much appreciate overriding my PR.

This quite aged bug should be resolved soonish.

epourmal:
@kmuehlbauer,

We will not have anyone right now to work on the issue. Please, go ahead and try implement what was suggested and submit new PR; mark it with "work in progress" (WIP) and we will be continue working with you. Don't worry about your C skills :-)

kmuehlbauer:
I'm sorry, but I'll not have time and energy to work on this PR. I've updated my issue #1388 with more details and would appreciate if this could be solved by HDF5. Thanks for the support!

PR #1395, opened 2022-01-26
Revert "Open bsd fixes"

lrknox:
Reverts HDFGroup/hdf5#1195

Switching DETECT_BYTE to DETECT_I for 8 byte types doesn't work on our big-endian Power8 machine because switching the last parameter of DETECT_I_BYTE_CORE from "int" to the 8 byte type incorrectly sets dt->shared->u.atomic.order = H5T_ORDER_LE on a big-endian machine.  

@seanm, no problems otherwise.

### DIFF
--- a/config/cmake_ext_mod/ConfigureChecks.cmake
+++ b/config/cmake_ext_mod/ConfigureChecks.cmake
@@ -223,63 +223,65 @@ set (LINUX_LFS 0)
 set (HDF_EXTRA_C_FLAGS)
 set (HDF_EXTRA_FLAGS)
 if (MINGW OR NOT WINDOWS)
-  if (CMAKE_SYSTEM_NAME MATCHES "Linux")
-	# Linux Specific flags
-	# This was originally defined as _POSIX_SOURCE which was updated to
-	# _POSIX_C_SOURCE=199506L to expose a greater amount of POSIX
-	# functionality so clock_gettime and CLOCK_MONOTONIC are defined
-	# correctly. This was later updated to 200112L so that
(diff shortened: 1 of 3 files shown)


seanm:
Why revert all of #1195 though?

Moreover, wouldn't it be better to fix this properly? :)

lrknox:
> Why revert all of #1195 though?
> 
> Moreover, wouldn't it be better to fix this properly? :)

Reverting to stop daily test failures.  However, fixing the problem would be fine.  Why was the H5detect change needed for BSD?

seanm:
> > Why revert all of #1195 though?
> > Moreover, wouldn't it be better to fix this properly? :)
> 
> Reverting to stop daily test failures. However, fixing the problem would be fine. Why was the H5detect change needed for BSD?

It's explained here: https://github.com/HDFGroup/hdf5/pull/1195/commits/61464d7ebe98a7ed84aeeecab1fa811a5ea0b831

This is invalid:

```
DETECT_BYTE(int_least8_t, INT_LEAST8, d_g[nd_g]);
```

`int_least8_t`, like its name suggests, is *at least* 8 bits, you can't assume it's *exactly* 8 bits which `DETECT_BYTE` does.

seanm:
>Reverting to stop daily test failures

Right, but can't you revert only the 1 commit instead of all 4?

lrknox:
Ok, github doesn't give me a button for that, but with a little more work I can create a PR to revert just the H5detect.c change.

lrknox:
PR #1396 created to revert only the H5detect.c changes

PR #1396, opened 2022-01-26
Revert 61464d7e

lrknox:
Reverts the H5detect.c change in PR #1195.


src/H5detect.c
```diff
@@ -205,12 +205,15 @@ precision(detected_t *d)
 }
 
 /*-------------------------------------------------------------------------
- * Function:    DETECT_I
+ * Function:    DETECT_I/DETECT_BYTE
  *
- * Purpose:     This macro takes a type like `int' and a base name like
+ * Purpose:     These macro takes a type like `int' and a base name like
```
byrnHDF, line 210: This instead of These

byrnHDF (approved):

derobins (approved):

src/H5detect.c
lrknox: Yes, I missed it when I reverted, then left it for an eventual fix that accomodates both big-endian and OpenBSD.

PR #1400, opened 2022-01-28
Replace H5detect's build-time detection of C99 integer properties with a

gnuoyd:
table-driven routine, `H5T__init_native_int()`, that is run at library
initialization time.

Always respect the alignment used by the compiler for integers.  The
library invites trouble by using different alignment than the compiler
expects.  Here and there update a comment about alignment.

Retire the H5detect code that tries to find the least permissible
integer alignment by running experiments and catching any signals thrown
or unexpected results.


gnuoyd:
@seanm I believe this overhaul of H5detect / native-type initialization should help with some of the issues you had?

gnuoyd:
Does anybody have an opinion about the precision of int_least8_t and int_fast8_t? Seems like it should be 8 bits rather than `8 * sizeof(int_least8_t)` or `8 * sizeof(int_fast8_t)`, respectively. It's an easy change to make in the `H5T__init_native_int()` table.

seanm:
Great, thanks for doing this!  Should probably fix the compiler error on OpenBSD that was re-introduced with [this revert](https://github.com/HDFGroup/hdf5/pull/1396).

Will you be doing some similar treatment for floats too?

seanm:
Also, looks like you are close to being able to remove `HDF_NO_UBSAN`, which would be great!

lrknox:
The errors below showed up in the "Windows Latest MSVC" test in https://github.com/HDFGroup/hdf5/actions/runs/1762792143.  The checks above won't complete because of the clang-format commit.  The test log is at https://github.com/HDFGroup/hdf5/runs/4984099680?check_suite_focus=true with these errors beginning at line 385 after a number of warnings beginning at line 342.

D:\a\hdf5\hdf5\src\H5Tnative.c(1134,1): error C2099: initializer is not a constant [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1134,1): warning C4047: 'initializing': 'size_t' differs in levels of indirection from 'hid_t *' [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1134,1): warning C4047: 'initializing': 'hid_t *' differs in levels of indirection from 'size_t' [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1134,7): error C2078: too many initializers [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1148,1): error C2099: initializer is not a constant [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1148,1): warning C4047: 'initializing': 'size_t' differs in levels of indirection from 'hid_t *' [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1148,1): warning C4047: 'initializing': 'hid_t *' differs in levels of indirection from 'size_t' [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1148,7): error C2078: too many initializers [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1162,1): error C2099: initializer is not a constant [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1162,1): warning C4047: 'initializing': 'size_t' differs in levels of indirection from 'hid_t *' [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1162,1): warning C4047: 'initializing': 'hid_t *' differs in levels of indirection from 'size_t' [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
D:\a\hdf5\hdf5\src\H5Tnative.c(1162,7): error C2078: too many initializers [D:\a\hdf5\build\src\hdf5-shared.vcxproj]
  H5Toffset.c

fortnern:
> Does anybody have an opinion about the precision of int_least8_t and int_fast8_t? Seems like it should be 8 bits rather than `8 * sizeof(int_least8_t)` or `8 * sizeof(int_fast8_t)`, respectively. It's an easy change to make in the `H5T__init_native_int()` table.

I thought those weren't required to be exactly 8 bits, only int8_t is guaranteed to be 8 bits. As far as I can tell this applies to precision as well as width. Then again, strictly speaking, looking at the size isn't guaranteed to give you the correct precision (though I doubt HDF5 would work correctly on a system with weird integer types).

fortnern:
Have we looked into the potential performance/memory effects of this change? I worked on a similar system a long time ago and we wanted to allow unaligned access wherever the system handled it correctly to maximize performance (for example, otherwise a compound may need to go through a more expensive type conversion process).

On a related note, have you looked at the H5_NO_ALIGNEMNT_RESTRICTIONS that is checked in configure and used in H5Tvlen.c?

I agree that conforming to the C standard is good as long as it doesn't substantially hurt performance.

jhendersonHDF:
> > Does anybody have an opinion about the precision of int_least8_t and int_fast8_t? Seems like it should be 8 bits rather than `8 * sizeof(int_least8_t)` or `8 * sizeof(int_fast8_t)`, respectively. It's an easy change to make in the `H5T__init_native_int()` table.
> 
> I thought those weren't required to be exactly 8 bits, only int8_t is guaranteed to be 8 bits. As far as I can tell this applies to precision as well as width. Then again, strictly speaking, looking at the size isn't guaranteed to give you the correct precision (though I doubt HDF5 would work correctly on a system with weird integer types).

This is correct. The 'least' types are guaranteed to be at least that size, but could be larger. I think we should generally keep the 'least' types as small as possible since they're more related to minimizing memory usage, while the 'fast' types should of course be the fastest size for the platform.

qkoziol:
Has this been checked for a cross-compile environment?  Probably one of the systems at an NNSA facility could be used.

lrknox:
> Has this been checked for a cross-compile environment? Probably one of the systems at an NNSA facility could be used.

Successfully compiled for knl nodes on haswell login nodes on theta at anl, cori at nersc, and mutrino at sandia with gcc, cce, llvm, and intel compilers.  Tests passed on theta and mutrino with the exception of testphdf5_tldsc with some compilers.  I wasn't able to run batch jobs at nersc;  allocation must have changed.  

32 bit and 64 bit on big-endian Power8 echidna with gcc and IBM XL compilers also built successfully and all serial tests passed.

gnuoyd:
On Fri, Feb 04, 2022 at 01:06:07PM -0800, jhendersonHDF wrote:
> > > Does anybody have an opinion about the precision of int_least8_t and int_fast8_t? Seems like it should be 8 bits rather than `8 * sizeof(int_least8_t)` or `8 * sizeof(int_fast8_t)`, respectively. It's an easy change to make in the `H5T__init_native_int()` table.
> > 
> > I thought those weren't required to be exactly 8 bits, only int8_t is guaranteed to be 8 bits. As far as I can tell this applies to precision as well as width. Then again, strictly speaking, looking at the size isn't guaranteed to give you the correct precision (though I doubt HDF5 would work correctly on a system with weird integer types).
> 
> This is correct. The 'least' types are guaranteed to be at least that size, but could be larger. I think we should generally keep the 'least' types as small as possible since they're more related to minimizing memory usage, while the 'fast' types should of course be the fastest size for the platform.

The question I have is a little bit different than the one I believe you
are answering.

Let us say that int_least8_t is 32 bits wide on platform X.  When
platform X saves H5T_NATIVE_INT_LEAST8-type data to an HDF5 file, do we
want for the in-file representation to represent that the data has 32
bits of precision, or 8 bits of precision?  One the one hand, the use of
H5T_NATIVE_INT_LEAST8 suggests that the user only really intended to use
8 of 32 bits; on the other hand, what if in actual fact they produced
some values outside of the [INT8_MIN, INT8_MAX] interval?

Dave

seanm:
>Let us say that int_least8_t is 32 bits wide on platform X.

And it's not just hypothetical. `int_fast8_t` is 32 bit on OpenBSD (at least on PowerPC).

>on the other hand, what if in actual fact they produced some values outside of the [INT8_MIN, INT8_MAX] interval?

Then they have created a bug I'd say.

fortnern:
> On Fri, Feb 04, 2022 at 01:06:07PM -0800, jhendersonHDF wrote: > > Does anybody have an opinion about the precision of int_least8_t and int_fast8_t? Seems like it should be 8 bits rather than `8 * sizeof(int_least8_t)` or `8 * sizeof(int_fast8_t)`, respectively. It's an easy change to make in the `H5T__init_native_int()` table. > > I thought those weren't required to be exactly 8 bits, only int8_t is guaranteed to be 8 bits. As far as I can tell this applies to precision as well as width. Then again, strictly speaking, looking at the size isn't guaranteed to give you the correct precision (though I doubt HDF5 would work correctly on a system with weird integer types). This is correct. The 'least' types are guaranteed to be at least that size, but could be larger. I think we should generally keep the 'least' types as small as possible since they're more related to minimizing memory usage, while the 'fast' types should of course be the fastest size for the platform.
> The question I have is a little bit different than the one I believe you are answering. Let us say that int_least8_t is 32 bits wide on platform X. When platform X saves H5T_NATIVE_INT_LEAST8-type data to an HDF5 file, do we want for the in-file representation to represent that the data has 32 bits of precision, or 8 bits of precision? One the one hand, the use of H5T_NATIVE_INT_LEAST8 suggests that the user only really intended to use 8 of 32 bits; on the other hand, what if in actual fact they produced some values outside of the [INT8_MIN, INT8_MAX] interval? Dave

We talked about it and decided we should set the precision to be the full width (unless we want to test if it really does have lower precision). The reasoning is we should save what the type actually is in memory, and not try to divine what the user was trying to use it for. Users can expect "native" types to be represented in exactly the same manner on disk as they are in memory.

gnuoyd:
On Fri, Feb 04, 2022 at 12:04:18PM -0800, Neil Fortner wrote:
> Have we looked into the potential performance/memory effects of this
> change? I worked on a similar system a long time ago and we wanted to
> allow unaligned access wherever the system handled it correctly to
> maximize performance (for example, otherwise a compound may need to go
> through a more expensive type conversion process).

I think that sometimes the in-memory representation of a compound may be
larger.  Depends how smart the library is about layout.

Can you recommend an existing program to use to investigate performance
changes?  If we find that there *is* a change, then I am sure that we
can bring the performance back up without invoking undefined behavior.

> On a related note, have you looked at the H5_NO_ALIGNEMNT_RESTRICTIONS
> that is checked in configure and used in H5Tvlen.c?

I see that, now.  Optimizing compilers obviate any need for it.  If you
`memcpy` to the temporary variable instead of invoking the undefined
behavior of loading through a (potentially) unaligned pointer, GCC 8.3.0
still emits a single `mov` instruction for x86_64.

Dave

src/H5Tnative.c
```diff
@@ -940,3 +940,270 @@ H5T__cmp_offset(size_t *comp_size, size_t *offset, size_t elem_size, size_t nele
...
+    return (endian_exemplar.u64 == 1) ? H5T_ORDER_LE : H5T_ORDER_BE;
+}
+
+/* Establish `H5T_t`s for C99 integer types including fixed- and
+ * minimum-width types (uint16_t, uint_least16_t, uint_fast16_t, ...).
+ */
+herr_t
+H5T__init_native_int(void)
+{
+    typedef struct {
```
fortnern, line 990: Probably a short comment here explaining how you're getting the alignment values.

src/H5Tnative.c
```diff
@@ -940,3 +940,270 @@ H5T__cmp_offset(size_t *comp_size, size_t *offset, size_t elem_size, size_t nele
...
+            dt->shared->state          = H5T_STATE_IMMUTABLE;
+            dt->shared->type           = H5T_INTEGER;
+            dt->shared->size           = table[j].size;
+            dt->shared->u.atomic       = table[j].atomic;
+            dt->shared->u.atomic.order = byte_order;
+            *table[j].alignmentp       = table[j].alignment;
+
+            if ((*table[j].hidp = H5I_register(H5I_DATATYPE, dt, FALSE)) < 0)
+                return FAIL;
+        }
```
fortnern, line 1253: Is there a reason you can't put the contents of this loop in the NATIVE_ENTRY_INITIALIZER macro and avoid the strange procedure with the tables? If not some comments should be added to explain what the loop is doing?

fortnern:
> On Fri, Feb 04, 2022 at 12:04:18PM -0800, Neil Fortner wrote: Have we looked into the potential performance/memory effects of this change? I worked on a similar system a long time ago and we wanted to allow unaligned access wherever the system handled it correctly to maximize performance (for example, otherwise a compound may need to go through a more expensive type conversion process).
> I think that sometimes the in-memory representation of a compound may be larger. Depends how smart the library is about layout. Can you recommend an existing program to use to investigate performance changes? If we find that there *is* a change, then I am sure that we can bring the performance back up without invoking undefined behavior.

It looks like these values are only used for H5Tget_native_type(), so this change should be fine overall. I see you removed the comment about it only being used for H5Tget_native_type() though, is there a reason for that?

src/H5Tnative.c
gnuoyd: I added a comment.

src/H5Tnative.c
gnuoyd: Compared with what came before, I use the preprocessor sparingly and replace some code with data.

I can add some general comments.

gnuoyd:
> It looks like these values are only used for H5Tget_native_type(), so this change should be fine overall. I see you removed the comment about it only being used for H5Tget_native_type() though, is there a reason for that?

Looks like a sentence got lost in the shuffle. I added it back.

gnuoyd:
> Will you be doing some similar treatment for floats too?

I will treat floats, yeah.

fortnern (approved):

seanm:
Any news on getting this glorious change merged?

src/H5Tnative.c
```diff
@@ -940,3 +940,324 @@ H5T__cmp_offset(size_t *comp_size, size_t *offset, size_t elem_size, size_t nele
...
+static H5T_order_t
+get_host_byte_order(void)
+{
+    static const union {
+        uint64_t u64;
+        char     byte[8];
+    } endian_exemplar = {.byte = {1}};
+
+    return (endian_exemplar.u64 == 1) ? H5T_ORDER_LE : H5T_ORDER_BE;
+}
```
soumagne, line 971: I think I'd be otherwise in favor of using endian.h when available but that's probably just fine for that purpose

src/H5Tnative.c
```diff
@@ -940,3 +940,324 @@ H5T__cmp_offset(size_t *comp_size, size_t *offset, size_t elem_size, size_t nele
...
+            hobj_ref_t x;
+        } hobjref;
+        struct {
+            char            c;
+            hdset_reg_ref_t x;
+        } hdsetregref;
+        struct {
+            char      c;
+            H5R_ref_t x;
+        } ref;
```
soumagne, line 1146: Maybe you could use a macro to define each of the structs, just to reduce the repetitiveness

src/H5Tnative.c
```diff
@@ -940,3 +940,324 @@ H5T__cmp_offset(size_t *comp_size, size_t *offset, size_t elem_size, size_t nele
...
+    , NATIVE_ENTRY_INITIALIZER(INT_LEAST32, int_least32_t, 0, true)
+    , NATIVE_ENTRY_INITIALIZER(UINT_LEAST32, uint_least32_t, 0, false)
+    , NATIVE_ENTRY_INITIALIZER(INT_FAST32, int_fast32_t, 0, true)
+    , NATIVE_ENTRY_INITIALIZER(UINT_FAST32, uint_fast32_t, 0, false)
+    , NATIVE_ENTRY_INITIALIZER(INT64, int64_t, 0, true)
+    , NATIVE_ENTRY_INITIALIZER(UINT64, uint64_t, 0, false)
+    , NATIVE_ENTRY_INITIALIZER(INT_LEAST64, int_least64_t, 0, true)
+    , NATIVE_ENTRY_INITIALIZER(UINT_LEAST64, uint_least64_t, 0, false)
+    , NATIVE_ENTRY_INITIALIZER(INT_FAST64, int_fast64_t, 0, true)
+    , NATIVE_ENTRY_INITIALIZER(UINT_FAST64, uint_fast64_t, 0, false)
```
soumagne, line 1217: Same here I feel like eventually the generation of the table could be simplified with macros but not a big deal. Or could be grouped with signed / unsigned and could also remove the extra 0.

soumagne (approved):
Definitely looks like a nice improvement over the old H5detect.

derobins (approved):

gnuoyd:
On Thu, Apr 14, 2022 at 09:19:03AM -0700, Sean McBride wrote:
> Any news on getting this glorious change merged?

It is waiting for one more approval.

Dave

seanm:
> On Thu, Apr 14, 2022 at 09:19:03AM -0700, Sean McBride wrote: Any news on getting this glorious change merged?
> It is waiting for one more approval.

Is it?  It shows at "merged". Back in April I think...

PR #1403, opened 2022-01-28
Make a buffer bigger: GCC 8.3.0 warned that a `snprintf` may be

gnuoyd:
truncated because the recipient buffer was too small, and the warning is
one that we promote to an error.  Now there is a warning that the buffer
is too big (-Wlarger-than=), but we don't promote those to errors, yet.


byrnHDF (approved):
I'm okay with this because the file it is in could use a bit of work for other reasons.
I would check the parallel version. pio_engine.c, for the same issue.

lrknox (approved):

PR #1411, opened 2022-01-30
Allow setting FORTRAN module install directory with --with-fmoddir

opoplawski:
Fedora does not install FORTRAN modules into the same location as include files (they are arch specific).  This allows setting the directory to install them into.

Not really tested yet.


configure.ac
```diff
@@ -612,6 +612,13 @@ if test "X$HDF_FORTRAN" = "Xyes"; then
   AC_PROG_FC([PAC_FC_SEARCH_LIST],)
   AC_F9X_MODS
 
+  ## Allow setting the fortran module install dir
+  AC_ARG_WITH([fmoddir],
```
byrnHDF, line 616: Need the CMake equivalent as well.

fortran/src/Makefile.am
```diff
@@ -22,7 +22,7 @@ include $(top_srcdir)/config/lt_vers.am
 # Include src directory in both Fortran and C flags (C compiler is used
 # for linking).
 AM_CPPFLAGS+=-I$(top_srcdir)/src
-AM_FCFLAGS+=-I$(top_srcdir)/src -I$(top_srcdir)/fortran/src
+AM_FCFLAGS+=-I$(top_srcdir)/src -I$(top_srcdir)/fortran/src $(FC_MODINC)$(fmoddir}
```
byrnHDF: There is also a fortran folder in the hl directory.

byrnHDF (changes requested):
Need CMake files updated as well as adding the changes to the hl/fortran files.

fortran/src/Makefile.am
opoplawski: Updated

configure.ac
opoplawski: So, I took a look at the CMake configuration, and I just don't think you have real FORTRAN support there in the first place.  You build h5fc from libhdf5cc.in which uses the stock libhdf5.pc file.  Is that going to contain the fortran libraries as well?

I'm just not very familiar with the HDF5 CMake build process to know for sure.

brtnfld: @byrnHDF does not CMake already have Fortran_MODULE_DIRECTORY/CMAKE_Fortran_MODULE_DIRECTORY for this purpose.

byrnHDF: Yes, but we would need to allow an option to override the default. Plus I would need to override the option that was added last year to copy the fortran mod files to the include folder for install.

opoplawski: So, I've added an initial HDF5_INSTALL_FMOD_DIR cmake change for discussion.   I have to say, from a Fedora packaging standpoint I find to whole /shared /static sub-directory thing very annoying.  We would prefer to build both static and shared libraries at the same time, but only install the shared executables as is done with the autotools build.  At least with gfortran the shared and static module files are identical.

Also, I think Fortran_MODULE_DIRECTORY/CMAKE_Fortran_MODULE_DIRECTORY is really only a build time variable.d

byrnHDF: Right, we added an option to copy the shared mod files to the include directory for the install. We can modify that to copy to the requested fortran folder. While gfortran on a linux machine might create the same file, other platforms and compilers do not; therefore the separate folders.

brtnfld (approved):

CMakeLists.txt
```diff
@@ -1123,6 +1123,14 @@ if (EXISTS "${HDF5_SOURCE_DIR}/fortran" AND IS_DIRECTORY "${HDF5_SOURCE_DIR}/for
         set (HDF5_INSTALL_MOD_FORTRAN "SHARED")
       endif ()
     endif ()
+    if (NOT DEFINED HDF5_INSTALL_FMOD_DIR)
+      set (HDF5_INSTALL_FMOD_DIR ${HDF5_INSTALL_INCLUDE_DIR})
+      set (HDF5_INSTALL_SHARED_FMOD_DIR ${HDF5_INSTALL_INCLUDE_DIR}/shared)
+      set (HDF5_INSTALL_STATIC_FMOD_DIR ${HDF5_INSTALL_INCLUDE_DIR}/static)
+    else ()
+      set (HDF5_INSTALL_SHARED_FMOD_DIR ${HDF5_INSTALL_FMOD_DIR})
```
byrnHDF, line 1131: I have an issue with this being the same folder if both static and shared is being built. One set will overwrite the other.
While this may be okay for the majority of the platform-compiler pairs - the mod files are still objects and are not always equivalent. 
I think it depends on HDF5_INSTALL_MOD_FORTRAN not being "no" - then this is okay if it is "no" then something needs to be done about one or the other. Maybe send static to a subdir and leave shared to the DIR?

byrnHDF:
Finally had time to look at this in depth.
Most of the changes are fine, but we need the documentation too;
Release.txt entry - new configuration - CMake and autotools Fortran mod file location path option.

byrnHDF:
Working on a refactor of the CMake changes, should be available within the hour.
See #1922

derobins:
The changes here were made in #1922

PR #1416, opened 2022-02-02
Fix issue with H5_now_usec() with 32 bit systems

fortnern:
Intermediate steps in the time calculations could cause overflow, which would trigger an abort. This was noticed with a new test for H5ESget_requests(), which is also the first test that calls H5ESinsert_request(), which was failing. Also slightly expand the test for H5ESget_requests (much more testing still needs to be done for event sets in general).


src/H5timer.c
```diff
@@ -193,17 +193,18 @@ H5_now_usec(void)
         struct timespec ts;
 
         HDclock_gettime(CLOCK_MONOTONIC, &ts);
-        now = (uint64_t)(ts.tv_sec * (1000 * 1000)) + (uint64_t)(ts.tv_nsec / 1000);
+        now = ((uint64_t)ts.tv_sec * ((uint64_t)1000 * (uint64_t)1000)) +
+              ((uint64_t)ts.tv_nsec / (uint64_t)1000);
```
gnuoyd, line 200: Those casts obscure the arithmetic. Why not `now = (uint64_t)ts.tv_sec * 1000 * 1000 + ts.tv_nsec / 1000;` ?
Both `*` and `+` should yield a result with precision as great as the highest-precision operand.

src/H5timer.c
```diff
@@ -193,17 +193,18 @@ H5_now_usec(void)
...
     {
         struct timeval now_tv;
 
         HDgettimeofday(&now_tv, NULL);
-        now = (uint64_t)(now_tv.tv_sec * (1000 * 1000)) + (uint64_t)now_tv.tv_usec;
+        now = ((uint64_t)now_tv.tv_sec * ((uint64_t)1000 * (uint64_t)1000)) + (uint64_t)now_tv.tv_usec;
     }
 #else  /* H5_HAVE_GETTIMEOFDAY */
-    now       = (uint64_t)(HDtime(NULL) * (1000 * 1000));
+    now       = ((uint64_t)HDtime(NULL) * ((uint64_t)1000 * (uint64_t)1000));
```
gnuoyd, line 215: Same comment about the casts for gettimeofday() and time().

gnuoyd (changes requested):
Simplify, please.

src/H5timer.c
fortnern: I prefer to be explicit that everything is done in uint64_t, since the precision of the arithmetic is important here. This way you don't need to know (or look up) the peculiarities of C type conversion to understand what is going on. I can remove some of the parentheses if you think they're a problem. Also, I believe the ts.tv_nsec / 1000 operation could otherwise be done in 32 bit.

fortnern: We can maybe pull out the uint64_t from the 1000*1000:

now = ((uint64_t)ts.tv_sec *((uint64_t)(1000 * 1000)) + ((uint64_t)ts.tv_nsec / (uint64_t)1000);

Does that work?

jhendersonHDF: > Those casts obscure the arithmetic. Why not `now = (uint64_t)ts.tv_sec * 1000 * 1000 + ts.tv_nsec / 1000;` ? Both `*` and `+` should yield a result with precision as great as the highest-precision operand.

I tend to also prefer being explicit about the conversion for those in the future that may or may not  understand type conversion well. Would it maybe help to have some intermediate uint64_t variables for the `((uint64_t)1000 * (uint64_t)1000)` and `((uint64_t)ts.tv_nsec / (uint64_t)1000)`? Seems that would make the arithmetic less obscure and still allow for some explicit casts, even if not necessary in some places.

soumagne: I would agree with Dave, i.e. just add the cast to the initial member that needs conversion. No need to hammer it down :)

gnuoyd:
On Wed, Feb 02, 2022 at 10:07:56AM -0800, Neil Fortner wrote:
> @fortnern commented on this pull request.
> 
> 
> 
> > @@ -193,17 +193,18 @@ H5_now_usec(void)
>          struct timespec ts;
>  
>          HDclock_gettime(CLOCK_MONOTONIC, &ts);
> -        now = (uint64_t)(ts.tv_sec * (1000 * 1000)) + (uint64_t)(ts.tv_nsec / 1000);
> +        now = ((uint64_t)ts.tv_sec * ((uint64_t)1000 * (uint64_t)1000)) +
> +              ((uint64_t)ts.tv_nsec / (uint64_t)1000);
> 
> I prefer to be explicit that everything is done in uint64_t, since the precision of the arithmetic is important here. This way you don't need to know (or look up) the peculiarities of C type conversion to understand what is going on. I can remove some of the parentheses if you think they're a problem.

Both the integer promotions and the relative precedence of arithmetic
operators are mainstream knowledge among C programmers, and the way that
they operate in this expression is straightforward.

The simple expression that I provided illuminates the meaning and
operates at great enough precision to produce a reliable uint64_t result
on ILP32, LP64, and LLP64 data models.

The C standard allows for `int` as narrow as 16 bits.  For extra
assurance, we can move the `uint64_t` cast to the first 1000:

```C
now = ts.tv_sec * (uint64_t)1000 * 1000 + ts.tv_nsec / 1000;
```

That product should be evaluated at 64-bit precision regardless of
either the precision of `int` or the associativity of the `*` operator.

Incidentally, the precision problem was probably not detected earlier
because *nobody read the original expression*, which was so badly
obfuscated by excess notation.

If you practice writing fluent C then your fluency will increase
and your colleagues will quickly come up to speed.  If you write C
expressions in the "belt and suspenders" style where every subexpression
is parenthesized *and* cast, then over the long term you lower the
productivity of everybody.  The blizzard of superfluous symbols will
prevent sight-reading by the C experts, and the C novices will not
improve their C-programming knowledge or practices.

Dave

src/H5timer.c
fortnern: I think Dave's suggestion sacrifices clarity for the sake of brevity, which I do not think is a good idea. Others have write access to this PR, if someone else wants to make the change go ahead, but I don't want the commit to have my name on it.

gnuoyd:
I have filed PR #1424 with simple, more correct H5_now_usec arithmetic.

epourmal (approved):
It is better to be safe than brief. 

When I read the code written this way, I understand that the author emphasizes importance of 64-bit precision.  I think the comment would help to explain the intent.

epourmal (commented):
Question: Why it is assumed that clock_gettime cannot fail?

fortnern:
> Question: Why it is assumed that clock_gettime cannot fail?

Yes we should be checking for failure of clock_gettime. I can make that change. Thanks!

fortnern:
> Question: Why it is assumed that clock_gettime cannot fail?

In order to add error checking and be consistent I will need to add error tracking to most of the functions in H5timer.c. I know this has caused consternation due to performance overhead. Is this ok?

epourmal:
I will not  insist on this change :-) Let's not introduce return values.

lrknox (approved):

PR #1424, opened 2022-02-06
H5_now_usec arithmetic fix

gnuoyd:
Sparingly use casts and the constant suffix "L" to ensure that the  H5_now_usec products are performed with 64 bits of precision or more and to quiet GCC 8.3.0's sign-conversion warnings.

While I am here, move the conditional compilation outside of H5_now_usec for clarity.


lrknox:
Fixed by PR #1416.

PR #1426, opened 2022-02-08
Remove H5_NO_ALIGNMENT_RESTRICTIONS

gnuoyd:
Do not conditionally compile code that uses a pointer dereference and assignment to copy a potentially unaligned variable to aligned automatic storage, or vice versa.  Instead, always use naked `memcpy(3)`s. Disassembling the generated code reveals that the `memcpy(3)`s optimize (`-O3`) to a single `mov` instruction for x86_64, which is not strict about alignment.

This change reduces the size of code and scripts by 143 lines, eases our way to cross-compilation, and avoids invoking undefined behavior.

### DIFF
--- a/src/H5Tvlen.c
+++ b/src/H5Tvlen.c
@@ -377,7 +377,10 @@ H5T__vlen_mem_seq_getlen(H5VL_object_t H5_ATTR_UNUSED *file, const void *_vl, si
     HDassert(_vl);
     HDassert(len);
 
-    memcpy(&vl, _vl, sizeof(hvl_t));
+    /* Copy to ensure correct alignment.  memcpy is best here because
+     * it optimizes to fast code.
+     */
+    HDmemcpy(&vl, _vl, sizeof(hvl_t));
 
     *len = vl.len;
 
@@ -405,7 +408,8 @@ H5T__vlen_mem_seq_getptr(void *_vl)
 
     /* check parameters, return result */
     HDassert(_vl);
-    memcpy(&vl, _vl, sizeof(hvl_t));
+    /* Copy to ensure correct alignment. */
+    HDmemcpy(&vl, _vl, sizeof(hvl_t));
 
     FUNC_LEAVE_NOAPI(vl.p)
 } /* end H5T__vlen_mem_seq_getptr() */
@@ -432,7 +436,8 @@ H5T__vlen_mem_seq_isnull(const H5VL_object_t H5_ATTR_UNUSED *file, void *_vl, hb
     /* Check parameters */
     HDassert(_vl);
 
-    memcpy(&vl, _vl, sizeof(hvl_t));
+    /* Copy to ensure correct alignment. */
+    HDmemcpy(&vl, _vl, sizeof(hvl_t));
 
     *isnull = ((vl.len == 0 || vl.p == NULL) ? TRUE : FALSE);
 
@@ -493,7 +498,8 @@ H5T__vlen_mem_seq_read(H5VL_object_t H5_ATTR_UNUSED *file, void *_vl, void *buf,
     /* check parameters, copy data */
     HDassert(buf);
     HDassert(_vl);
-    memcpy(&vl, _vl, sizeof(hvl_t));
+    /* Copy to ensure correct alignment. */
+    HDmemcpy(&vl, _vl, sizeof(hvl_t));
     HDassert(vl.p);
 
     H5MM_memcpy(buf, vl.p, len);
@@ -577,7 +583,8 @@ H5T__vlen_mem_str_getlen(H5VL_object_t H5_ATTR_UNUSED *file, const void *_vl, si
     /* check parameters */
     HDassert(_vl);
 
-    memcpy(&s, _vl, sizeof(char *));
+    /* Copy to ensure correct alignment. */
+    HDmemcpy(&s, _vl, sizeof(char *));
 
     *len = HDstrlen(s);
 
@@ -605,7 +612,8 @@ H5T__vlen_mem_str_getptr(void *_vl)
 
     /* check parameters */
     HDassert(_vl);
-    memcpy(&s, _vl, sizeof(char *));
+    /* Copy to ensure correct alignment. */
+    HDmemcpy(&s, _vl, sizeof(char *));
 
     FUNC_LEAVE_NOAPI(s)
 } /* end H5T__vlen_mem_str_getptr() */
@@ -629,7 +637,8 @@ H5T__vlen_mem_str_isnull(const H5VL_object_t H5_ATTR_UNUSED *file, void *_vl, hb
 
     FUNC_ENTER_STATIC_NOERR
 
-    memcpy(&s, _vl, sizeof(char *));
+    /* Copy to ensure correct alignment. */
+    HDmemcpy(&s, _vl, sizeof(char *));
 
     *isnull = (s == NULL ? TRUE : FALSE);
 
@@ -684,7 +693,8 @@ H5T__vlen_mem_str_read(H5VL_object_t H5_ATTR_UNUSED *file, void *_vl, void *buf,
         /* check parameters */
         HDassert(buf);
         HDassert(_vl);
-        memcpy(&s, _vl, sizeof(char *));
+        /* Copy to ensure correct alignment. */
+        HDmemcpy(&s, _vl, sizeof(char *));
 
         H5MM_memcpy(buf, s, len);
     } /* end if */
--- a/src/H5Tvlen.c
+++ b/src/H5Tvlen.c
@@ -369,11 +369,7 @@ done:
 static herr_t
 H5T__vlen_mem_seq_getlen(H5VL_object_t H5_ATTR_UNUSED *file, const void *_vl, size_t *len)
 {
-#ifdef H5_NO_ALIGNMENT_RESTRICTIONS
-    const hvl_t *vl = (const hvl_t *)_vl; /* Pointer to the user's hvl_t information */
-#else
     hvl_t vl; /* User's hvl_t information */
-#endif
 
     FUNC_ENTER_STATIC_NOERR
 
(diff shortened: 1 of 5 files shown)


fortnern:
I know this code was put in to avoid crashes we were seeing on some systems. Maybe Sun? It's possible it will never cause issues on gcc but we should make sure it won't cause issues on any of our supported platforms.

gnuoyd:
On Wed, Feb 09, 2022 at 07:29:58AM -0800, Neil Fortner wrote:
> I know this code was put in to avoid crashes we were seeing on some systems. Maybe Sun? It's possible it will never cause issues on gcc but we should make sure it won't cause issues on any of our supported platforms.

Testing is always a good idea, but my change removes the branch of
conditional compilation that could crash SPARC and other platforms,
produce the wrong result on certain ARM variants, et cetera.

Dave

src/H5Tvlen.c
```diff
@@ -369,25 +369,17 @@ H5T__vlen_set_loc(H5T_t *dt, H5VL_object_t *file, H5T_loc_t loc)
...
 
     /* Check parameter */
     HDassert(_vl);
     HDassert(len);
 
-#ifdef H5_NO_ALIGNMENT_RESTRICTIONS
-    *len = vl->len;
-#else
-    H5MM_memcpy(&vl, _vl, sizeof(hvl_t));
+    memcpy(&vl, _vl, sizeof(hvl_t));
```
byrnHDF: Is the change from H5MM_memcpy to just memcpy right?

gnuoyd:
On Wed, Mar 02, 2022 at 08:31:51AM -0800, Allen Byrne wrote:
> @byrnHDF commented on this pull request.
> -#ifdef H5_NO_ALIGNMENT_RESTRICTIONS
> -    *len = vl->len;
> -#else
> -    H5MM_memcpy(&vl, _vl, sizeof(hvl_t));
> +    memcpy(&vl, _vl, sizeof(hvl_t));
> 
> Is the change from H5MM_memcpy to just memcpy right?

Yes, because the compiler can and does inline memcpy() on x86 when
optimization is enabled, so the compiler will generate the same code as
before.

Dave

jhendersonHDF:
> On Wed, Mar 02, 2022 at 08:31:51AM -0800, Allen Byrne wrote: @byrnHDF commented on this pull request. -#ifdef H5_NO_ALIGNMENT_RESTRICTIONS - *len = vl->len; -#else - H5MM_memcpy(&vl, _vl, sizeof(hvl_t)); + memcpy(&vl, _vl, sizeof(hvl_t)); Is the change from H5MM_memcpy to just memcpy right?
> Yes, because the compiler can and does inline memcpy() on x86 when optimization is enabled, so the compiler will generate the same code as before. Dave

Unless H5_MEMORY_ALLOC_SANITY_CHECK is enabled, in which case it seems this can cause problems.

gnuoyd:
On Wed, Mar 02, 2022 at 08:51:11AM -0800, jhendersonHDF wrote:
> > On Wed, Mar 02, 2022 at 08:31:51AM -0800, Allen Byrne wrote: @byrnHDF commented on this pull request. -#ifdef H5_NO_ALIGNMENT_RESTRICTIONS - *len = vl->len; -#else - H5MM_memcpy(&vl, _vl, sizeof(hvl_t)); + memcpy(&vl, _vl, sizeof(hvl_t)); Is the change from H5MM_memcpy to just memcpy right?
> > Yes, because the compiler can and does inline memcpy() on x86 when optimization is enabled, so the compiler will generate the same code as before. Dave
> 
> Unless H5_MEMORY_ALLOC_SANITY_CHECK is enabled, in which case it seems this can cause problems.

Which problems are those?

Dave

jhendersonHDF:
> On Wed, Mar 02, 2022 at 08:51:11AM -0800, jhendersonHDF wrote: > On Wed, Mar 02, 2022 at 08:31:51AM -0800, Allen Byrne wrote: @byrnHDF commented on this pull request. -#ifdef H5_NO_ALIGNMENT_RESTRICTIONS - *len = vl->len; -#else - H5MM_memcpy(&vl, _vl, sizeof(hvl_t)); + memcpy(&vl, _vl, sizeof(hvl_t)); Is the change from H5MM_memcpy to just memcpy right? > Yes, because the compiler can and does inline memcpy() on x86 when optimization is enabled, so the compiler will generate the same code as before. Dave Unless H5_MEMORY_ALLOC_SANITY_CHECK is enabled, in which case it seems this can cause problems.
> Which problems are those? Dave

Note that I'm not well-versed enough in this particular section of the code to say definitively that this will cause problems, but if H5_MEMORY_ALLOC_SANITY_CHECK is enabled, then the various _vl parameters will presumably have been allocated with our heap canary info in some of the leading bytes (via H5MM_malloc), so a direct memcpy will likely not be valid.

gnuoyd:
On Wed, Mar 02, 2022 at 09:20:11AM -0800, jhendersonHDF wrote:
> Note that I'm not well-versed enough in this particular section of the code to say definitively that this will cause problems, but if H5_MEMORY_ALLOC_SANITY_CHECK is enabled, then the various _vl parameters will presumably have been allocated with our heap canary info in some of the leading bytes (via H5MM_malloc), so a direct memcpy will likely not be valid.

H5MM_memcpy doesn't tangle with the canaries.  It's fine.

Dave

jhendersonHDF:
> On Wed, Mar 02, 2022 at 09:20:11AM -0800, jhendersonHDF wrote: Note that I'm not well-versed enough in this particular section of the code to say definitively that this will cause problems, but if H5_MEMORY_ALLOC_SANITY_CHECK is enabled, then the various _vl parameters will presumably have been allocated with our heap canary info in some of the leading bytes (via H5MM_malloc), so a direct memcpy will likely not be valid.
> H5MM_memcpy doesn't tangle with the canaries. It's fine. Dave

My mistake then; it's been a while since I've gone through the H5MM code. This is probably ok then, notwithstanding the fact that we should use HD prefixed versions of memcpy until such a time as we decide to remove those all at once.

fortnern:
> On Wed, Feb 09, 2022 at 07:29:58AM -0800, Neil Fortner wrote: I know this code was put in to avoid crashes we were seeing on some systems. Maybe Sun? It's possible it will never cause issues on gcc but we should make sure it won't cause issues on any of our supported platforms.
> Testing is always a good idea, but my change removes the branch of conditional compilation that could crash SPARC and other platforms, produce the wrong result on certain ARM variants, et cetera. Dave

My mistake, looks like you changed it to always use the "safe but slow" behavior. Are there systems where the configure check would fail to detect that the fast code would cause problems? Have you investigated the performance penalty of this change? What does the compiler produce for the "fast" code?

fortnern:
> On Wed, Mar 02, 2022 at 08:31:51AM -0800, Allen Byrne wrote: @byrnHDF commented on this pull request. -#ifdef H5_NO_ALIGNMENT_RESTRICTIONS - *len = vl->len; -#else - H5MM_memcpy(&vl, _vl, sizeof(hvl_t)); + memcpy(&vl, _vl, sizeof(hvl_t)); Is the change from H5MM_memcpy to just memcpy right?
> Yes, because the compiler can and does inline memcpy() on x86 when optimization is enabled, so the compiler will generate the same code as before. Dave

Have you checked if the compiler inlines H5MM_memcpy() with assertions off and max optimization? If not, can you think of a way to make that happen? Maybe place it in a header file or turn it into a macro? We should try to avoid removing assertions from working code.

gnuoyd:
> My mistake, looks like you changed it to always use the "safe but slow" behavior. Are there systems where the configure check would fail to detect that the fast code would cause problems? Have you investigated the performance penalty of this change? What does the compiler produce for the "fast" code?

It's not "safe but slow" behavior.

The `memcpy` is sound and it lets the compiler generate the fastest code it can for the target architecture. On x86_64, I know that it is a single `mov` instruction because I checked.

gnuoyd:
> Have you checked if the compiler inlines H5MM_memcpy() with assertions off and max optimization? If not, can you think of a way to make that happen? Maybe place it in a header file or turn it into a macro? We should try to avoid removing assertions from working code.

The effort isn't justified: in context, the asserted conditions are redundant, undefined, or impossible.

qkoziol:
> > Have you checked if the compiler inlines H5MM_memcpy() with assertions off and max optimization? If not, can you think of a way to make that happen? Maybe place it in a header file or turn it into a macro? We should try to avoid removing assertions from working code.
> 
> The effort isn't justified: in context, the asserted conditions are redundant, undefined, or impossible.

I agree with Neil - converting H5MM_memcpy() to a macro would allow the extra sanity checking to be retained in debug builds.

Additionally, all the "memcpy" calls need an "HD" prefix.

seanm:
I'm no HDF5 expert, but I've just reviewed this change, and it looks glorious to me.

For many years now, compilers are very good at optimizing `memcpy`, trying to hack something faster than `memcpy` is no longer necessary.

As for 'extra sanity checking', I'm not aware what facilities HDF5 itself provides, but these days we have [Address Sanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer), [Undefined Behaviour Sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), [valgrind](https://valgrind.org/), [FORTIFY_SOURCE](https://www.redhat.com/en/blog/security-technologies-fortifysource), and so much more.  Would a `H5MM_memcpy` macro have 'extra sanity checking' that those tools don't cover?

fortnern (approved):
I still think there is value in keeping assertions even if they're not *currently* triggerable (that's kind of the point of assertions...), but I'm not going to die on this particular hill

derobins:
I'm going to approve this and merge it in. The reason for the H5MM_memcpy call was to watch for buffer overlap, which caught some bugs when introduced. Agree that other tools should be substituted for H5MM calls, though.

derobins (approved):

lrknox (approved):

PR #1429, opened 2022-02-09
(1.10 merge) Fix metadata cache bug when resizing a pinned/protected entry (#1358)

jhendersonHDF:
When resizing a pinned/protected cache entry, the metadata
cache code previously would wait until after resizing the
entry to attempt to log the newly-dirtied entry. This would
cause H5C_resize_entry to mark the entry as dirty and make
H5AC_resize_entry think that it doesn't need to add the
newly-dirtied entry to the dirty entries skiplist.

Thus, a subsequent H5AC__log_moved_entry would think it
needs to allocate a new entry for insertion into the dirty
entry skip list, since the entry doesn't exist on that list.
This causes an assertion failure, as the code to allocate a
new entry assumes that the entry is not dirty.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -119,6 +119,23 @@ Bug Fixes since HDF5-1.10.7 release
 ===================================
     Library
     -------
+    - Fixed a metadata cache bug when resizing a pinned/protected cache entry
+
+      When resizing a pinned/protected cache entry, the metadata
+      cache code previously would wait until after resizing the
+      entry to attempt to log the newly-dirtied entry. This would
+      cause H5C_resize_entry to mark the entry as dirty and make
+      H5AC_resize_entry think that it doesn't need to add the
+      newly-dirtied entry to the dirty entries skiplist.
+    
+      Thus, a subsequent H5AC__log_moved_entry would think it
+      needs to allocate a new entry for insertion into the dirty
+      entry skip list, since the entry doesn't exist on that list.
+      This causes an assertion failure, as the code to allocate a
+      new entry assumes that the entry is not dirty.
+
+      (JTH - 2022/01/12)
+
     - Unified handling of collective metadata reads to correctly fix old bugs
 
       Due to MPI-related issues occurring in HDF5 from mismanagement of the
--- a/src/H5AC.c
+++ b/src/H5AC.c
@@ -1514,10 +1514,6 @@ H5AC_resize_entry(void *thing, size_t new_size)
     cache_ptr = entry_ptr->cache_ptr;
     HDassert(cache_ptr);
 
-    /* Resize the entry */
-    if (H5C_resize_entry(thing, new_size) < 0)
-        HGOTO_ERROR(H5E_CACHE, H5E_CANTRESIZE, FAIL, "can't resize entry")
-
 #ifdef H5_HAVE_PARALLEL
     {
         H5AC_aux_t *aux_ptr;
@@ -1528,6 +1524,10 @@ H5AC_resize_entry(void *thing, size_t new_size)
                 HGOTO_ERROR(H5E_CACHE, H5E_CANTMARKDIRTY, FAIL, "can't log dirtied entry")
     }
 #endif /* H5_HAVE_PARALLEL */
+
+    /* Resize the entry */
+    if (H5C_resize_entry(thing, new_size) < 0)
+        HGOTO_ERROR(H5E_CACHE, H5E_CANTRESIZE, FAIL, "can't resize entry")
 
 done:
     /* If currently logging, generate a message */


qkoziol (changes requested):
I think the logical ordering of the previous code was better, and was correct in the case of H5C_resize_entry() failing.   I would suggest getting the dirty status of the entry before calling H5C_resize_entry() and then change the "if ((!entry_ptr->is_dirty) && (NULL != aux_ptr))" line to "if (!was_dirty && (entry_ptr->is_dirty) && (NULL != aux_ptr))" - i.e. the dirty state for the entry changed from not dirty to dirty after calling H5C_resize_entry().

Good catch though!

jhendersonHDF:
> I think the logical ordering of the previous code was better, and was correct in the case of H5C_resize_entry() failing. I would suggest getting the dirty status of the entry before calling H5C_resize_entry() and then change the "if ((!entry_ptr->is_dirty) && (NULL != aux_ptr))" line to "if (!was_dirty && (entry_ptr->is_dirty) && (NULL != aux_ptr))" - i.e. the dirty state for the entry changed from not dirty to dirty after calling H5C_resize_entry().
> 
> Good catch though!

Absolutely agree on the logical ordering being a little more, well, logical, in the previous case. The reason I went forward with this change was because similar routines in the cache code, like H5AC_move_entry, H5AC_mark_entry_dirty, H5AC_mark_entry_clean and so on follow the same pattern of "log doing the thing first, then do the thing". But there are also routines in the cache code that "do the thing, then log having done the thing", so it's not really consistent. I could fix this case with your suggestion, but likely the other cases need fixed as well.

qkoziol:
> > I think the logical ordering of the previous code was better, and was correct in the case of H5C_resize_entry() failing. I would suggest getting the dirty status of the entry before calling H5C_resize_entry() and then change the "if ((!entry_ptr->is_dirty) && (NULL != aux_ptr))" line to "if (!was_dirty && (entry_ptr->is_dirty) && (NULL != aux_ptr))" - i.e. the dirty state for the entry changed from not dirty to dirty after calling H5C_resize_entry().
> > Good catch though!
> 
> Absolutely agree on the logical ordering being a little more, well, logical, in the previous case. The reason I went forward with this change was because similar routines in the cache code, like H5AC_move_entry, H5AC_mark_entry_dirty, H5AC_mark_entry_clean and so on follow the same pattern of "log doing the thing first, then do the thing". But there are also routines in the cache code that "do the thing, then log having done the thing", so it's not really consistent. I could fix this case with your suggestion, but likely the other cases need fixed as well.

Well, in this case, there's non-error paths (as well as error paths) in H5C_resize_entry() that don't mark the entry dirty (i.e. if the size of the entry doesn't change), so this change creates a new bug in those cases.   The change I suggested avoids both bugs. ;-)

(I think that it would be a good idea to clean the inconsistency up too  :-) )

jhendersonHDF:
> > > I think the logical ordering of the previous code was better, and was correct in the case of H5C_resize_entry() failing. I would suggest getting the dirty status of the entry before calling H5C_resize_entry() and then change the "if ((!entry_ptr->is_dirty) && (NULL != aux_ptr))" line to "if (!was_dirty && (entry_ptr->is_dirty) && (NULL != aux_ptr))" - i.e. the dirty state for the entry changed from not dirty to dirty after calling H5C_resize_entry().
> > > Good catch though!
> > 
> > 
> > Absolutely agree on the logical ordering being a little more, well, logical, in the previous case. The reason I went forward with this change was because similar routines in the cache code, like H5AC_move_entry, H5AC_mark_entry_dirty, H5AC_mark_entry_clean and so on follow the same pattern of "log doing the thing first, then do the thing". But there are also routines in the cache code that "do the thing, then log having done the thing", so it's not really consistent. I could fix this case with your suggestion, but likely the other cases need fixed as well.
> 
> Well, in this case, there's non-error paths (as well as error paths) in H5C_resize_entry() that don't mark the entry dirty (i.e. if the size of the entry doesn't change), so this change creates a new bug in those cases. The change I suggested avoids both bugs. ;-)
> 
> (I think that it would be a good idea to clean the inconsistency up too :-) )

Ah ok, I briefly looked through the code and see your point about the non-error paths being problematic with this change. I'll need to fix this in develop since this was merged there. It looks like resizing an entry to the same size may be something we're not testing thoroughly enough. I turned on the most debugging and sanity checking possible for the cache code and this change passed all the library's tests, so we may need some extra coverage.

qkoziol:
Cool - Sorry I didn't notice it earlier when the PR was for the develop branch.   I'm starting to get back to a more normal balance now, so am trying to keep an eye on the PRs more.   (Yes, another parallel test that hit this case would be good too)

jhendersonHDF:
> Cool - Sorry I didn't notice it earlier when the PR was for the develop branch. I'm starting to get back to a more normal balance now, so am trying to keep an eye on the PRs more. (Yes, another parallel test that hit this case would be good too)

So I tried implementing what you suggested, but a new assertion crops up because H5AC__log_dirtied_entry has an assert of the form:

```HDassert(entry_ptr->is_dirty == FALSE);```

so the routine's expecting the entry to not be dirty before logging that it is and then actually making it dirty. Interestingly enough, H5AC__log_cleaned_entry has the exact same assertion, which is... confusing. So it seems like this is why the other similar functions were structured in the "log doing the thing, then do the thing" format.

I can update H5AC__log_dirtied_entry to expect the entry to be dirty, but I'm not sure about the current logic in the routine and whether the change is going to mess with things.

qkoziol:
> > Cool - Sorry I didn't notice it earlier when the PR was for the develop branch. I'm starting to get back to a more normal balance now, so am trying to keep an eye on the PRs more. (Yes, another parallel test that hit this case would be good too)
> 
> So I tried implementing what you suggested, but a new assertion crops up because H5AC__log_dirtied_entry has an assert of the form:
> 
> `HDassert(entry_ptr->is_dirty == FALSE);`
> 
> so the routine's expecting the entry to not be dirty before logging that it is and then actually making it dirty. Interestingly enough, H5AC__log_cleaned_entry has the exact same assertion, which is... confusing. So it seems like this is why the other similar functions were structured in the "log doing the thing, then do the thing" format.
> 
> I can update H5AC__log_dirtied_entry to expect the entry to be dirty, but I'm not sure about the current logic in the routine and whether the change is going to mess with things.

I agree - that assert in H5AC__log_dirty_entry looks wrong.   However, since it was being invoked with ""if ((!entry_ptr->is_dirty) && (NULL != aux_ptr))" before your change, it makes me believe that the codepath has _never_ been exercised (which is strange, too), so I believe that you should be OK to change the assert to be "HDassert(entry_ptr->is_dirty);" and then probably do the refactoring to clean up the inconsistencies that we've been talking about to make certain all the callers are making the same assumptions, etc.

jhendersonHDF:
> > > Cool - Sorry I didn't notice it earlier when the PR was for the develop branch. I'm starting to get back to a more normal balance now, so am trying to keep an eye on the PRs more. (Yes, another parallel test that hit this case would be good too)
> > 
> > 
> > So I tried implementing what you suggested, but a new assertion crops up because H5AC__log_dirtied_entry has an assert of the form:
> > `HDassert(entry_ptr->is_dirty == FALSE);`
> > so the routine's expecting the entry to not be dirty before logging that it is and then actually making it dirty. Interestingly enough, H5AC__log_cleaned_entry has the exact same assertion, which is... confusing. So it seems like this is why the other similar functions were structured in the "log doing the thing, then do the thing" format.
> > I can update H5AC__log_dirtied_entry to expect the entry to be dirty, but I'm not sure about the current logic in the routine and whether the change is going to mess with things.
> 
> I agree - that assert in H5AC__log_dirty_entry looks wrong. However, since it was being invoked with ""if ((!entry_ptr->is_dirty) && (NULL != aux_ptr))" before your change, it makes me believe that the codepath has _never_ been exercised (which is strange, too), so I believe that you should be OK to change the assert to be "HDassert(entry_ptr->is_dirty);" and then probably do the refactoring to clean up the inconsistencies that we've been talking about to make certain all the callers are making the same assumptions, etc.

I think it's wrong in the more logical sense, but it seems like the correct thing to do for the way that the code has been originally structured. It seems the code was clearly originally intended to log an action first and then perform the actual action in most cases. I think that these codepaths have been well-exercised, I just think that the original approach of logging before doing may not have been the best.

In any case, I'm not familiar enough with the code to know if making these changes will cause new problems, but I'll try testing it with all the cache sanity checks enabled and we might need to revisit.

jhendersonHDF:
It seems like this is not as trivial of a fix as we'd like it to be. Only H5AC__log_dirtied_entry and H5AC__log_cleaned_entry were able to be updated so that the logging occurs after performing an operation. However, H5AC_unprotect uses H5AC__log_dirtied_entry and in this case, the log must come before the H5C_unprotect call because H5C_unprotect might flush the cache entry and set its cache pointer to NULL, which results in an assertion in H5AC__log_dirtied_entry that expects it not to be NULL. Between the parallel compression tests that show these bugs and the actual cache tests, I can't seem to come up with a fix that agrees with both of them.

jhendersonHDF:
@qkoziol What if I just add a check for ```entry_ptr->size != new_size``` before logging the entry as dirtied and then doing the resize operation? It might be a little frail, but that's the check that H5C_resize_entry does anyway and it seems clear to me that the cache code was written for the "log an operation -> do the operation" pattern. Changing that assumption seems like it will need a big rewrite of the cache code.

PR #1430, opened 2022-02-10
Fixed H5Dget_storage_size_f for zero sized datasets

brtnfld:
The return value is not set if the size is zero. Instead, it was changed to always return the size.

Also, it should be changed never to return an error since the C API can never return an error status, but I left it as returning an error if zero size for now.


epourmal (approved):
Looks good to me.

byrnHDF (approved):

PR #1437, opened 2022-02-12
Revert "Fix metadata cache bug when resizing a pinned/protected entry"

lrknox:
Reverts HDFGroup/hdf5#1358

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -1082,23 +1082,6 @@ Bug Fixes since HDF5-1.12.0 release
 ===================================
     Library
     -------
-    - Fixed a metadata cache bug when resizing a pinned/protected cache entry
-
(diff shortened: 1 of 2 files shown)


jhendersonHDF (approved):
Reverting this partial fix for now; some context: the use of H5AC__log_dirtied_entry here is a bit of a special case, as the other places that use it call it before performing an operation, rather than after. When resizing an entry, though, we need to log the dirtied entry _after_ resizing it, or we might either:

- log the wrong amount of dirty bytes if the size of the entry changed
- incorrectly log a dirtied entry if the size of the entry didn't change

H5AC__log_dirtied_entry seems to have been written with the assumption that it will be called before performing an operation, so not adhering to that assumption in this case is what's eventually causing problems. However, just re-ordering the operations is not correct in this case, for the reasons above. A correct fix for this lies somewhere deeper.

PR #1440, opened 2022-02-15
Fix issue with copying null new references

fortnern:
Also fix assertion failure when reference copying fails. This does not introduce a new API function to set references to NULL.

The check for NULL new references is based on the check in H5T__ref_mem_isnull(), so this change is not introducing a new definition for what constitutes a NULL new reference, simply applying the existing definition to object copying.

For now, to create a NULL reference the user must make sure it is initialized to all 0 bytes (i.e. with calloc or memset). I'd like to hear opinions on how necessary a H5Rset_null() API function is - I think it's a good idea since requiring a specific bytestream (even if it's all 0s) seems hackish.

### DIFF
--- a/src/H5Ocache.c
+++ b/src/H5Ocache.c
@@ -346,7 +346,7 @@ H5O__cache_deserialize(const void *image, size_t len, void *_udata, hbool_t *dir
 done:
     /* Release the [possibly partially initialized] object header on errors */
     if (!ret_value && oh)
-        if (H5O__free(oh) < 0)
+        if (H5O__free(oh, FALSE) < 0)
             HDONE_ERROR(H5E_OHDR, H5E_CANTRELEASE, NULL, "unable to destroy object header data")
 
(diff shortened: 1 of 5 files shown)


byrnHDF (approved):

soumagne:
Hmm I believe I was making a similar comparison in the past. I can't remember when or why this might have been removed. I should probably look at the history...

soumagne:
Oh nvm, I see this is for new references. I think it might have been missed then, I think that makes sense to apply the same definition as `H5T__ref_mem_isnull`, we could also maybe improve it eventually to avoid the memcmp.

lrknox (approved):

epourmal:
> 

I think it is a good idea to provide helper functions for setting "NULL" for all complex types including references. I don't think it is the highest priority at this point. Do we need to open an issue?

soumagne (approved):

PR #1423, opened 2022-02-04
Sprinkle H5_ATTR_FORMAT over printf(3)-like functions in tools and fix issues

gnuoyd:

PR #1444, opened 2022-02-17
Oess 168 utils mirror vfd

lrknox:
Eliminates 8 clang warnings listed in PR #1310 adding 1 new ssize_t declarations after discussion.


utils/mirror_vfd/mirror_server.c
```diff
@@ -476,7 +476,7 @@ handle_requests(struct server_run *run)
 {
     int              connfd = -1;                       /**/
     char             mybuf[H5FD_MIRROR_XMIT_OPEN_SIZE]; /**/
-    int              ret;                               /* general-purpose error-checking */
+    ssize_t          ret;                               /* general-purpose error-checking */
```
lrknox, line 479: I think this can be long instead;  I must have copied this from the other PR.  Will check.

utils/mirror_vfd/mirror_writer.c
```diff
@@ -788,17 +788,16 @@ do_write(struct mirror_session *session, const unsigned char *xmit_buf)
...
-        nbytes_in_packet = HDread(session->sockfd, buf, H5FD_MIRROR_DATA_BUFFER_MAX);
-        if (-1 == nbytes_in_packet) {
+        if ((nbytes_in_packet = HDread(session->sockfd, buf, H5FD_MIRROR_DATA_BUFFER_MAX)) < 0) {
             mirror_log(session->loginfo, V_ERR, "can't read into databuffer");
             reply_error(session, "can't read data buffer");
             return -1;
         }
 
-        mirror_log(session->loginfo, V_INFO, "received %zd bytes", nbytes_in_packet);
+        mirror_log(session->loginfo, V_INFO, "received %zd bytes", (size_t)nbytes_in_packet);
```
gnuoyd: No cast should be necessary.

utils/mirror_vfd/mirror_server.c
gnuoyd: Why not ssize_t? This is not a POSIX program? It uses sigaction!

utils/mirror_vfd/mirror_writer.c
lrknox: Will remove this cast.

utils/mirror_vfd/mirror_server.c
```diff
@@ -456,7 +456,7 @@ accept_connection(struct server_run *run)
  * ---------------------------------------------------------------------------
  */
 static void
-wait_for_child(int sig)
+wait_for_child()
```
gnuoyd: A warning doesn't crop up at the assignment to `sa.sa_handler` in `handle_requests`? I think that if `H5_ATTR_UNUSED` is available in this context, then `int H5_ATTR_UNUSED sig` is appropriate.

utils/mirror_vfd/mirror_server.c
lrknox: Response to @derobins  comments at bottom of PR #1310 that ssize_t was not in C standard and should be avoided in new code.  Change from int to either ssize_t or to long will satisfy clang warning.

gnuoyd: It's perfectly fine to use a `ssize_t` (and `read`, for that matter) in a POSIX program. Looks like a POSIX program to me.

byrnHDF: Why? There is no restriction in the build code - should work on all supported platforms.

gnuoyd: Looks like it's restricted to systems having `fork(2)`?

```
# ----------------------------------------------------------------------
# Check whether we can build the Mirror VFD
# Header-check flags set in config/cmake_ext_mod/ConfigureChecks.cmake
# ----------------------------------------------------------------------
option (HDF5_ENABLE_MIRROR_VFD "Build the Mirror Virtual File Driver" OFF)
if (HDF5_ENABLE_MIRROR_VFD)
  if ( ${HDF_PREFIX}_HAVE_NETINET_IN_H AND
       ${HDF_PREFIX}_HAVE_NETDB_H      AND
       ${HDF_PREFIX}_HAVE_ARPA_INET_H  AND
       ${HDF_PREFIX}_HAVE_SYS_SOCKET_H AND
       ${HDF_PREFIX}_HAVE_FORK)
      set (${HDF_PREFIX}_HAVE_MIRROR_VFD 1)
  else()
    message(WARNING "The socket-based Mirror VFD was requested but cannot be bui
lt. System prerequisites are not met.")
  endif()
endif()
```

gnuoyd: (Not to mention all of those other prerequisites.)

utils/mirror_vfd/mirror_server.c
lrknox: I think "int sig" was removed in PR #1310 to eliminate  this clang warning:  mirror_server.c:459:20: warning: unused parameter 'sig' [-Wunused-parameter]
wait_for_child(int sig)
                   ^
It does however add 2 warnings for GCC 10.2:
utils/mirror_vfd/mirror_server.c:459:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
<   459 | wait_for_child()
<       | ^~~~~~~~~~~~~~
utils/mirror_vfd/mirror_server.c:459:1: warning: old-style function definition [-Wold-style-definition]

Using "wait for child(int H5_ATTR_UNUSED sig)" as suggested removes that clang warning without adding the GCC warnings.  Will add.  Thanks!

utils/mirror_vfd/mirror_server.c
byrnHDF: Oh, okay. I wonder though if we should should try to avoid piling on more restrictions preventing future expansion.

gnuoyd (approved):
LGTM.

lrknox:
Thanks.  Also, please check the comment in line 4529 of h5dump_xml.c in the PR #1423.  Was that intended to be kept?


byrnHDF (approved):

PR #1445, opened 2022-02-17
Updated VFD SWMR's Akadio license

derobins:


examples/h5_dtransform.c
```diff
@@ -1,6 +1,5 @@
 /* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
  * Copyright by The HDF Group.                                               *
- * Copyright by the Board of Trustees of the University of Illinois.         *
```
bmribler, line 3: Doesn't this UofI copyright need to stay for files created before 2006, when THG spinned off?  You might want to check with Les and Steve.

PR #1451, opened 2022-02-18
VFD SWMR: Aux process changes

derobins:
Removes the submodule, license changes, etc.


vchoi-hdfgroup:
Do we need to be concerned about the two CI failing checks?

derobins:
> Do we need to be concerned about the two CI failing checks?

It looks like a github problem and not a problem with the code. github is trying to download the ninja files and can't find them. The PR check yml files are identical with develop.

PR #1454, opened 2022-02-21
Replaced Richard's hyperslab selection because it has some bugs

raylu-hdf:
The new algorithm has the limitation that the hyperslab selection may be too big for the memory.  It doesn't subdivide the selection.


raylu-hdf:
Scot needs the changes for his test.  Merge it before review.

PR #1468, opened 2022-03-01
Disable memory alloc sanity checks by default for Autotools debug builds

jhendersonHDF:
Disable memory allocation sanity checking for Autotools debug builds as well. This mirrors what CMake is currently doing and should prevent issues with HDF5 filter plugins and checked-out-from-source HDF5 builds where debug is the default build mode. In the future, if we add a 'developer' build mode, we should consider enabling this option by default just for those types of builds.


derobins (approved):

derobins:
Approved but this needs a RELEASE.txt note

byrnHDF (approved):

lrknox (approved):

PR #1473, opened 2022-03-08
Phase one of removing the HD prefix: use plain C99 names

gnuoyd:
Phase one of removing the HD prefix: replace most `HD`-prefixed C99 function names with plain C99 names and poison the `HD`-prefixed versions.


qkoziol:
How does this work with Windows?   Aren't there still many functions that require compatibility shims?

doc/de-HD-notes.md
```diff
@@ -0,0 +1,94 @@
...
+```C
+#define HDcos(x) cos(x)
+```
+
+A few wrapper macros clamped a parameter to an integer range or cast
+a parameter to a different type before passing it to a C99 function.
+In every instance, the casts and clamping was not well justified so it
+was removed.
+
+## memset
```
byrnHDF, line 49: Why not make this a separate change so that any warnings can be isolated.

src/H5ES.c
```diff
@@ -269,7 +269,7 @@ H5ESget_requests(hid_t es_id, H5_iter_order_t order, hid_t *connector_ids, void
     herr_t  ret_value = SUCCEED; /* Return value */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE5("e", "iIo*i**xx", es_id, order, connector_ids, requests, count);
+    H5TRACE6("e", "iIo*i**xzx", es_id, order, connector_ids, requests, array_len, count);
```
byrnHDF: This change is necessary because of a change from HD??

gnuoyd:
On Tue, Mar 08, 2022 at 07:54:02AM -0800, Allen Byrne wrote:
> @byrnHDF commented on this pull request.
> 
> 
> 
> > @@ -269,7 +269,7 @@ H5ESget_requests(hid_t es_id, H5_iter_order_t order, hid_t *connector_ids, void
>      herr_t  ret_value = SUCCEED; /* Return value */
>  
>      FUNC_ENTER_API(FAIL)
> -    H5TRACE5("e", "iIo*i**xx", es_id, order, connector_ids, requests, count);
> +    H5TRACE6("e", "iIo*i**xzx", es_id, order, connector_ids, requests, array_len, count);
> 
> This change is necessary because of a change from HD??

I believe that is something that `sh autogen.sh` did.

Dave

byrnHDF:
This would be easier to review if you could split this up - just assert change first, followed by malloc/free, then printf, next str* functions, etc.
I know one big PR is faster, but the review is more time consuming! Plus it would be easier to merge to other branches.

src/H5win32defs.h
```diff
@@ -52,10 +55,9 @@ struct timezone {
 #define HDgetdrive()         _getdrive()
 #define HDgetlogin()         Wgetlogin()
 #define HDgettimeofday(V, Z) Wgettimeofday(V, Z)
-#define HDisatty(F)          _isatty(F)
+#define isatty(F)            _isatty(F)
```
byrnHDF: why is this one (and some others)  okay without an HD and others keep HD?
I would prefer not mixing concepts - in one PR.

gnuoyd:
On Tue, Mar 08, 2022 at 08:39:27AM -0800, Allen Byrne wrote:
> -#define HDisatty(F)          _isatty(F)
> +#define isatty(F)            _isatty(F)
> 
> why is this one (and some others)  okay without an HD and others keep HD?
> I would prefer not mixing concepts - in one PR.

That should not have changed.  I'm not sure how that happened.

Dave

gnuoyd:
On Tue, Mar 08, 2022 at 08:20:59AM -0800, Allen Byrne wrote:
> This would be easier to review if you could split this up - just assert change first, followed by malloc/free, then printf, next str* functions, etc.
> I know one big PR is faster, but the review is more time consuming! Plus it would be easier to merge to other branches.

To replicate the change on other branches, use the script.  I used the
script to make 99.9% of the changes.

I can provide a script that reverses the automatic de-HD-ification, if
that will help you review what remains.

Dave

byrnHDF:
After going through the first half, if just assert got changed first, I think half of those files would vanish from subsequent PRs. :{)

src/H5ES.c
derobins: It's from a prior commit where someone didn't run bin/trace via autogen.sh

derobins:
Why most? Do you have a plan for bringing this work to completion? How carefully are you considering Windows? Are you looking at Visual Studio warnings to make sure you aren't making a mess?

Having a weird mix of some C calls using HD and others not is going to be incredibly annoying.

src/H5win32defs.h
```diff
@@ -68,26 +70,38 @@ struct timezone {
...
+#define HDrmdir(S)          _rmdir(S)
+#define HDsetenv(N, V, O)   Wsetenv(N, V, O)
+#define sleep(S)            Sleep(S * 1000)
+#define HDstat(S, B)        _stati64(S, B)
+#define HDstrdup(S)         _strdup(S)
+#define HDstrtok_r(X, Y, Z) strtok_s(X, Y, Z)
+#define HDtzset()           _tzset()
+#define HDunlink(S)         _unlink(S)
+#define HDunsetenv(N)       Wsetenv(N, "", 1)
+#define HDwrite(F, M, Z)    _write(F, M, Z)
```
derobins, line 84: What's your plan for dealing with this? Removing HD from the C calls is great and all, but leaving a handful of calls that require the HD prefix is worse than leaving them all in place.

derobins (changes requested):
I'm blocking this until there's a comprehensive solution in place that handles the Windows redefinitions. I'll agree that the HD prefixes are annoying and need to go, but I'm even less excited about having to remember which calls need HD and which don't and the culture of "someone else will deal with Windows" has to stop.

byrnHDF:
I agree that the final result is that a solution for all HD removal is needed. 
Under the assumption that there is a plan for that, we need to not overload reviewers unless there is no other choice. I was thinking that we could work through a little at a time where we could test the effect of each change across our supported platforms. This includes not making the warnings on windows worse.

derobins:
> I agree that the final result is that a solution for all HD removal is needed. Under the assumption that there is a plan for that, we need to not overload reviewers unless there is no other choice. I was thinking that we could work through a little at a time where we could test the effect of each change across our supported platforms. This includes not making the warnings on windows worse.

While this PR is huge to the point of being nigh unmanageable, I feel like we should do all the changes at once. Having an ever-changing set of C/POSIX API calls that don't need the HD prefix would be even worse than a static subset.

derobins:
Also, we keep racking up technical debt when people have "plans" for future Windows changes (i.e., kicking the can down the road). I don't want plans. I want code. Anyone who wants to make changes at the portability layer level needs to deal with Windows. Visual Studio and the Win32 API are the key reasons why we still even *have* the portability layer.

derobins:
Another thing to keep in mind is that any Windows Unicode work is probably going to require additional shims.

gnuoyd:
On Tue, Mar 08, 2022 at 08:39:27AM -0800, Allen Byrne wrote:
> @byrnHDF commented on this pull request.
> 
> 
> 
> > @@ -52,10 +55,9 @@ struct timezone {
>  #define HDgetdrive()         _getdrive()
>  #define HDgetlogin()         Wgetlogin()
>  #define HDgettimeofday(V, Z) Wgettimeofday(V, Z)
> -#define HDisatty(F)          _isatty(F)
> +#define isatty(F)            _isatty(F)
> 
> why is this one (and some others)  okay without an HD and others keep HD?
> I would prefer not mixing concepts - in one PR.

I accidentally picked up an HDisatty with the `<ctype.h>`
functions (HDisalpha et cetera).  I just fixed it.

Dave

gnuoyd:
On Tue, Mar 08, 2022 at 07:19:42AM -0800, Allen Byrne wrote:
> +## memset
> 
> Why not make this a separate change so that any warnings can be isolated.

I found four instances in the MSVC CI where this change introduced
warnings.  That is so few that I think we can deal with them
individually.  I left a comment above each line where MSVC warns: search
the source for "warning C4090".

I cannot find any other MSVC warnings that the de-HD-ification
introduced.

Dave

gnuoyd:
On Tue, Mar 08, 2022 at 07:03:24AM -0800, Quincey Koziol wrote:
> How does this work with Windows?  Aren't there still many functions
> that require compatibility shims?

There are no more than five C99 functions that need shims.

Dave

byrnHDF:
> On Tue, Mar 08, 2022 at 07:19:42AM -0800, Allen Byrne wrote: +## memset Why not make this a separate change so that any warnings can be isolated.
> I found four instances in the MSVC CI where this change introduced warnings. That is so few that I think we can deal with them individually. I left a comment above each line where MSVC warns: search the source for "warning C4090". I cannot find any other MSVC warnings that the de-HD-ification introduced. Dave

Okay

qkoziol:
> I'm blocking this until there's a comprehensive solution in place that handles the Windows redefinitions. I'll agree that the HD prefixes are annoying and need to go, but I'm even less excited about having to remember which calls need HD and which don't and the culture of "someone else will deal with Windows" has to stop.

I agree - the HD prefixes help protect the project from "all the world is Linux" (i.e. there are other flavors of Unix still out there), inconsistent compiler implementations of "standard" routines, enable portability to Windows, and allow for a point of flexibility with other OS ports in the future (cloud, embedded, etc.)

Having an inconsistent situation where there are 10-100 (I don't know the exact #) of special cases where an HD-prefixed routine is needed is worse than "always use the HD prefix for interfacing with system and language libraries".

byrnHDF:
> > I'm blocking this until there's a comprehensive solution in place that handles the Windows redefinitions. I'll agree that the HD prefixes are annoying and need to go, but I'm even less excited about having to remember which calls need HD and which don't and the culture of "someone else will deal with Windows" has to stop.
> 
> I agree - the HD prefixes help protect the project from "all the world is Linux" (i.e. there are other flavors of Unix still out there), inconsistent compiler implementations of "standard" routines, enable portability to Windows, and allow for a point of flexibility with other OS ports in the future (cloud, embedded, etc.)
> 
> Having an inconsistent situation where there are 10-100 (I don't know the exact #) of special cases where an HD-prefixed routine is needed is worse than "always use the HD prefix for interfacing with system and language libraries".

Maybe we need to pull the HD defines out of the different private headers and move into a single HD define header?

derobins:
This is fundamentally a good idea, but needs a better implementation. Also, the branch in develop needs to be moved to a personal fork. I'm closing it while further development takes place.

I would probably stage this as multiple PRs:

- Handle Windows the way other open-source projects do, removing the Windows issues.
- Consider using H5_<function> instead of HD<function> to generate a coherent OS layer
- Strip HD from C library calls

PR #1480, opened 2022-03-10
Fix JNI bug, fix warnings noticed in MSVC CI, be modern & portable

gnuoyd:
See individual commit messages for the details.

### DIFF
--- a/java/src/jni/h5util.c
+++ b/java/src/jni/h5util.c
@@ -1160,11 +1160,11 @@ h5str_sprintf(JNIEnv *env, h5str_t *out_str, hid_t container, hid_t tid, void *i
 
                                         H5Otoken_to_str(tid, &oi.token, &token_str);
 
-                                        size_t this_len = 64 + strlen(token_str) + 1;
-                                        if (NULL == (this_str = (char *)HDmalloc(this_len)))
+                                        size_t that_len = 64 + strlen(token_str) + 1;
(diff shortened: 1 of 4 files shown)


derobins (approved):

src/H5system.c
```diff
@@ -1033,7 +1033,7 @@ H5_get_option(int argc, const char *const *argv, const char *opts, const struct
         HDfree(arg);
     }
     else {
-        register char *cp; /* pointer into current token */
```
derobins, line 1036: You might also want to yank the register keyword from H5VMprivate.h and the tools. We shouldn't be encouraging that level of micro-"optimization".

java/src/jni/h5util.c
```diff
@@ -813,7 +813,7 @@ h5str_sprintf(JNIEnv *env, h5str_t *out_str, hid_t container, hid_t tid, void *i
                     if (NULL == (this_str = (char *)HDmalloc(this_len)))
                         H5_OUT_OF_MEMORY_ERROR(ENVONLY, "h5str_sprintf: failed to allocate string buffer");
 
-                    if (HDsnprintf(this_str, "%g", this_len, tmp_double) < 0)
+                    if (HDsnprintf(this_str, this_len, "%g", tmp_double) < 0)
```
byrnHDF, line 816: This code likely started out in the tools lib, might want to dbl-chk the str_sprintf function in there.

jhendersonHDF (approved):

PR #1497, opened 2022-03-11
Removes the MANIFEST file and unused release files

derobins:


bin/bbrelease
```diff
@@ -330,64 +278,62 @@ for f in README.md release_docs/RELEASE.txt; do
...
     DOC_URL=https://git@bitbucket.hdfgroup.org/scm/hdffv/hdf5doc.git -b ${DOCVERSION}
 fi
 
 # Create the tar file
 test "$verbose" && echo "   Running tar..." 1>&2
 ( \
-    cd $tmpdir; \
-    tar cf $HDF5_VERS.tar $HDF5_IN_VERS/Makefile \
-	`sed 's/^\.\//'$HDF5_IN_VERS'\//' $MANIFEST` || exit 1 \
+    tar --transform "s,^\.,$HDF5_IN_VERS," -c --exclude=".git" -f "$tmpdir/$HDF5_VERS.tar" Makefile . || exit 1 \
```
derobins: I believe --transform is a gnu extension to tar. Not sure how that will fare on the BSDs or MacOS.

qkoziol: "man tar" on MacOS doesn't list a --transform option.

qkoziol:
I think the 'tar' command needs to be addressed, but otherwise is a good cleanup.

qkoziol (approved):

derobins:
> I think the 'tar' command needs to be addressed, but otherwise is a good cleanup.

I'll try it on MacOS over the weekend and I'll look at the man pages for the BSDs.

soumagne (approved):

byrnHDF (approved):

jhendersonHDF (commented):
Approved, outside of also agreeing that we need to find a solution for the tar issue here

qkoziol (approved):

bin/bbrelease
derobins: Fixed. Going through the symlink using -h should be portable.

lrknox (approved):

PR #1499, opened 2022-03-15
Misc fixes highlighted by the checkposix script

derobins:
When Dave's C99/POSIX work is done we can rip out the checkposix script and be done with this stuff. Until then, these fixes make the code uniform.

### DIFF
--- a/src/H5FDdevelop.h
+++ b/src/H5FDdevelop.h
@@ -223,6 +223,9 @@ struct H5FD_t {
     hbool_t paged_aggr; /* Paged aggregation for file space is enabled or not */
 };
 
+/* VFD initialization function */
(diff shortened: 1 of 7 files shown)


src/H5FDdevelop.h
```diff
@@ -223,6 +223,9 @@ struct H5FD_t {
     hbool_t paged_aggr; /* Paged aggregation for file space is enabled or not */
 };
 
+/* VFD initialization function */
+typedef hid_t (*H5FD_init_t)(void);
```
derobins, line 227: The library always typedefs a prototype for function parameters.

jhendersonHDF (approved):

lrknox (approved):

PR #1500, opened 2022-03-15
New API for memory management via VFDs

lucasvr:
With the introduction of VFD plugins, it is now possible to have virtual
file drivers that specialize on data transfer between storage and
hardware accelerators. The GDS (GPUDirect Storage) VFD is one such
example that enables I/O between NVMe storage and GPU memory.

GDS allocates device memory using vendor-specific APIs that are not meant
to be exposed elsewhere in the HDF5 core library. Yet, other components
such as I/O filters would benefit from having access to the memory
allocator used by the VFD, as that could save data copies across device
and host memory, for instance.

This commit introduces two high-level memory management interfaces that
enable memory allocation/deallocation through the VFD driver (falling
back to the existing `H5MM_malloc/xfree` functions when the VFD does not
provide a custom implementation). The functions are implemented on top
of the H5FD ctl interface and are described in details below.

`H5FDalloc_mem()`: allocate memory via the file driver. If the driver does
not support the `H5FD_CTL__MEM_ALLOC` opcode then the operation falls back
to allocation via system memory (i.e., `H5MM_malloc()`). The input
arguments to that CTL operation are given by the `H5FD_ctl_alloc_args_t`
structure:

```
typedef struct H5FD_ctl_alloc_args_t {
    hsize_t       size;      /**< Memory allocation size */
    hsize_t       alignment; /**< Desired memory alignment */
    unsigned long flags;     /**< Request flags (H5FD_ALLOC*) */
    const void   *args;      /**< Optional arguments wanted by the VFD
                                  given the request flags */
} H5FD_ctl_alloc_args_t;
```

The `flags` and `args` members have been introduced to support fancy
memory allocators such as `cudaMallocAsync()` (which takes a stream ID
as argument), and Xilinx' platform extension to OpenCL which enables
`clCreateBuffer()` to take a target memory bank ID argument.

`H5FDfree_mem()`: release memory previously allocated via `H5FDalloc_mem()`.
If the driver does not support the `H5FD_CTL__MEM_FREE` opcode, memory is
freed via `H5MM_xfree()`. Input arguments are given by the
`H5FD_ctl_free_args_t` structure:

```
typedef struct H5FD_ctl_free_args_t {
    void         *buf;   /**< Pointer to memory space to be freed */
    unsigned long flags; /**< Request flags (H5FD_MEM*) */
    const void   *args;  /**< Optional arguments wanted by the VFD
                              given the request flags */
} H5FD_ctl_free_args_t;
```

The same observations regarding `flags` and `args` apply here.

An application can query the supported flags by calling `H5FDctl` with
the new `H5FD_CTL__MEM_QUERY_SUPPORTED_FLAGS` opcode:

```
unsigned long supported_flags = 0;
unsigned long *flag_ptr = &supported_flags;
opcode = H5FD_CTL__MEM_QUERY_SUPPORTED_FLAGS;
H5FDctl(file_drv_ptr, opcode, 0, NULL, (void **) &flag_ptr);
```

This commit introduces a single `H5FD_MEM_ASYNC` flag. Other flags may be
added in the future depending on use-cases and demand.

Signed-off-by: Lucas C. Villa Real <lucasvr@br.ibm.com>

### DIFF
--- a/src/H5FD.c
+++ b/src/H5FD.c
@@ -24,6 +24,7 @@
 /****************/
 
 #define H5F_FRIEND      /* Suppress error about including H5Fpkg */
+#define H5E_FRIEND      /* Suppress error about including H5Epkg */
 #include "H5FDmodule.h" /* This source code file is part of the H5FD module */
 
 /***********/
@@ -33,6 +34,7 @@
 #include "H5CXprivate.h" /* API Contexts                             */
 #include "H5Dprivate.h"  /* Datasets                                 */
 #include "H5Eprivate.h"  /* Error handling                           */
+#include "H5Epkg.h"      /* Error macros                             */
 #include "H5Fpkg.h"      /* File access                              */
 #include "H5FDpkg.h"     /* File Drivers                             */
 #include "H5Iprivate.h"  /* IDs                                      */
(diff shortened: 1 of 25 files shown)


src/H5FD.c
```diff
@@ -24,6 +24,7 @@
 /****************/
 
 #define H5F_FRIEND      /* Suppress error about including H5Fpkg */
+#define H5E_FRIEND      /* Suppress error about including H5Epkg */
```
derobins: This is probably not a good idea. Private headers are for code that can be used outside of the package (H5E, H5FD, etc.). Package headers are for code that is only used within the package. If there is functionality that is needed outside of the package, we almost always move it to the private header. Using H5X_FRIEND is usually a band-aid and we've been working to minimize that.

derobins:
As with your other pull request, we should probably have a separate conversation about what you are trying to do. IIRC, there was some internal discussion about similar functionality in the recent past. We'll want to make sure that whatever memory allocation functionality we add meets everyone's needs.

lucasvr:
> As with your other pull request, we should probably have a separate conversation about what you are trying to do. IIRC, there was some internal discussion about similar functionality in the recent past. We'll want to make sure that whatever memory allocation functionality we add meets everyone's needs.

We're having a discussion about this topic on the HDF Group Forum. Could you join the conversation when you have a chance? This is the link to the RFC thread: https://forum.hdfgroup.org/t/memory-management-in-h5fd-class-t/9452

Thanks!

src/H5FD.c
```diff
@@ -1142,6 +1144,146 @@ H5FDfree(H5FD_t *file, H5FD_mem_t type, hid_t dxpl_id, haddr_t addr, hsize_t siz
...
+ *
+ * Purpose:     Checks if the most recent frame of the error stack have a
+ *              major and minor numbers that match the provided arguments.
+ *
+ * Return:      TRUE if arguments match the most recent error, FALSE otherwise
+ *
+ *----------------------------------------------------------------------------
+ */
+static hbool_t
+H5FD__error_matches(hid_t major, hid_t minor)
```
jhendersonHDF: Agreed with @derobins on the package vs. private separation. This kind of a method is likely something that can (and should) be part of the H5E private code and then we wouldn't need to bring in H5Epkg.h. There should probably also be a bit more sanity checking here, like making sure the error stack actually has any messages on it before accessing `slot[estack->nused-1]`.

src/H5FD.c
```diff
@@ -1142,6 +1144,146 @@ H5FDfree(H5FD_t *file, H5FD_mem_t type, hid_t dxpl_id, haddr_t addr, hsize_t siz
...
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "file class pointer cannot be NULL")
+    if (!args)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "args pointer cannot be NULL")
+    if (!out)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "pointer to output buffer cannot be NULL")
+
+    if (file->cls->ctl) {
+        /* Delegate memory allocation to the VFD driver */
+        uint64_t flags = H5FD_CTL__FAIL_IF_UNKNOWN_FLAG;
+        if ((file->cls->ctl)(file, H5FD_CTL__MEM_ALLOC, flags, args, out) < 0) {
```
jhendersonHDF: Rather than directly calling the ctl callback here, this should probably route through H5FD_ctl. Though, you probably will still need to check whether the driver has a ctl callback or not here so that you can properly allocate system memory in the case that the driver doesn't support this. Also, have you considered switching H5E_FCNTL to H5E_UNSUPPORTED from our discussion? That seems safer than checking H5E_FCNTL here.

test/vfd.c
```diff
@@ -3914,6 +3914,109 @@ test_ctl(void)
...
+        TEST_ERROR;
+
+    if (NULL == (file_drv_ptr = H5FDopen(filename, H5F_ACC_RDWR, fapl_id, HADDR_UNDEF)))
+        PUTS_ERROR("couldn't get pointer to H5FD_t structure");
+
+    /* Query supported alloc flags */
+    TESTING_2("mem query supported flags")
+    flag_ptr = &supported_flags;
+    ctl_flag = 0 == strcmp(driver_name, "sec2") ? 0 : H5FD_CTL__FAIL_IF_UNKNOWN_FLAG;
+    if (H5FDctl(file_drv_ptr, H5FD_CTL__MEM_QUERY_SUPPORTED_FLAGS, ctl_flag, NULL, (void **) &flag_ptr) < 0)
```
jhendersonHDF: Since we run these tests with all available VFDs, this is likely going to fail on everything except for the sec2 driver. For VFD test code, it's probably safest to never use the "fail if unknown" flag, unless it's something that's truly expected to be implemented by every single VFD.

jhendersonHDF:
A good idea here and I think this is a good start! A bit to cleanup here, but I agree that having memory allocation and freeing calls that can work regardless of the VFD being used is helpful. I still think it might be worth investigating if we can repurpose H5(allocate/resize/free)_memory so that applications don't have to mix and match between that and H5FDalloc_mem, but I'm not sure what compatibility issues might arise there.

One more suggestion might be to also have a realloc memory operation. I think the overhead of needing to allocate new memory, copy over the old memory and then free the old memory might be burdensome as compared to just having a realloc operation.

src/H5FD.c
lucasvr: Thanks for your comments, @jhendersonHDF. Yes, I did consider switching to H5E_UNSUPPORTED but ended up not bringing that one to this PR. I'll bring it to the next one.

test/vfd.c
lucasvr: > A good idea here and I think this is a good start! A bit to cleanup here, but I agree that having memory allocation and freeing calls that can work regardless of the VFD being used is helpful. I still think it might be worth investigating if we can repurpose H5(allocate/resize/free)_memory so that applications don't have to mix and match between that and H5FDalloc_mem, but I'm not sure what compatibility issues might arise there.

I'll take a look at the existing H5*_memory functions and will come back to you early next week with some more thoughts / updated PR.

> One more suggestion might be to also have a realloc memory operation. I think the overhead of needing to allocate new memory, copy over the old memory and then free the old memory might be burdensome as compared to just having a realloc operation.

Yes, a realloc is certainly a welcome operation. I also miss a "direction" flag for `H5FD_CTL__MEM_COPY` (i.e., host->host, host->device, device->host, device->device). I will submit that one as a separate PR early next week.

lucasvr:
@derobins @jhendersonHDF here's a revamped version of the patchset.

Note that I've found a few third-party projects on GitHub that rely on the existing `H5(allocate,resize,free)_memory` functions. For this reason I've created a new variant of such functions (with a suffix `2`). We can discuss what's best to do once you're happy with the code changes.

I'm especially interested to hear your opinion on my approach to obtain the file handle from filter callbacks. As you'll see, commit 1af8f2a3a955dd79042b464325c900184426c70d introduces a new API `hid_t H5FD_get_pipeline_file_id()` that filter callbacks can use to get that handle. Ultimately, that handle can be provided to the `H5(allocate,resize,free)_memory2` functions. The advantage of this approach is that the public I/O filter API remains the same. The disadvantage is that we're introducing _one more_ global-scoped variable in H5Z.c.

Thanks in advance for your feedback!

derobins:
I'm going to go ahead and close this PR, but I've created a discussion for this (#3082)

PR #1502, opened 2022-03-15
New H5F API to obtain a handle to the low-level file driver

lucasvr:
To date, the only way to obtain a file driver handle (`H5FD_t *`) is to
use `H5FDopen()`. This commit adds a new function that behaves like
`H5Fget_vfd_handle()`: given a valid file_id, it retrieves a handle that
can be used on calls to functions of the H5FD family.

The new API is especially useful for applications wishing to invoke CTL
operations on the VFD driver.

Signed-off-by: Lucas C. Villa Real <lucasvr@br.ibm.com>

### DIFF
--- a/src/H5F.c
+++ b/src/H5F.c
@@ -463,6 +463,49 @@ done:
 } /* end H5Fget_vfd_handle() */
 
 /*-------------------------------------------------------------------------
+ * Function:    H5Fget_file_driver_handle
+ *
+ * Purpose:     Returns a pointer to the handle of the low-level file
+ *              driver.
+ *
+ * Return:      Success:    Non-negative
+ *              Failure:    Negative
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Fget_file_driver_handle(hid_t file_id, void **file_handle /*out*/)
+{
+    H5F_t *                          obj;                 /* File driver handle */
+    H5VL_object_t *                  vol_obj;             /* File info */
+    H5VL_optional_args_t             vol_cb_args;         /* Arguments to VOL callback */
+    H5VL_native_file_optional_args_t file_opt_args;       /* Arguments for optional operation */
+    herr_t                           ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_API(FAIL)
+    H5TRACE2("e", "ix", file_id, file_handle);
+
+    /* Check args */
+    if (!file_handle)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "invalid file handle pointer")
+
+    /* Get the file object */
+    if (NULL == (vol_obj = (H5VL_object_t *)H5I_object(file_id)))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "invalid file identifier")
+
+    if (NULL == (obj = vol_obj->data))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "invalid file driver handle")
+
+    if (NULL == (obj->shared))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "invalid file driver shared object data")
+
+    /* Set output file handle */
+    *file_handle = obj->shared->lf;
+
+done:
+    FUNC_LEAVE_API(ret_value)
+} /* end H5Fget_vfd_handle() */
+
+/*-------------------------------------------------------------------------
  * Function:    H5Fis_accessible
  *
  * Purpose:     Check if the file can be opened with the given fapl.
--- a/src/H5Fpublic.h
+++ b/src/H5Fpublic.h
@@ -798,6 +798,9 @@ H5_DLL ssize_t H5Fget_obj_ids(hid_t file_id, unsigned types, size_t max_objs, hi
  *
  */
 H5_DLL herr_t H5Fget_vfd_handle(hid_t file_id, hid_t fapl, void **file_handle);
+
+H5_DLL herr_t H5Fget_file_driver_handle(hid_t file_id, void **file_handle);
+
 /**
  * \ingroup H5F
  *


derobins (changes requested):
This can't be invoked like this. You aren't going through the VOL via an 'optional' command so it only works for the native VOL connector.

There are also no tests. You'll need to add something to test/vfd.c to exercise this functionality.

What is the use case for this, btw?

src/H5F.c
```diff
@@ -462,6 +462,49 @@ H5Fget_vfd_handle(hid_t file_id, hid_t fapl_id, void **file_handle /*out*/)
...
+ *              Failure:    Negative
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Fget_file_driver_handle(hid_t file_id, void **file_handle /*out*/)
+{
+    H5F_t *                          obj;                 /* File driver handle */
+    H5VL_object_t *                  vol_obj;             /* File info */
+    H5VL_optional_args_t             vol_cb_args;         /* Arguments to VOL callback */
+    H5VL_native_file_optional_args_t file_opt_args;       /* Arguments for optional operation */
```
derobins, line 481: You need to look at the other 'file optional' calls to see how to use these and wire up this call through the VOL.

src/H5F.c
```diff
@@ -462,6 +462,49 @@ H5Fget_vfd_handle(hid_t file_id, hid_t fapl_id, void **file_handle /*out*/)
...
+ *
+ * Purpose:     Returns a pointer to the handle of the low-level file
+ *              driver.
+ *
+ * Return:      Success:    Non-negative
+ *              Failure:    Negative
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Fget_file_driver_handle(hid_t file_id, void **file_handle /*out*/)
```
derobins, line 476: Since this is native-VOL-only, the name should reflect that. Also, if you know you are returning a pointer to an H5FD_t, return that instead of a void pointer.

src/H5Fpublic.h
```diff
@@ -798,6 +798,9 @@ H5_DLL ssize_t H5Fget_obj_ids(hid_t file_id, unsigned types, size_t max_objs, hi
  *
  */
 H5_DLL herr_t H5Fget_vfd_handle(hid_t file_id, hid_t fapl, void **file_handle);
+
+H5_DLL herr_t H5Fget_file_driver_handle(hid_t file_id, void **file_handle);
+
```
derobins, line 803: If this is not really an API call for the general public, it should probably go in a new H5Fdevelop.h header file.

src/H5F.c
lucasvr: Hi Dana! Thanks for your review. This shouldn't be tied to the native VOL -- the `H5VL_optional_args_t` and `H5VL_native_file_optional_args_t` are dead variables (sorry about that! I'll remove them in my next PR). Do you see another assumption on the implementation of that function that's only guaranteed to work with the native VOL driver?

src/H5F.c
lucasvr: Since I'm simply returning the private file handle previously stored by the VFD under `obj->shared->lf`, would we have any advantage on implementing a "file optional" handler at the VFD driver to get that same data? Do non-native VFDs behave differently, somehow?

src/H5Fpublic.h
lucasvr: My intention with this API is to create the foundation for a next-generation I/O filter interface: filters would be able to retrieve the VFD handle via this new API and perform memory allocation via the VFD driver (see PR #1500), using functions from the H5FD* family.

Would you have any advice in addition to your comments above to make this API suitable for the general public? Thanks!

lucasvr:
> This can't be invoked like this. You aren't going through the VOL via an 'optional' command so it only works for the native VOL connector.

Ahh, got it. Thanks for your guidance. I'll work on this one and will make sure it works with non-native VOLs before asking for a new code review.

> There are also no tests. You'll need to add something to test/vfd.c to exercise this functionality.

Will do.

> What is the use case for this, btw?

I want to enable applications that have a valid file handle to invoke VFD CTL operations (so they can ask the VFD to allocate/deallocate device memory, for instance). Such operations require a valid VFD handle. To date, applications would need to open the same file twice if they need both a VFD handle and a regular file handle.

src/H5F.c
derobins: You are confusing the Virtual File Layer (VFL) / Virtual File Drivers (VFDs) with the Virtual Object Layer (VOL). The VOL maps HDF5 API calls to arbitrary storage. The VFL (and VFDs) are a part of the native VOL connector, which implements storing data in the HDF5 files we all know and love.

The obj->shared->lf pointer you are attempting to return only exists for the native VOL connector. Other connectors will return their own data. You're going to need to look at H5VLnative_file.c and the other native-specific API calls (e.g., H5Fget_vfd_handle) to see how to properly handle this.

src/H5F.c
derobins: See my other comment. What you are doing is 100% tied to the native VOL connector.

src/H5Fpublic.h
derobins: Instead of accepting your PRs, we should probably have a separate conversation about what you are trying to do. Shoot me an email at derobins@hdfgroup.org and I'll see if we can set something up.

lucasvr: Sounds good to me. I'll close this PR for now. Thanks!

jhendersonHDF:
I see that this PR is closed now, but I'll leave a comment that may help in reworking this. I ran into the same issue when implementing VFD plugin support, so I created a new private routine `H5F_shared_get_file_driver` in H5Fquery.c that does exactly this (return the `lf` field from a file object pointer). A public counterpart could probably use this rather than trying to directly return the `lf` field, but I think this may not be quite the right approach.

Returning an `H5FD_t *` publicly is probably a _bad idea_. Instead, a public routine to make a ctl call should accept an `hid_t` for the file ID and let the internals do the translating into an `H5FD_t *` work. `H5FDctl` needed to accept an `H5FD_t *` because that was the cleanest way for the ctl call to be passed down through a stack of drivers, but this makes it fundamentally not an application-level routine; it is more of a driver-internal routine, wherein `H5FD_t *` is the base working type for a driver. I think rather than trying to coerce `H5FD_t *` out of the library, we should consider that many of the routines in H5FD.c are meant to be driver-only and instead focus on how else the application-level interface to these features might look. This relates to why I was hoping we might be able to repurpose `H5(allocate/resize/free)_memory` without causing too much pain. Unfortunately, those don't accept a file ID like we'd need, but a change like that could be a good candidate for the next major release of HDF5.

lucasvr:
> I see that this PR is closed now, but I'll leave a comment that may help in reworking this. I ran into the same issue when implementing VFD plugin support, so I created a new private routine `H5F_shared_get_file_driver` in H5Fquery.c that does exactly this (return the `lf` field from a file object pointer). A public counterpart could probably use this rather than trying to directly return the `lf` field, but I think this may not be quite the right approach.

Ah, I wasn't aware of that function. I will definitely look into reusing it somehow.

> Returning an H5FD_t * publicly is probably a bad idea. Instead, a public routine to make a ctl call should accept an hid_t for the file ID and let the internals do the translating into an H5FD_t * work. H5FDctl needed to accept an H5FD_t * because that was the cleanest way for the ctl call to be passed down through a stack of drivers, but this makes it fundamentally not an application-level routine; it is more of a driver-internal routine, wherein H5FD_t * is the base working type for a driver. I think rather than trying to coerce H5FD_t * out of the library, we should consider that many of the routines in H5FD.c are meant to be driver-only and instead focus on how else the application-level interface to these features might look. 

OK, that makes sense. Thanks for the explanation.

> This relates to why I was hoping we might be able to repurpose H5(allocate/resize/free)_memory without causing too much pain. Unfortunately, those don't accept a file ID like we'd need, but a change like that could be a good candidate for the next major release of HDF5.

Sounds good. Thank you once again!

lucasvr:
> This relates to why I was hoping we might be able to repurpose `H5(allocate/resize/free)_memory` without causing too much pain. Unfortunately, those don't accept a file ID like we'd need, but a change like that could be a good candidate for the next major release of HDF5.

(reopening so that other people can have access to the ongoing conversation)

@jhendersonHDF, I/O filters don't have access to the file ID. Two possible approaches I can think of are:

1. to modify the filter API so that missing handlers are provided filter callback
2. to use thread-local storage to store the file ID information + introduce new API to let filters retrieve it

I'd be more inclined towards the second approach, but HDF5 thread-safe builds don't mix well with other build options (e.g., Fortran, Parallel I/O, C++, high-level lib). Do you have any comments about this?

lucasvr:
Here's a proposal.

I'm introducing a public API named `H5Zget_pipeline_file_id()`, to be called by I/O filters that want to manage memory using the repurposed `H5(allocate/resize/free)_memory` functions. This new API captures the file handle associated with the filter pipeline (`H5O_pline_t *`) being processed.

Filter implementations can then invoke e.g., `H5allocate_memory(hid_t file_handle, size_t size, hbool_t clear)` to ask the VFD driver (or the generic H5MM family of functions) to provide them with the requested memory area.

Please let me know if this sounds like a reasonable approach so I can introduce proper safety checks, tests, and the corresponding documentation.

Thanks,
Lucas

```
--- a/src/H5Z.c
+++ b/src/H5Z.c
@@ -60,6 +60,7 @@ typedef enum {
 static size_t        H5Z_table_alloc_g = 0;
 static size_t        H5Z_table_used_g  = 0;
 static H5Z_class2_t *H5Z_table_g       = NULL;
+static H5O_pline_t  *H5Z_curr_pline_g  = NULL;
 #ifdef H5Z_DEBUG
 static H5Z_stats_t *H5Z_stat_table_g = NULL;
 #endif /* H5Z_DEBUG */
@@ -714,6 +715,8 @@ H5Z__prelude_callback(const H5O_pline_t *pline, hid_t dcpl_id, hid_t type_id, hi
 
     HDassert(pline->nused > 0);
 
+    H5Z_curr_pline_g = pline;
+
     /* Iterate over filters */
     for (u = 0; u < pline->nused; u++) {
         /* Get filter information */
@@ -767,6 +770,8 @@ H5Z__prelude_callback(const H5O_pline_t *pline, hid_t dcpl_id, hid_t type_id, hi
 
 done:
 
+    H5Z_curr_pline_g = NULL;
+
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5Z__prelude_callback() */
 
@@ -1255,6 +1260,20 @@ done:
     FUNC_LEAVE_NOAPI(ret_value)
 } /* H5Z_find() */
 
+hid_t
+H5Zget_pipeline_file_id()
+{
+    hid_t file_id = -1;
+
+    if (NULL == H5Z_curr_pline_g)
+        return -1;
+
+    if (H5Z_curr_pline_g->sh_loc.file)
+        file_id = H5F_get_id(H5Z_curr_pline_g->sh_loc.file);
+
+    return file_id;
+}
+
  * Function: H5Z_pipeline
  *
@@ -1306,6 +1325,8 @@ H5Z_pipeline(const H5O_pline_t *pline, unsigned flags, unsigned *filter_mask /*i
     HDassert(buf && *buf);
     HDassert(!pline || pline->nused < H5Z_MAX_NFILTERS);
 
+    H5Z_curr_pline_g = pline;
+
 #ifdef H5Z_DEBUG
     H5_timer_init(&timer);
 #endif
@@ -1449,6 +1470,7 @@ H5Z_pipeline(const H5O_pline_t *pline, unsigned flags, unsigned *filter_mask /*i
     *filter_mask = failed;
 
 done:
+    H5Z_curr_pline_g = NULL;
     FUNC_LEAVE_NOAPI(ret_value)
 }
 
```

PR #1508, opened 2022-03-18
Re-promote some GNU warnings to errors

gnuoyd:


src/H5Dchunk.c
```diff
@@ -7711,21 +7711,16 @@ H5D__chunk_iter_cb(const H5D_chunk_rec_t *chunk_rec, void *udata)
...
 H5D__chunk_iter(const H5D_t *dset, H5D_chunk_iter_op_t op, void *op_data)
 {
-    const H5O_layout_t *layout = NULL;       /* Dataset layout */
-    const H5D_rdcc_t *  rdcc   = NULL;       /* Raw data chunk cache */
-    H5D_rdcc_ent_t *    ent;                 /* Cache entry index */
-    H5D_chk_idx_info_t  idx_info;            /* Chunked index info */
-    herr_t              ret_value = SUCCEED; /* Return value */
+    const H5D_shared_t *const shared = dset->shared;
+    const H5O_layout_t *const layout = &shared->layout;
+    const H5D_rdcc_t *const   rdcc   = &dset->shared->cache.chunk;
```
byrnHDF: Can &dset->shared be just &shared?

gnuoyd: Sure, I will make the change.

src/H5Dchunk.c
```diff
@@ -7711,21 +7711,16 @@ H5D__chunk_iter_cb(const H5D_chunk_rec_t *chunk_rec, void *udata)
...
+    const H5D_rdcc_t *const   rdcc   = &shared->cache.chunk;
+    H5D_rdcc_ent_t *          ent;                 /* Cache entry index */
+    H5D_chk_idx_info_t        idx_info;            /* Chunked index info */
+    herr_t                    ret_value = SUCCEED; /* Return value */
 
     FUNC_ENTER_PACKAGE_TAG(dset->oloc.addr)
 
-    /* Check args */
-    HDassert(dset);
-    HDassert(dset->shared);
```
fortnern: It's better to have an assertion here than to rely on a segfault in the initialization in case dset is NULL. Please reinstate these.

src/H5FDsec2.c
```diff
@@ -1073,21 +1073,15 @@ static herr_t
...
+    HDassert(_file);
 
-    switch (op_code) {
-        /* Unknown op code */
-        default:
-            if (flags & H5FD_CTL__FAIL_IF_UNKNOWN_FLAG)
-                HGOTO_ERROR(H5E_VFL, H5E_FCNTL, FAIL, "unknown op_code and fail if unknown flag is set")
-            break;
-    }
+    if (flags & H5FD_CTL__FAIL_IF_UNKNOWN_FLAG)
```
fortnern, line 1089: Please add a comment here that this file driver doesn't recognize any op codes.

src/H5Dchunk.c
derobins: This is fundamentally introducing a new paradigm of parameter handling. Instead of checking parameters via asserts before assigning to things like dset->shared, the asserts get tossed in favor of direct assignments at the top.

The problem with this is that now there are two paradigms for input parameter handling used in the library and it's not clear which would be the "right" one.

derobins (changes requested):
The new paradigm of parameter handling needs to come out

src/H5Dchunk.c
byrnHDF: Can the asserts be moved to the top of the function before the assignments?

derobins:
There's been no movement on fixing the issues and it's out of sync now, so I'm closing this.

src/H5Dchunk.c
gnuoyd: I have taken the change out because it is not integral to the PR and I really want the PR merged, but there is no reasonable defense for these assertions.

src/H5FDsec2.c
gnuoyd: Added.

qkoziol (approved):

derobins:
I feel like we should be working on the software via a 'developer mode' that turns on Werror as much as we can instead of inflicting Werror on users of the software. I'm not opposed to merging this in, but I feel like Jordan's changes are the way to go.

jhendersonHDF:
> I feel like we should be working on the software via a 'developer mode' that turns on Werror as much as we can instead of inflicting Werror on users of the software. I'm not opposed to merging this in, but I feel like Jordan's changes are the way to go.

Absolutely. A few werrors are maybe ok to turn on, like say -Werror=cast-align, but it is a terrible idea to turn on things like -Werror=maybe-uninitialized in the general case. Not opposed to merging this either, but the vast majority of these -Werror warnings need to go to a developer build mode and NEVER be turned on for user build modes (even debug). -Werror will constantly be a thorn across the various combinations of platforms, compilers, compiler versions and more.

config/gnu-warnings/error-general
```diff
@@ -3,18 +3,22 @@
...
 -Werror=old-style-definition
 -Werror=packed
 -Werror=pointer-sign
 -Werror=pointer-to-int-cast
--Werror=int-to-pointer-cast
 -Werror=redundant-decls
 -Werror=strict-prototypes
 -Werror=switch
+-Werror=unused-but-set-variable
+-Werror=unused-variable
```
jhendersonHDF, line 21: For example, this is a silly -Werror to turn on in the general case. Failing to build because of an unused variable is ridiculous. Perfectly ok in a developer build mode because it encourages us to have cleaner code, but even then developers may end up needing to disable it due to an external dependency being built.

gnuoyd:
On Fri, Apr 22, 2022 at 06:27:17AM -0700, jhendersonHDF wrote:
> > I feel like we should be working on the software via a 'developer mode' that turns on Werror as much as we can instead of inflicting Werror on users of the software. I'm not opposed to merging this in, but I feel like Jordan's changes are the way to go.
> 
> Absolutely. A few werrors are maybe ok to turn on, like say -Werror=cast-align, but it is a terrible idea to turn on things like -Werror=maybe-uninitialized in the general case. Not opposed to merging this either, but the vast majority of these -Werror warnings need to go to a developer build mode and NEVER be turned on for user build modes (even debug). -Werror will constantly be a thorn across the various combinations of platforms, compilers, compiler versions and more.

There are multiple layers of misunderstanding here.

Dave

jhendersonHDF:
> On Fri, Apr 22, 2022 at 06:27:17AM -0700, jhendersonHDF wrote: > I feel like we should be working on the software via a 'developer mode' that turns on Werror as much as we can instead of inflicting Werror on users of the software. I'm not opposed to merging this in, but I feel like Jordan's changes are the way to go. Absolutely. A few werrors are maybe ok to turn on, like say -Werror=cast-align, but it is a terrible idea to turn on things like -Werror=maybe-uninitialized in the general case. Not opposed to merging this either, but the vast majority of these -Werror warnings need to go to a developer build mode and NEVER be turned on for user build modes (even debug). -Werror will constantly be a thorn across the various combinations of platforms, compilers, compiler versions and more.
> There are multiple layers of misunderstanding here. Dave

In what way? I understand that these errors aren't _currently_ being turned on in the general case, unless the HDF5_ENABLE_WARNINGS_AS_ERRORS option is enabled (and only because of the added string replacement stuff); I'm referring to that configuration as "the general case". The point is that things like `maybe-uninitialized` can likely start cropping up as compilers get better at detecting these kinds of things in complex edge cases. So, "the general case" breaks differently between different compiler versions. I'm just attempting to give some food for thought since -Werror has annoyed me to no end on several projects.

Even if the user has asked for _some_ warnings as errors (as the comment near the CMake option implies), it seems odd to me to include `unused-variable` and `unused-but-set-variable` in that list. Doesn't seem to me to be the kind of thing that someone would be looking for, but I'm happy to be proven wrong.

Maybe we need to more clearly define what this CMake option is supposed to include; is it everything and the kitchen sink, something in between or nothing? Everything and the kitchen sink makes sense if someone is trying to build a warning-free HDF5, though obviously has all the inherent problems with -Werror. Something in between is where we currently are and I think it makes more sense of all of that to just fall under a specific build mode. Nothing implies (at least IMO) that the option could go away and a user could eventually just use the Developer build mode if they care about warnings as errors.

I believe we're both heading in the same direction, but to me "everything and the kitchen sink" is what the HDF5_ENABLE_WARNINGS_AS_ERRORS option should represent and all of the finer-grained -Werror subset flags should go under a developer build mode. Happy to hear other sides of the argument..

derobins (approved):

PR #1534, opened 2022-03-28
Removes cruft from a few header files

derobins:
* Unnecessary extern C guards in cache headers
* Non-existent H5Xpublic.h includes hidden behind NOT_YET symbols

### DIFF
--- a/src/H5ACpublic.h
+++ b/src/H5ACpublic.h
@@ -28,10 +28,6 @@
 #include "H5public.h"
 #include "H5Cpublic.h"
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 /****************************************************************************
  *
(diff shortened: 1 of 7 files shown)


byrnHDF (approved):

src/H5ACpublic.h
```diff
@@ -783,7 +779,4 @@ typedef struct H5AC_cache_image_config_t {
 
 //! <!-- [H5AC_cache_image_config_t_snip] -->
 
-#ifdef __cplusplus
-}
-#endif
```
derobins, line 788: extern "C" is only needed to prevent C++ name mangling in functions. It's not needed here since there are no public API calls in either H5AC or H5C.

jhendersonHDF (approved):

PR #1538, opened 2022-03-28
Fix bugs in test/vfd.c

fortnern:
Fix uninitialized variable and improper range checking.


fortnern:
I can move this to a clean branch to get rid of the spurious commits if that helps. They'll all be flattened when it's merged though right?

soumagne (approved):

lrknox (approved):

PR #1556, opened 2022-03-30
Add H5_CHECK_MPI_VERSION macro to check for at least MPI version of (version, subversion)

jhendersonHDF:
Also fixes an MPI version check for MPI_Aint_diff usage in parallel compression related code that I noticed after reviewing the selection I/O merge.


soumagne (commented):
I think it would be good to use a similar macro as in: https://github.com/HDFGroup/vol-daos/blob/master/src/daos_vol.h#L462

qkoziol (approved):

jhendersonHDF:
> I think it would be good to use a similar macro as in instead: https://github.com/HDFGroup/vol-daos/blob/master/src/daos_vol.h#L462

Yes I think that's a good idea; let me make that change.

lrknox (approved):

soumagne (approved):

PR #1565, opened 2022-03-31
Feature/subfiling

jhendersonHDF:
Only needs review by @brtnfld


brtnfld (approved):
Looks good to me.
But I see what you mean; it seems like a lot of work left to do...So many #if....
This is a great start.

For testing, we can also use utils/subfiling_vfd/h5fush.sh to combine the subfiles to make sure it is a valid HDF5 file.

PR #1574, opened 2022-04-01
1.12 New References for Dimension Scales (#1139)

lrknox:
* Enable usage of new-style references with dimension scale APIs

* Add API to check if an object ID represents a native connector object

* Modified code to use new function H5DSwith_new_ref to determine if new references should be used with
Dimension Scales. The new function return TRUE if non-native connector is used or if H5_DIMENSION_SCALES_WITH_NEW_REF
varible is define at configure time (--enable-dimension-scales-with-new-ref).

Tested on jelly.

ToDo: generate testing file on BE system and enable the test; add flag to CMake; test netCDF-4 with the new references.

* Adding new test files generated on BE system (hedgehog) created by 32 and 64-bit library.

test_ds chokes on test_ds_le_new_ref.h5  on BE system; test passes for test_ds_be_new_ref-32bit.h5
for the 32-bit library and fails for the 64-bit library, and vice versa. I am checking the files for further
investigation; but current implementation of the new references is not portable between LE and BE systems,
and 32 and 64-bit systems.

* Minor fixes for testing issues

* Update test_ds.c

Enabled broken test; tests pass now.

* Update RELEASE.txt

Documented new option to use new references with the HDF5 dimension scales APIs (H5DS*).

* Update MANIFEST for new 32-bit new-style references test file for H5DS APIs

* Update 'dimension scales w/ new-style refs' feature based on review

Co-authored-by: Elena <epourmal@hdfgroup.org>


src/H5VL.c
```diff
@@ -654,6 +654,37 @@ H5VLobject(hid_t id)
...
 } /* H5VLobject() */
 
+/*---------------------------------------------------------------------------
+ * Function:    H5VLobject_is_native
+ *
+ * Purpose:     Determines whether an object ID represents a native VOL
+ *              connector object.
+ *
+ * Return:      Success:    TRUE/FALSE
+ *              Failure:    FAIL
```
derobins: We need to fix this! This API call needs to be reworked so it returns herr_t and hbool_t is_native is an out parameter (htri_t is trash and should not be used in new code).

derobins (changes requested):
H5VLobject_is_native needs to be updated. I will do that over the weekend.

src/H5VL.c
lrknox: It was changed 5 days later.  I've pulled that change in, please review again.

derobins (approved):

jhendersonHDF:
I was under the impression that we weren't planning on bringing these changes back to the support branches. Do we have a reason to support it for 1.12?

lrknox:
Is it needed for netCDF-4?

PR #1578, opened 2022-04-04
added check for stdatomic.h

brtnfld:


CMakeLists.txt
```diff
@@ -748,6 +748,10 @@ if (HDF5_BUILD_SUBFILING)
   if (NOT HDF5_ENABLE_PARALLEL)
     message (FATAL_ERROR "Subfiling VFD requires a parallel HDF5 build")
   endif()
+  CHECK_INCLUDE_FILE("stdatomic.h" HAVE_STDATOMIC_H)
```
byrnHDF, line 751: CMake will need this added to the h5pubconf.h.in file

lrknox: You'll probably want to add that to this PR?

brtnfld: One needs to add standard C headers? What is the purpose of defining it in h5pubconf.h.in?

byrnHDF: configure will create a pubconf.h with whatever checks are used. pubconf captures the checks into defines, which can then used by the code. Because CMake works differently, we need to maintain h5pubconf.h.in for CMake to remain consistent with configure. 
I understand you are only checking for the requirement and probably won't need it afterwards, but configure will add it to it's pubconf.

jhendersonHDF (approved):

PR #1582, opened 2022-04-05
fix old-style-definition warning

brtnfld:
Treated as an error, fixes DT.


lrknox (approved):

brtnfld:
Seemed to have old commits; I'll fix it.

PR #1606, opened 2022-04-07
Minor C++ changes noticed while normalizing with 1.12

derobins:


c++/src/H5Attribute.cpp
```diff
@@ -11,6 +11,7 @@
  * help@hdfgroup.org.                                                        *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+#include <cstdlib>
```
derobins, line 14: Needed for free(). Not having this doesn't break the build, but only due to accidental inclusion via other files.

lrknox (approved):

byrnHDF (approved):

PR #1611, opened 2022-04-07
1.12: Brings 1.12 compiler options in line with develop

derobins:


src/H5public.h
```diff
@@ -623,7 +623,7 @@ H5_DLL herr_t H5set_free_list_limits(int reg_global_lim, int reg_list_lim, int a
  *          garbage collected with H5garbage_collect(). These lists are global
  *          for the entire library.
  *
- * \since 1.10.7
```
byrnHDF: No this was correct - they were introduced in 1.10.7 - see 1.10 file

derobins: Ah, then it's incorrect in develop. Will fix in this PR and develop.

byrnHDF (approved):

lrknox (approved):

PR #1619, opened 2022-04-08
Subfiling/IOC VFD fixes and tidying

jhendersonHDF:
Only needs review from @brtnfld


brtnfld (approved):
require_ioc is probably not getting passed to open_subfiles, unless I missed something.

jhendersonHDF:
> require_ioc is probably not getting passed to open_subfiles, unless I missed something.

It doesn't look like it ever was in the past either. The only place require_ioc is used is by the main subfiling VFD code and only in a few places. I don't think open_subfiles does anything with it.

PR #1622, opened 2022-04-08
Removes the STATIC flavor of FUNC_ENTER macros

derobins:
With the recent changes to initialization, the FUNC_ENTER_STATIC macros
are now identical to their FUNC_ENTER_PACKAGE counterparts and can
be replaced, simplifying the macro scheme.

Removed from H5private.h and replaced with PACKAGE variant in *.c/h:
    FUNC_ENTER_STATIC
    FUNC_ENTER_STATIC_NOERR
    FUNC_ENTER_STATIC_TAG

Converted to PACKAGE in H5private.h and replaced in *.c/h:
    FUNC_ENTER_STATIC_NOERR_NOFS
    FUNC_ENTER_STATIC_NAMECHECK_ONLY

Merging this commit into feature branches may be difficult due to the extent of
these changes. Alternatively, you can manually snip out and convert the
FUNC_ENTER macros in H5private.h as was done here, then run this command:

`find . -type f -name '*.c' -o -name '*.h' | xargs sed -i 's/FUNC_ENTER_STATIC/FUNC_ENTER_PACKAGE/g'`

Since the initialization scheme has not been moved to other branches,
this change does NOT need to be merged down to 1.12 or 1.10.

### DIFF
--- a/src/H5A.c
+++ b/src/H5A.c
@@ -117,7 +117,7 @@ H5A__create_common(H5VL_object_t *vol_obj, H5VL_loc_params_t *loc_params, const 
     void *attr      = NULL;            /* Attribute created */
     hid_t ret_value = H5I_INVALID_HID; /* Return value */
 
-    FUNC_ENTER_STATIC
+    FUNC_ENTER_PACKAGE
 
     /* Sanity checks */
     HDassert(vol_obj);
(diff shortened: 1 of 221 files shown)


derobins:
This touches a lot of files, but it's super easy to review. The only interesting code changes are in H5private.h. Everything else is a single line rename done via sed.

jhendersonHDF (approved):

fortnern (approved):

PR #1623, opened 2022-04-08
Undo increment of max_idx for addition of 1.14 version from develop in

lrknox:
make_vers.  max_idx must match 1.12 version, otherwise build error results
with --disable-deprecated-symbols.

### DIFF
--- a/bin/make_vers
+++ b/bin/make_vers
@@ -8,8 +8,8 @@ use warnings;
 #       is added (like support for 1.4, etc), the min_sup_idx parameter will
 #       need to be decremented.)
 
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):
When did this happen - plus when should it change?

lrknox:
> 

Happened this morning when changes for bin/make_vers were brought from develop, where the latest version is 1.14.  the latest version can't exceed the current version in the code.  The result was this error with --disable-deprecated-symbols:                                                                                                       current/src/H5version.h:47:4: error: #error "Can't choose old API versions when deprecated APIs are disabled"
   #error "Can't choose old API versions when deprecated APIs are disabled"

derobins (approved):

derobins:
Oops! Sorry!

PR #1624, opened 2022-04-08
Removes unused definitions from module headers

derobins:
This change probably does NOT need to be moved down to 1.12 or earlier.


byrnHDF (approved):

derobins:
If you dig down, all the tests pass. The formatting check failure and correction is just burying them.

lrknox (approved):

soumagne (approved):

jhendersonHDF (approved):

PR #1628, opened 2022-04-08
Create hdf5_1_12_2 preparation branch for HDF5 1.12.2 release.

lrknox:
Set version to 1.12.2-3.
Switch default build mode to production for release.
Switch AM_MAINTAINER_MODE to disabled.
Set HDF5_GENERATE_HEADERS to OFF.
Commit Makefile.ins and other generated files for release.


lrknox:
I will be doing these before merging this pull request:
    update so numbers.
    remove any generated files that shouldn't be added.
    correct misspelled words in generated files such as configure and libtool, or update the codespell ignored words list if necessary.

byrnHDF:
I thought those bin files were supposed to be removed?

lrknox:
Which ones?  I can go to the PR to remove the MANIFEST file and see which ones were removed, and then see if configure and make will work without them.  Remember that we check in files to release branches generated by autogen.sh that are not checked  into the support branches.  If configure and make need them, they should be checked in, but if they aren't needed I'll remove them.

There were 8 files generated in the bin directory.  4 of them are needed to run condifigure on jelly:  config.guess, config.sub, install-sh and missing.  The other 4 are generated by autotools and may be necessary for other systems/compilers:  compile, depcomp, ltmain.sh and test-driver.  Also, the generated files with misspelled words are not in bin.

byrnHDF (approved):

PR #1636, opened 2022-04-14
Retrieve MPI-IO hints used by MPI library after file open

wkliao:
H5Pget_fapl_mpio() should return an MPI info object containing all the
MPI-IO hints used by the MPI library underneath, after the file is
opened. Some hints, such as cb_nodes (number of I/O aggregators), are
useful for HDF5 applications and I/O libraries built on top of HDF5.

### DIFF
--- a/src/H5FDmpio.c
+++ b/src/H5FDmpio.c
@@ -839,6 +839,7 @@ H5FD__mpio_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t H5_ATTR
     H5P_genplist_t *plist;                /* Property list pointer */
     MPI_Comm        comm = MPI_COMM_NULL; /* MPI Communicator, from plist */
     MPI_Info        info = MPI_INFO_NULL; /* MPI Info, from plist */
+    MPI_Info        info_used;            /* MPI Info returned from MPI_File_open */
     MPI_File        fh;                   /* MPI file handle */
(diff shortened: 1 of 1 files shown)


qkoziol (approved):

derobins (approved):

src/H5FDmpio.c
```diff
@@ -906,6 +907,54 @@ H5FD__mpio_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t H5_ATTR
...
+                    HMPI_GOTO_ERROR(NULL, "MPI_Info_get failed", mpi_code)
+
+                /* copy the hint into info */
+                if (MPI_SUCCESS != (mpi_code = MPI_Info_set(info, key, value)))
+                    HMPI_GOTO_ERROR(NULL, "MPI_Info_set failed", mpi_code)
+            }
+
+            /* Free info_used allocated in the call to MPI_File_get_info() */
+            if (MPI_SUCCESS != (mpi_code = MPI_Info_free(&info_used)))
+                HMPI_GOTO_ERROR(NULL, "MPI_Info_free failed", mpi_code)
```
soumagne, line 951: you might want to add the info free also on error code path, if the previous calls fail. To do that cleanly it may be easier if this block of code is extracted into its own routine.

PR #1644, opened 2022-04-14
Hdf5 1 10 parallel comp

jhendersonHDF:


examples/CMakeTests.cmake
```diff
@@ -101,22 +101,26 @@ if (H5_HAVE_PARALLEL AND HDF5_TEST_PARALLEL AND NOT WIN32)
...
+      add_test (NAME MPI_TEST_EXAMPLES-${parallel_example} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${NUMPROCS} ${MPIEXEC_PREFLAGS} $<TARGET_FILE:${parallel_example}> ${MPIEXEC_POSTFLAGS})
+    else ()
+      add_test (NAME MPI_TEST_EXAMPLES-${parallel_example} COMMAND "${CMAKE_COMMAND}"
+          -D "TEST_PROGRAM=${MPIEXEC_EXECUTABLE};${MPIEXEC_NUMPROC_FLAG};${NUMPROCS};${MPIEXEC_PREFLAGS};$<TARGET_FILE:${parallel_example}>;${MPIEXEC_POSTFLAGS}"
+          -D "TEST_ARGS:STRING="
+          -D "TEST_EXPECT=0"
+          -D "TEST_SKIP_COMPARE=TRUE"
+          -D "TEST_OUTPUT=${parallel_example}.out"
+          -D "TEST_REFERENCE:STRING=PHDF5 example finished with no errors"
+          #-D "TEST_FILTER:STRING=PHDF5 tests finished with no errors"
```
byrnHDF, line 116: This TEST_FILTER is not needed? Maybe because grepTest.cmake just uses the TEST_REFERENCE line.

byrnHDF (approved):

lrknox (approved):

examples/CMakeTests.cmake
jhendersonHDF: Not sure. It didn't appear to be needed and was problematic since the new parallel examples I added don't print the same output as the old ph5example.

PR #1659, opened 2022-04-19
[WIP] Add Developer build mode to CMake

jhendersonHDF:
Looking to get some initial feedback on a Developer mode. This is just for CMake for now to see if people are happy with this approach and then Autotools should follow.

There a still a couple new SWMR test failures on this branch that I'm looking into.


config/cmake/HDF5DeveloperBuild.cmake
```diff
@@ -0,0 +1,207 @@
...
+mark_as_advanced (
+  CMAKE_CXX_FLAGS_DEVELOPER
+  CMAKE_C_FLAGS_DEVELOPER
+  CMAKE_EXE_LINKER_FLAGS_DEVELOPER
+  CMAKE_SHARED_LINKER_FLAGS_DEVELOPER
+)
+
+#-----------------------------------------------------------------------------
+# Define various HDF5 macros for debugging the library
+#-----------------------------------------------------------------------------
```
jhendersonHDF, line 50: For all the debug macros in the library, it might be more maintainable to have a developer-type header file in the source where they're all collected in one place instead, but I didn't know if that approach would be better. For this approach, I just have a CMake option for all the DEBUG type macros I know about that get added in to Developer builds.

config/cmake/HDF5DeveloperBuild.cmake
```diff
@@ -0,0 +1,207 @@
...
+mark_as_advanced (HDF5_ENABLE_DEBUG_H5FS_ASSERT)
+if (HDF5_ENABLE_DEBUG_H5FS_ASSERT)
+  add_compile_definitions (H5FS_DEBUG_ASSERT)
+endif ()
+
+# If HDF5 free list debugging wasn't specifically enabled, disable
+# free lists entirely for developer build modes, as they can
+# make certain types of issues (like references to stale pointers)
+# much more difficult to debug
+if (NOT HDF5_ENABLE_DEBUG_H5FL AND NOT HDF5_ENABLE_DEBUG_H5FL_TRACK)
```
jhendersonHDF, line 191: The library's free lists can make debugging of certain types of issues more difficult, so Developer build modes would have free lists disabled. There doesn't appear to be a noticeable impact for the library's tests, but there are a couple issues with the SWMR tests when free lists are disabled that I need to fix.

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -267,6 +271,16 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
     #  ADD_H5_FLAGS (H5_CFLAGS "${HDF5_SOURCE_DIR}/config/gnu-warnings/no-developer-10")
     endif ()
   endif ()
+
+  # Append more extra warning flags that only gcc 12.x+ knows about
+  # or which should only be enabled for gcc 12.x+
+  if (NOT CMAKE_C_COMPILER_VERSION VERSION_LESS 12.0)
+    if (HDF5_ENABLE_DEV_WARNINGS)
+      ADD_H5_FLAGS (H5_CFLAGS "${HDF5_SOURCE_DIR}/config/gnu-warnings/developer-12")
```
jhendersonHDF, line 280: Add in the flag for GCC's static analyzer for GCC 12+

test/tmisc.c
```diff
@@ -5847,7 +5847,7 @@ test_misc35(void)
     ret = H5get_free_list_sizes(&reg_size_start, &arr_size_start, &blk_size_start, &fac_size_start);
     CHECK(ret, FAIL, "H5get_free_list_sizes");
 
-#if !defined H5_USING_MEMCHECKER
+#if !defined H5_NO_FREE_LISTS && !defined H5_USING_MEMCHECKER
```
jhendersonHDF, line 5912: Fix tmisc.c when free lists are disabled via the H5_NO_FREE_LISTS #define

soumagne:
I think that looks good, you might want to check the `OUTPUT_NAME_DEVELOPER` is also set in the target properties, you probably want that to be the same as `OUTPUT_NAME_DEBUG`. (See HDFMacros.cmake)

config/cmake/HDF5DeveloperBuild.cmake
```diff
@@ -0,0 +1,207 @@
...
+
+#-----------------------------------------------------------------------------
+# Define various HDF5 macros for debugging the library
+#-----------------------------------------------------------------------------
+
+# Enable debugging of various HDF5 modules
+set (HDF5_ENABLE_DEBUG_APIS ON CACHE BOOL "Turn on extra debug output in all packages" FORCE)
+
+# Enable debugging of API contexts
+add_compile_definitions (H5CX_DEBUG)
```
byrnHDF: These defines are only needed for the library, right? add_compile_definitions is a global sledgehammer and discouraged by CMake devs.

see the src/CMakeLists.txt file about line 1183; this "target_compile_definitions" is preferred. In this case, create a HDF5_ENABLE_DEBUG_DEVEL_APIS option and then in src/CMakeLists.txt add annother  BOOL generator line to the target.
Same for the following options.

jhendersonHDF: If we don't consider other future debug options (like for the tools), then yes these are only relevant to the library, but there's a bit of an organization issue. I'd like to use a specific target_compile_definitions, but couldn't here because the target for the library isn't defined yet at the point where this .cmake file is included in the top-level CMakeLists.txt. I want to set all the debug options as early as possible so they apply to the whole project and are also defined before we start building the library and its components. If there's a way we can ensure the library's target is defined in the top-level CMakeLists.txt, then we could reduce the scope that these definitions are applied at.

As for the HDF5_ENABLE_DEBUG_DEVEL_APIS, there's already an option for this; the `set` call is just the best way I could see of force-overriding what was previously set for that option. I could have created another variable, but I didn't want to have to check 2 things before deciding whether to enable the debug APIs. I just wanted to re-use what already exists. It seems a bit unclean since I apparently have to specify the same docstring that the option was created with, but let me know if there's a better wayt to do this in CMake.

byrnHDF: set call is fine - it's the add_compile_definitions - by adding the BOOL generator lines to the target they are only applied to that target with the correct visibility. If you want the same defs applied to another target - you just add the same line to those targets. Only those targets you add the line to is affected.

byrnHDF: CMake is pushing target level adjustments rather then global adjustments.

byrnHDF: I probably need to start using generators more, since more CMake commands are supporting it.

jhendersonHDF: > set call is fine - it's the add_compile_definitions - by adding the BOOL generator lines to the target they are only applied to that target with the correct visibility. If you want the same defs applied to another target - you just add the same line to those targets. Only those targets you add the line to is affected.

Why create yet another generator line for the HDF5 lib target? It already has one for if the HDF5_ENABLE_DEBUG_APIS option is enabled, so re-using that makes sense instead of adding more maintenance burden with two options the target is pulling from.

byrnHDF: That's fine if you just want to extend the option instead of creating new options.

byrnHDF: Another thing we could do is create a "DEBUG_HDF5_APIS" var that is the list of defines to be used and just use that var in the generator line. Instead of the add_compile_definitions(xxxx) do a list (append ${DEBUG_HDF5_APIS} xxxx).
Then change the definitions generator to use ${DEBUG_HDF5_APIS}.

byrnHDF:
Let's do some refactoring.
1. Move the option - HDF5_ENABLE_DEBUG_APIS to config/cmake/HDFCompilerFlags just above the asserts option (about line 299).
2. Next in the option if block(new) create the DEBUG_HDF5_APIS list with the default defines: H5Z_DEBUG;H5T_DEBUG;H5S_DEBUG;H5O_DEBUG;H5I_DEBUG;H5HL_DEBUG;H5F_DEBUG;H5D_DEBUG;H5B2_DEBUG;H5AC_DEBUG
3. Replace the defines in src/CMakeLists.txt with the list variable.
4. Change all the developer file add_compile_defines to the list (APPEND ${DEBUG_HDF5_APIS} xxxx)

byrnHDF (approved):

jhendersonHDF:
I believe this PR is ready to go. I don't see any failures in the SWMR tests after merging with develop, but if any crop up I can go fix them.

derobins (approved):

PR #1673, opened 2022-04-21
added libs rt and pthreads

brtnfld:
will need to add an option to specify a directory, like the current --with-pthreads=DIR. But we don't want to enable pthreads for the HDF5 library itself.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3231,6 +3231,10 @@ if test "X$SUBFILING_VFD" = "Xyes"; then
       AC_MSG_ERROR([Subfiling VFD requires atomic operations support. C11 stdatomic.h NOT available.])
     fi
 
+# Checks for libraries.
+    AC_CHECK_LIB([rt], [shm_open])
+    AC_CHECK_LIB([pthread], [pthread_self],[], [echo "Error: Required library pthread not found." && exit 1])
+
 else
     AC_MSG_RESULT([no])
 fi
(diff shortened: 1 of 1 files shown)


jhendersonHDF (approved):

configure.ac
```diff
@@ -3231,6 +3231,10 @@ if test "X$SUBFILING_VFD" = "Xyes"; then
       AC_MSG_ERROR([Subfiling VFD requires atomic operations support. C11 stdatomic.h NOT available.])
     fi
 
+# Checks for libraries.
+    AC_CHECK_LIB([rt], [shm_open])
```
derobins: We already check for rt in the context of clock_gettime (line 2090). You could probably replace the clock_gettime check (posix4 is archaic, IIRC) with this one. Also, in VFD SWMR I have to always link to pthreads on MacOS (though I'd have to double check to be sure).

brtnfld: I saw that, but for some reason, the testing on Jelly showed that it was not adding it.

derobins: It's using AC_SEARCH_LIBS, which first checks to see if the symbol is available without linking to anything new. If you change that to AC_CHECK_LIB, you'll always link. AC_SEARCH_LIBS will also check for libposix4, but that functionality was moved to libc a long time ago, making libposix4 archaic and we can ignore it.

derobins: Or you could do two AC_SEARCH_LIBS calls, one for clock_gettime and one for shm_open. AC_SEARCH_LIBS is considered preferable to AC_CHECK_LIB, as the former won't bring in anything new unless it's necessary.

PR #1677, opened 2022-04-22
Adding the support of H5FDctl for the Onion VFD and the test case

raylu-hdf:


src/H5FDpublic.h
```diff
@@ -197,6 +197,7 @@
 #define H5FD_CTL__MEM_ALLOC                   5
 #define H5FD_CTL__MEM_FREE                    6
 #define H5FD_CTL__MEM_COPY                    7
+#define H5FD_CTL__GET_NUM_REVISIONS           8
```
derobins, line 200: We can do this in the short term, but this is going to be a problem going forward. Many VFDs will want to accept values like this, but they won't be able to modify H5FDpublic.h, so we need a system that generalizes to all VFDs.

test/onion.c
```diff
@@ -4263,6 +4263,221 @@ test_integration_dset_extension(void)
...
+     */
+    file = H5Fopen(paths->canon, H5F_ACC_RDWR, fapl_id);
+    if (H5I_INVALID_HID == file) {
+        TEST_ERROR;
+    }
+
+    /*
+     * Create the dataset using the dataset creation property list.
+     */
+    dset = H5Dcreate(file, "DS2", H5T_STD_I32LE, space, H5P_DEFAULT, dcpl, H5P_DEFAULT);
```
derobins, line 4386: Should check for dset == H5I_INVALID_HID like other calls

PR #1680, opened 2022-04-24
Warnings fixes

jhendersonHDF:


src/H5FDmpio.c
```diff
@@ -1865,7 +1865,7 @@ H5FD__mpio_vector_build_types(uint32_t count, H5FD_mem_t types[], haddr_t addrs[
                 if (!sub_types) {
                     HDassert(!sub_types_created);
 
-                    if (NULL == (sub_types = (int *)HDmalloc((size_t)count * sizeof(MPI_Datatype))))
+                    if (NULL == (sub_types = HDmalloc((size_t)count * sizeof(MPI_Datatype))))
```
jhendersonHDF, line 1868: sub_types is an MPI_Datatype *, not an int *

test/dt_arith.c
```diff
@@ -2694,18 +2694,18 @@ my_isnan(dtype_t type, void *val)
     if (FLT_FLOAT == type) {
         float x = 0.0;
         HDmemcpy(&x, val, sizeof(float));
-        retval = (x != x);
+        retval = !H5_FLT_ABS_EQUAL(x, x);
```
jhendersonHDF, line 2697: Don't compare floats, doubles, long doubles using != here. Use library macro to compare with an epsilon value instead.

test/vfd.c
```diff
@@ -4911,13 +4922,13 @@ test_vector_io(const char *vfd_name)
 
     for (i = 0; i < count; i++) {
 
-        HDfree(write_bufs_0[i]);
+        HDfree((void *)write_bufs_0[i]);
```
jhendersonHDF, line 4925: Since H5FDwrite_vector takes a const void ** argument, the cleanest way to fix the warnings in this file is to have them actually be const void **. Unfortunately that means we will always have const warnings from freeing those pointers, unless we use our flex_ptr union hack here. For now, the explicit cast to (void *) is meant to highlight that.

testpar/t_bigio.c
```diff
@@ -1151,14 +1151,14 @@ single_rank_independent_io(void)
         VRFY_G((mspace_id >= 0), "H5Screate_simple mspace_id succeeded");
 
         /* Write data */
-        H5Dwrite(dset_id, H5T_NATIVE_INT, mspace_id, fspace_id, H5P_DEFAULT, data);
+        ret = H5Dwrite(dset_id, H5T_NATIVE_INT, mspace_id, fspace_id, H5P_DEFAULT, data);
         VRFY_G((ret >= 0), "H5Dwrite succeeded");
```
jhendersonHDF, line 1155: ret had never been initialized or set in this test file, so it's very possible t_bigio might've had sporadic failures from checking an uninitialized ret.

src/H5Dcontig.c
```diff
@@ -615,13 +615,15 @@ H5D__contig_may_use_select_io(const H5D_io_info_t *io_info, H5D_io_op_type_t op_
     HDassert(dataset);
     HDassert(op_type == H5D_IO_OP_READ || op_type == H5D_IO_OP_WRITE);
 
+    dataset = io_info->dset;
```
qkoziol, line 618: This should be immediately before the HDassert(dataset) line, otherwise that assert will always be tripped.

tools/src/h5diff/ph5diff_main.c
```diff
@@ -312,5 +312,5 @@ h5diff_exit(int status)
     /* Always exit(0), since MPI implementations do weird stuff when they
      *  receive a non-zero exit value. - QAK
      */
-    HDexit(0);
+    HDexit(status);
```
qkoziol, line 315: This shouldn't be changed (per my comment immediately above :-)

qkoziol (changes requested):
Two small fixes needed, but otherwise looks good

tools/src/h5diff/ph5diff_main.c
jhendersonHDF: Interesting if there are still MPI implementations that behave weird. On the other hand, just a few lines up we always set status to EXIT_SUCCESS when in parallel, and EXIT_SUCCESS should be equal to 0, so (I think?) this shouldn't be a problem.

src/H5Dcontig.c
jhendersonHDF: Was building in release so missed this one! Just fixed the assertion and pushed.

tools/src/h5diff/ph5diff_main.c
qkoziol: Worth a try :-)

qkoziol (approved):

src/H5FDmpio.c
derobins: There's no need to cast the allocated pointers in C anyways.

tools/src/h5diff/ph5diff_main.c
derobins: EXIT_SUCCESS is used instead of 0/1 because the "good" and "bad" error codes are platform-specific. VMS/OpenVMS were weird about return codes, for example. I wouldn't count on EXIT_SUCCESS always being zero, even if that's a practical assumption today. That said, I would assume that MPI on a "weird" system would do the right thing with EXIT_SUCCESS.

derobins (approved):

PR #1684, opened 2022-04-25
Address some warnings from casting away of const

jhendersonHDF:
Attempt to address some of the const cast warnings that are reasonable to fix. Most of the remaining ones are involved enough to be separate PRs on their own.

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -1143,8 +1143,7 @@ done:
     fm->mem_space  = NULL;
 
     if (file_space_normalized == TRUE)
-        if (H5S_hyper_denormalize_offset((H5S_t *)file_space, old_offset) <
-            0) /* (Casting away const OK -QAK) */
+        if (H5S_hyper_denormalize_offset(file_space, old_offset) < 0)
             HDONE_ERROR(H5E_DATASET, H5E_CANTSET, FAIL, "can't denormalize selection")
 
     FUNC_LEAVE_NOAPI(ret_value)
@@ -1371,15 +1370,16 @@ done:
  *-------------------------------------------------------------------------
  */
 void *
-H5D__chunk_mem_alloc(size_t size, const H5O_pline_t *pline)
+H5D__chunk_mem_alloc(size_t size, void *pline)
 {
-    void *ret_value = NULL; /* Return value */
+    H5O_pline_t *_pline    = (H5O_pline_t *)pline;
+    void *       ret_value = NULL; /* Return value */
 
     FUNC_ENTER_STATIC_NOERR
 
     HDassert(size);
 
-    if (pline && pline->nused)
+    if (_pline && _pline->nused)
         ret_value = H5MM_malloc(size);
     else
         ret_value = H5FL_BLK_MALLOC(chunk, size);
@@ -1402,14 +1402,14 @@ H5D__chunk_mem_alloc(size_t size, const H5O_pline_t *pline)
  *-------------------------------------------------------------------------
  */
 void *
-H5D__chunk_mem_xfree(void *chk, const void *_pline)
+H5D__chunk_mem_xfree(void *chk, const void *pline)
 {
-    const H5O_pline_t *pline = (const H5O_pline_t *)_pline;
+    const H5O_pline_t *_pline = (const H5O_pline_t *)pline;
 
     FUNC_ENTER_STATIC_NOERR
 
     if (chk) {
-        if (pline && pline->nused)
+        if (_pline && _pline->nused)
             H5MM_xfree(chk);
         else
             chk = H5FL_BLK_FREE(chunk, chk);
@@ -1426,9 +1426,9 @@ H5D__chunk_mem_xfree(void *chk, const void *_pline)
  *-------------------------------------------------------------------------
  */
 void
-H5D__chunk_mem_free(void *chk, const void *_pline)
+H5D__chunk_mem_free(void *chk, void *pline)
 {
-    (void)H5D__chunk_mem_xfree(chk, _pline);
+    (void)H5D__chunk_mem_xfree(chk, pline);
 }
 
 /*-------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


src/H5Dchunk.c
```diff
@@ -1143,8 +1143,7 @@ H5D__chunk_io_init(H5D_io_info_t *io_info, const H5D_type_info_t *type_info, hsi
     fm->mem_space  = NULL;
 
     if (file_space_normalized == TRUE)
-        if (H5S_hyper_denormalize_offset((H5S_t *)file_space, old_offset) <
-            0) /* (Casting away const OK -QAK) */
+        if (H5S_hyper_denormalize_offset(file_space, old_offset) < 0)
```
jhendersonHDF, line 1146: `file_space` here was already changed to not be const-qualified. Just remove the comment and cast.

src/H5Dchunk.c
```diff
@@ -1371,15 +1370,16 @@ H5D__chunk_io_init_selections(const H5D_io_info_t *io_info, const H5D_type_info_
  *-------------------------------------------------------------------------
  */
 void *
-H5D__chunk_mem_alloc(size_t size, const H5O_pline_t *pline)
+H5D__chunk_mem_alloc(size_t size, void *pline)
```
jhendersonHDF, line 1373: These routines in H5Dchunk.c didn't match the H5MM_allocate_t and H5MM_free_t function pointer typedefs, but this was hidden by casting them to such. Correct that and do the casting to H5O_pline_t in the routine. I recall this being changed from void * to H5O_pline_t * previously, but let's actually match our specification for the typedefs and not hide that.

src/H5Dchunk.c
```diff
@@ -1580,8 +1580,7 @@ H5D__create_chunk_map_single(H5D_chunk_map_t *fm, const H5D_io_info_t
     chunk_info->fspace_shared = TRUE;
 
     /* Just point at the memory dataspace & selection */
-    /* (Casting away const OK -QAK) */
-    chunk_info->mspace = (H5S_t *)fm->mem_space;
+    chunk_info->mspace = fm->mem_space;
```
jhendersonHDF, line 1583: fm->mem_space already had const removed.

src/H5Dint.c
```diff
@@ -488,7 +488,7 @@ H5D__new(hid_t dcpl_id, hid_t dapl_id, hbool_t creating, hbool_t vl_type)
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5D__init_type(H5F_t *file, const H5D_t *dset, hid_t type_id, const H5T_t *type)
+H5D__init_type(H5F_t *file, const H5D_t *dset, hid_t type_id, H5T_t *type)
```
jhendersonHDF, line 491: Since we might take the input type and directly assign it to a modifiable `dset->shared->type` field, let's not lie about the `type` parameter's const-ness here.

src/H5Dint.c
```diff
@@ -2192,7 +2192,7 @@ H5D_oloc(H5D_t *dataset)
  *-------------------------------------------------------------------------
  */
 H5G_name_t *
-H5D_nameof(const H5D_t *dataset)
+H5D_nameof(H5D_t *dataset)
```
jhendersonHDF, line 2195: While in theory returning the name of an object should be able to accept a const qualified pointer to the object, the returned value can potentially be used in a non-const way, so until we can properly architect things, the _nameof routines for objects should just accept non-const-qualified pointers to objects. Note that this follows the same as the _oloc routines for objects.

src/H5Dio.c
```diff
@@ -174,21 +174,21 @@ H5D__read(H5D_t *dataset, hid_t mem_type_id, H5S_t *mem_space, H5S_t *file_space
...
-        const void *adj_buf = NULL; /* Pointer to the location in buf corresponding  */
-                                    /* to the beginning of the projected mem space.  */
+        ptrdiff_t buf_adj = 0;
 
         /* Attempt to construct projected dataspace for memory dataspace */
         if (H5S_select_construct_projection(mem_space, &projected_mem_space,
-                                            (unsigned)H5S_GET_EXTENT_NDIMS(file_space), buf, &adj_buf,
-                                            type_info.dst_type_size) < 0)
+                                            (unsigned)H5S_GET_EXTENT_NDIMS(file_space),
+                                            type_info.dst_type_size, &buf_adj) < 0)
```
jhendersonHDF, line 182: In the read case, having H5S_select_construct_projection always take const-qualified pointers to buffers and try to return const-qualified pointers to potentially adjusted buffers is problematic. The non-const read buffer goes in as const-qualified and comes back as const-qualified, then we cast away the const since the buffer should be modifiable.

Rather than having H5S_select_construct_projection try to do the adjusting of buffers, just return an adjustment amount and let the higher level do the adjusting of buffers, which works fine for both the read and write case.

src/H5Gent.c
```diff
@@ -300,7 +300,7 @@ H5G_ent_encode(const H5F_t *f, uint8_t **pp, const H5G_entry_t *ent)
  *-------------------------------------------------------------------------
  */
 void
-H5G__ent_copy(H5G_entry_t *dst, const H5G_entry_t *src, H5_copy_depth_t depth)
+H5G__ent_copy(H5G_entry_t *dst, H5G_entry_t *src, H5_copy_depth_t depth)
```
jhendersonHDF, line 303: While this routine should theoretically accept a const-qualified source H5G_entry_t, the pesky H5G_ent_reset call below means we're lying about the const-ness of the source parameter. Let's not do that for now, until future re-architecting. This is a common pattern in a few other spots (like H5G_name_copy), but fixing those is very involved.

src/H5Gloc.c
```diff
@@ -927,13 +926,15 @@ H5G__loc_set_comment_cb(H5G_loc_t H5_ATTR_UNUSED *grp_loc /*in*/, const char H5_
 
     /* Add the new message */
     if (udata->comment && *udata->comment) {
-        /* Casting away const OK -QAK */
-        comment.s = (char *)udata->comment;
+        if (NULL == (comment.s = HDstrdup(udata->comment)))
```
jhendersonHDF, line 928: Short of playing games with our flexible const pointer type, the easiest way to solve this appears to be taking a copy of the object comment passed in from above, since H5O_msg_create accepts a non-const message pointer.

soumagne (approved):

derobins (approved):

PR #1696, opened 2022-04-26
Fixes stack size warnings in dtransform

derobins:


test/dtransform.c
```diff
@@ -129,6 +131,10 @@ const int transformData[ROWS][COLS] = {{36, 31, 25, 19, 13, 7, 1, 5, 11, 16, 22,
         H5T_order_t order;                                                                                   \
         hsize_t     dim[2] = {ROWS, COLS};                                                                   \
                                                                                                              \
+        /* NOTE: If this part of the test macro encounters errors, this memory will leak */                  \
+        if (NULL == (array = HDcalloc(1, sizeof(*array))))                                                   \
+            TEST_ERROR;                                                                                      \
```
derobins, line 136: Due to the macros, cleaning the memory on errors (when the test is just going to die anyway) seems like it's not worth the extra code complexity, macros, and goto statements.

jhendersonHDF (approved):

test/dtransform.c
qkoziol: Yeah, seems fine in a limited way for the test code

qkoziol (approved):

soumagne (approved):

PR #1700, opened 2022-04-28
Fix some const cast and stack/static object size warnings

jhendersonHDF:


testpar/t_pflush2.c
```diff
@@ -213,10 +217,18 @@ main(int argc, char *argv[])
...
+        data_g = NULL;
+    }
+
     MPI_Finalize();
 
     HDexit(EXIT_SUCCESS);
 
 error:
+    if (data_g)
+        HDfree(data_g);
```
derobins, line 231: The check for data_g isn't necessary. You can call free() on a NULL pointer.

testpar/t_cache.c
```diff
@@ -6938,6 +6938,26 @@ main(int argc, char **argv)
...
+        if (MAINPROCESS)
+            HDprintf("    Couldn't allocate data index array.  Exiting.\n");
+        MPI_Abort(MPI_COMM_WORLD, -1);
+    }
+
+    HDmemset(filenames, 0, sizeof(filenames));
+    for (int i = 0; i < NFILENAME; i++) {
+        if (NULL == (filenames[i] = HDmalloc(PATH_MAX))) {
+            if (MAINPROCESS)
+                HDprintf("couldn't allocate filename array\n");
```
derobins: If anything other than rank 0 can't allocate memory, the program will exit without an error message.

test/vfd.c
```diff
@@ -140,6 +140,13 @@ static int __k;
         }                                                                                                    \
     } while (FALSE)
 
+/*
+ * Ugly hack to cast away const for freeing const-qualified memory
+ */
+#ifndef HDfree_const
+#define HDfree_const(mem) HDfree((void *)(uintptr_t)mem)
+#endif
```
derobins: We should put this in H5private.h. We already have n H5MM flavor, but this could be used more widely.

test/vds_env.c
```diff
@@ -44,13 +44,13 @@ const char *FILENAME[] = {"vds_env_virt_0", "vds_env_virt_3", "vds_env_src_2", "
 static int
 test_vds_prefix_second(unsigned config, hid_t fapl)
 {
-    char        srcfilename[FILENAME_BUF_SIZE];
-    char        srcfilename_map[FILENAME_BUF_SIZE];
-    char        vfilename[FILENAME_BUF_SIZE];
-    char        vfilename2[FILENAME_BUF_SIZE];
-    char        srcfilenamepct[FILENAME_BUF_SIZE];
-    char        srcfilenamepct_map[FILENAME_BUF_SIZE];
     const char *srcfilenamepct_map_orig = "vds%%%%_src";
```
derobins, line 47: What platform/compiler is triggering stack size warnings on these? I don't think I've been seeing them with whatever flavor of gcc I'm using. Is clang picking these up?

derobins (approved):
I'd probably move HDfree_const into H5private.h, though.

testpar/t_pflush2.c
jhendersonHDF: I know; bad habit

testpar/t_cache.c
jhendersonHDF: On the other hand, if they all can't allocate memory, then stdout will blow up

test/vfd.c
jhendersonHDF: I do hate this hack, but I thought about putting it in H5private.h. I don't want to encourage people to do this, but it is kind of necessary in some cases. I can put it in H5private and leave a stern comment.

test/vds_env.c
jhendersonHDF: GCC is complaining because the total stack frame size is over 9000 bytes

src/H5Tref.c
```diff
@@ -753,6 +753,14 @@ H5T__ref_mem_write(H5VL_object_t *src_file, const void *src_buf, size_t src_size
...
  *-------------------------------------------------------------------------
  */
+/*
+ * It is the correct thing to do to have the reference buffer
+ * be const-qualified here, but the VOL "blob specific" routine
+ * needs a non-const pointer since an operation might modify
+ * the "blob". Disable this warning here (at least temporarily)
+ * for this reason.
+ */
+H5_GCC_CLANG_DIAG_OFF("cast-qual")
```
soumagne, line 763: I feel like that could happen in several cases, we should probably fix that VOL callback to handle const and non-const

src/H5private.h
```diff
@@ -1561,6 +1561,15 @@ H5_DLL int HDvasprintf(char **bufp, const char *fmt, va_list _ap);
...
 #endif
 
+/*
+ * Ugly hack to cast away const for freeing const-qualified pointers.
+ * Should only be used sparingly, where the alternative (like keeping
+ * an equivalent non-const pointer around) is far messier messier.
+ */
+#ifndef HDfree_const
+#define HDfree_const(mem) HDfree((void *)(uintptr_t)mem)
+#endif
```
soumagne: Hmm ok if that's only used in tests, I think you should leave it in tests because having that in H5private.h seems ugly

test/vfd.c
soumagne: I see your original comment now Dana, I feel like I'm more on Jordan's side on that, it was better to leave it in that particular test imo. I don't think encouraging that kind of hack is a good thing :)

src/H5private.h
jhendersonHDF: I don't really like the way that this works, but we also do the same thing in H5MM_xfree_const. Right now it's only used in the tests so I only had it in the one test file, but moved it to H5private.h from Dana's suggestion. I can move it back if we don't really see it being useful elsewhere.

src/H5Tref.c
jhendersonHDF: Agreed. Just thought I'd temporarily disable it until we can do that. I'd like to eventually search for all the places we've turned things off with H5_GCC_CLANG_DIAG_OFF in the past and correctly fix things if we can, but that's for a later date.

src/H5private.h
derobins: Not being able to free const pointers is often considered to be a deficiency in the standard. If we leave it here, it shouldn't get HD as that implies it's a part of the platform-independence stuff in H5private.h. Maybe h5_free_const to go with the other tools functionality.

src/H5private.h
```diff
@@ -1561,6 +1561,15 @@ H5_DLL int HDvasprintf(char **bufp, const char *fmt, va_list _ap);
 #define HDwrite(F, M, Z) write(F, M, Z)
 #endif
 
+/*
+ * Ugly hack to cast away const for freeing const-qualified pointers.
+ * Should only be used sparingly, where the alternative (like keeping
+ * an equivalent non-const pointer around) is far messier messier.
```
derobins: Probably just need one "messier"

test/vds_env.c
derobins: Ah

src/H5private.h
jhendersonHDF: Decided to move it to h5test.h for now, since the tests are where this macro will most likely be used. We can always move it back to H5private.h in the future if we use it more widely.

soumagne (approved):

qkoziol (approved):

PR #1701, opened 2022-04-29
Fix object size warnings in cache.c test

jhendersonHDF:


test/cache.c
```diff
@@ -8420,1013 +8764,125 @@ check_flush_cache__flush_op_test(H5F_t *file_ptr, int test_num, unsigned int flu
...
-         FALSE,
-         FALSE,
-         {0, 0, 0, 0, 0, 0, 0, 0},
-         {0, 0, 0, 0, 0, 0, 0, 0},
-         0,
-         0,
-         0,
-         -1,
-         FALSE}};
+        for (i = 0; i < num_variable_entries; i++) {
```
jhendersonHDF: In this and all the following cases, I verified that the loop-generated structures had the same values as previously.

derobins (approved):

qkoziol (approved):
*whew*   Nice cleanup!

PR #1702, opened 2022-04-29
Fix SWMR/refresh bug hidden by library free lists

jhendersonHDF:
Use-after-free warnings in valgrind for SWMR tests that perform refresh operations. Was hidden by library's free lists and seems to have always worked, but definitely a bug. May or may not have been problematic when free lists are on since the memory is still around, but could have been overwritten at some point.


derobins (approved):

qkoziol (approved):
Nice catch!

PR #1705, opened 2022-04-29
Add a workaround for a bug in PGI C Compiler version 19.10

gnuoyd:
PGI C Compiler version 19.10 quits with the unhelpful and misleading error message `PGC-S-0053-Illegal use
of void type` when I put a `static` qualifier on my tables.

### DIFF
--- a/src/H5Tnative.c
+++ b/src/H5Tnative.c
@@ -1171,10 +1171,20 @@ H5T__init_native_internal(void)
 
     /* clang-format off */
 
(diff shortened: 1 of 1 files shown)


qkoziol (changes requested):
Scattering system or compiler ifdef's through the code is a long-term maintenance difficulty.   Please find a way to localize this into a header file.

brtnfld:
Has this been reported to PGI?

derobins:
Merging this in. I'll address Quincey's concern in a subsequent PR.

PR #1708, opened 2022-05-01
H5S: fix const correctness for `space` parameters

mathstuf:
---
Split from #1640.

### DIFF
--- a/src/H5Sall.c
+++ b/src/H5Sall.c
@@ -64,7 +64,7 @@ static herr_t   H5S__all_adjust_u(H5S_t *space, const hsize_t *offset);
 static herr_t   H5S__all_adjust_s(H5S_t *space, const hssize_t *offset);
 static herr_t   H5S__all_project_scalar(const H5S_t *space, hsize_t *offset);
 static herr_t   H5S__all_project_simple(const H5S_t *space, H5S_t *new_space, hsize_t *offset);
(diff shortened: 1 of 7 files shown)


jhendersonHDF:
Hey @mathstuf,

looks like we were tackling the same issue as each other! I ended up fixing this by removing const rather than adding it some places since it seems the selection iterator code may currently need to operate on the space during iteration and so the const seems a bit misleading.

mathstuf:
Really? Where is the `const` being stripped away then as I think it was the iterator code that led me to adding `const` everywhere.

jhendersonHDF:
> Really? Where is the `const` being stripped away then as I think it was the iterator code that led me to adding `const` everywhere.

Here's one such place in a previous PR: https://github.com/HDFGroup/hdf5/pull/1684/files#diff-ca61bfabd09b881fe4c59df5be0650fe189ebe672bb355e69392662c686e31a6.

Following a similar example in that file, I had marked the `dst_space` and `src_space` parameters as const, but then casted const away since the selection iterator code currently accepts a non-const `space` parameter. IMO, the selection iterator code should probably continue to accept a non-const `space` pointer for now, so I just removed const from the parameters to `H5D_select_io_mem`. It would likely be a good idea to have const and non-const versions of selection iterators in the future, but for now it seems best to leave this open.

In any case, if you build the latest HDF5 develop branch you should find all the H5S warnings addressed, unless I happened to miss one.

derobins:
This has already been addressed in a different way.

PR #1709, opened 2022-05-01
test/dt_arith: use `isnan` instead of self comparison

mathstuf:
It is more obvious what is being asked this way.

---
Split from #1640.


qkoziol (approved):

derobins (approved):

derobins:
I feel like this PR could be further improved if we remember that it's 2022 and not 1997 and simply remove my_isnan(), replacing it with isnan.

PR #1714, opened 2022-05-02
added subfiling_dir read

brtnfld:
Use the  subfile path in the filename.


derobins:
fwiw, you might want to run shellcheck over this file.

jhendersonHDF (approved):

PR #1716, opened 2022-05-03
CMake: respect GNUInstallDirs

thesamesam:
This makes life a lot easier for in particular Linux distributions
as it means CMake will respect a standard set of arguments for
libdir, docdir, etc.

It also brings the CMake in line with autotools in terms of paths.

We've been using this downstream in Gentoo for quite some time
with no problems.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/c++/src/CMakeLists.txt
+++ b/c++/src/CMakeLists.txt
@@ -173,8 +173,8 @@ endif ()
 #-----------------------------------------------------------------------------
 set (_PKG_CONFIG_PREFIX ${CMAKE_INSTALL_PREFIX})
 set (_PKG_CONFIG_EXEC_PREFIX \${prefix})
-set (_PKG_CONFIG_LIBDIR \${exec_prefix}/lib)
-set (_PKG_CONFIG_INCLUDEDIR \${prefix}/include)
+set (_PKG_CONFIG_LIBDIR \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
+set (_PKG_CONFIG_INCLUDEDIR \${prefix}/${CMAKE_INSTALL_INCLUDEDIR})
 set (_PKG_CONFIG_LIBNAME "${HDF5_CPP_LIB_CORENAME}")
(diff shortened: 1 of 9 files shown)


byrnHDF:
Need to determine how these changes will affect none linux platforms first.

c++/src/CMakeLists.txt
```diff
@@ -173,8 +173,8 @@ endif ()
 #-----------------------------------------------------------------------------
 set (_PKG_CONFIG_PREFIX ${CMAKE_INSTALL_PREFIX})
 set (_PKG_CONFIG_EXEC_PREFIX \${prefix})
-set (_PKG_CONFIG_LIBDIR \${exec_prefix}/lib)
-set (_PKG_CONFIG_INCLUDEDIR \${prefix}/include)
+set (_PKG_CONFIG_LIBDIR \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
```
byrnHDF, line 176: I like this change to use a variable instead of hard-coded dir names. But we already have variables defined in the macro that should be used.

byrnHDF, line 176: I like this change to use a variable instead of hard-coded dir names.

byrnHDF:
Also will need a new configuration Release.txt entry.

byrnHDF (changes requested):
The changes need to go into the HDFMacros file

byrnHDF:
Investigating again, because of history and not wanting to break other projects that rely on certain settings, I propose we add a build option "UseGNUInstallDirs" that calls the module and then sets the HDF5_INSTALL_* correctly.
This means changing all un-prefixed settings to use the appropriate HDF5_INSTALL_* variable.

byrnHDF:
See #2175 for reworked implementation - uses the option "HDF5_USE_GNU_DIRS". The option is OFF by default to not interfere with existing applications that depend on the installation dir structure.

thesamesam:
Fantastic -- I'd missed your earlier comments but I really appreciate you finishing this off! I'll close in favour of yours.

PR #1718, opened 2022-05-03
Hdf5 1 12 Miscellaneous warnings fixes

jhendersonHDF:
Also cherry picks a few relevant commits from develop.

### DIFF
--- a/hl/src/H5PT.c
+++ b/hl/src/H5PT.c
@@ -29,7 +29,7 @@ static H5I_type_t H5PT_ptable_id_type = H5I_UNINIT;
 #define H5PT_HASH_TABLE_SIZE 64
 
 /* Packet Table private functions */
-static herr_t H5PT_free_id(void *id, void **_ctx);
+static herr_t H5PT_free_id(void *id);
 static herr_t H5PT_close(htbl_t *table);
 static herr_t H5PT_create_index(htbl_t *table_id);
 static herr_t H5PT_set_index(htbl_t *table_id, hsize_t pt_index);
@@ -87,8 +87,7 @@ H5PTcreate(hid_t loc_id, const char *dset_name, hid_t dtype_id, hsize_t chunk_si
 
     /* Register the packet table ID type if this is the first table created */
     if (H5PT_ptable_id_type < 0)
-        if ((H5PT_ptable_id_type =
-                 H5Iregister_type((size_t)H5PT_HASH_TABLE_SIZE, 0, (H5I_free_t)H5PT_free_id)) < 0)
+        if ((H5PT_ptable_id_type = H5Iregister_type((size_t)H5PT_HASH_TABLE_SIZE, 0, H5PT_free_id)) < 0)
             goto error;
 
     /* Get memory for the table identifier */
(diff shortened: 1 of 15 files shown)


hl/src/H5PT.c
```diff
@@ -29,7 +29,7 @@ static H5I_type_t H5PT_ptable_id_type = H5I_UNINIT;
 #define H5PT_HASH_TABLE_SIZE 64
 
 /* Packet Table private functions */
-static herr_t H5PT_free_id(void *id, void **_ctx);
+static herr_t H5PT_free_id(void *id);
```
jhendersonHDF, line 32: The ctx parameter was added to the public H5I_free_t typedef in 1.13 for async operations, but isn't in the 1.12 branch.

src/H5B2int.c
```diff
@@ -52,9 +52,9 @@
 /********************/
 /* Local Prototypes */
 /********************/
-static herr_t H5B2__update_child_flush_depends(H5B2_hdr_t *hdr, unsigned depth,
-                                               const H5B2_node_ptr_t *node_ptrs, unsigned start_idx,
-                                               unsigned end_idx, void *old_parent, void *new_parent);
+static herr_t H5B2__update_child_flush_depends(H5B2_hdr_t *hdr, unsigned depth, H5B2_node_ptr_t *node_ptrs,
+                                               unsigned start_idx, unsigned end_idx, void *old_parent,
+                                               void *new_parent);
```
jhendersonHDF, line 57: Cherry-picked Dana's Btree const fixes

tools/lib/h5tools.c
```diff
@@ -1843,15 +1843,21 @@ render_bin_output(FILE *stream, hid_t container, hid_t tid, void *_mem, hsize_t
                     hid_t        region_id    = H5I_INVALID_HID;
                     hid_t        region_space = H5I_INVALID_HID;
                     H5S_sel_type region_type;
+                    H5R_ref_t    tref;
```
jhendersonHDF, line 1846: Cherry-picked Dave's reference misalignment fix

src/H5trace.c
```diff
@@ -120,8 +120,8 @@ H5_trace(const double *returning, const char *func, const char *type, ...)
     void *            vp                  = NULL;
     FILE *            out                 = H5_debug_g.trace;
     static hbool_t    is_first_invocation = TRUE;
-    H5_timer_t        function_timer      = {{0}, {0}, {0}, FALSE};
-    H5_timevals_t     function_times;
+    H5_timer_t        function_timer;
+    H5_timevals_t     function_times = {0.0, 0.0, 0.0};
```
bmribler, line 124: Hmm, I'm confused.  I thought they need to line up...

src/H5trace.c
```diff
@@ -156,11 +160,10 @@ H5_trace(const double *returning, const char *func, const char *type, ...)
         H5_timer_init(&running_timer);
```
bmribler, line 160: So, in some condition, the timer is initialized again?  Is that ok?

src/H5trace.c
byrnHDF: They would if you swap 123 and 124 because 123 breaks the continuous lineup.

src/H5trace.c
derobins: H5_timer_init() just resets the timer fields.

jhendersonHDF: This one tripped me up when I first read it. Note that there's a `running_timer` and a `function_timer`

src/H5trace.c
derobins: Also, we let the formatter deal with this.

derobins (approved):

PR #1720, opened 2022-05-04
Fixes a bug where t_cache fails due to a string size being too small

derobins:
Recent warning reductions led to an incorrect string size being passed
to h5_fixname, causing the test to silently fail. In addition to
fixing the bug, the test will now fail noisily on setup failures.

### DIFF
--- a/testpar/t_cache.c
+++ b/testpar/t_cache.c
@@ -7087,8 +7087,8 @@ finish:
     MPI_Barrier(MPI_COMM_WORLD);
     if (MAINPROCESS) { /* only process 0 reports */
         HDprintf("===================================\n");
-        if (failures) {
-            HDprintf("***metadata cache tests detected %d failures***\n", failures);
+        if (nerrors || failures) {
+            HDprintf("***metadata cache tests detected %d failures***\n", nerrors + failures);
         }
         else {
             HDprintf("metadata cache tests finished with no failures\n");
@@ -7105,5 +7105,5 @@ finish:
     MPI_Finalize();
 
     /* cannot just return (failures) because exit code is limited to 1byte */
-    return (failures != 0);
+    return (nerrors != 0 || failures != 0);
 }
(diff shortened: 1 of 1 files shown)


testpar/t_cache.c
```diff
@@ -1124,6 +1124,8 @@ setup_derived_types(void)
     MPI_Aint      displs[9];
     struct mssg_t sample; /* used to compute displacements */
 
+    HDmemset(&sample, 0, sizeof(struct mssg_t));
```
derobins, line 1127: Cleans a "maybe ininitialized" warning

testpar/t_cache.c
```diff
@@ -6981,7 +6983,7 @@ main(int argc, char **argv)
 
     /* fix the file names */
     for (u = 0; u < sizeof(FILENAME) / sizeof(FILENAME[0]) - 1; ++u) {
-        if (h5_fixname(FILENAME[u], fapl, filenames[u], sizeof(filenames[u])) == NULL) {
+        if (h5_fixname(FILENAME[u], fapl, filenames[u], PATH_MAX) == NULL) {
```
derobins, line 6986: I checked and all other similar changes in the PR that introduced the bug were switched to use PATH_MAX

testpar/t_cache.c
```diff
@@ -7103,5 +7105,5 @@ main(int argc, char **argv)
     MPI_Finalize();
 
     /* cannot just return (failures) because exit code is limited to 1byte */
-    return (failures != 0);
+    return (nerrors != 0 || failures != 0);
```
derobins, line 7108: This is a little weird, but setup errors and test errors are tracked separately.

jhendersonHDF (approved):
Nice catch! I believe I must have mismerged this file during the course of cleaning up warnings among different branches.

soumagne (approved):

PR #1724, opened 2022-05-04
Update to new multi dataset Fortran API and tests.

brtnfld:
Sync Fortran with develop.


epourmal (approved):
Looks good!

brtnfld:
@fortnern for the multidataset branch: In one of the Fortran parallel tests, it checks that a contiguous dataset was collective MPI I/O with h5pget_mpio_actual_mpio. But it is returning that it did chunk_collective (2) instead of contig_coll. (4). This test has nothing to do with multi-dataset.

brtnfld:
@fortnern  I skipped the test until a solution to the missing contiguous collective path is addressed.

PR #1727, opened 2022-05-05
Fix a few Clang sanitizer warnings

jhendersonHDF:
Noticed while doing other testing

### DIFF
--- a/src/H5Tconv.c
+++ b/src/H5Tconv.c
@@ -3266,9 +3266,10 @@ H5T__conv_vlen(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts, si
                     /* If we're down to the last few elements, just wrap up */
                     /* with a "real" reverse copy */
                     if (safe < 2) {
-                        s        = (uint8_t *)buf + (nelmts - 1) * (size_t)s_stride;
-                        d        = (uint8_t *)buf + (nelmts - 1) * (size_t)d_stride;
-                        b        = (uint8_t *)bkg + (nelmts - 1) * (size_t)b_stride;
+                        s = (uint8_t *)buf + (nelmts - 1) * (size_t)s_stride;
+                        d = (uint8_t *)buf + (nelmts - 1) * (size_t)d_stride;
+                        if (bkg)
+                            b = (uint8_t *)bkg + (nelmts - 1) * (size_t)b_stride;
                         s_stride = -s_stride;
                         d_stride = -d_stride;
                         b_stride = -b_stride;
@@ -3278,7 +3279,8 @@ H5T__conv_vlen(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts, si
                     else {
                         s = (uint8_t *)buf + (nelmts - safe) * (size_t)s_stride;
                         d = (uint8_t *)buf + (nelmts - safe) * (size_t)d_stride;
-                        b = (uint8_t *)bkg + (nelmts - safe) * (size_t)b_stride;
+                        if (bkg)
+                            b = (uint8_t *)bkg + (nelmts - safe) * (size_t)b_stride;
                     } /* end else */
                 }     /* end if */
                 else {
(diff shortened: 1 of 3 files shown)


src/H5Tconv.c
```diff
@@ -3426,7 +3428,9 @@ H5T__conv_vlen(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts, si
                     /* Advance pointers */
                     s += s_stride;
                     d += d_stride;
-                    b += b_stride;
+
+                    if (b)
+                        b += b_stride;
```
jhendersonHDF, line 3433: Don't do arithmetic on `b` if `b` is a NULL pointer.

src/H5Tconv.c
```diff
@@ -3278,7 +3279,8 @@ H5T__conv_vlen(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts, si
                     else {
                         s = (uint8_t *)buf + (nelmts - safe) * (size_t)s_stride;
                         d = (uint8_t *)buf + (nelmts - safe) * (size_t)d_stride;
-                        b = (uint8_t *)bkg + (nelmts - safe) * (size_t)b_stride;
+                        if (bkg)
+                            b = (uint8_t *)bkg + (nelmts - safe) * (size_t)b_stride;
```
jhendersonHDF, line 3283: Don't do arithmetic on `bkg` if `bkg` is a NULL pointer.

src/H5VLdyn_ops.c
```diff
@@ -166,9 +166,12 @@ H5VL__term_opt_operation(void)
     FUNC_ENTER_PACKAGE_NOERR
 
     /* Iterate over the VOL subclasses */
-    for (subcls = 0; subcls < NELMTS(H5VL_opt_vals_g); subcls++)
-        if (H5VL_opt_ops_g[subcls])
+    for (subcls = 0; subcls < NELMTS(H5VL_opt_vals_g); subcls++) {
+        if (H5VL_opt_ops_g[subcls]) {
             H5SL_destroy(H5VL_opt_ops_g[subcls], H5VL__term_opt_operation_cb, NULL);
+            H5VL_opt_ops_g[subcls] = NULL;
```
jhendersonHDF, line 172: Set dynamic operations skip list array table entries to NULL to avoid a use after free during package termination.

test/dsets.c
```diff
@@ -12514,10 +12514,19 @@ test_bt2_hdr_fd(const char *env_h5_driver, hid_t fapl)
...
     hid_t             msid = -1;
     H5D_chunk_index_t idx_type;
-    const hsize_t     shape[2]    = {8, 8};
-    const hsize_t     maxshape[2] = {H5S_UNLIMITED, H5S_UNLIMITED};
-    const hsize_t     chunk[2]    = {8, 8};
-    const int         buffer[8]   = {0, 1, 2, 3, 4, 5, 6, 7};
+    const hsize_t     shape[2]     = {8, 8};
+    const hsize_t     maxshape[2]  = {H5S_UNLIMITED, H5S_UNLIMITED};
+    const hsize_t     chunk[2]     = {8, 8};
+    const int         buffer[8][8] = {
```
jhendersonHDF: Test was previously writing to an 8x8 dataset with an 8x1 array, underflowing into the following variables and causing a stack overflow.

test/dsets.c
```diff
@@ -15267,7 +15276,7 @@ test_h5s_plist(void)
     /* Attempt to 'OR' block with invalid dimensions into the selection */
     H5E_BEGIN_TRY
     {
-        ret = H5Pset_dataset_io_hyperslab_selection(dxpl_id_copy, 2, H5S_SELECT_OR, &start, &stride, &count,
+        ret = H5Pset_dataset_io_hyperslab_selection(dxpl_id_copy, H5S_MAX_RANK + 1, H5S_SELECT_OR, &start, &stride, &count,
```
jhendersonHDF: Specifying 2 for the dimensionality here caused an overflow since start, stride, count, block are single values, not 2-dimensional arrays. Since the test was trying to 'OR' an invalid block, use H5S_MAX_RANK + 1 for the dimensionality, which should fail early.

byrnHDF (approved):

derobins (approved):

PR #1739, opened 2022-05-06
Minor things noticed while merging to 1.10

derobins:


test/vfd.c
```diff
@@ -5975,7 +5975,7 @@ main(void)
      */
     env_h5_drvr = HDgetenv(HDF5_DRIVER);
     if (env_h5_drvr) {
-        HDprintf(" -- SKIPPED VFD tests because %s is set -- \n", HDF5_DRIVER);
+        HDprintf(" -- SKIPPED VFD tests because %s is set -- \n", env_h5_drvr);
```
derobins: Probably want the actual environment variable string instead of "HDF5_DRIVER" here.

derobins:
A lot of this was just fixing mixed tabs and spaces

test/vfd.c
jhendersonHDF: HDF5_DRIVER IS the actual environment string. The message was meant to tell the user that "HDF5_DRIVER" is set, not what the value was

jhendersonHDF (changes requested):
HDF5_DRIVER is a macro that was added which maps to "HDF5_DRIVER" in case it needs to be changed in the future.

derobins:
> HDF5_DRIVER is a macro that was added which maps to "HDF5_DRIVER" in case it needs to be changed in the future.

Ah, I see what you were going for. I'll change that back.

test/vfd.c
derobins: Fixed

jhendersonHDF (approved):

lrknox (approved):

PR #1760, opened 2022-05-11
Allows parallel + thread-safe combination in Autotools and CMake

derobins:


byrnHDF (approved):

epourmal (commented):
Was this combination tested and where?

derobins:
> Was this combination tested and where?

It's been tested informally for years and the async VOL connector requires it. There's also no obvious reason for this configuration to be unsupported as the parallel and thread-safe features do not interact in the HDF5 library.

We'll update the daily tests to cover thread-safe + parallel and I'll be reworking the GitHub tests in the next week or two to be more comprehensive. That work will include some parallel + thread-safe configurations.

qkoziol (approved):

lrknox (approved):

PR #1797, opened 2022-05-30
CMake: Allow to set try_run result

xantares:
In cross-compiling contexts, this allows to set the different run
error code and outputs per try_run call.
See https://cmake.org/cmake/help/latest/command/try_run.html#behavior-when-cross-compiling


xantares:
closing; hdf5 cannot really be compiled if no emulator is available anyways

PR #1798, opened 2022-06-01
Check that copy of type when inserting inside cmpd is valid

alkino:
Here is a fix for the problem encountered here: https://forum.hdfgroup.org/t/bug-h5tcopy-of-empty-enum/9634/2

Today, the error is arriving when using the compound type for example in H5Tclose in the test.

I add the test a bit randomly, and it is not finish, i was not able to understand really where it should go and how I need to write it.

### DIFF
--- a/src/H5Tcompound.c
+++ b/src/H5Tcompound.c
@@ -473,7 +473,12 @@ H5T__insert(H5T_t *parent, const char *name, size_t offset, const H5T_t *member)
     parent->shared->u.compnd.memb[idx].name   = H5MM_xstrdup(name);
     parent->shared->u.compnd.memb[idx].offset = offset;
     parent->shared->u.compnd.memb[idx].size   = total_size;
-    parent->shared->u.compnd.memb[idx].type   = H5T_copy(member, H5T_COPY_ALL);
+    {
+        H5T_t* t = H5T_copy(member, H5T_COPY_ALL);
+        if (!t)
+            HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "member type cannot be copied")
+        parent->shared->u.compnd.memb[idx].type   = t;
+    }
 
     parent->shared->u.compnd.sorted = H5T_SORT_NONE;
     parent->shared->u.compnd.nmembs++;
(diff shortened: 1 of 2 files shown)


src/H5Tcompound.c
```diff
@@ -473,7 +473,12 @@ H5T__insert(H5T_t *parent, const char *name, size_t offset, const H5T_t *member)
     parent->shared->u.compnd.memb[idx].name   = H5MM_xstrdup(name);
     parent->shared->u.compnd.memb[idx].offset = offset;
     parent->shared->u.compnd.memb[idx].size   = total_size;
-    parent->shared->u.compnd.memb[idx].type   = H5T_copy(member, H5T_COPY_ALL);
+    {
+        H5T_t* t = H5T_copy(member, H5T_COPY_ALL);
```
fortnern, line 477: This variable seems unnecessary here. We could probably replace this block with:

`if (NULL == (parent->shared->u.compnd.memb[idx].type = H5T_copy(member, H5T_COPY_ALL)))
    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "member type cannot be copied")`

test/cmpd_dset.c
```diff
@@ -2210,6 +2210,46 @@ test_ooo_order(char *filename, hid_t fapl_id)
...
+    if ((tenum = H5Tenum_create(H5Tcopy(H5T_NATIVE_INT))) < 0)
+        TEST_ERROR;
+
+    if ((tcmp = H5Tcreate(H5T_COMPOUND, 2 * H5Tget_size(H5T_NATIVE_LONG))) < 0)
+        TEST_ERROR;
+
+    if ((H5Tinsert(tcmp, "long", 0, H5T_NATIVE_LONG)) < 0)
+        TEST_ERROR;
+
+    if ((H5Tinsert(tcmp, "enum", H5Tget_size(H5T_NATIVE_LONG), tenum)) < 0)
```
fortnern, line 2229: Is this call supposed to fail? If so, why does the test pass if it doesn't (and fail if it does)?

alkino: I was not good at understand how to write tests, so I don't think test means something.

src/H5Tcompound.c
alkino: Yes, seems ok

derobins:
This issue is fixed in #3088

PR #1801, opened 2022-06-06
Support for UnifyFS with MPI_File_sync

houjun:
This is a prototype implementation and tested with a couple of simple codes. Link to the [design document](https://docs.google.com/document/d/1IJMxZrEeppLlRcwZnsz0iqHKbk1L8t7xKW56Z9dykrg). 
Main changes are:
- Enable UnifyFS support can be done by checking an MPI hint (will be in a future MPICH release), or setting an environment variable: export HDF5_DO_MPI_FILE_SYNC=1 at file open/create time, and record the result in a flag in the HDF5 file structure.
- When UnifyFS is on, MPI_File_sync will be called after collective write operations to make data visible to all ranks. 
- Will fail and report error for any independent write call when UnifyFS is on.


src/H5FDmpio.c
```diff
@@ -964,6 +965,10 @@ H5FD__mpio_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t H5_ATTR
     file->mpi_rank = mpi_rank;
     file->mpi_size = mpi_size;
 
+    /* Retrieve the flag indicating whether MPI_File_sync is needed after each write */
+    if (H5_mpio_get_file_sync_required(fh, &file->mpi_file_sync_required) < 0)
```
jhendersonHDF, line 969: Can the status for this change between separate file opens? Otherwise, it seems to me like this is something that should be checked only once when the MPI I/O driver is initialized so that the library will get the correct value if it queries for the status before a file has been opened.

src/H5mpi.c
```diff
@@ -782,4 +782,57 @@ H5_mpio_gatherv_alloc_simple(void *send_buf, int send_count, MPI_Datatype send_t
...
+    herr_t   ret_value = SUCCEED;
+    MPI_Info info_used;
+    int      flag;
+    char     value[MPI_MAX_INFO_VAL];
+
+    FUNC_ENTER_NOAPI(FAIL)
+
+    HDassert(file_sync_required);
+
+    *file_sync_required = true;
```
jhendersonHDF: Shouldn't this be initialized to false so that we retain the existing (though as I understand it, incorrect) behavior and only turn on MPI_File_sync when ROMIO reports needing it?

src/H5FDmpio.c
houjun: Yes, the checking should be in each file open as the application may open a file that is not on UnifyFS.

src/H5mpi.c
houjun: @jhendersonHDF Please check the latest commit.

src/H5mpi.c
```diff
@@ -782,4 +782,58 @@ H5_mpio_gatherv_alloc_simple(void *send_buf, int send_count, MPI_Datatype send_t
...
+ *
+ * Purpose:     Retrieve the MPI hint indicating whether the data written
+ *              by the MPI ROMIO driver is immediately visible to all MPI
+ *              ranks.
+ *
+ * Notes:       This routine is designed for supporting UnifyFS that needs
+ *              MPI_File_sync in order to make the written data available
+ *              to all ranks.
+ *
+ * Return:      True when MPI_File_sync is not needed/False otherwise
```
derobins: The return value is SUCCEED/FAIL. The out hbool_t parameter is set to TRUE/FALSE.

src/H5mpi.c
```diff
@@ -782,4 +782,58 @@ H5_mpio_gatherv_alloc_simple(void *send_buf, int send_count, MPI_Datatype send_t
...
+        HGOTO_ERROR(H5E_LIB, H5E_CANTGET, FAIL, "can't get MPI info")
+
+    if (flag && !HDstrcmp(value, "false"))
+        *file_sync_required = TRUE;
+
+    if (MPI_SUCCESS != MPI_Info_free(&info_used))
+        HGOTO_ERROR(H5E_LIB, H5E_CANTFREE, FAIL, "can't free MPI info")
+
+    /* Force setting the flag via env variable (temp solution before the flag is implemented in MPI) */
+    char *sync_env_var = HDgetenv("HDF5_DO_MPI_FILE_SYNC");
```
derobins, line 830: We put all variable declarations at the top of the function

src/H5mpi.c
```diff
@@ -782,4 +782,58 @@ H5_mpio_gatherv_alloc_simple(void *send_buf, int send_count, MPI_Datatype send_t
...
+ * Return:      True when MPI_File_sync is not needed/False otherwise
+ *
+ * Programmer:  Houjun Tang,  April 7, 2022
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5_mpio_get_file_sync_required(MPI_File fh, hbool_t *file_sync_required)
+{
+    herr_t   ret_value = SUCCEED;
```
derobins: Usually the ret_value variable is the last thing to be declared in the variable block

src/H5Fprivate.h
```diff
@@ -974,6 +974,7 @@ H5_DLL herr_t   H5F_mpi_retrieve_comm(hid_t loc_id, hid_t acspl_id, MPI_Comm *mp
 H5_DLL herr_t  H5F_mpi_get_file_block_type(hbool_t commit, MPI_Datatype *new_type, hbool_t *new_type_derived);
 H5_DLL hbool_t H5F_get_coll_metadata_reads(const H5F_t *f);
 H5_DLL void H5F_set_coll_metadata_reads(H5F_t *f, H5P_coll_md_read_flag_t *file_flag, hbool_t *context_flag);
+H5_DLL herr_t H5F_shared_get_mpi_file_sync_required(const H5F_shared_t *f_sh, hbool_t *flag);
```
derobins, line 977: I'm surprised the parallel shared_get bits aren't a part of the H5Fquery.c scheme

src/H5Fmpi.c
```diff
@@ -588,4 +588,35 @@ H5F_mpi_get_file_block_type(hbool_t commit, MPI_Datatype *new_type, hbool_t *new
...
+    herr_t ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_NOAPI(FAIL)
+
+    HDassert(f_sh);
+    HDassert(flag);
+
+    /* Dispatch to driver */
+    if ((ret_value = H5FD_mpi_get_file_sync_required(f_sh->lf, flag)) < 0)
+        HGOTO_ERROR(H5E_FILE, H5E_CANTGET, (-1), "driver get_file_sync_required request failed")
```
derobins: FAIL, not (-1)

src/H5FDmpi.c
```diff
@@ -252,6 +252,43 @@ H5FD_mpi_haddr_to_MPIOff(haddr_t addr, MPI_Offset *mpi_off /*out*/)
...
+ *              May 19, 2022
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5FD_mpi_get_file_sync_required(H5FD_t *file, hbool_t *file_sync_required)
+{
+    const H5FD_class_t *cls;
+    uint64_t            flags     = H5FD_CTL_FAIL_IF_UNKNOWN_FLAG | H5FD_CTL_ROUTE_TO_TERMINAL_VFD_FLAG;
+    herr_t              ret_value = SUCCEED;
```
derobins: ret_value is usually the last thing in the variable block

src/H5Dio.c
```diff
@@ -830,6 +830,18 @@ H5D__ioinfo_adjust(H5D_io_info_t *io_info, const H5D_t *dset, const H5S_t *file_
             } /* end if */
         }     /* end if */
         else {
+            /* Fail when file sync is required, since it requires collective write */
+            if (io_info->op_type == H5D_IO_OP_WRITE) {
+                hbool_t mpi_file_sync_required = FALSE;
+                if (H5F_shared_get_mpi_file_sync_required(io_info->f_sh, &mpi_file_sync_required) < 0)
+                    HGOTO_ERROR(H5E_FILE, H5E_CANTGET, FAIL, "can't get MPI file_sync_required flag")
```
derobins: H5E_DATASET for the major category

src/H5Dio.c
```diff
@@ -830,6 +830,18 @@ H5D__ioinfo_adjust(H5D_io_info_t *io_info, const H5D_t *dset, const H5S_t *file_
...
+
+                if (mpi_file_sync_required)
+                    HGOTO_ERROR(
+                        H5E_IO, H5E_NO_INDEPENDENT, FAIL,
```
derobins: H5E_DATASET for the major category

houjun: @derobins I have updated the code to address your comments.

src/H5FDmpio.c
```diff
@@ -964,6 +965,10 @@ H5FD__mpio_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t H5_ATTR
...
+        HDONE_ERROR(H5E_VFL, H5E_CANTGET, NULL, "unable to get mpi_file_sync_required hint")
```
jhendersonHDF: You may want this to be a goto_error instead of a done_error, since a failure to get the hint probably means we don't want to continue opening the file anyway.

src/H5FDmpi.c
```diff
@@ -252,6 +252,43 @@ H5FD_mpi_haddr_to_MPIOff(haddr_t addr, MPI_Offset *mpi_off /*out*/)
...
+ * Programmer:	Houjun Tang
+ *              May 19, 2022
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5FD_mpi_get_file_sync_required(H5FD_t *file, hbool_t *file_sync_required)
+{
+    const H5FD_class_t *cls;
+    uint64_t            flags = H5FD_CTL_FAIL_IF_UNKNOWN_FLAG | H5FD_CTL_ROUTE_TO_TERMINAL_VFD_FLAG;
```
jhendersonHDF: Thinking about it more, I'm hesitant to use the "fail if unknown" flag here, because it doesn't seem like something we necessarily want to force every MPI-based VFD to implement. Or at the very least, a failure due to an unknown op code should be gracefully handled here, but that's a little trickier since there isn't a good way to distinguish between that and real failures currently.

houjun: @jhendersonHDF , just changed the previous goto_error and removed the "fail if unknown" flag.

jhendersonHDF (approved):

src/H5mpi.c
```diff
@@ -782,4 +782,58 @@ H5_mpio_gatherv_alloc_simple(void *send_buf, int send_count, MPI_Datatype send_t
...
+
+    HDassert(file_sync_required);
+
+    *file_sync_required = FALSE;
+
+    if (MPI_SUCCESS != MPI_File_get_info(fh, &info_used))
+        HGOTO_ERROR(H5E_LIB, H5E_CANTGET, FAIL, "can't get MPI info")
+
+    if (MPI_SUCCESS !=
+        MPI_Info_get(info_used, "romio_visibility_immediate", MPI_MAX_INFO_VAL - 1, value, &flag))
```
ax3l, line 820: I think this flag leads now to the effect that NFS writes are also enabling the file sync, which crashes independent I/O:
https://github.com/openPMD/openPMD-api/issues/1423

PR #1803, opened 2022-06-07
Fix bugs in parallel selection I/O

fortnern:


fortnern:
This will be tested by the multi dataset tests

epourmal (approved):
Looks very good to me. Question: why we cannot use `size_t` for `count`?

jhendersonHDF (approved):

qkoziol (approved):

fortnern:
> Looks very good to me. Question: why we cannot use `size_t` for `count`?

I think it's because MPI_Type_create_struct(), etc. take an int for the count. I'm not sure why it's uin32_t instead of int though.

epourmal:
> > Looks very good to me. Question: why we cannot use `size_t` for `count`?
> 
> I think it's because MPI_Type_create_struct(), etc. take an int for the count. I'm not sure why it's uin32_t instead of int though.

If it comes from `MPI_Type_create_struct()`, we should use `int`. My guess about `uin32_t `is that someone wanted to enforce unsigned 32-bit value, but if MPI needs `int,` we should use `int`. There is not guarantee, that `int `is a 32-bit number, right? If we need to check the range to avoid MPI bugs or something like this, it should be done explicitly and not by using C datatype. But this is my "non-programmer" opinion :-)

PR #1812, opened 2022-06-14
Various warning fixes

seanm:


seanm:
eed5d37 is WIP. Is there a canonical mechanism in HDF5 to rename public constants and still provide old spelling?

release_docs/HISTORY-1_13.txt
```diff
@@ -538,7 +538,7 @@ New Features
       certain operations.
 
       These changes also introduce a new "ctl" callback
-      operation identified by the H5FD_CTL__MEM_COPY op code.
+      operation identified by the H5FD_CTL_MEM_COPY op code.
```
jhendersonHDF: I was under the impression that in C99 reserved identifiers were just those with leading underscores. Why are double underscores problematic in the middle of a macro name?

seanm: They are [disallowed in C++](https://en.cppreference.com/w/c/language/identifier):

```
In file included from /Users/sean/external/hdf5/c++/src/H5CommonFG.cpp:16:
In file included from /Users/sean/external/hdf5/c++/src/H5Include.h:15:
In file included from /Users/sean/external/hdf5/src/hdf5.h:29:
/Users/sean/external/hdf5/src/H5FDpublic.h:191:9: warning: macro name is a reserved identifier [-Wreserved-id-macro]
#define H5FD_CTL__INVALID_OPCODE              0
        ^
```

jhendersonHDF: I see. This may be a bit trickier to handle since the macro names are already public, as you mentioned. Also, I believe that we stopped supporting compiling the C library with a C++ compiler a while ago due to some problems we were having in continuing to support both. That would be why this wasn't immediately seen at least.

Since the macros are public and the C++ wrappers are including them, we might need to work around the issue in a different way than just renaming them.

seanm: I kept the old spellings in the header file too.  Ideally the `#if 0` I have there for now would change to something allowing the old names to be defined or not. Other projects I work with do something like:

```
#if COMPILE_LEGACY_STUFF
#define OldSpelling NewSpelling
#endif
```

Does HDF5 have a similar mechanism?

jhendersonHDF: The closest I can think of would be the H5_NO_DEPRECATED_SYMBOLS approach that we use for removing or compiling deprecated APIs, where H5_NO_DEPRECATED_SYMBOLS being defined means that we _don't_ compile the old functionality. However, changing a public symbol like this is something we'd typically avoid doing until the next major release of HDF5 (in this case, HDF5 1.14 since these symbols are in the 1.13 releases).

seanm: I tweaked my commit accordingly.

Other projects I work with have something like `H5_NO_DEPRECATED_SYMBOLS` but also something like `H5_FUTURE_NO_DEPRECATED_SYMBOLS`.  Then when a major release comes out, there's a mass rename of `H5_FUTURE_NO_DEPRECATED_SYMBOLS` to `H5_NO_DEPRECATED_SYMBOLS`.

config/cmake_ext_mod/HDFTests.c
```diff
@@ -82,12 +82,6 @@ SIMPLE_TEST(timezone=0);
 
 #ifdef PRINTF_LL_WIDTH
```
derobins: Now that Microsoft supports C99 and HDF5 requires it, this entire test is useless.

hl/c++/src/H5PacketTable.h
```diff
@@ -163,7 +163,7 @@ class H5_HLCPPDLL FL_PacketTable : virtual public PacketTable {
     /* Destructor
      * Cleans up the packet table
      */
-    virtual ~FL_PacketTable()
+    ~FL_PacketTable() override
```
derobins, line 166: This is in a public header and should get a note in RELEASE.txt

release_docs/HISTORY-1_13.txt
derobins: Regardless, this is a history document and the change should be reverted in this file.

src/H5FDpublic.h
```diff
@@ -249,9 +249,23 @@
...
+#define H5FD_CTL__TEST_OPCODE                 H5FD_CTL_TEST_OPCODE
+#define H5FD_CTL__GET_MPI_COMMUNICATOR_OPCODE H5FD_CTL_GET_MPI_COMMUNICATOR_OPCODE
+#define H5FD_CTL__GET_MPI_RANK_OPCODE         H5FD_CTL_GET_MPI_RANK_OPCODE
+#define H5FD_CTL__GET_MPI_SIZE_OPCODE         H5FD_CTL_GET_MPI_SIZE_OPCODE
+#define H5FD_CTL__MEM_ALLOC                   H5FD_CTL_MEM_ALLOC
+#define H5FD_CTL__MEM_FREE                    H5FD_CTL_MEM_FREE
+#define H5FD_CTL__MEM_COPY                    H5FD_CTL_MEM_COPY
+#define H5FD_CTL__FAIL_IF_UNKNOWN_FLAG        H5FD_CTL_FAIL_IF_UNKNOWN_FLAG
+#define H5FD_CTL__ROUTE_TO_TERMINAL_VFD_FLAG  H5FD_CTL_ROUTE_TO_TERMINAL_VFD_FLAG
+#endif
```
derobins: You can just do the rename and remove the double underscore compatibility versions you've defined here. HDF5 1.13.x is an experimental release and we don't promise compatibility between releases. The ctl callback is 1.13.x only, so we don't have to worry about 1.12, etc.

As Jordan commented earlier, we don't support compiling the library with a C++ compiler anymore, but this file will be consumed by anyone consuming the library from C++, so we do need to be mindful of C++ issues in the public headers.

derobins (changes requested):
Looks great, but a couple of things:

* Revert the history file (there's no need to update past text files)
* Make a small note in RELEASE.txt about the packet table change
* Remove the compatibility defines from H5FDdevelop.h (no need to keep that around)
* Probably also want to add a RELEASE.txt note for the H5FDdevelop.h symbol changes

derobins:
Also, removing the LL checks entirely from CMake would be nice. No need keeping pre-c99 stuff around. If you don't have time, I'll do it after this is merged.

src/H5FDpublic.h
seanm: Done.

config/cmake_ext_mod/HDFTests.c
seanm: Done. Hopefully I removed the correct code, am not 100% I got enough/too much.

release_docs/HISTORY-1_13.txt
seanm: Reverted history change. Added small note in RELEASE.txt.

hl/c++/src/H5PacketTable.h
seanm: What would you want to say?  This doesn't affect API or codegen.

seanm:
> Also, removing the LL checks entirely from CMake would be nice. No need keeping pre-c99 stuff around. If you don't have time, I'll do it after this is merged.

Done I think, but not certain if I got too much/too little.

derobins:
> > Also, removing the LL checks entirely from CMake would be nice. No need keeping pre-c99 stuff around. If you don't have time, I'll do it after this is merged.
> 
> Done I think, but not certain if I got too much/too little.

Looks like there's a bunch of stuff that uses H5_PRINTF_LL_WIDTH

tools/src/h5perf/sio_engine.c
tools/src/h5perf/sio_perf.c
tools/src/h5perf/pio_perf.c
tools/src/h5perf/pio_engine.c
tools/lib/h5tools_str.c
tools/lib/h5diff_array.c
test/dt_arith.c
java/src/jni/h5util.c

*sigh*

Most of the changes are trivial, though. You just have to plug "ll" in to couple places.

derobins:
There's remaining __int64 bits in the Windows code. I'll look into removing that in a later PR. It's only used on Windows, where __int64 always exists, so the CMake bits you removed are unnecessary (and unused).

seanm:
>Most of the changes are trivial, though. You just have to plug "ll" in to couple places.

Done.

>There's remaining __int64 bits in the Windows code

I had left that deliberately thinking it more literally matches the Windows API declarations.

derobins (approved):

derobins:
Thanks for the changes!

seanm:
Thanks for the review & merge!

seanm:
I'm curious, why do you squash everything into one commit?  I had deliberately kept things grouped into logically separated commits...

PR #1815, opened 2022-06-22
sprintf to snprintf

seanm:


jhendersonHDF (approved):
A good improvement!

derobins (approved):

seanm:
Note that this patch did not get rid of every last sprintf, there are still a dozen or so left that were too tricky for me to confidently change, as I don't know the codebase well.

I'd appreciate if someone else could purge the remaining ones though... shall I create an issue for that?

seanm:
So actually there are 0 uses of sprintf anymore, but there are 16 uses of HDsprintf left.

jhendersonHDF:
> Note that this patch did not get rid of every last sprintf, there are still a dozen or so left that were too tricky for me to confidently change, as I don't know the codebase well.
> 
> I'd appreciate if someone else could purge the remaining ones though... shall I create an issue for that?

If you'd like to create an issue for this, I'll assign it to myself and take a look when I get a chance.

PR #1819, opened 2022-06-22
Removes VERIFY/CHECK macros from the chunk_info test

derobins:
Mixing testhdf5 and h5test macros is a bad idea. Also quiets a couple
of MSVC warnings.

### DIFF
--- a/test/chunk_info.c
+++ b/test/chunk_info.c
@@ -179,11 +179,18 @@ verify_get_chunk_info(hid_t dset, hid_t dspace, hsize_t chk_index, hsize_t exp_c
 
     if (H5Dget_chunk_info(dset, dspace, chk_index, out_offset, &read_flt_msk, &addr, &size) < 0)
         TEST_ERROR;
-    CHECK(addr, HADDR_UNDEF, "H5Dget_chunk_info");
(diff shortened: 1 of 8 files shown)


derobins:
Didn't mean to push all those at once. Fat-fingered commit -a...

test/chunk_info.c
```diff
@@ -179,11 +179,18 @@ verify_get_chunk_info(hid_t dset, hid_t dspace, hsize_t chk_index, hsize_t exp_c
...
+        FAIL_PUTS_ERROR("address cannot be HADDR_UNDEF");
+    if (size != exp_chk_size)
+        FAIL_PUTS_ERROR("unexpected chunk size");
+    if (read_flt_msk != exp_flt_msk)
+        FAIL_PUTS_ERROR("unexpected filter mask");
+    if (out_offset[0] != exp_offset[0])
+        FAIL_PUTS_ERROR("unexpected offset[0]");
+    if (out_offset[1] != exp_offset[1])
+        FAIL_PUTS_ERROR("unexpected offset[1]");
+
```
bmribler, line 193: What is the reason for this change?  We've been using VERIFY for years and many places, haven't we?  Can VERIFY be modified instead?

PR #1822, opened 2022-06-22
Set a constant in H5SM to UINT_MAX instead of -1U

derobins:
MSVC complains about using 80s tricks instead of the appropriate integer constant

`30>C:\Users\Dana Robinson\Desktop\hdf5\hdf5\src\H5SM.c(1079,30): warning C4146: unary minus operator applied to unsigned type, result still unsigned`


jhendersonHDF (approved):

fortnern (approved):
Kind of pedantic of Windows, but fine I suppose

PR #1832, opened 2022-06-29
Fixes a few minor parallel warnings

derobins:


jhendersonHDF (approved):

byrnHDF (approved):

src/H5Dmpio.c
```diff
@@ -4068,10 +4068,18 @@ H5D__mpio_share_chunk_modification_data(H5D_filtered_collective_io_info_t *chunk
...
 
-            /* Determine if all send requests have completed */
+            /* Determine if all send requests have completed
+             *
+             * gcc 11 complains about passing MPI_STATUSES_IGNORE as an MPI_Status
+             * array. See the discussion here:
+             *
+             * https://github.com/pmodels/mpich/issues/5687
+             */
+            H5_GCC_DIAG_OFF("stringop-overflow")
```
derobins, line 4078: The MPICH people defined MPI_STATUSES_IGNORE to be 1, which gcc doesn't recognize as a "special" value so it does size checking on it and complains. This is only an MPICH + "gcc of a certain vintage" issue. Suppressing the warning is useful, though, as the benign code looks like a potential memory access issue.

PR #1833, opened 2022-06-29
Adds a -Werror debug Autotools test and updates build modes

derobins:
The Autotools build modes didn't look like they were doing the right thing, so I updated that while I was in there.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
 #      fail-fast: false
       matrix:
-        name: ["Windows Latest MSVC", "Ubuntu Latest GCC", "Ubuntu Debug GCC", "macOS Latest Clang", "Ubuntu Autotools GCC", "Windows TS MSVC", "Ubuntu TS GCC", "TS Debug GCC", "macOS TS Clang", "TS Autotools GCC"]
+        name: ["Windows Latest MSVC", "Ubuntu Latest GCC", "Ubuntu Debug GCC", "macOS Latest Clang", "Ubuntu Autotools GCC", "-Werror Ubuntu Debug Autotools GCC", "Windows TS MSVC", "Ubuntu TS GCC", "TS Debug GCC", "macOS TS Clang", "TS Autotools GCC"]
         include:
           - name: "Windows Latest MSVC"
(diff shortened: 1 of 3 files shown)


lrknox (approved):

derobins:
There are some gcc 12 warnings that I was not seeing on my gcc 11 VM that cause failures with -Werror. I'll submit a separate PR with gcc 12 warning fixes and merge this PR after that one goes in.

.github/workflows/pr-check.yml
```diff
@@ -81,6 +86,21 @@ jobs:
             parallel: disable
             toolchain: ""
             generator: "autogen"
+            flags: ""
+          # Build that ensures no new warnings sneak into the library.
+          - name: "-Werror Ubuntu Debug Autotools GCC"
+            artifact: "LinuxDBG.tar.xz"
+            os: ubuntu-latest
+            build_type: "debug"
```
byrnHDF, line 107: Capital "D" - "Debug"

byrnHDF (approved):

.github/workflows/pr-check.yml
derobins: Not in the Autotools. We need --enable-build-mode=debug.

derobins:
This PR should NOT get merged elsewhere. Only develop has been made warning-free.

gnuoyd:
Re: commit [5b56fb3](https://github.com/HDFGroup/hdf5/pull/1833/commits/5b56fb362aca39b24095d38771ba9503b416cbc5), it looks like you may not be using `--enable-warnings-as-errors`. That will turn on the errors promotion without breaking `configure`.

configure.ac
```diff
@@ -181,6 +183,14 @@ saved_user_JAVAFLAGS="$JAVAFLAGS"
 saved_user_LDFLAGS="$LDFLAGS"
 saved_user_CPPFLAGS="$CPPFLAGS"
 
+## Strip out -Werror from CFLAGS since that can cause checks to fail when
+## compiling the test program fails due to warnings
+CFLAGS="`echo $CFLAGS | sed -e 's/-Werror//g'`"
```
opoplawski, line 188: This breaks compilation on Fedora which uses `-Werror=format-security`

PR #1843, opened 2022-07-01
Adds a parallel Autotools build to the GitHub actions

derobins:


derobins:
Okay, this functions now but takes forever to run as you only get two processors. I'm probably going to close this until we have a way of severely restricting what runs in testpar.

PR #1862, opened 2022-07-08
Persist dataset access properties when using H5Fstart_swmr_write

fortnern:
This fix addresses an issue where using H5Fstart_swmr_write would wipe out the access properties for all open datasets. This PR modifies H5Fstart_swmr_write to cache the DAPL ids for all open datasets then use these property lists when closing/reopening the datasets for SWMR reasons.

### DIFF
--- a/src/H5Dint.c
+++ b/src/H5Dint.c
@@ -3703,9 +3703,11 @@ H5D_get_access_plist(const H5D_t *dset)
 {
     H5P_genplist_t *   old_plist;                   /* Stored DAPL from dset */
     H5P_genplist_t *   new_plist;                   /* New DAPL */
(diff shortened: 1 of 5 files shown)


qkoziol (approved):
Library change looks fine, but would be nice if it had a regression test also.

fortnern:
Added a regression test, and fixed some bugs exposed by the test. Quincey, do you know why H5Dget_access_plist() (and similar functions) works the way it does, where it copies the DAPL used to open the dataset, then goes over all the properties one by one and retrieves them from the dataset struct? Seems redundant to me.

fortnern:
Github tests are hung. I ran the code tests manually and it seems to pass, as did codespell.

derobins (approved):

derobins:
Oops. Clicked wrong button.

PR #1864, opened 2022-07-09
Onion VFD

derobins:


derobins:
This still has warnings so the -Werror check will fail until I fix that over the weekend.

src/H5FDonion_header.c
```diff
@@ -0,0 +1,231 @@
...
+ *
+ *              Implementation must correspond with
+ *              H5FD__onion_header_encode().
+ *
+ * Return:      Success:    Number of bytes read from buffer
+ *              Failure:    0
+ *-----------------------------------------------------------------------------
+ */
+size_t
+H5FD__onion_header_decode(unsigned char *buf, H5FD_onion_header_t *header)
```
jhendersonHDF, line 121: It seems unwise not to pass the size of the buffer in here, since decoding of corrupt file contents is usually where our CVEs stem from.

src/H5FDonion_header.c
```diff
@@ -0,0 +1,231 @@
...
+ *              The destination buffer must be sufficiently large to hold the
+ *              encoded contents (H5FD__ONION_ENCODED_SIZE_HEADER).
+ *
+ * Return:      Number of bytes written to buffer.
+ *              The checksum of the generated buffer contents (excluding the
+ *              checksum itself) is stored in the pointer `checksum`).
+ *-----------------------------------------------------------------------------
+ */
+size_t
+H5FD__onion_header_encode(H5FD_onion_header_t *header, unsigned char *buf, uint32_t *checksum /*out*/)
```
jhendersonHDF, line 203: Same on not knowing the size of the buffer here.

src/H5FDonion_history.c
```diff
@@ -0,0 +1,275 @@
...
+    if (history_out->n_revisions > 0 && NULL == history_out->record_locs)
+        HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "can't allocate record pointer list")
+
+    if (H5FD__onion_history_decode(buf, history_out) != size)
+        HGOTO_ERROR(H5E_VFL, H5E_CANTDECODE, FAIL, "can't decode history (final)")
+
+done:
+    H5MM_xfree(buf);
+    if (ret_value == FAIL)
+        H5MM_xfree(history_out->record_locs);
```
jhendersonHDF, line 75: This line can cause a crash either:

- if history_out is NULL and we fail
- if history_out was partially initialized and we try to free record_locs when it isn't NULL and not allocated

src/H5FDonion_history.c
```diff
@@ -0,0 +1,275 @@
...
+ *              Then the decode operation is called a second time, and all
+ *              components will be populated (and again number of bytes read is
+ *              returned).
+ *
+ * Return:      Success:    Number of bytes read from buffer
+ *              Failure:    0
+ *-----------------------------------------------------------------------------
+ */
+size_t
+H5FD__onion_history_decode(unsigned char *buf, H5FD_onion_history_t *history)
```
jhendersonHDF, line 150: Same old comment about buffer size for decode and encode

src/H5FDonion_history.c
```diff
@@ -0,0 +1,275 @@
...
+    unsigned char *ptr         = NULL;
+    size_t         ret_value   = 0;
+
+    FUNC_ENTER_PACKAGE;
+
+    HDassert(buf != NULL);
+    HDassert(history != NULL);
+    HDassert(H5FD__ONION_HISTORY_VERSION_CURR == history->version);
+
+    if (HDstrncmp((const char *)buf, "OWHS", 4))
```
jhendersonHDF, line 166: This should probably be H5FD__ONION_HISTORY_SIGNATURE instead of a "magic" string

src/H5FDonion_index.c
```diff
@@ -0,0 +1,935 @@
...
+    if (aix->n_entries == 0 || logical_page > aix->list[high].logical_page ||
+        logical_page < aix->list[0].logical_page)
+        HGOTO_DONE(0)
+
+    /*
+     * Binary search on sorted list
+     */
+
+    /* Winnow down to first of found or one element */
+    while (range > 0) {
```
jhendersonHDF, line 251: Seems we could replace some or all of these binary searches with bsearch, but that's just a nitpick

tools/lib/h5diff.c
```diff
@@ -648,8 +648,15 @@ h5diff(const char *fname1, const char *fname2, const char *objname1, const char
...
+    if (opts->vfd_info[0].u.name) {
+        if ((fapl1_id = h5tools_get_fapl(H5P_DEFAULT, NULL, &(opts->vfd_info[0]))) < 0) {
+            parallel_print("h5diff: unable to create fapl for input file\n");
+            H5TOOLS_GOTO_ERROR(H5DIFF_ERR, "unable to create input fapl\n");
+        }
+    }
+
     if (opts->custom_vol[0] || opts->custom_vfd[0]) {
-        if ((fapl1_id = h5tools_get_fapl(H5P_DEFAULT, opts->custom_vol[0] ? &(opts->vol_info[0]) : NULL,
+        if ((fapl1_id = h5tools_get_fapl(fapl1_id, opts->custom_vol[0] ? &(opts->vol_info[0]) : NULL,
```
jhendersonHDF, line 659: I'm not sure what the reason for this and the below change is, but this is going to copy fapl1_id and assign it back to itself, losing the original reference. Same for fapl2_id below

tools/src/h5diff/h5diff_common.c
```diff
@@ -56,6 +56,28 @@ static struct h5_long_options l_opts[] = {{"help", no_arg, 'h'},
                                           {"vfd-info-2", require_arg, 'Z'},
                                           {NULL, 0, '\0'}};
 
+static H5FD_onion_fapl_info_t onion_fa_g_1 = {
```
jhendersonHDF, line 59: I've probably mentioned this before, but definitely the sooner we can move the VFD-specific info out of the tools, the better maintenance-wise.

byrnHDF:
h5diff, h5dump, h5repack tests not added to the CMakeTests.cmake files

derobins:
Holy crap git has made a giant mess of this branch. Will redo and recreate the PR.

PR #1865, opened 2022-07-09
Updates the patterns in Doxyfile.in to accept H5*develop.h

derobins:


gheber (approved):
How many other `H5*develop.h` headers do we anticipate? What's the rationale? Is that documented somewhere?

byrnHDF (approved):

lrknox (approved):

PR #1871, opened 2022-07-10
Adds the mirror and direct VFDs to GitHub actions

derobins:
Both Autotools and CMake. Also turns on C++ for the -Werror check.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,6 +34,8 @@ jobs:
             ts: OFF
             hl: ON
             parallel: OFF
+            mirror_vfd: OFF
+            direct_vfd: OFF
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -106,12 +118,14 @@ jobs:
...
             build_type: "debug"
-            cpp: disable
+            cpp: enable
             fortran: disable
             java: disable
             ts: disable
             hl: enable
             parallel: disable
+            mirror_vfd: disable
+            direct_vfd: enable
```
derobins, line 128: The mirror VFD still spits out warnings, so it can't be enabled here until I fix that.

byrnHDF (approved):

lrknox (approved):

PR #1872, opened 2022-07-10
Reduce -Wstrict-overflow= from 5 to 2

derobins:
The signal-to-noise ratio of the higher warning level is very low
and the noise obscures things we should fix

gcc docs here:

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

### DIFF
--- a/config/gnu-warnings/4.8
+++ b/config/gnu-warnings/4.8
@@ -8,7 +8,7 @@
 -Wsync-nand
 
 # warning flag added for GCC >= 4.5
--Wstrict-overflow=5
+-Wstrict-overflow=2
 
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

lrknox (approved):

qkoziol (commented):
Why go all the way down to 2?   Why not go to 4 and fix those?

derobins:
> Why go all the way down to 2? Why not go to 4 and fix those?

Everything greater than 2 is noise. We can set a different level in the developer warnings, but it's almost entirely worthless.

PR #1874, opened 2022-07-11
Fixes for warnings identified by MSVC

derobins:


byrnHDF (approved):

jhendersonHDF (approved):

src/H5B2internal.c
```diff
@@ -636,24 +636,27 @@ H5B2__update_internal(H5B2_hdr_t *hdr, uint16_t depth, unsigned *parent_cache_in
 
                     if (idx == 0) { /* Left-most child */
                         /* Check for left-most child and its neighbor being close to full */
-                        if ((internal->node_ptrs[idx].node_nrec + internal->node_ptrs[idx + 1].node_nrec) >=
+                        if ((int)(internal->node_ptrs[idx].node_nrec +
```
qkoziol: I believe these checks would be better the other way: casting the right side to unsigned

qkoziol (changes requested):
Suggest switching the cast for the H5B2 changes

src/H5B2internal.c
qkoziol: (All the values are positive)

derobins: I was worried more about casting -1 to an unsigned very large number than casting a small unsigned to an int, which has a zero percent chance of badness.

derobins: And for this reason, I disagree with the requested change.

qkoziol: split_nrec will be >0, so 2*(>0)-1 will always be greater than zero

qkoziol: Except for overflow scenarios, but then that would be true for casting node_nrec to an int also.  "Pick your badness" :-/

derobins: Right. To-MAY-to / to-MAH-to. Especially since we're trying to guess which highly unlikely edge case (or future change + edge case) will be a problem.

PR #1876, opened 2022-07-11
H5Oflush fails for parallel

raylu-hdf:
H5Oflush causes H5Fclose to trigger an assertion failure in metadata cache for parallel.  This commit makes sure H5Oflush fails for parallel until this problem is solved in the future.

### DIFF
--- a/src/H5Dint.c
+++ b/src/H5Dint.c
@@ -3225,9 +3225,8 @@ H5D__flush(H5D_t *dset, hid_t dset_id)
 
     /* Currently, H5Oflush causes H5Fclose to trigger an assertion failure in metadata cache.
      * Leave this situation for the future solution */
-    if (H5FD_MPIO == H5F_get_driver_id(dset->oloc.file))
-        HGOTO_ERROR(H5E_DATASET, H5E_CANTFLUSH, FAIL,
-                    "H5Oflush doesn't work properly with MPIO driver for the current design")
+    if (H5F_HAS_FEATURE(dset->oloc.file, H5FD_FEAT_HAS_MPI))
+        HGOTO_ERROR(H5E_DATASET, H5E_UNSUPPORTED, FAIL, "H5Oflush isn't supported for parallel")
 
     /* Flush any dataset information still cached in memory */
     if (H5D__flush_real(dset) < 0)
--- a/src/H5Dint.c
+++ b/src/H5Dint.c
@@ -3223,6 +3223,11 @@ H5D__flush(H5D_t *dset, hid_t dset_id)
     HDassert(dset);
     HDassert(dset->shared);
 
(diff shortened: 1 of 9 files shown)


src/H5Dint.c
```diff
@@ -3223,6 +3223,12 @@ H5D__flush(H5D_t *dset, hid_t dset_id)
     HDassert(dset);
     HDassert(dset->shared);
 
+    /* Currently, H5Oflush causes H5Fclose to trigger an assertion failure in metadata cache.
+     * Leave this situation for the future solution */
+    if (H5FD_MPIO == H5F_get_driver_id(dset->oloc.file))
+        HGOTO_ERROR(H5E_DATASET, H5E_CANTFLUSH, FAIL,
+                    "H5Oflush doesn't work properly with MPIO driver for the current design")
```
derobins: I feel like the HGOTO_ERROR should be protected by an ifdef instead of checking for the MPI-I/O VFD. There may be other parallel VFDs in the future (subfiling, MPI POSIX may return).

jhendersonHDF: Not an #ifdef. We should use `H5F_HAS_FEATURE(f, H5FD_FEAT_HAS_MPI)`.

Also, we should check that the MPI Comm size is > 1 because at least in theory this should be fine when only 1 rank is involved. Would be good to test that though.

src/H5VLnative_datatype.c
```diff
@@ -237,6 +237,12 @@ H5VL__native_datatype_specific(void *obj, H5VL_datatype_specific_args_t *args, h
     switch (args->op_type) {
         /* H5VL_DATATYPE_FLUSH */
         case H5VL_DATATYPE_FLUSH: {
+            /* Currently, H6Oflush causes H5Fclose to trigger an assertion failure in metadata cache.
+             * Leave this situation for the future solution */
+            if (H5FD_MPIO == H5F_get_driver_id(dt->oloc.file))
+                HGOTO_ERROR(H5E_DATASET, H5E_CANTFLUSH, FAIL,
```
jhendersonHDF: I'd change the minor code on these to H5E_UNSUPPORTED and just say that H5Oflush isn't supported in parallel in all the error messages rather than saying it doesn't work properly.

src/H5Dint.c
raylu-hdf: I changed the use of `H5F_get_driver_id `to `H5F_HAS_FEATURE `.  Running with one rank doesn't make `H5Oflush` succeed.  I tested it in the beginning.

src/H5VLnative_datatype.c
raylu-hdf: Fixed.

derobins (approved):

jhendersonHDF (approved):

PR #1883, opened 2022-07-12
Subfiling VFD

jhendersonHDF:
This is the current development branch for the Subfiling VFD feature, outlined in the attached RFC. The feature itself is entirely contained inside the `H5FDsubfiling` directory under `src/`. There are several other supporting changes mostly surrounding VFD testing in this PR as well.

Subfiling uses [Mercury's](https://github.com/mercury-hpc/mercury) threading utilities to manage a pool of worker threads and so brings along a bit of the code from that project with it.

[RFC_VFD_subfiling_200424.pdf](https://github.com/HDFGroup/hdf5/files/9097547/RFC_VFD_subfiling_200424.pdf)


CMakeLists.txt
```diff
@@ -237,6 +237,8 @@ set (HDF5_JAVA_LOGGING_NOP_JAR     ${HDF5_SOURCE_DIR}/java/lib/ext/slf4j-nop-1.7
 set (HDF5_JAVA_LOGGING_SIMPLE_JAR  ${HDF5_SOURCE_DIR}/java/lib/ext/slf4j-simple-1.7.33.jar)
 set (HDF5_DOXYGEN_DIR              ${HDF5_SOURCE_DIR}/doxygen)
 
+set (HDF5_SRC_INCLUDE_DIRS ${HDF5_SRC_DIR})
```
jhendersonHDF, line 240: Since the subfiling VFD is composed of several files, I tried to contain the feature to its own subdirectory under `src/`. CMake then needed to be adjusted so that everywhere ${HDF5_SRC_DIR} was added to a target's include directories, the subfiling subdirectory would be as well. This is so that the public H5FDsubfiling.h and H5FDioc.h headers are available for anything that includes hdf5.h.

Going forward, new subdirectories under `src/` can be added to this `HDF5_SRC_INCLUDE_DIRS` variable and should work seamlessly across the library's components (C++, Java, etc.)

testpar/t_subfiling_vfd.c
```diff
@@ -0,0 +1,260 @@
...
+    h5_clean_files(test_filenames, fapl_id);
+
+    if (H5P_DEFAULT != subfiling_config.ioc_fapl_id)
+        VRFY((H5Pclose(subfiling_config.ioc_fapl_id) >= 0), "FAPL close succeeded");
+
+    return;
+}
+
+int
+main(int argc, char **argv)
```
jhendersonHDF, line 172: Unfortunately, the current regression testing for subfiling mostly consists of a simple file create and close test in this file, as well as some selection I/O testing in testpar/t_vfd.c and the serial HDF5 tests that can be run with `make check-vfd` or `ctest` when the CMake VFD testing option is enabled. The plan is to flesh out more specific regression testing for subfiling, but development work on the feature has been the highest priority so far and the feature has mostly been tested with external applications like IOR and HACC.

src/H5FDsubfiling/H5subfiling_common.h
```diff
@@ -0,0 +1,257 @@
...
+ * This file is part of HDF5.  The full HDF5 copyright notice, including     *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the COPYING file, which can be found at the root of the source code       *
+ * distribution tree, or in https://support.hdfgroup.org/ftp/HDF5/releases.  *
+ * If you do not have access to either file, you may request a copy from     *
+ * help@hdfgroup.org.                                                        *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+/*
+ * Header file for shared code between the HDF5 Subfiling VFD and IOC VFD
```
jhendersonHDF, line 14: Originally, the subfiling and I/O concentrator VFDs were tightly couple together, with most of their routines being marked extern and directly shared between source files. To make the feature more modular and allow subfiling to use different I/O concentrator backends, the VFDs were separated from each other and common subfiling functionality was moved into this "module" so that one can write their own I/O concentrator backend and just include this header.

src/H5FDsubfiling/H5FDioc.c
```diff
@@ -0,0 +1,1778 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
```
jhendersonHDF, line 1: This VFD is the reference I/O concentrator implementation. It works by creating a "main" thread on each MPI rank that is designated as an I/O concentrator rank, as well as a thread pool with a configurable number of worker threads on each of those ranks. "Main" threads will await incoming MPI messages containing I/O requests and will farm those out to the worker threads. The I/O concentrator ranks currently use the sec2 VFD for the underlying subfile, but it should be fairly easy to use other VFDs for file I/O.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -0,0 +1,3264 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
```
jhendersonHDF, line 1: This is the Subfiling VFD, which has the main purpose of breaking I/O requests down into vector I/O requests according to how the I/O requests fall across the different subfiles that the logical HDF5 file is being striped across. It then sends the vector I/O requests down to the underlying I/O concentrator, which handles the details of writing the I/O portions to each subfile.

src/H5FDsubfiling/mercury/src/util/CMakeLists.txt
```diff
@@ -0,0 +1,295 @@
+project(MERCURY_UTIL C)
```
byrnHDF: Can we separate the configure checks from the build commands?
Plus can the configure checks be in a file that can be optioned into the existing config checks files?

jhendersonHDF: In this case, I don't believe we even use this CMake file, so I'll revisit what we actually use out of Mercury.

utils/subfiling_vfd/h5fuse.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
byrnHDF, line 1: Isn't this the post configuration version of the .in file?

jhendersonHDF: Ah yes. Its checkin might have predated creation of the .in file. Will remove

PR #1884, opened 2022-07-13
Fix for a tools compile issue when deprecated symbols are disabled

derobins:
Introduced in #1811, also uses VOL token instead of addr

### DIFF
--- a/java/src/jni/h5util.c
+++ b/java/src/jni/h5util.c
@@ -1180,8 +1180,9 @@ h5str_sprintf(JNIEnv *env, h5str_t *out_str, hid_t container, hid_t tid, void *i
                     CHECK_JNI_EXCEPTION(ENVONLY, JNI_FALSE);
             }
             else if (H5R_OBJ_REF_BUF_SIZE == typeSize) {
-                H5O_info1_t oi;
+                H5O_info2_t oi;
                 hid_t       obj = H5I_INVALID_HID;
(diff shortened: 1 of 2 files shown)


byrnHDF:
> Introduced in #1811, also uses VOL token instead of addr

Does this also apply to 1.12 or was it after 1.12 these were added?

byrnHDF (approved):

derobins:
> > Introduced in #1811, also uses VOL token instead of addr
> 
> Does this also apply to 1.12 or was it after 1.12 these were added?

I'll check and merge if appropriate. Pretty sure it goes to 1.12.

jhendersonHDF (approved):

lrknox (changes requested):
There are similar errors for versioned H5O_info_1_t, etc in java/src/jni/h5util.c.  I'll take a look - probably the fix is similar there.

derobins:
> There are similar errors for versioned H5O_info_1_t, etc in java/src/jni/h5util.c. I'll take a look - probably the fix is similar there.

I just built the fix branch w/ Java and --disable-deprecated-symbols and don't see problems. Everything in java/src/jni/h5util.c uses H5O_info2_t (in develop, at least). I'll keep an eye out for problems in 1.12.

derobins:
I don't see other failures, so I'm going to merge this now so it at least fixes one problem. I'll fix any others that crop up in separate PRs.

Issue #99, opened 2020-11-18
There's no documentation for `H5Fformat_convert`.


gheber:
Is this function deprecated? When was that?

derobins:
Fixed in develop (like internal calls in H5Dpublic.h), will propagate

PR #1910, opened 2022-07-19
Marks H5Fformat_convert as internal in Doxygen

derobins:
Addresses GitHub #99


src/H5Fpublic.h
```diff
@@ -1777,6 +1771,11 @@ H5_DLL herr_t H5Fset_mpi_atomicity(hid_t file_id, hbool_t flag);
 H5_DLL herr_t H5Fget_mpi_atomicity(hid_t file_id, hbool_t *flag);
 #endif /* H5_HAVE_PARALLEL */
 
+/// \cond DEV
+/* Internal API routines */
+H5_DLL herr_t H5Fformat_convert(hid_t fid);
+/// \endcond
```
derobins, line 1777: I feel like we could do something better with this, but for now this is what we do with H5D internal calls.

byrnHDF (approved):

jhendersonHDF (approved):

PR #1911, opened 2022-07-19
Switch to vector I/O for collective metadata writes

jhendersonHDF:
Switch to vector I/O for performing collective metadata writes, instead of using an MPI derived datatype. This allows other MPI-based VFDs (like the Subfiling VFD) to support collective metadata writes, since it's generally not feasible to try and inspect/interpret or even use the derived types when a VFD isn't directly using MPI I/O. It also cleans up this particular code a bit.

I believe this should be equivalent functionally since:

- The old code creates hindexed types, which are also what the MPI I/O VFD generally creates for vector I/O
- The old code will bypass both the page buffer and metadata accumulator, which the new H5F_shared_vector_read/write also need to do for the time being

### DIFF
--- a/src/H5Fio.c
+++ b/src/H5Fio.c
@@ -327,6 +327,85 @@ done:
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5F_shared_select_write() */
 
+herr_t
+H5F_shared_vector_read(H5F_shared_t *f_sh, uint32_t count, H5FD_mem_t types[],
+                       haddr_t addrs[], size_t sizes[], void *bufs[])
+{
+    herr_t ret_value = SUCCEED;
+
+    FUNC_ENTER_NOAPI(FAIL)
+
+    /* Sanity checks */
+    HDassert(f_sh);
+    HDassert((types) || (count == 0));
+    HDassert((addrs) || (count == 0));
+    HDassert((sizes) || (count == 0));
+    HDassert((bufs) || (count == 0));
+
+    /*
+     * Note that we don't try to map global heap data to raw
+     * data here, as it may become expensive to check for when
+     * I/O vectors are large. This may change in the future, but,
+     * for now, assume the caller has done this already.
+     */
+#ifndef NDEBUG
+    for (uint32_t i = 0; i < count; i++)
+        HDassert(types[i] != H5FD_MEM_GHEAP);
+#endif
+
+    /* Pass down to file driver layer (bypass page buffer for now) */
+    if (H5FD_read_vector(f_sh->lf, count, types, addrs, sizes, bufs) < 0)
+        HGOTO_ERROR(H5E_IO, H5E_READERROR, FAIL, "vector read through file driver failed")
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
+
+/*-------------------------------------------------------------------------
+ * Function:    H5F_shared_vector_write
+ *
+ * Purpose:     Writes data from `count` buffers (from the `bufs` array) to
+ *              a file/server/etc. at the offsets provided in the `addrs`
+ *              array, with the data sizes specified in the `sizes` array
+ *              and data memory types specified in the `types` array. The
+ *              addresses are relative to the base address for the file.
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5F_shared_vector_write(H5F_shared_t *f_sh, uint32_t count, H5FD_mem_t types[],
+                        haddr_t addrs[], size_t sizes[], const void *bufs[])
+{
+    herr_t ret_value = SUCCEED;
+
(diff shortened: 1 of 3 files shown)


src/H5Fio.c
```diff
@@ -327,6 +327,67 @@ H5F_shared_select_write(H5F_shared_t *f_sh, H5FD_mem_t type, uint32_t count, H5S
...
+    FUNC_ENTER_NOAPI(FAIL)
+
+    /* Sanity checks */
+    HDassert(f_sh);
+    HDassert((types) || (count == 0));
+    HDassert((addrs) || (count == 0));
+    HDassert((sizes) || (count == 0));
+    HDassert((bufs) || (count == 0));
+
+    /* TODO: map global heap data to raw data here? */
```
jhendersonHDF: Not sure if global heap data needs to be mapped to raw data for each type entry here, or if that's not appropriate?

brtnfld:
@fortnern can you review this? 
@qkoziol if you can find time, can you review this?

Since most of the apps are using collective metadata, we currently have to bypass this for subfiling, which is a pain.

fortnern:
Have we done any performance profiling on this yet?

jhendersonHDF:
> Have we done any performance profiling on this yet?

No performance testing of any kind on this yet; it's more of a functionality over performance thing. That said, did the initial results of comparing selection I/O to the current implementation seem reasonable?

fortnern:
> > Have we done any performance profiling on this yet?
> 
> No performance testing of any kind on this yet; it's more of a functionality over performance thing. That said, did the initial results of comparing selection I/O to the current implementation seem reasonable?

They did, but we should still check the performance of collective metadata since it's replacing a different code path in the library and the usage patterns are different.

src/H5Fio.c
fortnern: Presumably that would be done in H5C.c, not here. The cache client should report what type of data it is if it doesn't already.

fortnern: I suppose you could set it to MEM_GHEAP in H5C.c then scan the array here to make it like selection and scalar I/O, but I'd try to come up with a consistent way to avoid a second pass over that array. We should look into what exactly these types are used for and think if we really need to change them around or just pass the real values to the file driver.

fortnern (approved):
Approved as long as performance testing doesn't show a substantial regression.

src/H5Fio.c
derobins: The remapping is done in H5F_block_write(), which we are replacing with this function, so perhaps that should be done here as well?

derobins: If we update this, might want to update the comment in H5C__verify_len_eoa() as well.

brtnfld: No discernable performance difference.

[summit_pr.pdf](https://github.com/HDFGroup/hdf5/files/9160553/summit_pr.pdf)

brtnfld:
No discernible performance difference. It uses collective metadata. The benchmark was run 4 times for the stats.

summit_pr

src/H5Fio.c
jhendersonHDF: > I suppose you could set it to MEM_GHEAP in H5C.c then scan the array here to make it like selection and scalar I/O, but I'd try to come up with a consistent way to avoid a second pass over that array. We should look into what exactly these types are used for and think if we really need to change them around or just pass the real values to the file driver.

At least when the write comes from H5Cmpio.c, the type should always be H5FD_MEM_DEFAULT, so it shouldn't (at least I think) be a concern for the current PR's changes. For future things making calls to H5F_shared_vector_write though, it might be important to catch MEM_GHEAP; that's what I'm not sure of. Maybe it makes more sense for the upper layers to do the translation before making this call to avoid the extra pass over the vectors like you mention. I'm assuming the translation was simply added to H5F_block_write/read to try to keep it to one place, but that's a little more complicated of a question if you have large I/O vectors.

src/H5Fio.c
```diff
@@ -327,6 +327,85 @@ H5F_shared_select_write(H5F_shared_t *f_sh, H5FD_mem_t type, uint32_t count, H5S
...
+    HDassert((addrs) || (count == 0));
+    HDassert((sizes) || (count == 0));
+    HDassert((bufs) || (count == 0));
+
+    /*
+     * Note that we don't try to map global heap data to raw
+     * data here, as it may become expensive to check for when
+     * I/O vectors are large. This may change in the future, but,
+     * for now, assume the caller has done this already.
+     */
```
jhendersonHDF, line 350: Since this mapping could possibly change in the future and the only current use case shouldn't be using H5FD_MEM_GHEAP anyway, I'll just leave a note for the future here and a debug check to catch the situation if it changes.

fortnern:
Looks good to me

derobins (approved):

qkoziol (commented):
Mostly moot, but looks fine to me

PR #1925, opened 2022-07-23
Switches the tagged metadata list from a skip list to a hash table

derobins:
#1928 Should be merged before this PR

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -139,6 +139,15 @@
 #define R_OK 04
 #endif
 
+/* uthash is an external, header-only hash table implementation.
+ *
+ * We include the file directly in src/ and #define a few functions
+ * to use our internal memory calls.
+ */
+#define uthash_malloc(sz)    H5MM_malloc(sz)
+#define uthash_free(ptr, sz) H5MM_free(ptr) /* Ignoring sz is intentional */
+#include "uthash.h"
+
 /*
  * MPE Instrumentation support
  */
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -8120,7 +8120,6 @@ H5C_cork(H5C_t *cache_ptr, haddr_t obj_addr, unsigned action, hbool_t *corked)
             /* Set the corked status for the entire object */
             tag_info->corked = TRUE;
             cache_ptr->num_objs_corked++;
-
         }
         else {
             /* Sanity check */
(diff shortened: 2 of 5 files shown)


derobins:
This is a non-trivial performance increase in metadata-heavy workflows

src/H5private.h
```diff
@@ -139,6 +139,15 @@
 #define R_OK 04
 #endif
 
+/* uthash is an external, header-only hash table implementation.
```
qkoziol, line 142: If this is brought in up here, the similar code in H5Ipkg.h and H5Dmpio.c should be removed.

qkoziol: Also - I understand that we don't want to redefine "HASH*" in the uthash.h header, but some sense of self-defense here if "HASH*" was already defined in a header that H5private has already included would be good.   Something like "HDcompile_assert(!defined(HASH_FIND));" before including the header?

qkoziol:
Also suggest updating the version of uthash from the most recent one on github:   https://github.com/troydhanson/uthash   (Note different URL than in the header file in 'src')

qkoziol:
Another thing - probably should define HASH_NONFATAL_OOM, so that the uthash code doesn't call exit() within the library.

qkoziol:
And may want to see if uthash_fatal() could be defined to return an error from within the library.

qkoziol:
Oops - accidentally clicked 'Close', sorry!

derobins:
I'll probably sweep up any uthash changes in a separate PR that we can merge before this one. I like to keep to one "coding idea" per PR.

derobins:
> Also suggest updating the version of uthash from the most recent one on github: https://github.com/troydhanson/uthash (Note different URL than in the header file in 'src')

Done (in #1928)

derobins:
> Another thing - probably should define HASH_NONFATAL_OOM, so that the uthash code doesn't call exit() within the library.

Done (in #1928)

src/H5private.h
derobins: Done (in #1928)

H5Ipkg.h was removed in this PR. I missed H5Dmpio.c.

derobins:
> And may want to see if uthash_fatal() could be defined to return an error from within the library.

I added this to my to-do list, but I'm probably not going to address it here.

qkoziol (approved):

jhendersonHDF (approved):

PR #1945, opened 2022-07-28
H5Location::getLinkval: Fix a bug that it always returns empty string

wenjuno:
Update:
Actually, we could initialize the string and use it's internal buffer as the char array for `H5Lget_val` so we don't have to copy from the char array to string.

Original:
The return of `H5Lget_val` for external link always starts with a null so the return from `getLinkval` is always an empty string. Moreover, the values in the char array buffer is null terminated so we must specify the length of char array in the string constructor.

### DIFF
--- a/c++/src/H5Location.cpp
+++ b/c++/src/H5Location.cpp
@@ -1842,9 +1842,10 @@ H5Location::getLinkval(const char *name, size_t size) const
 
         size_t last_notnull_pos(value.find_last_not_of('\0'));
         if (last_notnull_pos == H5std_string::npos) {
-          value.resize(0);
-        } else {
-          value.resize(last_notnull_pos + 1);
+            value.resize(0);
+        }
+        else {
+            value.resize(last_notnull_pos + 1);
         }
     }
     return (value);
--- a/c++/src/H5Location.cpp
+++ b/c++/src/H5Location.cpp
@@ -1834,7 +1834,8 @@ H5Location::getLinkval(const char *name, size_t size) const
     // if link has value, retrieve the value, otherwise, return null string
     if (val_size > 0) {
         // Create buffer for C string
-        value_C = new char[val_size + 1]();
+        ++val_size;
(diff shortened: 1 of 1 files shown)


derobins:
There should be a test for this and if this is correcting bad behavior, there should be a note in RELEASE.txt

bmribler (approved):

qkoziol (approved):

wenjuno:
@derobins Sure. After reading the test for the soft link I'm not 100% sure whether current behavior of returning an empty string for external link is intentional. For the soft link the return from `H5Lget_val` is a null-terminated string such that if you convert to string you can use it like any other string. For external link the return from `H5Lget_val` is _two_ null-terminated strings with a null in the first character (one for file path and the other for object path). So maybe the `getLinkval` is for soft link only? I'm OK with using the C API if that's the case.

c++/src/H5Location.cpp
```diff
@@ -1834,16 +1833,20 @@ H5Location::getLinkval(const char *name, size_t size) const
     // if link has value, retrieve the value, otherwise, return null string
     if (val_size > 0) {
         // Create buffer for C string
-        value_C = new char[val_size + 1]();
+        value.assign(val_size, '\0');
 
-        ret_value = H5Lget_val(getId(), name, value_C, val_size, H5P_DEFAULT);
+        ret_value = H5Lget_val(getId(), name, &value[0], val_size, H5P_DEFAULT);
```
bmribler, line 1838: This will only work for HDF5 1.13 and later because C++11 is required since 1.13. Without the use of C++11, one cannot guarantee that the C++ string is stored contiguously, so passing &value[0] may cause undefined behavior.

wenjuno:
Closing the PR. I think it's a bit premature for me to propose this behavior change. Thanks for the helps.

bmribler:
> Closing the PR. I think it's a bit premature for me to propose this behavior change. Thanks for the helps.

Thanks!

PR #1951, opened 2022-07-29
Adds platform-independent basename and dirname

derobins:
Bringing this over from the VFD SWMR feature branch so it can be used elsewhere


src/H5system.c
```diff
@@ -937,6 +937,146 @@ H5_expand_windows_env_vars(char **env_var)
...
+ */
+
+#ifdef H5_HAVE_WIN32_API
+#define H5_FILE_SEPARATOR '\\'
+#else
+#define H5_FILE_SEPARATOR '/'
+#endif
+
+static size_t
+H5__find_last_file_separator(const char *path, size_t len, hbool_t *no_separator)
```
jhendersonHDF: This should be as simple as calling strrchr().

I also believe empty strings passed to this will return -1 and cause the callers to access invalid memory.

src/H5system.c
```diff
@@ -937,6 +937,146 @@ H5_expand_windows_env_vars(char **env_var)
...
+
+    if (!path)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "path can't be NULL")
+    if (!basename)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "dirname can't be NULL")
+
+    path_len = HDstrlen(path);
+    pos      = H5__find_last_file_separator(path, path_len, &no_separator);
+
+    if (no_separator)
```
jhendersonHDF: If there's no separator, out_len will be set to strlen(path) and not include space for the NUL terminator.

gnuoyd:
On Fri, Jul 29, 2022 at 10:23:04AM -0700, Dana Robinson wrote:
> -- Commit Summary --
> 
>   * Adds platform-independent basename and dirname

It is not necessary for these utilities to be so complicated, is it?

Here are h5_basename() and h5_dirname() in about half as many lines of
standard C:

```
#include <stdlib.h>
#include <string.h>

#ifdef H5_HAVE_WIN32_API
#	define H5_FILE_SEPARATOR '\\'
#else
#	define H5_FILE_SEPARATOR '/'
#endif

char *
h5_strndup(const char *s, size_t n)
{
	char *r;
	size_t len = strlen(s);

	if (len > n)
		len = n;

	if ((r = malloc(len + 1)) == NULL)
		return NULL;

	r[len] = '\0';
	return memcpy(r, s, len);
}

char *
h5_strdup(const char *s)
{
	return h5_strndup(s, strlen(s));
}

char *
h5_dirname(const char *path)
{
	const char *sep;

	if (path == NULL)
		return NULL;

	if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL)
		return h5_strdup(".");

	return h5_strndup(path, sep - path);
}

char *
h5_basename(const char *path)
{
	const char *sep;

	if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL)
		return h5_strdup(path);

	return h5_strdup(sep + 1);
}
```

Dave

jhendersonHDF:
> On Fri, Jul 29, 2022 at 10:23:04AM -0700, Dana Robinson wrote: -- Commit Summary -- * Adds platform-independent basename and dirname
> It is not necessary for these utilities to be so complicated, is it? Here are h5_basename() and h5_dirname() in about half as many lines of standard C: ``` #include <stdlib.h> #include <string.h> #ifdef H5_HAVE_WIN32_API # define H5_FILE_SEPARATOR '\\' #else # define H5_FILE_SEPARATOR '/' #endif char * h5_strndup(const char *s, size_t n) { char *r; size_t len = strlen(s); if (len > n) len = n; if ((r = malloc(len + 1)) == NULL) return NULL; r[len] = '\0'; return memcpy(r, s, len); } char * h5_strdup(const char *s) { return h5_strndup(s, strlen(s)); } char * h5_dirname(const char *path) { const char *sep; if (path == NULL) return NULL; if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL) return h5_strdup("."); return h5_strndup(path, sep - path); } char * h5_basename(const char *path) { const char *sep; if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL) return h5_strdup(path); return h5_strdup(sep + 1); } ``` Dave

@gnuoyd +1 on the elegant basename. For that matter, we don't necessarily need h5_basename to return a copy of a string if we're willing to accept the asymmetry between h5_basename and h5_dirname, e.g.

```
herr_t
h5_basename(const char *path, char **basename)
{
    char *substr = strrchr(path, H5_FILE_SEPARATOR);
    *basename = substr ? substr + 1 : path;
}
```

But these implementations also don't handle the empty string returns "." case. I plan to tidy this up a little bit and add some test cases.

gnuoyd:
On Fri, Jul 29, 2022 at 11:43:59AM -0700, jhendersonHDF wrote:
> But these implementations also don't handle the empty string returns "." case. I plan to tidy this up a little bit and add some test cases.

Ahh, I see what you mean.  This version seems to match the output from
`basename` and `dirname` utilities on Linux:

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#ifdef H5_HAVE_WIN32_API
#	define H5_FILE_SEPARATOR '\\'
#else
#	define H5_FILE_SEPARATOR '/'
#endif

char *
h5_strndup(const char *s, size_t n)
{
	char *r;
	size_t len = strlen(s);

	if (len > n)
		len = n;

	if ((r = malloc(len + 1)) == NULL)
		return NULL;

	r[len] = '\0';
	return memcpy(r, s, len);
}

char *
h5_strdup(const char *s)
{
	return h5_strndup(s, strlen(s));
}

char *
h5_dirname(const char *path)
{
	const char *sep;

	if (path == NULL)
		return NULL;

	if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL)
		return h5_strdup(".");
	else if (sep == path)
		return h5_strdup("/");

	return h5_strndup(path, sep - path);
}

char *
h5_basename(const char *path)
{
	const char *sep;

	if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL)
		return h5_strdup(path);
	else if (sep == path && sep[1] == '\0')
		return h5_strdup("/");

	return h5_strdup(sep + 1);
}

int
main(void)
{
	size_t i;
	char *string[] = {".", "a/b/c", "d/e", "f", "", "/", "/a", "/p/q/r",
	                  "./."};

	for (i = 0; i < sizeof(string) / sizeof(string[0]); i++) {
		char *bn = h5_basename(string[i]);
		char *dn = h5_dirname(string[i]);

		printf("%s dirname %s basename %s\n", string[i], dn, bn);
		free(bn);
		free(dn);
	}
	return EXIT_SUCCESS;
}
```

Dave

jhendersonHDF:
> On Fri, Jul 29, 2022 at 11:43:59AM -0700, jhendersonHDF wrote: But these implementations also don't handle the empty string returns "." case. I plan to tidy this up a little bit and add some test cases.
> Ahh, I see what you mean. This version seems to match the output from `basename` and `dirname` utilities on Linux: ``` #include <stdio.h> #include <stdlib.h> #include <string.h> #ifdef H5_HAVE_WIN32_API # define H5_FILE_SEPARATOR '\\' #else # define H5_FILE_SEPARATOR '/' #endif char * h5_strndup(const char *s, size_t n) { char *r; size_t len = strlen(s); if (len > n) len = n; if ((r = malloc(len + 1)) == NULL) return NULL; r[len] = '\0'; return memcpy(r, s, len); } char * h5_strdup(const char *s) { return h5_strndup(s, strlen(s)); } char * h5_dirname(const char *path) { const char *sep; if (path == NULL) return NULL; if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL) return h5_strdup("."); else if (sep == path) return h5_strdup("/"); return h5_strndup(path, sep - path); } char * h5_basename(const char *path) { const char *sep; if ((sep = strrchr(path, H5_FILE_SEPARATOR)) == NULL) return h5_strdup(path); else if (sep == path && sep[1] == '\0') return h5_strdup("/"); return h5_strdup(sep + 1); } int main(void) { size_t i; char *string[] = {".", "a/b/c", "d/e", "f", "", "/", "/a", "/p/q/r", "./."}; for (i = 0; i < sizeof(string) / sizeof(string[0]); i++) { char *bn = h5_basename(string[i]); char *dn = h5_dirname(string[i]); printf("%s dirname %s basename %s\n", string[i], dn, bn); free(bn); free(dn); } return EXIT_SUCCESS; } ``` Dave

Definitely seems to handle more of the odd edge cases for basename and dirname. Makes me wonder though, does it make more sense to return H5_FILE_SEPARATOR instead of "/" on all platforms? Certainly, returning "/" on Windows would be conforming to the basename and dirname specs, but might be surprising on Windows.

derobins:
We should probably return the Windows file separator, not "/" to avoid Windows-specific path munging in other code.

src/H5system.c
```diff
@@ -935,8 +935,170 @@ H5_expand_windows_env_vars(char **env_var)
 
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5_expand_windows_env_vars() */
+
```
qkoziol, line 922: Needs comment header

src/H5system.c
```diff
@@ -935,8 +935,170 @@ H5_expand_windows_env_vars(char **env_var)
...
+    if (FAIL == ret_value) {
+        HDfree(out);
+        if (dirname)
+            *dirname = NULL;
+    }
+
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5_dirname() */
+
+herr_t
```
qkoziol, line 1108: I see that there's discussion about basename in the comment above dirname, but if a developer is looking at basename <n> years from now, they probably won't see it.  Helpful comments here should be added.

src/H5system.c
```diff
@@ -935,8 +935,170 @@ H5_expand_windows_env_vars(char **env_var)
...
+
+    if (!s)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, NULL, "string cannot be NULL")
+
+    len = HDstrlen(s);
+
+    if (len > n)
+        len = n;
+
+    if (NULL == (ret_value = HDmalloc(len + 1)))
```
qkoziol: Should use H55MM_malloc here for consistency with other routines in this file.

test/th5_system.c
```diff
@@ -0,0 +1,276 @@
...
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the COPYING file, which can be found at the root of the source code       *
+ * distribution tree, or in https://www.hdfgroup.org/licenses.               *
+ * If you do not have access to either file, you may request a copy from     *
+ * help@hdfgroup.org.                                                        *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+#define H5_SYSTEM_TEST_PATH_MAX 4096
+
+/***********************************************************
```
qkoziol, line 15: Needs a test for new strndup routine.

qkoziol (changes requested):
Implementation is fine, but carries technical debt without some small fixes.

src/H5system.c
jhendersonHDF: Added a header comment

src/H5system.c
jhendersonHDF: Added a minimal header to each one, presuming that developers know how to `man dirname`/`man basename` for any other relevant details.

src/H5system.c
jhendersonHDF: Had to add a new H5MM_strndup to satisfy this for H5_dirname and H5_basename, but done.

test/th5_system.c
jhendersonHDF: Added

jhendersonHDF:
@gnuoyd As it turns out, sticking to a dirname/basename that don't modify their input strings and which conform to POSIX semantics in basename's case is actually fairly involved. The dirname implementation here is very similar to the glibc implementation, but slightly more complex since modifying the path is a nifty shortcut we can't use here.

jhendersonHDF:
@qkoziol Any chance you have time for a quick once-over on this? I believe I've addressed your concerns and I could make use of these changes right away.

src/H5system.c
qkoziol: Looks good

qkoziol (approved):

PR #1952, opened 2022-07-29
Run bin/reconfigure to propagate fortran mod change to configure and

lrknox:
Makefile.in files.


byrnHDF (approved):

brtnfld (approved):
I assume CMake is still an issue?

PR #1953, opened 2022-07-30
Onion VFD

derobins:


derobins:
There is still some technical debt here. In particular, the tools integration needs rework, but that has exposed larger bugs in the tools regarding specifying VFDs by name, so we'll fix that in 1.13.3.

I'll add a release note and add the tools tests to CMake over the weekend.

derobins:
All the daily test failures should be addressed now and CMake runs the new tools tests. I'll add a release note tomorrow.

derobins:
This VFD cannot be merged to 1.12 or earlier due to reliance on the VFL ctl callback.

byrnHDF (approved):

tools/lib/h5diff.c
```diff
@@ -648,8 +648,15 @@ h5diff(const char *fname1, const char *fname2, const char *objname1, const char
...
+    if (opts->vfd_info[0].u.name) {
+        if ((fapl1_id = h5tools_get_fapl(H5P_DEFAULT, NULL, &(opts->vfd_info[0]))) < 0) {
+            parallel_print("h5diff: unable to create fapl for input file\n");
+            H5TOOLS_GOTO_ERROR(H5DIFF_ERR, "unable to create input fapl\n");
+        }
+    }
+
     if (opts->custom_vol[0] || opts->custom_vfd[0]) {
-        if ((fapl1_id = h5tools_get_fapl(H5P_DEFAULT, opts->custom_vol[0] ? &(opts->vol_info[0]) : NULL,
+        if ((fapl1_id = h5tools_get_fapl(fapl1_id, opts->custom_vol[0] ? &(opts->vol_info[0]) : NULL,
```
jhendersonHDF, line 659: Comment for future reference from old PR:

This is going to copy fapl1_id and assign it back to itself, losing the original reference. Same for fapl2_id below

src/H5FDonion_index.c
```diff
@@ -0,0 +1,935 @@
...
+ *              Then the decode operation is called a second time, and all
+ *              components will be populated (and again number of bytes read is
+ *              returned).
+ *
+ * Return:      Success:    Number of bytes read from buffer
+ *              Failure:    0
+ *-----------------------------------------------------------------------------
+ */
+size_t
+H5FD__onion_revision_record_decode(unsigned char *buf, H5FD_onion_revision_record_t *record)
```
jhendersonHDF, line 587: Comment for future reference:

The decode and encode routines in this file should probably be aware of how large buf is so we don't run off the end of it on accident.

src/H5FDonion_history.c
```diff
@@ -0,0 +1,297 @@
...
+    unsigned char *ptr         = NULL;
+    size_t         ret_value   = 0;
+
+    FUNC_ENTER_PACKAGE;
+
+    HDassert(buf != NULL);
+    HDassert(history != NULL);
+    HDassert(H5FD__ONION_HISTORY_VERSION_CURR == history->version);
+
+    if (HDstrncmp((const char *)buf, "OWHS", 4))
```
jhendersonHDF: Comment from old PR:

This should probably be H5FD__ONION_HISTORY_SIGNATURE instead of a "magic" string

src/H5FDonion_history.c
```diff
@@ -0,0 +1,297 @@
...
+ *              Then the decode operation is called a second time, and all
+ *              components will be populated (and again number of bytes read is
+ *              returned).
+ *
+ * Return:      Success:    Number of bytes read from buffer
+ *              Failure:    0
+ *-----------------------------------------------------------------------------
+ */
+size_t
+H5FD__onion_history_decode(unsigned char *buf, H5FD_onion_history_t *history)
```
jhendersonHDF, line 158: Comment for future reference:

The decode and encode routines in this file should probably be aware of how large buf is so we don't run off the end of it on accident.

src/H5FDonion_header.c
```diff
@@ -0,0 +1,231 @@
...
+ *
+ *              Implementation must correspond with
+ *              H5FD__onion_header_encode().
+ *
+ * Return:      Success:    Number of bytes read from buffer
+ *              Failure:    0
+ *-----------------------------------------------------------------------------
+ */
+size_t
+H5FD__onion_header_decode(unsigned char *buf, H5FD_onion_header_t *header)
```
jhendersonHDF, line 121: Comment for future reference:

The decode and encode routines in this file should probably be aware of how large buf is so we don't run off the end of it on accident.

src/H5FDonion_history.c
```diff
@@ -0,0 +1,297 @@
...
+    if (history_out->n_revisions > 0 && NULL == history_out->record_locs)
+        HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "can't allocate record pointer list")
+
+    if (H5FD__onion_history_decode(buf, history_out) != size)
+        HGOTO_ERROR(H5E_VFL, H5E_CANTDECODE, FAIL, "can't decode history (final)")
+
+done:
+    H5MM_xfree(buf);
+    if (ret_value == FAIL)
+        H5MM_xfree(history_out->record_locs);
```
jhendersonHDF, line 83: Comment for future reference from old PR:

This line can cause a crash either:

- if history_out is NULL and we fail
- if history_out was partially initialized and we try to free record_locs when it isn't NULL and not allocated

tools/src/h5diff/h5diff_common.c
```diff
@@ -56,6 +56,28 @@ static struct h5_long_options l_opts[] = {{"help", no_arg, 'h'},
                                           {"vfd-info-2", require_arg, 'Z'},
                                           {NULL, 0, '\0'}};
 
+static H5FD_onion_fapl_info_t onion_fa_g_1 = {
```
jhendersonHDF, line 59: Comment for future reference:

The tools should have basically no idea about the Onion VFD or its structure definitions, but for now this is good enough to make the tools play nice.

jhendersonHDF:
I'd prefer for the comment about the magic string in H5FDonion_history.c to get fixed, but other than that I think this is good enough to merge and take care of the leftover bits following that.

jhendersonHDF (approved):

src/H5FDonion_history.c
derobins: Fixed

src/H5FDonion_history.c
derobins: Fixed

tools/src/h5diff/h5diff_common.c
derobins: Agree. Aren't we waiting on some library fixes, though? We can make the changes after the merge.

derobins:
Also, this VFD lacks Java/Fortran/C++ wrappers. I'll add those in the future.

lrknox (approved):

PR #1906, opened 2022-07-18
Greatly simplifies Autotools bits of autogen.sh

derobins:
Should fix #1343

PR #1954, opened 2022-07-30
Reverts the autogen.sh changes from e3800e1 (#1906)

derobins:
This is causing issues with the NAG Fortran compiler in our daily
testing due to -shared getting inappropriately added. Reverting while
we investigate.


derobins:
This was never pushed downstream, so it doesn't need to go to 1.12 or 1.10

lrknox (approved):

PR #1956, opened 2022-07-30
Fixes clang warnings in H5Tconv.c (long) double macros

derobins:
This might fix the failures in @jhendersonHDF's memory PR


qkoziol (approved):
Nice catch!

jhendersonHDF (approved):

PR #1961, opened 2022-08-01
Set entry pointer mem type for vector I/O in collective metadata writes

jhendersonHDF:
Whereas the old collective metadata write code used to just use H5FD_MEM_DEFAULT for the memory type for writing, the switch to vector I/O lets us specify the memory type for each metadata entry being written and so lets the underlying VFD handle things with more granularity. In particular, the Subfiling VFD mirrors superblock writes to a stub file for compatibility with legacy POSIXy HDF5 applications.

### DIFF
--- a/src/H5Cmpio.c
+++ b/src/H5Cmpio.c
@@ -997,11 +997,12 @@ H5C__collective_write(H5F_t *f)
             HGOTO_ERROR(H5E_CACHE, H5E_NOTFOUND, FAIL, "can't retrieve skip list item")
 
         /* Set up initial array position & buffer base address */
+        HDassert(entry_ptr->type);
         base_buf = entry_ptr->image_ptr;
         addrs[0] = entry_ptr->addr;
         sizes[0] = entry_ptr->size;
         bufs[0]  = base_buf;
-        types[0] = H5FD_MEM_DEFAULT;
+        types[0] = entry_ptr->type->mem_type;
 
         node = H5SL_next(node);
         i    = 1;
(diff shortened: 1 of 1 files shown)


derobins (approved):

src/H5Cmpio.c
```diff
@@ -1010,18 +1012,21 @@ H5C__collective_write(H5F_t *f)
...
+            if (types[i] != types[i - 1])
+                identical_types = FALSE;
 
             /* Advance to next node & array location */
             node = H5SL_next(node);
             i++;
         } /* end while */
 
         /* Optimization for vector I/O */
-        if (count > 1)
```
fortnern, line 1024: I'm not sure if this identical types optimization here really buys us anything. We already have the array, might as well pass it to the FD layer. I suppose you could delay allocating the types array until the main loop is finished, and do a second pass if the types are not identical.

src/H5Cmpio.c
```diff
@@ -1010,18 +1012,21 @@ H5C__collective_write(H5F_t *f)
...
+        if (count > 1 && identical_types)
```
fortnern: I'm not sure if this identical types optimization here really buys us anything. We already have the array, might as well pass it to the FD layer. I suppose you could delay allocating the types array until the main loop is finished, and do a second pass if the types are not identical.

jhendersonHDF: I thought the idea was to tell the vector I/O implementation that it doesn't need to do any processing on the rest of the types array after the first element? Not that there's likely to be any measurable overhead in either case, but it's essentially free to do here, whereas if we don't the vector I/O code will walk the types array.

jhendersonHDF: @fortnern I've removed that particular piece of code for now. I still think it'll perform slightly better for large metadata writes, but I'll keep an eye on it in the future.

fortnern (approved):

PR #1963, opened 2022-08-03
Subfiling updates for release

jhendersonHDF:


tools/lib/h5tools.c
```diff
@@ -81,11 +81,20 @@ const char *volnames[] = {
...
+    [WINDOWS_VFD_IDX]   = "windows",
+    [STDIO_VFD_IDX]     = "stdio",
+    [CORE_VFD_IDX]      = "core",
+    [FAMILY_VFD_IDX]    = "family",
+    [SPLIT_VFD_IDX]     = "split",
+    [MULTI_VFD_IDX]     = "multi",
+    [MPIO_VFD_IDX]      = "mpio",
+    [ROS3_VFD_IDX]      = "ros3",
+    [HDFS_VFD_IDX]      = "hdfs",
+    [SUBFILING_VFD_IDX] = H5FD_SUBFILING_NAME,
```
derobins, line 96: Should we define the other names as constants as well?

tools/lib/h5tools.h
```diff
@@ -578,15 +578,15 @@ typedef struct h5tools_vfd_info_t {
...
 } h5tools_vfd_info_t;
 
-/* This enum should match the entries in the above 'volnames'
- * since they are indices into the 'volnames' array. */
+/* This enum should match the entries in the 'volnames'
+ * array since they are indices into that array. */
 typedef enum {
     NATIVE_VOL_IDX = 0,
     PASS_THROUGH_VOL_IDX,
 } vol_idx;
```
derobins, line 586: We should probably do the same this with the VOL array that we do with the VFD array.

tools/lib/h5tools.c
jhendersonHDF: It would probably be a good idea to keep strings in one single place and use a macro everywhere else. I don't know how many VFDs follow this pattern though. Some of them may need some tweaks.

tools/lib/h5tools.h
jhendersonHDF: Agreed. Anything to make the maintainability there better.

tools/lib/h5tools.c
derobins: We don't have to make all the changes here. We can have separate PRs later to address orthogonal issues like this.

byrnHDF (approved):

derobins (approved):

PR #1967, opened 2022-08-03
Sanity checks for H5Xcreate_anon functions to make sure the IDs passed in are valid

raylu-hdf:
vol_tests tests H5Dcreate_anon, H5Gcreate_anon, and H5Tcommit_anon with invalid property list IDs.  I verified the fixes with vol_tests.

### DIFF
--- a/src/H5D.c
+++ b/src/H5D.c
@@ -279,11 +279,20 @@ H5Dcreate_anon(hid_t loc_id, hid_t type_id, hid_t space_id, hid_t dcpl_id, hid_t
     FUNC_ENTER_API(H5I_INVALID_HID)
     H5TRACE5("i", "iiiii", loc_id, type_id, space_id, dcpl_id, dapl_id);
 
+    /* Sanity check */
+    if (loc_id < 0 || type_id < 0 || space_id < 0 || dcpl_id < 0 || dapl_id < 0)
+        HGOTO_ERROR(H5E_ID, H5E_BADID, (-1), "ID must be valid")
+
     /* Check arguments */
     if (H5P_DEFAULT == dcpl_id)
         dcpl_id = H5P_DATASET_CREATE_DEFAULT;
     else if (TRUE != H5P_isa_class(dcpl_id, H5P_DATASET_CREATE))
         HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, H5I_INVALID_HID, "not dataset create property list ID")
+
+    if (H5P_DEFAULT == dapl_id)
+        dapl_id = H5P_DATASET_ACCESS_DEFAULT;
+    else if (TRUE != H5P_isa_class(dapl_id, H5P_DATASET_ACCESS))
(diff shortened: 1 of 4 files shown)


src/H5D.c
```diff
@@ -279,12 +279,21 @@ H5Dcreate_anon(hid_t loc_id, hid_t type_id, hid_t space_id, hid_t dcpl_id, hid_t
     FUNC_ENTER_API(H5I_INVALID_HID)
     H5TRACE5("i", "iiiii", loc_id, type_id, space_id, dcpl_id, dapl_id);
 
+    /* Sanity check */
+    if (loc_id < 0 || type_id < 0 || space_id < 0 || dcpl_id < 0 || dapl_id < 0)
```
jhendersonHDF: I don't think these are necessary. `dcpl_id` and `dapl_id` will be checked by H5P_isa_class below. `loc_id` will be checked by H5I_object. There might be a case for `type_id` and `space_id`, but I'm assuming those will fail a bit further when we try to get an object out of them.

src/H5Iint.c
```diff
@@ -1069,7 +1069,8 @@ H5I_dec_ref(hid_t id)
     FUNC_ENTER_NOAPI((-1))
 
     /* Sanity check */
-    HDassert(id >= 0);
+    if (id < 0)
```
jhendersonHDF: Since this is library-internal, I think the assert is fine and there shouldn't be a need to check for id < 0. That's the point of the assertion in debug builds.

src/H5Tcommit.c
```diff
@@ -337,6 +337,10 @@ H5Tcommit_anon(hid_t loc_id, hid_t type_id, hid_t tcpl_id, hid_t tapl_id)
     FUNC_ENTER_API(FAIL)
     H5TRACE4("e", "iiii", loc_id, type_id, tcpl_id, tapl_id);
 
+    /* Sanity check */
+    if (loc_id < 0 || type_id < 0 || tcpl_id < 0 || tapl_id < 0)
```
jhendersonHDF: Same as above. `loc_id`, `type_id`, `tcpl_id` and `tapl_id` are all checked immediately below.

qkoziol:
I am in the middle on these - I would suggest writing tests that exercise all of the parameters that are checked for and then only add in the checks for tests that operate incorrectly.

raylu-hdf:
> I am in the middle on these - I would suggest writing tests that exercise all of the parameters that are checked for and then only add in the checks for tests that operate incorrectly.

I did what you suggested and removed several unnecessary sanity checks.  Thanks.

src/H5D.c
raylu-hdf: Corrected.

src/H5Iint.c
raylu-hdf: Corrected.

src/H5Tcommit.c
raylu-hdf: Fixed.

jhendersonHDF (approved):

derobins (approved):

Issue #1419, opened 2022-02-03
H5Dchunk_iter offset is not scaled by chunk dimensions


wenjuno:
I'm not very sure if this is intentional but I find it inconsistent with `offset` in `H5Dget_chunk_info`.
I think the relevant code for `H5Dchunk_iter` is this:
https://github.com/HDFGroup/hdf5/blob/1d598182d6097052060f15a6b0d744c42a2fd80a/src/H5Dchunk.c#L7519-L7520
Comparing to how the `offset` in `H5Dget_chunk_info` is calculated below, there are missing multipliers of `dset->shared->layout.u.chunk.dim[ii]`:
https://github.com/HDFGroup/hdf5/blob/1d598182d6097052060f15a6b0d744c42a2fd80a/src/H5Dchunk.c#L7347-L7349
In order to use `H5Dchunk_iter` I will need to apply the chunk dimensions multipliers myself. Is this a bug or a conscious design choice? Thanks.

mkitti:
This is an important inconsistency. @gauteh was this intended?

gauteh:
Hm, that was probably not intended: I can see in my library code that I scale it when this function is available, and that I don't scale when I use `chunk_info` directly. 

I also noticed that in https://github.com/HDFGroup/hdf5/blob/1d598182d6097052060f15a6b0d744c42a2fd80a/src/H5Dchunk.c#L7487 the *offset seems to not be written to, that also seems like a bug?

Neither it seems that the offset is set here: https://github.com/HDFGroup/hdf5/blob/1d598182d6097052060f15a6b0d744c42a2fd80a/src/H5Dchunk.c#L7391

mkitti:
> the *offset seems to not be written to, that also seems like a bug?

You're looking at `H5Dget_chunk_info_by_coord` which takes `offset` as an input parameter.

https://docs.hdfgroup.org/hdf5/develop/group___h5_d.html#ga408a49c6ec59c5b65ce4c791f8d26cb0

gauteh:
Doh, of course.

mkitti:
I'll tackle if this unless someone else has started already.

PR #1968, opened 2022-08-04
Backport H5Dchunk_iter to 1.10 branch

mkitti:
This backports H5Dchunk_iter #6 to the 1.10 branch from the 1.13 development branch.
xref: https://forum.hdfgroup.org/t/possibility-of-backporting-h5dchunk-iter-to-1-12/9971/2

To do:
- [x] Backport tests from `test/chunk_iter.c`
- [x] Backport #1969 to fix offsets that were not multiplied by the chunk dimensions (see #1419)
- [x] Finish backport of `H5Dchunk_iter` to 1.12: #1970

### DIFF
--- a/test/chunk_info.c
+++ b/test/chunk_info.c
@@ -45,7 +45,7 @@ const char *FILENAME[] = {"tchunk_info_18", "tchunk_info_110", "chunk_info", NUL
 /* From original test */
 #define DATASETNAME "2d"
 
-#define BASIC_FILE      "basic_query"
+#define BASIC_FILE "basic_query"
 
 /* Parameters for testing chunk querying */
 #define RANK                 2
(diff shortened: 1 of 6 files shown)


mkitti:
The test failure encountered in #666 may be an issue here as well

derobins:
This needs to go to 1.12 first. We move features down the chain of maintenance branches.

test/chunk_info.c
```diff
@@ -1666,6 +1806,280 @@ test_chunk_info_version2_btrees(char *filename, hid_t fapl)
...
+    H5E_BEGIN_TRY
+    {
+        ret = H5Dget_chunk_info(dset, dspace, chk_index, out_offset, &read_flt_msk, &addr, &size);
+    }
+    H5E_END_TRY;
+    /* In HDF5 1.10 this will not fail. It will fail in 1.12 and older. */
+    /*
+    if (ret != FAIL)
+        TEST_ERROR;
+    */
```
mkitti, line 1968: Trying to fix this creates new issues on HDF5 1.10.

mkitti:
I backported the `test_basic_query` in `test/chunkinfo.c`. This mostly involved changing the names of some constants to older variants. Also an actual dataspace identifier instead of `H5S_ALL` needed to be used.

I noticed some difference in return value behavior of `H5Dget_chunk_info` when trying to get information on a chunk greater than the number of chunks. Thus I had to disable one test that I noted above.

Since the HDF5 1.12 version of `H5Dchunk_iter` is advancing, the time for this may be appropriate now.

byrnHDF (approved):

PR #1969, opened 2022-08-05
H5Dchunk_iter now passes offsets in units of dataset elements, fix #1419

### LINKED ISSUE #1419: H5Dchunk_iter offset is not scaled by chunk dimensions

mkitti:
This fixes #1419 by adding a `H5O_layout_chunk_t` pointer to `H5D_chunk_iter_ud_t` and using the `dim` and `ndim` fields.

This uses an statically sized array `hsize_t                    offset[H5O_LAYOUT_NDIMS]` which is passed to the user callback function.

`test/chunk_info.c` is modified to match the changes and is now consistent with `H5Dget_chunk_info`.

## Tasks
- [x] Update the documentation (H5Dpublic.h) to clarify that "logical coordinates of chunk" uses units of dataset elements

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -254,6 +254,7 @@ typedef struct H5D_chunk_coll_fill_info_t {
 typedef struct H5D_chunk_iter_ud_t {
     H5D_chunk_iter_op_t op;      /* User defined callback */
     void               *op_data; /* User data for user defined callback */
+    H5O_layout_chunk_t *chunk;   /* Chunk layout */
 } H5D_chunk_iter_ud_t;
 
 /********************/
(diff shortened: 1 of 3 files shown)


fortnern (approved):

fortnern:
We should probably update the documentation (H5Dpublic.h> to clarify that "logical coordinates of chunk" uses units of dataset elements

derobins:
I'll approve and merge this once the documentation is updated.

mkitti:
We should regularize the description of the arguments. The descriptions of the arguments should be the same between `H5Dchunk_iter`, `H5Dget_chunk_info`, and `H5Dget_chunk_info_by_coord`. I revised the documentation for all three and regularized the changes so identical parameters have the same descriptions.

Also I changed `nbytes` to `size` since that has precedence. However, perhaps `nbytes` is a clearer name for this argument.

fortnern:
Looks good to me. Thanks!

derobins (approved):

derobins:
This also needs a note in RELEASE.txt, but I'll take care of that.

mkitti:
Thank you for the reviews. I will leave this branch alone now, and begin moving these changes into 1.12 via #1970 .

src/H5Dchunk.c
```diff
@@ -8066,13 +8067,20 @@ static int
 H5D__chunk_iter_cb(const H5D_chunk_rec_t *chunk_rec, void *udata)
 {
     const H5D_chunk_iter_ud_t *data      = (H5D_chunk_iter_ud_t *)udata;
+    const H5O_layout_chunk_t  *chunk     = data->chunk;
     int                        ret_value = H5_ITER_CONT;
+    hsize_t                    offset[H5O_LAYOUT_NDIMS];
+    unsigned                   ii; /* Match H5O_layout_chunk_t.ndims */
+
+    /* Similar to H5D__get_chunk_info */
+    for (ii = 0; ii < chunk->ndims; ii++)
```
ZhaoMJ, line 8076: I think this is off by one?  The `ndims` here is 1 larger than the rank I set. I also see in other places, `layout->ndims` is decreased by 1 to determine the number of dimensions.
https://github.com/HDFGroup/hdf5/blob/509fe962d42b3f02d356d2668b04712df06daf7f/src/H5Dearray.c#L1095-L1100

ZhaoMJ: If you add an assertion here and run [this test](https://github.com/HDFGroup/hdf5/blob/509fe962d42b3f02d356d2668b04712df06daf7f/test/chunk_info.c#L1579), it's easy to see that `chunk->ndims` is 3 but we clearly expect the `offset` array to have two valid elements.

mkitti: You are correct that there are ndims - 1 dimensions, but this is an internal implementation detail of the library. That specific loop comes from `H5D__get_chunk_info`.

What is the actual problem though? At worse, we do one more multiplication than needed while avoiding a subtraction and another variable on the stack.

PR #1971, opened 2022-08-05
Develop cmake clean

byrnHDF:
This updates the CMake test code to remove generated files - this will not remove files copied from the source to the test folder.


derobins (approved):

lrknox (approved):

c++/examples/CMakeTests.cmake
```diff
@@ -16,17 +16,31 @@
...
-             Group.h5
-             SDS.h5
-             SDScompound.h5
-             SDSextendible.h5
-             Select.h5
+        -E remove ${CPP_EX_CLEANFILES}
+)
+set_tests_properties (CPP_ex-clear-objects PROPERTIES
+    FIXTURES_SETUP clear_cppex
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
```
jhendersonHDF, line 33: These two are identical right? Can we not just set the FIXTURES_SETUP and FIXTURES_CLEANUP properties on the same test and not have duplicates everywhere?

byrnHDF: If I read the CMake documentation correctly, FIXTURES_SETUP and FIXTURES_CLEANUP cannot be the same test. Therefore the dups. Actually the SETUP could probably be deleted since the clean should have cleanedthe files.

PR #1972, opened 2022-08-05
Update java code with doxygen tags, add menu files.

byrnHDF:


config/cmake/UseJava.cmake
```diff
@@ -1448,6 +1448,7 @@ function(create_javadoc _target)
 
     add_custom_target(${_target}_javadoc ALL
         COMMAND ${Java_JAVADOC_EXECUTABLE}
+                -Xdoclint:none
```
byrnHDF, line 1451: this needs added to the autotools version

derobins (approved):

PR #1973, opened 2022-08-05
Fix issues in tools code after Subfiling VFD updates

jhendersonHDF:


src/H5FDsubfiling/H5FDioc.c
```diff
@@ -841,6 +843,12 @@ H5FD__ioc_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxaddr)
     /* Fill in the file config values */
     HDmemcpy(&file_ptr->fa, config_ptr, sizeof(H5FD_ioc_config_t));
 
+    /* Copy the ioc FAPL. */
+    if (H5FD__copy_plist(config_ptr->under_fapl_id, &(file_ptr->fa.under_fapl_id)) < 0) {
+        file_ptr->fa.under_fapl_id = H5I_INVALID_HID;
```
jhendersonHDF, line 848: The tools code tries all available VFDs when it can't open with sec2. Here it tried Subfiling and when Subfiling couldn't resolve file paths below (when it didn't create the files), it was closing the file FAPL out from under the tools code and causing an infinite loop.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -354,13 +354,7 @@ H5FD_subfiling_init(void)
                     "Subfiling VFD requires the use of MPI_Init_thread with MPI_THREAD_MULTIPLE");
         }
         else {
-            char *env_var;
-            int   required = MPI_THREAD_MULTIPLE;
-
-            /* Ensure that Subfiling VFD has been loaded dynamically */
-            env_var = HDgetenv(HDF5_DRIVER);
-            if (!env_var || HDstrcmp(env_var, H5FD_SUBFILING_NAME))
-                H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTINIT, H5I_INVALID_HID, "MPI isn't initialized");
```
jhendersonHDF, line 363: If the Subfiling VFD is called from serial code (the tools code in this case), it always needs to initialize MPI, even when it wasn't dynamically loaded.

derobins (approved):

lrknox (approved):

PR #1976, opened 2022-08-08
Turns off clang optimizations on a couple of onion VFD functions

derobins:
Ancient clang (7.7.3 on OS X) emits problematic code with -O2 or -O3

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -289,6 +289,18 @@
 #       define H5_ATTR_MALLOC /*void*/
 #  endif
 
+/* Turns off optimizations for a function. Goes after the return type.
+ * Not generally needed in the library, but ancient versions of clang
+ * (7.3.3, possibly others) have trouble with some of the onion VFD decode
+ * functions and need the optimizer turned off. This macro can go away when
+ * we figure out what's going on and can engineer another solution.
+ */
+#  if defined(__clang__)
+#       define H5_ATTR_NO_OPTIMIZE __attribute__((optnone))
+#  else
+#       define H5_ATTR_NO_OPTIMIZE /*void*/
+#  endif
+
 #else
 #   define H5_ATTR_FORMAT(X, Y, Z) /*void*/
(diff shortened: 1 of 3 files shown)


src/H5private.h
```diff
@@ -289,6 +289,18 @@
 #       define H5_ATTR_MALLOC /*void*/
 #  endif
 
+/* Turns off optimizations for a function. Goes after the return type.
+ * Not generally needed in the library, but ancient versions of clang
+ * (7.3.3, possibly others) have trouble with some of the onion VFD decode
+ * functions and need the optimizer turned off. This macro can go away when
+ * we figure out what's going on and can engineer another solution.
+ */
+#  if defined(__clang__)
```
byrnHDF, line 298: Can it be targeted to a less then version of clang?

byrnHDF (approved):

src/H5private.h
derobins: Maybe but I'm not going to investigate that now

jhendersonHDF (approved):
It's a bit unfortunate of a need, but I don't immediately see what the decode routines are doing wrong. Doesn't seem to be any pointer aliasing (which was my first thought).

byrnHDF:
> It's a bit unfortunate of a need, but I don't immediately see what the decode routines are doing wrong. Doesn't seem to be any pointer aliasing (which was my first thought).

I don't think it is code but the actual compiler on that version of mac. I remember there was an issue to do with macs and optimization at one time. There existed an override in maybe one of the autotools config files that changed/forced the optimization.

gnuoyd:
On Mon, Aug 08, 2022 at 10:08:15AM -0700, Dana Robinson wrote:
> Ancient clang (7.7.3 on OS X) emits problematic code with -O2 or -O3
> You can view, comment on, or merge this pull request online at:

A comment probably should say how the emitted code is problematic so
that the next person to work with the source knows what to look for.

There is not a subtle bug lurking?

Dave

PR #1979, opened 2022-08-09
Remove stray files, incorrect vds and mpi clean files

byrnHDF:


examples/CMakeTests.cmake
```diff
@@ -116,27 +116,6 @@ if (H5_HAVE_PARALLEL AND HDF5_TEST_PARALLEL AND NOT WIN32)
...
-  )
-  add_test (
-      NAME MPI_TEST_EXAMPLES-clean-objects
-      COMMAND    ${CMAKE_COMMAND} -E remove ${mpi_test_ex_CLEANFILES}
-  )
-  set_tests_properties (MPI_TEST_EXAMPLES-clean-objects PROPERTIES
-      FIXTURES_CLEANUP clear_MPI_TEST_EXAMPLES
-      WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
-  )
-
```
lrknox, line 139: Will these be removed permanently, or just until it is known which files should be removed?  I think the test error occurred because the list of files to delete was empty?

lrknox (approved):

examples/CMakeTests.cmake
byrnHDF: I looked at the tests - they seem to cleanup properly.

lrknox: Ok, so this clean/clear code wasn't needed and needs only to be removed.

derobins (approved):

PR #1980, opened 2022-08-09
Merge UG from 1.10 and add prelim vol section

byrnHDF:
Only new is the limited and incomplete VOL chapter


java/src/hdf/hdf5lib/H5.java
```diff
@@ -6434,64 +6972,16 @@ public static H5O_info_t H5Oget_info(long loc_id) throws HDF5LibraryException, N
...
-     *            Error from the HDF-5 Library.
+     *            Error from the HDF5 Library.
      * @exception NullPointerException
      *            name is null.
      **/
     public synchronized static native H5O_info_t H5Oget_info(long loc_id, int fields)
         throws HDF5LibraryException, NullPointerException;
 
     /**
-     * H5Oget_info_by_name retrieves the metadata for an object, identifying the object by location and
```
jhendersonHDF, line 6445: Lots of deletions below. Were these H5O routines deleted or just moved around? It looks like the H5Oget_native_info_XXX were removed.

jhendersonHDF (approved):
I did a fair bit of skimming, but since it's a merge, best to get the changes in to make it easier to synchronize things between branches.

java/src/hdf/hdf5lib/H5.java
byrnHDF: Working in a merge view I moved them to the bottom of the section.

derobins (approved):

PR #1981, opened 2022-08-09
Add documentation for parallel compression feature

jhendersonHDF:


byrnHDF (approved):
As a standalone doc it's fine.
If it is moved to under doxygen in the future, then cross refs to the Library could be easily added

jhendersonHDF:
> As a standalone doc it's fine. If it is moved to under doxygen in the future, then cross refs to the Library could be easily added

I believe the idea is to eventually move it into the user's guide once all of that is settled.

derobins (approved):

derobins:
I think interim docs can go in doc/ while we get the UG/RM up to speed wrt newer features.

PR #1992, opened 2022-08-10
Adjustments to t_bigio.c for >2GiB I/O test

jhendersonHDF:
The single_rank_independent_io test currently writes 6GiB rather than 2GiB from a single rank and also doesn't reset the "bigio_count" so creates an unnecessarily complicated derived type. Instead, calculate just the size of dataset we need to cross the INT_MAX threshold for MPI and make sure the bigio_count is reset so we only create a derived type consisting of 2 sub-types.

Really, the test for >2GiB I/O follows a different pathway than the rest of the tests in this file and should probably be in a different test file and/or gated behind HDF_TEST_EXPRESS, but it currently tends to expose bugs in MPI implementations, so it's somewhat useful.

### DIFF
--- a/testpar/t_bigio.c
+++ b/testpar/t_bigio.c
@@ -27,7 +27,6 @@ const char *FILENAME[3] = {"bigio_test.h5", "single_rank_independent_io.h5", NUL
 #define DXFER_COLLECTIVE_IO  0x1 /* Collective IO*/
 #define DXFER_INDEPENDENT_IO 0x2 /* Independent IO collectively */
 #define DXFER_BIGCOUNT       (1 < 29)
-#define LARGE_DIM            1610612736
 
 #define HYPER 1
 #define POINT 2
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

derobins (approved):

testpar/t_bigio.c
```diff
@@ -1135,47 +1139,35 @@ single_rank_independent_io(void)
...
 
         /* Initialize data */
-        for (i = 0; i < LARGE_DIM; i++)
+        for (i = 0; i < dims[0]; i++)
             data[i] = (int)(i % (uint64_t)DXFER_BIGCOUNT);
 
-        if (mpi_rank_g == 0)
-            H5Sselect_all(fspace_id);
-        else
-            H5Sselect_none(fspace_id);
```
fortnern, line 1147: How did this work previously? mspace = ALL, fspace = NONE?

fortnern (approved):

fortnern:
IIRC the independent I/O code doesn't look at the bigio_count setting, but resetting it for this test is probably the right thing to do anyways

jhendersonHDF:
> IIRC the independent I/O code doesn't look at the bigio_count setting, but resetting it for this test is probably the right thing to do anyways

It doesn't follow the same code pathways as the rest of the test code, but eventually down in the MPI I/O driver the code creates types if the I/O size overflows an int. There it was creating on the order of 2.6k sub-types in the derived type vs. only a few needed. Not that it probably makes any difference, but I figured best to be minimal anyway.

PR #2001, opened 2022-08-11
1.12 Correct target name for szaec lib - merge #2000

byrnHDF:
Also sync'd the HDF5config.cmake file - toolchain option move to top.


config/cmake/CTestCustom.cmake
```diff
@@ -20,6 +20,7 @@ set (CTEST_CUSTOM_WARNING_EXCEPTION
     ${CTEST_CUSTOM_WARNING_EXCEPTION}
     "note.*expected.*void.*but argument is of type.*volatile"
     "plugin-build.*:[ \t]*warning"
+    "stamp.verify"
```
derobins, line 23: What is this?

byrnHDF: This file and this section are REGEX for ignoring warnings.

derobins (approved):

PR #2009, opened 2022-08-13
Merge C++ fix to 1.3.2

lrknox:


c++/src/H5File.cpp
```diff
@@ -119,6 +119,53 @@ H5File::H5File(const H5std_string &name, unsigned int flags, const FileCreatProp
     }
 }
 
+//--------------------------------------------------------------------------
+// Function:    H5File overloaded constructor
+///\brief       Opens an HDF5 file using a non-default access property list
+///\param       name         - IN: Name of the file
+///\param       flags        - IN: File access flags
+///\param       access_plist - IN: File access property list.  Default to
+///             FileAccPropList::DEFAULT
```
bmribler, line 128: "Default to FileAccPropList::DEFAULT" need to be removed.

PR #2021, opened 2022-08-17
Removes the bin/gcov_script file

derobins:
We don't need fragile scripts to run basic profiling programs


qkoziol (approved):
RIP kagiso  :-)

jhendersonHDF (approved):

PR #2023, opened 2022-08-17
Removes bin/chkconfigure

derobins:
We check autoconf in configure.ac and can check for anything else in
autogen.sh


lrknox (approved):

qkoziol (changes requested):
If this is possible to check somewhere else, please add that.   Otherwise, this is a valuable sanity check for a certain type of silent errors.

derobins:
> If this is possible to check somewhere else, please add that. Otherwise, this is a valuable sanity check for a certain type of silent errors.

We use AC_PREREQ([2.69]) in configure.ac so this script is redundant. The script doesn't appear to check anything else.

qkoziol:
> > If this is possible to check somewhere else, please add that. Otherwise, this is a valuable sanity check for a certain type of silent errors.
> 
> We use AC_PREREQ([2.69]) in configure.ac so this script is redundant. The script doesn't appear to check anything else.

OK, cool, thanks for checking

qkoziol (approved):

jhendersonHDF (approved):

PR #2026, opened 2022-08-17
Removes the bin/iostats script

derobins:
We don't need to maintain scripts to perform simple tasks

I'm a little more open to keeping this one around, but I'd also be willing to guess that nobody has used this script in over a decade


byrnHDF (approved):

qkoziol (commented):
Dunno about removing this - it's pretty useful to have around.  Maybe move into the h5bench directory?

derobins:
Closing this while we decide its fate

PR #2031, opened 2022-08-17
Persist dataset access properties when using H5Fstart_swmr_write - merge to 1.12 branch

fortnern:
Persist DAPL properties during H5Fstart_swmr_write()

### DIFF
--- a/src/H5Fint.c
+++ b/src/H5Fint.c
@@ -3596,6 +3596,7 @@ H5F__start_swmr_write(H5F_t *f)
     size_t      grp_dset_count = 0;      /* # of open objects: groups & datasets */
     size_t      nt_attr_count  = 0;      /* # of opened named datatypes  + opened attributes */
     hid_t      *obj_ids        = NULL;   /* List of ids */
+    hid_t      *obj_apl_ids    = NULL;   /* List of access property lists */
     H5G_loc_t  *obj_glocs      = NULL;   /* Group location of the object */
     H5O_loc_t  *obj_olocs      = NULL;   /* Object location */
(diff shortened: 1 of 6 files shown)


fortnern:
All tests passed except for codespell (spelling errors in files that this PR doesn't change): https://github.com/HDFGroup/hdf5/actions/runs/2878540588

derobins (approved):

src/H5Fint.c
```diff
@@ -3800,6 +3854,14 @@ H5F__start_swmr_write(H5F_t *f)
     if (obj_paths)
         H5MM_xfree(obj_paths);
 
+    /* Free access property lists */
+    if (obj_apl_ids) {
+        for (u = 0; u < grp_dset_count; u++)
+            if (obj_apl_ids[u] != H5P_DEFAULT && obj_apl_ids[u] >= 0 && H5I_dec_ref(obj_apl_ids[u]) < 0)
+                HGOTO_ERROR(H5E_ATOM, H5E_CANTDEC, FAIL, "decrementing property list ID failed")
```
jhendersonHDF: Should be a DONE_ERROR instead of a GOTO since this is past the done label

derobins: This is also a problem in develop and needs to be fixed there, too

byrnHDF (approved):

src/H5Fint.c
fortnern: Done and (h)done

derobins (approved):

lrknox (approved):

PR #2032, opened 2022-08-17
Fixes codespell issues

derobins:
It remains a mystery why codespell didn't catch these a long time ago...

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -11,5 +11,5 @@ jobs:
       - uses: actions/checkout@v2
       - uses: codespell-project/actions-codespell@master
         with:
-          skip: ./bin/trace,./hl/tools/h5watch/h5watch.c,./tools/test/h5jam/tellub.c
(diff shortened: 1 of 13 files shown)


byrnHDF (approved):

fortnern (approved):

.github/workflows/codespell.yml
```diff
@@ -11,5 +11,5 @@ jobs:
       - uses: actions/checkout@v2
       - uses: codespell-project/actions-codespell@master
         with:
-          skip: ./bin/trace,./hl/tools/h5watch/h5watch.c,./tools/test/h5jam/tellub.c
+          skip: ./bin/trace,./hl/tools/h5watch/h5watch.c,./tools/test/h5jam/tellub.c,./config/sanitizer/LICENSE
```
derobins, line 14: codespell thinks "complies" should actually be "compiles" and I'm disinclined to change the Apache LICENSE

fortnern: it doesn't think "complies" is a word?

fortnern: Or is it trying to infer which it should be from grammar rules or something?

jhendersonHDF (approved):

PR #2034, opened 2022-08-18
Cleanup some warnings in Java JNI code

jhendersonHDF:
Picks off some of the more concerning warnings in the Java JNI code.

@byrnHDF Probably worth testing references after this PR to make sure I didn't make a mistake, but the JNI code was doing some bad things with reference buffers that would certainly cause issues down the line.

### DIFF
--- a/java/src/jni/h5aImp.c
+++ b/java/src/jni/h5aImp.c
@@ -1067,10 +1067,9 @@ Java_hdf_hdf5lib_H5_H5AreadVL(JNIEnv *env, jclass clss, jlong attr_id, jlong mem
     H5T_class_t type_class;
     hsize_t     dims[H5S_MAX_RANK];
     hid_t       sid = H5I_INVALID_HID;
-    jsize       n;
+    jsize       n   = 0;
     htri_t      vl_data_class;
-    herr_t      status = FAIL;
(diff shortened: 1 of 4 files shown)


byrnHDF (approved):
All looks good! 
I believe the JUnit tests would fail if there were any problems with refs

java/src/jni/h5aImp.c
```diff
@@ -1123,16 +1121,18 @@ Java_hdf_hdf5lib_H5_H5AreadVL(JNIEnv *env, jclass clss, jlong attr_id, jlong mem
             H5_LIBRARY_ERROR(ENVONLY);
 
         /* Cache class types */
-        jclass cBool   = ENVPTR->FindClass(ENVONLY, "java/lang/Boolean");
+        /* jclass cBool   = ENVPTR->FindClass(ENVONLY, "java/lang/Boolean"); */
```
derobins, line 1124: Why are these commented out?

jhendersonHDF: They aren't currently used because the boolean type handling code is commented out. I think I asked @byrnHDF about this, but I can't remember how the boolean stuff is handled.

derobins (approved):

PR #2038, opened 2022-08-18
Spelling fixes for codespell

derobins:
Unclear why codespell missed these earlier, but here's the fixes.

### DIFF
--- a/test/cache.c
+++ b/test/cache.c
@@ -9853,15 +9853,6 @@ check_flush_cache__flush_op_eviction_test(H5F_t *file_ptr)
  * Programmer:    John Mainzer
  *              1/12/05
  *
- * Modifications:
- *
- *         JRM -- 3/29/06
- *         Added tests for pinned entries.
- *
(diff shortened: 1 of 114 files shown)


test/cache.c
```diff
@@ -9853,15 +9853,6 @@ check_flush_cache__flush_op_eviction_test(H5F_t *file_ptr)
  * Programmer:    John Mainzer
  *              1/12/05
  *
- * Modifications:
```
fortnern, line 9856: Why were these lines removed?

jhendersonHDF (approved):

test/cache.c
derobins: Modification lines in general are going away. It was easier to delete them than correct them.

bmribler: I know you've been doing that and I heard your reasoning.  It's just sometime I found some good explanation in these Modifications notes very helpful.  I'm not pushing to keep them though...

PR #2039, opened 2022-08-18
Removes config/conclude_fc.am

derobins:
This file set up Autotools Fortran build rules. Automake has no problem
figuring out how to compile Fortran files so this file is no longer
needed.

With this change, Fortran builds now respect silent rules.


derobins:
Note that this doesn't mess with the Fortran dependencies, which are still hard-coded due to long-standing Autotools issues with figuring those out.

bmribler (approved):

derobins:
It looks like the hacks in this file were still necessary in 2015. I may have to close this PR and try the hackier way to get silent rules to work.

derobins:
Closing to try another way of quieting compiler output. Removing the CPP flags is apparently important for the xlf compiler.

PR #2041, opened 2022-08-19
Removes cmake_ext_mod (files moved to cmake)

derobins:
At one time this content was a git submodule shared between several
projects but that time has long passed.


derobins:
This change is @byrnHDF approved, btw.

derobins:
For anyone reviewing:

- cmake_ext_mod files moved to cmake
- Only two files had to be combined: CtestCustom.cmake and ConfigureChecks.cmake
- All other file changes involved path renames

byrnHDF (approved):
We should consider removing the FindSZIP.cmake file as it is out of date and does not even consider the aec szip replacement library.
The other choice is to fix it.

lrknox (approved):

derobins:
> We should consider removing the FindSZIP.cmake file as it is out of date and does not even consider the aec szip replacement library. The other choice is to fix it.

I'll create an issue for that

PR #2046, opened 2022-08-21
Adds a compile-only option to GitHub actions

derobins:
Each test in the matrix has a run_tests field.

Currently, just the -Werror check and a new parallel check run with compile-only, but more will be added soon.


derobins:
Closing. Made this feature branch off of another feature branch. Will redo.

PR #2047, opened 2022-08-21
GitHub workflow cleanup

derobins:
- Combines main.yml and pr-check.yml
- Reduces config redundancy via the matrix scheme
- Adds build-only configs for parallel, no deprec symbols, all API defaults
- The -Werror config has been converted to build-only
- Uniform naming scheme
- main.yml is easier to read and includes helpful comments
- As a bonus, fixes an issue in the mirror VFD test code w/ 1.6 default API


test/mirror_vfd.c
```diff
@@ -1464,8 +1464,8 @@ _create_chunking_ids(hid_t file_id, unsigned min_dset, unsigned max_dset, hsize_
             FAIL_PUTS_ERROR(mesg);
         }
 
-        if ((dataset_ids[m] = H5Dcreate(file_id, dset_name, H5T_STD_I32BE, dataspace_ids[m], H5P_DEFAULT,
-                                        dcpl_id, H5P_DEFAULT)) < 0) {
+        if ((dataset_ids[m] = H5Dcreate2(file_id, dset_name, H5T_STD_I32BE, dataspace_ids[m], H5P_DEFAULT,
+                                         dcpl_id, H5P_DEFAULT)) < 0) {
```
derobins, line 1468: I had to fix this to get the code to compile w/ default v1.6 API

.github/workflows/main.yml
```diff
@@ -4,279 +4,418 @@ name: hdf5 dev CI
 on:
   workflow_dispatch:
   push:
+  pull_request:
```
derobins, line 7: Aside from the name, the 'on' key is the only difference between the two files.

This file has changed so much, it's probably better to look at the before and after individually rather than the diff.

.github/workflows/main.yml
```diff
@@ -4,279 +4,419 @@ name: hdf5 dev CI
...
+      - name: Autotools Run Tests
+        run: make check
+        working-directory: ${{ runner.workspace }}/build
+        if: (matrix.generator == 'autogen') && (matrix.run_tests)
+
+      - name: CMake Run Tests
+        run: ctest --build . -C ${{ matrix.build_mode.cmake }} -V
+        working-directory: ${{ runner.workspace }}/build
+        # Skip Debug MSVC while we investigate H5L Java test timeouts
+        if: (matrix.generator != 'autogen') && (matrix.run_tests) && ! ((matrix.name == 'Windows MSVC CMake') && (matrix.build_mode.cmake == 'Debug'))
```
derobins, line 422: Windows Debug builds time out on H5L Java tests. We currently don't test this, so it's been hidden. We'll investigate and fix (see #2048), but in the meantime we'll just bypass.

derobins:
The yellow tests are due to tests I removed being flagged as 'required' on GitHub. I'll update the 'required' list after this is merged.

byrnHDF (approved):

lrknox (approved):

PR #2053, opened 2022-08-22
Quiets Fortran warnings about large arrays

derobins:


derobins:
Quiets warnings like this:

Warning: Array ‘dset_data’ at (1) is larger than limit set by ‘-fmax-stack-var-size=’, moved from stack to static storage. This makes the procedure unsafe when called recursively, or concurrently from multiple threads. Consider increasing the ‘-fmax-stack-var-size=’ limit (or use ‘-frecursive’, which implies unlimited ‘-fmax-stack-var-size’) - or change the code to use an ALLOCATABLE array. If the variable is never accessed concurrently, this warning can be ignored, and the variable could also be declared with the SAVE attribute. [-Wsurprising]

byrnHDF (approved):

brtnfld (approved):
lgtm

PR #2059, opened 2022-08-23
Corrects more Fortran warnings

derobins:
* Moves -Wunused-dummy-arguments to the developer warnings since this
  generates a lot of spurious, unfixable noise due to things like
  callback wrappers
* Changes the return cast in h5fget_fileno_c, fixes a comment, and comment
  on potential badness
* Initializes a variable in tH5L_F03.F90 to quiet a 'maybe
  uninitialized' warning


config/gnu-warnings/gfort-developer-general
```diff
@@ -0,0 +1,3 @@
+# There is no easy way in Fortran to indicate parameters are intentionally
+# unused and many callbacks will have these, generating a lot of noise.
+-Wunused-dummy-argument
```
derobins, line 3: This generates a LOT of noise due to unused parameters in callbacks

fortran/src/H5Ff.c
```diff
@@ -664,11 +664,15 @@ h5fget_fileno_c(hid_t_f *file_id, int_f *fileno)
...
+     * Call H5Fget_fileno function
      */
     if ((ret_value = H5Fget_fileno((hid_t)*file_id, &fileno_c)) < 0)
         HGOTO_DONE(FAIL);
-    *fileno = (hsize_t_f)fileno_c;
+
+    /* XXX: This will have problems if the library fileno value doesn't fit
+     * into an int_f.
+     */
+    *fileno = (int_f)fileno_c;
```
derobins, line 675: It seems weird that we don't wrap long values properly in Fortran. Maybe we should consider that for HDF5 2.0?

byrnHDF (approved):

fortran/src/H5Ff.c
brtnfld: Unfortunately, this API chose to use integer instead of C_LONG, which got passed me.

brtnfld (approved):
lgtm

PR #2063, opened 2022-08-24
Removes setvbuf() from H5win32defs.h

derobins:
Replaces setvbuf(stream, NULL, _IOLBF, 0) with setbuf(stream, NULL),
which is equivalent, but without the undefined behavior of the 0 size
of the former.


src/H5.c
```diff
@@ -831,36 +831,37 @@ H5__debug_mask(const char *s)
...
+            }
         }
         else if (HDisdigit(*s)) {
             int                     fd = (int)HDstrtol(s, &rest, 0);
             H5_debug_open_stream_t *open_stream;
 
             if ((stream = HDfdopen(fd, "w")) != NULL) {
-                (void)HDsetvbuf(stream, NULL, _IOLBF, (size_t)0);
+                /* This is for debugging info, so turn off buffering */
+                (void)HDsetbuf(stream, NULL);
```
derobins, line 846: The size of zero is undefined behavior and makes MSVC sad.

PR #2074, opened 2022-08-27
Align arg types of H5D_chunk_iter_op_t with H5Dget_chunk_info

### LINKED ISSUE #2056: [BUG] H5D_chunk_iter_op_t exposes `size` (`nbytes`) as `uint32_t` rather than `hsize_t`
**Describe the bug**
[`H5D_chunk_iter_op_t`](https://docs.hdfgroup.org/hdf5/develop/_h5_dpublic_8h.html#ac5da85e7fc125ab783d7796afecd7f45) exposes the argument `size` (formerly `nbytes`) as `uint32_t` rather than `hsize_t`.

**Expected behavior**
The type of `size` should be `hsize_t` as in [`H5Dget_chunk_info`](https://docs.hdfgroup.org/hdf5/develop/group___h5_d.html#gaccff213d3e0765b86f66d08dd9959807).

 - 1.13, 1.12
 - All OSes, compilers, build systems

**Additional context**
`nbytes` was renamed to `size` in https://github.com/HDFGroup/hdf5/pull/1969

mkitti:
Fix #2056 

For `H5D_chunk_iter_op_t`
* `filter_mask` changes from type `uint32_t` to `unsigned`
* `size` changes from `uint32_t` to `hsize_t`

The practical change is that `size` argument is now a [64-bit unsigned integer](https://github.com/HDFGroup/hdf5/blob/08aebb6d1af481272bda391f19e03c330fddf149/src/H5public.h#L292).

These changes make `H5D_chunk_iter_op_t` similar the argument similar to the `size` argument in [`H5Dget_chunk_info`](https://github.com/HDFGroup/hdf5/blob/08aebb6d1af481272bda391f19e03c330fddf149/src/H5Dpublic.h#L736-L737).

```C
H5_DLL herr_t H5Dget_chunk_info(hid_t dset_id, hid_t fspace_id, hsize_t chk_idx, hsize_t *offset,
                                unsigned *filter_mask, haddr_t *addr, hsize_t *size);
```

The change is important because it allows the internal chunk size to change to be a 64-bit value in the future without modifying the public API.

### DIFF
--- a/src/H5D.c
+++ b/src/H5D.c
@@ -2515,18 +2515,18 @@ done:
  *
  *        typedef int (*H5D_chunk_iter_op_t)(
  *            const hsize_t *offset,
- *            uint32_t filter_mask,
+ *            unsigned filter_mask,
  *            haddr_t addr,
- *            uint32_t nbytes,
+ *            hsize_t size,
  *            void *op_data);
  *
  *      H5D_chunk_iter_op_t parameters:
- *          hsize_t *offset;        IN/OUT: Array of starting logical coordinates of chunk.
- *          uint32_t filter_mask;   IN: Filter mask of chunk.
- *          haddr_t addr;           IN: Offset in file of chunk data.
- *          uint32_t nbytes;        IN: Size in number of bytes of chunk data in file.
- *          void *op_data;          IN/OUT: Pointer to any user-defined data
- *                                  associated with the operation.
+ *          hsize_t *offset;        IN/OUT: Logical position of the chunk’s first element in units of dataset
+ *                                          elements
+ *          unsigned filter_mask;   IN: Bitmask indicating the filters used when the chunk was written haddr_t
+ *          addr;                   IN: Chunk address in the file
+ *          hsize_t;                IN: Chunk size in bytes, 0 if the chunk does not exist
+ *          void *op_data;          IN/OUT: Pointer to any user-defined data associated with the operation.
  *
  *      The return values from an operator are:
  *          Zero (H5_ITER_CONT) causes the iterator to continue, returning zero when all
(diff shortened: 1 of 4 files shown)


derobins:
> The change is important because it allows the internal chunk size to change to be a 64-bit value in the future without modifying the public API.

The problem with this, though, is that using a 64-bit value now makes the function signature misleading until we update the file format. That will definitely not happen in 1.14.0, so we should save this change until HDF5 2.0 (or whenever we get around to updating the file format).

mkitti:
I find the inconsistency with the rest of the public chunk API more problematic at the present.

The mismatch between `hsize_t` and the current maximum chunk size has existed for as long as `hsize_t` became a 64-bit integer.

derobins:
While I agree with the sentiment behind this PR, H5D_chunk_iter_op_t already exists in a released version of HDF5, so we'd have to run these changes through our backward compatibility scheme.

Since that's kind of a big deal and I'm unhappy with the types anyway, I'm going to label this WIP/draft and we can turn this into a fix in HDF5 2.0, where I'd like to introduce a chunk dimension size type and we can change the filter mask to a fixed size type.

derobins:
I'll approve this and we can merge this in, but:

(A) This HAS to go to 1.12 ASAP since the chunk iteration stuff was merged there (but not released in 1.12.2).
(B) I'm not excited by the types in these functions and we should update them in 2.0 with a chunk dimension type and a fixed-size mask type.

derobins (approved):

mkitti:
I rebased on develop since it's been a while.

fortnern (approved):

mkitti:
I cherry picked this on the hdf5_1_12 branch: https://github.com/HDFGroup/hdf5/pull/2329

PR #2078, opened 2022-08-29
Minor changes to CMake comments, etc.

derobins:
Minor tidying of things noticed while documenting CMake files

### DIFF
--- a/config/cmake/mccacheinit.cmake
+++ b/config/cmake/mccacheinit.cmake
@@ -11,9 +11,9 @@
 #
 # This is the CMakeCache file.
 
-########################
+#########################
 # EXTERNAL cache entries
-########################
+#########################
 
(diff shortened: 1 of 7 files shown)


config/cmake/mccacheinit.cmake
```diff
@@ -11,9 +11,9 @@
 #
 # This is the CMakeCache file.
 
-########################
+#########################
 # EXTERNAL cache entries
```
byrnHDF, line 15: mc stands for "memory check".

byrnHDF:
One minor note on usage of case in naming. There are some packaging conventions that use lower case intentionally.
(Nothing in this PR)

config/cmake/mccacheinit.cmake
lrknox: So line 15 should be "# EXTERNAL cache entries for memory checking"?

byrnHDF: Or something like it.

byrnHDF (approved):

lrknox (approved):

PR #2080, opened 2022-08-31
Update CMake code to 3.18 minimum version

byrnHDF:
Two types of changes the cmake_minimum_required line changed the version from 3.12 to 3.18
And message command checks for at least 3.15 have been removed.


release_docs/RELEASE.txt
```diff
@@ -47,6 +47,13 @@ New Features
 
     Configuration:
     -------------
+    - Update CMake minimum version to 3.18
+
+      Updated CMake minimum version to 3.12 and removed version checks
```
```suggestion
      Increased CMake minimum version from 3.12 to 3.18 and removed version checks for features that required versions less than 3.18. 
```

lrknox (approved):

derobins:
Why is the version check duplicated everywhere? Shouldn't that just be done once? We can merge this, but that's a LOT of repetitive build code.

derobins (approved):

Issue #2082, opened 2022-09-01
[Feature Request] Remove bitbucket.hdfgroup.org dependency.


hyoklee:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**

[cacheinit.cmake](https://github.com/HDFGroup/hdf5/blob/develop/config/cmake/cacheinit.cmake) still refers to bitbucket.

```
set (BLOSC_ZLIB_GIT_URL "https://git@bitbucket.hdfgroup.org/scm/test/zlib.git" CACHE STRING "Use BLOSC_ZLIB from HDF Group repository" FORCE)
```

I hope it's migrated to GitHub.

**Describe alternatives you've considered**

Points to the original source project repo. and apply patches on-demand for HDF5's needs.

**Additional context**

What happens if bitbucket.hdfgroup.org is down?

byrnHDF:
Actually those repos never moved because we use the tgz files in the plugins lib folder.
The bitbucket refs will need to be removed from the plugin code also.

byrnHDF:
#2086 created

PR #2086, opened 2022-09-06
Remove bitbucket.hdfgroup.org dependency. #2082

byrnHDF:


hyoklee (changes requested):
Remove git@ from git@someurl.

hyoklee (approved):

lrknox (approved):

PR #2087, opened 2022-09-06
Change `test` to `example`.

hyoklee:
Or should they be more descriptive like below?

```
         -D "TEST_REFERENCE:STRING=PHDF5 example reference test finished with no errors"
          #-D "TEST_FILTER:STRING=PHDF5 example filter test finished with no errors"
```

### DIFF
--- a/examples/CMakeTests.cmake
+++ b/examples/CMakeTests.cmake
@@ -99,7 +99,6 @@ if (HDF5_TEST_SERIAL)
           -D "TEST_EXPECT=0"
           -D "TEST_SKIP_COMPARE=TRUE"
           -D "TEST_OUTPUT=${example}.txt"
-          #-D "TEST_REFERENCE=${example}.out"
           -D "TEST_FOLDER=${PROJECT_BINARY_DIR}"
           -P "${HDF_RESOURCES_DIR}/runTest.cmake"
       )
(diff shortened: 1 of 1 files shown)


examples/CMakeTests.cmake
```diff
@@ -129,7 +129,7 @@ if (H5_HAVE_PARALLEL AND HDF5_TEST_PARALLEL AND NOT WIN32)
           -D "TEST_SKIP_COMPARE=TRUE"
           -D "TEST_OUTPUT=${parallel_example}.out"
           -D "TEST_REFERENCE:STRING=PHDF5 example finished with no errors"
-          #-D "TEST_FILTER:STRING=PHDF5 tests finished with no errors"
+          #-D "TEST_FILTER:STRING=PHDF5 example finished with no errors"
```
byrnHDF: It is disabled and probably should be removed - likely an artifact of the creation process.

byrnHDF (approved):

lrknox (commented):
commented line 102 was removed;  commented line 131 is currently kept as changed.  Is this intended, or should line 131 also be removed?

hyoklee:
Please remove 131. Good catch!

examples/CMakeTests.cmake
```diff
@@ -129,7 +128,7 @@ if (H5_HAVE_PARALLEL AND HDF5_TEST_PARALLEL AND NOT WIN32)
...
```
lrknox:
```suggestion
```

lrknox (approved):

PR #2096, opened 2022-09-12
Fix memory bug in selection I/O

fortnern:
H5FD_read/write_selection() is passed an array of H5S_t *s and in some cases needs to create to pass an array of hid_ts referencing these H5S_t *s to the underlying callback. The function does this using H5I_register(). When freeing memory at the end of the function, it would previously call H5I_dec_app_ref on the ids, which had the side effect of freeing the H5S_t *s, even though they were not created by H5FD_read/write_selection(), causing problems when the calling function(s) tried to use these spaces afterwards. Changed to H5I_remove().

### DIFF
--- a/src/H5FDint.c
+++ b/src/H5FDint.c
@@ -1171,7 +1171,7 @@ H5FD_read_selection(H5FD_t *file, H5FD_mem_t type, uint32_t count, H5S_t **mem_s
             if ((file_space_ids[num_spaces] = H5I_register(H5I_DATASPACE, file_spaces[num_spaces], TRUE)) <
                 0) {
                 if (NULL == H5I_remove(mem_space_ids[num_spaces]))
-                    HDONE_ERROR(H5E_VFL, H5E_CANTDEC, FAIL, "problem freeing id")
+                    HDONE_ERROR(H5E_VFL, H5E_CANTREMOVE, FAIL, "problem removing id")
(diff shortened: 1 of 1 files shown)


src/H5FDint.c
```diff
@@ -1170,7 +1170,7 @@ H5FD_read_selection(H5FD_t *file, H5FD_mem_t type, uint32_t count, H5S_t **mem_s
 
             if ((file_space_ids[num_spaces] = H5I_register(H5I_DATASPACE, file_spaces[num_spaces], TRUE)) <
                 0) {
-                if (H5I_dec_app_ref(mem_space_ids[num_spaces]) < 0)
+                if (NULL == H5I_remove(mem_space_ids[num_spaces]))
                     HDONE_ERROR(H5E_VFL, H5E_CANTDEC, FAIL, "problem freeing id")
```
bmribler: Does error code/message need to be changed accordingly?  Same below.

fortnern: I think it's still valid - the ID itself is still being freed.

fortnern: Maybe H5E_CANTDEC should be changed

bmribler (approved):
Thank you!

byrnHDF (approved):

jhendersonHDF (approved):

PR #2101, opened 2022-09-14
Use case-insensitive comparison for headers fix #2100

### LINKED ISSUE #2100: ROS3 header search is case-sensitive
**Describe the bug**
Looking up the content length header in the ROS3 VFD is case-sensitive at [H5FDs3comms.c:944](https://github.com/HDFGroup/hdf5/blob/05a0411140688722a39855dcfa902c64bf6215a9/src/H5FDs3comms.c#L944). That means, ROS3 finds `Content-Length: ...`, but fails to find `content-length: ...`.

**Expected behavior**
Headers are supposed to be case-insensitive ASCII strings ([RFC 7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.2), [RFC 7540](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2)). So, both `Content-Length: ...` and `content-length: ...` should be matched.

 - HDF5 version 1.12.2
 - OS and version: Debian 11
 - Compiler and version: gcc (Debian 10.2.1-6) 10.2.1 20210110
 - Build system: Package: autotools-dev Version: 20180224.1+nmu1
 - Any configure options you specified: --enable-ros3_vfd

**Additional context**
Opening files using ROS3 VFD from a Ceph implementation of S3 fails. Opening the same file/object from AWS works. The reported error is that the 'Content-Length' header cannot be found. Indeed, Ceph advertises `content-length` while AWS advertises `Content-Length`:

```
$ curl --head "https://cephs3.../hdf5/snp500.h5"
HTTP/1.1 200 OK
content-length: 113404160
...
```

```
$ curl --head "http://aws.../hdf5/snp500.h5"
HTTP/1.1 200 OK
Content-Length: 113404160
```

It is rather unfortunate that case-insensitive string search is non-standard.

berghaus:
Use a case-insensitive search to locate header information. As this is a non-standard extension of C select the appropriate function depending on operating system and define a common alias:

- Microsoft docs suggest [StrStrIA](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-strstria) should be selected on Windows
- Linux docs point to [strcasestr](https://man7.org/linux/man-pages/man3/strcasestr.3.html)
- Apple docs suggest the same convention as Linux [here](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/strcasestr.3.html)

*Disclaimer:* I only tested on Linux (Debian 11).


src/H5private.h
```diff
@@ -1394,6 +1394,13 @@ H5_DLL H5_ATTR_CONST int Nflock(int fd, int operation);
 #ifndef HDstrcasecmp
 #define HDstrcasecmp(X, Y) strcasecmp(X, Y)
 #endif
+#ifndef HDstrcasestr
+#ifdef _WIN32
+#define HDstrcasestr(X, Y) StrStrIA(X, Y)
+#else
+#define HDstrcasestr(X, Y) strcasestr(X, Y)
```
jhendersonHDF: Hey @berghaus, two things here:

- The #define HDstrcasestr for Windows should be moved into H5win32defs.h since that's where all the Windows compatibility stuff goes

- Though it's been around long enough that there probably aren't many (or any) systems where we'd need to worry about having strcasestr, it would probably be better and more portable for HDstrcasestr to point to an implemented function, say H5strcasestr, so we guarantee we have the functionality across platforms.

berghaus: Thank you for the feedback @jhendersonHDF :-)

Regarding your second point: I had been considering this as well - but ran into trouble when dealing with localization. Here would be a possible solution, but I am unsure where to best put this into the existing code [1]:
```c
char *
H5strcasestr(const char *haystack, const char* needle)
{
    /* Check arguments. */
    HDassert(haystack);
    HDassert(needle);

    /* begin once from each character of haystack, until needle is found */
    do {
        const char *h = haystack;
        const char *n = needle;
        /* loop while lowercase strings match, or needle ends */
        while (HDtolower(*h) == HDtolower(*n) && *n) {
            h++;
            n++;
        }
        /* if all characters in needle matched we found it */
        if (*n == 0) {
            return (char *)haystack;
        }
    } while (*haystack++);
    return 0;
} /* end H5strcasestr() */
```

I am a bit uncertain where to best place this, and how to decorate it. I was trying to follow what is done with `H5_timer_get_time_string` as an example - but I still get linking errors when trying. Shall I mark this as a draft and push this code, so you can more concretely see what I am doing wrong?

[1] Inspired by [this discussion](https://stackoverflow.com/questions/27303062/strstr-function-like-that-ignores-upper-or-lower-case).

jhendersonHDF: > I am a bit uncertain where to best place this

Though not the most ideal place, I'd say the best candidate currently would be for this to go into H5system.c and then the prototype for the function would go in H5private.h towards the bottom (making sure that the prototype is marked with H5_DLL).

> Shall I mark this as a draft and push this code, so you can more concretely see what I am doing wrong?

Sure, it never hurts to have the current code here to view! Your implementation looks fine to me though, so I guess we just need to figure out why you end up with linking errors.

berghaus: > Sure, it never hurts to have the current code here to view! Your implementation looks fine to me though, so I guess we just need to figure out why you end up with linking errors.

I made a mistake, trying to add a new source file in the make instructions. Following your suggestion, the linking errors went away :-)

src/H5system.c
```diff
@@ -1350,3 +1350,42 @@ H5_get_option(int argc, const char *const *argv, const char *opts, const struct
...
+ *              Failure:  NULL
+ *
+ * Programmer:  Frank Berghaus
+ *              Sep 2022
+ *
+ *-------------------------------------------------------------------------
+ */
+char *
+H5strcasestr(const char *haystack, const char *needle)
+{
```
jhendersonHDF, line 1368: @berghaus ~~Looks like this function will need to be updated with FUNC_ENTER and FUNC_LEAVE. In this case, it's probably fine to use FUNC_ENTER_NOAPI_NOINIT. You can refer to H5_basename and H5_dirname in this same file for an example how to instrument the function.~~

On second thought, renaming the function to H5_strcasestr since it's for internal use only should probably be sufficient.

src/H5private.h
```diff
@@ -593,6 +593,7 @@ H5_DLL herr_t   H5_timer_stop(H5_timer_t *timer /*in,out*/);
 H5_DLL herr_t   H5_timer_get_times(H5_timer_t timer, H5_timevals_t *times /*in,out*/);
 H5_DLL herr_t   H5_timer_get_total_times(H5_timer_t timer, H5_timevals_t *times /*in,out*/);
 H5_DLL char    *H5_timer_get_time_string(double seconds);
+H5_DLL char    *H5strcasestr(const char *haystack, const char *needle);
```
derobins: Private functions need an underscore after the 'H5', so H5_strcasestr().

src/H5system.c
```diff
@@ -1350,3 +1350,42 @@ H5_get_option(int argc, const char *const *argv, const char *opts, const struct
...
+ *
+ * Purpose:     Find the first occurrence of the substring needle in the
+ *              string haystack ignoring case.
+ *
+ * Return:      Success:  Pointer to the beginning of the located substring
+ *
+ *              Failure:  NULL
+ *
+ * Programmer:  Frank Berghaus
+ *              Sep 2022
```
derobins: Although it's still common in the library, we're not adding author information to new code. Existing author comments will eventually be removed.

src/H5private.h
```diff
@@ -1385,6 +1386,9 @@ H5_DLL H5_ATTR_CONST int Nflock(int fd, int operation);
 #ifndef HDstrcat
 #define HDstrcat(X, Y) strcat(X, Y)
 #endif
+#ifndef HDstrcasestr
+#define HDstrcasestr(X, Y) H5strcasestr(X, Y)
+#endif
```
derobins, line 1395: The code here would use  H5_strcasestr() on all platforms, which is not what we want.

What you want to do is check for strcasestr in the configure files (both Autotools and CMake) and then the code here will look something like:

```
#ifndef HDstrcasestr
#if defined(H5_HAVE_STRCASESTR)                                                       
#define HDstrcasestr(X, Y) strcasestr(X, Y)                                                                                     
#else                                                                            
#define HDstrcasestr(X, Y) H5_strcasestr(X, Y)                                              
#endif
#endif
```
This defines the macro to be the C library call if we found it or the new private function if not (both only in the case that we didn't do something else in H5win32defs.h or some other OS-specific file in the future).

derobins: You'll also want to #define HDstrcasestr() to StrStrlA or whatnot in H5win32defs.h so this call will use the most "native" thing on Windows. The GitHub actions will test this on Windows if you don't have a Windows machine handy.

src/H5system.c
```diff
@@ -1350,3 +1350,42 @@ H5_get_option(int argc, const char *const *argv, const char *opts, const struct
...
+            h++;
+            n++;
+        }
+        /* if all characters in needle matched we found it */
+        if (*n == 0) {
+            return (char *)haystack;
+        }
+    } while (*haystack++);
+    return 0;
+} /* end H5strcasestr() */
```
derobins: This function needs a quick test in th5_system.c to ensure it works properly as a substitute on all platforms. You should test this function explicitly and HDstrcasestr to make sure the redefinition works.

src/H5FDs3comms.c
```diff
@@ -941,7 +941,7 @@ H5FD_s3comms_s3r_getsize(s3r_t *handle)
      * PARSE RESPONSE *
      ******************/
 
-    start = HDstrstr(headerresponse, "\r\nContent-Length: ");
+    start = HDstrcasestr(headerresponse, "\r\nContent-Length: ");
```
derobins, line 944: Note that the read-only S3 VFD is currently neither built nor tested via GitHub actions at this time. We should ask @lrknox to test this independently (even though it's a trivial change).

This is also a reminder to us to set up ros3 VFD testing in the very near future.

src/H5system.c
derobins: I don't think we need to use FUNC_ENTER macros here. Most of the replacement functions don't use them and I plan to rip out the name checks in the future anyway.

src/H5private.h
derobins: For the Autotools, look in configure.ac and add it to the block of AC_CHECK_FUNCS (note that they are in alphabetical order).

derobins: For CMake, add it to the CHECK_FUNCTION_EXISTS section in config/cmake/ConfigureChecks.cmake (around line 450) and also add a section to config/cmake/H5pubconf.h.in that matches the Autotools output (we try to keep the pubconf headers identical between the Autotools and CMake).

derobins: If strcasestr were POSIX, we'd probably decorate the replacement function with a 'W' (Wstrcasestr), not do the function tests, and only replace on Windows, but since it's "nonstandard" instead of POSIX we have to worry about it not being present on uncommon Unix-y platforms and the H5_strcasestr() scheme w/ configure checks is the right way to go.

derobins:
The reason this is failing, btw, is that when we run autogen.sh the bin/trace script attempts to add trace information to public API calls and your new function lacks an underscore after the H5, making it look like a public API call, but with no function enter macro.

src/H5FDs3comms.c
berghaus: In addition to the regular requirements, I needed to install `libcurl4-openssl-dev` to build and test - in case that helps :-)

berghaus:
Thank you for all the feedback and help to get this done @derobins. I followed through on each of the suggestions.

To avoid conflicts rebased on the current head. I am a little surprised how dirty that makes this PR look - so if you want me to close and restart to remove that noise, I am happy to.

berghaus:
> I am a little surprised how dirty that makes this PR look

Fixed it. Must have missed something over the weekend :-)

test/th5_system.c
```diff
@@ -437,6 +437,31 @@ test_h5_basename(void)
     HDfree(path);
 }
 
+static void
+test_h5_strcasestr(void)
+{
+    const char *const haystack = "My test string";
+    char * str = NULL;
```
jhendersonHDF: Almost there! Seems clang-format is unhappy,

```
--- ./test/th5_system.c	(original)
+++ ./test/th5_system.c	(reformatted)
@@ -441,7 +441,7 @@
 test_h5_strcasestr(void)
 {
     const char *const haystack = "My test string";
-    char * str = NULL;
+    char             *str      = NULL;
 
     MESSAGE(5, ("Testing H5_strcasestr\n"));
```

test/th5_system.c
```diff
@@ -437,6 +437,31 @@ test_h5_basename(void)
...
+
+    /* Check that H5_strcasestr find a string of same case */
+    str = H5_strcasestr(haystack, "string");
+    CHECK_PTR_EQ(str, &(haystack[8]), "H5_strcasestr search same case");
+
+    /* Check that H5_strcasestr find a string of different case */
+    str = H5_strcasestr(haystack, "sTrInG");
+    CHECK_PTR_EQ(str, &(haystack[8]), "H5_strcasestr search different case");
+
+    /* Check that H5_strcasestr find a string of different case */
```
jhendersonHDF: Minor nitpick - inaccurate comment

berghaus: Good catch! I admit it, I copied and pasted :-)

test/th5_system.c
berghaus: Let's hope I got it this time. Running clang-format on my machine ended up with different choices where the '*' should go :facepalm:

lrknox:
Note also this cast discards 'const' error in https://github.com/HDFGroup/hdf5/actions/runs/3090306304/jobs/4999421418#step:13:33:
/home/runner/work/hdf5/hdf5/src/H5system.c: In function ‘H5_strcasestr’:
/home/runner/work/hdf5/hdf5/src/H5system.c:1384:20: error: cast discards ‘const’ qualifier from pointer target type [-Werror=cast-qual]
 1384 |             return (char *)haystack;

jhendersonHDF:
> Note also this cast discards 'const' error in https://github.com/HDFGroup/hdf5/actions/runs/3090306304/jobs/4999421418#step:13:33: /home/runner/work/hdf5/hdf5/src/H5system.c: In function ‘H5_strcasestr’: /home/runner/work/hdf5/hdf5/src/H5system.c:1384:20: error: cast discards ‘const’ qualifier from pointer target type [-Werror=cast-qual] 1384 | return (char *)haystack;

Yes, this is an unfortunate issue with the prototype for strcasestr. @berghaus you'll probably need to silence the warning here using the H5_GCC_CLANG_DIAG_OFF/H5_GCC_CLANG_DIAG_ON macros. See https://github.com/HDFGroup/hdf5/blob/develop/src/H5Gname.c#L449-L451 for example.

jhendersonHDF (approved):

src/H5FDs3comms.c
lrknox: Tested berghaus:develop with ros3 enabled and S3 bucket for tests configured.  All tests passed.

lrknox (approved):

byrnHDF:
Fixed window linker error: #2114

PR #2106, opened 2022-09-16
Subfiling VFD updates

jhendersonHDF:
This PR contains all the Subfiling development work since its release in 1.13.2. It mostly involves performance improvements, but also contains additional regression testing and the ability for a Subfiling I/O concentrator to control more than 1 subfile, which allows reading back files with a different number of MPI ranks than they were created with.

There's quite a bit to review here, so it's probably much easier to review by commit.


testpar/t_subfiling_vfd.c
```diff
@@ -138,103 +180,1765 @@ create_subfiling_ioc_fapl(void)
...
+
+    curr_nerrors = nerrors;
+
+    if (MAINPROCESS)
+        TESTING_2("h5fuse utility");
+
+#if defined(H5_HAVE_FORK) && defined(H5_HAVE_WAITPID)
+
+    /*
+     * Check if h5fuse script exists in current directory;
```
jhendersonHDF, line 1487: @brtnfld I updated the CMake code to copy the h5fuse.sh script into the testpar directory and setup this test to only run if it's available. I wasn't sure what the best way to accomplish that for Autotools was, but that may be a todo item.

utils/subfiling_vfd/CMakeLists.txt
```diff
@@ -1,10 +1,20 @@
 cmake_minimum_required (VERSION 3.10)
```
byrnHDF, line 1: develop cmake minimum is 3.18 now.

utils/subfiling_vfd/CMakeLists.txt
```diff
@@ -1,10 +1,20 @@
...
 project (HDF5_UTILS_SUBFILINGVFD C)
 
-configure_file (${HDF5_UTILS_SUBFILINGVFD_SOURCE_DIR}/h5fuse.sh.in ${HDF5_BINARY_DIR}/h5fuse.sh @ONLY)
+configure_file (${HDF5_UTILS_SUBFILINGVFD_SOURCE_DIR}/h5fuse.sh.in ${HDF5_UTILS_SUBFILINGVFD_BINARY_DIR}/h5fuse.sh @ONLY)
```
byrnHDF, line 4: Optional: Note that HDF5_UTILS_SUBFILINGVFD_SOURCE(BINARY)_DIR could also be PROJECT_SOURCE(BINARY)_DIR. I do like the specific name though.

byrnHDF (approved):

examples/ph5_subfiling.c
```diff
@@ -35,12 +35,13 @@
...
 #define EXAMPLE_FILE  "h5_subfiling_default_example.h5"
 #define EXAMPLE_FILE2 "h5_subfiling_custom_example.h5"
+#define EXAMPLE_FILE3 "h5_subfiling_precreate_example.h5"
 
 #define EXAMPLE_DSET_NAME "DSET"
 #define EXAMPLE_DSET_DIMS 2
 
-/* Have each MPI rank write 64MiB of data */
-#define EXAMPLE_DSET_NY 16777216
+/* Have each MPI rank write 16MiB of data */
```
brtnfld, line 43: 4MiB

testpar/t_subfiling_vfd.c
brtnfld: Can you create a to-do issue.

brtnfld (approved):
lgtm

examples/ph5_subfiling.c
brtnfld: never mind, is correct.

PR #2114, opened 2022-09-22
Add missing windows include/lib configure for StrStrIA

byrnHDF:
Linker fix for [ROS3 header search is case-sensitive] #2100


lrknox (approved):

jhendersonHDF (approved):

config/cmake/ConfigureChecks.cmake
```diff
@@ -154,6 +157,10 @@ if (MINGW OR NOT WINDOWS)
   CHECK_LIBRARY_EXISTS_CONCAT ("wsock32" gethostbyname ${HDF_PREFIX}_HAVE_LIBWSOCK32)
 endif ()
 
+if (WINDOWS)
+  CHECK_LIBRARY_EXISTS_CONCAT ("shlwapi" StrStrIA ${HDF_PREFIX}_HAVE_SHLWAPI)
```
jhendersonHDF, line 161: Once this is merged, I should probably update the H5win32defs header to check for this before using StrStrIA and use H5_strcasestr if it's not available.

byrnHDF: Good idea.

lrknox: The test in #2115 failed because it is now testing for this.  I'll go head and merge this one.

PR #2115, opened 2022-09-22
Add test for HDstrcasestr macro

jhendersonHDF:


byrnHDF (approved):

test/th5_system.c
```diff
@@ -447,21 +447,58 @@ test_h5_strcasestr(void)
...
+    CHECK_PTR(str, "H5_strcasestr for substring of same case");
+    if (str)
+        VERIFY_STR(str, "string", "comparing H5_strcasestr for substring of same case");
 
     /* Check that H5_strcasestr find a string of different case */
     str = H5_strcasestr(haystack, "sTrInG");
-    CHECK_PTR_EQ(str, &(haystack[8]), "H5_strcasestr search different case");
+    CHECK_PTR(str, "H5_strcasestr for substring of different case");
+    if (str)
+        VERIFY_STR(str, "string", "comparing H5_strcasestr for substring of different case");
```
byrnHDF, line 464: I think the 2nd parameter should be "sTrInG"?

test/th5_system.c
```diff
@@ -447,21 +447,58 @@ test_h5_strcasestr(void)
...
+    str = HDstrcasestr(haystack, "string");
+    CHECK_PTR(str, "HDstrcasestr for substring of same case");
+    if (str)
+        VERIFY_STR(str, "string", "comparing HDstrcasestr for substring of same case");
+
+    /* Check that HDstrcasestr find a string of different case */
+    str = HDstrcasestr(haystack, "sTrInG");
+    CHECK_PTR(str, "HDstrcasestr for substring of different case");
+    if (str)
+        VERIFY_STR(str, "string", "comparing HDstrcasestr for substring of different case");
```
byrnHDF, line 495: Here too.

test/th5_system.c
jhendersonHDF: In `VERIFY_STR` we're checking to make sure that `H5_strcasestr` returns the actual original string of "string" from `haystack`. We just pass "sTrInG" to `H5_strcasestr` to make sure that it can find that portion even though the search uses a string with weird casing to try and find it.

jhendersonHDF:
This PR is ready for review again. I had to add a wrapper function for windows because StrStrIA does not exhibit the same GNU behavior as strcasestr where searching for an empty substring returns the input string.

byrnHDF (approved):

derobins (approved):

PR #2134, opened 2022-09-28
Fix mostly java warnings, and cpp cast

byrnHDF:


test/tmisc.c
```diff
@@ -3978,7 +3978,7 @@ test_misc22(void)
                 if ((prec[j] + offsets[k]) > (H5Tget_size(idts[i]) * 8))
                     continue;
 
-                MESSAGE(5, ("  Testing datatypes size=%d precision=%u offset=%d\n", H5Tget_size(idts[i]),
```
jhendersonHDF, line 3981: H5Tget_size returns a size_t so that specifier should be %zu here. Then, since the offsets entry is cast to an unsigned, that specifier should either be %u or the cast to unsigned should be removed and the correct specifier should be used.

java/src/jni/h5jni.h
```diff
@@ -257,6 +257,23 @@
     do {                                                                                                     \
         (*envptr)->ReleaseStringUTFChars(envptr, pinnedString, stringToRelease);                             \
     } while (0)
+/*
+ * Above String macros may be incorrect, suggested code for getting a cstr from java
+ * int jstr_to_cstr(JNIEnv *jenv, jstring j_str, char *c_str, size_t cstr_len)
```
jhendersonHDF, line 262: We should decide what to do about this instead of just adding a commented out section of code

byrnHDF: I ran across this info in a post, and this was outside the scope of this PR. So I put in the comment for near future investigation.

test/tmisc.c
byrnHDF: The compiler suggested %ld, but I can change it.

jhendersonHDF (approved):

lrknox (approved):

PR #2139, opened 2022-10-03
Fix bugs in multi-dataset I/O.  Expand serial multi dataset test.

fortnern:
Update macro in parallel multi dataset test.


src/H5Dcompact.c
```diff
@@ -416,7 +416,7 @@ H5D__compact_writevv(const H5D_io_info_t *io_info, const H5D_dset_io_info_t *dse
     }
 
     /* Mark the compact dataset's buffer as dirty */
-    *io_info->dsets_info[0].store->compact.dirty = TRUE;
+    *dset_info->store->compact.dirty = TRUE;
```
bmribler, line 420: Just checking...  *dset_info, not dset_info?

PR #2143, opened 2022-10-04
Subfiling Fortran wrapper work.

brtnfld:
Added subfiling and ioc FD Fortran wrappers with tests. Added H5Pget/set_mpi_params with tests. 

CMake tests does not yet work, CMake fortran testpar is not working in general.


brtnfld:
CMake issues has been resolved.

brtnfld:
The only outstanding issue is the check for MPI_Init_thread with MPI_THREAD_MULTIPLE when calling h5open_f, even when the subfiling VFD is not being used.

byrnHDF (approved):

derobins (approved):

PR #2147, opened 2022-10-05
1.10 Fix some Java warnings

byrnHDF:


java/src/jni/h5util.c
```diff
@@ -717,7 +717,7 @@ size_t
 h5str_sprintf(JNIEnv *env, h5str_t *out_str, hid_t container, hid_t tid, void *in_buf, int expand_data)
 {
     unsigned char *ucptr = (unsigned char *)in_buf;
-    static char    fmt_llong[8], fmt_ullong[8];
```
jhendersonHDF, line 720: Was there a reason for removing static from this? This might cause problems with formatting since the buffers were static and would persist between function calls before this was changed.

byrnHDF: It was removed at some point in time in develop - so I merged the change down.

jhendersonHDF: Looks like it was removed in https://github.com/HDFGroup/hdf5/commit/4ac6ccc0213767be935b2686524937ae93c46e3f but that was probably wrong.

byrnHDF: I think this comment from that commit is why: 
Removed unnecessary use of static on small buffers

byrnHDF: BTW, this function is only used to export a dataset from java.

jhendersonHDF: > I think this comment from that commit is why: Removed unnecessary use of static on small buffers

Right it was done to help thread safety but I think this was a mistake here.

jhendersonHDF: > BTW, this function is only used to export a dataset from java.

There are also other places where we do the same thing and static was removed. We probably need to look at those again.

jhendersonHDF (approved):

PR #2153, opened 2022-10-07
Fix for HDFFV-10840: Instead of using fill->buf for datatype conversion

vchoi-hdfgroup:
if it is large enough, a buffer is allocated regardless so that the element in fill->buf can later be reclaimed.
Valgrind is run on test/set_extent.c and there is no memory leak.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -137,6 +137,19 @@ Bug Fixes since HDF5-1.13.2 release
 
       (NAF - 2022/08/22, GitHub #2016)
 
+    - Memory leak 
+    
+      A memory leak was observed with variable-length fill value in 
+      H5O_fill_convert() function in H5Ofill.c. The leak is
+      manifested by running valgrind on test/set_extent.c.
+
+      Previously, fill->buf is used for datatype conversion 
+      if it is large enough and the variable-length information 
+      is therefore lost.  A buffer is now allocated regardless 
+      so that the element in fill->buf can later be reclaimed.
+
+      (VC - 2022/10/10, HDFFV-10840)
+ 
 
     Java Library
     ------------
--- a/src/H5Ofill.c
+++ b/src/H5Ofill.c
@@ -1007,6 +1007,8 @@ H5O_fill_convert(H5O_fill_t *fill, H5T_t *dset_type, hbool_t *fill_changed)
 
     /* Don't bother doing anything if there will be no actual conversion */
     if (!H5T_path_noop(tpath)) {
+        size_t fill_type_size;
+
         if ((src_id = H5I_register(H5I_DATATYPE, H5T_copy(fill->type, H5T_COPY_ALL), FALSE)) < 0 ||
             (dst_id = H5I_register(H5I_DATATYPE, H5T_copy(dset_type, H5T_COPY_ALL), FALSE)) < 0)
             HGOTO_ERROR(H5E_OHDR, H5E_CANTINIT, FAIL, "unable to copy/register data type")
@@ -1015,13 +1017,11 @@ H5O_fill_convert(H5O_fill_t *fill, H5T_t *dset_type, hbool_t *fill_changed)
          * Datatype conversions are always done in place, so we need a buffer
          * that is large enough for both source and destination.
          */
-        if (H5T_get_size(fill->type) >= H5T_get_size(dset_type))
-            buf = fill->buf;
-        else {
-            if (NULL == (buf = H5MM_malloc(H5T_get_size(dset_type))))
-                HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "memory allocation failed for type conversion")
-            H5MM_memcpy(buf, fill->buf, H5T_get_size(fill->type));
-        } /* end else */
+        fill_type_size = H5T_get_size(fill->type);
+
+        if (NULL == (buf = H5MM_malloc(MAX(fill_type_size, H5T_get_size(dset_type)))))
+            HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "memory allocation failed for type conversion")
+        H5MM_memcpy(buf, fill->buf, fill_type_size);
 
         /* Use CALLOC here to clear the buffer in case later the library thinks there's
          * data in the background. */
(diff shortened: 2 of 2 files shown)


vchoi-hdfgroup:
The commit message should be:
Instead of using fill->buf for datatype conversion if it is large enough, 
a buffer is allocated regardless so that the element in fill->buf can later be reclaimed.
Valgrind is run on test/set_extent.c and there is no memory leak.

derobins:
Nice fix! This needs a RELEASE.txt note too, since it's fixing a memory leak.

fortnern:
Probably not an issue since this isn't critical for performance, but would it work if you took the old code and moved the call to H5T_vlen_reclaim_elmt() outside the if (buf != fill->buf) block?

fortnern:
I guess I'm a little confused as to where the leak is occurring. Does H5T_convert() not reclaim the source type if it needs to alloc new vlen buffers? That seems like it would cause problems elsewhere.

Edit: I read the Jira report, I don't really see how this fixes that. Since buf/fill->buf is shallow copied, shouldn't the H5Tconvert() reclaim the old buffer from both of them, so the explicit reclaim call frees the vl data inside the active fill buffer? I mean I can see how this fixes the memory leak but if the fill value is used again it'd cause problems unless I'm missing something.

vchoi-hdfgroup:
Before the fix: fill->buf is reused.  
--See function description for H5T__conv_vlen() in H5Tconv.c
--See also H5T__vlen_disk_write() in H5Tvlen.c:
    The info in vl (which is fill->buf) after UINT32ENCODE(vl, seq_len) is lost and couldn't be reclaimed later.

fortnern:
> 

The function descriiption says it frees the old data. I'm still concerned that this fix frees the VL data in the fill value at dataset open time.

fortnern:
A few things:
-Should probably add a comment explaining why you're shallow copying the fill value buffer and that H5T_convert() leaks memory otherwise.
- This behavior of H5T_convert() (and presumably the public H5Tconvert()), where you need to keep a copy of the unconverted buffer around to free embedded storage, while necessary for how data I/O is handled internally, is surprising and should be documented, at least at the public API level for H5Tconvert(). Anyone using H5Tconvert() on VL data is probably expecting it to reuse or realloc the original buffers.
- The jira issue states that the library is unable to free the VL data at close time because the fill value message's datatype is NULL. Is this not the case? If this is still the case, how does that VL data get freed at dataset close time?

vchoi-hdfgroup:
May I assume the existing code (before fix) in H5O_fill_convert() (H5Ofill.c) is correct when the buffer is not large enough and a new buffer needs to be allocated and later on reclaimed?
        
        if (H5T_get_size(fill->type) >= H5T_get_size(dset_type))
            buf = fill->buf;
        else {
            if (NULL == (buf = H5MM_malloc(H5T_get_size(dset_type))))
                HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "memory allocation failed for type conversion")
            H5MM_memcpy(buf, fill->buf, H5T_get_size(fill->type));
        } /* end else */
       if (buf != fill->buf) {
            H5T_vlen_reclaim_elmt(fill->buf, fill->type);
            H5MM_xfree(fill->buf);
            fill->buf = buf;
        } /* end if */

If the above is correctly done, the proposed fix is just allocating a buffer regardless.

Also for the question "The jira issue states that the library is unable to free the VL data at close time because the fill value message's datatype is NULL. Is this not the case? If this is still the case, how does that VL data get freed at dataset close time?":
Please see the function description for H5O_fill_convert() about the type field being null.  The VL data cannot be freed at dataset close time because the info is lost as described previously and therefore the fix in this PR is proposed.

fortnern:
> May I assume the existing code (before fix) in H5O_fill_convert() (H5Ofill.c) is correct when the buffer is not large enough and a new buffer needs to be allocated and later on reclaimed?
> 
> ```
>     if (H5T_get_size(fill->type) >= H5T_get_size(dset_type))
>         buf = fill->buf;
>     else {
>         if (NULL == (buf = H5MM_malloc(H5T_get_size(dset_type))))
>             HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "memory allocation failed for type conversion")
>         H5MM_memcpy(buf, fill->buf, H5T_get_size(fill->type));
>     } /* end else */
>    if (buf != fill->buf) {
>         H5T_vlen_reclaim_elmt(fill->buf, fill->type);
>         H5MM_xfree(fill->buf);
>         fill->buf = buf;
>     } /* end if */
> ```
> 
> If the above is correctly done, the proposed fix is just allocating a buffer regardless.
> 
> Also for the question "The jira issue states that the library is unable to free the VL data at close time because the fill value message's datatype is NULL. Is this not the case? If this is still the case, how does that VL data get freed at dataset close time?": Please see the function description for H5O_fill_convert() about the type field being null. The VL data cannot be freed at dataset close time because the info is lost as described previously and therefore the fix in this PR is proposed.

Are you saying this PR somehow allows the VL data to be freed at dataset close time when it wasn't before? That doesn't make sense to me. I agree that this PR is effectively just always taking the "alloc" branch in H5O_fill_convert(), but if the VL data isn't freed at close time it's either leaked or it was freed too soon. I think we might need to meet to try to understand this.

fortnern:
Put another way: after H5T_convert(), there are 2 VL buffers. H5T_vlen_reclaim_elmt() frees one of them. Where is the other one freed?

vchoi-hdfgroup:
The other buffer is freed at dataset close time: 
--In H5D_close():  H5O_msg_reset(H5O_FILL_ID, &dataset->shared->dcpl_cache.fill) is called, which calls H5O_fill_reset_dyn(fill)
--In H5O_fill_reset_dyn(), the fill value buffer is freed:  fill->buf = H5MM_xfree(fill->buf);

fortnern:
This frees the fill buffer, not the VL data inside the fill buffer (unless I'm missing something). The Jira issue describes, and I can see in the code, that the VL data is only freed if fill->type is not NULL.

fortnern (approved):
Turns out the Jira issue was misleading - the converted fill value buffer doesn't need to be reclaimed at dataset close time since it's a disk type.

derobins (approved):

PR #2156, opened 2022-10-12
JNI: Improve varlen datatype handling in H5A/H5D read/write functions

byrnHDF:
Existing code works if VL datatype is top level datatype.
This change moves the read/write to a separate function so it can be called recursively for VL types in container datatpyes (VL, Compound, Array). Java passes in an ArrayList of Arrays for the data buffer. This needs to be translated to C structures.

Tests need to be written for the other types (see the tvldtypes.h5 files generated by h5dumpgentest.c).

### DIFF
--- a/java/src/jni/h5aImp.c
+++ b/java/src/jni/h5aImp.c
@@ -181,17 +181,9 @@ Java_hdf_hdf5lib_H5_H5Aread(JNIEnv *env, jclass clss, jlong attr_id, jlong mem_t
             H5_BAD_ARGUMENT_ERROR(ENVONLY, "H5Aread: readBuf length < 0");
         }
 
-        dims[0] = (hsize_t)vl_array_len;
-        if ((sid = H5Screate_simple(1, dims, NULL)) < 0)
+        if (!(typeSize = H5Tget_size(mem_type_id)))
             H5_LIBRARY_ERROR(ENVONLY);
-    }
 
-    if (!(typeSize = H5Tget_size(mem_type_id)))
-        H5_LIBRARY_ERROR(ENVONLY);
-
-    if ((type_class = H5Tget_class((hid_t)mem_type_id)) < 0)
-        H5_LIBRARY_ERROR(ENVONLY);
-    if (vl_data_class) {
         if (NULL == (readBuf = HDcalloc((size_t)vl_array_len, typeSize)))
             H5_OUT_OF_MEMORY_ERROR(ENVONLY, "H5Aread: failed to allocate raw VL read buffer");
     }
@@ -208,13 +200,22 @@ Java_hdf_hdf5lib_H5_H5Aread(JNIEnv *env, jclass clss, jlong attr_id, jlong mem_t
     if ((status = H5Aread((hid_t)attr_id, (hid_t)mem_type_id, (void *)readBuf)) < 0)
         H5_LIBRARY_ERROR(ENVONLY);
 
-    if (vl_data_class)
+    if (vl_data_class) {
+        if ((type_class = H5Tget_class((hid_t)mem_type_id)) < 0)
+            H5_LIBRARY_ERROR(ENVONLY);
+
         translate_rbuf(env, buf, mem_type_id, type_class, vl_array_len, readBuf);
+    }
 
 done:
     if (readBuf) {
         if ((status >= 0) && vl_data_class) {
+            dims[0] = (hsize_t)vl_array_len;
+            if ((sid = H5Screate_simple(1, dims, NULL)) < 0)
+                H5_LIBRARY_ERROR(ENVONLY);
+                
             H5Treclaim(attr_id, sid, H5P_DEFAULT, readBuf);
+            
             if (sid >= 0)
                 H5Sclose(sid);
         }
--- a/java/src/jni/h5aImp.c
+++ b/java/src/jni/h5aImp.c
@@ -156,44 +156,60 @@ JNIEXPORT jint JNICALL
 Java_hdf_hdf5lib_H5_H5Aread(JNIEnv *env, jclass clss, jlong attr_id, jlong mem_type_id, jbyteArray buf,
                             jboolean isCriticalPinning)
 {
-    jboolean readBufIsCopy;
-    jbyte   *readBuf = NULL;
-    hsize_t  dims[H5S_MAX_RANK];
-    hid_t    sid = H5I_INVALID_HID;
-    jsize    n;
-    htri_t   vl_data_class;
-    herr_t   status = FAIL;
+    jboolean    readBufIsCopy;
+    jbyte      *readBuf = NULL;
+    hsize_t     dims[H5S_MAX_RANK];
+    hid_t       sid = H5I_INVALID_HID;
+    size_t      typeSize;
+    H5T_class_t type_class;
+    jsize       n;
+    htri_t      vl_data_class;
+    herr_t      status = FAIL;
 
     UNUSED(clss);
 
     if (NULL == buf)
         H5_NULL_ARGUMENT_ERROR(ENVONLY, "H5Aread: read buffer is NULL");
 
+    /* Get size of data array */
+    if ((n = ENVPTR->GetArrayLength(ENVONLY, buf)) < 0) {
+        CHECK_JNI_EXCEPTION(ENVONLY, JNI_TRUE);
+        H5_BAD_ARGUMENT_ERROR(ENVONLY, "H5Aread: readBuf length < 0");
+    }
+
     if ((vl_data_class = h5str_detect_vlen(mem_type_id)) < 0)
         H5_LIBRARY_ERROR(ENVONLY);
 
     if (vl_data_class) {
-        /* Get size of data array */
-        if ((n = ENVPTR->GetArrayLength(ENVONLY, buf)) < 0) {
-            CHECK_JNI_EXCEPTION(ENVONLY, JNI_TRUE);
-            H5_BAD_ARGUMENT_ERROR(ENVONLY, "H5Aread: readBuf length < 0");
-        }
-
         dims[0] = (hsize_t)n;
         if ((sid = H5Screate_simple(1, dims, NULL)) < 0)
             H5_LIBRARY_ERROR(ENVONLY);
     }
 
-    if (isCriticalPinning) {
-        PIN_BYTE_ARRAY_CRITICAL(ENVONLY, buf, readBuf, &readBufIsCopy,
-                                "H5Aread: read buffer not critically pinned");
+    if (!(typeSize = H5Tget_size(mem_type_id)))
+        H5_LIBRARY_ERROR(ENVONLY);
+
+    if ((type_class = H5Tget_class((hid_t)mem_type_id)) < 0)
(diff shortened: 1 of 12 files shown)


byrnHDF:
Removing WIP label for the 1.13.3 release as this works for VL datatypes.
Compound and array types will be addressed next in a separate PR for each.

java/src/jni/h5aImp.c
```diff
@@ -156,45 +156,61 @@ JNIEXPORT jint JNICALL
...
+    htri_t      vl_data_class;
+    herr_t      status = FAIL;
 
     UNUSED(clss);
 
     if (NULL == buf)
         H5_NULL_ARGUMENT_ERROR(ENVONLY, "H5Aread: read buffer is NULL");
 
+    /* Get size of data array */
+    if ((n = ENVPTR->GetArrayLength(ENVONLY, buf)) < 0) {
```
jhendersonHDF: Shouldn't this stay inside the `vl_data_class` block? Looks like it's only used in those blocks.

java/src/jni/h5aImp.c
```diff
@@ -239,7 +262,7 @@ Java_hdf_hdf5lib_H5_H5Awrite(JNIEnv *env, jclass clss, jlong attr_id, jlong mem_
     /* Get size of data array */
     if ((n = ENVPTR->GetArrayLength(ENVONLY, buf)) < 0) {
```
jhendersonHDF: Same here. I think this can just be moved inside the first `if (vl_data_class)` block.

java/src/jni/h5aImp.c
```diff
@@ -239,7 +262,7 @@ Java_hdf_hdf5lib_H5_H5Awrite(JNIEnv *env, jclass clss, jlong attr_id, jlong mem_
...
         CHECK_JNI_EXCEPTION(ENVONLY, JNI_TRUE);
-        H5_BAD_ARGUMENT_ERROR(ENVONLY, "H5Aread: readBuf length < 0");
+        H5_BAD_ARGUMENT_ERROR(ENVONLY, "H5Awrite: write buffer length < 0");
     }
 
     dims[0] = (hsize_t)n;
```
jhendersonHDF: Along with the creation of the dataspace here and below.

java/src/jni/h5aImp.c
```diff
@@ -1064,244 +1106,9 @@ Java_hdf_hdf5lib_H5_H5Awrite_1string(JNIEnv *env, jclass clss, jlong attr_id, jl
...
-        if ((status >= 0) && vl_data_class) {
-            H5Treclaim(attr_id, sid, H5P_DEFAULT, readBuf);
-            if (sid >= 0)
-                H5Sclose(sid);
-        }
+    herr_t status = FAIL;
 
-        UNPIN_BYTE_ARRAY(ENVONLY, buf, readBuf, (status < 0) ? JNI_ABORT : 0);
-    }
+    status = Java_hdf_hdf5lib_H5_H5Aread(env, clss, attr_id, mem_type_id, (jbyteArray)buf, JNI_TRUE);
```
jhendersonHDF, line 1115: Do these still work as intended (what is their intention in the first place?)? I'm suspicious about casting the jobjectArray buf into a jbyteArray and passing it to the regular read/write funcs.

java/src/jni/h5dImp.c
```diff
@@ -182,47 +182,69 @@ Java_hdf_hdf5lib_H5_H5Dread(JNIEnv *env, jclass clss, jlong dataset_id, jlong me
...
+        }
     }
 
     if ((status = H5Dread((hid_t)dataset_id, (hid_t)mem_type_id, (hid_t)mem_space_id, (hid_t)file_space_id,
-                          (hid_t)xfer_plist_id, readBuf)) < 0)
+                          (hid_t)xfer_plist_id, (void *)readBuf)) < 0)
         H5_LIBRARY_ERROR(ENVONLY);
 
+    if (vl_data_class)
+        translate_rbuf(env, buf, mem_type_id, type_class, n, (jobjectArray)readBuf);
```
jhendersonHDF: Just a thought - might be worthing naming these `translate_vl_rbuf`/`translate_vl_wbuf` just so there's no confusion at first glance about what type of buffers they're translating.

java/src/jni/h5util.c
```diff
@@ -4043,6 +4043,551 @@ Java_hdf_hdf5lib_H5_H5export_1attribute(JNIEnv *env, jclass clss, jstring file_e
...
+     jmethodID boolValueMid =
+     ENVPTR->GetStaticMethodID(ENVONLY, cBool, "valueOf", "(Z)Ljava/lang/Boolean;");
+     */
+    jmethodID byteValueMid  = ENVPTR->GetStaticMethodID(ENVONLY, cByte, "valueOf", "(B)Ljava/lang/Byte;");
+    jmethodID shortValueMid = ENVPTR->GetStaticMethodID(ENVONLY, cShort, "valueOf", "(S)Ljava/lang/Short;");
+    jmethodID intValueMid   = ENVPTR->GetStaticMethodID(ENVONLY, cInt, "valueOf", "(I)Ljava/lang/Integer;");
+    jmethodID longValueMid  = ENVPTR->GetStaticMethodID(ENVONLY, cLong, "valueOf", "(J)Ljava/lang/Long;");
+    jmethodID floatValueMid = ENVPTR->GetStaticMethodID(ENVONLY, cFloat, "valueOf", "(F)Ljava/lang/Float;");
+    jmethodID doubleValueMid =
+        ENVPTR->GetStaticMethodID(ENVONLY, cDouble, "valueOf", "(D)Ljava/lang/Double;");
```
jhendersonHDF, line 4080: In the future, it's probably best to figure out what the type actually is and not grab the classes/methods/etc. for every primitive type. It's likely that some of these calls can be expensive.

java/src/jni/h5util.c
```diff
@@ -4043,6 +4043,551 @@ Java_hdf_hdf5lib_H5_H5export_1attribute(JNIEnv *env, jclass clss, jstring file_e
...
+    }
+
+done:
+
+    return (jint)status;
+}
+
+herr_t
+translate_wbuf(JNIEnv *env, jobjectArray in_buf, jlong mem_type_id, H5T_class_t type_class, jsize count,
+               jobjectArray raw_buf)
```
jhendersonHDF: For writes, this result is always going to be a byte buffer that gets passed to H5Awrite/H5Dwrite, correct? In that case, I'd not mark this as returning a jobjectArray because it's misleading. In fact, we're doing direct `HDmemcpy` on a jobjectArray which might only be working by coincidence. Probably best to mark raw_buf as a `void *` or similar.

java/src/jni/h5util.c
```diff
@@ -4043,6 +4043,551 @@ Java_hdf_hdf5lib_H5_H5export_1attribute(JNIEnv *env, jclass clss, jstring file_e
         H5Aclose(attr_id);
 } /* end Java_hdf_hdf5lib_H5_H5export_1attribute */
 
+herr_t
+translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t type_class, jsize count,
+               jobjectArray raw_buf)
```
jhendersonHDF: Same here that raw_buf should probably be marked as a `void *` or similar since it's just a byte buffer and not a jobjectArray.

java/src/jni/h5aImp.c
byrnHDF: I had to split the vl_data_class in three parts and each needed the number.

byrnHDF: I guess I could have duplicated it.

java/src/jni/h5aImp.c
byrnHDF: Oh right the n is a function var not a block var.

java/src/jni/h5aImp.c
byrnHDF: The issue is that the H5.java read, checks the type of the buffer and unfortunately will just loop through the wrong function.  I couldn't figure out a way to get the correct call and the correct cast another way.

java/src/jni/h5dImp.c
byrnHDF: No, because they will also do array and compounds

java/src/jni/h5util.c
byrnHDF: Sounds right

java/src/jni/h5util.c
byrnHDF: Yeah, that is true, but that is why I cached them at the top instead of in the switch later. I'm thinking compound may need different types.

java/src/jni/h5util.c
byrnHDF: yeah, void * would be better.

java/src/jni/h5dImp.c
byrnHDF: Right now they are only triggered by a vl type, but the function is written such that it isn't a requirement.

java/src/jni/h5util.c
byrnHDF: How about jbyte * as that is the declared type of writeBuf?

java/src/jni/h5aImp.c
jhendersonHDF: Each part can still use the number. The variable itself is in the scope of all the blocks, but you can still only set it in the first block.

java/src/jni/h5aImp.c
jhendersonHDF: I think that's a different issue than what I was getting at. My question is are the readVL/writeVL routines still used anywhere and if so, why/what for? And also, I still have concerns about the casting of the buffer here since it's not clear why it's being done.

java/src/jni/h5dImp.c
jhendersonHDF: > Right now they are only triggered by a vl type, but the function is written such that it isn't a requirement.

But isn't that effectively no different than a VL type being a requirement? Also, what if they're triggered by a VL type inside a compound? Doesn't that mean the data for a compound will be in a different format than Java expects since it'll get translated into a list?

java/src/jni/h5util.c
jhendersonHDF: That would probably be best if that's what matches the type for writeBuf/readBuf.

java/src/jni/h5dImp.c
byrnHDF: No, the function just translate raw buffer (bytes) to / from Lists.
So the functions could do just compound or array or any combo of vl, compound or array. The Java buffer is an ArrayList the raw buffer bytes to the hdf5 library

java/src/jni/h5aImp.c
byrnHDF: Right, take a look at H5.java H5Dread - VL will have a buffer of ArrayLists --- and the default assumes byte array which then goes through HDFArray function. 
That HDFArray call probably needs to recognize ArraryLists and pass them through, but I couldn't figure that out, so I just kept the VL check and redirected for now.
Obviously for compounds and array combos I will need to solve that.

jhendersonHDF (approved):

lrknox (approved):

PR #2161, opened 2022-10-13
Fix H5Dfill_f test

brtnfld:
Changed legacy tests to an integer type.


brtnfld:
hold on, my repo is not in sync

PR #2162, opened 2022-10-13
Fix H5Dfill_f test

brtnfld:
Changed to integer for legacy H5Dfill_f testing.


lrknox (approved):
I've tested 7 of the configurations that previously failed to compile on echidna, jelly, kite and moohan.   All of them build without compile errors.  Only the configuration on jelly with MPICH3.3 and check-passthrough-vol failed Fortran API: Test log for parallel_test with VOL connector: native in the contiguous and chunked layout with collective MPI I/O tests, which is apparently a separate issue.

byrnHDF (approved):

PR #2164, opened 2022-10-15
Remove memory sanity checks

derobins:
Also simplifies the H5MM package, replacing most H5MM replacement functions with defines to C standard library functions

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -55,13 +55,29 @@ New Features
 
       (ADB - 2022/08/29, HDFFV-11329)
 
-    -
+    - Removed the --enable-memory-alloc-sanity-check Autotools configure option
+      and the HDF5_MEMORY_ALLOC_SANITY_CHECK CMake option
+
+      The memory allocation sanity checks have been removed from the library.
+
+      (DER - 2022/10/15)
 
 
     Library:
     --------
-    -
+    - Removed the memory allocation sanity checks from the library
 
+      This feature added heap canaries to memory allocated inside the library
+      that were checked on library close. Although potentially helpful, other
+      methods for investigating memory leaks exist (valgrind, -fsanitize, etc.)
+      and the heap canaries caused problems with filters and API calls that
+      return library-allocated memory.
+
+      As a part of this change, the H5get_alloc_stats() API call has been
+      removed from the library. The Autotools and CMake features that
+      enabled this feature have also been removed (see above).
+
+      (DER - 2022/10/15)
 
     Parallel Library:
     -----------------
@@ -75,7 +91,7 @@ New Features
       Added Fortran H5Dfill_f, which is fully equivalent to the C API. It accepts pointers,
       fill value datatype and datatype of dataspace elements.
 
-      (MSB - 2022/10/10, HDFFV-10734.)
+      (MSB - 2022/10/10, HDFFV-10734)
 
     C++ Library:
     ------------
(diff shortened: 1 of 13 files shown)


byrnHDF (approved):

release_docs/RELEASE.txt
```diff
@@ -55,13 +55,29 @@ New Features
 
       (ADB - 2022/08/29, HDFFV-11329)
 
-    -
+    - Removed the --enable-memory-alloc-sanity-check Autotools configure option
```
byrnHDF, line 58: newer notes at top of section, right?

bmribler: That's what I thought...

src/H5.c
```diff
@@ -1262,7 +1219,10 @@ H5resize_memory(void *mem, size_t size)
     FUNC_ENTER_API_NOINIT
     H5TRACE2("*x", "*xz", mem, size);
 
-    ret_value = H5MM_realloc(mem, size);
+    if (size != 0)
+        ret_value = H5MM_realloc(mem, size);
+    else if (mem)
+        ret_value = H5MM_xfree(mem);
```
bmribler, line 1225: I read here https://www.tutorialspoint.com/c_standard_library/c_function_realloc.htm, it said
size − This is the new size for the memory block, in bytes. If it is 0 and ptr points to an existing block of memory, the memory block pointed by ptr is deallocated and a NULL pointer is returned.
Do we need to use H5MM_xfree instead?

src/H5MM.c
```diff
@@ -449,122 +72,24 @@ H5MM_xstrdup(const char *s)
...
-    FUNC_LEAVE_NOAPI(ret_value)
-} /* end H5MM_strndup() */
-
 /*-------------------------------------------------------------------------
  * Function:    H5MM_xfree
  *
- * Purpose:     Just like free(3) except null pointers are allowed as
- *              arguments, and the return value (always NULL) can be
- *              assigned to the pointer whose memory was just freed:
+ * Purpose:     Just like free(3) except the return value (always NULL) can
```
fortnern, line 86: Is H5MM_xfree still needed? Do any supported systems fail if we remove it (could test by #define H5MM_xfree H5MM_free)?

fortnern:
It won't let me approve it for some reason but "approved"

fortnern (approved):

src/H5MM.c
jhendersonHDF: If it gets mapped to just plain `free` then I think we still have several places in the library that would do `mem = free` which won't compile, so we'd need to update those. But I don't think it's really needed.

lrknox:
Test that failed was one of the dt_arith tests in the MacOS Clang CMake REL test configuration.
23: Testing hard normalized long double -> signed char conversions        
...
  22/2452 Test   #23: H5TEST-dt_arith ............................................................***Failed    0.24 sec

There were also 75+ "Failed ..." messages in the same test configuration, mostly in h5repack (and h5dump) tests, but all of those tests passed.

jhendersonHDF:
> Test that failed was one of the dt_arith tests in the MacOS Clang CMake REL test configuration. 23: Testing hard normalized long double -> signed char conversions ... 22/2452 Test #23: H5TEST-dt_arith ............................................................***Failed 0.24 sec
> 
> There were also 75+ "Failed ..." messages in the same test configuration, mostly in h5repack (and h5dump) tests, but all of those tests passed.

This is the same test that fails in https://github.com/HDFGroup/hdf5/pull/1929. @derobins and I need to figure out what's going on there because it looks like we have a memory issue that's been hidden on MacOS for a long time.

src/H5MM.c
fortnern: We should be able to convert H5MM_xfree to a macro:
#define H5MM_xfree(A) (H5MM_free(A), NULL)

derobins:
Need to investigate the long double conversion failures on MacOS

PR #2175, opened 2022-10-19
Implement option for using CMake GNUInstallDirs

byrnHDF:
Default for option HDF5 "HDF5_USE_GNU_DIRS" is OFF to use existing install structure.


lrknox (approved):

config/cmake/HDF5PluginCache.cmake
```diff
@@ -27,3 +27,5 @@ set (H5PL_TGZ_NAME "${PLUGIN_TGZ_NAME}" CACHE STRING "Use plugins from compresse
 
 set (PL_PACKAGE_NAME "${PLUGIN_PACKAGE_NAME}" CACHE STRING "Name of plugins package" FORCE)
 set (H5PL_CPACK_ENABLE OFF CACHE BOOL "Enable CPack include and components" FORCE)
+
+set (H5PL_USE_GNU_DIRS ${HDF5_USE_GNU_DIRS} CACHE BOOL "TRUE to use GNU Coding Standard install directory variables" FORCE)
```
thesamesam, line 31: Can I suggest:
```suggestion
set (H5PL_USE_GNU_DIRS ${HDF5_USE_GNU_DIRS} CACHE BOOL "TRUE to use GNU Coding Standard install directory variables (recommended)" FORCE)
```

or
```suggestion
set (H5PL_USE_GNU_DIRS ${HDF5_USE_GNU_DIRS} CACHE BOOL "TRUE to use GNU Coding Standard install directory variables (recommended for Linux distros)" FORCE)
```

thesamesam (approved):
Thanks for working on this. This will make a life a lot easier for Linux distros.

config/cmake/HDF5PluginCache.cmake
byrnHDF: The trouble with those is that it is not just for linux distros. It can be used on any platform. The key is the GNU standard and I guess your second suggestion is most accruate.  I think that should be in the missing release note and explained in the INSTALL_CMAKE.txt file. The cache hint should probably be shortened.

jhendersonHDF (approved):

derobins (approved):

PR #2181, opened 2022-10-24
Fix problem with variable length attributes being accessed through multiple file handles

fortnern:
Previously, if a variable length attribute was held open while its file was opened through another handle, the same attribute was opened through the second file handle, and the second file and attribute handles were closed, attempting to write to the attribute through the first handle would cause an error.


jhendersonHDF (approved):

byrnHDF (approved):

bmribler:
The user wants to know which releases this will go in.  It made it to 1.13.3, right?  And then all the other branches of their next release, I assume.

Issue #2209, opened 2022-11-02
[CVE-2019-8396, HDFFV-10712] Heap Buffer Overflow in  H5O__pline_decode() 


e4t:
**Description**
[CVE-2019-8396](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-8396) provides a [reproducer](https://github.com/magicSwordsMan/PAAFS/tree/master/vul4) that causes a heap buffer overflow in `H5O__pline_decode()`. The issue is still seen in hdf5 version 1.13.3.
The cause of this is an incomplete bounds checking in said function as only the the address of the filter descriptions is checked against the bounds. If the overflow happens before, because there isn't even enough space for the filter pipeline header, an overflow will be triggered.
Once this overflow condition is handled properly, the following overflow in `H5O__layout_decode` no longer occurs.
On version 1.13 the SEGV seen in the link above no longer occurs. The heap overflow itself appears to be harmless from a security perspective. This may not be true for later memory accesses (possibly writes) based on inconsistent/invalid data read.
The overflow can be seen both with valgrind and the gcc address sanitizer (`make CFLAGS="-g -fsanitize=address"  CXXFLAGS="-g -fsanitize=address"  FCFLAGS="-g -fsanitize=address"`)

**Expected behavior**
Detect invalid hdf5 data properly and trigger the appropriate error handler. 
Handle buffer overflow condition properly to avoid possible vulnerabilities thru manipulated hdf5 data files brought in circulation by bad actors.

**Platform**
 - HDF5 version 1.13.3
 - openSUSE Leap 15.4 / SLE 15.4 / openSUSE Tumbleweed
 - gcc 7 / gcc 11
 - autotools
 - configure --enable-fortran --enable-unsupported --enable-hl --enable-shared --enable-threadsafe --enable-build-mode=production --enable-cxx --with-pthread

mattjala:
Fixed in #2210.

PR #2210, opened 2022-11-02
H5O__pline_decode() Make more resilient to out-of-bounds read

e4t:
Malformed hdf5 files may have trunkated content which does not match the expected size. When this function attempts to decode these it may read past the end of the allocated space leading to heap overflows as bounds checking is incomplete.
Make sure each element is within bounds before reading.

This fixes CVE-2019-8396 / HDFFV-10712 / github bug #2209.

### DIFF
--- a/src/H5Opline.c
+++ b/src/H5Opline.c
@@ -110,6 +110,7 @@ H5FL_DEFINE(H5O_pline_t);
  *
  *-------------------------------------------------------------------------
  */
+
 static void *
 H5O__pline_decode(H5F_t H5_ATTR_UNUSED *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSED mesg_flags,
                   unsigned H5_ATTR_UNUSED *ioflags, size_t p_size, const uint8_t *p)
@@ -131,6 +132,9 @@ H5O__pline_decode(H5F_t H5_ATTR_UNUSED *f, H5O_t H5_ATTR_UNUSED *open_oh, unsign
         HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL, "memory allocation failed")
 
     /* Version */
+    if (p + 4 - 1 > p_end) /* 4 byte is minimum for all versions */
+        HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, NULL, "ran off the end of the buffer: current p = %p, p_end = %p",
+                    p + 4, p_end)
     pline->version = *p++;
     if (pline->version < H5O_PLINE_VERSION_1 || pline->version > H5O_PLINE_VERSION_LATEST)
         HGOTO_ERROR(H5E_PLINE, H5E_CANTLOAD, NULL, "bad version number for filter pipeline message")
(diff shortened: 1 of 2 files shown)


hyoklee (approved):

src/H5Opline.c
```diff
@@ -110,6 +110,14 @@ H5FL_DEFINE(H5O_pline_t);
  *
  *-------------------------------------------------------------------------
  */
+static char err[] = "ran off the end of the buffer: current p = %p, p_end = %p";
+
+#define VERIFY_LIMIT(p,s,l)                                        \
+  if (p + s - 1 > l) {                                             \
+    HCOMMON_ERROR(H5E_RESOURCE, H5E_NOSPACE, err, p + s, l);       \
```
jhendersonHDF: Typically, HCOMMON_ERROR isn't used directly in the code other than being part of the other error macros. Also, most of the error macros should be able to handle printf-like arguments, so these two lines should be able to just be an `HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL, "ran off the end of the buffer: current p = %p, p_end = %p", p + s, l)`

e4t: The use of HCOMMON_ERROR() was a leftover from a previous version of the code - I don't remember why I believed I needed it. Will change this.
The use of the variable` err `instead of the string was to avoid needless repetition of the same string in the binary -  just found out, modern compilers will detect & merge these. 
Will fix and push again.

src/H5Opline.c
```diff
@@ -110,6 +110,14 @@ H5FL_DEFINE(H5O_pline_t);
...
  *-------------------------------------------------------------------------
  */
+
+#define VERIFY_LIMIT(p,s,l)                                                  \
+  if (p + s - 1 > l) {                                                       \
+    HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL,                             \
+                "ran off the end of the buffer: current p = %p, p_end = %p", \
+                p + s, l);                                                   \
+  };
+
```
derobins: I am not a fan of adding a local convenience macro here.

src/H5Opline.c
```diff
@@ -110,6 +110,14 @@ H5FL_DEFINE(H5O_pline_t);
  *
  *-------------------------------------------------------------------------
  */
+
+#define VERIFY_LIMIT(p,s,l)                                                  \
+  if (p + s - 1 > l) {                                                       \
+    HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL,                             \
```
derobins: The library is inconsistant, but we generally make the major error category match the package, so this should be H5E_OHDR instead of H5E_RESOURCE.

src/H5Opline.c
```diff
@@ -179,9 +190,12 @@ H5O__pline_decode(H5F_t H5_ATTR_UNUSED *f, H5O_t H5_ATTR_UNUSED *open_oh, unsign
         /* Filter name, if there is one */
         if (name_length) {
             size_t actual_name_length; /* Actual length of name */
-
+            size_t len = (size_t)(p_end - p + 1);
             /* Determine actual name length (without padding, but with null terminator) */
-            actual_name_length = HDstrlen((const char *)p) + 1;
+            actual_name_length = HDstrnlen((const char *)p, len);
+            if (actual_name_length == len)
+                HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL, "filter name not null terminated")
```
derobins: Same here: H5E_OHDR

src/H5Opline.c
e4t: I've removed the macro, now, and included explicit tests.

src/H5Opline.c
e4t: I've changed this and pushed the new version. 
Now it's inconsistent with the other error messages in the same function. These should probably be fixed separately.

jhendersonHDF:
@e4t If you could make the last change mentioned above for changing `H5E_RESOURCE` -> `H5E_OHDR`, I think the rest of this looks good to me.

jhendersonHDF (approved):

derobins (approved):

PR #2218, opened 2022-11-03
Removes the memory allocation sanity checks feature

derobins:


jhendersonHDF (approved):

src/H5MM.c
```diff
@@ -260,58 +77,8 @@ H5MM_malloc(size_t size)
...
-            if (H5MM_curr_alloc_blocks_count_s > H5MM_peak_alloc_blocks_count_s)
-                H5MM_peak_alloc_blocks_count_s = H5MM_curr_alloc_blocks_count_s;
-
-            /* Set buffer to return */
-            ret_value = block->b + H5MM_HEAD_GUARD_SIZE;
-        } /* end if */
-        else
-            ret_value = NULL;
-#else  /* H5_MEMORY_ALLOC_SANITY_CHECK */
+    if (size)
```
fortnern: Is this check necessary on any systems we support?

src/H5MM.c
```diff
@@ -347,14 +109,8 @@ H5MM_calloc(size_t size)
     /* Use FUNC_ENTER_NOAPI_NOINIT_NOERR here to avoid performance issues */
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
-    if (size) {
-#if defined H5_MEMORY_ALLOC_SANITY_CHECK
-        if (NULL != (ret_value = H5MM_malloc(size)))
-            HDmemset(ret_value, 0, size);
-#else  /* H5_MEMORY_ALLOC_SANITY_CHECK */
+    if (size)
```
fortnern: Same

src/H5MM.c
```diff
@@ -395,35 +147,12 @@ H5MM_realloc(void *mem, size_t size)
         /* Not defined in the standard, return NULL */
```
fortnern, line 141: From reading the standard it looks like NULL != mem && 0 == size is technically undefined. And if we consider malloc(0) to be defined (in the standard it looks like there are two possible behaviors), then NULL == mem && 0 == size is defined

src/H5MM.c
derobins: It's not necessary according to the standard. I didn't bother with looking carefully at existing behavior since I was just scissoring out the sanity checks, especially since the next step is to simply map H5MM calls to C standard library calls, but I can change this pretty easily.

src/H5MM.c
```diff
@@ -489,14 +203,8 @@ H5MM_strdup(const char *s)
 
     if (!s)
```
fortnern, line 198: Should this be an assertion?

src/H5MM.c
```diff
@@ -521,29 +229,15 @@ H5MM_strdup(const char *s)
...
 H5MM_strndup(const char *s, size_t n)
 {
-#if defined H5_MEMORY_ALLOC_SANITY_CHECK
-    size_t len;
-#endif
     char *ret_value = NULL;
 
     FUNC_ENTER_NOAPI(NULL)
 
     if (!s)
```
fortnern, line 230: Same

fortnern (approved):
The questions I raised shouldn't hold this back, just looking for places to trim unnecessary code

PR #2224, opened 2022-11-04
1.8 Merge #2175 Implement option for using CMake GNUInstallDirs

byrnHDF:


lrknox:
Title should be "1.8 Merge HDFGroup#2175 Implement option for using CMake GNUInstallDirs"

lrknox (approved):

Issue #1289, opened 2021-12-13
test/accum.c  CTest fails on macOS Monterey


hyoklee:
```
19: Testing SWMR write of large metadata: with latest format              *FAILED*
19:    at /Users/hyoklee/src/hdf5/test/accum.c:2291 in test_swmr_write_big()...
19: child process exited abnormally
19: Testing SWMR write of large metadata: with non-latest-format          *FAILED*
19:    at /Users/hyoklee/src/hdf5/test/accum.c:2291 in test_swmr_write_big()...
19: child process exited abnormally
19: *** TESTS FAILED ***
```

seanm:
What are the test names exactly?

I've just tried on older macOS 10.14:

```
$ ctest -R swmr
Test project /Users/sean/external/hdf5-bin
    Start  6: H5TEST-clear-objects
1/4 Test  #6: H5TEST-clear-objects .............   Passed    0.02 sec
    Start 65: H5TEST-swmr
2/4 Test #65: H5TEST-swmr ......................   Passed    0.33 sec
    Start 96: H5SHELL-testswmr
3/4 Test #96: H5SHELL-testswmr .................   Passed  154.97 sec
    Start 97: H5SHELL-testvdsswmr
4/4 Test #97: H5SHELL-testvdsswmr ..............   Passed   24.44 sec

100% tests passed, 0 tests failed out of 4
```

hyoklee:
19 - H5TEST-accum (Failed)

seanm:
I don't reproduce on my Intel Mac with macOS 10.14

I do reproduce with my ARM Mac with macOS 12.1.

I wonder if it's ARM vs Intel or 10.14 vs 12.1....

hyoklee:
No. My Mac is x86_64:
```Darwin nene.ad.hdfgroup.org 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64```

**Comment by seanm** [comment]
OK, let me try on my macOS 11.6.1 bot then...

**Comment by seanm** [comment]
Passes on 11.6.2 (on Intel):

```
builder@builder20 hdf5-bin % sw_vers
ProductName:	macOS
ProductVersion:	11.6.2
BuildVersion:	20G314

builder@builder20 hdf5-bin % uname -a
[...]

PR #2232, opened 2022-11-07
Refactor JNI translate functions to a recursive switch on datatype

byrnHDF:
Refactored into a switch on datatype class.
Added array implementation and tests for datasets and attributes. (Will add more tests for other combos later)
Still need to implement compounds.
Requires use of H5.H5DwriteVL and H5.H5DreadVL because of existing implementation of non-VL functions in H5.java file.
This implementation of read/write should be able to supersede the others once tested.

### DIFF
--- a/java/src/jni/h5util.c
+++ b/java/src/jni/h5util.c
@@ -3426,10 +3426,8 @@ Java_hdf_hdf5lib_H5_H5AreadComplex(JNIEnv *env, jclass clss, jlong attr_id, jlon
     size = (((H5Tget_size(mem_type_id)) > (H5Tget_size(p_type))) ? (H5Tget_size(mem_type_id))
                                                                  : (H5Tget_size(p_type)));
 
-    if ((n = ENVPTR->GetArrayLength(ENVONLY, buf)) <= 0) {
-        CHECK_JNI_EXCEPTION(ENVONLY, JNI_TRUE);
+    if ((n = ENVPTR->GetArrayLength(ENVONLY, buf)) <= 0)
         H5_BAD_ARGUMENT_ERROR(ENVONLY, "H5AreadComplex: read buffer length <= 0");
-    }
 
     if (NULL == (readBuf = (char *)HDmalloc((size_t)n * size)))
         H5_OUT_OF_MEMORY_ERROR(ENVONLY, "H5AreadComplex: failed to allocate read buffer");
@@ -4050,8 +4048,8 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
     hid_t        memb   = H5I_INVALID_HID;
     H5T_class_t  vlClass;
     size_t       vlSize;
-    jobjectArray jList = NULL;
-    size_t       i, j, x;
+    size_t       i, x;
+    size_t       typeSize;
 
     /* retrieve the java.util.ArrayList interface class */
     jclass    arrCList      = ENVPTR->FindClass(ENVONLY, "java/util/ArrayList");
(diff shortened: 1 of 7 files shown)


java/src/jni/h5util.c
```diff
@@ -4078,335 +4076,297 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
+                /* Get the number of sequence elements */
+                HDmemcpy(&vl_elem, (char *)raw_buf + i * sizeof(hvl_t), sizeof(hvl_t));
+                jsize nelmts = (jsize)vl_elem.len;
+                if (vl_elem.len != (size_t)nelmts)
+                    H5_JNI_FATAL_ERROR(ENVONLY, "translate_rbuf: overflow of number of VL elements");
+
+                if (nelmts < 0)
+                    H5_BAD_ARGUMENT_ERROR(ENVONLY, "translate_rbuf: number of VL elements < 0");
+
+                ret_buflen = ENVPTR->GetArrayLength(ENVONLY, ret_buf);
```
jhendersonHDF, line 4107: No need to do this multiple times for each loop.

java/src/jni/h5util.c
```diff
@@ -4078,335 +4076,297 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
+                /* The list we're going to return: */
+                if (i < ret_buflen) {
+                    jList = ENVPTR->GetObjectArrayElement(ENVONLY, (jobjectArray)ret_buf, (jsize)i);
```
jhendersonHDF, line 4110: Instead, I'd just check if i >= ret_buflen and stop the loop. Otherwise it looks like there's a chance we could pass a NULL jList to translate_rbuf when i >= ret_buflen and probably cause a crash if that ever happens.

Beyond that though, should i ever be >= ret_buflen?

java/src/jni/h5util.c
byrnHDF: Right - I missed that.

java/src/jni/h5util.c
byrnHDF: The idea was that if ret_buf is only initialized (new List() has no members) then we need to create a new member, jList.

byrnHDF: GetObjectArrayElement will error if we try to get a member that doesn't exist.

java/src/jni/h5util.c
byrnHDF: But then again if ret_buf is only partially inited, I should get the existing members before we create new ones?

byrnHDF: Will move the GetArrayLength call out of the loop.

java/src/jni/h5util.c
```diff
@@ -4078,335 +4076,297 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
                 jobject jobj = NULL;
-                switch (vlClass) {
-                    /*case H5T_BOOL: {
-                        jboolean boolValue;
-                        for (x = 0; x < vlSize; x++) {
-                            ((char *)&boolValue)[x] = ((char *)vl_elem.p)[j*vlSize+x];
+                switch (typeSize) {
+                    case sizeof(jbyte): {
+                        jbyte byteValue;
+                        for (x = 0; x < typeSize; x++) {
```
byrnHDF, line 4249: These loops probably should be changed to memcpy, maybe?

jhendersonHDF: Yes, I think it would be more appropriate to just do a memcpy of size `sizeof(jbyte)`. It's still dangerous if typeSize isn't correct, but the loops below are somewhat misleading and may not necessarily be correct.

byrnHDF: Done for next PR.

java/src/jni/h5util.c
byrnHDF: Fixed in next PR

jhendersonHDF (approved):

lrknox (approved):

hyoklee:
@byrnHDF Mac test failed. Revert this PR.

byrnHDF:
This PR cannot cause that failure because it only deals with the JNI. That test failing only uses the C Library!

hyoklee:
I know but your PR triggered it, didn't it?

byrnHDF:
the accum test is broken on macOS 12

hyoklee:
I know.
Why does CI suddenly fail today after your merge?

lrknox:
> Why does CI suddenly fail today after your merge?

macos-latest in the workflows/main.yaml file resolved to macOS 11 previously but to macOS 12 today. PR #2260 changes macos-latest to macos-11 to allow github macOS tests to pass the SWMR tests until the issue is fixed. See  [HDFGroup#2261](https://github.com/HDFGroup/hdf5/issues/2261) and [test/accum.c CTest fails on macOS Monterey · Issue #1289 · HDFGroup/hdf5 (github.com)](https://github.com/HDFGroup/hdf5/issues/1289).

PR #2246, opened 2022-11-09
Add text about release notes

byrnHDF:


CONTRIBUTING.md
```diff
@@ -48,17 +53,69 @@ If the pull request does not have a clear purpose and benefits, it will not be a
...
+
+      Problem/Solution
+
+      Signature
+```
+
+* **Entry Elements** - The elements of the entry - title, problem, solution, and signature - are described in more detail in the table
+below. Descriptions of the problem and the solution should be clear without any ambiguities and should be short without losing clarity or specifics.
+
+    * **Title** - The title or tag should identify one or more categories that will help readers decide if the entry is something they need to study. Can be combined with the `Problem` element
```
lrknox, line 103: Should this "**Title**" line be two lines?

lrknox (approved):

CONTRIBUTING.md
lrknox: I don't think there is any indication of where to put a release note - at the top of the appropriate section either under "New Features" or under "Bug Fixes since HDF5-<previous release>", so that entries within a section appear from latest date to earliest date.  This could be in "Where to Write a Release Note" at the bottom?

byrnHDF: I don't like the title description either!

CONTRIBUTING.md
```diff
@@ -48,17 +53,69 @@ If the pull request does not have a clear purpose and benefits, it will not be a
...
+    * **Solution** - Describe the solution in another paragraph.
+You might also consider the following as you describe the solution:
+        * What was done to resolve the issue?
+        * What is the functional impact?
+        * Is there a workaround – a way for users design their software so as not to encounter the issue? If so, what is the workaround?
+        * For a performance fix, how has the performance improved? Links to published documentation would be good.
+    * **Signature** - Each entry must be signed with the initials of the author, the date in YYYY/MM/DD format, and the JIRA ticket number. The
+following is an example entry written by developer Xavier Zolo on April 16, 2014 about JIRA ticket HDFFV-5555: (XYZ - 2014/04/16, HDFFV-5555). The
+signature is enclosed in parentheses. JIRA or Github numbers should not be used in the description of the problem or the solution. They are like
+abbreviations that customers and external users will not be able to interpret.
```
derobins, line 118: I don't think adding the developer is helpful. As an external person reading RELEASE.txt would you care?

derobins: For internal use, just `git blame` the file.

derobins (approved):

PR #2248, opened 2022-11-10
Bug Fix: Segfault in H5CX_get_vol_wrap_ctx when H5VLwrap_register is called from an application

raylu-hdf:
Jira HDFFV-10881.  A quick and simple fix to make it fail with a relevant error message without touching the other parts of the library.

### DIFF
--- a/src/H5CX.c
+++ b/src/H5CX.c
@@ -1631,7 +1631,8 @@ H5CX_get_vol_wrap_ctx(void **vol_wrap_ctx)
      * which doesn't reset the API context and there is no context, returns an relevant error here
      */
     if (!head || !(*head))
-        HGOTO_ERROR(H5E_CONTEXT, H5E_CANTGET, FAIL, "the API context isn't available, e.g. when an application calls H5VLwrap_register")
+        HGOTO_ERROR(H5E_CONTEXT, H5E_CANTGET, FAIL,
+                    "the API context isn't available, e.g. when an application calls H5VLwrap_register")
 
     /* Check for value that was set */
     if ((*head)->ctx.vol_wrap_ctx_valid)
--- a/src/H5CX.c
+++ b/src/H5CX.c
@@ -1621,12 +1621,17 @@ H5CX_get_vol_wrap_ctx(void **vol_wrap_ctx)
     H5CX_node_t **head      = NULL;    /* Pointer to head of API context list */
     herr_t        ret_value = SUCCEED; /* Return value */
 
-    FUNC_ENTER_NOAPI_NOERR
+    FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity check */
     HDassert(vol_wrap_ctx);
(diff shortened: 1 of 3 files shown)


lrknox:
@raylu-hdf the clang-format Check failed, and doesn't seem to push a fix anymore.  You can see the changes needed by clicking "Details" for the check, or if you have a clone of your fork with the branch on jelly you can use "module load LLVM" to load clang version 13.0.0, check that the clone is up-to-date, and then run "bin/format_source" in the top-level directory of your clone, view the changes with "git diff" followed by "git push", provided your clone only has the one remote.

lrknox (approved):

src/H5CX.c
```diff
@@ -1621,12 +1621,18 @@ H5CX_get_vol_wrap_ctx(void **vol_wrap_ctx)
...
     HDassert(vol_wrap_ctx);
     head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
-    HDassert(head && *head);
+
+    /* No error is expected at this point.  But in case an application calls H5VLwrap_register
+     * which doesn't reset the API context and there is no context, returns an relevant error here
+     */
+    if (!head || !(*head))
+        HGOTO_ERROR(H5E_CONTEXT, H5E_CANTGET, FAIL,
+                    "the API context isn't available, e.g. when an application calls H5VLwrap_register")
```
jhendersonHDF: I think this should be two error checks instead, one for `!head` with an error message such as `H5E_CONTEXT, H5E_UNINITIALIZED, FAIL, "API context is not available"` and one for `!(*head)` with an error message such as `H5E_CONTEXT, H5E_CANTGET, FAIL, "unable to get current API context"`.

Rather than having "when an application calls H5VLwrap_register" in the error message, I think that's something we should document in the doxygen for the function in H5VLconnector_passthru.h. That header file currently doesn't have any doxygen in it, but we should probably add it at least for this function so users are aware that applications shouldn't call it instead of just running into a failure case.

raylu-hdf: I made changes according to your suggestion.

src/H5CX.c
```diff
@@ -1621,12 +1621,20 @@ H5CX_get_vol_wrap_ctx(void **vol_wrap_ctx)
...
-    FUNC_ENTER_NOAPI_NOERR
+    FUNC_ENTER_NOAPI(FAIL)
 
     /* Sanity check */
     HDassert(vol_wrap_ctx);
     head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
-    HDassert(head && *head);
+
+    /* No error is expected at this point.  But in case an application calls H5VLwrap_register
+     * which doesn't reset the API context and there is no context, returns an relevant error here
```
lrknox:
```suggestion
     * which doesn't reset the API context and there is no context, returns a relevant error here
```

jhendersonHDF (approved):

Issue #2250, opened 2022-11-11
[BUG, CVE-2018-11205, HDFFV-10479] 


e4t:
**Description**
The [reproducer](https://github.com/TeamSeri0us/pocs/tree/master/hdf5) of CVE-2018-11205 triggers an out of bound read in memcpy() leading to a SEGV: The crash happens when reading compact chunked data. The hdf5 file seems to specify a selection whose size exceeds the size of the chunked data.

**Expected behavior**
It should be verified that selected data lies within the chunk before attempting to read.

**Platform**
  - 1.10.7 and earlier
 - openSUSE Leap 15.4/ SLE 15 SP4 / openSUSE Tumbleweed
 - gcc7
 - Autotools
 -  config --enable-fortran --enable-unsupported --enable-hl --enable-shared --enable-threadsafe --enable-build-mode=production --enable-cxx --with-pthread

fortnern:
This appears to be fixed in newer versions of HDF5 (tested 1.10 branch and develop). @e4t is there anything to do here?

e4t:
> This appears to be fixed in newer versions of HDF5 (tested 1.10 branch and develop). @e4t is there anything to do here?

@fortnern : I can see this fixed in the head of the 1.14 branch (commit cac778a4c), however, I can still reproduce the segfault with the head of the 1.10 branch (commit 779db127). I cannot test 1.12 ATM as the head seems to require autoconf v. 2.71.
Not sure if you plan to include all CVE fixes into 1.10 as well. The changes that fixed CVE-2018-11205 in the end were different than what I had proposed in #2251 (as stated, I was not convinced of it either). I cannot find a reference to the CVE in the changelog so I assume this was fixed as part of a larger restructuring.

Output 1.14:
```
$ h5dump ./1093663/H5VM_memcpyvv-H5VM.c_1626-oob_read
HDF5 "./1093663/H5VM_memcpyvv-H5VM.c_1626-oob_read" {
GROUP "/" {
   DATASET "Compressed_Data" {
      DATATYPE  H5T_STD_I32BE
      DATASPACE  SIMPLE { ( 100, 20 ) / ( 100, 20 ) }
      DATA {h5dump error: unable to print data

      }
   }
}
}
```
Output 1.10:
```
h5dump ./1093663/H5VM_memcpyvv-H5VM.c_1626-oob_read
HDF5 "./1093663/H5VM_memcpyvv-H5VM.c_1626-oob_read" {
GROUP "/" {
[...]

PR #2251, opened 2022-11-11
Pass compact chunk size info to ensure requested elements are within …

e4t:
…bounds

Pass compact chunk size info to ensure requested elements are within bounds

A selection in a malformed hdf5 file may reference data that does not lie within the dataset.

To avoid out of bounds reads/writes of elements of a compact chunk, pass size info to the read and write functions and check whether all elements are within size before attempting to read/write them.

This fixes CVE-2018-11205 / HDFFV-10479 / Bug #2250

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -172,6 +172,18 @@ Bug Fixes since HDF5-1.13.3 release
 ===================================
     Library
     -------
+    - Fix CVE-2018-11205 / GHSA-8rc4-w39h-mcmg
+
+      Pass compact chunk size info to ensure requested elements are within bounds
+
+      A selection in a malformed hdf5 file may reference data that does not lie
+      within the dataset.
+      To avoid out of bounds reads/writes of elements of a compact chunk, pass
+      size info to the read and write functions and check whether all elements
+      are within size before attempting to read/write them.
+
+      (EFE - 2022/10/01 HDFFV-10479, GH-2250)
+
     - Fix CVE-2018-13867 / GHSA-j8jr-chrh-qfrf
  
       Validate location (offset) of the accumulated metadata when comparing.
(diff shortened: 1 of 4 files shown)


e4t:
I'm less sure about this fix than about the previous ones.
For once, I'm not convinced that this is the correct location to perform these checks. Also, I wonder if there are not other paths `H5D__compact_readvv/writevv()` can be called from where `...->compact.size` would have to be set from as well. I was not able to find any and `...store->compact.buf` seems to be set only in the locations where `...->compact.size` is set. 
Therefore, this should be looked at by someone who's more familiar with the code.

e4t:
I've fixed the formatting as suggested by the format checker although I'm not a great fan of changes like:
```
         HDassert(!ent->locked);
-        ent->locked = TRUE;
-        chunk       = ent->chunk;
+        ent->locked    = TRUE;
+        chunk          = ent->chunk;
```
They require to touch limes that have not been changed and will cause additional effort when using `git annotate` or `git blame`.

fortnern:
This doesn't seem to be the right place to put this check. Generally file consistency checks should be done at the point where the values are read from the file. I tried to figure out what is going on here but it's hard to find specific information about the CVE. Does it reference https://github.com/TeamSeri0us/pocs/blob/master/hdf5/vuln/H5VM_memcpyvv-OUT_OF_BOUND_READ or https://github.com/TeamSeri0us/pocs/blob/master/hdf5/vuln2/H5VM_memcpyvv-heap-buffer-overflow ? I get assertion failures for both of these, and the first seems to occur due to a compact dataset of size 0. If that's the problem, we should be checking when decoding the layout message that the decoded compact dataset size matches the size expected from the dataspace extent and datatype.

fortnern:
Assertions at read time would be fine but it's probably not worth passing around a bunch of extra data to do it.

derobins:
I was hoping to get this into 1.14.0, but I think we need a little more time to consider this. Hopefully this (or some other fix for the CVE) will get into 1.14.1.

fortnern:
The bug seems to be fixed in versions of HDF5 newer than 1.10.7. Why is this change necessary? Is there still a case where we could hit this problem? Is there a reproducer, or a trace from a source code analyzer?

derobins:
This CVE issue has been fixed in develop so I'm going to go ahead and close this.

Issue #2254, opened 2022-11-11
[CVE-2021-46242] Heap use after free in H5AC_unpin_entry()


e4t:
**Describe the bug**
The [reproducer]() of  CVE-2021-46242 exposes a heap use after free error on version 1.10.8 which makes the application crash. This crash can no longer be reproduced on 1.13 - presumably due to commit  aa688c68c447c. This commit, however, seems to be unrelated to the issue that causes the crash and I believe the analysis below remain valid regardless:
The use-after-free occurs in
```
H5AC_unpin_entry(void *thing)
{
...
    entry_ptr = (H5AC_info_t *)thing;
    cache_ptr = entry_ptr->cache_ptr;
...

    if (cache_ptr != NULL && cache_ptr->log_info != NULL)
        if (cache_ptr->log_info->logging)
-->             if (cache_ptr->log_info->logging)
...
}
```
called from:
```
herr_t
H5F__dest(H5F_t *f, hbool_t flush)
{
...
            if (f->shared->drvinfo)
                if (H5AC_unpin_entry(f->shared->drvinfo) < 0)
...
}
```
Valgrind reveals that `f->shared->drvinfo` was freed in 
```
H5F__super_read(..)
{
...
   if (ret_value < 0) {
        /* Unpin and discard drvinfo cache entry */
        if (f->shared->drvinfo) {
            if (H5AC_unpin_entry(f->shared->drvinfo) < 0)
                HDONE_ERROR(H5E_FILE, H5E_CANTUNPIN, FAIL, "unable to unpin driver info")

            /* Evict the driver info block from the cache */
-->        if (sblock && H5AC_expunge_entry(f, H5AC_DRVRINFO, sblock->driver_addr, H5AC__NO_FLAGS_SET) < 0)
                HDONE_ERROR(H5E_FILE, H5E_CANTEXPUNGE, FAIL, "unable to expunge driver info block")
        } /* end if */
 ...
  }
...
}
```
Since `H5C_expunge_entry()` (called from `H5AC_expunge_entry()`)  sets the `H5C__FLUSH_INVALIDATE_FLAG`, the entry is destroyed (ie. deallocated). The pointer `f->shared->drvinfo` is however not NULLed. When `H5AC_unpin_entry()` is called again from
`H5F__dest()` a use-after-free occurs.
  
**Expected behavior**
[...]

PR #2255, opened 2022-11-11
CVE 2021 46242 develop

e4t:
Fix two potential causes of usr-after-free:

- When evicting the superblock, NULL the corresponding entry
- When evicting driver info block, NULL the corresponding entry (CVE-2021-46242, Bug #2254)

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -172,6 +172,18 @@ Bug Fixes since HDF5-1.13.3 release
 ===================================
     Library
     -------
+    - Fix CVE-2021-46242 / GHSA-x9pw-hh7v-wjpf
+
+      When evicting driver info block, NULL the corresponding entry.
+
+      Since H5C_expunge_entry() called (from H5AC_expunge_entry()) sets the  flag
+      H5C__FLUSH_INVALIDATE_FLAG, the driver info block will be freed. NULLing
+      the pointer in f->shared->drvinfo will prevent use-after-free  when it is
+      used in other functions (like  H5F__dest()) - as other places will check
+      whether the pointer is initialized before using its value.
+
+      (EFE - 2022/09/29 GH-2254)
+
     - Fix CVE-2018-13867 / GHSA-j8jr-chrh-qfrf
  
       Validate location (offset) of the accumulated metadata when comparing.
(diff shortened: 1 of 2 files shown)


e4t:
This needs review by someone who is familiar with `H5F__super_read()`: I'm unsure whether `f->shared->sblock = NULL` is really needed. I've added a check whether it's value is the same as `sblock` as I cannot rule out, that it may have retained its previous value.

release_docs/RELEASE.txt
```diff
@@ -172,6 +172,18 @@ Bug Fixes since HDF5-1.13.3 release
...
     -------
+    - Fix CVE-2021-46242 / GHSA-x9pw-hh7v-wjpf
+
+      When evicting driver info block, NULL the corresponding entry.
+
+      Since H5C_expunge_entry() called (fromH5AC_expunge_entry()) sets the  flag
+      H5C__FLUSH_INVALIDATE_FLAG, the driver info block will be freed. NULLing
+      the pointer in f->shared->drvinfo will prevent use-after-free  when it is
+      used in other functions (like  H5F__dest()) - as other places will check
+      whether the pointer is initialzed before using its value.
```
hyoklee: @e4t Please fix typo. Otherwise, this is amazingly good patch! Thank you!

hyoklee (changes requested):

release_docs/RELEASE.txt
e4t: > @e4t Please fix typo. Otherwise, this is amazingly good patch! Thank you!

Thank you! Typo fixed.
I've rechecked the code, I'm now convinced that `f->shared->sblock` should be NULLed if expunged in the error path so I removed it. It seems there has been trouble with this before as the same function contains the code:
```
#if 1 /* bug fix test code -- tidy this up if all goes well */ /* JRM */
                    /* KLUGE ALERT!!                                                                               
                     *                                                                                             
                     * H5F__super_ext_write_msg() expects f->shared->sblock to                                     
                     * be set -- verify that it is NULL, and then set it.                                          
                     * Set it back to NULL when we are done.                                                       
                     */
                    HDassert(f->shared->sblock == NULL);
                    f->shared->sblock = sblock;
#endif /* JRM */
```
twice. Maybe the `HDassert(f->shared->sblock == NULL)` should be moved to the beginning of the function.

hyoklee: Yes, this code is very complicated and requires a careful review. I really appreciate your effort to dig deep!

jhendersonHDF (approved):

lrknox (approved):

lrknox:
Test file POC1-H5AC_unpin_entry-useafterfree fails to open with error message h5dump error: unable to open file "POC1-H5AC_unpin_entry-useafterfree"

PR #2256, opened 2022-11-13
Make sure testsuite can handle multiple dependent scripts

e4t:
Commit afc54d75a19 to the test suite to eliminate a separate shell script did not take into account that there may be multiple dependent test scripts which resulted in the message:
`/bin/sh: line 7: [: too many arguments`.
when running the test suite.

Beware that this new version still makes use of non-posix GNU extensions to `test`.


hyoklee (approved):
Thanks for another great fix!

lrknox (approved):

byrnHDF (approved):

Issue #2257, opened 2022-11-14
[BUG, CVE-2020-10812] Null pointer dereference SEGV in Exit Handler 


e4t:
**Description**
The reproducer of CVE-2020-10812 triggers a segfault in `H5F_get_nrefs()` - (from gdb):
```
0x00007ffff7b1bf74 in H5F_get_nrefs (f=f@entry=0x44bbb0) at H5Fquery.c:568
568         FUNC_LEAVE_NOAPI(f->shared->nrefs)
```
The backtrace shows:
```
#0  0x00007ffff7b1bf74 in H5F_get_nrefs (f=f@entry=0x44bbb0) at H5Fquery.c:568
#1  0x00007ffff7d1e836 in H5VL__native_file_close (file=0x44bbb0, dxpl_id=<optim
    at H5VLnative_file.c:763
#2  0x00007ffff7d0c3be in H5VL__file_close (cls=<optimized out>, req=0x0, dxpl_i
    obj=<optimized out>) at H5VLcallback.c:4309
#3  H5VL_file_close (vol_obj=vol_obj@entry=0x44fd50, dxpl_id=792633534417207304,
#4  0x00007ffff7b12806 in H5F__close_cb (file_vol_obj=0x44fd50, request=<optimiz
#5  0x00007ffff7b935c9 in H5I__mark_node (key=0x0, _udata=<synthetic pointer>, _
#6  H5I_clear_type (type=type@entry=H5I_FILE, force=force@entry=false, app_ref=a
#7  0x00007ffff7b13c91 in H5F_term_package () at H5Fint.c:173
#8  0x00007ffff7a5b542 in H5_term_library () at H5.c:464
#9  0x00007ffff78450c5 in __run_exit_handlers () from /usr/lib64/libc.so.6
#10 0x00007ffff7845250 in exit () from /usr/lib64/libc.so.6
```
Thus, the crash happens in an `exit()` handler.
[...]

PR #2258, opened 2022-11-14
Hot fix for CVE-2020-10812 / Bug #2257

e4t:
CVE-2020-10812 unveils a more fundamental design flaw in H5F__dest(): this function returns FAIL if one of multiple operations fail (in this case H5AC_prep_for_file_close()) while it still proceeds to prepare the close operation, free the 'shared' member in struct H5F_t and ulimately deallocate the structure itself.
When H5F__dest() signals back FAIL to the caller, the caller itself (H5F_try_close() in this case) will fail. This failure is signalled up the stack, thus the file will not be considered closed and another attempt will be made to close it - latest in the exit handler. The next attempt to close will however need the already deallocated H5F_t structure and the H5T_shared_t structure in its 'shared' member, however.
This fix papers over the failure of H5AC_prep_for_file_close() by not changing the return status of H5F__dest() to fail. There are numerous other opportunities where this will happen.
This may call for a more fundamental solution.

### DIFF
--- a/src/H5Fint.c
+++ b/src/H5Fint.c
@@ -1380,7 +1380,7 @@ H5F__dest(H5F_t *f, hbool_t flush)
          */
         if (H5AC_prep_for_file_close(f) < 0)
             /* Push error, but keep going */
-            HDONE_ERROR(H5E_FILE, H5E_CANTFLUSH, FAIL, "metadata cache prep for close failed")
+            HDONE_ERROR(H5E_FILE, H5E_CANTFLUSH, ret_value, "metadata cache prep for close failed")
 
         /* Flush at this point since the file will be closed (phase 2).
          * Only try to flush the file if it was opened with write access, and if


e4t:
Note: I've pull-requested this merely as a starting point for a discussion how to properly address this issue. Please also check the analysis on Bug #2257. The patch here is only covering up the problem, not solving it. I've therefore not created a release note entry, yet.

jhendersonHDF:
> [CVE-2020-10812](https://github.com/advisories/GHSA-g838-rwc8-xv7q) unveils a more fundamental design flaw in H5F__dest(): this function returns FAIL if one of multiple operations fail (in this case H5AC_prep_for_file_close()) while it still proceeds to prepare the close operation, free the 'shared' member in struct H5F_t and ulimately deallocate the structure itself.

Hey @e4t, I ended up running into something similar when trying to fix the issue in https://forum.hdfgroup.org/t/page-buffer-with-phdf5/10435, except it was an issue with the `sblock` structure not being available due to a failure in the middle of file opening. As you mentioned in #2257, I think a proper fix would probably be to go find all the places where the `shared` structure and other members of a file object are needed at close time and do the best we can of cleaning things up (NULLing them along the way) while checking that the fields we're accessing are valid and simply failing if they aren't. More than a few places in the library still only assert on in-memory contents that may or may not be valid before accessing them. Some of the cases are instances where we assume everything should be in order since we've had full control over the in-memory contents but hadn't yet exposed these types of bugs through testing. Other cases of course come from in-memory contents that we crafted from reading data out of a file which we assume is well-formed.

In any case, H5F__dest likely needs a rethink because I believe there are more than a few similar bugs in it that simply haven't been exposed yet.

derobins:
Already fixed via a different commit

PR #2263, opened 2022-11-17
Simplify & fix check for szlib encoder

### LINKED ISSUE #2262: [BUG] configure script thinks szip encoder is not enabled but it is
**Describe the bug**
When configured with `--with-szlib=yes`, hdf5 fails to determine that szip has been compiled with encoder support, when `-Werror=implicit-function-declaration` is in effect. `-Werror=implicit-function-declaration` is in effect by default with the version of the clang compiler included with Xcode 12 and later on macOS.

Even though szip is installed with encoder support, hdf5 configure output is:

```
checking for szlib encoder... no
```

**Expected behavior**
```
checking for szlib encoder... yes
```

 - HDF5 version: 1.12.2
 - OS and version: macOS 10.15.7
 - Compiler and version: Apple clang version 12.0.0 (clang-1200.0.32.29)
 - Build system (e.g. CMake, Autotools) and version: autotools
 - Any configure options you specified --with-szlib=yes
 - MPI library and version (parallel HDF5): n/a

**Additional context**
config.log contains:

```
configure:29563: checking for szlib encoder
configure:29614: ccache /usr/bin/clang -o conftest  -std=c99  -Wall -Warray-bounds -Wcast-qual -Wconversion -Wdouble-promotion -Wextra -Wformat=2 -Wframe-larger-than=16384 -Wimpli
cit-fallthrough -Wnull-dereference -Wunused-const-variable -Wwrite-strings -Wpedantic -Wvolatile-register-var -Wno-c++-compat    -pipe -Os -isysroot/Library/Developer/CommandLineT
ools/SDKs/MacOSX10.15.sdk -arch x86_64     -I/opt/local/include -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk  -L/opt/local/lib -Wl,-headerpad_max_install_name
s -Wl,-rpath,/opt/local/lib/libgcc -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -arch x86_64 conftest.c -lsz -lz -ldl -lm  >&5
conftest.c:121:21: error: implicitly declaring library function 'exit' with type 'void (int) __attribute__((noreturn))' [-Werror,-Wimplicit-function-declaration]
                    exit(0);
                    ^
conftest.c:121:21: note: include the header <stdlib.h> or explicitly provide a declaration for 'exit'
1 error generated.
configure:29614: $? = 1
configure: program exited with status 1
```
```
configure:29637: result: no
```

The configure test in question is:

```
  AC_CACHE_VAL([hdf5_cv_szlib_can_encode],
        [AC_RUN_IFELSE(
            [AC_LANG_PROGRAM([
                #include "szlib.h"
            ],[[
                /* SZ_encoder_enabled returns 1 if encoder is present */
                if(SZ_encoder_enabled() == 1)
                    exit(0);
                else
                    exit(1);
            ]])]
        , [hdf5_cv_szlib_can_encode=yes], [hdf5_cv_szlib_can_encode=no],)]
   )
```

`exit` is being called without `<stdlib.h>` (where `exit` is declared) having been included.

I will submit a PR that fixes the problem by simplifying the code to use `return` instead of `exit`.

ryandesign:
Return the result rather than setting the exit code. `return` is a language keyword whereas `exit` is a function for which the <stdlib.h> header has to be included which it wasn't in this test, therefore the test would previously fail to identify that the encoder was enabled if `-Werror=implicit-function-declaration` was used, which it is by default with clang from Xcode 12 and later.

Fixes #2262

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1737,10 +1737,7 @@ if test "x$HAVE_SZLIB" = "xyes" -a "x$HAVE_SZLIB_H" = "xyes"; then
                 #include "szlib.h"
             ],[[
                 /* SZ_encoder_enabled returns 1 if encoder is present */
-                if(SZ_encoder_enabled() == 1)
-                    exit(0);
-                else
-                    exit(1);
+                return SZ_encoder_enabled() != 1;
             ]])]
         , [hdf5_cv_szlib_can_encode=yes], [hdf5_cv_szlib_can_encode=no],)]
    )


byrnHDF (approved):

hyoklee (approved):
Thank you for the great patch @ryandesign !

qkoziol (approved):

PR #2266, opened 2022-11-21
Develop jni trans

byrnHDF:


byrnHDF:
Code is complete except that string[] variable vlen fail in the H5write calls. The buffer looks correct and seems to match all other tests/examples. Creating PR for others to view for suggestions.

java/test/TestH5A.java
```diff
@@ -1863,4 +1870,134 @@ public void testH5AArraywr()
...
+
+        for (int j = 0; j < dims.length; j++)
+            lsize *= dims[j];
+
+        ArrayList[] arr_readbuf = new ArrayList[6];
+        for (int j = 0; j < lsize; j++)
+            arr_readbuf[j] = new ArrayList<String>();
+
+        try {
+            H5.H5AreadVL(att_str_id, atype_str_id, arr_readbuf);
```
jhendersonHDF, line 1975: Are the read and write going to assume that the user is passing in an ArrayList[]? I can see that being problematic for several reasons.

java/src/jni/h5util.c
```diff
@@ -4125,64 +4641,50 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
-                        (jList = (jobjectArray)ENVPTR->NewObject(ENVONLY, arrCList, arrListMethod, 0)))
-                        H5_OUT_OF_MEMORY_ERROR(ENVONLY,
-                                               "translate_rbuf: failed to allocate list read buffer");
+                    translate_atomic_rbuf(ENVONLY, jList, memb, memb_vlClass,
+                                          ((char *)raw_buf) + i * typeSize + memb_offset);
                 }
-
-                translate_rbuf(ENVONLY, jList, memb, vlClass, (jsize)1, objBuf);
-                if (!found_jList) {
+                H5Tclose(memb);
```
jhendersonHDF: This close call should be moved inside the loop. Otherwise `memb` will be leaked on each iteration.

java/src/jni/h5util.c
```diff
@@ -4477,91 +4911,36 @@ translate_wbuf(JNIEnv *env, jobjectArray in_buf, jlong mem_type_id, H5T_class_t
...
-                }
+                translate_atomic_wbuf(ENVONLY, jobj, mem_type_id, type_class,
+                                      ((char *)raw_buf) + i * typeSize);
                 ENVPTR->DeleteLocalRef(ENVONLY, jobj);
             }
             break;
-        } /* H5T_FLOAT */
-        case H5T_REFERENCE: {
+        }
+        case H5T_STRING: {
```
jhendersonHDF, line 4926: The implementation below for converting strings looks the same as for the group of types above. I'm guessing it can just use the same logic in the way that translate_rbuf includes H5T_STRING among the types that do the same thing for conversion?

java/src/jni/h5util.c
```diff
@@ -4125,64 +4641,50 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
-                        (jList = ENVPTR->GetObjectArrayElement(ENVONLY, (jobjectArray)ret_buf, (jsize)i)))
-                        found_jList = JNI_FALSE;
-                }
-                if (NULL == jList) {
-                    if (NULL ==
-                        (jList = (jobjectArray)ENVPTR->NewObject(ENVONLY, arrCList, arrListMethod, 0)))
-                        H5_OUT_OF_MEMORY_ERROR(ENVONLY,
-                                               "translate_rbuf: failed to allocate list read buffer");
+                    translate_atomic_rbuf(ENVONLY, jList, memb, memb_vlClass,
+                                          ((char *)raw_buf) + i * typeSize + memb_offset);
```
jhendersonHDF: More of a nitpick since I'm assuming this works correctly as is, but for "container" types like compound, vlen, etc. it would probably be more efficient to check whether each member is also a container type and pass them directly back to translate_rbuf/wbuf rather than passing to translate_atomic_rbuf/wbuf just to have those pass back up to translate_rbuf/wbuf. It would also make more sense considering the "atomic" part of the functions names and would save on some code duplication, but this is mostly just a design issue.

java/src/jni/h5util.c
```diff
@@ -4079,6 +4081,521 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
+            jobject jobj        = NULL;
+            char   *cstr        = raw_buf;
+            htri_t  is_variable = 0;
+
+            if ((is_variable = H5Tis_variable_str(mem_type_id)) < 0)
+                H5_LIBRARY_ERROR(ENVONLY);
+
+            /* Convert each element */
+            if (is_variable) {
+                char **new_buf = (char *)raw_buf;
```
jhendersonHDF: This line looks suspicious to me. Did you mean `char **new_buf = &(char *)raw_buf` or even `char **new_buf = &cstr`? (cstr looks to be unused).

java/src/jni/h5util.c
```diff
@@ -4079,6 +4081,521 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
+                H5_BAD_ARGUMENT_ERROR(ENVONLY, "translate_atomic_wbuf: number of VL elements < 0");
+
+            vl_elem.len = (size_t)jnelmts;
+
+            if (NULL == (vl_elem.p = HDmalloc((size_t)jnelmts * vlSize)))
+                H5_OUT_OF_MEMORY_ERROR(ENVONLY, "translate_atomic_wbuf: failed to allocate vlen ptr buffer");
+
+            translate_wbuf(ENVONLY, (jobjectArray)in_obj, memb, vlClass, (jsize)jnelmts, vl_elem.p);
+
+            HDmemcpy((char *)raw_buf + sizeof(hvl_t), &vl_elem, sizeof(hvl_t));
```
jhendersonHDF: Why the addition of `sizeof(hvl_t)` here? Do we need to avoid copying over the initial `sizeof(hvl_t)` bytes of raw_buf?

java/src/jni/h5util.c
```diff
@@ -4079,6 +4081,521 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
+
+            if (jnelmts < 0)
+                H5_BAD_ARGUMENT_ERROR(ENVONLY, "translate_atomic_wbuf: number of array elements < 0");
+
+            if (NULL == (objBuf = HDmalloc((size_t)jnelmts * vlSize)))
+                H5_OUT_OF_MEMORY_ERROR(ENVONLY, "translate_atomic_wbuf: failed to allocate buffer");
+
+            translate_wbuf(ENVONLY, array, memb, vlClass, (jsize)jnelmts, objBuf);
+
+            HDmemcpy((char *)raw_buf + vlSize * jnelmts, (char *)objBuf, vlSize * jnelmts);
```
jhendersonHDF: Same here with the addition of `vlSize * jnelmts`?

java/test/TestH5A.java
byrnHDF: No only when it is a container type.

java/src/jni/h5util.c
byrnHDF: Done

java/src/jni/h5util.c
byrnHDF: I was waiting for the fix for variable strings - done.

java/src/jni/h5util.c
byrnHDF: Sounds correct, will need to investigate.

java/src/jni/h5util.c
byrnHDF: yes cstr is now removed.
The assignment is because I couldn't get the compiler to agree and this worked (debugged the actual values of the pointer to verify).

java/src/jni/h5util.c
byrnHDF: Will debug it, I agree that looks strange.

java/src/jni/h5util.c
byrnHDF: Might be a holdover from when the loops were there before being moved.

java/src/jni/h5aImp.c
```diff
@@ -1154,7 +1156,10 @@ Java_hdf_hdf5lib_H5_H5AreadVL(JNIEnv *env, jclass clss, jlong attr_id, jlong mem
             if (sid >= 0)
                 H5Sclose(sid);
         }
-
+        if (is_variable) {
+            for (size_t i = 0; i < (size_t)vl_array_len; i++)
+                HDfree(((char **)readBuf)[i]);
```
jhendersonHDF, line 1161: Isn't this what H5Treclaim is supposed to be for?

java/test/TestH5D.java
```diff
@@ -966,25 +966,121 @@ public void testH5Dvlen_get_buf_size()
     }
 
     @Ignore
```
jhendersonHDF, line 968: Do you want to test these things in this test file or are they still to be ignored for some reason?

java/src/jni/h5util.c
jhendersonHDF: The cast here is just wrong since you're assigning a `char *` to a `char **`. It may be that a `char **` is getting passed in for `raw_buf` and making things work by coincidence but this is just not quite right.

byrnHDF: variable len strings are just weird! The issue is that var string data are pointers , everything else is the data. I haven't tried removing the right hand cast - that might make it less wrong looking.

java/src/jni/h5aImp.c
byrnHDF: No, reclaim is for the internal library allocs, this is for the top-level var types that JNI creates.

java/test/TestH5D.java
byrnHDF: I have left that one there as I am undecided on its usefulness. Maybe I should just merge the len check into the test that actually reads. I'm thinking the len check might not actually even need the H5write call.

java/src/jni/h5util.c
jhendersonHDF: It's not just that it's wrong looking, it's misleading about what's going on here since var. strings are a different buffer type. We should probably have a separate small function for translating var. strings that accepts a `char **`, but if we want to keep it simple and all in one function here, then I really think this should be fixed like:

1. Change `raw_buf` to be a `void *` rather than a `jbyte *`
2. At the start of this function, declare something like `char *char_buf = (char *)raw_buf` to be used in all the places where `raw_buf` is currently cast to a `char *` throughout this function
3. Here, you can then do `char **var_str_buf = (char **) raw_buf` and then do `NewStringUTF(ENVONLY, *var_str_buf)` (no need to cast to `char *`)

It's still playing games with hiding types by cramming a `char **` into a `void *`, but it's more explicit about the fact that this function can take and work with different buffer types that aren't just a single `jbyte *`. The rest of this PR looks fine to me.

byrnHDF: okay I can do that

java/test/TestH5D.java
byrnHDF: Missed this in the last update - will address in next PR.

java/src/jni/h5util.c
byrnHDF: done.

java/src/jni/h5util.c
byrnHDF: Working on a solution for non-ArrayList objects as data (next PR soon) which may address this.

jhendersonHDF (approved):

lrknox (approved):

PR #2267, opened 2022-11-23
Develop ubuntu 20.04

lrknox:
In Ubuntu 20.04, apt install openmpi-bin also installed openmpi-common and libopenmpi-dev as dependencies.  In Ubuntu 22.04, libopenmpi-dev is not installed as a dependency of openmpi-bin, and without it, links such as libmpi.so are not created, resulting in autotools configure failure when trying to compile a basic C program with mpicc.  Specifically adding libopenmpi-dev restores the previous behavior.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-* @lrknox @derobins @byrnHDF @fortnern @jhendersonHDF @ChristopherHogan @gnuoyd @qkoziol @vchoi-hdfgroup @bmribler @raylu-hdf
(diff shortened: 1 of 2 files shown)


lkurz (approved):

gheber:
Same issue on Debian. This fixes that as well.

jhendersonHDF:
Relevant: https://bugs.launchpad.net/ubuntu/+source/openmpi/+bug/1870780

>* Don't have openmpi-bin depend on libopenmpi-dev.
    THIS MAY BREAK THINGS EXPECTING libopenmpi-dev TO SHIP mpicc*, etc.
    THEY SHOULD DEPEND ON MPI-DEFAULT-DEV instead.

derobins (approved):

lrknox:
PR was merged before I finished testing apt install openmpi-bin mpi-default-dev.  It fixes configure problem, and if there are no other issues I’ll create a new PR.

PR #2269, opened 2022-11-30
H5E_ERR_CLS_g symbol missing when compiling C++ plugins

vasole:
C++ HDF5 filter plugin SZ3 fails to build under windows with failure at linking time unless that extern "C" block is added.

image

### DIFF
--- a/src/H5Epublic.h
+++ b/src/H5Epublic.h
@@ -64,7 +64,6 @@ typedef struct H5E_error2_t {
  * can miss the symbol at linking time (for instance with Visual Studio)
  *
  */
-
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/src/H5Epublic.h
+++ b/src/H5Epublic.h
@@ -59,8 +59,20 @@ typedef struct H5E_error2_t {
 #endif /* H5private_H */
 
 /* HDF5 error class */
+/* 
+ * If this extern "C" block is not added, filter plugins compiled with C++ (ex. SZ3)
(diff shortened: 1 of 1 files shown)


src/H5Epublic.h
```diff
@@ -59,8 +59,15 @@ typedef struct H5E_error2_t {
...
 /* HDF5 error class */
+/* Extern "C" block needed to compile C++ filter plugins with some compilers */
+#ifdef __cplusplus
+extern "C" {
+#endif
 #define H5E_ERR_CLS (H5OPEN H5E_ERR_CLS_g)
 H5_DLLVAR hid_t H5E_ERR_CLS_g;
+#ifdef __cplusplus
+}
+#endif
```
derobins, line 70: Instead of having two extern C blocks, just move this to the extern C block later in this file.

vasole: With so many macros depending on each other, I made the safest, most conservative, change.

If I move that piece of code:

```
#define H5E_ERR_CLS (H5OPEN H5E_ERR_CLS_g)
H5_DLLVAR hid_t H5E_ERR_CLS_g;
``` 

to the extern C block later on in the file, other plugins fail at compilation time.

The alternative is to extend the extern C block instead of moving the code.

That also works, but would it be acceptable for you?

derobins: Let's leave it as-is and merge it.

derobins (approved):

lrknox (approved):

PR #2272, opened 2022-12-01
WIP: Remove file "post open" VOL callback

fortnern:
Also add "get_wrap_ctx_pre_open" VOL callback and add "wrap_ctx" parameter to file create and file open VOL
callbacks to allow this to work.

The main aim of these changes is to set a valid wrap context *before* making the file open/create VOL callback. This removes the need for post open since we can do everything inside the main open/create calls. This should also allow more flexibility for VOL connector authors to do things in the open/create calls that were previously precluded due to not having a wrap context.

Passthrough authors need to implement the "get_wrap_ctx_pre_open" callback, but they can ignore the "wrap_ctx" parameter to file create/open unless they want to use the same wrap context for all objects in a file.

### DIFF
--- a/src/H5ES.c
+++ b/src/H5ES.c
@@ -143,7 +143,8 @@ H5ESinsert_request(hid_t es_id, hid_t connector_id, void *request)
         HGOTO_ERROR(H5E_EVENTSET, H5E_CANTINSERT, FAIL, "can't insert request into event set")
 
 done:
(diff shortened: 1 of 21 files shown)


fortnern:
We've decided not to go this route, at least for 1.14. I'll leave this up for a bit longer as a draft incase anyone wants to weigh in on the post open/object wrapping code

derobins:
Closing this for now

Issue #2270, opened 2022-12-01
[BUG] h5repack in 1.12.2 displaying output without verbosity options


EricR86:
**Describe the bug**
`h5repack` is displaying output to terminal without any option to do so.

**Expected behavior**
No output to terminal

**Platform (please complete the following information)**
 - HDF5 version (if building from a maintenance branch, please include the commit hash)
 1.12.2

 - OS and version
 Ubuntu 20.04

 - Compiler and version
 gcc 11.2.0

 - Build system (e.g. CMake, Autotools) and version
Autoconf  2.69

 - Any configure options you specified
 Specified a `prefix` for install

 - MPI library and version (parallel HDF5)
 No MPI

**Additional context**
This problem exists as well in the conda build on the `conda-forge` channel. Verified this issue without conda from building from source. Older versions do not exhibit this behavior.

byrnHDF:
What documentation indicates that no output should be produced?

EricR86:
[This is the reference I'm using and there is no explicit documentation that output should be produced](https://portal.hdfgroup.org/display/HDF5/h5repack). There *is* an explicit option you can enable however with `verbosity`:
"-v, --verbose           Verbose mode, print object information"

The output I get from `h5repack` seems to be of that type:
```
-----------------------------------------
 Type     Filter (Compression)     Name
-----------------------------------------
 group                       /
...
```

byrnHDF:
Understandable - and do you get the same or more if you add the verbose option?
Just trying to understand both the expectation and the actual result.

EricR86:
My belief is the expectation is nothing to output since this behavior did not exist in earlier versions.

When I add `-v` I get additional preamble and the `Filter` column filled out. Otherwise it is the same. Looks like:
```
No all objects to modify layout
All objects to apply filter are...
 All with GZIP, parameter 1
Making new file ...
-----------------------------------------
 Type     Filter (Compression)     Name
-----------------------------------------
```

byrnHDF:
I will investigate today and maybe something was changed that did not check the verbose setting.

byrnHDF:
Found it, a check for timing had an improper if/else construct in h5repack_copy source file.

PR #2275, opened 2022-12-02
1.10 h5repack should only print if verbose

byrnHDF:
Also change dump test to use long option
Fix for #2270


lrknox:
Do we need a RELEASE.txt entry?

lrknox (approved):

PR #2277, opened 2022-12-06
Fix for HDFFV-10840: Instead of using fill->buf for datatype conversi…

vchoi-hdfgroup:
…on (#2153)

* Fix for HDFFV-10840: Instead of using fill->buf for datatype conversion if it is large enough, a buffer is allocated regardless so that the element in fill->buf can later be reclaimed.
Valgrind is run on test/set_extent.c and there is no memory leak.

* Add information of this fix to release notes.


lrknox (approved):

lrknox:
All tests passed in action #3046.

PR #2279, opened 2022-12-06
WIP: Improved const usage in dataset write calls

derobins:
The data array parameter in dataset write calls should be `const void * const buf[]` instead of `const void * buf[]` as the latter allows for the const-ness to be subverted.  This affects both the VOL and VFD layers.


derobins:
This is incomplete, btw - the VOL changes are coming later today. This is also missing parallel. That "flexible const pointer" hack needs some fixin'.

byrnHDF (approved):

derobins:
Closing for now

PR #2301, opened 2022-12-14
Adds 'unused' hints for MDS io_info parameters

derobins:


byrnHDF (approved):

derobins:
This is for sure going to fail the formatting check.

For one of the calls, I opted to use (void) instead of an ATTR_UNUSED macro. We don't have one for "used in debug parallel only" and those macros are already ridiculous.

derobins:
Once this is merged, I'll fix the GitHub actions that are incorrectly using the default build mode.

src/H5Dchunk.c
```diff
@@ -2466,7 +2466,12 @@ H5D__chunk_cacheable(const H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info
     FUNC_ENTER_PACKAGE
 
     /* Sanity check */
+#ifdef H5_HAVE_PARALLEL
     HDassert(io_info);
+#else
+        (void)
+        io_info;
```
derobins: lol - yay, formatter

byrnHDF: strange -  I wonder if there is a setting for that?

jhendersonHDF (approved):

src/H5Dchunk.c
```diff
@@ -2466,7 +2466,13 @@ H5D__chunk_cacheable(const H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info
     FUNC_ENTER_PACKAGE
 
     /* Sanity check */
+#ifdef H5_HAVE_PARALLEL
     HDassert(io_info);
+#else
+        /* clang-format off */
```
derobins: wtf? Must be due to the FUNC_ENTER macro.

derobins: This is less ugly, and I have zero interest in adding to the already-over-complicated ATTR_UNUSED scheme

byrnHDF (approved):

PR #2308, opened 2022-12-16
Converts the YAML build mode arrays to objects in special runs

derobins:


.github/workflows/main.yml
```diff
@@ -174,12 +174,12 @@ jobs:
             flags: ""
             run_tests: false
             thread_safety:
-              - enabled: false
-                text: ""
+              enabled: false
+              text: ""
```
derobins, line 178: The matrix setup creates an array of objects. When copying one of those down here, this adds a single element array to the matrix instead of an object, causing the build mode to be blank in the configure step.

PR #2311, opened 2022-12-16
1.12: Brings the updated CI in from develop

derobins:
Does not include the -Werror checks, changes 1.14 API to 1.12 where needed.


byrnHDF (approved):

hyoklee (changes requested):
CI fails so I don't have to review further until everything passes.

derobins:
> CI fails so I don't have to review further until everything passes.

Looks like it's a bug in the mirror VFD. I think we also tripped over this in develop. I'll fix it.

PR #2324, opened 2022-12-19
Fix doxygen warnings and remove javadocs

byrnHDF:


derobins (approved):

release_docs/RELEASE.txt
```diff
@@ -81,7 +81,8 @@ New Features
       Autotools:
           --enable-memory-alloc-sanity-check
 
-      CMake:
+      CMake:    <tab type="user" url="@ref VOL_Connector" title="VOL Connector Author Guide" />
```
lrknox: I'm not sure what this CMake: line does in RELEASE.txt.  Does it belong here?

I suppose it will be cleared out soon anyway.  the 1_14 and 1_14_10 branches don't have the extra text.

lrknox (approved):

lrknox (approved):

PR #2325, opened 2022-12-20
Fix up the H5Pencode2 test for virtual layout.

vchoi-hdfgroup:


test/enc_dec_plist.c
```diff
@@ -44,62 +44,57 @@ test_encode_decode(hid_t orig_pl, H5F_libver_t low, H5F_libver_t high, hbool_t s
     }
     H5E_END_TRY;
 
-    if (support_virtual && high < H5F_LIBVER_V110)
-        VERIFY(ret, FAIL, "H5Pencode2");
-    else {
+    VERIFY(ret, SUCCEED, "H5Pencode2");
```
derobins: If there's an error here, it won't cause the test to fail. The VERIFY, etc. macros only work in the testhdf5 program, which checks the global error value.

derobins (changes requested):
The testhdf5 macros have to be converted to use the macros in h5test.h

derobins (approved):

PR #2329, opened 2022-12-20
Port h5dchunk_iter_hsizet_size #2074 to hdf5_1_12 branch

mkitti:
* Align arg types of H5D_chunk_iter_op_t with H5Dget_chunk_info

* Modify chunk_info test to for unsigned / hsize_t types

* Fix types in test

* Port of #2074 to hdf5_1_12 branch

cc: @derobins

### DIFF
--- a/src/H5D.c
+++ b/src/H5D.c
@@ -1138,32 +1138,6 @@ done:
  *              H5D_chunk_iter_op_t cb  IN: User callback function, called for every chunk.
  *              void *op_data           IN/OUT: Optional user data passed on to user callback.
  *
- * Callback information:
- *      H5D_chunk_iter_op_t is defined as:
- *
- *        typedef int (*H5D_chunk_iter_op_t)(
- *            const hsize_t *offset,
- *            unsigned filter_mask,
- *            haddr_t addr,
- *            hsize_t size,
- *            void *op_data);
- *
- *      H5D_chunk_iter_op_t parameters:
- *          hsize_t *offset;        IN/OUT: Logical position of the chunk’s first element in units of dataset
- *                                          elements
- *          unsigned filter_mask;   IN: Bitmask indicating the filters used when the chunk was written haddr_t
- *          addr;                   IN: Chunk address in the file
- *          hsize_t;                IN: Chunk size in bytes, 0 if the chunk does not exist
- *          void *op_data;          IN/OUT: Pointer to any user-defined data associated with the operation.
- *
- *      The return values from an operator are:
- *          Zero (H5_ITER_CONT) causes the iterator to continue, returning zero when all
- *              elements have been processed.
- *          Positive (H5_ITER_STOP) causes the iterator to immediately return that positive
- *              value, indicating short-circuit success.
- *          Negative (H5_ITER_ERROR) causes the iterator to immediately return that value,
- *              indicating failure.
- *
  * Return:      Non-negative on success, negative on failure
  *
  * Programmer:  Gaute Hope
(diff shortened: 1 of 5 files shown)


derobins:
The failed test was due to the mirror VFD test not being able to connect to a socket. We see this test failure on GitHub from time to time. Re-running the failed tests fixes the problem so we think it's a GitHub issue.

mkitti:
Note that we should also update the doxygen example as in https://github.com/HDFGroup/hdf5/pull/2380

src/H5D.c
```diff
@@ -1143,18 +1143,18 @@ H5Dget_chunk_info_by_coord(hid_t dset_id, const hsize_t *offset, unsigned *filte
...
  *      H5D_chunk_iter_op_t parameters:
- *          hsize_t *offset;        IN/OUT: Array of starting logical coordinates of chunk.
- *          uint32_t filter_mask;   IN: Filter mask of chunk.
- *          haddr_t addr;           IN: Offset in file of chunk data.
- *          uint32_t nbytes;        IN: Size in number of bytes of chunk data in file.
- *          void *op_data;          IN/OUT: Pointer to any user-defined data
- *                                  associated with the operation.
+ *          hsize_t *offset;        IN/OUT: Logical position of the chunk’s first element in units of dataset
+ *                                          elements
+ *          unsigned filter_mask;   IN: Bitmask indicating the filters used when the chunk was written haddr_t
```
jhendersonHDF: I don't know if clang-format caused this, but looks like the `haddr_t` should be on the following line? as well as `size` missing from the `hsize_t` below.

jhendersonHDF (approved):
Everything looks good to me here; just minor nitpicks

mkitti:
Could we have some determination on https://github.com/HDFGroup/hdf5/pull/2380 ? I would prefer to keep the documentation in sync with the code if possible.

jhendersonHDF:
> Could we have some determination on #2380 ? I would prefer to keep the documentation in sync with the code if possible.

That's probably more of a question for @derobins or @byrnHDF, but I'll note that to address that first warning the example should probably use the `PRIuHSIZE` macro in [H5public.h](https://github.com/HDFGroup/hdf5/blob/develop/src/H5public.h#L307) to print out the hsize_t value rather than %ld

src/H5D.c
mkitti: Yes, but we should also fix this in develop.

I'm questioning the need for this documentation here since `H5D_chunk_iter_op_t` is documented in H5Dpublic.h. As far as I can tell the documentation here does not actually appear in doxygen. Why are we documenting this in more than one place?

mkitti: Could we make a decision on whether to fix this as indicated or remove this redundant, unused documentation of `H5D_chunk_iter_op_t` in deference to the existing correct documentation in H5Dpublic.h?

mkitti: I'm going to remove the documentation of `H5D_chunk_iter_op_t` from here also make a PR to remove it from the `develop` branch. Having the redundant documentation is just going to lead to misdocumentation.

mkitti: Fixed in [9f252f1](https://github.com/HDFGroup/hdf5/pull/2329/commits/9f252f193a64ab380db149fb89be7213c8296710) by removing the redundant documentation

byrnHDF (approved):

mkitti:
I plan on pulling the Doxygen example correction from https://github.com/HDFGroup/hdf5/pull/2380/files#diff-a8a358e3ee19fd11ac351047b6fd4d25f291254c986970da70fbbebb69d0ce39 into this pull request, without the changes to cmake.

mkitti:
I cherry picked ef3bed6 into this branch for consistency. Minimal pull request to develop is #2519 .

PR #2334, opened 2022-12-21
Jira issue OESS-337: Create test for H5VLconnector_info_to_str

raylu-hdf:
Added a test case for H5VLconnector_info_to_str, H5VLconnector_str_to_info, and H5VLconnector_free_info.  The test may need to change after the possible changes of the parameters of the API functions.


test/vol.c
```diff
@@ -558,6 +561,68 @@ reg_opt_datatype_get(void H5_ATTR_UNUSED *obj, H5VL_datatype_get_args_t *args, h
...
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+fake_vol_info_to_str(const void *info, char **str)
+{
+    herr_t ret_value = SUCCEED; /* Return value */
+
+    *str = (char *)malloc(8);
+
+    sprintf(*str, "%d", *((const int *)info));
```
jhendersonHDF: The GitHub action is complaining here because 8 bytes isn't enough to print out all ints. Something like INT_MIN would overflow the buffer. Since this is just a test, the simplest way to fix this would probably be to convert to:

`snprintf(*str, 8, "%d", *((const int *)info));`

If you choose to do that, I would add a check like:

```C
const int val = *((const int *)info);

if (val != 7) {
    print error message;
    return FAIL;
}
```

just to make sure a future change won't have that number being cut off if it's changed to something larger than 7 characters when printed. I would probably also move the 7 value to a macro but that's just a nitpick.

raylu-hdf: @jhendersonHDF Please have a look again.  Thanks.

test/vol.c
```diff
@@ -558,6 +561,75 @@ reg_opt_datatype_get(void H5_ATTR_UNUSED *obj, H5VL_datatype_get_args_t *args, h
...
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+fake_vol_info_to_str(const void *info, char **str)
+{
+    herr_t    ret_value = SUCCEED; /* Return value */
+    const int val       = *(const int *)info;
+
+    /* Verify the info is correct before continuing */
+    if (val != INT_MAX) {
```
jhendersonHDF, line 581: This new check is just going to fail since INT_MAX isn't what's being passed in here. The changes failed right away for the Ubuntu action, though I don't know why that isn't showing up for this PR.

raylu-hdf: Fixed.  I thought I changed the info value to INT_MAX.

jhendersonHDF (approved):

test/vol.c
```diff
@@ -558,6 +561,75 @@ reg_opt_datatype_get(void H5_ATTR_UNUSED *obj, H5VL_datatype_get_args_t *args, h
...
+        HDprintf("The value of info (%d) is incorrect\n", val);
+        return FAIL;
+    }
+
+    *str = (char *)malloc(16);
+
+    HDsnprintf(*str, 16, "%d", *((const int *)info));
```
bmribler: What is 16?  If you must use numerical value, some comment would be helpful, I think...

jhendersonHDF: In this case it's an arbitrary size picked for the VOL connector message, but it's probably not a terrible idea to make sure the malloc and snprintf calls use the same value, even if it's just test code. @raylu-hdf can you update this to use some macro for the magic number here?

jhendersonHDF (changes requested):
Looks fine, but we might as well update the magic number Binh-Minh mentioned.

test/vol.c
raylu-hdf: Done.

jhendersonHDF (approved):

mattjala (approved):

PR #2349, opened 2022-12-22
Update testh5cc.sh.in for new major version 1.14.

lrknox:
Fixes test failure in make installcheck for autotools builds.
Not needed for HDF5 1.14.


byrnHDF (approved):
Should add this to the confluence page with the other major changes.

derobins (approved):

PR #2350, opened 2022-12-22
Correct VOL connector env string parsing issue

qkoziol:
Allow spaces (and tabs) in VOL connector info string from HDF5_VOL_CONNECTOR environment variable.


qkoziol:
Tabs aren't really important, but spaces are.

qkoziol:
The fix for the mirror VFD test should allow it to run concurrently in "make check" invocations from different builds, possibly cutting down on intermittent test failures.

byrnHDF (approved):

qkoziol:
> The fix for the mirror VFD test should allow it to run concurrently in "make check" invocations from different builds, possibly cutting down on intermittent test failures.

Looks like this change was merged back in April and my merge history is slightly confused.  So, this just contains the VOL connector info parsing update.

jhendersonHDF (approved):

PR #2351, opened 2022-12-22
Clean up develop release docs

derobins:


lrknox (approved):

byrnHDF (approved):

byrnHDF (approved):

derobins:
Fixed in #2614 (slightly different scheme than used here)

PR #2360, opened 2022-12-23
HL GIF tools changes

derobins:
* Fixes the broken Autotools option
* Removes the "build HL tools" option

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -874,37 +874,6 @@ else
   AC_MSG_RESULT([no])
 fi
 
-AC_MSG_CHECKING([if the high-level tools are enabled])
-AC_ARG_ENABLE([hltools],
-     [AS_HELP_STRING([--enable-hltools],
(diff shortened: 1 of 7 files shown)


configure.ac
```diff
@@ -874,37 +874,6 @@ else
...
-                     [default=yes]
-                     ])],
-     [HDF5_HL_TOOLS=$enableval])
-
-if test "X${HDF5_HL}" = "Xyes" -a "X-$HDF5_HL_TOOLS" = "X-yes"; then
- AC_MSG_RESULT([yes])
- HL_TOOLS="tools"
-else
-  AC_MSG_RESULT([no])
-fi
```
derobins, line 890: This option was added to disable the GIF tools,. Now that we have a specific option for that, there's no reason to have a separate option for the HL tools.

lrknox (approved):

PR #2362, opened 2022-12-24
1.14.0: Brings the HL GIF tools configure changes from develop

derobins:


derobins:
I had to regenerate build files, so a lot of uninteresting changes :(

lrknox (approved):

PR #2364, opened 2022-12-27
Adds missing Doxygen for H5ESpublic.h

derobins:


src/H5ESpublic.h
```diff
@@ -297,8 +308,58 @@ H5_DLL herr_t H5ESget_err_count(hid_t es_id, size_t *num_errs);
...
+/**
+ * \ingroup H5ES
+ *
+ * \brief Convenience routine to free an array of H5ES_err_info_t structs
+ *
+ * \param[in] num_err_info The number of elements in \p err_info array
+ * \param[in] err_info Array of structures
+ * \returns \herr_t
+ *
+ * \since 1.13.0
```
lrknox, line 320: I'm not sure how this works, but will "since 1.13.0" still do what it is supposed to do in view of 1.13.x not being real releases"?  grep -r "since 1." * at the top level of the HDF5 code shows a few instances of 1.13.0, but all the other version numbers after 1.x are even:  0,  6, 8, 10, 12.

derobins: There's a LOT of 1.13.x API calls. I'll go through the library after this is merged and replace them with 1.14.0.

derobins: Alternatively, I could do that in this PR and make it a "pan-library Doxygen" PR

brtnfld: H5Dpublic.h: * \since 1.13.0
H5Dpublic.h: * \since 1.13.3
H5Dpublic.h: * \since 1.13.3
H5ESmodule.h: * listed in the table below were added to HDF5 library version 1.13.0 and later. The async version
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5ESpublic.h: * \since 1.13.0
H5Epublic.h: * \since 1.13.0
H5Ppublic.h: * \since 1.13.0
H5Ppublic.h: * \since 1.13.0
H5VLmodule.h: * VOL users should target HDF5 1.13.X, which is currently under development. The 1.13.X releases are
H5VLmodule.h: * 1.13.X release branches and we do not guarantee binary compatibility (”unstable” does NOT mean buggy).
H5VLmodule.h: * buffers obtained from the HDF5 library. This feature should be disabled by default in HDF5 1.13.X.
H5VLmodule.h: * In HDF5 1.13.0, a way to determine support for optional calls has been added.
H5VLpublic.h: * \since 1.13.0
H5public.h: * \since 1.13.0
H5public.h: * \since 1.13.0

brtnfld (approved):
LGTM

src/H5ESpublic.h
derobins: I'm going to do this in a separate PR. Should be up shortly after this one is merged.

PR #2382, opened 2023-01-01
Convert some H5MM calls to standard C equivalents

derobins:
Similar to #1929 but converts a smaller subset of calls. I've left the ones that paper over implementation-defined behavior concerning NULL parameters.


derobins:
I'm debugging the MacOS long double issue...

byrnHDF (approved):

bmribler (approved):

src/H5Tconv.c
```diff
@@ -887,7 +887,17 @@ done:
...
+         *                                                                                                   \
+         * The typed pointer arrives via a void pointer, which may have any alignement.                      \
+         * We then cast it to a pointer to a type that is assumed to be aligned, which                       \
+         * is undefined behavior (section 6.3.2.3 paragraph 7 of the C99 standard).                          \
+         * In the past this hasn't caused many problems, but in some cases (e.g.                             \
+         * converting long doubles on macOS), an optimizing compiler might do the                            \
+         * wrong thing (in the macOS case, the conversion uses SSE, which has stricter                       \
+         * requirements about alignment).                                                                    \
+         */                                                                                                  \
+        H5MM_memcpy(&src_aligned, (uint8_t *)src, sizeof(ST));                                               \
```
derobins: These comments essentially duplicate the one in H5MMprivate.h, but I'm leaving these here since we've historically had frequent long double conversion bugs.

qkoziol (changes requested):
I think this is going to cause trouble, by exposing the library to many strict-aliasing issues.   Previously, the H5MM_memcpy() code was telling the compiler that the pointers were void *, so it couldn't "know" the alignment of them.  This is likely compensating for many strict-aliasing issues in the library, as the datatype conversion code shows.  More thorough testing is probably necessary before we can be confident of removing H5MM_memcpy (in this way).

derobins:
> I think this is going to cause trouble, by exposing the library to many strict-aliasing issues. Previously, the H5MM_memcpy() code was telling the compiler that the pointers were void *, so it couldn't "know" the alignment of them. This is likely compensating for many strict-aliasing issues in the library, as the datatype conversion code shows. More thorough testing is probably necessary before we can be confident of removing H5MM_memcpy (in this way).

H5MM_memcpy was added by me in 2019 to look for overlap in the src and dst buffer. The library has happily existed for most of its life without it. Claiming that we have to have it because the library is so poorly coded that we can't exist without it seems unreasonable. The only reason we even tripped over this bug is that we have a test that deliberately mis-aligns memory in a way that is both difficult to accomplish via normal programming practices and unwise.

I strongly disagree with making memcpy slow in all cases on the off chance that a programmer is going to do something dumb that we need to insulate them from. I also don't want the library to be full of HDF5-specific hacks that serve little practical purpose. HDF5 is a C library. People who can't make proper use of pointers should be writing their code in Python or Java.

qkoziol:
> > I think this is going to cause trouble, by exposing the library to many strict-aliasing issues. Previously, the H5MM_memcpy() code was telling the compiler that the pointers were void *, so it couldn't "know" the alignment of them. This is likely compensating for many strict-aliasing issues in the library, as the datatype conversion code shows. More thorough testing is probably necessary before we can be confident of removing H5MM_memcpy (in this way).
> 
> H5MM_memcpy was added by me in 2019 to look for overlap in the src and dst buffer. The library has happily existed for most of its life without it. Claiming that we have to have it because the library is so poorly coded that we can't exist without it seems unreasonable. The only reason we even tripped over this bug is that we have a test that deliberately mis-aligns memory in a way that is both difficult to accomplish via normal programming practices and unwise.
> 
> I strongly disagree with making memcpy slow in all cases on the off chance that a programmer is going to do something dumb that we need to insulate them from. I also don't want the library to be full of HDF5-specific hacks that serve little practical purpose. HDF5 is a C library. People who can't make proper use of pointers should be writing their code in Python or Java.

I agree on all points!   I am suggesting going ahead with making the change, with more testing, preferably before merging.

qkoziol:
If you could update the PR with just the H5MM change (now that the other PR https://github.com/HDFGroup/hdf5/pull/3038#pullrequestreview-1459461770 is up), I can download this patch and try it on my Apple Silicon laptop and on some Graviton instances at AWS.

derobins:
> I agree on all points! I am suggesting going ahead with making the change, with more testing, preferably before merging.

I'm going to create a discussion item so we can come back to this. I think there are few things that need a global overhaul and this is one of them (another is integer type usage).

hyoklee:
> If you could update the PR with just the H5MM change (now that the other PR [#3038 (review)](https://github.com/HDFGroup/hdf5/pull/3038#pullrequestreview-1459461770) is up), I can download this patch and try it on my Apple Silicon laptop and on some Graviton instances at AWS.

@qkoziol rocks! I'm really curious about Graviton result.

src/H5Olayout.c
```diff
@@ -582,9 +582,15 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
...
-                        HGOTO_ERROR(H5E_OHDR, H5E_CANTALLOC, NULL, "unable to allocate heap block");
+                    if (tmp_hsize > 0) {
+                        if (NULL == (mesg->storage.u.virt.list = (H5O_storage_virtual_ent_t *)H5MM_calloc(
+                                         (size_t)tmp_hsize * sizeof(H5O_storage_virtual_ent_t))))
+                            HGOTO_ERROR(H5E_OHDR, H5E_CANTALLOC, NULL, "unable to allocate heap block");
+                    }
+                    else {
+                        /* Avoid zero-size allocation */
+                        mesg->storage.u.virt.list = NULL;
+                    }
```
derobins, line 593: This fixes a possible zero-size allocation warning

jhendersonHDF (approved):

qkoziol (approved):

PR #2383, opened 2023-01-03
Cleans up Java JNI warnings and add java to -Werror checks

derobins:


derobins:
Fixes several types of issues:

* Unused variables and parameters
* Not duplicating const strings that will be tortured with strtok
* Casts and checks for signed / unsigned integer type issues
* Paper over badness in multi VFD constants

java/src/jni/h5Constants.c
```diff
@@ -1658,16 +1658,24 @@ Java_hdf_hdf5lib_HDF5Constants_H5FD_1MEM_1DEFAULT_1GHEAP_1SIZE(JNIEnv *env, jcla
 {
     return (hsize_t)(3 * (HADDR_MAX / (H5FD_MEM_NTYPES - 1)));
 }
+H5_GCC_CLANG_DIAG_OFF("sign-conversion")
 JNIEXPORT jlong JNICALL
 Java_hdf_hdf5lib_HDF5Constants_H5FD_1MEM_1DEFAULT_1LHEAP_1SIZE(JNIEnv *env, jclass cls)
 {
+    /* XXX: BADNESS! - This value cannot fit into a jlong!
```
byrnHDF, line 1665: These are used only by one java test: testH5P_fapl_muti_defaults. I suggest we delete the test and all the java constants that start with "H5FD_MEM_DEFAULT_"

byrnHDF: If you do not want to remove the public constants then just set them to a value of -1.

java/src/jni/h5pFAPLImp.c
```diff
@@ -1254,6 +1254,9 @@ Java_hdf_hdf5lib_H5_H5Pget_1cache(JNIEnv *env, jclass clss, jlong plist, jintArr
                        "H5Pget_cache: nbytesArray array not pinned");
 
     {
+        if (NULL == nbytesArray)
```
byrnHDF: This should be above the bracket, immediately after above if statement.

java/src/jni/h5pDAPLImp.c
```diff
@@ -82,6 +82,9 @@ Java_hdf_hdf5lib_H5_H5Pget_1chunk_1cache(JNIEnv *env, jclass clss, jlong dapl, j
                        "H5Pget_chunk_cache: nbytesArray array not pinned");
 
     {
+        if (NULL == nbytesArray)
```
byrnHDF: Move above the bracket

java/src/jni/h5pDAPLImp.c
```diff
@@ -80,11 +80,10 @@ Java_hdf_hdf5lib_H5_H5Pget_1chunk_1cache(JNIEnv *env, jclass clss, jlong dapl, j
     if (NULL != rdcc_nbytes)
         PIN_LONG_ARRAY(ENVONLY, rdcc_nbytes, nbytesArray, &isCopy,
                        "H5Pget_chunk_cache: nbytesArray array not pinned");
+    if (NULL == nbytesArray)
+        H5_NULL_ARGUMENT_ERROR(ENVONLY, "nbytesArray should not be NULL after pinning");
```
jhendersonHDF, line 84: The above code is only pinning when the input arguments aren't NULL, so won't this throw an error if NULL is passed in for `rdcc_nbytes`?

java/src/jni/h5pFAPLImp.c
```diff
@@ -1252,11 +1252,10 @@ Java_hdf_hdf5lib_H5_H5Pget_1cache(JNIEnv *env, jclass clss, jlong plist, jintArr
     if (NULL != rdcc_nbytes)
         PIN_LONG_ARRAY(ENVONLY, rdcc_nbytes, nbytesArray, &isCopy,
                        "H5Pget_cache: nbytesArray array not pinned");
+    if (NULL == nbytesArray)
+        H5_NULL_ARGUMENT_ERROR(ENVONLY, "nbytesArray should not be NULL after pinning");
```
jhendersonHDF, line 1256: Same comment here about an error being thrown for this

java/src/jni/h5pDAPLImp.c
derobins: @byrnHDF  asked me to move it here. I don't normally do much work in the Java wrappers. Which is the correct behavior?

jhendersonHDF: In this case it looks like we probably need a block starting at `if (NULL != rdcc_nbytes)` and have the check for `NULL == nbytesArray` inside that block. Someone can call H5.H5Pget_chunk_cache with NULL pointer arguments so that only the non-NULL pointer arguments passed in get populated. Since one could pass a NULL `rdcc_nbytes`, the check for `nbytesArray` being NULL after the PIN call should only happen if we actually tried to pin it.

jhendersonHDF: However, there are other issues here in that the below code seems to assume `nbytesArray` is non-NULL and dereferences it.

jhendersonHDF: I guess adding this check is better than the current situation that if NULL comes in for `rdcc_nbytes`, the code would crash the JVM. The code here probably still needs to be revised at some point though.

jhendersonHDF (approved):

java/src/jni/h5pDAPLImp.c
byrnHDF: So the three jlongArray buffers must exit in order to return information, so the three checks should have the else like:
`    if (NULL != rdcc_w0)
        PIN_DOUBLE_ARRAY(ENVONLY, rdcc_w0, w0Array, &isCopy, "H5Pget_chunk_cache: rdcc_w0 array not pinned");
    else
        H5_NULL_ARGUMENT_ERROR(ENVONLY, "rdcc_w0 should not be NULL ");
    if (NULL != rdcc_nslots)
        PIN_LONG_ARRAY(ENVONLY, rdcc_nslots, rdcc_nslotsArray, &isCopy,
                       "H5Pget_chunk_cache: rdcc_nslots array not pinned");
    else
        H5_NULL_ARGUMENT_ERROR(ENVONLY, "rdcc_nslots should not be NULL ");
    if (NULL != rdcc_nbytes)
        PIN_LONG_ARRAY(ENVONLY, rdcc_nbytes, nbytesArray, &isCopy,
                       "H5Pget_chunk_cache: nbytesArray array not pinned");
    else
        H5_NULL_ARGUMENT_ERROR(ENVONLY, "rdcc_nbytes should not be NULL");
Member
`

byrnHDF:
No problem with the changes, just note that the Compound section of translate functions need to be fixed/finished.

java/src/jni/h5pDAPLImp.c
jhendersonHDF: In the C API, `rdcc_nslots`, `rdcc_nbytes` and `rdcc_w0` are optional. If we don't want them to be optional in Java, then yes we should throw an error if any are NULL.

byrnHDF: So that is why it doesn't throw error. But why is it checking just one?

byrnHDF: What should be done then is that the assignments after the call and before done is only done if the array is not NULL;
`        if (rdcc_nslotsArray!=NULL)
              *rdcc_nslotsArray = (jlong)rdcc_nslots_t;
        if (nbytesArray!=NULL)
              *nbytesArray      = (jlong)nbytes_t;
`

java/src/jni/h5pFAPLImp.c
byrnHDF: Same issue - don't assign if the array is NULL.

java/src/jni/h5pDAPLImp.c
jhendersonHDF: Yes I think the code needs some revising depending on what we want the semantics of the Java API to be

derobins:
> No problem with the changes, just note that the Compound section of translate functions need to be fixed/finished.

Can you approve then, so I can merge this?

PR #2386, opened 2023-01-03
HDFFV-10368 (OESS-319): Making two useful macros (H5L_EXT_FLAGS_ALL and H5L_EXT_VERSION) public

raylu-hdf:
I moved two useful macros (H5L_EXT_FLAGS_ALL and H5L_EXT_VERSION) from H5Lpkg.h to H5Lpublic.h for public usage. The test case in vol.c only envokes H5Lcreate_external using the passthru VOL.  To test if the macros are public, one can use them in the link creation of the passthru VOL (H5VL_pass_through_link_create), such as printing out their values.  It's kind of manual test.  But for such a simple change, testing it once may be good enough.


test/vol.c
```diff
@@ -2364,6 +2364,79 @@ test_wrap_register(void)
...
+ * Function:    test_external_link
+ *
+ * Purpose:     Tests VOL using H5Lcreate_external
+ *
+ * Return:      SUCCEED/FAIL
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+test_external_link(void)
```
jhendersonHDF: We already have tests for external links and we don't need to involve the passthrough connector at all here. Rather than this test, we need a test that uses H5Lget_val to get the external link buffer from the external link and then compares that buffer to the expected external link buffer format.

The external link version info should be in the first byte of the external link buffer returned by H5Lget_val, so that should be obtained similar to https://github.com/HDFGroup/hdf5/blob/develop/src/H5Lexternal.c#L130, which will use the H5L_EXT_VERSION macro for comparison. The link flags should also be in the first byte, so it can be obtained similar to https://github.com/HDFGroup/hdf5/blob/develop/src/H5Lexternal.c#L132, which will use the other H5L_EXT_FLAGS_ALL macro.

You can also obtain the link flags by calling http://portal.hdfgroup.org/display/HDF5/H5L_UNPACK_ELINK_VAL on the link value buffer; either way should be fine.

raylu-hdf: @jhendersonHDF Please have a look at the changes and let me know what you think.  Thanks.

test/links.c
```diff
@@ -13716,6 +13716,107 @@ external_link_with_committed_datatype(hid_t fapl, hbool_t new_format)
...
+ */
+static int
+external_link_public_macros(hid_t fapl, hbool_t new_format)
+{
+    hid_t       fid = -1;               /* File ID */
+    H5L_info2_t linfo;                  /* Link information */
+    char        objname[NAME_BUF_SIZE]; /* Object name */
+    char        filename1[NAME_BUF_SIZE];
+    char        filename2[NAME_BUF_SIZE];
+    unsigned   *flags = NULL; /* External link flags, packed as a bitmap */
```
jhendersonHDF: This needs to be a plain `unsigned flags = NULL;` so that it can be passed to H5Lunpack_elink_val as `&flags`.

raylu-hdf: Fixed.

raylu-hdf:
@derobins @jhendersonHDF 

I moved the two macros from H5Lpublic.h to H5Ldevelop.h.  Please have a look.  Thanks.

qkoziol (approved):

raylu-hdf:
@byrnHDF: I reran the test.  It passed this time.  Please have a look at it.

byrnHDF (approved):

jhendersonHDF (approved):

PR #2391, opened 2023-01-05
1.12: Fixes broken CI

derobins:


derobins:
Ignore the changes in H5private.h - I just needed something that will trigger the CI to run. I don't think it picks up changes to main.yml.

derobins:
Huh - looks like the failing CI is a false positive from the changes that were made previously.

byrnHDF (approved):

PR #2397, opened 2023-01-06
Determine if native connector is terminal

qkoziol:
Use H5VLobject_is_native to determine if the native connector is the terminal connector.


derobins (approved):

byrnHDF (changes requested):
Of course that changes a public API, but we don't publish the tools API, so it should be in a release note.

derobins:
I don't consider the tools API a public API. It's like the test API.

qkoziol:
That's confusing - if it's unpublished, why is a release note required?

byrnHDF (approved):
Would prefer a release note.

PR #2407, opened 2023-01-17
Include shlwapi.h explicitly on Windows

### LINKED ISSUE #2396: src\H5FDs3comms.c(944,13): : error : call to undeclared function 'StrStrIA'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
ONEapi Intel C Compiler on windows cannot find StrStrIA(A, B)

30>    start = HDstrcasestr(headerresponse, "\r\nContent-Length: ");
30>            ^
30>hdfsrc\src/H5win32defs.h(84,31): note: expanded from macro 'HDstrcasestr'
30>#define HDstrcasestr(A, B)    StrStrIA(A, B)

mkitti:
This fixes builds on mingw32 and other compilers.

I suspect this will fix https://github.com/HDFGroup/hdf5/issues/2396 , #2408

### DIFF
--- a/config/cmake/ConfigureChecks.cmake
+++ b/config/cmake/ConfigureChecks.cmake
@@ -131,6 +131,8 @@ CHECK_INCLUDE_FILE_CONCAT ("netdb.h"         ${HDF_PREFIX}_HAVE_NETDB_H)
 CHECK_INCLUDE_FILE_CONCAT ("arpa/inet.h"     ${HDF_PREFIX}_HAVE_ARPA_INET_H)
 if (WINDOWS)
   CHECK_INCLUDE_FILE_CONCAT ("shlwapi.h"         ${HDF_PREFIX}_HAVE_SHLWAPI_H)
+  # Checking for StrStrIA in the library is not relaible for mingw32 to stdcall
+  set (LINK_LIBS ${LINK_LIBS} "shlwapi")
 endif ()
 
 ## Check for non-standard extension quadmath.h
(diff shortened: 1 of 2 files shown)


mkitti:
xref: https://learn.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-strstria

jhendersonHDF (approved):

mkitti:
<strike>Hmm, this alone may not be sufficient. `shlwapi` may also needed to be added to `LINK_LIB` when the header is detected. Testing again.</strike>

edit: [d300feb](https://github.com/HDFGroup/hdf5/pull/2407/commits/d300febacda44b5f81d7d882c6f45b56c0424c97) is sufficient, I just forgot to apply the patch.

mkitti:
During configuration this still occurs in the mingw32 environment.

```
-- Looking for StrStrIA in shlwapi;m;ws2_32;wsock32
-- Looking for StrStrIA in shlwapi;m;ws2_32;wsock32 - not found
```

This is due to name mangling that occurs in 32-bit windows with stdcall:
```
$ nm /mingw64/lib/libshlwapi.a  | grep StrStrIA
0000000000000000 I __imp_StrStrIA
0000000000000000 T StrStrIA

$ nm /mingw32/lib/libshlwapi.a  | grep StrStrIA
00000000 I __imp__StrStrIA@8
00000000 T _StrStrIA@8
```

hmaarrfk:
Do you think it cleanly applies to 1.14.0? 
If so I'm going to test it in https://github.com/conda-forge/hdf5-feedstock/pull/188 if that is ok.

hmaarrfk:
I can confirm that this fixes things for the "windows" compiler, at least in terms of the segfault.

derobins (approved):

hmaarrfk:
didn't you want to fix the `mingw32` configuration?

mkitti:
> didn't you want to fix the `mingw32` configuration?

This addresses the issues for Windows-based compilation in general actually. Was there something else that needed to be addressed here?

hmaarrfk:
No. I'm just glad I got to test the fix before you merged anything.

emmenlau:
I think this PR opened a problem, because `shlwapi` is now a transitive dependency of hdf5, but this is not exposed via the cmake properties. Linking hdf5 on Windows leads to errors that can be resolved by linking `shlwapi`. But in cmake world, this dependency should be exposed from the cmake target.

So, since this commit, the following is no longer sufficient to link hdf5 for me:
```
  find_package(HDF5 MODULE REQUIRED COMPONENTS C)

  target_include_directories(libdeps SYSTEM INTERFACE ${HDF5_INCLUDE_DIRS})
  target_link_libraries(libdeps INTERFACE ${HDF5_C_LIBRARIES}) 
  target_compile_definitions(libdeps INTERFACE ${HDF5_DEFINITIONS})
```

mkitti:
@emmenlau I think the problem would be easier to track if to created a new issue.

PR #2409, opened 2023-01-18
Apply szip LINK_COMP_LIBS CmakeFilters.cmake patch from msys2 

mkitti:
Originally by @Biswa96

See
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-hdf5/hdf5-fix-find-szip.patch


byrnHDF (approved):

Biswa96:
Warning: A valid patch in msys2 may not mean it can be applied upstream. Please check if the change is compatible with the expected behavior in all use cases.

Also, I guess that the condition in the patch would be better if it was written like this `if (ZLIB_FOUND) ... else() ... endif()`.

CMakeFilters.cmake
```diff
@@ -59,9 +59,9 @@ if (HDF5_ENABLE_Z_LIB_SUPPORT)
       find_package (ZLIB NAMES ${ZLIB_PACKAGE_NAME}${HDF_PACKAGE_EXT} COMPONENTS static shared)
       if (NOT ZLIB_FOUND)
         find_package (ZLIB) # Legacy find
-        if (ZLIB_FOUND)
-          set (LINK_COMP_LIBS ${LINK_COMP_LIBS} ${ZLIB_LIBRARIES})
-        endif ()
+      endif ()
+      if (ZLIB_FOUND)
+        set (LINK_COMP_LIBS ${LINK_COMP_LIBS} ${ZLIB_LIBRARIES})
```
byrnHDF, line 64: This line should just be added to the if block that is next. (line 67)

byrnHDF (changes requested):
Looking further at this, there is another problem - once pass this section and ignoring the External build section, there is another set LINK_COMP_LIBS command.

CMakeFilters.cmake
byrnHDF: Look at line 94.  I think this set (LINK_COMP_LIBS line should replace the one at line 94 as that is the end of any ZLIB_FOUND checking and ZLIB_LIBRARIES is set correctly no matter which path was taken.

byrnHDF:
I will incorporate this into a new PR that will add FetchContent options that was implemented in hdf4 CMake.

mkitti:
I believe this was fixed by https://github.com/HDFGroup/hdf5/pull/2492

PR #2419, opened 2023-01-26
Correct TEST_REFERENCE string for ph5example.

lrknox:


examples/CMakeTests.cmake
```diff
@@ -93,8 +93,8 @@ if (H5_HAVE_PARALLEL AND HDF5_TEST_PARALLEL AND NOT WIN32)
         -D "TEST_ARGS:STRING="
         -D "TEST_EXPECT=0"
         -D "TEST_OUTPUT=ph5example.out"
-        -D "TEST_REFERENCE:STRING=PHDF5 tests finished with no errors"
-        -D "TEST_FILTER:STRING=PHDF5 tests finished with no errors"
+        -D "TEST_REFERENCE:STRING=PHDF5 example finished with no errors"
+        #-D "TEST_FILTER:STRING=PHDF5 tests finished with no errors"
```
lrknox, line 97: I'm not sure what this TEST_FILTER line does (suspect it does nothing), but it's commented out in 1.10 and 1.12 and is not in develop, so I commented it out here to match 1.10 and 1.12.

byrnHDF: Well it should be example too, but the reference one is all that is needed.

byrnHDF (approved):

jhendersonHDF (approved):

derobins (approved):

PR #2424, opened 2023-01-30
Update hdf5 1 8 after HDF5 1.8.23 release

lrknox:


byrnHDF:
why is  
config/cmake/scripts/CTestScript.cmake
empty?

lrknox:
> why is config/cmake/scripts/CTestScript.cmake empty?

The change was only for permissions.  The file isn't empty, but the code change is.

byrnHDF (approved):

PR #2425, opened 2023-01-31
corrected/added note about closing creation property list identifier

brtnfld:


src/H5Gpublic.h
```diff
@@ -242,11 +242,11 @@ H5_DLL hid_t  H5Gopen_async(hid_t loc_id, const char *name, hid_t gapl_id, hid_t
  *          property list associated with the group specified by \p group_id.
  *
  *          The creation property list identifier should be released with
- *          H5Gclose() to prevent resource leaks.
+ *          H5Pclose() to prevent resource leaks.
  *
  * \since 1.8.0
  *
- * \see H5Gcreate2(), H5Gclose()
+ * \see H5Gcreate2(), H5Pclose()
```
brtnfld: I don't think having the "\see ..." line for all these H5G functions is necessary, especially if the "see" API is already mentioned in the details.

byrnHDF (approved):

src/H5Gpublic.h
brtnfld: Plus, for some reason, H5Gcreate is the go-to routine when it can also be H5Gopen H5Gclose has nothing to do with most of the APIs.

brtnfld: I removed \see that did not seem relevant.

lrknox (approved):

PR #2443, opened 2023-02-10
1.14: Update release_docs HISTORY scheme

derobins:
We will no longer maintain HISTORY files for other maintenance branches, so those have been removed. Adds a README.md explaining the purpose and procedures of HDF5 HISTORY. Also adds a newsletter template and updates the RELEASE.txt note part of CONTRIBUTING.md.

Also cleans out RELEASE.txt post-HDF5-1.14.0


doc/release_txt_notes.md
```diff
@@ -0,0 +1,107 @@
...
+
+Generally, a release note must be written for every change that is made to the
+code for which users might see a change in the way the software works. In other
+words, if a user might see a difference in the way the software works, a note
+should be written. By code we mean the text that will be compiled into one of
+the company's software products. The code includes configuration changes and
+changes to tools users might work with to configure and build our software.
+
+A release note does not need to be written for changes to the code that users
+will not see.
```
lrknox, line 31: Is this intended to mean "A release note does not need to be written for changes to the code that will have no effect on the way the software works"?

release_docs/README.md
```diff
@@ -0,0 +1,102 @@
...
+This file contains the changes that were made to a maintenance branch since
+it split off from `develop`. It will also be found in the `develop` branch
+when experimental releases have been created.
+
+### RELEASE.txt
+
+This is the changelog for the current version of the library.
+
+For a MAJOR release (or in `develop`) this files lists all the changes since the
+last major version. For a MINOR release (or in a maintenance branch), this file
```
lrknox, line 29: There could be some ambiguity to "since the last major version".  Would "since the initial release of the previous major version" be correct and more precise?  Also it should be "this file lists" and I don't think the spell checker will catch the extra 's' in "files".  Probably "last release" below could also be "previous release"?
The examples below do make it pretty clear.

doc/release_txt_notes.md
derobins: I feel like what matters is the user experience. Changing a data structure under the hood would be an effect on the way the software works, but would not be visible to the user, and thus not worthy of a RELEASE.txt note.

byrnHDF (approved):

PR #2444, opened 2023-02-11
1.10: Bring main.yml GitHub workflow over from develop

derobins:
Adds many builds, particularly default API checks, an expanded matrix, and parallel (build, no test). Also sets the target branches to just hdf5_1_10.

Does NOT bring over -Werror builds since hdf5_1_10 still has many warnings. Also does not bring over the no deprecated symbols check, since that uses the `default` option, which we did not bring over yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   push:
   pull_request:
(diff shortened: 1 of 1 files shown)


derobins:
Closing - I'll cherry pick over the individual commits.

hyoklee (commented):
LGTM

PR #2384, opened 2023-01-03
Adds parallel -Werror checks

derobins:

PR #2457, opened 2023-02-11
Fix parallel warnings in H5Dmpio.c

derobins:
* Mark a parameter as unused when not using a special debug define
* Check for a chunk_entry NULL pointer after using HASH_FIND

These should be the last parallel warnings so we can start building parallel with -Werror.

### DIFF
--- a/src/H5Dmpio.c
+++ b/src/H5Dmpio.c
@@ -1404,7 +1404,11 @@ done:
  *-------------------------------------------------------------------------
  */
 static herr_t
+#ifdef H5Dmpio_DEBUG
 H5D__link_piece_collective_io(H5D_io_info_t *io_info, int mpi_rank)
+#else
+H5D__link_piece_collective_io(H5D_io_info_t *io_info, int H5_ATTR_UNUSED mpi_rank)
+#endif
 {
     MPI_Datatype  chunk_final_mtype; /* Final memory MPI datatype for all chunks with selection */
     hbool_t       chunk_final_mtype_is_derived = FALSE;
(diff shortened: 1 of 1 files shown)


derobins:
When this goes in, I can update and merge #2384

src/H5Dmpio.c
```diff
@@ -4697,8 +4701,8 @@ H5D__mpio_collective_filtered_chunk_update(H5D_filtered_collective_io_info_t *ch
 
             /* Find the chunk entry according to its chunk index */
             HASH_FIND(hh, chunk_hash_table, &chunk_idx, sizeof(hsize_t), chunk_entry);
-            HDassert(chunk_entry);
-            HDassert(mpi_rank == chunk_entry->new_owner);
+            if (chunk_entry == NULL || mpi_rank != chunk_entry->new_owner)
```
jhendersonHDF: I feel like the new_owner portion of the logic here should probably still be an assertion since it's an internal logic issue.

derobins: I'm not a fan of asserts. They are crude and I don't like the way they cause the process to abort. I also don't want people to waste time arguing over whether something warrants a library error or an assert. I feel like asserts should just be used for things that would cause an abort anyway, like dereferencing a NULL pointer or dividing by zero.

jhendersonHDF: The problem I see is that if the new_owner check fails, it's likely going to abort or hang anyway because internal library state is broken. 

At the very least though, I think the new_owner check should at least be in a separate error check to distinguish between not finding a chunk entry and the chunk entry's state being messed up. Giving an error to the user that the state is messed up won't really be helpful to them in any way from their perspective, but I guess it at least won't be misleading.

derobins: I moved the new_owner check into a separate error check

jhendersonHDF (approved):

Issue #2470, opened 2023-02-13
Missing or misplaced entries in API reference tables


jhendersonHDF:
Several entries appear to be missing or incorrect from the Property List API reference tables rooted at https://docs.hdfgroup.org/hdf5/develop/group___h5_p.html.

- File Creation Property Lists
    - H5Pset_file_image/H5Pget_file_image should be listed under FAPL table, rather than FCPL

- File Access Property Lists
[...]

PR #2472, opened 2023-02-15
Missing or misplaced entries in API reference tables #2470

byrnHDF:


jhendersonHDF (approved):

doxygen/examples/tables/propertyLists.dox
```diff
@@ -179,11 +171,19 @@ creation property list.</td>
 <td>Sets/retrieves metadata cache and raw data chunk cache parameters.</td>
```
bmribler, line 171: Shouldn't this be Sets/Retrieves?  Same for Sets/gets.
Also, some are Set/get instead of Sets/gets.

byrnHDF: Maybe, but this is how it was typically used before in the tables in the portal.

bmribler: Didn't mean correct, right?

derobins (approved):

PR #2480, opened 2023-02-16
Fix issue with collective metadata writes of global heap data

### LINKED ISSUE #2433: [BUG] NetCDF fails its parallel tests with HDF 1.14.0
**Describe the bug**
The NetCDF parallel tests fail due to an assertion in H5F_shared_vector_write (H55Fioc).

**Expected behavior**
NetCDF Parallel tests should not fail.

 - HDF5 version (if building from a maintenance branch, please include the commit hash)
 - OS and version
 - Compiler and version
 - Build system (e.g., CMake, Autotools) and version
 - Any configure options you specified
 - MPI library and version (parallel HDF5)

tested  HDF5 13.1 (passes), 13.2(fails), 14.0(fails) and develop (fails).
Linux, GCC 11, MPICH 4, Autotools, NetCDF 4.9.1.

nc_create_par will fail assertion:
../../src/H5Fio.c:397: H5F_shared_vector_write: Assertion `types[i] != H5FD_MEM_GHEAP' failed.

All the NetCDF tests pass when the assert is removed.

jhendersonHDF:
Fixes #2433

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -207,6 +207,21 @@ Bug Fixes since HDF5-1.13.3 release
 ===================================
     Library
     -------
+    - Fixed an issue with collective metadata writes of global heap data
+
+      New test failures in parallel netCDF started occurring with debug
+      builds of HDF5 due to an assertion failure and this was reported in
+      GitHub issue #2433. The assertion failure began happening after the
+      collective metadata write pathway in the library was updated to use
+      vector I/O so that parallel-enabled HDF5 Virtual File Drivers (other
+      than the existing MPI I/O VFD) can support collective metadata writes.
+
+      The assertion failure was fixed by updating collective metadata writes
+      to treat global heap metadata as raw data, as done elsewhere in the
+      library. 
+
+      (JTH - 2023/02/16, GH #2433)
+
     - Seg fault on file close
 
       h5debug fails at file close with core dump on a file that has an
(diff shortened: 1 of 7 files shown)


brtnfld (approved):
LGTM, I verified that it fixes the NetCDF failures 🔧

qkoziol (changes requested):
lgtm, but should have a new regression test also

jhendersonHDF:
> lgtm, but should have a new regression test also

I agree we need a reproducer, though it might be difficult to come up with one. Any idea how a parallel netCDF program is causing a write of global heap metadata? My only initial thought was reference and/or VL types, but we don't support access to the global heap in parallel.

qkoziol:
VL or ref datatype in an attribute?

jhendersonHDF:
> VL or ref datatype in an attribute?

Ah ok.. I suppose that wouldn't really be problematic parallel access since all ranks would be doing the same thing.

qkoziol (approved):
Great!

lrknox (approved):

release_docs/RELEASE.txt
```diff
@@ -207,6 +207,21 @@ Bug Fixes since HDF5-1.13.3 release
 ===================================
     Library
     -------
+    - Fixed an issue with collective metadata writes of global heap data
```
wkliao, line 210: Shouldn't this statement be under "Bug Fixes since HDF5-1.14.0 release", instead of 1.13.3?

PR #2495, opened 2023-02-23
Subfiling VFD - add option to specify config file prefix

jhendersonHDF:
Add an environment variable that the Subfiling VFD checks for so an application can place a config file in a different directory than where the subfiles go.

### DIFF
--- a/src/H5FDsubfiling/H5FDioc.c
+++ b/src/H5FDsubfiling/H5FDioc.c
@@ -1455,6 +1455,7 @@ H5FD__ioc_del(const char *name, hid_t fapl)
     if (mpi_rank == 0) {
         int64_t read_n_subfiles = 0;
         int32_t n_subfiles      = 0;
+        char   *prefix_env      = NULL;
         int     num_digits      = 0;
 
         if (HDstat(name, &st) < 0)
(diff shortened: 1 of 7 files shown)


brtnfld (approved):
lgtm

src/H5FDsubfiling/H5FDioc.c
```diff
@@ -1470,9 +1471,13 @@ H5FD__ioc_del(const char *name, hid_t fapl)
             H5_SUBFILING_GOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL,
                                     "can't allocate config file name buffer");
 
+        /* Check if a prefix has been set for the configuration file name */
+        prefix_env = HDgetenv(H5FD_SUBFILING_CONFIG_FILE_PREFIX);
+
         /* TODO: No support for subfile directory prefix currently */
```
lrknox, line 1477: Should this comment "No support for subfile directory prefix currently" be removed?

jhendersonHDF: In this case the comment refers to a separate prefix the user can specify that deals with files other than what's covered for this PR.

derobins (approved):

PR #2500, opened 2023-02-24
Use LINK_COMP_LIBS instead of STATIC_LIBRARY for depends

byrnHDF:


CMakeFilters.cmake
```diff
@@ -96,14 +96,15 @@ if (HDF5_ENABLE_Z_LIB_SUPPORT)
...
       endif ()
-    endif ()
-    if (ZLIB_FOUND)
-      set (H5_HAVE_FILTER_DEFLATE 1)
-      set (H5_HAVE_ZLIB_H 1)
-      set (H5_HAVE_LIBZ 1)
-      set (H5_ZLIB_HEADER "zlib.h")
-      set (ZLIB_INCLUDE_DIR_GEN ${ZLIB_INCLUDE_DIR})
-      set (ZLIB_INCLUDE_DIRS ${ZLIB_INCLUDE_DIRS} ${ZLIB_INCLUDE_DIR})
+      if (ZLIB_FOUND)
```
lrknox, line 99: should this be 
      else ()
instead of 
      if (NOT ZLIB_FOUND)
      endif ()
      if (ZLIB_FOUND)
        ...
?

lrknox: Nevermind, I guess it's a sequence

lrknox (approved):

PR #2511, opened 2023-03-05
Fix an errant bracket that confused formatter

derobins:
An extra { in an ifdef'd-out block caused bad indentation in H5C.c

### DIFF
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -3566,575 +3566,575 @@ done:
 
 #if H5C_DO_EXTREME_SANITY_CHECKS
     if ((H5C_validate_protected_entry_list(cache_ptr) < 0) ||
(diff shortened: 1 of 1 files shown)


src/H5C.c
```diff
@@ -3566,575 +3566,575 @@ H5C_unprotect(H5F_t *f, haddr_t addr, void *thing, unsigned flags)
 
 #if H5C_DO_EXTREME_SANITY_CHECKS
     if ((H5C_validate_protected_entry_list(cache_ptr) < 0) ||
-        (H5C_validate_pinned_entry_list(cache_ptr) < 0) || (H5C_validate_lru_list(cache_ptr) < 0)) {
+        (H5C_validate_pinned_entry_list(cache_ptr) < 0) || (H5C_validate_lru_list(cache_ptr) < 0))
```
derobins, line 3569: Removing this bracket is the only change I made. Everything else was from the formatter.

byrnHDF (approved):

PR #2380, opened 2022-12-30
Fix H5Dchunk_iter doxygen example, add doxygen example to tests

mkitti:
This updates the doxygen example for `H5Dchunk_iter` following changes made in #2074 .

The primary goal is to change the `chunk_cb` function to the following signature:
```
chunk_cb(const hsize_t *offset, unsigned filter_mask, haddr_t addr, hsize_t size, void *op_data)
```

To ensure the doxygen examples can be compiled, the cmake infrastructure is extended in a limited way to now cover `doxygen/examples/H5D_examples.c`. I anticipate another pull request may extend this further to cover other files in `doxygen/examples`.

PR #2519, opened 2023-03-08
Fix H5Dchunk_iter doxygen example, cherry-pick of ef3bed6

mkitti:
This is a minimalist version of #2380 that only modifies the doxygen documentation example for the develop branch.

The main reason this exists is so that this minimal change can be ported to the 1.12 and 1.10 branches.

This pull request should affect documentation only. Unlike #2380 it makes no change to the build process.

### DIFF
--- a/doxygen/examples/H5D_examples.c
+++ b/doxygen/examples/H5D_examples.c
@@ -7,10 +7,10 @@
 
 //! <!-- [H5Dchunk_iter_cb] -->
 int
-chunk_cb(const hsize_t *offset, uint32_t filter_mask, haddr_t addr, uint32_t nbytes, void *op_data)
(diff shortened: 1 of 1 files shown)


mkitti:
codespell is weird... why is it detecting these issues now?

lrknox:
> 

We also saw new spelling errors today in a merge  to HDF4 master, which had been there since well before we started using codespell or github for that matter.  Why did it miss them before?  Maybe codespell was upgraded?

Some of them are actual errors to be corrected, others are strings in code that will need to be ignored.  I see the same issues in multiple branches - I'll create  PR for develop to fix them, and then propagate it to the PRs for the other branches.

byrnHDF (approved):

lrknox (approved):

PR #2521, opened 2023-03-08
Fix new codespell issues

mkitti:


lrknox:
I see you've already created a PR to fix.

I think we need to add ndefined to the list to be skipped in .github/workflows/codespell.yaml.  It is part of [Uu]ndefined in cmake sanitzer code, so it is correct.   There is one more misspelled word in doxygen/examples/Fileformat.html:33 - knowledgable needs e before able.

lrknox:
@mkitti, do you want to add ndefined to the ignore_words_list in .github/workflows.codespell.yaml?

mkitti:
Hmm... `ndefined` is likely to be an actual misspelling at some point. Perhaps we should just skip checking `./config/sanitizer/sanitizers.cmake`

lrknox:
> Hmm... `ndefined` is likely to be an actual misspelling at some point. Perhaps we should just skip checking `./config/sanitizer/sanitizers.cmake`

Sure.  You'll add it to the "skip:" list in codespell.yaml?

lrknox (approved):

byrnHDF (approved):

PR #2524, opened 2023-03-09
Addressed various Doxygen grammar issues.

brtnfld:
Doxygen Doc. Changes.


byrnHDF (approved):

src/H5Cpkg.h
```diff
@@ -4398,7 +4398,7 @@ typedef struct H5C_tag_info_t {
  * order, we scan the index repeatedly, once for each flush dependency
  * height in increasing order.
  *
- * This operation is complicated by the fact that entries other the the
+ * This operation is complicated by the fact that entries other the
```
lrknox:
```suggestion
 * This operation is complicated by the fact that entries other than the
```

src/H5Lpublic.h
```diff
@@ -209,7 +209,7 @@ H5_DLL herr_t H5Lmove(hid_t src_loc, const char *src_name, hid_t dst_loc, const
  *
  *          H5Lcopy() retains the creation time and the target of the original
  *          link. However, since the link may be renamed, the character
- *          encoding is that specified in \p lcpl_id rather than that of the
+ *          encoding is  specified in \p lcpl_id rather than that of the
```
lrknox:
```suggestion
 *          encoding is specified in \p lcpl_id rather than in that of the
```

src/H5MF.c
```diff
@@ -3236,7 +3236,7 @@ H5MF_settle_meta_data_fsm(H5F_t *f, hbool_t *fsm_settled)
          *           multi file drivers, as the self referential free space
          *           manager header and section info can be stored in up to
          *           two different files -- requiring that up to two EOA's
-         *           be stored in the the free space managers super block
+         *           be stored in the free space managers super block
```
lrknox:
```suggestion
         *           be stored in the free space manager's superblock
```

src/H5Opublic.h
```diff
@@ -790,9 +790,9 @@ H5_DLL herr_t H5Olink(hid_t obj_id, hid_t new_loc_id, const char *new_name, hid_
  *          H5Odecr_refcount() should be used.
  *
  *          An object’s reference count is the number of hard links in the
- *          file that point to that object. See the “Programming Model”
+ *          file that points to that object. See the “Programming Model”
```
lrknox: I think this should be 'point':  it's the hard links that point, not the file.

src/H5Spublic.h
```diff
@@ -293,7 +293,7 @@ H5_DLL hid_t H5Screate(H5S_class_t type);
  *       H5Sselect_elements() selects array elements in a dataspace. These
  *       functions are used for subsetting. H5Sselect_none() removes all
  *       selections from a dataspace and is used in Parallel HDF5 when a process
- *       does not have or need to write data.
+ *       does not have or needs to write data.
```
lrknox: I don't know the rule for this, but the subject is process and the verbs with number agreement are does have or does need.  I don't think process does not needs is correct.

src/H5Tenum.c
```diff
@@ -333,7 +333,7 @@ H5Tenum_nameof(hid_t type, const void *value, char *name /*out*/, size_t size)
 /*-------------------------------------------------------------------------
  * Function:	H5T__enum_nameof
  *
- * Purpose:	Finds the symbol name that corresponds the the specified
+ * Purpose:	Finds the symbol name that corresponds the specified
```
lrknox:
```suggestion
 * Purpose:	Finds the symbol name that corresponds to the specified
```

src/H5Tenum.c
```diff
@@ -470,7 +470,7 @@ H5Tenum_valueof(hid_t type, const char *name, void *value /*out*/)
 /*-------------------------------------------------------------------------
  * Function:	H5T__enum_valueof
  *
- * Purpose:	Finds the value that corresponds the the specified symbol
+ * Purpose:	Finds the value that corresponds the specified symbol
```
lrknox:
```suggestion
 * Purpose:	Finds the value that corresponds to the specified symbol
```

src/H5Tpublic.h
```diff
@@ -1138,7 +1138,7 @@ H5_DLL herr_t H5Tlock(hid_t type_id);
  * \ingroup H5T
  *
  * \brief Commits a transient datatype, linking it into the file and creating
- *        a new committed datatype
+ *        a newly committed datatype
```
lrknox: Judging from only this sentence, it seems that a committed datatype that did not yet exist will be created.  If true, I think new is better.   I also think that after it's created, newly-committed makes sense in **both** lines 1161 and 1164.  Seeing newly in line 1238, though, the whole process being described should be considered.

lrknox (changes requested):

src/H5Tpublic.h
brtnfld: fixed

src/H5Spublic.h
brtnfld: fixed.

lrknox (approved):

PR #2537, opened 2023-03-13
Many -Wextra-semi-stmt fixes

seanm:


qkoziol:
What does a "function-like macro" look like?

seanm:
>What does a "function-like macro" look like?

https://gcc.gnu.org/onlinedocs/cpp/Function-like-Macros.html

seanm:
@qkoziol the idea is that if it's a function-wannabe, it should end with a semi-colon, like a real function call.

src/H5FD.c
```diff
@@ -929,10 +929,9 @@ H5FD_cmp(const H5FD_t *f1, const H5FD_t *f2)
 {
     int ret_value = -1; /* Return value */
 
-    FUNC_ENTER_NOAPI_NOERR; /* return value is arbitrary */
+    FUNC_ENTER_NOAPI_NOERR /* return value is arbitrary */
 
-    if ((!f1 || !f1->cls) && (!f2 || !f2->cls))
-        HGOTO_DONE(0)
+        if ((!f1 || !f1->cls) && (!f2 || !f2->cls)) HGOTO_DONE(0)
```
derobins: I'd like to find something that makes both clang and clang-format happy. This is an example where the formatting is worse when an otherwise useless semicolon is removed.

src/H5Dfarray.c
```diff
@@ -1435,7 +1435,7 @@ H5D__farray_idx_copy_setup(const H5D_chk_idx_info_t *idx_info_src, const H5D_chk
             HGOTO_ERROR(H5E_DATASET, H5E_CANTOPENOBJ, FAIL, "can't open fixed array")
 
     /* Set copied metadata tag */
-    H5_BEGIN_TAG(H5AC__COPIED_TAG);
+    H5_BEGIN_TAG(H5AC__COPIED_TAG)
```
derobins, line 1438: I don't like macros that float in space like they aren't C statements. We should fix up the macros so they are C-like when used.

derobins:
I'm excited to fix some long-standing code deficiencies in the library, but I feel like this PR needs a few more tweaks, particularly in  the FUNC_ENTER/LEAVE macros, where this PR sometimes removes semicolons from function-like macros. Those macros can be a bit of a snarl - perhaps @qkoziol and I can help with that.

derobins:
I've converted this to a draft while we discuss.

derobins:
What does clang have to say about HGOTO_ERROR and the like?

src/H5Dfarray.c
seanm: >I don't like macros that float in space like they aren't C statements.

Meaning you'd prefer them with no indentation?

>We should fix up the macros so they are C-like when used.

C-like?  You mean function-like?  I don't see how.  Recall the macro is:

```C
#define H5_BEGIN_TAG(tag)                    \
    {                                        \
        haddr_t prv_tag = HADDR_UNDEF;       \
        H5AC_tag(tag, &prv_tag);
```

It's not function-like at all, it declares a variable that survives in scope.  As it's not function-like, it's better without the trailing semi, IMHO.

seanm:
>...particularly in the FUNC_ENTER/LEAVE macros, where this PR sometimes removes semicolons from function-like macros.

Again, they don't seem function-like to me... ex:

```C
#define FUNC_LEAVE_NOAPI_VOID                                                                                \
    ;                                                                                                        \
    } /*end scope from end of FUNC_ENTER*/                                                                   \
    H5_POP_FUNC                                                                                              \
    return;                                                                                                  \
    } /*end scope from beginning of FUNC_ENTER*/
```

You certainly can't wrap them in a do-while idiom.

seanm:
>What does clang have to say about HGOTO_ERROR and the like?

I maybe didn't get to that one... I thought fixing the `-Wextra-semi-stmt` warnings would be a short diversion, but after descending the rabbit hole quite far, I thought better to share what I started before continuing.

src/H5FD.c
seanm: Search current `develop` branch. This was the only `FUNC_ENTER_NOAPI_NOERR` *with* a semi colon. The other 100 or so *already* don't have a semi colon.  Given that this macro is:

```C
#define FUNC_ENTER_NOAPI_NOERR                         \
    {                                                  \
        FUNC_ENTER_COMMON_NOERR(!H5_IS_API(__func__)); \
        H5_PUSH_FUNC                                   \
        {
```

clang-format's choice isn't even so weird as this macro opens two new scopes with `{`.

seanm:
@derobins @qkoziol friendly ping...

Do you understand now what I mean about some of those macros not being function-like?

How can we get this PR moved forward?  Would it help to break it into smaller pieces?

derobins:
> @derobins @qkoziol friendly ping...
> 
> Do you understand now what I mean about some of those macros not being function-like?
> 
> How can we get this PR moved forward? Would it help to break it into smaller pieces?

Can you come to one of the HDF5 Working Group meetings on Thursdays at 10am Central? I can send you a Teams invite if you shoot me an email.

seanm:
@derobins yes, I could this coming Thursday.

derobins:
> @derobins yes, I could this coming Thursday.

Shoot me an email at derobins at hdfgroup dot org and I'll send you a Teams invite.

hl/src/H5LT.c
```diff
@@ -1898,7 +1898,6 @@ H5LTget_attribute_ndims(hid_t loc_id, const char *obj_name, const char *attr_nam
     /* End access to the attribute */
     if (H5Aclose(attr_id))
         goto out;
-    ;
```
derobins, line 1901: Weird that all these lone semicolons exist. I wonder where they all came from?

seanm: I suspect someone fat-fingered `goto out;;` then clang-format put the second "statement" on a separate line.

src/H5FD.c
derobins: I'm sure we figure something out that will make clang-format not due weird stuff

seanm: [`StatementMacros`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) I believe is the way.  I'll give it a try...

seanm: Commit added. You already had `StatementMacros` actually, I added items and alphabetized the list.

derobins (approved):

seanm:
thanks @derobins . After this is merged, now that I know you're ok with the concept, I'll try to fix the remaining such warnings (there are tonnes).

derobins:
> thanks @derobins . After this is merged, now that I know you're ok with the concept, I'll try to fix the remaining such warnings (there are tonnes).

Thanks so much! We really appreciate outside contributions!

qkoziol (approved):
Nice!  Thanks for taking the time to improve the code quality!

PR #2550, opened 2023-03-13
Rename clang-format fix action

derobins:
The clang-format fix and check actions have the same name. This also makes some small changes to the action's text fields.

### DIFF
--- a/.github/workflows/clang-format-fix.yml
+++ b/.github/workflows/clang-format-fix.yml
@@ -1,15 +1,15 @@
-name: clang-format Check
+name: clang-format Commit Changes
 on: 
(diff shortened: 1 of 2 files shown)


.github/workflows/clang-format-fix.yml
```diff
@@ -1,15 +1,15 @@
-name: clang-format Check
+name: clang-format Commit Changes
```
derobins, line 1: This is the most important line. Both clang-format-fix and clang-format-check should not use the same name.

lrknox (approved):

bmribler (approved):

PR #2623, opened 2023-03-23
Minor fixes for CMake code and install docs

byrnHDF:


lrknox (approved):

brtnfld:
Why not just use "XX" or something for the 1.# versions, similar to using the 1.15."X" now, instead of continually updating these values in the docs?

byrnHDF:
> Why not just use "XX" or something for the 1.# versions, similar to using the 1.15."X" now, instead of continually updating these values in the docs?
The major number only should be updated once. And it should be done automatically anyway.

lrknox:
> > Why not just use "XX" or something for the 1.# versions, similar to using the 1.15."X" now, instead of continually updating these values in the docs?
> > The major number only should be updated once. And it should be done automatically anyway.

Doing it automatically could probably be done by h5vers.  If you want to do it that way, I suggest an issue and a PR in the near future to do it, but to go ahead with the changes in this PR for now.

derobins:
Agree with Scot's comment, but this is okay for now.

derobins (approved):

PR #2626, opened 2023-03-23
Fix a memory corruption issue in H5S__point_project_simple

jhendersonHDF:


src/H5Spoint.c
```diff
@@ -2319,7 +2319,7 @@ H5S__point_project_simple(const H5S_t *base_space, H5S_t *new_space, hsize_t *of
             /* Copy over the point's coordinates */
             HDmemset(new_node->pnt, 0, sizeof(hsize_t) * rank_diff);
             H5MM_memcpy(&new_node->pnt[rank_diff], base_node->pnt,
-                        (new_space->extent.rank * sizeof(hsize_t)));
+                        (base_space->extent.rank * sizeof(hsize_t)));
```
jhendersonHDF, line 2322: If you expand the code above, you can see that this block assumes `new_space->extent.rank > base_space->extent.rank`. This copying code allocates a `new_space->extent.rank`-sized array of `hsize_t` for the point coordinates, then offsets into the `pnt` array by `rank_diff` which is calculated by `new_space->extent.rank - base_space->extent.rank`. However, it previously tried copying a `new_space->extent.rank` amount of points into the array AFTER offsetting into the array, whereas there would only be room left for a `base_space->extent.rank` amount of points. This issue cropped up with a Developer build that disables free lists, so this issue has likely been around for a long time but went unnoticed because the free lists were allocating enough room that we could happily write over.

qkoziol: Nice catch!

qkoziol (approved):

lrknox (approved):

PR #2628, opened 2023-03-23
Synchronize HDF5 tests with VOL tests

jhendersonHDF:


testpar/t_bigio.c
```diff
@@ -26,7 +26,7 @@ const char *FILENAME[3] = {"bigio_test.h5", "single_rank_independent_io.h5", NUL
 #define DATASET4             "DSET4"
 #define DXFER_COLLECTIVE_IO  0x1 /* Collective IO*/
 #define DXFER_INDEPENDENT_IO 0x2 /* Independent IO collectively */
-#define DXFER_BIGCOUNT       (1 < 29)
```
jhendersonHDF, line 29: This seems to have been overlooked, but I'm also a bit hesitant about this change because it seems the original intention was to have the bigio test creating datasets with 500M elements in one of their two dimensions, which quickly locks my machine up when testing. In the VOL tests I've adjusted this value to 1310720 to be a little more reasonable, but I think the test needs reviewing and/or rewriting as a whole.

lrknox (approved):

brtnfld:
Is the return state of H5Fdelete not an issue in these tests?

jhendersonHDF:
> Is the return state of H5Fdelete not an issue in these tests?

I suppose it is, though we never checked the status from HDremove before either and that could fail. If a test fails to run correctly and doesn't create a test file, an H5Fdelete/HDremove call would fail to remove the file which doesn't exist, but that's sort of a spurious error anyway.

brtnfld:
But now that would not be detected because of the begin_try

testpar/t_bigio.c
brtnfld: Agreed, we probably need a reasonable value depending on the resources available.

brtnfld (approved):
lgtm

jhendersonHDF:
> But now that would not be detected because of the begin_try

I can remove the begin_try, but then we also will probably see spurious failures in the tests. I'm fine with either way, as long as we remember those H5Fdelete failures are sometimes expected.

testpar/t_bigio.c
jhendersonHDF: Yes, @lrknox we should probably hold off on merging this until we decide a reasonable value for this test. It takes a long long time after the above is correctly fixed.

brtnfld:
> I can remove the begin_try, but then we also will probably see spurious failures in the tests. I'm fine with either way, as long as we remember those H5Fdelete failures are sometimes expected.

I would leave the begin_trys since we don't want sporadic H5Fdelete failures.

PR #2632, opened 2023-03-24
Avoid suppressing error output for non-tentative file opens

jhendersonHDF:


qkoziol (changes requested):
The goal is good here, but the changes here and code in the existing next 'if' block should be integrated into one coherent set of code, instead of overlapping / duplicating code.

jhendersonHDF:
@qkoziol Does that look a little better? I'm not super fond of the organization but it was the only way I could think of since we can't really conditionally use H5E_BEGIN/END_TRY without duplicating code.

qkoziol (approved):
This reads better and the code is clearer now, thanks for tuning it up!

lrknox (approved):

PR #2633, opened 2023-03-24
Minor fixes for Subfiling VFD tests

jhendersonHDF:


brtnfld (approved):
LGTM

testpar/t_subfiling_vfd.c
```diff
@@ -1971,9 +2032,13 @@ int
 main(int argc, char **argv)
 {
     unsigned seed;
-    char    *env_value = NULL;
-    int      required  = MPI_THREAD_MULTIPLE;
-    int      provided  = 0;
+    hbool_t  unset_stripe_size_env      = FALSE;
```
mattjala: Nitpick: I think the names of these booleans are ambiguous - they could mean either that this variable has already been unset, or that it needs to be unset later. The latter case is true here, so these should be renamed to "must_unset_stripe_size_env" or similar.

jhendersonHDF: Agreed. Let me make that quick change.

mattjala (approved):

PR #2638, opened 2023-03-27
Undo revert 2615

lrknox:


config/cmake/scripts/HDF5config.cmake
```diff
@@ -38,7 +38,7 @@ cmake_minimum_required (VERSION 3.12)
 ##############################################################################
 
 set (CTEST_SOURCE_VERSION "1.10.10")
-set (CTEST_SOURCE_VERSEXT "-2")
+set (CTEST_SOURCE_VERSEXT "-1")
```
byrnHDF: Do you want to do this change?

byrnHDF (approved):

config/cmake/scripts/HDF5config.cmake
lrknox: Probably not.  The 1_10 branch will be going to 1.10.11-1, and the 1_10_10 branch is 1.10.10-3, but I'll probably take this change out.

lrknox:
```suggestion
set (CTEST_SOURCE_VERSEXT "-2")
```

byrnHDF: Don't forget the other files that have that version issue.

byrnHDF (approved):

PR #2639, opened 2023-03-27
Move vol-tests back into library

glennsong09:


qkoziol:
What's the Kitware System library used for?

jhendersonHDF:
> What's the Kitware System library used for?

This is used for a test driver program for running the VOL tests. For the DAOS VOL it was basically used to bring up the server and wait for that to come up before running tests, then shut down the server afterward. It's still unclear whether we'll keep it, but this is a feature branch in-progress for now.

jhendersonHDF (approved):

PR #2656, opened 2023-04-02
Fix memory leaks and aborts in H5O EFL decode

### LINKED ISSUE #2605: [BUG] detected memory leaks in H5O__efl_decode hdf5/src/H5Oefl.c
detected memory leaks in H5O__efl_decode hdf5/src/H5Oefl.c

Built from git commit hash 0553fb7
OS: Ubuntu 20.04
Compiler: clang version 11.0

Build options:
- Shared library disabled
CC=clang
CXX=clang++
CFLAGS="-g -fno-inline -fsanitize=address"
CXXFLAGS="-g -fno-inline -fsanitize=address"
LDFLAGS="$LDFLAGS -fsanitize=address"
ASAN_OPTIONS=abort_on_error=1

Command: ./h5dump pov

POV file:
[pov.zip](https://github.com/HDFGroup/hdf5/files/11014096/pov.zip)

Stack trace:
```
==1948296==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 1 byte(s) in 1 object(s) allocated from:
#0 0x4c18c7 in calloc /home/chaitra/aflgo-top/build/llvm_tools/llvm-11.0.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:154:3
#1 0xcc2c04 in H5O__efl_decode hdf5/src/H5Oefl.c:137:37
#2 0x8738c8 in H5O_msg_read_oh hdf5/src/H5Omessage.c:513:5
#3 0x8732df in H5O_msg_read hdf5/src/H5Omessage.c:454:30
#4 0x63d32f in H5D__layout_oh_read hdf5/src/H5Dlayout.c:627:21
#5 0x623906 in H5D_open hdf5/src/H5Dint.c:1498:13
#6 0x622d37 in H5D__open_name hdf5/src/H5Dint.c:1433:25
#7 0xb81750 in H5VL__native_dataset_open hdf5/src/H5VLnative_dataset.c:318:25
#8 0xb5638b in H5VL__dataset_open hdf5/src/H5VLcallback.c:1946:30
SUMMARY: AddressSanitizer: 1 byte(s) leaked in 1 allocation(s).
```

derobins:
H5O__efl_decode() would call assert() and not properly clean up memory when parsing malformed external data files messages. This change corrects that behavior and fixes #2605.

### DIFF
--- a/src/H5Oefl.c
+++ b/src/H5Oefl.c
@@ -96,19 +96,19 @@ H5O__efl_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSED
         HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, NULL, "memory allocation failed")
 
     /* Version (1 byte) */
-    if (p + 1 - 1 > p_end)
+    if ((p + 1 - 1) > p_end)
         HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, NULL, "ran off end of input buffer while decoding")
     version = *p++;
     if (version != H5O_EFL_VERSION)
         HGOTO_ERROR(H5E_OHDR, H5E_CANTLOAD, NULL, "bad version number for external file list message")
 
     /* Reserved (3 bytes) */
-    if (p + 3 - 1 > p_end)
+    if ((p + 3 - 1) > p_end)
         HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, NULL, "ran off end of input buffer while decoding")
     p += 3;
 
     /* Number of slots (2x 2 bytes) */
-    if (p + 4 - 1 > p_end)
+    if ((p + 4 - 1) > p_end)
         HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, NULL, "ran off end of input buffer while decoding")
     UINT16DECODE(p, mesg->nalloc);
     if (mesg->nalloc <= 0)
@@ -118,7 +118,7 @@ H5O__efl_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSED
         HGOTO_ERROR(H5E_OHDR, H5E_CANTLOAD, NULL, "bad number of in-use slots when parsing efl msg")
 
     /* Heap address */
-    if (p + H5F_SIZEOF_ADDR(f) - 1 > p_end)
+    if ((p + H5F_SIZEOF_ADDR(f) - 1) > p_end)
         HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, NULL, "ran off end of input buffer while decoding")
     H5F_addr_decode(f, &p, &(mesg->heap_addr));
     if (H5F_addr_defined(mesg->heap_addr) == FALSE)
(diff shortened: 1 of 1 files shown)


src/H5Oefl.c
```diff
@@ -111,64 +103,69 @@ H5O__efl_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSED
 
     /* Number of slots */
     UINT16DECODE(p, mesg->nalloc);
-    HDassert(mesg->nalloc > 0);
+    if (mesg->nalloc <= 0)
+        HGOTO_ERROR(H5E_OHDR, H5E_CANTLOAD, NULL, "bad number of allocated slots when parsing efl msg")
```
derobins, line 115: Parsing a malformed file should never result in an assert() call.

src/H5Oefl.c
```diff
@@ -111,64 +103,69 @@ H5O__efl_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSED
...
 
 done:
     if (ret_value == NULL)
-        if (mesg != NULL)
+        if (mesg != NULL) {
+            if (mesg->slot != NULL) {
+                for (size_t u = 0; u < mesg->nused; u++)
+                    H5MM_xfree(mesg->slot[u].name);
+                H5MM_xfree(mesg->slot);
+            }
```
derobins, line 182: mesg->slot and the names could be allocated but were not cleaned on errors.

src/H5Oefl.c
qkoziol: Agree

However, it's an unsigned value, so can't be <0

qkoziol (approved):

src/H5Oefl.c
derobins: I could change that to be != 0, but then that would be broken if anyone changes the type to be signed so they can have a "bad" value.

qkoziol (changes requested):
Suggest parentheses around math expressions involved in comparisons.   Memorizing the precedence of operations shouldn't be the most important thing on your mind when looking at this code.

derobins:
> Suggest parentheses around math expressions involved in comparisons. Memorizing the precedence of operations shouldn't be the most important thing on your mind when looking at this code.

I went ahead and added parens, but I honestly don't think that remembering math > logic is too much to remember. If I get some free time, I'll add a function or macro to hide all the awkward -1 stuff.

qkoziol (approved):
Cool, thanks :-)

mattjala (approved):

src/H5Oefl.c
```diff
@@ -67,108 +67,121 @@ const H5O_msg_class_t H5O_MSG_EFL[1] = {{
...
 {
-    H5O_efl_t  *mesg = NULL;
-    int         version;
-    const char *s = NULL;
-    H5HL_t     *heap;
-    size_t      u;                /* Local index variable */
-    void       *ret_value = NULL; /* Return value */
+    H5O_efl_t     *mesg = NULL;
+    int            version;
+    const uint8_t *p_end     = p + p_size - 1; /* pointer to last byte in p */
```
jhendersonHDF, line 83: Suggest possibly moving this assignment past the assert for `p` since addition on a NULL pointer is undefined. That said, we don't have a normal check for `p != NULL` so it won't be caught in release builds as is anyway if this ever happens. Hopefully the assert would catch it during development though.

jhendersonHDF (approved):

PR #2661, opened 2023-04-04
Remove -Wunused-but-set-variable warnings in tests

byrnHDF:
These three tests produced warnings on variables set but unused.
These changes were simple, but maybe others have better recommendations.

### DIFF
--- a/test/tattr.c
+++ b/test/tattr.c
@@ -7103,8 +7103,14 @@ attr_iterate_check(hid_t fid, const char *dsetname, hid_t obj_id, H5_index_t idx
             VERIFY(iter_info->visited[v], TRUE, "H5Aiterate2");
     } /* end if */
     else {
+        unsigned nvisit = 0; /* # of links visited */
+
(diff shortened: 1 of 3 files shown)


test/tattr.c
```diff
@@ -7103,13 +7103,7 @@ attr_iterate_check(hid_t fid, const char *dsetname, hid_t obj_id, H5_index_t idx
...
     } /* end if */
     else {
-        unsigned nvisit = 0; /* # of links visited */
-
         HDassert(order == H5_ITER_NATIVE);
-        for (v = 0; v < max_attrs; v++)
-            if (iter_info->visited[v] == TRUE)
-                nvisit++;
-
         VERIFY(skip, (max_attrs / 2), "H5Aiterate2");
```
bmribler: I think this for loop shouldn't be removed.  The VERIFY statement should compare nvisit against max_attrs instead.

derobins (approved):

bmribler (approved):

Issue #2601, opened 2023-03-20
[BUG] heap-buffer-overflow in H5O__ginfo_decode hdf5/src/H5Oginfo.c


cniddodi:
heap-buffer-overflow in H5O__ginfo_decode hdf5/src/H5Oginfo.c

Built from git commit hash 0553fb7
OS: Ubuntu 20.04
Compiler: clang version 11.0

Build options:
- Shared library disabled
CC=clang
CXX=clang++
CFLAGS="-g -fno-inline -fsanitize=address"
CXXFLAGS="-g -fno-inline -fsanitize=address"
LDFLAGS="$LDFLAGS -fsanitize=address"
ASAN_OPTIONS=abort_on_error=1

Command: ./h5dump pov

POV file:
[pov.zip](https://github.com/HDFGroup/hdf5/files/11013600/pov.zip)

Stack trace:
```
==1947747==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6060000055d8 at pc 0x000000cc6ada bp 0x7ffd8d283c90 sp 0x7ffd8d283c88
READ of size 1 at 0x6060000055d8 thread T0
#0 0xcc6ad9 in H5O__ginfo_decode hdf5/src/H5Oginfo.c:104:9
#1 0x8738c8 in H5O_msg_read_oh hdf5/src/H5Omessage.c:513:5
#2 0x8732df in H5O_msg_read hdf5/src/H5Omessage.c:454:30
#3 0x753261 in H5G_get_create_plist hdf5/src/H5Gint.c:1312:21
#4 0xb8b1a3 in H5VL__native_group_get hdf5/src/H5VLnative_group.c:182:48
#5 0xb62da3 in H5VL__group_get hdf5/src/H5VLcallback.c:4611:9
#6 0xb62b66 in H5VL_group_get hdf5/src/H5VLcallback.c:4642:9
#7 0x7458d0 in H5Gget_create_plist hdf5/src/H5G.c:525:9
#8 0x5017a5 in dump_group hdf5/tools/src/h5dump/h5dump_ddl.c:770:20
#9 0x4fe5d6 in dump_all_cb hdf5/tools/src/h5dump/h5dump_ddl.c:228:25
#10 0x751262 in H5G__iterate_cb hdf5/src/H5Gint.c:865:25
#11 0x760355 in H5G__node_iterate hdf5/src/H5Gnode.c:966:25
[...]

PR #2663, opened 2023-04-04
Fix invalid memory access in H5O__ginfo_decode

mattjala:
A malformed file would cause H5O__ginfo_decode to read beyond allocated memory. Fix for #2601.


src/H5Oginfo.c
```diff
@@ -68,6 +68,8 @@ const H5O_msg_class_t H5O_MSG_GINFO[1] = {{
 #define H5O_GINFO_STORE_PHASE_CHANGE   0x01
 #define H5O_GINFO_STORE_EST_ENTRY_INFO 0x02
 #define H5O_GINFO_ALL_FLAGS            (H5O_GINFO_STORE_PHASE_CHANGE | H5O_GINFO_STORE_EST_ENTRY_INFO)
+/* Check if access at this point of given size is valid */
+#define IS_BUFFER_OVERFLOW(ptr, size, buffer_end) (((ptr) + (size) - 1) > (buffer_end))
```
derobins: Adding a macro for this is a good idea, but this should be located in a central location for general use, like H5private.h.

derobins: Also, such a central macro would start with H5_

src/H5private.h
```diff
@@ -2576,3 +2576,8 @@ H5_DLL herr_t H5_buffer_dump(FILE *stream, int indent, const uint8_t *buf, const
                              size_t buf_offset, size_t buf_size);
 
 #endif /* H5private_H */
+
+/* Check if a read of size bytes starting at ptr would overflow past
+ * the last valid byte, pointed to by buffer_end .
+ */
+#define H5_IS_BUFFER_OVERFLOW(ptr, size, buffer_end) (((ptr) + (size) - 1) > (buffer_end))
```
jhendersonHDF: This macro needs to be moved so that it's at least inside the big `#ifndef H5private_h .. #endif` block. For consistency, it should probably be located somewhere next to the other macros defined in this file.

qkoziol (approved):
Nice update w/macro to clarify what's going on.

Minor error codes should be updated to reflect actual problem though.  Possibly H5_NOSPACE to H5_OVERFLOW?

src/H5Omessage.c
```diff
@@ -499,6 +499,11 @@ H5O_msg_read_oh(H5F_t *f, H5O_t *oh, unsigned type_id, void *mesg)
     type = H5O_msg_class_g[type_id]; /* map the type ID to the actual type object */
     HDassert(type);
 
+    if ((oh->nmesgs * sizeof(H5O_mesg_t)) > oh->cache_info.size) {
```
mattjala: There was an issue with how I was executing the test suite before.  This isn't correct - it throws a false positive for most HDF5 files.

lrknox (approved):

Issue #2599, opened 2023-03-20
Detected memory leaks in H5FL__malloc hdf5/src/H5FL.c


cniddodi:
detected memory leaks in H5FL__malloc hdf5/src/H5FL.c

Built from git commit hash 0553fb7
OS: Ubuntu 20.04
Compiler: clang version 11.0

Build options:
- Shared library disabled
CC=clang
CXX=clang++
CFLAGS="-g -fno-inline -fsanitize=address"
CXXFLAGS="-g -fno-inline -fsanitize=address"
LDFLAGS="$LDFLAGS -fsanitize=address"
ASAN_OPTIONS=abort_on_error=1

Command: ./h5dump pov

POV file:
[pov.zip](https://github.com/HDFGroup/hdf5/files/11013588/pov.zip)

Stack trace:
```
==1947206==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 56 byte(s) in 1 object(s) allocated from:
#0 0x4c170f in malloc /home/chaitra/aflgo-top/build/llvm_tools/llvm-11.0.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cpp:145:3
#1 0x72cebb in H5FL__malloc hdf5/src/H5FL.c:237:30
SUMMARY: AddressSanitizer: 56 byte(s) leaked in 1 allocation(s).
```

vchoi-hdfgroup:
It was fixed and merged to develop, 1.14, 1.12 and 1.10.

PR #2665, opened 2023-04-04
Fix for github issue #2599:

vchoi-hdfgroup:
As indicated in the description, memory leak is detected when running "./h5dump pov".

The problem is: when calling H5O__add_cont_msg() from H5O__chunk_deserialize(), memory is allocated for cont_msg_info->msgs.  Eventually, when the library tries to load the continuation message via H5AC_protect() in H5O_protect(), error is encountered due to illegal info in the continuation message. Due to the error, H5O_protect() exits but the memory allocated for cont_msg_info->msgs is not freed.

When we figure out how to handle fuzzed files that we didn't generate, a test needs to be added to run h5dump with the provided "pov" file.


hyoklee (approved):

qkoziol (approved):

lrknox:
Please add a release note for Bugs fixed since 1.14.0.

lrknox (approved):

PR #2666, opened 2023-04-05
CMake: Let runtime checks through stdout only

jschueller:
This avoids mixing emulators errors (on stderr) to configure results while cross-compiling with eg wine or qemu.
For fortran runtime-checks we send the results to stdout instead of stderr.


byrnHDF (changes requested):
RUN_OUTPUT_STDOUT_VARIABLE isn't available until CMake 3.25, our minimum is 3.18.

jschueller:
ok, lets use  RUN_OUTPUT_VARIABLE or RUN_OUTPUT_STDOUT_VARIABLE depending on CMAKE_VERSION

byrnHDF (approved):

lrknox (approved):

PR #2679, opened 2023-04-06
Add buffer overrun checks to H5O__layout_decode and H5O__sdspace_decode

jhendersonHDF:


src/H5private.h
```diff
@@ -325,6 +325,11 @@
 /* Raise an integer to a power of 2 */
 #define H5_EXP2(n) (1 << (n))
 
+/* Check if a read of size bytes starting at ptr would overflow past
+ * the last valid byte, pointed to by buffer_end.
+ */
+#define H5_IS_BUFFER_OVERFLOW(ptr, size, buffer_end) (((ptr) + (size)-1) > (buffer_end))
```
jhendersonHDF, line 331: Borrowing this until we determine final placement..

qkoziol (approved):
Suggest H5E_NOSPACE -> H5E_OVERFLOW

jhendersonHDF:
> Suggest H5E_NOSPACE -> H5E_OVERFLOW

"Address overflowed" seems a bit awkward to me as well to describe the issue that occurred, but I don't have any strong opinion either way. I can change it.

src/H5Olayout.c
```diff
@@ -164,29 +180,46 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
              * size in the dataset code, where we've got the dataspace
              * information available also.  - QAK 5/26/04
              */
+            if (H5_IS_BUFFER_OVERFLOW(p, (ndims * sizeof(uint32_t)), p_end))
+                HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding")
             p += ndims * 4; /* Skip over dimension sizes (32-bit quantities) */
```
mattjala: Could this '4' be replaced with a size call to match the overflow check?

jhendersonHDF: Sure. Changed.

lrknox (approved):

PR #2681, opened 2023-04-06
Update HDF5 version after 1.10.10 release.

lrknox:
Update so numbers to match 1.10.10 release so numbers. 
Update HISTORY-1_10.txt.
Clean RELEASE.txt entries.


lrknox:
RELEASE.txt from the HDF5 1.10.9 release was missing from HISTORY-1_10.txt.  I've added it as well as RELEASE.txt from the HDF5 1.10.10 release.

byrnHDF (approved):

bmribler (approved):

PR #2683, opened 2023-04-07
Subfiling VFD - check if MPI is finalized during VFD termination

jhendersonHDF:
Fixes potential use of MPI routines after MPI_Finalize if user leaves an HDF5 ID open. We have code that calls H5_term_library in response to MPI_Finalize, but there seems to be an ordering issue as it doesn't cover this case.

### DIFF
--- a/src/H5FDsubfiling/H5subfiling_common.c
+++ b/src/H5FDsubfiling/H5subfiling_common.c
@@ -338,7 +338,17 @@ done:
 static herr_t
 H5_free_subfiling_object_int(subfiling_context_t *sf_context)
 {
+    int    mpi_finalized;
+    int    mpi_code;
+    herr_t ret_value = SUCCEED;
+
     HDassert(sf_context);
+
+    if (MPI_SUCCESS != (mpi_code = MPI_Finalized(&mpi_finalized))) {
+        /* Assume MPI is finalized or worse, and try to clean up what we can */
+        H5_SUBFILING_MPI_DONE_ERROR(FAIL, "MPI_Finalized failed", mpi_code);
+        mpi_finalized = 1;
+    }
 
     sf_context->sf_context_id           = -1;
     sf_context->h5_file_id              = UINT64_MAX;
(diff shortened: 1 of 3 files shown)


qkoziol:
This is reasonable defensive programming, but these routines should be getting invoked when H5__mpi_delete_cb() is called on MPI_Finalize() (see the code in src/H5.c for how this is set up).  Why aren't they getting invoked when H5_term_library() is called?   Do they need to be added to the SHUTDOWN macro there?  (Or some other special code?)

jhendersonHDF:
> This is reasonable defensive programming, but these routines should be getting invoked when H5__mpi_delete_cb() is called on MPI_Finalize() (see the code in src/H5.c for how this is set up). Why aren't they getting invoked when H5_term_library() is called? Do they need to be added to the SHUTDOWN macro there? (Or some other special code?)

The issue with that code is that it only gets invoked if MPI has been initialized at the time that the library gets initialized. This doesn't always end up being the case due to weird things like predefined datatypes that initialize the library through assignment to a variable. For example, here's a tiny example program that causes the same issue for me:

```
#include "hdf5.h"

int
main(int argc, char **argv)
{
    hid_t fapl;
    hid_t type_id = H5T_NATIVE_INT;

    MPI_Init(&argc, &argv);

    fapl = H5Pcreate(H5P_FILE_ACCESS);
    H5Pset_mpi_params(fapl, MPI_COMM_WORLD, MPI_INFO_NULL);

    H5Fclose(H5Fcreate("open_id.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT));

    /* Leave FAPL open to trigger comm free after finalize */
    /* H5Pclose(fapl); */

    MPI_Finalize();

    return 0;

}
```

This has caused issues in the past like when we tried running the DAOS VOL connector with a c++ application that was causing HDF5 to be initialized in a manner that appeared to be serial so the connector didn't realize it was running in parallel. IIRC, we worked around the issue but it's still a present problem in the library.

mattjala (approved):

bmribler (approved):

lrknox (approved):

brtnfld (approved):

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -338,8 +338,18 @@ H5_free_subfiling_object(int64_t object_id)
...
 H5_free_subfiling_object_int(subfiling_context_t *sf_context)
 {
+    int    mpi_finalized;
+    int    mpi_code;
+    herr_t ret_value = SUCCEED;
+
     HDassert(sf_context);
 
+    if (MPI_SUCCESS != (mpi_code = MPI_Finalized(&mpi_finalized))) {
+        /* Assume MPI is finalized or worse, and try to clean up what we can */
```
hdfeos, line 348: worse than what?

hyoklee:
This fix doesn't work on Sunspot.

PR #2693, opened 2023-04-11
Cache tidy

qkoziol:
Bring style closer to library's agreed coding style


src/H5Cpkg.h
```diff
@@ -855,7 +828,7 @@ if ( ( ( ( (head_ptr) == NULL ) || ( (tail_ptr) == NULL ) ) &&             \
 #define H5C__UPDATE_STATS_FOR_HT_DELETION(cache_ptr)
 #define H5C__UPDATE_STATS_FOR_HT_SEARCH(cache_ptr, success, depth)
 #define H5C__UPDATE_STATS_FOR_INSERTION(cache_ptr, entry_ptr)
-#define H5C__UPDATE_STATS_FOR_CLEAR(cache_ptr, entry_ptr)
+#define H5C__UPDATE_STATS_FOR_CLEAR(cache_ptr, entry_ptr)                       {}
```
qkoziol, line 831: This quirk will be address in an upcoming patch

src/H5AC.c
```diff
@@ -439,34 +439,25 @@ H5AC_dest(H5F_t *f)
...
+        if (curr_logging)
+            if (H5C_log_write_destroy_cache_msg(f->shared->cache) < 0)
+                HGOTO_ERROR(H5E_CACHE, H5E_LOGGING, FAIL, "unable to emit log message")
 
         if (H5C_log_tear_down(f->shared->cache) < 0)
-
-            HGOTO_ERROR(H5E_CACHE, H5E_LOGGING, FAIL, "mdc logging tear-down failed")
-    }
+            HGOTO_ERROR(H5E_CACHE, H5E_LOGGING, FAIL, "metadata cache logging tear-down failed")
+    } /* end if */
```
derobins, line 452: I'm not going to die on this hill today, but I consider almost all "end <thing>" comments to be noise. If something is so messy that they are necessary, then that code should be refactored to extract functions, not get poorly-aligned noise comments.

derobins (approved):
Good tidying that brings the H5C coding style in line with the rest of the library

src/H5AC.c
qkoziol: Sorry, I like them, but I'm not trying to fight that battle - this slipped in on accident

lrknox (approved):

Issue #2603, opened 2023-03-20
[BUG] heap-buffer-overflow in H5O__linfo_decode hdf5/src/H5Olinfo.c


cniddodi:
heap-buffer-overflow in H5O__linfo_decode hdf5/src/H5Olinfo.c

Built from git commit hash 0553fb7
OS: Ubuntu 20.04
Compiler: clang version 11.0

Build options:
- Shared library disabled
CC=clang
CXX=clang++
CFLAGS="-g -fno-inline -fsanitize=address"
CXXFLAGS="-g -fno-inline -fsanitize=address"
LDFLAGS="$LDFLAGS -fsanitize=address"
ASAN_OPTIONS=abort_on_error=1

Command: ./h5dump pov

POV file: 
[pov.zip](https://github.com/HDFGroup/hdf5/files/11013615/pov.zip)

Stack trace:
```
==1948081==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6060000053f8 at pc 0x00000086bf45 bp 0x7ffddb4bf4d0 sp 0x7ffddb4bf4c8
READ of size 1 at 0x6060000053f8 thread T0
#0 0x86bf44 in H5O__linfo_decode hdf5/src/H5Olinfo.c:121:9
#1 0x8738c8 in H5O_msg_read_oh hdf5/src/H5Omessage.c:513:5
#2 0x8732df in H5O_msg_read hdf5/src/H5Omessage.c:454:30
#3 0x76434d in H5G__obj_get_linfo hdf5/src/H5Gobj.c:307:21
#4 0x751776 in H5G_visit hdf5/src/H5Gint.c:1221:25
#5 0xb8eeac in H5VL__native_link_specific hdf5/src/H5VLnative_link.c:375:30
#6 0xb667f3 in H5VL__link_specific hdf5/src/H5VLcallback.c:5482:22
#7 0xb665bd in H5VL_link_specific hdf5/src/H5VLcallback.c:5516:14
#8 0x7ec80d in H5Lvisit_by_name2 hdf5/src/H5L.c:1986:22
#9 0x555522 in traverse hdf5/tools/lib/h5trav.c:282:17
#10 0x55769d in h5trav_visit hdf5/tools/lib/h5trav.c:1051:9
#11 0x55357c in init_objs hdf5/tools/lib/h5tools_utils.c:628:22
#12 0x4f6c86 in table_list_add hdf5/tools/src/h5dump/h5dump.c:425:9
[...]

PR #2697, opened 2023-04-12
Fixed GH-2603, heap-buffer-overflow in H5O__linfo_decode

bmribler:
Verified with valgrind -v --tool=memcheck --leak-check=full h5dump POV-GH-2603
The several invalid reads shown originally are now gone.
The test file is POV-GH-2603.


src/H5Olinfo.c
```diff
@@ -136,20 +141,30 @@ H5O__linfo_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUS
...
+    if (linfo->track_corder) {
+        if (H5_IS_BUFFER_OVERFLOW(p, 8, p_end))
+            HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding")
         INT64DECODE(p, linfo->max_corder)
+    }
     else
         linfo->max_corder = 0;
 
+    /* Check input buffer before decoding the next two addresses */
+    if (H5_IS_BUFFER_OVERFLOW(p, 8 + 8, p_end))
```
derobins: The address checks should use H5F_sizeof_addr() to get the size of addresses in the file instead of 8.

release_docs/RELEASE.txt
```diff
@@ -147,11 +147,20 @@ Support for new platforms, languages and compilers
 ==================================================
     -
     
-Bug Fixes since HDF5-1.13.3 release
+Bug Fixes since HDF5-1.14.0 release
```
jhendersonHDF, line 150: Also is this change something we only make at release time, or was it just forgotten?

bmribler: I'm pretty sure it was a typo, i.e., forgotten.

derobins: Forgotten. Good catch.

derobins (approved):

jhendersonHDF (approved):

PR #2712, opened 2023-04-13
Check for invalid AAPL in H5Aopen

jhendersonHDF:
The AAPL isn't used in H5Aopen, but the VOL tests check for failure when passing an invalid AAPL and the library doesn't currently check the AAPL for serial builds. It only gets checked in a roundabout way for parallel builds in H5CX_set_apl when collective metadata reads aren't enabled globally for the file.

### DIFF
--- a/src/H5VLnative_attr.c
+++ b/src/H5VLnative_attr.c
@@ -135,18 +135,22 @@ done:
  *-------------------------------------------------------------------------
  */
 void *
-H5VL__native_attr_open(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_name,
-                       hid_t H5_ATTR_UNUSED aapl_id, hid_t H5_ATTR_UNUSED dxpl_id, void H5_ATTR_UNUSED **req)
+H5VL__native_attr_open(void *obj, const H5VL_loc_params_t *loc_params, const char *attr_name, hid_t aapl_id,
+                       hid_t H5_ATTR_UNUSED dxpl_id, void H5_ATTR_UNUSED **req)
 {
-    H5G_loc_t loc;         /* Object location */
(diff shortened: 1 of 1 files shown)


jhendersonHDF:
@byrnHDF Does the failure in Java for MSVC here look like another case of a dependency issue when running the tests in parallel?

derobins:
A quick test for this would be helpful to ensure it doesn't regress.

src/H5VLnative_attr.c
```diff
@@ -135,19 +135,23 @@ H5VL__native_attr_create(void *obj, const H5VL_loc_params_t *loc_params, const c
...
+    H5A_t          *attr = NULL; /* Attribute opened */
+    void           *ret_value;
 
     FUNC_ENTER_PACKAGE
 
     /* check arguments */
     if (H5G_loc_real(obj, loc_params->obj_type, &loc) < 0)
         HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, NULL, "not a file or file object")
 
+    if (NULL == (plist = (H5P_genplist_t *)H5I_object_verify(aapl_id, H5I_GENPROP_LST)))
```
qkoziol: H5P_object_verify would be a better (stronger) check.

qkoziol (changes requested):

jhendersonHDF:
> A quick test for this would be helpful to ensure it doesn't regress.

How do we want to handle the duplicated test between adding one here and the existing one in the VOL tests?

jhendersonHDF:
@qkoziol pushed those changes. Once this PR is merged I have a GitHub action for the VOL tests hopefully ready to go that regression tests this. It should fix a single failure on Ubuntu.

mattjala (approved):

derobins (approved):

PR #2720, opened 2023-04-13
Fix java depends and pdb lib install

byrnHDF:


java/examples/datasets/CMakeLists.txt
```diff
@@ -97,6 +98,9 @@ if (BUILD_TESTING AND HDF5_TEST_EXAMPLES AND HDF5_TEST_SERIAL)
               ${HDFJAVA_EXAMPLES_BINARY_DIR}/${example}.h5
       )
     endif ()
+    if (last_test)
+      set_tests_properties (JAVA_datasets-${example}-clear-objects PROPERTIES DEPENDS ${last_test})
```
jhendersonHDF, line 102: What was the ordering here that was causing problems? Looks like clear objects just would've cleared testing files if they were left behind

byrnHDF: Because the clear did not depend on anything it could run at anytime, now it should be in a depends path - I hope

byrnHDF: Better would be a fixture process - I need to think more about that.

jhendersonHDF: Ahh, if you look below around line 144 it looks like the dependency is only getting set for tests that aren't "H5Ex_D_External" due to it accidentally being inside the else block

jhendersonHDF (changes requested):
Still think we need the dependency chain added here, but the fix is in the dependency setting code that exists

java/examples/datasets/CMakeLists.txt
byrnHDF: Of course!

jhendersonHDF: I guess that brings up the question though whether the clear objects of the next loop needs to depend on the clean objects of the previous or if all those tests can run in parallel.

jhendersonHDF (approved):

derobins (approved):

PR #2734, opened 2023-04-14
[1.10 Merge] H5O__pline_decode() Make more resilient to out-of-bounds read (#2210)

jhendersonHDF:
Malformed hdf5 files may have trunkated content which does not match the expected size. When this function attempts to decode these it may read past the end of the allocated space leading to heap overflows as bounds checking is incomplete.
Make sure each element is within bounds before reading.

This fixes CVE-2019-8396 / HDFFV-10712 / github bug #2209.


jhendersonHDF:
Merging this CVE fix back to the 1.10 branch because it's useful and contains a definition for HDstrnlen which I need in other PRs.

derobins (approved):

PR #2740, opened 2023-04-14
Run parallel clang-format tasks

derobins:
This won't affect the GitHub actions since those run clang-format via python script.

On local machines, though, this dramatically speeds up running clang-format via bin/format_source (10s --> 3s on my machine).

### DIFF
--- a/bin/format_source
+++ b/bin/format_source
@@ -21,6 +21,6 @@ find . \( -type d -path ./config -prune -and -not -path ./config \) \
         -or -name H5overflow.h \
         \) \) \
     -and \( -iname *.h -or -iname *.c -or -iname *.cpp -or -iname *.hpp -or -iname *.java \) \) \
-    | xargs -P0 --max-args=1 clang-format -style=file -i -fallback-style=none
+    | xargs -P0 -n1 clang-format -style=file -i -fallback-style=none
 
 exit 0
--- a/bin/format_source
+++ b/bin/format_source
@@ -21,6 +21,6 @@ find . \( -type d -path ./config -prune -and -not -path ./config \) \
         -or -name H5overflow.h \
         \) \) \
     -and \( -iname *.h -or -iname *.c -or -iname *.cpp -or -iname *.hpp -or -iname *.java \) \) \
-    | xargs clang-format -style=file -i -fallback-style=none
+    | xargs -P0 --max-args=1 clang-format -style=file -i -fallback-style=none
 
 exit 0


bin/format_source
```diff
@@ -21,6 +21,6 @@ find . \( -type d -path ./config -prune -and -not -path ./config \) \
         -or -name H5overflow.h \
         \) \) \
     -and \( -iname *.h -or -iname *.c -or -iname *.cpp -or -iname *.hpp -or -iname *.java \) \) \
-    | xargs clang-format -style=file -i -fallback-style=none
+    | xargs -P0 --max-args=1 clang-format -style=file -i -fallback-style=none
```
qkoziol: I don't see the --max-args flag for xargs in my Mac's manpage.  What does it do?   If it's not on the Mac, is it portable?

derobins: Does MacOS have -n?

derobins: -n max-args, --max-args=max-args
              Use at most max-args arguments per command line.  Fewer than max-args arguments will be used if the size (see the
              -s option) is exceeded, unless the -x option is given, in which case xargs will exit.

       -P max-procs, --max-procs=max-procs
              Run up to max-procs processes at a time; the default is 1.  If max-procs is 0, xargs will run as  many  processes
              as possible at a time.  Use the -n option or the -L option with -P; otherwise chances are that only one exec will
              be done.  While xargs is running, you can send its process a SIGUSR1 signal to increase the number of commands to
              run  simultaneously, or a SIGUSR2 to decrease the number.  You cannot increase it above an implementation-defined
              limit (which is shown with --show-limits).  You cannot decrease it below 1.  xargs never terminates its commands;
              when asked to decrease, it merely waits for more than one existing command to terminate before starting another.

              Please  note  that  it is up to the called processes to properly manage parallel access to shared resources.  For
              example, if more than one of them tries to print to stdout, the output will be produced in an indeterminate order
              (and  very  likely  mixed  up)  unless the processes collaborate in some way to prevent this.  Using some kind of
              locking scheme is one way to prevent such problems.  In general, using a locking scheme will help ensure  correct
              output but reduce performance.  If you don't want to tolerate the performance difference, simply arrange for each
              process to produce a separate output file (or otherwise use separate resources).

derobins: The value for -n doesn't seem to matter, but if you don't have it, you get the same behavior as if you did not have -P0.

UPDATE: Ah, I think I get it now - If you don't have -n equal some small value (like 1), one process will snarf up all the arguments up to the default max.

derobins: There is no mention of --max-args not being portable:

       As  of GNU xargs version 4.2.9, the default behaviour of xargs is not to have a logical end-of-file marker.  POSIX (IEEE
       Std 1003.1, 2004 Edition) allows this.

       The -l and -i options appear in the 1997 version of the POSIX standard, but do not appear in the  2004  version  of  the
       standard.  Therefore you should use -L and -I instead, respectively.

       The -o option is an extension to the POSIX standard for better compatibility with BSD.

       The  POSIX  standard  allows implementations to have a limit on the size of arguments to the exec functions.  This limit
       could be as low as 4096 bytes including the size of the environment.  For scripts to be portable, they must not rely  on
       a larger value.  However, I know of no implementation whose actual limit is that small.  The --show-limits option can be
       used to discover the actual limits in force on the current system.

qkoziol: From the Mac's xargs manpage:

```
     -L number
             Call utility for every number non-empty lines read.  A line
             ending with a space continues to the next non-empty line.  If EOF
             is reached and fewer lines have been read than number then
             utility will be called with the available lines.  The -L and -n
             options are mutually-exclusive; the last one given will be used.

     -n number
             Set the maximum number of arguments taken from standard input for
             each invocation of utility.  An invocation of utility will use
             less than number standard input arguments if the number of bytes
             accumulated (see the -s option) exceeds the specified size or
             there are fewer than number arguments remaining for the last
             invocation of utility.  The current default value for number is
             5000.

     -o      Reopen stdin as /dev/tty in the child process before executing
             the command.  This is useful if you want xargs to run an
             interactive application.

     -P maxprocs
             Parallel mode: run at most maxprocs invocations of utility at
             once.  If maxprocs is set to 0, xargs will run as many processes
             as possible.

 ```

derobins: I'll change it to -n

qkoziol (approved):
Super!  Definitely happy to get this running faster

lrknox (approved):

PR #2742, opened 2023-04-14
Merge "Use additional CPUs" and related fixes

derobins:
* 2720 Fix java depends and pdb lib install
* 2695 Use additional CPUs in GitHub actions


byrnHDF (approved):
Why didn't this pick up #2720?  Nevermind - I was confused with a different fix

PR #2748, opened 2023-04-15
Sanitize the attribute info message decode fxn

derobins:
Adds bounds checking on the buffer in the attribute info message's decode function (H5O__ainfo_decode).

### DIFF
--- a/src/H5Oainfo.c
+++ b/src/H5Oainfo.c
@@ -13,8 +13,6 @@
 /*-------------------------------------------------------------------------
  *
  * Created:             H5Oainfo.c
- *                      Mar  6 2007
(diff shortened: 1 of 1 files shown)


derobins:
There is no RELEASE.txt note because this doesn't fix a specific GitHub issue. I will probably add a blanket statement about hardening the decode layer before release.

byrnHDF (approved):

jhendersonHDF (approved):

hyoklee:
I think this generated 2 more errors on Sunspot vol-daos testing.

https://github.com/hyoklee/vol-daos

hyoklee:
Sunspot is now back to 1 error.

PR #2749, opened 2023-04-16
Sanitize the attribute object header msg code

derobins:
Adds:
* Bounds checks on buffer access
* Better memory cleanup

### DIFF
--- a/src/H5Oattr.c
+++ b/src/H5Oattr.c
@@ -189,7 +189,7 @@ H5O__attr_decode(H5F_t *f, H5O_t *open_oh, unsigned H5_ATTR_UNUSED mesg_flags, u
      */
     if (H5_IS_BUFFER_OVERFLOW(p, name_len, p_end))
         HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
-    if (NULL == (attr->shared->name = H5MM_strndup((const char *)p, name_len)))
+    if (NULL == (attr->shared->name = H5MM_strndup((const char *)p, name_len - 1)))
         HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL, "memory allocation failed")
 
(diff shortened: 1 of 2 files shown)


test/titerate.c
```diff
@@ -1011,8 +1011,11 @@ test_corrupted_attnamelen(void)
     hbool_t        driver_is_default_compatible;
     const char    *testfile = H5_get_srcdir_filename(CORRUPTED_ATNAMELEN_FILE); /* Corrected test file name */
 
-    const char *err_message = "attribute name has different length than stored length";
-    /* the error message produced when the failure occurs */
+    /* The error message produced when the failure occurs
+     *
+     * FIXME: This is incredibly fragile!
+     */
+    const char *err_message = "ran off end of input buffer while decoding";
```
derobins, line 1018: I am not a fan of tests that are this fragile, but I'll save fixing this for another day.

src/H5Oattr.c
```diff
@@ -234,17 +268,18 @@ H5O__attr_decode(H5F_t *f, H5O_t *open_oh, unsigned H5_ATTR_UNUSED mesg_flags, u
     if ((attr->shared->data_size / dt_size) != ds_size)
         HGOTO_ERROR(H5E_RESOURCE, H5E_OVERFLOW, NULL, "data size exceeds addressable range")
 
-    /* Go get the data */
+    /* Get the data */
     if (attr->shared->data_size) {
         /* Ensure that data size doesn't exceed buffer size, in case of
-           it's being corrupted in the file */
-        if (attr->shared->data_size > p_size)
-            HGOTO_ERROR(H5E_RESOURCE, H5E_OVERFLOW, NULL, "data size exceeds buffer size")
```
derobins, line 242: This check was incorrect and would not catch cases where the data size was too big (and could run off the end of the buffer) but less than the size of the entire message.

jhendersonHDF (approved):

mattjala (approved):

PR #2755, opened 2023-04-16
Clean up H5Oefl decode function

derobins:
* Use the H5_IS_BUFFER_OVERFLOW macro
* Attempt to close local heap on errors if left open

### DIFF
--- a/src/H5Oefl.c
+++ b/src/H5Oefl.c
@@ -64,14 +64,14 @@ const H5O_msg_class_t H5O_MSG_EFL[1] = {{
 /*-------------------------------------------------------------------------
  * Function:    H5O__efl_decode
  *
- * Purpose:	Decode an external file list message and return a pointer to
- *          the message (and some other data).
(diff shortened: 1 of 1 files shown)


jhendersonHDF (approved):

src/H5Oefl.c
```diff
@@ -90,36 +85,38 @@ H5O__efl_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSED
     /* Check args */
     HDassert(f);
     HDassert(p);
-    HDassert(p_size > 0);
```
bmribler, line 93: So, does p_size need to be checked?  I haven't seen it being checked/asserted in other functions

bmribler (approved):

src/H5Oefl.c
derobins: No. If it's 0 the first check will fail and we'll jump to the error condition.

PR #2761, opened 2023-04-16
Add no subsets option to h5diff like h5dump

byrnHDF:


tools/src/h5diff/h5diff_common.c
```diff
@@ -24,7 +24,7 @@ static int check_d_input(const char *);
  * Command-line options: The user can specify short or long-named
  * parameters.
  */
-static const char         *s_opts   = "hVrv*qn:d:p:NcelxE:A:S*";
+static const char         *s_opts   = "hVrv*qn:d:p:NcelxE:A:CS*";
 static struct long_options l_opts[] = {{"help", no_arg, 'h'},
```
jhendersonHDF: Should the argument re-ordering be applied to this branch as well to keep it in sync?

derobins:
Why are these hanging?

tools/src/h5diff/h5diff_common.c
byrnHDF: Not for 1.10 branch

jhendersonHDF (approved):

derobins (approved):

PR #2765, opened 2023-04-17
Turn off GitHub concurrent actions

derobins:
This cancels way too much stuff inappropriately. We need to better understand how this feature works.

### DIFF
--- a/.github/workflows/hdfeos5.yml
+++ b/.github/workflows/hdfeos5.yml
@@ -15,9 +15,9 @@ on:
       - '**.md'
 
 # Using concurrency to cancel any in-progress job or run
-concurrency:
-  group: ${{ github.ref }}
-  cancel-in-progress: true
+#concurrency:
+#  group: ${{ github.ref }}
+#  cancel-in-progress: true
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


brtnfld:
Can you be more specific about what the issue is? Which PR actions failed due to their actions being canceled.

derobins:
It might just have been the HDF-EOS5 actions that Jordan fixed, but I had multiple PRs over the weekend that would get cancelled and required constant restarts. It was driving me bananas.

bmribler (approved):

jhendersonHDF:
I think as a workaround for the issue of cancelling older workflows for merge commits so that we only run workflows for the latest develop commit, we should change the CI so it doesn't run `on: push` and instead setup the workflow to run maybe once daily using https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

byrnHDF:
I am working on a workflow that does exactly that - still under dev but progressing.  Maybe we need to separate push (to fork) from pull-request and run different things depending on the situation.  
Once I get the daily workflow working we will have a chance to implement true multi-level CI from fork to DailyTesting.

PR #2778, opened 2023-04-19
Correct CMake option defaults - deprecate szip for libaec

byrnHDF:
Updated the default state for some options referenced elsewhere.
libaec has been the default for a few years and removing the option for the old szip library simplifies usage.


mattjala (approved):

release_docs/RELEASE.txt
```diff
@@ -47,6 +47,11 @@ New Features
 
     Configuration:
     -------------
+    - Deprecate and remove old SZIP library in favor of LIBAEC library
```
hyoklee: Use past tense for consistency - `Deprecated`.

hyoklee (changes requested):

hyoklee (approved):

release_docs/RELEASE.txt
lrknox: Deprecated and removed ...
and maybe "We removed the options in the sentence below.

lrknox (approved):

hyoklee (approved):

byrnHDF:
Fixes #2781

PR #2783, opened 2023-04-20
H5Spoint.c:  Comment fix, back porting to 1.14

Dave-Allured:
Comment here should be "point", not "hyperslab".


hyoklee (approved):
Thank you for your patch!

jhendersonHDF (approved):

PR #2786, opened 2023-04-21
Correct CMake option defaults - deprecate szip for libaec

byrnHDF:


derobins (changes requested):
Please add a blurb to release_docs/NEWSLETTER.txt since this is a change users should be aware of.

derobins (approved):

PR #2795, opened 2023-04-23
Sanitize object header message decode functions

derobins:
Cleans up all remaining decode calls besides H5Odtype's


derobins:
There will be additional cleanup passes over this code after 1.14.1 releases. Feel free to offer suggestions. Right now I'm going for "uniform" and I'll do future cleanup in en masse passes through the H5O code.

jhendersonHDF (approved):

byrnHDF (approved):

PR #2797, opened 2023-04-24
H5Odtype.c decode cleanup

derobins:
Cleans up H5Odtype.c's decode call, which was complicated by H5Tdecode()
taking a buffer but not a size.

### DIFF
--- a/src/H5Odtype.c
+++ b/src/H5Odtype.c
@@ -24,6 +24,15 @@
 #include "H5Tpkg.h"      /* Datatypes            */
 #include "H5VMprivate.h" /* Vectors and arrays   */
 
+/* Variant boundary-checking macro, used here since H5Tdecode() doesn't take a
+ * size parameter so we need to ignore the bounds checks.
+ *
+ * This is a separate macro since we don't want to inflict that behavior on
+ * the rest of the library.
+ */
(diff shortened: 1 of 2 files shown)


src/H5Odtype.c
derobins, line 1: I thought I was going to have to debug here so I did some tidying, but that wasn't the case.

test/dtypes.c
derobins, line 1: I thought I was going to have to debug in this file, so I did some tidying, but it turned out that wasn't required.

src/H5Odtype.c
```diff
@@ -24,6 +24,15 @@
...
 
+/* Variant boundary-checking macro, used here since H5Tdecode() doesn't take a
+ * size parameter so we need to ignore the bounds checks.
+ *
+ * This is a separate macro since we don't want to inflict that behavior on
+ * the rest of the library.
+ */
+#define H5_DTYPE_IS_BUFFER_OVERFLOW(skip, ptr, size, buffer_end)                                             \
+    (skip ? FALSE : ((ptr) + (size)-1) > (buffer_end))
+
```
derobins, line 35: Ugly hack, but necessary to work around H5Tencode()'s deficiencies. I'll probably version that call to take a size parameter in 2.0.0.

jhendersonHDF (approved):

byrnHDF (approved):

byrnHDF (approved):

PR #2802, opened 2023-04-24
chore: match comments in HDF5config.cmake

hyoklee:
Top comment is not consistent with `Usage` comment block.

### DIFF
--- a/config/cmake/scripts/HDF5config.cmake
+++ b/config/cmake/scripts/HDF5config.cmake
@@ -11,7 +11,8 @@
 #
 #############################################################################################
 ### ${CTEST_SCRIPT_ARG} is of the form OPTION=VALUE                                       ###
-### BUILD_GENERATOR required [Unix, VS2022, VS2019, VS201964, VS2017, VS201764, VS2015, VS201564] ###
+### BUILD_GENERATOR required:                                                             ###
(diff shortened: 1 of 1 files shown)


config/cmake/scripts/HDF5config.cmake
```diff
@@ -24,9 +25,9 @@ cmake_minimum_required (VERSION 3.18)
 #            MinGW     * MinGW Makefiles
 #            Unix      * Unix Makefiles
 #            VS2022    * Visual Studio 17 2022
-#            VS202264  * Visual Studio 17 2022
+#            VS202264  * Visual Studio 17 2022 Win64
```
byrnHDF, line 28: No that Win64 is not correct for VS2022 and VS2019

hyoklee: What do you mean? 
What's the difference between VS2022 and VS202264? 
Is VS202264 redundant?

byrnHDF: No, The generator field is the same, there is an additional option to choose the architecture. Maybe the comments just should be deleted and leave the explanation to CMake --help

hyoklee: @byrnHDF, I don't understand what you're saying.
I'll close this PR.
You can delete comments and create another PR.

hyoklee:
@byrnHDF will finish the rest.

PR #2803, opened 2023-04-24
Harden the v1 B-tree and local heap cache clients

derobins:
* Add bounds checks to the deserialize code
* Convert asserts to regular error checks
* General cleanup

### DIFF
--- a/src/H5Bcache.c
+++ b/src/H5Bcache.c
@@ -13,10 +13,8 @@
 /*-------------------------------------------------------------------------
  *
  * Created:		H5Bcache.c
- *			Oct 31 2005
- *			Quincey Koziol
  *
- * Purpose:		Implement B-tree metadata cache methods.
+ * Purpose:		Implement B-tree metadata cache methods
  *
  *-------------------------------------------------------------------------
  */
@@ -83,13 +81,9 @@ const H5AC_class_t H5AC_BT[1] = {{
 /*-------------------------------------------------------------------------
  * Function:    H5B__cache_get_initial_load_size
  *
- * Purpose:     Compute the size of the data structure on disk.
+ * Purpose:     Compute the size of the data structure on disk
  *
- * Return:      Non-negative on success/Negative on failure
- *
- * Programmer:  Quincey Koziol
- *              May 18, 2010
- *
+ * Return:      SUCCEED/FAIL
  *-------------------------------------------------------------------------
  */
 static herr_t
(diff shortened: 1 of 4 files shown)


src/H5Bcache.c
```diff
@@ -156,57 +146,70 @@ H5B__cache_deserialize(const void *_image, size_t H5_ATTR_UNUSED len, void *_uda
...
+    /* Magic number */
+    if (H5_IS_BUFFER_OVERFLOW(image, H5_SIZEOF_MAGIC, p_end))
+        HGOTO_ERROR(H5E_BTREE, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
     if (HDmemcmp(image, H5B_MAGIC, (size_t)H5_SIZEOF_MAGIC) != 0)
         HGOTO_ERROR(H5E_BTREE, H5E_BADVALUE, NULL, "wrong B-tree signature")
     image += H5_SIZEOF_MAGIC;
 
-    /* node type and level */
+    /* Node type and level */
+    if (H5_IS_BUFFER_OVERFLOW(image, 1, p_end))
```
jhendersonHDF: 2 instead of 1 here

src/H5Bcache.c
```diff
@@ -156,57 +146,70 @@ H5B__cache_deserialize(const void *_image, size_t H5_ATTR_UNUSED len, void *_uda
...
+    /* Child/key pairs */
     native = bt->native;
     for (u = 0; u < bt->nchildren; u++) {
         /* Decode native key value */
+        if (H5_IS_BUFFER_OVERFLOW(image, shared->sizeof_rkey, p_end))
+            HGOTO_ERROR(H5E_BTREE, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
         if ((udata->type->decode)(shared, image, native) < 0)
             HGOTO_ERROR(H5E_BTREE, H5E_CANTDECODE, NULL, "unable to decode key")
         image += shared->sizeof_rkey;
         native += udata->type->sizeof_nkey;
```
jhendersonHDF, line 199: Should we be tracking whether we run off the end of the native buffer as well, or is that handled elsewhere?

src/H5Bcache.c
derobins: Fixed

src/H5Bcache.c
derobins: For this pass, I was just worried about the image since that's where most of the CVE issues come from.

jhendersonHDF (approved):

byrnHDF (approved):

Issue #2779, opened 2023-04-20
Add CMakePresets.json support


byrnHDF:
This would allow the deprecation of the cacheinit.cmake files currently used. Presets were added in CMake 3.19.

CMake supports two main files, CMakePresets.json and CMakeUserPresets.json, that allow users to specify common configure options and share them with others. CMakePresets.json may be checked into a version control system, and CMakeUserPresets.json should NOT be checked in. For example, if a project is using Git, CMakePresets.json may be tracked, and CMakeUserPresets.json should be added to the .gitignore.

CMakePresets.json and CMakeUserPresets.json live in the project's root directory. They both have exactly the same format, and both are optional (though at least one must be present if --preset is specified). CMakePresets.json is meant to specify project-wide build details, while CMakeUserPresets.json is meant for developers to specify their own local build details.

byrnHDF:
Another reason is that presets have a consistent setting for any number of "standard" configurations and can be used multiple ways.

If a project has presets available, either through CMakePresets.json or CMakeUserPresets.json, the list of presets will appear in a drop-down menu in cmake-gui(1) between the source directory and the binary directory. Choosing a preset sets the binary directory, generator, environment variables, and cache variables, but all of these options can be overridden after a preset is selected.

byrnHDF:
Completed

PR #2817, opened 2023-04-26
Add support for CMakePresets and fix example download

byrnHDF:
See the changes to INSTALL_CMake.txt and RELEASE.txt for basic overview.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,237 @@
+{
+  "version": 6,
+  "include": [
+    "config/cmake-presets/hidden-presets.json"
+  ],
+  "configurePresets": [
+    {
+      "name": "ci-base-tgz",
+      "hidden": true,
+      "inherits": "ci-base",
+      "cacheVariables": {
+        "HDF5_ALLOW_EXTERNAL_SUPPORT": "NO",
+        "TGZPATH": {"type": "STRING", "value": "${sourceParentDir}/temp"}
+      }
+    },
+    {
+      "name": "ci-StdCompression",
+      "hidden": true,
+      "inherits": "ci-base-tgz",
+      "cacheVariables": {
+        "HDF5_ENABLE_Z_LIB_SUPPORT": "ON",
+        "HDF5_ENABLE_SZIP_SUPPORT": "ON",
+        "HDF5_ENABLE_SZIP_ENCODING": "ON",
+        "BUILD_ZLIB_WITH_FETCHCONTENT": "ON",
+        "ZLIB_PACKAGE_NAME": {"type": "STRING", "value": "zlib"},
+        "ZLIB_TGZ_ORIGPATH": {"type": "STRING", "value": "https://github.com/madler/zlib/releases/download/v1.2.13"},
+        "ZLIB_TGZ_ORIGNAME": {"type": "STRING", "value": "zlib-1.2.13.tar.gz"},
+        "ZLIB_USE_LOCALCONTENT": "OFF",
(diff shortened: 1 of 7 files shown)


lrknox (approved):

bmribler (approved):

CMakePresets.json
jhendersonHDF, line 1: Since this is a config type of thing, seems like it'd make sense for this file to be under the config dir instead of at the top level

byrnHDF: Not my choice - it is a CMake req!

jhendersonHDF: Ah ok. Doesn't seem great but if it's a Cmake req it is what it is.

release_docs/INSTALL_CMake.txt
```diff
@@ -1042,6 +1043,99 @@ The HDF5 CMake variables;
...
+Using the workflow preset to configure, build, test and package the standard configuration is:
+    change directory to the hdf5 source folder
+    execute "cmake --workflow --presets=ci-StdShar-<compiler-type> --fresh"
+          where <compiler-type> is  GNUC or MSVC or Clang
+
+Creating your own configurations
+--------------------------------
+The quickest way is to copy CMakePresets.json to CMakeUserPresets.json and
+edit CMakeUserPresets.json configuration names from ci-* to my-*. Change the
+configurePresets inherits only for those that you have provided alternates for
```
jhendersonHDF: This sentence reads a bit weird to me

byrnHDF: The explanation is more complicated then the process - Just a first attempt, open for suggestions.

jhendersonHDF: I think the other explanations here are fine. I'm just not sure what "Change the
configurePresets inherits only for" means here. What is configurePresets? Does "inherits" here refer to the inherits configuration setting in the file?

byrnHDF: This refers to : 
 "packagePresets": [
    {
      "name": "ci-StdShar-MSVC",
      "configurePreset": "ci-StdShar-MSVC",
      "inherits": "ci-x64-Release-MSVC"
    },

byrnHDF: And the actual configurePresets section which has "inherits"
i.e.
     {
      "name": "my-StdPlugins",
      "hidden": true,
      "inherits": ["ci-base-plugins", "my-base-tgz"],
      "cacheVariables": {
        "HDF5_ENABLE_PLUGIN_SUPPORT": "ON",
        "PLUGIN_TGZ_ORIGNAME": {"type": "STRING", "value": "hdf5_plugins-1.14.0.tar.gz"}
      }
    },

byrnHDF: "my-base-tgz" changed from "ci-base-tgz"

byrnHDF: I could have just included my User file in the docs folder maybe?

jhendersonHDF: I think wrapping inherits in quotes and maybe changing to 'the "inherits" field' would do a lot to clarify that it refers to editing the "inherits" field in the configuration file. As for "configurePresets" maybe adding "entries" after that word or showing an inline example would make it clearer.

byrnHDF: Wrapped the presets names in quotes

jhendersonHDF (approved):

byrnHDF:
Fixes #2779

PR #2823, opened 2023-04-26
Implement selection I/O with type conversion

fortnern:
Also enables collective I/O with type conversion. Also added H5Pset_selection_io(), H5Pget_selection_io(), and H5Pget_no_selection_io_cause() API functions to manage the selection I/O feature, and H5Pset_modify_write_buf() and H5Pget_modify_write_buf() to allow the library to modify write buffers.


test/select_io_dset.c
```diff
@@ -0,0 +1,3212 @@
...
+/*
+ * Programmer:
+ *
+ * Purpose:    Tests selection IO for the dataset interface (H5D)
+ */
+
+#include "testhdf5.h"
+#include "H5srcdir.h"
+
+const char *FILENAME[] = {"select_io", /* 0 */
```
byrnHDF, line 22: Is this file created and deleted - or does it need to be added to the build system cleanup sections?
(Also same for parallel test)

fortnern: It looks like the file is deleted by h5_cleanup() in the serial test and MPI_File_delete() in the parallel test

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,190 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ */
+H5_DLL herr_t H5Pset_selection_io(hid_t dxpl_id, H5D_selection_io_mode_t selection_io_mode);
+
+/**
+ *
+ * \ingroup DXPL
+ *
+ * \brief Retrieves the selection I/O mode
+ *
+ * \param[in] plist_id Property list identifier
```
brtnfld: Should be dxpl_id, but I think it should use the alias \plist_id{dxpl_id}

But, what is the convention? Some APIs use dxpl_id and some used plist_id, and yet they all have the same description. I think all of the new ones should use plist_id.

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,190 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ *          Selection I/O was not performed because of chunk cache
+ *          - #H5D_SEL_IO_TCONV_BUF_TOO_SMALL
+ *          Selection I/O was not performed because the type conversion buffer is too small
+ *          - #H5D_SEL_IO_BKG_BUF_TOO_SMALL
+ *          Selection I/O was not performed because the type conversion background buffer is too small
+ *          - #H5D_SEL_IO_DEFAULT_OFF
+ *          Selection I/O was not performed because the selection I/O mode is DEFAULT and the library chose it
+ * to be off for this case
+ *
+ * \since 1.14.3
```
vchoi-hdfgroup: Should be 1.14.1

src/H5Ppublic.h
fortnern: Updated, thanks. Not sure on the convention.

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,190 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+
+/**
+ * \ingroup DXPL
+ *
+ * \brief Retrieves the cause for not performing selection or vector I/O on the
+ *        last parallel I/O call
+ *
+ * \dxpl_id{plist_id}
+ * \param[out] no_selection_io_cause A bitwise set value indicating the relevant
+ *                                   causes that prevented selection I/O to be performed
```
brtnfld: to be -> from being

src/H5Ppublic.h
```diff
@@ -357,11 +357,72 @@ typedef enum H5D_mpio_no_collective_cause_t {
...
-    H5D_MPIO_NO_COLLECTIVE_MAX_CAUSE = 0x100
+    H5D_MPIO_NO_SELECTION_IO = 0x100,
+    /**< Collective I/O would be supported by selection or vector I/O but that feature was disabled
+       (see causes via H5Pget_no_selection_io_cause()) */
+    H5D_MPIO_NO_COLLECTIVE_MAX_CAUSE = 0x200
     /**< Sentinel */
 } H5D_mpio_no_collective_cause_t;
 //! <!-- [H5D_mpio_no_collective_cause_t_snip] -->
 
+/**
```
brtnfld, line 368: Why are these not in a struct like the other no*cause cases? Why unsigned?

src/H5Pdxpl.c
```diff
@@ -2386,3 +2496,248 @@ H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned rank, H5S_seloper
...
+ * Purpose:	    Retrieves causes for not performing selection I/O
+ *
+ * Return:	    Non-negative on success/Negative on failure
+ *
+ * Programmer:	Vailin Choi
+ *              April 17, 2023
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Pget_no_selection_io_cause(hid_t plist_id, uint32_t *no_selection_io_cause /*out*/)
```
brtnfld, line 2592: why is this marked` /*out*/ `and others are not?

testpar/t_subfiling_vfd.c
```diff
@@ -1011,12 +1014,19 @@ test_read_different_stripe_size(void)
...
     hid_t                   dset_id      = H5I_INVALID_HID;
+    hid_t                   dxpl_id      = H5I_INVALID_HID;
     hid_t                   fspace_id    = H5I_INVALID_HID;
     char                   *tmp_filename = NULL;
     void                   *buf          = NULL;
 
     curr_nerrors = nerrors;
 
+    dxpl_id = H5Pcreate(H5P_DATASET_XFER);
+    VRFY((dxpl_id >= 0), "DCPL creation succeeded");
```
bmribler: DXPL?  More below.

testpar/t_subfiling_vfd.c
```diff
@@ -427,6 +427,9 @@ test_stripe_sizes(void)
     dxpl_id = H5Pcreate(H5P_DATASET_XFER);
     VRFY((dxpl_id >= 0), "DCPL creation succeeded");
 
+    /* Set selection I/O mode on DXPL */
+    VRFY((H5Pset_selection_io(dxpl_id, H5D_SELECTION_IO_MODE_ON) >= 0), "H5Pset_selection_io succeeded");
```
bmribler, line 431: There is a minor inconsistency in the messages.  Sometimes, "< function name > succeeded", and other times, "Doing something succeeded" or "doing something succeeded".

testpar/t_dset.c
```diff
@@ -3538,34 +3555,59 @@ test_no_collective_cause_mode(int selection_mode)
...
         test_name = "Broken Collective I/O - Datatype Conversion";
-        no_collective_cause_local_expected |= H5D_MPIO_DATATYPE_CONVERSION;
-        no_collective_cause_global_expected |= H5D_MPIO_DATATYPE_CONVERSION;
+
         /* set different sign to trigger type conversion */
         data_type = H5T_NATIVE_UINT;
+
+        /* Disable selection I/O since datatype conversion is supported in collective with selection I/O */
+        ret = H5Pset_selection_io(dxpl_write, H5D_SELECTION_IO_MODE_OFF);
+
```
bmribler: Extra line?

testpar/t_dset.c
```diff
@@ -3348,13 +3348,26 @@ test_actual_io_mode(int selection_mode)
...
     MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
+    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
 
     /* Only run these tests if selection I/O is not being used - selection I/O
      * bypasses this IO mode decision - it's effectively always multi chunk
      * currently */
-    if (!H5_use_selection_io_g) {
+
+    dxpl_id = H5Pcreate(H5P_DATASET_XFER);
```
bmribler, line 3364: No verifying

testpar/t_2Gio.c
```diff
@@ -4072,6 +4094,20 @@ test_no_collective_cause_mode(int selection_mode)
                                           &no_collective_cause_global_write);
     VRFY((ret >= 0), "retrieving no collective cause succeeded");
 
+    ret = H5Pget_no_selection_io_cause(dxpl_write, &no_selection_io_cause_read);
```
bmribler: no_selection_io_cause_write?

src/H5Ppublic.h
fortnern: The others are in enums. We decided to avoid putting bitwise values in enums because that is inconsistent with the normal usage of enums and potentially confusing, since the flags parameter can take on a value that does not match one of the enum values. In addition, unless I'm missing something there's no advantage using enum values instead of simple #defines.

fortnern: They are unsigned because they are bitflag values and need to be able to be combined in a bitwise fashion

fortnern: We plan to change the other no*cause values to #defines with 2.0

testpar/t_subfiling_vfd.c
fortnern: I think this is fine

testpar/t_dset.c
fortnern: You mean you think the comment should be broken into two lines? I generally rely on the formatter to pick line breaks, which I think is set at 110. This is 107.

jhendersonHDF:
Just an overall question on this PR: is there a downside currently to always enabling selection I/O as opposed to needing to set it with the new API routines?

test/select_io_dset.c
```diff
@@ -0,0 +1,3212 @@
...
+        HDfree(l_total_wbuf2);
+        HDfree(l_total_wbuf2_bak);
+        HDfree(s_total_rbuf2);
+
+        HDfree(s1_total_wbuf3);
+        HDfree(s1_total_wbuf3_bak);
+        HDfree(s3_total_rbuf3);
+        HDfree(s4_total_wbuf3);
+        HDfree(s4_total_wbuf3_bak);
+        HDfree(s1_total_rbuf3);
```
jhendersonHDF, line 2672: All these buffers above should probably be set to NULL. Otherwise if the next iteration fails somewhere we might double-free segfault in the error cleanup section.

testpar/t_select_io_dset.c
```diff
@@ -0,0 +1,3775 @@
...
+        HDfree(l_total_wbuf2);
+        HDfree(l_total_wbuf2_bak);
+        HDfree(s_total_rbuf2);
+
+        HDfree(s1_total_wbuf3);
+        HDfree(s1_total_wbuf3_bak);
+        HDfree(s3_total_rbuf3);
+        HDfree(s4_total_wbuf3);
+        HDfree(s4_total_wbuf3_bak);
+        HDfree(s1_total_rbuf3);
```
jhendersonHDF, line 3387: Suggest setting all these buffers to NULL to prevent double-free segfault.

testpar/t_select_io_dset.c
```diff
@@ -0,0 +1,3775 @@
...
+
+    /* Barrier to make sure all ranks are done before deleting the file, and
+     * also to clean up output (make sure PASSED is printed before any of the
+     * following messages) */
+    if (MPI_Barrier(MPI_COMM_WORLD) != MPI_SUCCESS)
+        P_TEST_ERROR;
+
+    /* Delete file */
+    if (MAINPROCESS)
+        if (MPI_File_delete(FILENAME, MPI_INFO_NULL) != MPI_SUCCESS)
```
jhendersonHDF: Can this test be run with other parallel-enabled VFDs? If so this should probably be a call to H5Fdelete (either collectively with the file fapl or by MAINPROCESS with a serial fapl) so that VFDs like the Subfiling VFD can correctly delete the test file.

testpar/t_dset.c
```diff
@@ -3538,34 +3556,59 @@ test_no_collective_cause_mode(int selection_mode)
...
+        no_selection_io_cause_expected |= H5D_SEL_IO_DISABLE_BY_API;
     }
 
     if (selection_mode & TEST_DATA_TRANSFORMS) {
         test_name = "Broken Collective I/O - DATA Transforms";
-        no_collective_cause_local_expected |= H5D_MPIO_DATA_TRANSFORMS;
-        no_collective_cause_global_expected |= H5D_MPIO_DATA_TRANSFORMS;
+
+        /* Set transform */
+        ret = H5Pset_data_transform(dxpl_write, "x+1");
```
jhendersonHDF, line 3585: As an aside, what's the current level of support for collective selection I/O in the library? There's an open issue about supporting data transformations with parallel compression and since parallel compressed writes require collective I/O, seems like we could at least partially support that now.

src/H5CXprivate.h
```diff
@@ -115,6 +115,9 @@ H5_DLL herr_t H5CX_get_filter_cb(H5Z_cb_t *filter_cb);
 H5_DLL herr_t H5CX_get_data_transform(H5Z_data_xform_t **data_transform);
 H5_DLL herr_t H5CX_get_vlen_alloc_info(H5T_vlen_alloc_info_t *vl_alloc_info);
 H5_DLL herr_t H5CX_get_dt_conv_cb(H5T_conv_cb_t *cb_struct);
+H5_DLL herr_t H5CX_get_selection_io_mode(H5D_selection_io_mode_t *selection_io_mode);
+H5_DLL herr_t H5CX_get_no_selection_io_cause(uint32_t *no_selection_io_cause);
+H5_DLL herr_t H5CX_get_modify_write_buf(hbool_t *modify_write_buf);
```
jhendersonHDF, line 120: No need for set routines for selection io mode or modify write buf?

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,191 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ * \brief Sets the selection I/O mode
+ *
+ * \dxpl_id{plist_id}
+ * \param[in] selection_io_mode    The selection I/O mode to be set
+ *
+ * \return \herr_t
+ *
+ * \details H5Pset_selection_io() sets the selection I/O mode
+ *          \p selection_io_mode in the dataset transfer property
+ *          list \p dxpl_id.
```
jhendersonHDF: Should this be updated to plist_id?

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,191 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ *
+ * \brief Retrieves the selection I/O mode
+ *
+ * \dxpl_id{plist_id}
+ * \param[out] selection_io_mode   The selection I/O mode
+ *
+ * \return \herr_t
+ *
+ * \details H5Pget_selection_io() queries the selection I/O mode set in
+ *          in the dataset transfer property list \p dxpl_id.
```
jhendersonHDF: plist_id?

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,191 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ * \details H5Pset_selection_io() sets the selection I/O mode
+ *          \p selection_io_mode in the dataset transfer property
+ *          list \p dxpl_id.
+ *
+ *          This can be used to enable collective I/O with type conversion, or
+ *          with custom VFDs that support vector or selection I/O.
+ *
+ *          Values that can be set in \p selection_io_mode:
+ *          \snippet this H5D_selection_io_mode_t_snip
+ *          \click4more
```
jhendersonHDF, line 8288: Is this some special doxygen command?

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,191 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ * \brief Allows the library to modify the contents of the write buffer
+ *
+ * \dxpl_id{plist_id}
+ * \param[in] modify_write_buf   Whether the library can modify the contents of the write buffer
+ *
+ * \return \herr_t
+ *
+ * \details H5Pset_modify_write_buf() sets whether the library is allowed to
+ *          modify the contents of write buffers passed to HDF5 API routines
+ *          that are passed the dataset transfer property list \p dxpl_id.  The
```
jhendersonHDF: plist_id?

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,191 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ *
+ * \brief Retrieves the "modify write buffer" property
+ *
+ * \dxpl_id{plist_id}
+ * \param[out] modify_write_buf   Whether the library can modify the contents of the write buffer
+ *
+ * \return \herr_t
+ *
+ * \details H5Pget_modify_write_buf() gets the "modify write buffer" property
+ *          from the dataset transfer property list \p dxpl_id.  This property
```
jhendersonHDF: plist_id?

src/H5Ppublic.h
```diff
@@ -8204,6 +8265,191 @@ H5_DLL herr_t H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned ran
...
+ *          determines whether the library is allowed to  modify the contents of
+ *          write buffers passed to HDF5 API routines that are passed
+ *          \p dxpl_id.  The default value for modify_write_buf is FALSE.
```
jhendersonHDF: plist_id?

src/H5Pdxpl.c
```diff
@@ -441,6 +461,24 @@ H5P__dxfr_reg_prop(H5P_genclass_t *pclass)
                            H5D_XFER_DSET_IO_SEL_CLOSE) < 0)
         HGOTO_ERROR(H5E_PLIST, H5E_CANTINSERT, FAIL, "can't insert property into class")
 
+    if (H5P__register_real(pclass, H5D_XFER_SELECTION_IO_MODE_NAME, H5D_XFER_SELECTION_IO_MODE_SIZE,
+                           &H5D_def_selection_io_mode_g, NULL, NULL, NULL, H5D_XFER_SELECTION_IO_MODE_ENC,
+                           H5D_XFER_SELECTION_IO_MODE_DEC, NULL, NULL, NULL, NULL) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTINSERT, FAIL, "can't insert property into class")
+
+    /* Register the cause of no selection I/O property */
+    /* (Note: this property should not have an encode/decode callback -QAK) */
```
jhendersonHDF: Remove copied comment from QAK

src/H5Pdxpl.c
```diff
@@ -2386,3 +2496,248 @@ H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned rank, H5S_seloper
...
+ * Return:      Success:    Non-negative
+ *              Failure:    Negative
+ *
+ * Programmer:  Vailin Choi
+ *              March 5, 2023
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Pget_selection_io(hid_t plist_id, H5D_selection_io_mode_t *selection_io_mode /*out*/)
```
jhendersonHDF, line 2559: Should check for H5P_DEFAULT and set plist_id to correct library default DXPL if that's passed in

src/H5Pdxpl.c
```diff
@@ -2386,3 +2496,248 @@ H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned rank, H5S_seloper
...
+ * Purpose:	    Retrieves causes for not performing selection I/O
+ *
+ * Return:	    Non-negative on success/Negative on failure
+ *
+ * Programmer:	Vailin Choi
+ *              April 17, 2023
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Pget_no_selection_io_cause(hid_t plist_id, uint32_t *no_selection_io_cause /*out*/)
```
jhendersonHDF, line 2592: Should check for H5P_DEFAULT and set plist_id to correct library default DXPL if that's passed in

src/H5Pdxpl.c
```diff
@@ -2386,3 +2496,248 @@ H5Pset_dataset_io_hyperslab_selection(hid_t plist_id, unsigned rank, H5S_seloper
...
+ *
+ * Purpose:     Retrieves the "modify write buffer" property
+ *
+ * Return:      Success:    Non-negative
+ *              Failure:    Negative
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Pget_modify_write_buf(hid_t plist_id, hbool_t *modify_write_buf /*out*/)
```
jhendersonHDF, line 2724: Should check for H5P_DEFAULT and set plist_id to correct library default DXPL if that's passed in

fortnern:
> Just an overall question on this PR: is there a downside currently to always enabling selection I/O as opposed to needing to set it with the new API routines?

It could be slower for parallel I/O with strided selections since there's currently no compact way to pass a regular selection (only vector I/O is implemented). When used with type conversion it requires potentially very large type conversion buffers. And there are also intermittent failures in the parallel tests with MPICH (most likely an MPICH issue).

testpar/t_dset.c
fortnern: Selection I/O supports transforms with collective, but selection I/O doesn't support compressions, so you can't have both yet unfortunately

src/H5CXprivate.h
fortnern: Not needed by the library since the library never changes it. There are several other properties where this is the case. For something like this it's best not to create this routine since we really want to make sure we don't violate const if the app didn't set this property.

src/H5Ppublic.h
fortnern: It seems to print "(Click on a enumerator, field, or type for more information.)" in the html

src/H5Pdxpl.c
fortnern: Are you sure about this? None of the other H5Pget* functions seem to do this

src/H5Pdxpl.c
fortnern: See above

src/H5Pdxpl.c
fortnern: See above

src/H5Pdxpl.c
fortnern: I guess it might be useful to make this work, especially since we're bad about documenting the default values for properties (I just had to scour through the code to find some to fix one of the above tests), but we should really apply it to all at once as a separate PR

test/select_io_dset.c
fortnern: Fixed, thanks

testpar/t_select_io_dset.c
fortnern: Fixed, thanks

src/H5Pdxpl.c
jhendersonHDF: Yeah I think we should check for and handle H5P_DEFAULT in every API routine where a plist is passed in. I think it's just something that's easy to forget and so many routines are missing the check. Maybe we should move the logic down into H5P_object_verify since it accepts a "plist type" parameter. Other places where H5I_object is used won't be covered but it'd be a step in the right direction.

testpar/t_dset.c
bmribler: No, I meant there is an extra line, 3574.  It's ok if it's intentional, hence, the question mark. :)

src/H5Dio.c
```diff
@@ -833,25 +910,41 @@ H5D__ioinfo_init(size_t count, H5D_dset_io_info_t *dset_info, H5D_io_info_t *io_
...
+
     /* Determine if the file was opened with an MPI VFD */
     if (count > 0)
         io_info->using_mpi_vfd = H5F_HAS_FEATURE(dset_info[0].dset->oloc.file, H5FD_FEAT_HAS_MPI);
 #endif /* H5_HAVE_PARALLEL */
 
+    /* Check if we could potentially use in-place type conversion */
+    if (op_type == H5D_IO_OP_READ)
+        /* Always on for read (modulo other restrictions that are handled in layout callbacks) */
+        io_info->may_use_in_place_tconv = (op_type == H5D_IO_OP_READ);
```
jhendersonHDF: Should this be hoisted out of the if block here since the conditional is the same as the check in the if? Though I guess the field gets initialized in the `op_type == H5D_IO_OP_READ` case anyway due to the memset of io_info above.

fortnern: Oops, that logic was left over from before I implemented modify_write_buf. I'll fix it.

testpar/t_dset.c
fortnern: Ok I'll reformat thanks

src/H5Dpkg.h
```diff
@@ -67,7 +67,49 @@
...
+        if ((IO_INFO)->may_use_in_place_tconv) {                                                             \
+            size_t mem_type_size  = ((IO_INFO)->op_type == H5D_IO_OP_READ)                                   \
+                                        ? (DINFO)->type_info.dst_type_size                                   \
+                                        : (DINFO)->type_info.src_type_size;                                  \
+            size_t file_type_size = ((IO_INFO)->op_type == H5D_IO_OP_READ)                                   \
+                                        ? (DINFO)->type_info.src_type_size                                   \
+                                        : (DINFO)->type_info.dst_type_size;                                  \
+                                                                                                             \
+            /* Make sure the memory type is not smaller than the file type, otherwise the memory buffer      \
+             * won't be big enough tp serve as the type conversion buffer */                                 \
```
jhendersonHDF: tp -> to. Guess codespell didn't catch that one.

jhendersonHDF (approved):
I'm not as familiar with selection I/O as I'd like to be for reviewing this PR, but everything here looks reasonable enough to me.

fortnern:
I'll fix the typo and have one minor change then I think it's good to go

fortnern:
Working on a couple tests that we discussed at the meeting, but should be fine to merge without if we're in a hurry. Edit: done.

derobins (approved):

PR #2837, opened 2023-04-27
Add bug note to H5Dget_space_status documentation (#2788)

jhendersonHDF:


src/H5Dpublic.h
```diff
@@ -423,6 +423,20 @@ H5_DLL hid_t H5Dget_space(hid_t dset_id);
  * \details H5Dget_space_status() determines whether space has been allocated
  *          for the dataset \p dset_id.
  *
+ * \note \Bold{BUG:} Prior to the HDF5 1.13.1, 1.12.2 and 1.10.9 releases,
```
bmribler: I thought we treat 1.13.x as if they never exist...

jhendersonHDF: That's true, although changing this to 1.14.0 would be slightly inaccurate. I've also seen 1.13 releases being used in the wild, which complicates documenting this a bit as well.

byrnHDF (approved):

PR #2844, opened 2023-04-28
Remove unused variable warning in H5C.c

hyoklee:
Sunspot compiler (icx) throws a warning.

`CMAKE_C_COMPILER:FILEPATH=/soft/restricted/CNDA/updates/2022.12.30.001/oneapi/\
cmpiler/trunk-20230201/compiler/linux/bin/icx`

See https://my.cdash.org/viewBuildError.php?type=1&buildid=2327793 for details.

### DIFF
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -6579,7 +6579,7 @@ H5C__make_space_in_cache(H5F_t *f, size_t space_needed, hbool_t write_permitted)
     H5C_cache_entry_t *entry_ptr;
     H5C_cache_entry_t *prev_ptr;
     H5C_cache_entry_t *next_ptr;
-    herr_t             ret_value          = SUCCEED; /* Return value */
+    herr_t             ret_value = SUCCEED; /* Return value */
 
     FUNC_ENTER_PACKAGE
 
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -6579,8 +6579,8 @@ H5C__make_space_in_cache(H5F_t *f, size_t space_needed, hbool_t write_permitted)
     H5C_cache_entry_t *entry_ptr;
     H5C_cache_entry_t *prev_ptr;
     H5C_cache_entry_t *next_ptr;
-    uint32_t num_corked_entries = 0;
-    herr_t ret_value = SUCCEED; /* Return value */
+    uint32_t           num_corked_entries = 0;
+    herr_t             ret_value          = SUCCEED; /* Return value */
 
     FUNC_ENTER_PACKAGE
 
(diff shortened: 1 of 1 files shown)


hyoklee:
This warning occurs with the following configuration:

`cmake -DHDF5_BUILD_FORTRAN:BOOL=ON -DHDF5_ENABLE_MAP_API:BOOL=ON ..`

src/H5C.c
```diff
@@ -6579,8 +6579,10 @@ H5C__make_space_in_cache(H5F_t *f, size_t space_needed, hbool_t write_permitted)
     H5C_cache_entry_t *entry_ptr;
     H5C_cache_entry_t *prev_ptr;
     H5C_cache_entry_t *next_ptr;
-    uint32_t           num_corked_entries = 0;
-    herr_t             ret_value          = SUCCEED; /* Return value */
+#if !defined(SYCL_LANGUAGE_VERSION) || !defined(__INTEL_LLVM_COMPILER)
```
byrnHDF: I don't understand how this can be undefined - it is incremented in line 6631. And the value is used in an assert at line 6758.

hyoklee: Ask the developer who wrote H5C.c.
The responsible coder should understand this behavior.

jhendersonHDF: This is likely a false positive by the compiler based on the code below. And since it failed to compile on ubuntu when the variable was removed before `#if !defined(SYCL_LANGUAGE_VERSION) || !defined(__INTEL_LLVM_COMPILER)` was added, this will probably fail to compile for ICC in the absence of any compiler bugs. At a minimum, further uses of `num_corked_entries` below would need to also be surrounded by `#if !defined(SYCL_LANGUAGE_VERSION) || !defined(__INTEL_LLVM_COMPILER)`, but I don't believe that's the correct fix.

hyoklee:
The https://cs.illinois.edu/academics/courses/cs426 in 1996 reminded me how to fix it.

jhendersonHDF:
> The https://cs.illinois.edu/academics/courses/cs426 in 1996 reminded me how to fix it.

Hmm interesting, does ICX still give a warning if you use `num_corked_entries++;` rather than `num_corked_entries = num_corked_entries + 1;` or `++num_corked_entries`? It's a little more compact and readable to use the post-increment and if ICC throws a warning for that then that seems like a fundamental compiler flaw.

jhendersonHDF:
Regardless this is a compiler flaw so I don't see why we should be working around it.

byrnHDF (approved):

derobins (approved):

qkoziol:
This was probably about the variable only being used for the later assert().  I have a fix for it in my current cache PR, so will resolve this issue there (and revert this particular change, probably).

PR #2855, opened 2023-04-28
Don't use strnlen when len is not known

derobins:
The datatype object header message decode function was updated to do bounds checking on the decode buffer. This buffer may arrive with no buffer size via H5Tdecode(), in which case the buffer size will have been set to SIZE_MAX by the library.

This fix changes the string length calls to strlen when we don't know the buffer size (and avoids a potential compiler bug with icc 17).

### DIFF
--- a/src/H5Odtype.c
+++ b/src/H5Odtype.c
@@ -336,18 +336,31 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
             for (dt->shared->u.compnd.nmembs = 0; dt->shared->u.compnd.nmembs < nmembs;
                  dt->shared->u.compnd.nmembs++) {
 
-                size_t   actual_name_length;                /* Actual length of name */
-                size_t   max   = (size_t)(p_end - *pp + 1); /* Max possible name length */
-                unsigned ndims = 0;                         /* Number of dimensions of the array field */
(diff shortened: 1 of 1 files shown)


src/H5Odtype.c
```diff
@@ -336,18 +336,31 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
...
+
+                    size_t max = (size_t)(p_end - *pp + 1); /* Max possible name length */
+
+                    actual_name_length = HDstrnlen((const char *)*pp, max);
+                    if (actual_name_length == max)
+                        HGOTO_ERROR(H5E_OHDR, H5E_NOSPACE, FAIL, "field name not null terminated")
+                }
+                else {
+                    /* The buffer end can't be determined when it's an unbounded buffer
+                     * passed via H5Tdecode(), so don't bounds check and hope for
```
gagsports, line 358: `don't bounds check` -> `don't check bounds`

lrknox (approved):
Also passed all tests on jelly with icc 17.0

jhendersonHDF (approved):

src/H5Odtype.c
hyoklee, line 358: ```don't bounds check``` -> ```don't check bounds```

PR #2856, opened 2023-04-29
Harden H5O cache deserialize calls

derobins:


lrknox (approved):

byrnHDF (approved):

bmribler (approved):

src/H5Ocache.c
```diff
@@ -172,33 +164,26 @@ H5O__cache_get_initial_load_size(void H5_ATTR_UNUSED *_udata, size_t *image_len)
...
     HDassert(udata);
     HDassert(actual_len);
     HDassert(*actual_len == image_len);
 
     /* Deserialize the object header prefix */
-    if (H5O__prefix_deserialize((const uint8_t *)image, udata) < 0)
+    if (H5O__prefix_deserialize((const uint8_t *)image, image_len, udata) < 0)
         HGOTO_ERROR(H5E_OHDR, H5E_CANTDECODE, FAIL, "can't deserialize object header prefix")
 
     /* Sanity check */
```
jhendersonHDF, line 189: Should the assertion below be converted to a normal error check similar to what was done further down in this file?

src/H5Ocache.c
```diff
@@ -268,53 +248,50 @@ H5O__cache_verify_chksum(const void *_image, size_t len, void *_udata)
...
     if (NULL == udata->oh) {
         /* Deserialize the object header prefix */
-        if (H5O__prefix_deserialize((const uint8_t *)image, udata) < 0)
+        if (H5O__prefix_deserialize((const uint8_t *)image, len, udata) < 0)
             HGOTO_ERROR(H5E_OHDR, H5E_CANTDECODE, NULL, "can't deserialize object header prefix")
 
-        /* Sanity check */
-        HDassert(udata->oh);
-    } /* end if */
+        if (NULL == udata->oh)
```
jhendersonHDF: Or should this go back to being an assert? If we didn't set udata->oh after decoding that seems like something more appropriately caught by an assert rather than a normal error check.

src/H5Ocache.c
```diff
@@ -1127,14 +1022,20 @@ H5O__prefix_deserialize(const uint8_t *_image, H5O_cache_ud_t *udata)
     /* (indicates version 2 or later) */
     if (!HDmemcmp(image, H5O_HDR_MAGIC, (size_t)H5_SIZEOF_MAGIC)) {
         /* Magic number */
+        if (H5_IS_BUFFER_OVERFLOW(image, H5_SIZEOF_MAGIC, p_end))
```
jhendersonHDF: Should probably move this check outside the block before the memcmp call

src/H5Ocache.c
derobins: I thought about doing that, but there's no obvious error path where the above call succeeded and udata->oh isn't set.

jhendersonHDF (approved):

src/H5Ocache.c
derobins: Fixed

src/H5Ocache.c
derobins: Switched back to an assert

PR #2859, opened 2023-04-30
Fix memory leaks in H5Dwrite w/ selection I/O

derobins:


derobins:
I'll look deeper, but this may also need a RELEASE.txt note since it looks like this has been around since 1.13 days when selection I/O was first implemented.

byrnHDF (approved):

fortnern (approved):
Looks fine, though the formatting/ordering should ideally be the same between read and write. A release note probably isn't necessary because selection I/O was only an experimental feature in previous versions.

derobins:
> Looks fine, though the formatting/ordering should ideally be the same between read and write. A release note probably isn't necessary because selection I/O was only an experimental feature in previous versions.

Good call. I didn't even look at the read case, but I'll make those look the same before pushing.

Issue #2791, opened 2023-04-21
CMake can't generate Xcode project.


hyoklee:
**Describe the bug**

1. git clone repo
2. mkdir build && cd build && cmake -G Xcode ..

```
-- Found Perl: /usr/bin/perl (found version "5.30.3") 
CMake Error at CMakeLists.txt:614 (if):
  if given arguments:

    "MATCHES" "Debug" "OR" "MATCHES" "Developer"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
See also "/Users/hyoklee/src/hdf5.HDFGroup/build/CMakeFiles/CMakeOutput.log".
See also "/Users/hyoklee/src/hdf5.HDFGroup/build/CMakeFiles/CMakeError.log".
```

**Expected behavior**

Create Xcode project.

**Platform (please complete the following information)**
 - HDF5 version: develop def21b1e33f2499ed4d51af0708d12242d63fa63
 - OS and version: Darwin nene.ad.hdfgroup.org 21.6.0 Darwin Kernel Version 21.6.0: Thu Mar  9 20:08:59 PST 2023; root:xnu-8020.240.18.700.8~1/RELEASE_X86_64 x86_64
 - Compiler and version: AppleClang 14.0.0.14000029
 - Build system (e.g. CMake, Autotools) and version: cmake version 3.25.2

hyoklee:
If I run `cmake -G Xcode -D HDF_CFG_NAME:STRING="Debug" ..`, 
I get 547 lines of these repeating messages:

```
CMake Error in src/CMakeLists.txt:
  The custom command generating

    /Users/hyoklee/src/hdf5/build/src/gen_SRCS.stamp1

  is attached to multiple targets:

    gen_hdf5-static
    gen_hdf5-shared

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".


CMake Error in examples/CMakeLists.txt:
  The custom command generating

    /Users/hyoklee/src/hdf5/build/src/gen_SRCS.stamp1

  is attached to multiple targets:

    gen_hdf5-static
    gen_hdf5-shared

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".

```

byrnHDF:
Use CMAKE_BUILD_TYPE

hyoklee:
Can you make `CMAKE_BUILD_TYPE` not required?

byrnHDF:
No. CMake required.

[...]

PR #2863, opened 2023-04-30
Correct usage of CMAKE_BUILD_TYPE

byrnHDF:


config/cmake/HDFMacros.cmake
```diff
@@ -16,8 +16,13 @@ macro (SET_HDF_BUILD_TYPE)
   if (_isMultiConfig)
     set_property (CACHE CMAKE_CONFIGURATION_TYPES PROPERTY STRINGS "Debug" "Release"
       "MinSizeRel" "RelWithDebInfo" "Developer" FORCE)
-    set (HDF_CFG_NAME "RelWithDebInfo")
-    set (HDF_BUILD_TYPE ${CMAKE_CFG_INTDIR})
+    if (CMAKE_BUILD_TYPE)
+      set (HDF_CFG_NAME ${CMAKE_BUILD_TYPE})
+      set (HDF_BUILD_TYPE ${CMAKE_BUILD_TYPE})
```
hyoklee, line 21: Is the difference between `HDF_CFG_NAME` and `HDF_BUILD_TYPE` documented somewhere? Can we consolidate them?

byrnHDF: You're right, they should be documented, but aren't yet. I don't believe they can be combined, but will check.

lrknox (approved):

config/cmake/HDFMacros.cmake
derobins: One of the things I was doing in the patch I was putting together included a comment block where I documented things like this. I think it's something that needs to be done, but I don't think that should hold this PR.

derobins (approved):

byrnHDF:
This should fix #2791

PR #2876, opened 2023-05-02
2my feature select io tconv

vchoi-hdfgroup:
Add H5Pget_actual_selection_io_mode().


fortnern:
My original intent was that this would be a bitflags-type parameter, so instead of assigning the mode the library would get / bitwise-or / set. I need to think some more about whether this is the right approach though

vchoi-hdfgroup:
I would wait for your decision on how to proceed with this PR.

fortnern:
After thinking more about this, I think that what I mentioned earlier is fine. We'll just have to be clear in the documentation that this is a bitflags parameter than can include multiple options, and that it also includes metadata I/O performed (i.e. chunk index reads/writes, vlen, etc). We could potentially only include raw data I/O by checking the I/O type, but we'd potentially have to check an array of types for vector I/O, every time even if the user doesn't care about this function at all, which I don' think is worth it.

fortnern:
So the places where you currently set the selection I/O mode will look like:
H5CX_get_actual_selection_io_mode(&selection_io_mode);
selection_io_mode |= <I/O mode>
H5CX_set_actual_selection_io_mode(selection_io_mode);

src/H5Pdxpl.c
```diff
@@ -473,6 +477,13 @@ H5P__dxfr_reg_prop(H5P_genclass_t *pclass)
                            NULL) < 0)
         HGOTO_ERROR(H5E_PLIST, H5E_CANTINSERT, FAIL, "can't insert property into class")
 
+    /* Register the actual selection I/O mode property */
+    /* (Note: this property should not have an encode/decode callback -QAK) */
```
fortnern: Should probably remove Quincey's initials here. You can add your own if you want, or just leave it unsigned

src/H5Ppublic.h
```diff
@@ -8397,6 +8402,45 @@ H5_DLL herr_t H5Pget_selection_io(hid_t plist_id, H5D_selection_io_mode_t *selec
...
 
+/**
+ * \ingroup DXPL
+ *
+ * \brief Retrieves the type of I/O that HDF5 actually performed on
+ *        last I/O call (not necessarily the type requested)
+ *
+ * \dxpl_id{plist_id}
+ * \param[out] actual_selection_io_mode A bitwise set value indicating the
+ *                                      type of I/O performed
```
fortnern: type(s)

src/H5Ppublic.h
```diff
@@ -8397,6 +8402,45 @@ H5_DLL herr_t H5Pget_selection_io(hid_t plist_id, H5D_selection_io_mode_t *selec
...
+ * \return \herr_t
+ *
+ * \par Motivation:
+ *      A user can request selection I/O to be performed via a data transfer
+ *      property list (DXPL).  This can be used to enable collective I/O with
+ *      type conversion, or with custom VFDs that support vector or selection
+ *      I/O.  However, there are conditions that can cause HDF5 to forgo
+ *      selection or vector I/O and perform legacy (scalar) I/O instead.
```
fortnern, line 8421: <after existing text> This function allows the user to determine which type or types of I/O were actually performed.

test/select_io_dset.c
```diff
@@ -184,6 +200,13 @@ test_no_type_conv(hid_t fid, unsigned chunked, unsigned dtrans, unsigned mwbuf)
     if (H5Dwrite(did, H5T_NATIVE_INT, H5S_ALL, H5S_ALL, dxpl, wbuf) < 0)
         FAIL_STACK_ERROR;
 
+    if (H5Pget_no_selection_io_cause(dxpl, &no_selection_io_cause) < 0)
+        FAIL_STACK_ERROR;
+
+    if (!no_selection_io_cause)
```
fortnern: Why do you call H5Pget_no_selection_io_cause() here (and in similar places)?

fortnern: Also does check-vfd run with the MPIO vfd (even for serial tests like this)? If so this might fail since it'll use vector I/O.

vchoi-hdfgroup: > Why do you call H5Pget_no_selection_io_cause() here (and in similar places)?

I think I am a bit confused.  I will remove all those changes from the tests in test/select_io_dset.c.

src/H5Pdxpl.c
vchoi-hdfgroup: Fixed.

src/H5Ppublic.h
vchoi-hdfgroup: Fixed

src/H5Ppublic.h
vchoi-hdfgroup: Fixed.

fortnern: We should add a note stating: "Be aware that this function will include the types of all I/O that were performed during this operation, including any metadata operations that may be incidental to the requested I/O. To make sure this function is only capturing raw data I/O, one can temporarily disable metadata cache evictions by calling H5Fset_mdc_config() with evictions_enabled set to false."

fortnern: My intent with the original comment was to add that sentence to the end of the Motivation section. I'll go ahead and make my changes and paste the entire doxygen comment here

fortnern: > /**
 * \ingroup DXPL
 *
 * \brief Retrieves the type(s) of I/O that HDF5 actually performed on
 *        last I/O call (not necessarily the type requested)
 *
 * \dxpl_id{plist_id}
 * \param[out] actual_selection_io_mode A bitwise set value indicating the
 *                                      type(s) of I/O performed
 * \return \herr_t
 *
 * \par Motivation:
 *      A user can request selection I/O to be performed via a data transfer
 *      property list (DXPL).  This can be used to enable collective I/O with
 *      type conversion, or with custom VFDs that support vector or selection
 *      I/O.  However, there are conditions that can cause HDF5 to forgo
 *      selection or vector I/O and perform legacy (scalar) I/O instead.  This
 *      function allows the user to determine which type or types of I/O were
 *      actually performed.
 *
 * \details H5Pget_actual_selection_io_mode() allows the user to determine which
 *          type(s) of I/O were actually performed during the last I/O operation
 *          which used \p plist_id.  This property is set after all I/O is
 *          completed; if I/O fails, it will not be set.
 *
 *          H5Pget_no_selection_io_cause() can be used to determine the reason
 *          why selection or vector I/O was not performed.
 *
 *          Valid values returned in \p actual_selection_io_mode are listed
 *          as follows.
 *
 *          - #H5D_SCALAR_IO
 *          Scalar (or legacy MPIO) I/O was performed
 *          - #H5D_VECTOR_IO
 *          Vector I/O was performed
 *          - #H5D_SELECTION_IO
 *          Selection I/O was performed
 *
 *          Be aware that this function will include the types of all I/O that
 *          were performed during this operation, including any metadata
 *          operations that may be incidental to the requested I/O. To make sure
 *          this function is only capturing raw data I/O, one can temporarily
 *          disable metadata cache evictions by calling H5Fset_mdc_config() with
 *          evictions_enabled set to false.
 *
 * \since 1.14.2
 *
 */

fortnern: I can't get github to avoid parsing the doxygen (even in a code block) so I'll jsut email it to you.

fortnern (approved):

PR #2877, opened 2023-05-02
Add initial version of HDF5 API tests

jhendersonHDF:


test/API/CMake/CheckAsan.cmake
```diff
@@ -0,0 +1,37 @@
+set(ASAN_FLAG "-fsanitize=address")
```
byrnHDF, line 1: Isn't this a duplicate of cmake/sanitizer/sanitizers.cmake?

test/API/CMake/CheckUbsan.cmake
```diff
@@ -0,0 +1,37 @@
+set(UBSAN_FLAG "-fsanitize=undefined")
```
byrnHDF, line 1: This too, it would be better to augment saniitzers.cmake for an special settings based on the HDF5_TEST_API value.

test/API/CMakeLists.txt
```diff
@@ -0,0 +1,314 @@
+# Copyright by The HDF Group.
```
byrnHDF, line 1: It may be debatable, but for consistency we include the following lines at the top of CMakeLists.txt:
cmake_minimum_required (VERSION 3.18)
project (HDF5_XXXX C)

test/API/CMake/CheckAsan.cmake
jhendersonHDF: More or less, yes. Asan and Ubsan were used in the external VOL tests repo during testing.

test/API/CMakeLists.txt
```diff
@@ -0,0 +1,314 @@
...
+# the COPYING file, which can be found at the root of the source code
+# distribution tree, or in https://www.hdfgroup.org/licenses.
+# If you do not have access to either file, you may request a copy from
+# help@hdfgroup.org.
+#
+
+#------------------------------------------------------------------------------
+# Set module path
+#------------------------------------------------------------------------------
+set(HDF5_TEST_API_CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake")
```
byrnHDF, line 15: These would go away if the those macros were merged with sanitizers.cmake

test/API/CMakeLists.txt
```diff
@@ -0,0 +1,314 @@
...
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${HDF5_TEST_API_CMAKE_MODULE_PATH})
+
+# TODO: probably not necessary
+#------------------------------------------------------------------------------
+# Setup CMake Environment
+#------------------------------------------------------------------------------
+if (WIN32)
+  message("The HDF5 API test suite is currently not supported on this platform." FATAL_ERROR)
```
byrnHDF, line 23: Shouldn't this be in root CMakeLists.txt so that the option could be disabled instead of halting the build?

test/API/CMakeLists.txt
jhendersonHDF: The VOL tests used to be a separate project and had its own set of these (though note the minimum there was only 2.8 I think), but as part of the library this would just fall under the umbrella of test/CMakeLists. It could be a separate project, but it doesn't seem like that really buys us anything.

test/API/CMakeLists.txt
```diff
@@ -0,0 +1,314 @@
...
+endif ()
+
+#------------------------------------------------------------------------------
+# Setup testing configuration file
+#------------------------------------------------------------------------------
+if (HDF5_TEST_PARALLEL)
```
byrnHDF, line 29: Why the duplication?

test/API/CMakeLists.txt
```diff
@@ -0,0 +1,314 @@
...
+  set (HDF5_TEST_API_HAVE_PARALLEL 1)
+endif ()
+if (HDF5_TEST_API_ENABLE_ASYNC)
+  set (H5_API_TEST_HAVE_ASYNC 1)
+endif ()
+
+configure_file(
```
byrnHDF, line 36: @ONLY not needed?

test/API/CMakeLists.txt
```diff
@@ -0,0 +1,314 @@
...
+#------------------------------------------------------------------------------
+
+# Ported HDF5 tests
+set (HDF5_API_TESTS_EXTRA
+  testhdf5
+)
+
+# List of files generated by the HDF5 API tests which
+# should be cleaned up in case the test failed to remove
+# them
```
byrnHDF, line 59: We usually separate test handling into a CMakeTests.cmake file that is included

test/API/README.md
```diff
@@ -0,0 +1,86 @@
+# HDF5 API Tests
+
+This directory contains several test applications that exercise [HDF5](https://github.com/HDFGroup/hdf5)'s
+public API and serve as regression tests for HDF5 [VOL Connectors](https://portal.hdfgroup.org/display/HDF5/Virtual+Object+Layer).
+
+## Build Process and options
+
+These HDF5 API tests are enabled and built by default, but can be disabled if desired.
```
byrnHDF, line 8: disabled by default?

test/API/README.md
```diff
@@ -0,0 +1,86 @@
...
+option should be edited to point to the server executable that the driver program should attempt
+to launch before running the API tests.
+
+### Autotools
+
+Currently unsupported
+
+### Usage
+
+These API tests currently only support usage with HDF5 VOL connectors that can be loaded dynamically
```
byrnHDF, line 42: So does HDF5_TEST_API depend on a VOL optopn being enabled? Then that option should guard HDF5_TEST_API in root CMakeLists.txt

test/API/driver/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 2.8.12.2 FATAL_ERROR)
```
byrnHDF, line 1: minimum is 3.18 now

test/API/driver/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+project(H5_API_TEST_DRIVER CXX)
+
+include(CheckAsan)
+include(CheckUbsan)
+
+set(CMAKE_CXX_STANDARD 11)
+
+set(KWSYS_NAMESPACE h5_api_test_sys)
```
byrnHDF, line 9: kwsys!  This should be in a FetchContent call not copied into here.

test/API/driver/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 2.8.12.2 FATAL_ERROR)
+project(H5_API_TEST_DRIVER CXX)
```
byrnHDF, line 2: This has a language requirement for C++ -  another guard for the root/CMakeLists.txt option.

test/CMakeLists.txt
```diff
@@ -395,6 +395,24 @@ if (HDF5_BUILD_UTILS)
   set (H5_TESTS ${H5_TESTS} mirror_vfd)
 endif ()
 
+set (HDF5_API_TESTS
```
byrnHDF, line 398: This should be in a separate CMakeAPITests.cmake file that can be guarded by the option value.

testpar/API/CMakeLists.txt
```diff
@@ -0,0 +1,279 @@
+# Copyright by The HDF Group.
```
byrnHDF, line 1: Same issues as test/API file

byrnHDF (changes requested):
Changes needed after merge.

test/API/README.md
jhendersonHDF: There's going to be some weirdness because the intention was that they would be enabled by default. However, since this is close to a release, they've been temporarily disabled by default.

test/API/CMakeLists.txt
jhendersonHDF: Yes though at this point I don't know that the check is needed in the first place. There shouldn't be an issue with building and running these test on Windows, though it hasn't been tested.

test/API/CMakeLists.txt
jhendersonHDF: Duplication? This is just adding parallel-related defines when configuring the single configuration file used by the tests

test/API/CMakeLists.txt
jhendersonHDF: Probably not, but it shouldn't hurt to add in this case

test/API/CMakeLists.txt
jhendersonHDF: Personally I feel that just makes it harder to find things, but these can be separate that way as well

test/API/driver/CMakeLists.txt
jhendersonHDF: Right but seeing as this is Kitware's system library, our minimum doesn't exactly apply across everything. Though as noted below FetchContent could be used rather than having the driver in place. It was a topic of discussion that still needed to take place because it was unclear how to handle the test driver code.

test/CMakeLists.txt
jhendersonHDF: I would think it would be just as simple to guard the option here rather than creating yet another file just for these tests. These tests aren't particularly special and should be run with the native VOL connector as well.

test/API/README.md
jhendersonHDF: An issue of documentation with the difference between being in the library after being external. These tests should also get run against the native VOL connector, but their primary purpose is being available for use with external VOL connectors.

CMakeLists.txt
```diff
@@ -958,6 +961,25 @@ if (BUILD_TESTING)
   math (EXPR CTEST_LONG_TIMEOUT "${DART_TESTING_TIMEOUT} * 2")
   math (EXPR CTEST_VERY_LONG_TIMEOUT "${DART_TESTING_TIMEOUT} * 3")
 
+  option (HDF5_TEST_API "Execute HDF5 API tests" OFF)
```
brtnfld, line 964: You should be more explicit about this being for VOL developers (experts, advanced users), etc...

byrnHDF (approved):

derobins (approved):

PR #2878, opened 2023-05-02
Minor tidying of API tests files

jhendersonHDF:
Pick off a few of the small suggested changes


test/API/CMakeLists.txt
```diff
@@ -9,20 +9,6 @@
...
-#------------------------------------------------------------------------------
-set(HDF5_TEST_API_CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake")
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${HDF5_TEST_API_CMAKE_MODULE_PATH})
-
-# TODO: probably not necessary
-#------------------------------------------------------------------------------
-# Setup CMake Environment
-#------------------------------------------------------------------------------
-if (WIN32)
-  message("The HDF5 API test suite is currently not supported on this platform." FATAL_ERROR)
```
jhendersonHDF: There shouldn't be any reason that these tests can't be built and run on Windows with the native VOL connector. Support for other VOL connectors on Windows is a different matter.

byrnHDF: if (HDF5_TEST_PARALLEL)
  set (HDF5_TEST_API_HAVE_PARALLEL 1)
endif ()

I can't find any other us of this HDF5_TEST_API_HAVE_PARALLEL variable.

test/API/CMakeLists.txt
```diff
@@ -9,20 +9,6 @@
 # help@hdfgroup.org.
 #
 
-#------------------------------------------------------------------------------
-# Set module path
```
byrnHDF: I can take care of these changes in this file

byrnHDF (approved):

test/API/CMakeLists.txt
jhendersonHDF: Ah yes, that should have been removed when I switched over to checking for H5_HAVE_PARALLEL instead.

lrknox (approved):

PR #2899, opened 2023-05-05
Remove H5TB discussion

derobins:


qkoziol (approved):

gheber:
This particular instance is probably a flat liner, but I would like some *context* in a wider discussion, starting w/ the forum, on the future, or the lack thereof, of high-level interfaces.

byrnHDF (approved):

derobins:
> This particular instance is probably a flat liner, but I would like some _context_ in a wider discussion, starting w/ the forum, on the future, or the lack thereof, of high-level interfaces.

I'm not sure how to interpret that comment in this context. This is just an obsolete comment about how the metadata cache was implemented. H5TB was an old package that was removed a long time ago. I just tripped over it while working on a python script to automagically clean up our include statement blocks and removed it since it's confusing and unhelpful to anyone looking over the code.

https://github.com/HDFGroup/hdf5/blob/hdf5-1_2_0/src/H5TB.c

Issue #2906, opened 2023-05-05
Incorrect cmake_minimum_required for Fortran support


Mittagskogel:
`cmake_minimum_required` in HDF5 1.13.3 and above should be at least 3.23. 
When building with older CMake, a CMake generator expression causes extraneous `$<1:` and `>` in paths, leading to errors such as the following:

```
  >> 2070    make[2]: *** No rule to make target '/lib64/libpthread.so>', needed by 'bin/libhdf5_fortran.so.310.0.0'.  Stop.
```
or
```
  >> 2977    /usr/bin/ld: cannot find $<1:/home/user/spack/opt/spack/linux-rocky8-zen3/gcc-13.1.0/intel-oneapi-mpi-2021.9.0-p4vdumpdjsjhxb67uy4lyd6iradfmz7z/mpi/2021.9.0/lib/libmpifort.so: No such file or directory
```

This issue seems to occur only when building with Fortran support enabled.

In short: Building HDF5 with Fortran support, versions 1.13.3 and above with CMake, versions 3.22 and below, causes the above errors. `cmake_minimum_required (VERSION 3.18)` is incorrect.

byrnHDF:
I need a bit more context (Note that 1.13.x are not supported since the release of 1.14.0). 
I assume linux (rocky 8), compiler gcc 13.1.0, but what options other then fortran and was this a build of hdf5 or a link to an installed hdf5.

Mittagskogel:
1.14.0 (and above) have the same problem.
This is a build of hdf5 done with Spack. Linux (rocky 8), and multiple compilers.

wdconinc:
Possibly due to lack of quoting in https://github.com/HDFGroup/hdf5/blob/develop/fortran/src/CMakeLists.txt#L375, see https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#whitespace-and-quoting where the first example of what can go wrong seems to be exactly that when `MPI_Fortran_LIBRARIES` is expanded and has a space between libraries.

byrnHDF:
**Possibly due to lack of quoting**
Thanks, that helped - Adding testing of minimum CMake reqs is on our todo list as we have retired some older test machines that should have caught this.

wdconinc:
> Possibly due to lack of quoting

Just want to point that's just a hunch, and still untested (but I'll let you know if it fixes it for me).

byrnHDF:
> > Possibly due to lack of quoting
> 
[...]

PR #2909, opened 2023-05-05
Fix CMake generator expression syntax

byrnHDF:


byrnHDF:
Addresses #2906

byrnHDF:
see https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#whitespace-and-quoting

derobins (approved):

jhendersonHDF (approved):

PR #2914, opened 2023-05-05
Reduce code duplication in macros

qkoziol:
Reduce copies of the macros and create generic versions for all lists

### DIFF
--- a/src/H5Cpkg.h
+++ b/src/H5Cpkg.h
@@ -13,16 +13,9 @@
 /*
  * Programmer: John Mainzer -- 10/12/04
  *
- * Purpose:     This file contains declarations which are normally visible
- *              only within the H5C package.
- *
- *        Source files outside the H5C package should include
- *        H5Cprivate.h instead.
- *
- *        The one exception to this rule is test/cache.c.  The test
- *        code is easier to write if it can look at the cache's
- *        internal data structures.  Indeed, this is the main
- *        reason why this file was created.
+ * Purpose:	This file contains declarations which are visible only within
+ *		the H5C package.  Source files outside the H5C package should
+ *		include H5Cprivate.h instead.
  */
 
 /* clang-format off */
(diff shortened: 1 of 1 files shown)


src/H5Cpkg.h
```diff
@@ -82,424 +75,167 @@
...
+    ((entry_ptr)->list_next == NULL && (tail_ptr) != (entry_ptr)) ||        \
+    ((len) == 1 &&                                                           \
+     !((head_ptr) == (entry_ptr) && (tail_ptr) == (entry_ptr) &&             \
+       (entry_ptr)->list_next == NULL && (entry_ptr)->list_prev == NULL && \
+       (list_size) == (entry_ptr)->size                                      \
+      )                                                                      \
+    )                                                                        \
    ) {                                                                       \
-    HGOTO_ERROR(H5E_CACHE, H5E_SYSTEM, (fail_val), "DLL pre remove SC failed")     \
+    HGOTO_ERROR(H5E_CACHE, H5E_SYSTEM, (fail_val), "DLL pre insert SC failed")               \
```
derobins: The macro says REMOVE but the error message says "insert"

derobins (approved):
Need to change that error string, though

src/H5Cpkg.h
qkoziol: Done

hyoklee (approved):

src/H5Cpkg.h
```diff
@@ -82,424 +75,167 @@
...
-       )                                                                 \
-     ) ||                                                                \
-     ( ( (len) >= 1 ) &&                                                 \
-       ( ( (head_ptr) == NULL ) || ( (head_ptr)->prev != NULL ) ||       \
-         ( (tail_ptr) == NULL ) || ( (tail_ptr)->next != NULL )          \
-       )                                                                 \
-     )                                                                   \
-   ) {                                                                   \
-    HGOTO_ERROR(H5E_CACHE, H5E_SYSTEM, (fail_val), "DLL sanity check failed")  \
+#define H5C__GEN_DLL_SC(list_next, list_prev, head_ptr, tail_ptr, len, list_size, fail_val) \
```
fortnern: This macro doesn't seem to be used for anything

qkoziol: Ah, interesting...    I looked in older versions of the file and it doesn't appear used in the ones I've looked at either.   I'll remove it.

src/H5Cpkg.h
```diff
@@ -82,424 +75,151 @@
...
-     ( ( (len) == 1 ) &&                                                     \
-       ( ! ( ( (head_ptr) == (entry_ptr) ) &&                                \
-             ( (tail_ptr) == (entry_ptr) ) &&                                \
-             ( (entry_ptr)->next == NULL ) &&                                \
-             ( (entry_ptr)->prev == NULL ) &&                                \
-             ( (list_size) == (entry_ptr)->size )                                 \
-           )                                                                 \
-       )                                                                     \
-     )                                                                       \
+#define H5C__GEN_DLL_PRE_REMOVE_SC(entry_ptr, list_next, list_prev, head_ptr, tail_ptr, len, list_size, fail_val) \
```
jhendersonHDF, line 78: It doesn't look like the new list_next and list_prev params are used in this macro. Were they supposed to replace entry_ptr->list_next/list_prev?

qkoziol: :-?

qkoziol: They look correct to me

qkoziol: I think I understand what you are saying - the way they are used: (entry_ptr)->list_next/prev is the required syntax to get the substitution correct by the compiler.

jhendersonHDF: Ah ok I see what you did now. Seeing only entry_ptr in parenthesis was a little misleading to me. Didn't realize the field accessed through entry_ptr is also genericized through the list_next and list_prev params.

qkoziol: Yeah, I wanted the put list_next/prev in parens also, but the compiler barfs on that

jhendersonHDF (approved):

Issue #2926, opened 2023-05-07
Change CMake defaults for enabling szip and gzip compression


byrnHDF:
It is requested that the options should be on by default.

byrnHDF:
Currently, if the filter is enabled and not found, the build will halt. Should this be changed?

jhendersonHDF:
> Currently, if the filter is enabled and not found, the build will halt. Should this be changed?

Yes, AFAIK Autotools will skip building the filter if support for it can't be found so CMake should do the same.

byrnHDF:
This has been addressed in all branches.

PR #2927, opened 2023-05-07
Change defaults of standard filters.

byrnHDF:
This addresses issue #2926


lrknox:
the default for szip is disabled in autotools.  Do you want to match that?

lrknox (approved):

jhendersonHDF (approved):

.github/workflows/main.yml
```diff
@@ -496,7 +496,7 @@ jobs:
...
+          cmake -C $GITHUB_WORKSPACE/config/cmake/cacheinit.cmake ${{ matrix.generator }} -DCMAKE_BUILD_TYPE=${{ matrix.build_mode.cmake }} -DCMAKE_TOOLCHAIN_FILE=${{ matrix.toolchain }} -DBUILD_SHARED_LIBS=ON -DHDF5_ENABLE_ALL_WARNINGS=ON -DHDF5_ENABLE_PARALLEL:BOOL=${{ matrix.parallel }} -DHDF5_BUILD_CPP_LIB:BOOL=${{ matrix.cpp }} -DHDF5_BUILD_FORTRAN=${{ matrix.fortran }} -DHDF5_BUILD_JAVA=${{ matrix.java }} -DBUILD_SZIP_WITH_FETCHCONTENT=${{ matrix.libaecfc }} -DBUILD_ZLIB_WITH_FETCHCONTENT=${{ matrix.zlibfc }} -DHDF5_ENABLE_MIRROR_VFD:BOOL=${{ matrix.mirror_vfd }} -DHDF5_ENABLE_DIRECT_VFD:BOOL=${{ matrix.direct_vfd }} $GITHUB_WORKSPACE
```
jhendersonHDF, line 499: I wonder if it wouldn't be better here to have a separate build that uses a cacheinit file so that this doesn't paper over potential issues by things that get set in the cacheinit file that aren't set by default.

byrnHDF: The other choice is to move the paths/names for compression into the CMakeLists/CMakeFilters files.

PR #2930, opened 2023-05-08
More code duplication reduction

qkoziol:
Add failure value where it's missing from 1+ macros.  Clean up whitespace / continuation characters ('\').  Made hash-table macros generic for use in both the package header and test header.  Remove duplicated copy & pasted macros (by hoisting difference into #ifdef'd macro). Updated and re-flowed comments to read better.

Also clean up a few compiler warnings in production builds.

### DIFF
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -3435,7 +3435,7 @@ H5C_create_flush_dependency(void *parent_thing, void *child_thing)
 
         for (u = 0; u < child_entry->flush_dep_nparents; u++)
             HDassert(child_entry->flush_dep_parent[u] != parent_entry);
-    }  /* end block */
+    } /* end block */
 #endif
 
     /* More sanity checks */
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -1500,7 +1500,7 @@ H5C_mark_entry_clean(void *_thing)
         if (was_dirty)
             H5C__UPDATE_INDEX_FOR_ENTRY_CLEAN(cache_ptr, entry_ptr, FAIL)
         if (entry_ptr->in_slist)
-            H5C__REMOVE_ENTRY_FROM_SLIST(cache_ptr, entry_ptr, FALSE)
+            H5C__REMOVE_ENTRY_FROM_SLIST(cache_ptr, entry_ptr, FALSE, FAIL)
 
         /* Update stats for entry being marked clean */
         H5C__UPDATE_STATS_FOR_CLEAR(cache_ptr, entry_ptr)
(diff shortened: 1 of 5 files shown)


hyoklee (approved):

src/H5C.c
```diff
@@ -6579,8 +6579,10 @@ H5C__make_space_in_cache(H5F_t *f, size_t space_needed, hbool_t write_permitted)
     H5C_cache_entry_t *entry_ptr;
     H5C_cache_entry_t *prev_ptr;
     H5C_cache_entry_t *next_ptr;
-    uint32_t           num_corked_entries = 0;
-    herr_t             ret_value          = SUCCEED; /* Return value */
+#ifndef NDEBUG
+    uint32_t num_corked_entries = 0;
+#endif                          /* NDEBUG */
```
derobins: We should probably try to use H5C_DEBUG instead of NDEBUG for package-level debugging, as we do elsewhere in the library

src/H5Cimage.c
```diff
@@ -2650,10 +2662,12 @@ static herr_t
...
 {
     H5C_cache_entry_t *entry_ptr;
-    H5C_image_entry_t *image_entries   = NULL;
-    uint32_t           entries_visited = 0;
-    unsigned           u;                   /* Local index variable */
-    herr_t             ret_value = SUCCEED; /* Return value */
+    H5C_image_entry_t *image_entries = NULL;
+#ifndef NDEBUG
+    uint32_t entries_visited = 0;
+#endif                            /* NDEBUG */
```
derobins: Also, tail comments for ifdefs that are one line above just turn into malformed noise, as is the case here. They are really only useful for stuff that rolls off the screen.

qkoziol: OK, I'll update those

src/H5C.c
qkoziol: Agree - I'll add it to my TODO list

src/H5Cpkg.h
```diff
@@ -3409,227 +2721,196 @@ typedef struct H5C_tag_info_t {
...
+ * total_failed_ht_search_depth: The total number of entries examined in
+ *        unsuccessful searches of the hash table in the current epoch.
+ *
+ * max_index_len: Largest value attained by the index_len field in the
+ *        current epoch.
  *
- * max_index_size:  Largest value attained by the index_size field in the
- *              current epoch.
+ * max_index_size: Largest value attained by the index_size field in the
+ *        current epoch.
```
derobins, line 2839: It'd be nice if all this could be made to line up

derobins (approved):

src/H5Cpkg.h
qkoziol: > It'd be nice if all this could be made to line up

Agree, but I don't want to hold this PR on that.  I'll try to polish them over the next few PRs

jhendersonHDF (approved):

Issue #2680, opened 2023-04-06
Potential buffer overrun issues with H5S_SELECT_DESERIALIZE


jhendersonHDF:
**Describe the bug**
H5S_SELECT_DESERIALIZE/H5S_select_deserialize accepts a parameter for a buffer to decode a dataspace selection into, but doesn't accept a parameter to provide the size of that buffer. Therefore, these buffers can be overrun during decoding of the dataspace selection if the buffers aren't big enough to contain the decoded selection. A malformed HDF5 file could cause this to happen for example.

**Expected behavior**
H5S_select_deserialize should not decode into buffers that aren't large enough to contain the dataspace selection being decoded.

**Platform**
 - Develop (526f0769a3353b20d1265c09b3035fa4c2610933)
 - Slackware Linux (Kernel 6.3.0-rc2)
 - GCC 12.2.0
 - CMake / Autotools / N/A
 - N/A
 - N/A

**Additional context**
H5S_select_deserialize needs to be updated to accept a parameter that specifies the size of the buffer being passed in.

PR #2931, opened 2023-05-08

mattjala:
Fix for #2680.

The call to `H5S_select_deserialize` from `H5S_decode` doesn't have the buffer size available to it, so to allow decoding there I set it to assume a max size buffer for now.

Making the buffer size known in `H5S_decode` could be done by modifying the external API's `H5Sdecode`, or splitting `H5Sdecode` into two functions using a macro (similar to `H5Sencode`), with the macro taking one argument and assuming a max buffer size.

### DIFF
--- a/src/H5Olayout.c
+++ b/src/H5Olayout.c
@@ -635,12 +635,12 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
 
                         /* Source selection */
                         if (H5S_SELECT_DESERIALIZE(&mesg->storage.u.virt.list[i].source_select,
-                                                   &heap_block_p) < 0)
+                                                   &heap_block_p, heap_block_p_end - heap_block_p + 1) < 0)
                             HGOTO_ERROR(H5E_OHDR, H5E_CANTDECODE, NULL, "can't decode source space selection")
 
                         /* Virtual selection */
                         if (H5S_SELECT_DESERIALIZE(&mesg->storage.u.virt.list[i].source_dset.virtual_select,
-                                                   &heap_block_p) < 0)
+                                                   &heap_block_p, heap_block_p_end - heap_block_p + 1) < 0)
                             HGOTO_ERROR(H5E_OHDR, H5E_CANTDECODE, NULL,
                                         "can't decode virtual space selection")
 
--- a/src/H5Olayout.c
+++ b/src/H5Olayout.c
@@ -634,13 +634,26 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
                         heap_block_p += tmp_size;
 
                         /* Source selection */
+                        avail_buffer_space = heap_block_p_end - heap_block_p + 1;
+
+                        if (avail_buffer_space <= 0)
+                            HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, NULL, "buffer overflow while decoding layout")
+
+
                         if (H5S_SELECT_DESERIALIZE(&mesg->storage.u.virt.list[i].source_select,
-                                                   &heap_block_p, heap_block_p_end - heap_block_p + 1) < 0)
+                                                   &heap_block_p, (size_t) (avail_buffer_space)) < 0)
                             HGOTO_ERROR(H5E_OHDR, H5E_CANTDECODE, NULL, "can't decode source space selection")
 
                         /* Virtual selection */
+
+                        /* Buffer space must be updated after previous deserialization */
+                        avail_buffer_space = heap_block_p_end - heap_block_p + 1;
+                        
+                         if (avail_buffer_space <= 0)
+                            HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, NULL, "buffer overflow while decoding layout")
+
                         if (H5S_SELECT_DESERIALIZE(&mesg->storage.u.virt.list[i].source_dset.virtual_select,
-                                                   &heap_block_p, heap_block_p_end - heap_block_p + 1) < 0)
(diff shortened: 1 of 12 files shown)


mattjala:
Now uses the macro Dana defined in #2797, after renaming it and moving it to `H5private.h`

src/H5Olayout.c
```diff
@@ -634,13 +634,22 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
                         heap_block_p += tmp_size;
 
                         /* Source selection */
-                        if (H5S_SELECT_DESERIALIZE(&mesg->storage.u.virt.list[i].source_select,
-                                                   &heap_block_p) < 0)
+                        if (heap_block_p_end - heap_block_p + 1 <= 0)
```
jhendersonHDF: Suggest doing the math here just once and storing it in a `ptrdiff_t` variable just for future maintenance reasons.

src/H5S.c
```diff
@@ -1655,9 +1655,10 @@ H5S_decode(const unsigned char **p)
     if (H5S_select_all(ds, FALSE) < 0)
         HGOTO_ERROR(H5E_DATASPACE, H5E_CANTSET, NULL, "unable to set all selection")
 
-    /* Decode the select part of dataspace.  I believe this part always exists. */
+    /* Decode the select part of dataspace.  I believe this part always exists.
+     *  Because size of buffer is unknown, just assume arbitrarily large buffer to allow decoding. */
```
jhendersonHDF: Just a nitpick: I'm guessing at this point we can probably cleanup the part of the comment about this part always existing.

src/H5Spoint.c
```diff
@@ -1444,14 +1465,26 @@ H5S__point_deserialize(H5S_t **space, const uint8_t **p)
         for (j = 0; j < (unsigned)rank; j++, tcoord++)
             switch (enc_size) {
                 case H5S_SELECT_INFO_ENC_SIZE_2:
+                    if (H5_IS_KNOWN_BUFFER_OVERFLOW(skip, pp, sizeof(uint16_t), p_end))
+                        HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, FAIL,
+                                    "buffer overflow while decoding selection coordinates")
```
jhendersonHDF: For most other cases I don't think these checks should have added much in terms of overhead, but for this specific case of point selections, I might lean more toward calculating the total buffer size needed here based on the combination of `num_elem`, `rank` and `enc_size` and then just performing a single check. Otherwise, large point selections might incur a noticeable overhead during decoding.

src/H5Rint.c
```diff
@@ -1209,7 +1209,7 @@ H5R__decode_region(const unsigned char *buf, size_t *nbytes, H5S_t **space_ptr)
         HGOTO_ERROR(H5E_REFERENCE, H5E_CANTDECODE, FAIL, "Buffer size is too small")
     if (H5S_set_extent_simple(space, rank, NULL, NULL) < 0)
         HGOTO_ERROR(H5E_REFERENCE, H5E_CANTSET, FAIL, "can't set extent rank for selection")
-    if (H5S_SELECT_DESERIALIZE(&space, &p) < 0)
+    if (H5S_SELECT_DESERIALIZE(&space, &p, *nbytes - 2 * sizeof(uint32_t)) < 0)
```
jhendersonHDF: I'd suggest calculating an end_ptr in this function and then making the buffer size calculation here based on that rather than manual subtraction of sizes so that any future revisions to the file format don't miss updating this calculation.

src/H5Rint.c
```diff
@@ -1508,7 +1508,7 @@ H5R__decode_token_region_compat(H5F_t *f, const unsigned char *buf, size_t *nbyt
             HGOTO_ERROR(H5E_REFERENCE, H5E_NOTFOUND, FAIL, "not found")
 
         /* Unserialize the selection */
-        if (H5S_SELECT_DESERIALIZE(&space, &p) < 0)
+        if (H5S_SELECT_DESERIALIZE(&space, &p, data_size - token_size) < 0)
```
jhendersonHDF: Same comment here about using an end_ptr to do the calculation.

src/H5private.h
```diff
@@ -327,6 +327,15 @@
...
 #define H5_IS_BUFFER_OVERFLOW(ptr, size, buffer_end) (((ptr) + (size)-1) > (buffer_end))
 
+/* Variant of H5_IS_BUFFER_OVERFLOW, used with functions such as H5Tdecode()
+ * that don't take a size parameter, where we need to skip the bounds checks.
+ *
+ * This is a separate macro since we don't want to inflict that behavior on
+ * the entire library.
+ */
+#define H5_IS_KNOWN_BUFFER_OVERFLOW(skip, ptr, size, buffer_end)                                             \
+    (skip ? FALSE : ((ptr) + (size)-1) > (buffer_end))
```
derobins, line 337: I'm not a fan of this name but I can't think of a great one. We can find-n-replace later.

derobins (approved):

src/H5Olayout.c
```diff
@@ -634,13 +634,28 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
...
+                        avail_buffer_space = heap_block_p_end - heap_block_p + 1;
+
+                        if (avail_buffer_space <= 0)
+                            HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, NULL,
+                                        "buffer overflow while decoding layout")
+
                         if (H5S_SELECT_DESERIALIZE(&mesg->storage.u.virt.list[i].source_dset.virtual_select,
-                                                   &heap_block_p) < 0)
+                                                   &heap_block_p,
+                                                   (size_t)(heap_block_p_end - heap_block_p + 1)) < 0)
```
jhendersonHDF: Use `avail_buffer_space` var. here

jhendersonHDF (approved):
Just a minor last cleanup

PR #2933, opened 2023-05-09
Update t_bigio.c

jeanbez:
Fix failed test in Perlmutter. This fixed is already in the develop branch.


derobins (changes requested):
Let's find a better solution for this issue than re-introducing a bug

PR #2940, opened 2023-05-10
Fix CMake generator expression syntax

byrnHDF:


bmribler (approved):

bmribler (approved):

CMakePresets.json
```diff
@@ -182,7 +182,12 @@
       "configurePreset": "ci-StdShar-MSVC",
       "inherits": [
         "ci-x64-Release-MSVC"
-      ]
+      ],
+      "filter": {
+        "exclude": {
+          "name": "H5DUMP-tfloatsattrs"
```
jhendersonHDF, line 188: As an aside, is this test excluded due to floating point error differences between testing on Linux and Windows or due to some other reason?

jhendersonHDF (approved):

CMakePresets.json
byrnHDF: Correct.

jhendersonHDF: Interesting - the H5Z ZFP tests run into the same issue with floating point accuracy on Windows. Would need to do some research to figure out if the results can be made close enough to identical across platforms.

byrnHDF: This test just runs into rounding errors - ZFP I think has bigger differences

PR #2944, opened 2023-05-11
Bump Autoconf version to 2.71

derobins:
Required for building with Intel's oneAPI

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@
 ## ----------------------------------------------------------------------
 ## Initialize configure.
 ##
-AC_PREREQ([2.69])
+AC_PREREQ([2.71])
 
 ## AC_INIT takes the name of the package, the version number, and an
 ## email address to report bugs. AC_CONFIG_SRCDIR takes a unique file


byrnHDF (approved):

derobins:
I feel like we are at the point where we *have* to merge this in. If we don't, we risk breaking oneAPI unless we are SUPER careful (and I'm not a fan of fragile manual steps).

If anyone has any other ideas about how we can manage the tension between "need to build HDF5 on older machines" and "need to support oneAPI", I'm all ears.

jhendersonHDF (approved):

jhendersonHDF:
> I feel like we are at the point where we _have_ to merge this in. If we don't, we risk breaking oneAPI unless we are SUPER careful (and I'm not a fan of fragile manual steps).
> 
> If anyone has any other ideas about how we can manage the tension between "need to build HDF5 on older machines" and "need to support oneAPI", I'm all ears.

Considering that Slackware moved to 2.71 back in Feb. of last year, I think it's a safe bet at this point that most others will have moved along too.

brtnfld (approved):
Plus, if a user really wants to regenerate using 2.69, it is a one-line change since there are no 2.71 nuances.

Although, a search on GitHub reported ~12k using 2.69 compared to 760 using 2.71

qkoziol (approved):

hyoklee:
Where is oneAPI Action to test?

derobins:
> Where is oneAPI Action to test?

That's going to take some time to set up, but it's on the agenda. One thing at a time.

hyoklee:
configure fails with `--enable-parallel`.

PR #2911, opened 2023-05-05
ci: add spack action

hyoklee:

PR #2946, opened 2023-05-11
ci(Spack): add Ubuntu action

hyoklee:
Redo #2911 using my fork branch per @derobins request.
Address @brtnfld review.

### DIFF
--- a/.github/workflows/spack_u_dev.yml
+++ b/.github/workflows/spack_u_dev.yml
@@ -1,4 +1,4 @@
-name: Spack Ubuntu hdf5@develop+mpi+fortran
+name: Spack hdf5@develop-1.15+mpi+fortran
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


derobins (changes requested):
In the standup, we recommended modifying this to use the system resources instead of installing via spack, to get the time of this test down.

hyoklee:
I won't change because it defeats the purpose of Spack testing (i.e., builds everything from source for HPC) if you use the system resource.
Time doesn't matter since Actions are faster than HDF5 PR merge frequency.

brtnfld:
A couple of things. You should use the system's mpi compiler without having spack install it. The HPC facilities will always use the system's mpi and system compilers. So, as you are CI testing on Sunspot, Frontier, etc., make sure you are using the system compilers, not a spack-installed package. Secondly, we want to know if there are issues with system-supplied packages in our CI testing. We need to test these cases, in addition to, but run less frequently, the case where spack builds everything.

hyoklee:
@brtnfld , it's irrelevant. I don't use this Action for HPC testing.

brtnfld:
What use case are you testing, then? Most spack users try to use system resources when available.

hyoklee:
This is a quick CI check if HDF5 develop branch accidentally breaks `spack install hdf5@develop` from Spack develop branch.

derobins:
We're going to take a different strategy when it comes to Spack testing

Issue #92, opened 2020-11-13
Alert users to not change collections in user callbacks while iterating


gheber:
`H5Iiterate` is a case in point, but all modules should be reviewed.

hyoklee:
https://docs.hdfgroup.org/hdf5/develop/group___h5_i_u_d.html#ga3bc2a11a594bf484cc5dc69ac987d0f4

hyoklee:
Update H5Ipublic.h

hyoklee:
@gheber , what do you mean by "collections" in this issue?

gheber:
Mostly groups (= collections of links), attribute collections of objects, etc. I.e., don't change a group that you are iterating. OK?

hyoklee:
I still don't understand. Please assign this to someone who wrote this function or is familiar with.

PR #2950, opened 2023-05-12
Warn about changing collections during iteration 

mattjala:
Resolution to #92

### DIFF
--- a/src/H5Dpublic.h
+++ b/src/H5Dpublic.h
@@ -1286,6 +1286,8 @@ H5_DLL herr_t H5Dread_chunk(hid_t dset_id, hid_t dxpl_id, const hsize_t *offset,
  *            be restarted at the point of exit; a second H5Diterate()
  *            call will always restart at the beginning.
  *
+ * \warning   Adding or removing elements to the dataspace during iteration
+ *            will lead to undefined behavior. 
  *
  * \since 1.10.2
  *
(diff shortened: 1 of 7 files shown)


derobins (approved):

jhendersonHDF (approved):

src/H5Dpublic.h
```diff
@@ -1286,6 +1286,8 @@ H5_DLL herr_t H5Dread_chunk(hid_t dset_id, hid_t dxpl_id, const hsize_t *offset,
  *            be restarted at the point of exit; a second H5Diterate()
  *            call will always restart at the beginning.
  *
+ * \warning   Adding or removing elements to the dataspace during iteration
```
fortnern: Maybe this should be worded "Modifying the selection of \p space_id during iteration will lead to undefined behavior." or something like that

src/H5Gpublic.h
```diff
@@ -1002,9 +1002,13 @@ H5_DLL int H5Gget_comment(hid_t loc_id, const char *name, size_t bufsize, char *
...
  *          required, the application must handle the recursion, explicitly
  *          calling H5Giterate() on discovered subgroups.
-
+ *
+ * \warning  Adding or removing members to the group during iteration
+ *           will lead to undefined behavior.
+ *
  * \version 1.8.0 Function deprecated in this release.
  *
+ *
```
fortnern: Unnecessary line

PR #2951, opened 2023-05-12
Remove unnecessary fields from cache structs

qkoziol:
Remove unnecessary 'magic' fields from cache structs

### DIFF
--- a/src/H5Cdbg.c
+++ b/src/H5Cdbg.c
@@ -86,7 +86,6 @@ H5C_dump_cache(H5C_t *cache_ptr, const char *cache_name)
 
     /* Sanity check */
     HDassert(cache_ptr != NULL);
-    HDassert(cache_ptr->magic == H5C__H5C_T_MAGIC);
     HDassert(cache_name != NULL);
 
     /* First, create a skip list */
@@ -101,7 +100,6 @@ H5C_dump_cache(H5C_t *cache_ptr, const char *cache_name)
         entry_ptr = cache_ptr->index[i];
 
         while (entry_ptr != NULL) {
-            HDassert(entry_ptr->magic == H5C__H5C_CACHE_ENTRY_T_MAGIC);
             if (H5SL_insert(slist_ptr, entry_ptr, &(entry_ptr->addr)) < 0)
                 HGOTO_ERROR(H5E_CACHE, H5E_BADVALUE, FAIL, "can't insert entry in skip list")
 
@@ -131,8 +129,6 @@ H5C_dump_cache(H5C_t *cache_ptr, const char *cache_name)
     i         = 0;
     entry_ptr = (H5C_cache_entry_t *)H5SL_remove_first(slist_ptr);
     while (entry_ptr != NULL) {
-        HDassert(entry_ptr->magic == H5C__H5C_CACHE_ENTRY_T_MAGIC);
-
         /* Print entry */
         HDfprintf(stdout, "%s%5d ", cache_ptr->prefix, i);
         HDfprintf(stdout, "  0x%16llx ", (long long)(entry_ptr->addr));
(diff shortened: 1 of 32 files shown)


derobins (approved):

src/H5Cdbg.c
```diff
@@ -415,12 +405,7 @@ H5C_stats(H5C_t *cache_ptr, const char *cache_name,
...
     FUNC_ENTER_NOAPI(FAIL)
 
-    HDassert(cache_ptr->magic == H5C__H5C_T_MAGIC);
-
-    /* This would normally be an assert, but we need to use an HGOTO_ERROR
-     * call to shut up the compiler.
-     */
-    if ((NULL == cache_ptr) || (cache_ptr->magic != H5C__H5C_T_MAGIC) || (NULL == cache_name))
+    if (NULL == cache_ptr || NULL == cache_name)
         HGOTO_ERROR(H5E_CACHE, H5E_SYSTEM, FAIL, "Bad cache_ptr or cache_name")
```
derobins, line 409: Not a big deal, but we could use H5_ATTR_UNUSED to "shut up the compiler" now

src/H5Centry.c
```diff
@@ -885,12 +881,6 @@ H5C__flush_single_entry(H5F_t *f, H5C_cache_entry_t *entry_ptr, unsigned flags)
                                 "can't notify client about entry dirty flag cleared")
             } /* end if */
 
-            /* we are about to discard the in core representation --
-             * set the magic field to bad magic so we can detect a
-             * freed entry if we see one.
-             */
-            entry_ptr->magic = H5C__H5C_CACHE_ENTRY_T_BAD_MAGIC;
-
```
derobins, line 893: I feel like we need to expunge the term "in core" from the library. "Core" stopped being a common synonym for "memory" a long time ago.

src/H5Cprivate.h
```diff
@@ -976,25 +972,6 @@ typedef int H5C_ring_t;
...
- *
- *              This in turn can cause a recursive call to
- *              H5C__make_space_in_cache() which may either flush or evict
- *              the next entry that the first invocation of that function
- *              was about to examine.
- *
- *              The magic field allows H5C__make_space_in_cache() to
- *              detect this case, and re-start its scan from the bottom
- *              of the LRU when this situation occurs.
- *
```
derobins, line 997: Is this magic number still needed to avoid this problem? I don't see a place where this is checked in the code, so I'm assuming this is a stale comment.

src/H5Cdbg.c
qkoziol: Yes, true, I should review these

src/H5Cprivate.h
qkoziol: I was concerned when I read the comment also and investigated the algorithm(s) that were involved.  They have all moved away from this method of using the magic field to restart scans, and the comment was indeed stale. :-)

src/H5Cdbg.c
derobins: Probably out of scope for this PR. We can clean in a subsequent one.

test/cache.c
```diff
@@ -24793,17 +24793,6 @@ check_auto_cache_resize_input_errs(unsigned paged)
         }
     }
 
-    if (pass) {
-
-        result = H5C_get_cache_auto_resize_config((const H5C_t *)&test_auto_size_ctl, &test_auto_size_ctl);
```
jhendersonHDF, line 24798: Was this test meant to be removed?

qkoziol: Yes, that was an internal test that was checking the magic field explicitly and, along with the field, unnecessary.

jhendersonHDF (approved):

PR #2955, opened 2023-05-14
Update H5private.h to fix errors with vastprintf() and astprintf() 

cross85:
When building hd5f with intel compilers on ubuntu 22.04, I got the following error message:

`
H5E.c:1339:9: error: call to undeclared function 'vasprintf'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    if (HDvasprintf(&tmp, fmt, ap) < 0)
        ^
./H5private.h:1498:34: note: expanded from macro 'HDvasprintf'
#define HDvasprintf(RET, FMT, A) vasprintf(RET, FMT, A)
                                 ^
H5E.c:1339:9: note: did you mean 'vsprintf'?
./H5private.h:1498:34: note: expanded from macro 'HDvasprintf'
#define HDvasprintf(RET, FMT, A) vasprintf(RET, FMT, A)
                                 ^
/usr/include/stdio.h:373:12: note: 'vsprintf' declared here
extern int vsprintf (char *__restrict __s, const char *__restrict __format,
           ^
1 error generated.
`

This error message is fixed adding `#define _GNU_SOURCE` to [H5private.h](https://github.com/HDFGroup/hdf5/commit/0228a16b26049108f8c9165362fcad0e755ca24d)

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -17,6 +17,8 @@
  *          defined in H5config.h which is included by H5public.h.
  */
 
+#define _GNU_SOURCE
+
 #ifndef H5private_H
 #define H5private_H
 


derobins (changes requested):
This needs to be a change to the Autotools and CMake, not the header file.

derobins:
We'll need to update the Intel checks to define _GNU_SOURCE for the platform you are building on. Which build system did you use? CMake or the Autotools? And which version of Intel's compilers?

cross85:
I used the autotools, and I think the latest Intel version (2023.1.0)

byrnHDF:
> I used the autotools, and I think the latest Intel version (2023.1.0)

classic compilers or the newer ifx?
Our instance of 22.04 and Intel compilers do not show this error for the new compiler.

cross85:
I used the new ifx compileres with hdf5 1.14.1-2

derobins:
Can you create an issue for this and close this PR? I want to track this as we implement oneAPI testing, but this fix is incorrect and can't go into the library.

cross85:
> Can you create an issue for this and close this PR? I want to track this as we implement oneAPI testing, but this fix is incorrect and can't go into the library.

It's the new compilers problem, I used oneAPI with icc and ifort instead of icx and ifx, and works ok.

derobins:
> > Can you create an issue for this and close this PR? I want to track this as we implement oneAPI testing, but this fix is incorrect and can't go into the library.
> 
> It's the new compilers problem, I used oneAPI with icc and ifort instead of icx and ifx, and works ok.

Okay, but we don't see it in our testing and this fix breaks other compilers, so it couldn't go in even if it were the right fix.

derobins:
I'm going to go ahead and close this since this change can't go into the library. A proper fix should go in the Autotools and CMake files. In the meantime, it'd be useful to create an issue for this.

PR #2963, opened 2023-05-16

mattjala:


src/H5Spoint.c
```diff
@@ -1291,17 +1303,65 @@ H5S__point_deserialize(H5S_t **space, const uint8_t **p)
...
+    /* decode the number of points */
+    switch (enc_size) {
+        case H5S_SELECT_INFO_ENC_SIZE_4:
+            if (H5_IS_KNOWN_BUFFER_OVERFLOW(skip, pp, sizeof(uint32_t), p_end))
+                HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, FAIL,
+                            "buffer overflow while decoding number of points")
+
+            UINT32DECODE(pp, num_elem);
+            break;
+        case H5S_SELECT_INFO_ENC_SIZE_8:
```
jhendersonHDF: Since it looks like in the 1.10 branch the encoding size here was always essentially H5S_SELECT_INFO_ENC_SIZE_4, it's probably better to enforce that size with an error check for if the enc. size is anything different.

src/H5Spoint.c
```diff
@@ -1291,17 +1303,65 @@ H5S__point_deserialize(H5S_t **space, const uint8_t **p)
...
+    if (H5_IS_KNOWN_BUFFER_OVERFLOW(skip, pp, coordinate_buffer_requirement, p_end))
+        HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, FAIL, "buffer overflow while decoding selection coordinates")
+
     /* Retrieve the coordinates from the buffer */
     for (tcoord = coord, i = 0; i < num_elem; i++)
         for (j = 0; j < (unsigned)rank; j++, tcoord++)
-            UINT32DECODE(pp, *tcoord);
+            switch (enc_size) {
+                case H5S_SELECT_INFO_ENC_SIZE_4:
+                    UINT32DECODE(pp, *tcoord);
```
jhendersonHDF: This change block can probably remain as just `UINT32DECODE(pp, *tcoord);` if an enc. size check is added as in the above comment.

src/H5Spoint.c
```diff
@@ -1291,17 +1303,65 @@ H5S__point_deserialize(H5S_t **space, const uint8_t **p)
...
     /* Allocate space for the coordinates */
     if (NULL == (coord = (hsize_t *)H5MM_malloc(num_elem * rank * sizeof(hsize_t))))
         HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "can't allocate coordinate information")
 
+    /* Determine necessary size of buffer for coordinates */
+    size_t enc_type_size = 0;
+
+    switch (enc_size) {
+        case H5S_SELECT_INFO_ENC_SIZE_4:
+            enc_type_size = sizeof(uint32_t);
```
jhendersonHDF: can probably just set this directly to `sizeof(uint32_t)` if an enc. size check is added as in the above comment.

src/H5Odtype.c
```diff
@@ -123,20 +123,38 @@ const H5O_msg_class_t H5O_MSG_DTYPE[1] = {{
...
 
     FUNC_ENTER_NOAPI_NOINIT
 
     /* check args */
     HDassert(pp && *pp);
-    HDassert(dt && dt->shared);
+
+    if (!dt || !dt->shared) {
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "attempted to decode bad datatype");
+    }
```
jhendersonHDF: This should remain consistent across branches and be an assert rather than an error check

mattjala: Having this (and the matching check in [H5T_close_real](https://github.com/HDFGroup/hdf5/blob/ec4c67733fa55771edc8102a13adc2ab0b7522ad/src/H5T.c#LL3914C20-L3914C21)) as an assert raises a couple failures during testing. `H5TEST-dtypes` and `H5TEST-version_bounds_1_10` fail when accessing objects with compound datatypes that have no fields, since they expect to get an error back from the library but execution is interrupted by the assert. I think it's expecting the error to be caught earlier - I'll  look into it.

src/H5Odtype.c
```diff
@@ -1099,30 +1280,47 @@ H5O_dtype_encode_helper(const H5F_t *f, uint8_t **pp, const H5T_t *dt)
...
 
     /* Set return value */
     ret_value = dt;
 
 done:
+    /* Cleanup on error */
+    if (!ret_value) {
+        if (dt != NULL) {
+            if (dt->shared != NULL)
+                dt->shared = H5FL_FREE(H5T_shared_t, dt->shared);
```
mattjala: Doing this instead of `H5T_close_real(dt)` avoids problems in cases where `dt->shared` is `NULL`.

jhendersonHDF: Is the develop branch not failing in the same way? Since it uses `H5T_close_real(dt)` we should figure out the difference between the two branches so that they are kept in sync with each other. Also, using just H5FL_FREE here could miss freeing memory for things inside `dt->shared`.

mattjala: This was just an issue with this branch.  Upon further examination, the actual root issue was duplication of freeing logic introduced during the merge; I've set this back to `H5T_close_real`.

src/H5Odtype.c
```diff
@@ -563,14 +752,6 @@ H5O_dtype_decode_helper(H5F_t *f, unsigned *ioflags /*in,out*/, const uint8_t **
     } /* end switch */
 
 done:
-    if (ret_value < 0) {
-        if (dt != NULL) {
-            if (dt->shared != NULL)
-                dt->shared = H5FL_FREE(H5T_shared_t, dt->shared);
-            dt = H5FL_FREE(H5T_t, dt);
-        } /* end if */
-    }     /* end if */
```
jhendersonHDF, line 572: This should match develop in that develop has:

```C
    /* Cleanup on error */
    if (ret_value < 0)
        /* Release (reset) dt but do not free it - leave it as an empty datatype as was the case on
         * function entry */
        if (H5T__free(dt) < 0)
            HDONE_ERROR(H5E_DATATYPE, H5E_CANTRELEASE, FAIL, "can't release datatype info")
```

here to fix a separate issue.

src/H5Odtype.c
```diff
@@ -1099,30 +1280,43 @@ H5O_dtype_encode_helper(const H5F_t *f, uint8_t **pp, const H5T_t *dt)
...
+    if (H5O_dtype_decode_helper(f, ioflags, &p, dt, skip, p_end) < 0)
         HGOTO_ERROR(H5E_DATATYPE, H5E_CANTDECODE, NULL, "can't decode type")
 
     /* Set return value */
     ret_value = dt;
 
 done:
+    /* Cleanup on error */
+    if (!ret_value) {
+        H5T_close_real(dt);
```
jhendersonHDF: Should check for errors here. Develop looks like

```C
    /* Cleanup on error */
    if (!ret_value)
        /* Free dt */
        if (H5T_close_real(dt) < 0)
            HDONE_ERROR(H5E_DATATYPE, H5E_CANTRELEASE, NULL, "can't release datatype info")
```

jhendersonHDF (approved):

PR #2964, opened 2023-05-16
Change UBSAN for undefined instead of address

### LINKED ISSUE #2947: HDF_NO_UBSAN should (also) disable UB sanitizer
Hi!

We find that when running GCC or Clang UBSAN on code that uses HDF5, it still throws warnings despite the `HDF_NO_UBSAN` flag:

```
hdf5/src/H5detect.c:1295:5: runtime error: store to misaligned address 0x503000000341 for type 'short int', which requires 2 byte alignment
0x503000000341: note: pointer points here
 00 00 00  be be be be be be be be  be be be be be be be be  be be 00 00 00 00 00 00  00 00 00 00 00
              ^ 
```

The reason for this is that this particular warning comes from the UB sanitizer, not the ASAN sanitizer. Minimal example:

https://godbolt.org/z/YMYcqhPEz

Replacing it with `no_sanitize_undefined` solves the issue.

byrnHDF:


byrnHDF:
Addresses #2947

jhendersonHDF (approved):

qkoziol (approved):

PR #2974, opened 2023-05-17
Add new API function H5Pget_actual_select_io_mode()

fortnern:
This function allows the user to determine if the library performed selection I/O, vector I/O, or scalar (legacy) I/O during the last HDF5 operation performed with the provided DXPL. Expanded existing tests to check this functionality.


test/select_io_dset.c
```diff
@@ -127,7 +142,8 @@ test_no_type_conv(hid_t fid, unsigned chunked, unsigned dtrans, unsigned mwbuf)
     int         trans_wbuf[DSET_SELECT_DIM];
     int         rbuf[DSET_SELECT_DIM];
     char        dset_name[DSET_NAME_LEN];
-    const char *expr = "2*x";
+    const char *expr                  = "2*x";
+    uint32_t    no_selection_io_cause = 0;
```
vchoi-hdfgroup: I forgot to remove the declaration (and subsequent ones) for "no_selection_io_cause".

fortnern: I'm planning to add the tests back in, needs more work though

src/H5Pdxpl.c
```diff
@@ -2610,6 +2621,39 @@ H5Pget_no_selection_io_cause(hid_t plist_id, uint32_t *no_selection_io_cause /*o
...
+{
+    H5P_genplist_t *plist;
+    herr_t          ret_value = SUCCEED; /* return value */
+
+    FUNC_ENTER_API(FAIL)
+    H5TRACE2("e", "ix", plist_id, actual_selection_io_mode);
+
+    /* Get the plist structure */
+    if (NULL == (plist = H5P_object_verify(plist_id, H5P_DATASET_XFER)))
+        HGOTO_ERROR(H5E_ID, H5E_BADID, FAIL, "can't find object for ID")
```
derobins: Should probably be H5E_PLIST, not H5E_ID

src/H5Pdxpl.c
```diff
@@ -2610,6 +2621,39 @@ H5Pget_no_selection_io_cause(hid_t plist_id, uint32_t *no_selection_io_cause /*o
...
+
+    /* Return values */
+    if (actual_selection_io_mode)
```
derobins, line 2495: Shouldn't it be an error to pass a NULL pointer to the only output variable?

src/H5CX.c
```diff
@@ -2657,6 +2669,43 @@ H5CX_get_no_selection_io_cause(uint32_t *no_selection_io_cause)
...
+/*-------------------------------------------------------------------------
+ * Function:    H5CX_get_no_selection_io_cause
+ *
+ * Purpose:     Retrieves the cause for not performing selection I/O
+ *              for the current API call context.
+ *
+ * Return:      Non-negative on success / Negative on failure
+ *
+ * Programmer:  Vailin Choi
+ *              April 15, 2023
```
derobins: I've been removing the "Programmer" lines.

src/H5CX.c
```diff
@@ -2657,6 +2669,43 @@ H5CX_get_no_selection_io_cause(uint32_t *no_selection_io_cause)
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5CX_get_no_selection_io_cause() */
 
+/*-------------------------------------------------------------------------
+ * Function:    H5CX_get_no_selection_io_cause
```
derobins: The name doesn't match the function name

src/H5CX.c
```diff
@@ -3702,7 +3751,42 @@ H5CX_set_no_selection_io_cause(uint32_t no_selection_io_cause)
...
+    head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
+    HDassert(head && *head);
+    HDassert((*head)->ctx.dxpl_id != H5P_DEFAULT);
+
+    /* If we're using the default DXPL, don't modify it */
+    if ((*head)->ctx.dxpl_id != H5P_DATASET_XFER_DEFAULT) {
+        /* Cache the value for later, marking it to set in DXPL when context popped */
+        (*head)->ctx.actual_selection_io_mode     = actual_selection_io_mode;
+        (*head)->ctx.actual_selection_io_mode_set = TRUE;
+    } /* end if */
```
derobins: /* end if */ comments are noise, especially when the if block is just a few lines long

test/select_io_dset.c
```diff
@@ -106,6 +106,21 @@ typedef enum {
 #define TEST_TCONV_BUF_TOO_SMALL               0x100
 #define TEST_IN_PLACE_TCONV                    0x200
 
+static herr_t
+check_actual_selection_io_mode(hid_t dxpl, uint32_t sel_io_mode_expected)
+{
+    uint32_t actual_sel_io_mode;
+
+    if (H5Pget_actual_selection_io_mode(dxpl, &actual_sel_io_mode) < 0)
+        FAIL_STACK_ERROR;
```
derobins: API calls should probably use TEST_ERROR, not FAIL_STACK_ERROR, which will duplicate the error stack dump (TEST_ERROR already dumps the stack when h5_reset() has been called)

test/select_io_dset.c
```diff
@@ -106,6 +106,21 @@ typedef enum {
...
+    if (actual_sel_io_mode != sel_io_mode_expected)
+        TEST_ERROR;
```
derobins, line 117: Non-API errors could use TEST_ERROR, though FAIL_PUTS_ERROR can provide more information

src/H5Ppublic.h
```diff
@@ -420,6 +420,11 @@ typedef enum H5D_selection_io_mode_t {
 } H5D_selection_io_mode_t;
 //! <!--[H5D_selection_io_mode_t_snip] -->
 
+/* Actual selection I/O modes for H5Pget_actual_selection_io_mode() property */
+#define H5D_SCALAR_IO    0x1 /* Scalar (or legacy MPIO) I/O was performed */
+#define H5D_VECTOR_IO    0x2 /* Vector I/O was performed */
+#define H5D_SELECTION_IO 0x4 /* Selection I/O was performed */
```
derobins: The default is 0, which isn't a value here. What does it mean when the function returns zero? That's not indicated in the Doxygen for the new API call, but it's tested for in the test.

derobins (changes requested):
In addition to a bunch of minor changes, I think we should talk about this a little more. I think the 0 / default case needs addressing. Also, H5D_SELECTION_IO is a value but never appears in the tests.

qkoziol:
What is the right value to return when a terminal VOL connector that's not the native connector is accessing the container?

fortnern:
> What is the right value to return when a terminal VOL connector that's not the native connector is accessing the container?

Possibly create a new "unknown/not set" value, and set it on the plist when entering the API above the VOL, so the app knows when it has no information on what happened. Reset to 0 when entering the native VOL

src/H5Ppublic.h
fortnern: This is a bitflags type parameter. 0 indicates no flags set (no raw data I/O peformed). Flags should only be set when I/O is performed so it always starts out as 0.

fortnern: It is possible for this function to return a combination of flags.

src/H5Pdxpl.c
fortnern: The way this works currently is consistent with other H5Pget* functions, even ones with only one output variable (and that return herr_t).

src/H5Ppublic.h
fortnern: We could make this clearer in the documentation but I don't think this needs to be changed. It's worth keeping in mind that the "default" is not really a default in the same way as other properties since this property is not set by the user.

src/H5Pdxpl.c
fortnern: This is consistent with how H5P_object_verify is handled everywhere else. I don't have strong feelings either way but I think if we're going to change it we should change it everywhere in a separate PR. Or we can just wait until we're able to get rid of major/minor errors classes.

fortnern:
> > What is the right value to return when a terminal VOL connector that's not the native connector is accessing the container?
> 
> Possibly create a new "unknown/not set" value, and set it on the plist when entering the API above the VOL, so the app knows when it has no information on what happened. Reset to 0 when entering the native VOL

The more I think about this the more I think we just shoudn't do anything for now. I actually implemented most of this and it's kind of awkward and probably adds more confusion to the docs. We should, however, add developer level "set" versions of this API and similar "get_actual" APIs. Then it will simply be a case of VOL connectors either reporting it or not. I don't think we need to do this now however.

src/H5Ppublic.h
```diff
@@ -420,6 +420,11 @@ typedef enum H5D_selection_io_mode_t {
 } H5D_selection_io_mode_t;
 //! <!--[H5D_selection_io_mode_t_snip] -->
 
+/* Actual selection I/O modes for H5Pget_actual_selection_io_mode() property */
+#define H5D_SCALAR_IO    0x1 /* Scalar (or legacy MPIO) I/O was performed */
```
brtnfld: I think the comment statements need to be Doxygen comments.

src/H5Ppublic.h
```diff
@@ -420,6 +420,13 @@ typedef enum H5D_selection_io_mode_t {
 } H5D_selection_io_mode_t;
 //! <!--[H5D_selection_io_mode_t_snip] -->
 
+/**
+ * Causes for H5Pget_actual_selection_io_mode() property
+ */
+#define H5D_SCALAR_IO    (0x0001u) /**< Scalar (or legacy MPIO) I/O was performed */
+#define H5D_VECTOR_IO    (0x0002u) /**< Vector I/O was performed */
+#define H5D_SELECTION_IO (0x0004u) /**< Selection I/O was performed */
+
```
brtnfld, line 429: Why not do it like the other bitmap assignments and use  enum? 
typedef enum selection_io_mode_t{ 
H5D_SCALAR_IO    (0x0001u) ,
H5D_VECTOR_IO    (0x0002u) ,
H5D_SELECTION_IO (0x0004u) ,
}

fortnern: Dana and I agreed that stuffing bitflag values into an enum is confusing and inconsistent with normal usage of both enums and bitfields, and that we should change the types that use that pattern to use simple macros in 2.0.

doxygen/examples/tables/propertyLists.dox
```diff
@@ -711,6 +711,10 @@ of the library for reading or writing the actual data.</td>
 <td>Gets the cause for not performing selection or vector I/O on the last parallel I/O call.</td>
 </tr>
 <tr>
+<td>#H5Pget_actual_selection_io_mode</td>
+<td>Gets the type(s) (scalar, vector, selection) of raw data I/O performed I/O on the last I/O call.</td>
```
jhendersonHDF: possibly "data I/O performed I/O on" -> "data I/O performed on"?

src/H5CX.c
```diff
@@ -2509,6 +2521,47 @@ H5CX_get_no_selection_io_cause(uint32_t *no_selection_io_cause)
...
+    FUNC_ENTER_NOAPI(FAIL)
+
+    /* Sanity check */
+    assert(actual_selection_io_mode);
+    head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
+    assert(head && *head);
+    assert(H5P_DEFAULT != (*head)->ctx.dxpl_id);
+
+    /* This property is a special case - we want to wipe out any previous setting.  Copy the default setting
+     * if  */
```
jhendersonHDF: unfinished comment

src/H5CX.c
```diff
@@ -3467,7 +3520,39 @@ H5CX_set_no_selection_io_cause(uint32_t no_selection_io_cause)
...
+    /* Sanity checks */
+    head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
+    assert(head && *head);
+    assert((*head)->ctx.dxpl_id != H5P_DEFAULT);
+
+    /* If we're using the default DXPL, don't modify it */
+    if ((*head)->ctx.dxpl_id != H5P_DATASET_XFER_DEFAULT) {
+        /* Cache the value for later, marking it to set in DXPL when context popped */
+        (*head)->ctx.actual_selection_io_mode     = actual_selection_io_mode;
+        (*head)->ctx.actual_selection_io_mode_set = TRUE;
```
jhendersonHDF: `TRUE` -> `true`, although I think we still #define the macro

src/H5FDint.c
```diff
@@ -455,13 +476,27 @@ H5FD_read_vector(H5FD_t *file, uint32_t count, H5FD_mem_t types[], haddr_t addrs
...
 
     /* if the underlying VFD supports vector read, make the call */
     if (file->cls->read_vector) {
-
         if ((file->cls->read_vector)(file, dxpl_id, count, types, addrs, sizes, bufs) < 0)
-
             HGOTO_ERROR(H5E_VFL, H5E_READERROR, FAIL, "driver read vector request failed");
+
+        /* Set actual selection I/O mode, if this is a raw data operation */
+        if (is_raw) {
```
jhendersonHDF, line 493: Since the scalar I/O pathway in `H5FD_write`/`H5FD_read` is checking that the raw data I/O is `size > 0`, should we be checking that as well in the vector I/O pathway, or does that matter?

src/H5Pdxpl.c
```diff
@@ -2456,6 +2467,39 @@ H5Pget_no_selection_io_cause(hid_t plist_id, uint32_t *no_selection_io_cause /*o
...
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Pget_actual_selection_io_mode(hid_t plist_id, uint32_t *actual_selection_io_mode /*out*/)
+{
+    H5P_genplist_t *plist;
+    herr_t          ret_value = SUCCEED; /* return value */
+
+    FUNC_ENTER_API(FAIL)
+    H5TRACE2("e", "ix", plist_id, actual_selection_io_mode);
```
jhendersonHDF, line 2488: Should this be `i*Iu` for the trace macro? Looks like that's used for both unsigned and uint32_t

src/H5Ppublic.h
```diff
@@ -8299,6 +8306,61 @@ H5_DLL herr_t H5Pget_selection_io(hid_t plist_id, H5D_selection_io_mode_t *selec
...
+ *          indicates no raw data I/O was performed during the operation.
+ *
+ *          Be aware that this function will only include raw data I/O performed
+ *          to/from disk as part of the last I/O operation.  Any metadata
+ *          I/O, including attribute and compact dataset I/O, is disregarded.
+ *          It is also possible that data was cached in the dataset chunk cache
+ *          or sieve buffer, which may prevent I/O from hitting the disk, and
+ *          thereby prevent it from being counted by this function.
+ *
+ * \since 1.14.2
```
jhendersonHDF: 1.14.3

testpar/t_select_io_dset.c
```diff
@@ -3446,16 +3571,16 @@ test_no_selection_io_cause_mode(const char *filename, hid_t fapl, uint32_t test_
...
          */
-        no_selection_io_cause_read_expected |= H5D_SEL_IO_NO_VECTOR_OR_SELECTION_IO_CB;
+        //    no_selection_io_cause_read_expected |= H5D_SEL_IO_NO_VECTOR_OR_SELECTION_IO_CB;
 
         /* Exception case: This will turn off selection I/O landing at H5FD__mpio_write() */
         if ((test_mode & TEST_TCONV_BUF_TOO_SMALL) && !(test_mode & TEST_IN_PLACE_TCONV))
             no_selection_io_cause_write_expected |= H5D_SEL_IO_TCONV_BUF_TOO_SMALL;
         else
-            no_selection_io_cause_write_expected |= H5D_SEL_IO_NO_VECTOR_OR_SELECTION_IO_CB;
+            //       no_selection_io_cause_write_expected |= H5D_SEL_IO_NO_VECTOR_OR_SELECTION_IO_CB;
```
jhendersonHDF: Looks like commenting out this line will cause the setting of selection I/O below to only occur in the else clause. Is that intended?

jhendersonHDF:
Just some minor comments/possible fixes but otherwise this looks good to me

src/H5FDint.c
fortnern: So we could do this for vector without too much problem, but to implement it for selection we'd have to check all the dataspaces which could be expensive for the "_id" routines which would have to resolve all of those IDs into H5S_t *s. Vector already avoids modifying the io mode when passed 0 vectors (count == 0),, and that could easily be added to selection. I suppose you could make an argument for reverting this change for scalar (it is making the callback after all), though I'd have to change the test code.

src/H5Pdxpl.c
fortnern: This is what bin/trace spits out. It might be an error but it seems to be the same for other similar functions.

testpar/t_select_io_dset.c
fortnern: I reworked this section, I think it's cleaner now

src/H5FDint.c
fortnern: I decided to revert the checking for size == 0. There's a fairly lengthy comment explaining some of this behavior in the test now.

src/H5Pdxpl.c
jhendersonHDF: It seems inconsistent based on https://github.com/HDFGroup/hdf5/blob/b916ce2419da45f7a51503e3bf2774e71f7db815/src/H5D.c#L1204

jhendersonHDF (approved):

src/H5Pdxpl.c
fortnern: It might be getting confused by the "/* out */"

fortnern: This looks like a bug in the bin/trace script that affects many existing API functions. I think the fix is outside the scope of this PR

fortnern: Filed issue #3733

derobins (approved):

Issue #2960, opened 2023-05-16
[BUG] `release_docs/README_HPC` refers 1.13.0.


hyoklee:
**Describe the bug**

I don't know why 1.13.0 appears in 1.14.1-2 release source package.

```
release_docs/README_HPC:be renamed "hdf5-1.13.0", or for the first hdf5_1_12_0 \
pre-release version,
release_docs/README_HPC:       "hdf5-1.13.0".
release_docs/README_HPC:       hdf5-1.13.0/config/cmake/scripts/HDF5config.cmak\
e
release_docs/README_HPC:       hdf5-1.13.0/config/cmake/scripts/CTestScript.cma\
ke
release_docs/README_HPC:       hdf5-1.13.0/config/cmake/scripts/HDF5options.cma\
ke
release_docs/README_HPC:       hdf5-1.13.0
release_docs/README_HPC:can be edited in the batch files in hdf5-1.13.0/bin/batch.
```

**Expected behavior**

1.13.x is experimental. Replace it with 1.14.x.

**Platform (please complete the following information)**
 - HDF5 version: 1.14.1-2.
 - OS and version: n/a
 - Compiler and version: n/a
 - Build system (e.g. CMake, Autotools) and version: n/a
 - Any configure options you specified: n/a
 - MPI library and version (parallel HDF5): n/a

**Additional context**

See [RELEASE SCHEDULE](https://github.com/HDFGroup/hdf5)

PR #2613, opened 2023-03-20
Remove version specific examples from README_HPC.

lrknox:

PR #2975, opened 2023-05-17
Remove version specific examples from README_HPC. (#2613)

lrknox:


lrknox:
Cherry-picked from previous fix in HDF5 1.10.10.

hyoklee (approved):

release_docs/README_HPC
```diff
@@ -145,19 +141,19 @@ cori, another CrayXC40, that line is replaced by "#SBATCH -C knl,quad,cache".
 For cori (and other machines), the values in LOCAL_BATCH_SCRIPT_NAME and
 LOCAL_BATCH_SCRIPT_PARALLEL_NAME in the config/cmake/scripts/HPC/sbatch-HDF5options.cmake
 file can be replaced by cori_knl_ctestS.sl and cori_knl_ctestS.sl, or the lines
-can be edited in the batch files in hdf5-1.13.0/bin/batch.
+can be edited in the batch files in hdf5-<version string(see section II)>/bin/batch.
```
jhendersonHDF: Having the (see section II) here in a pathname seems odd. Was this intended?

lrknox: Intended, but agreed that it is odd.  The purpose was to let the user know how to find the version, but maybe it's unnecessary.  I can take it out if you think it would be better.

jhendersonHDF: I think the referral to section II from section I above is helpful, but here my first thought was just that it was a copy-paste issue. We could maybe still refer to section II somewhere in the sentence but I think it'd be better outside the path name.

lrknox:
In the source file for the PR, the "(see section II) came after the closing '>'.  I don't know if the PR is just showing it inside, or if it was actually moved.  I think I'll just make it a complete sentence and remove the ()s.

jhendersonHDF:
> In the source file for the PR, the "(see section II) came after the closing '>'. I don't know if the PR is just showing it inside, or if it was actually moved. I think I'll just make it a complete sentence and remove the ()s.

Ah, it looks like we were looking at separate occurrences. The first mention of (see section II) in section I looked fine to me. It's the second occurrence in the file where it's in the middle of a path name that I was looking at.

lrknox:
I  updated it.

jhendersonHDF:
It looks like this PR still has "hdf5-<version string(see section II)>" under section IV., which is what I was referring to

lrknox:
> It looks like this PR still has "hdf5-<version string(see section II)>" under section IV., which is what I was referring to

Yes, I was looking at section III.  I think I've got it now.

jhendersonHDF (approved):

hyoklee:
Close #2960

PR #2980, opened 2023-05-18
ci: set minimal permissions on GitHub Workflows

### LINKED ISSUE #2973: CI: Set minimal permissions on GitHub Workflows
Hi!

I'm here to suggest that you set minimal permissions to your GitHub Workflows, because currently they don't specify the permissions for their jobs and their privileges are being determined by GitHub's defaults. If you define minimal permissions, you would be secured against erroneous or malicious behaviours from external jobs you call from your workflow. It's specially important for the case they get compromised, for example.

The ideal would be to set a top-level read-only permission on all workflows, so that they would be inherited by any job that does not define job-level permissions. Additionally, you can set any write permissions as job-level permissions, only where they are needed -- as I see its the case for your job in [clang-format-fix.yml](https://github.com/HDFGroup/hdf5/blob/develop/.github/workflows/clang-format-fix.yml), which requires `contents: write` permission.

Setting minimum permissions for workflows is recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

Let me know what you think about this. If you agree, I'd be happy to contribute with a PR =).

#### Context
I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://github.com/diogoteles08#about-gosst-ghost)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

diogoteles08:
Closes #2973

I've tried to evaluate the permissions needed to all your workflows but I couldn't be 100% sure of some of them. It went all fine in my testings on the fork, but if you have any issues on runs on your repo, we might need to add some extra permissions.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -1,6 +1,8 @@
 name: clang-format Check
 on:
   pull_request:
+permissions:
+  contents: read
 jobs:
   formatting-check:
     name: Formatting Check
--- a/.github/workflows/clang-format-fix.yml
+++ b/.github/workflows/clang-format-fix.yml
@@ -11,11 +11,15 @@ name: clang-format Commit Changes
 on: 
   workflow_dispatch:
   push:
+permissions:
+  contents: read
 jobs:
   formatting-check:
     name: Commit Format Changes
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'skip-ci')"
+    permissions:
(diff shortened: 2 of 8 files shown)


byrnHDF (approved):

jhendersonHDF (approved):
Thanks for the PR @diogoteles08!

PR #2987, opened 2023-05-19
Updated H5Ovisit_f and H5Ovisit_by_name_f wrappers internals.

brtnfld:
* Removed C wrappers for H5Ovisit_f and H5Ovisit_by_name_f, modifying the Fortran source accordingly.

* The intent for op_data was declared INOUT, even though the pointer address INTENT is, in actuality, IN. gfortran was optimizing out op_data in tests where the values were repeatedly reset to the same value. The values were reset in the test because the data the pointer targeted was updated in the callback.

* Made use of the 'verify' function to check value correctness.

* Updated the H5(A&L)exists_async tests to use INTEGER(C_INT) instead of C_BOOL, and update the documentation to suggest using C_INT over C_BOOL.

### DIFF
--- a/fortran/src/H5Off.F90
+++ b/fortran/src/H5Off.F90
@@ -1025,33 +1025,35 @@ CONTAINS
     INTEGER(HID_T), INTENT(IN) :: object_id
     INTEGER, INTENT(IN) :: index_type
     INTEGER, INTENT(IN) :: order
-
     TYPE(C_FUNPTR), INTENT(IN) :: op
-    TYPE(C_PTR), INTENT(IN)    :: op_data
+    TYPE(C_PTR), INTENT(INOUT) :: op_data ! Declare INOUT to bypass gfortran 4.8.5 issue
     INTEGER, INTENT(OUT) :: return_value
     INTEGER, INTENT(OUT) :: hdferr
-    INTEGER, INTENT(IN), OPTIONAL  :: fields
-    INTEGER :: fields_c
+    INTEGER, INTENT(IN), OPTIONAL :: fields
+
+    INTEGER(C_INT) :: fields_c
+    INTEGER(C_INT) :: return_value_c
 
     INTERFACE
-       INTEGER FUNCTION h5ovisit_c(object_id, index_type, order, op, op_data, fields) &
-            BIND(C, NAME='h5ovisit_c')
-         IMPORT :: C_FUNPTR, C_PTR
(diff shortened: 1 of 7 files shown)


lrknox:
I ran the daily test configurations for the visit branch on jelly and echidna.  The previous fortran failures are fixed.  The H5L async API tests failed with MPICH 3.3, MPICH 4.0.3 (module MPICH/master-GCC-7.2.0-2.29), and module OpenMPI/master on jelly with these errors:

ERROR: INCORRECT VALIDATION H5Lexists_async_f (x12)

All tests including async API tests passed with OpenMPI 3.1.3 and OpenMPI 4.1.2.  module OpenMPI/master was installed July28 2022, so probably all tests will pass with latest MPICH and OpenMPI versions.

lrknox (approved):

brtnfld:
I just realized the particularity of logical with C_BOOL. I can not determine how to compare the value to the correct true or false. It was a mistake to pass a pointer instead of a logical for exists APIs.

byrnHDF:
Should there be a release note about the removal of the functions? Or does it not affect the public API?

brtnfld:
Updated the tests to be portable, updated the documentation. It needs to be a pointer since it is async and the value gets set later.

lrknox:
I ran the daily test configurations for the visit branch again on jelly and echidna. The previous fortran and H5L async API test failures are fixed. The only failing configurations were 2 cmakehdf5 configurations that failed because ZLib.tar.gz was not found, which is a different issue with PR 2988 that fixes it.

byrnHDF (approved):

fortran/src/H5Off.F90
```diff
@@ -1025,33 +1025,35 @@ SUBROUTINE h5ovisit_f(object_id, index_type, order, op, op_data, return_value, h
     INTEGER(HID_T), INTENT(IN) :: object_id
     INTEGER, INTENT(IN) :: index_type
     INTEGER, INTENT(IN) :: order
-
     TYPE(C_FUNPTR), INTENT(IN) :: op
-    TYPE(C_PTR), INTENT(IN)    :: op_data
+    TYPE(C_PTR), INTENT(INOUT) :: op_data ! Declare INOUT to bypass gfortran 4.8.5 issue
```
derobins, line 1029: Ugh. It'd be nice to have a better way around this.

derobins (approved):

PR #2988, opened 2023-05-19
Fix plugin cache variable name

byrnHDF:
Requires change in plugins project also


lrknox (approved):

config/cmake/HDF5PluginCache.cmake
```diff
@@ -8,7 +8,7 @@
 set (H5PL_BUILD_TESTING ON CACHE BOOL "Enable H5PL testing" FORCE)
 set (BUILD_EXAMPLES ON CACHE BOOL "Build H5PL Examples" FORCE)
 
-set (HDF5_HDF5_HEADER "H5pubconf.h" CACHE STRING "Name of HDF5 header" FORCE)
+set (H5PL_HDF5_HEADER "H5pubconf.h" CACHE STRING "Name of HDF5 header" FORCE)
 set (HDF5_LINK_LIBS ${HDF5_LIBSH_TARGET} CACHE STRING "HDF5 target" FORCE)
```
jhendersonHDF: I think this one also needs to be updated if I understand correctly.

jhendersonHDF (approved):

config/cmake/HDF5PluginCache.cmake
byrnHDF: Maybe, I need to see where that is used but most likely it needs to be prefixed with "H5PL_"

release_docs/INSTALL_CMake.txt
```diff
@@ -325,17 +325,17 @@ IV. Further considerations
             ZLIB_TGZ_ORIGPATH:STRING="some_location"
 
             HDF5_ALLOW_EXTERNAL_SUPPORT:STRING="TGZ"
-         where "some_location" is the URL or full path to the compressed
-         file and ext is the type of compression file. The individual filters are
-         enabled by setting the BUILD_<filter>_WITH_FETCHCONTENT CMake variable to ON.
+         where "some_location" is the URLand by setting
```
lrknox:
```suggestion
         where "some_location" is the URL and by setting
```

PR #3002, opened 2023-05-23
Develop release workflow

byrnHDF:
This adds a SNAPSHOT tag and creates artifacts for download (location/url still under investigation).
Will also add doxygen artifacts before merging

### DIFF
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -27,5 +27,6 @@ jobs:
     uses: ./.github/workflows/release.yml
     with:
       file_base: ${{ needs.call-workflow-tarball.outputs.file_base }}
+      file_sha: ${{ needs.call-workflow-tarball.outputs.file_sha }}
     if: ${{ needs.call-workflow-tarball.outputs.has_changes == 'true' }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/daily-build.yml
```diff
@@ -6,9 +6,6 @@ on:
   schedule:
     - cron: "6 0 * * *"
 
-permissions:
-  contents: read
```
jhendersonHDF, line 10: Does this action need write permissions? Can those be given on a per-job basis while leaving this line in?

byrnHDF: Maybe - It might be from debugging. I will try

byrnHDF: No - a deeper understanding of permissions will be needed.

jhendersonHDF: I'm guessing that if you leave in the global read permissions line here, you still need to set write permissions for each job that needs it in this file, even if the workflows for those jobs are setting that.

derobins (approved):

jhendersonHDF (approved):

PR #3018, opened 2023-05-26
API test updates

jhendersonHDF:
Most of the changes in this PR just replace macros from test/API/H5_api_tests_disabled.h that were disabling API tests with appropriate checks for VOL connector capability flags. However, there are a few bug fixes and CMake code updates to implement support for pulling in, building and testing external VOL connectors.


testpar/API/CMakeLists.txt
jhendersonHDF, line 1: The changes in this file hook up external built VOL connectors into the API tests

test/CMakeLists.txt
```diff
@@ -49,6 +49,22 @@ if (BUILD_STATIC_LIBS)
   endif ()
   H5_SET_LIB_OPTIONS (${HDF5_TEST_LIB_TARGET} ${HDF5_TEST_LIB_NAME} STATIC 0)
   set_target_properties (${HDF5_TEST_LIB_TARGET} PROPERTIES FOLDER libraries/test)
+
+  if (HDF5_EXPORTED_TARGETS)
```
jhendersonHDF: @byrnHDF I just realized for the purposes of the API tests I was unconditionally installing the test lib here and below. Would it be appropriate to surround it with `if (HDF5_TEST_API_INSTALL)` or should we have something else since there may be other reasons to install the test lib?

test/API/H5_api_test_util.c
jhendersonHDF, line 1: The changes in this file mostly just separated the code for generating different datatypes into sub-functions, but it also fixed a bug in enum. datatype generation where we tried to insert duplicate name/value pairs into the datatype.

test/API/CMakeLists.txt
jhendersonHDF, line 1: The changes in this file hook up external built VOL connectors into the API tests

src/H5VLnative_attr.c
jhendersonHDF, line 1: The changes in this file mirror a check that was added to H5Aopen where we want to make sure we return failure when an invalid AAPL is passed in. The API tests were failing for the native VOL connector due to checking this.

CMakeVOL.cmake
jhendersonHDF, line 1: This new file contains all the code to generate CMake options and variables for specifying external VOL connectors and then pulling in, building those connectors and setting up for later testing of the connectors.

byrnHDF (approved):

glennsong09 (approved):

doc/cmake-vols-fetchcontent.md
```diff
@@ -0,0 +1,195 @@
...
+    HDF5_VOL_VOL-ASYNC_NAME
+    HDF5_VOL_VOL-ASYNC_TEST_PARALLEL
+
+**NOTE**
+If a VOL connector requires extra information to be passed in its
+HDF5_VOL_<VOL name>_NAME variable and that information contains any semi-colons,
+those semi-colons should be escaped with a single backslash so that CMake
+doesn't parse the string as a list. If `cmake` is run from a shell, extra care
+may need to be taken when escaping the semi-colons depending on how the
+shell interprets backslashes.
```
lrknox, line 100: 'semicolons' shouldn't have a '-'

lrknox (approved):

doc/cmake-vols-fetchcontent.md
jhendersonHDF: Fixed

jhendersonHDF:
The EOS workflow appears to be failing due to too many requests being made, likely because the same actions are running in the feature/api_tests feature branch and those end up running two sets of the same actions for some reason, so I believe there are 122 checks running simultaneously, which is a bit overkill

PR #3026, opened 2023-05-29
Only pass efc pointer to H5F__efc_open

derobins:
The H5F__efc_open() API call only needs the efc pointer, so there is no need to pass the entire file struct in.

### DIFF
--- a/src/H5Fefc.c
+++ b/src/H5Fefc.c
@@ -137,9 +137,8 @@ done:
  *-------------------------------------------------------------------------
  */
 H5F_t *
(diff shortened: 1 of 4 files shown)


src/H5Fefc.c
```diff
@@ -149,8 +148,6 @@ H5F__efc_open(H5F_t *parent, const char *name, unsigned flags, hid_t fcpl_id, hi
     FUNC_ENTER_PACKAGE
 
     /* Sanity checks */
-    HDassert(parent);
-    HDassert(parent->shared);
     HDassert(name);
```
derobins, line 151: NULL is a valid value for the efc pointer

qkoziol (approved):

lrknox (approved):

PR #3030, opened 2023-06-01
ROS3: (feature) Temporary security credentials

jwsblokland:
- Implemented support for AWS temporary security credentials. For this kind of credentials also a session/security token should be included in the request by adding the x-amz-security-token header.

Related HDF forum discussion: https://forum.hdfgroup.org/t/ros3-temporary-security-credentials/11135


src/H5FDros3.c
```diff
@@ -599,6 +603,236 @@ H5FD__ros3_fapl_free(void *_fa)
...
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "failed to check if property token exists in plist")
+    if (token_exists) {
+        if (H5P_get(plist, ROS3_TOKEN_PROP_NAME, token_src) < 0)
+            HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "unable to get token value")
+    }
+
+    /* Copy the token data out */
+    tokenlen = HDstrlen(token_src);
+    if (size < tokenlen) {
+        tokenlen = size - 1;
```
mattjala, line 628: If a user passes in `size = 0`, this could assign a negative value to a `size_t`.

src/H5FDros3.c
```diff
@@ -599,6 +603,236 @@ H5FD__ros3_fapl_free(void *_fa)
...
+ *
+ * Return:      SUCCEED/FAIL
+ *
+ * Programmer:  Jan-Willem Blokland
+ *              2023-05-26
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5FD__ros3_str_token_copy(const char *name, size_t size, void *_value)
```
mattjala: It would be more consistent with the naming scheme if the argument were `value` and the internal cast were  `_value`.

src/H5FDros3.c
```diff
@@ -599,6 +603,236 @@ H5FD__ros3_fapl_free(void *_fa)
...
+
+    if (value1) {
+        if (value2)
+            ret_value = HDstrcmp(value1, value2);
+        else
+            ret_value = 1;
+    }
+    else {
+        if (value2)
+            ret_value = -1;
```
mattjala, line 700: The `FAIL`/`SUCCEED` macros would be more clear.

src/H5FDros3.c
jwsblokland: Thanks for spotting this. I will implement an additional check to ensure that `size` is at least 1. Furthermore, as far as I am aware of `size_t` is an unsigned integer data type therefore it cannot be negative. Assigning to, for example, -1 will result in the `MAX_SIZE` value. Also not something we want to happen.

src/H5FDros3.c
jwsblokland: I am happy to change this. However, I do have a question about this. 

The reason why I named it `_value` is if I look at the other `H5FD__ros3_*` functions all the pointer names start with an underscore. Therefore, I assumed there is internal convention which prescribes that the pointer name should start with an underscore for certain functions. Is this actually true? If not should I also rename the other arguments in this source file?

jwsblokland:
While looking at my changes, I noticed I made a mistake in the function `H5Pset_fapl_ros3_token()`. Currently, the function always returns `FAIL`. I will fix this.

src/H5FDros3.c
jwsblokland: It turns out I made a mistake in the function declaration the return type should have been `int` instead of `herr_t`. I will fix it. This means that your suggestion to make use of `FAIL`/`SUCCEED` marco make less sense now.

src/H5FDros3.c
mattjala: I checked this, and you're correct - leaving the argument with the underscore is consistent with the rest of this file.

jwsblokland: Ok. I will force push my updates so you can take another look at my changes.

By the way, thanks for the review. We found a few things that needs to be improved.

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
...
+        HGOTO_ERROR(H5E_PLIST, H5E_BADVALUE, FAIL, "incorrect VFL driver")
+    if (HDstrlen(token) >= H5FD_ROS3_MAX_SECRET_TOK_LEN)
+        HGOTO_ERROR(H5E_PLIST, H5E_BADVALUE, FAIL,
+                    "specified token exceeds the internally specified maximum string length")
+
+    if ((token_exists = H5P_exist_plist(plist, ROS3_TOKEN_PROP_NAME)) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "failed to check if property token exists in plist")
+
+    if (token_exists) {
+        if (H5P_set(plist, ROS3_TOKEN_PROP_NAME, &token) < 0)
```
mattjala: I think this should be `(void *) token` instead of `&token` - H5P_set expects a pointer to a value, not a pointer to a pointer to a value.

jwsblokland: Yes, you are right. I will fix this too.

src/H5FDs3comms.c
```diff
@@ -1091,6 +1096,12 @@ H5FD_s3comms_s3r_open(const char *url, const char *region, const char *id, const
         if (handle->signing_key == NULL)
             HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, NULL, "could not malloc space for handle key copy.");
         H5MM_memcpy(handle->signing_key, signing_key, tmplen);
+
+        tmplen        = HDstrlen(token) + 1;
+        handle->token = (char *)H5MM_malloc(sizeof(char) * tmplen);
```
mattjala, line 1081: Needs a corresponding `H5MM_xfree` in `H5FD_s3comms_s3r_close`

src/H5FDros3.h
```diff
@@ -68,13 +68,21 @@
  *
  *     String: "Secret Access Key" associated with the ID and resource.
  *
+ * `secret_tok` (char[])
```
mattjala: Since `H5FD_ros3_fapl_t` doesn't contain this as a top-level field, this information should be moved to the declarations for `H5Pget_fapl_ros3_token`/`H5Pget_fapl_ros3_token`.

src/H5FDros3.c
jhendersonHDF: @jwsblokland This was actually correct as `&token`. That way your callbacks receive a `void **` that is converted to a `char **`. The `char **` value here is implicitly converted to a `void *` for passing to `H5P_set`. In general, we use `&value` everywhere to be consistent across property callbacks.

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
     FUNC_LEAVE_NOAPI(SUCCEED)
 } /* end H5FD__ros3_fapl_free() */
 
+/*-------------------------------------------------------------------------
+ * Function:    H5Pget_fapl_ros3_token
+ *
+ * Purpose:     Returns session/secutiry token of the ros3 file access
```
jhendersonHDF: nitpick: secutiry -> security

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
...
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "token_dst is NULL")
+
+    if (NULL == (plist = H5P_object_verify(fapl_id, H5P_FILE_ACCESS)))
+        HGOTO_ERROR(H5E_PLIST, H5E_BADTYPE, FAIL, "not a file access property list")
+    if (H5FD_ROS3 != H5P_peek_driver(plist))
+        HGOTO_ERROR(H5E_PLIST, H5E_BADVALUE, FAIL, "incorrect VFL driver")
+    if ((token_exists = H5P_exist_plist(plist, ROS3_TOKEN_PROP_NAME)) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "failed to check if property token exists in plist")
+    if (token_exists) {
+        if (H5P_get(plist, ROS3_TOKEN_PROP_NAME, token_src) < 0)
```
jhendersonHDF: If the call to set the token is changed back to pass a pointer to pointer, this should then be `&token_src` in this case since all the callbacks would be expecting to receive a `void **` that gets converted to a `char **`

tools/lib/h5tools.h
```diff
@@ -569,6 +569,7 @@ typedef struct h5tools_vfd_info_t {
 
     /* Pointer to information to be passed to the driver for its setup */
     const void *info;
+    const char *token;
```
jhendersonHDF: Since this new token is currently specific to the ROS3 VFD, it'd be much better from a maintenance perspective if this field was wrapped in a new structure that encloses both the old `H5FD_ros3_fapl_t` structure and this field so that it doesn't get mixed in with other VFDs. Then, that structure can be passed for the `info` field here rather than a `H5FD_ros3_fapl_t` structure.

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
...
+
+    if ((token_exists = H5P_exist_plist(plist, ROS3_TOKEN_PROP_NAME)) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "failed to check if property token exists in plist")
+
+    if (token_exists) {
+        if (H5P_set(plist, ROS3_TOKEN_PROP_NAME, (void *)token) < 0)
+            HGOTO_ERROR(H5E_PLIST, H5E_CANTSET, FAIL, "unable to set value")
+    }
+    else {
+        if (H5P_insert(plist, ROS3_TOKEN_PROP_NAME, sizeof(char) * H5FD_ROS3_MAX_SECRET_TOK_LEN,
```
jhendersonHDF: Typically, this would just use `sizeof(char *)` for the property size since all we really need to store is the pointer to the string. There are a few approaches to this really, but since your callbacks for this property appear to work with the pointer and the call to `H5P_set` above will just set a pointer, this is probably just wasting space in the property list.

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
...
+ *              2023-05-26
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5FD__ros3_str_token_delete(hid_t prop_id, const char *name, size_t size, void *value)
+{
+    herr_t ret_value = SUCCEED;
+
+    FUNC_ENTER_PACKAGE_NOERR
```
jhendersonHDF, line 758: Similar to the way in which we do this for string properties elsewhere, I'd recommend that `H5Pset_fapl_ros3_token` should make a copy of the token string with strdup and set that pointer on the property list. That allows the caller to free the string without stale pointers being left around in any FAPL the property is set on. That would mean that this callback and `H5FD__ros3_str_token_close` would need implementations where they just free the memory for the string after retrieving the pointer for it.

src/H5FDs3comms.c
```diff
@@ -1304,23 +1315,24 @@ H5FD_s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest)
...
-        char authorization[512 + 1];
-        /*   512 := approximate max length...
-         *    67 <len("AWS4-HMAC-SHA256 Credential=///s3/aws4_request,"
-         *           "SignedHeaders=,Signature=")>
-         * +   8 <yyyyMMDD>
-         * +  64 <hex(sha256())>
-         * + 128 <max? len(secret_id)>
-         * +  20 <max? len(region)>
-         * + 128 <max? len(signed_headers)>
+        char authorization[2048 + 1];
```
jhendersonHDF: When we went to clean up warnings in HDF5 in recent times, one of the things responsible for several warnings was too large of stack-allocated buffers. Before this change the buffers were probably small enough at 513 bytes that they didn't cause warnings, but at 2049 bytes they likely will. For buffers of this size, the code here will probably need to change to using dynamic allocation to avoid warnings. We don't currently run into this in our -Werror builds because they don't build the ROS3 VFD.

src/H5FDros3.h
```diff
@@ -104,6 +112,20 @@ H5_DLL herr_t H5Pget_fapl_ros3(hid_t fapl_id, H5FD_ros3_fapl_t *fa_out);
  */
 H5_DLL herr_t H5Pset_fapl_ros3(hid_t fapl_id, const H5FD_ros3_fapl_t *fa);
 
+/**
+ * \ingroup FAPL
+ *
+ * \todo Add missing documentation
+ */
+H5_DLL herr_t H5Pget_fapl_ros3_token(hid_t fapl_id, size_t size, char *token);
```
jhendersonHDF, line 183: It would be great if we could have documentation for these API routines before this PR is merged.

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
...
+            HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "unable to get token value")
+    }
+
+    /* Copy the token data out */
+    tokenlen = HDstrlen(token_src);
+    if (size < tokenlen) {
+        tokenlen = size - 1;
+    }
+    H5MM_memcpy(token_dst, token_src, sizeof(char) * tokenlen);
+    token_dst[tokenlen] = '\0';
```
jhendersonHDF, line 631: If the user passes in a value for `size` where `size == tokenlen`, this seems to access memory outside of `token_dst`, presuming that that buffer is actually only `size` bytes big. Switching to strncpy instead could be one way of ensuring the resulting string in `token_dst` is NUL-terminated, though there are tradeoffs there. You could also probably just change the check above to `if (size <= tokenlen) {`

src/H5FDs3comms.c
```diff
@@ -1388,8 +1411,9 @@ H5FD_s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest)
         /**** COMPUTE AUTHORIZATION ****/
 
         /* buffer1 -> canonical request */
-        if (FAIL == H5FD_s3comms_aws_canonical_request(buffer1, 512, signed_headers, 48, request))
+        if (FAIL == H5FD_s3comms_aws_canonical_request(buffer1, 2048, signed_headers, 48 + 1024, request)) {
```
jhendersonHDF: I hadn't reviewed the previous code here as much as I would have liked to, but seeing this now I'd much prefer some #defines for the magic numbers here to make things more readable and reduce the chances of a bad buffer size being passed around somewhere

src/H5FDs3comms.c
```diff
@@ -1664,17 +1689,17 @@ H5FD_s3comms_aws_canonical_request(char *canonical_request_dest, int _cr_size, c
 
         HDassert(node->magic == S3COMMS_HRB_NODE_MAGIC);
 
-        ret = HDsnprintf(tmpstr, 256, "%s:%s\n", node->lowername, node->value);
-        if (ret < 0 || ret >= 256)
+        ret = HDsnprintf(tmpstr, 1024, "%s:%s\n", node->lowername, node->value);
```
jhendersonHDF, line 1694: This should really be "1024 + 1" since that's how big the buffer is, but I think the "1024 + 1" above should just be "1024" anyway and should be put behind some #define. Not really necessary to getting this PR merged, but it would be nice cleanup.

src/H5FDros3.c
```diff
@@ -599,6 +608,240 @@ H5FD__ros3_fapl_free(void *_fa)
...
+    FUNC_ENTER_PACKAGE_NOERR
+
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* H5FD__ros3_str_token_delete */
+
+/*-------------------------------------------------------------------------
+ * Function:    H5Pset_fapl_ros3_token()
+ *
+ * Purpose:     Modify the file access property list to use the H5FD_ROS3
+ *              driver defined in this source file.
```
jhendersonHDF: Documentation needs updated here, though the most descriptive explanation for the function should probably go in the doxygen in src/H5FDros3.h

src/H5FDs3comms.c
jwsblokland: Oke. I have change memory allocation of the variable `authorization`, `buffer` and `signed_headers` to a dynamic one.

src/H5FDros3.h
jwsblokland: I have added doxygen-style documentation for the defines, struct and functions. Most likely, not perfect but at least there is something now.

src/H5FDs3comms.c
jwsblokland: This has been added.

src/H5FDros3.c
jwsblokland: Fixed

src/H5FDros3.c
jwsblokland: This comment and comment of the function `H5FD__ros3_str_token_copy()`, `H5FD__ros3_str_token_close()` and `H5FD__ros3_str_token_delete()` are more or less related to each other.

As you noticed, my current implementation is slightly different compared to code snippets you posted on the forum. The main reason for it is that I want the HDF5 library to store specified token string instead of the pointer to pointer. I assume that is done when calling `H5Pset_ros3_fapl()` for the strings inside the struct `H5FD_ros3_fapl_t`. If this is not true, I need to rethink certain things when implementing ROS3 functionality on my library by making use of HDF5. My aim is that my implementation behaves similar as the existing implementation regarding specifying the security credentials.

Furthermore, specifying the token is optional and if the token is not specified no additional memory will be allocated.

src/H5FDs3comms.c
jwsblokland: This has been added. Also found another place where it should be added.

src/H5FDros3.h
jwsblokland: Yes, you are right. Removed this description from the header file.

src/H5FDros3.c
jwsblokland: Corrected this typo.

src/H5FDs3comms.c
jwsblokland: Change the implementation such that `authorization`, `buffer1` and `signed_headers` are now allocated dynamically.

src/H5FDs3comms.c
jwsblokland: On various places make use of the define `H5FD_ROS3_MAX_SECRET_TOK_LEN` found in the header file `H5FDros3.h`.

src/H5FDros3.c
jwsblokland: Slightly improved of the purpose description.

src/H5FDros3.h
jwsblokland: Added doxygen-style inline documentation for the defines, struct and functions. Most likely, not perfect but at least now there is something.

src/H5FDros3.c
jwsblokland: This has been fixed.

src/H5FDs3comms.c
jwsblokland: The last element of the tmpstr (position: 1024) is the \0 character to ensure that tmpstr is properly NULL-terminated in all cases. This necessary for the function `HDstrlen(tmpstr)`.

tools/lib/h5tools.h
jwsblokland: This has been modified based on your suggestion.

src/H5FDs3comms.c
jhendersonHDF: Sure, but snprintf should always NUL-terminate the buffer.

src/H5FDros3.h
```diff
@@ -84,26 +135,94 @@ typedef struct H5FD_ros3_fapl_t {
...
+ *      Configuration structure for H5Pset_fapl_ros3() / H5Pget_fapl_ros3().
+ *
+ * \var char H5FD_ros3_fapl_ext_t::token[H5FD_ROS3_MAX_SECRET_TOK_LEN + 1]
+ *      A string which specifies the session/security token.
+ *
+ */
+typedef struct H5FD_ros3_fapl_ext_t {
+    H5FD_ros3_fapl_t fa;
+    char             token[H5FD_ROS3_MAX_SECRET_TOK_LEN + 1];
+} H5FD_ros3_fapl_ext_t;
```
jhendersonHDF: Ah, I had mostly imagined this structure solely existing in the tools code. The issue I see with putting it into the public API here is that if any new functionality is added, it would be another API-breaking change to add it into this structure, so at that point we may as well have just added everything to the original H5FD_ros3_fapl_t structure. One other issue I see here and that I ran into elsewhere is that having large char buffer fields in API structures can make it difficult for applications to have these variables on the stack for certain platforms and usually causes warnings in example code if they are used as stack variables. I think that part shouldn't necessarily hold this PR though since we don't have the ROS3 VFD built with -Werror at the moment and warnings can be cleaned up later.

src/H5FDros3.h
jhendersonHDF: Thanks for going above and beyond and adding all the doxygen in this file!

src/H5FDros3.h
```diff
@@ -84,26 +135,94 @@ typedef struct H5FD_ros3_fapl_t {
...
+ *
+ * \fapl_id
+ * \param[in] token Session/security token.
+ * \returns \herr_t
+ *
+ * \details H5Pset_fapy_ros3_token() modifies the File Access Property List to use the
+ *          #H5FD_ROS3 driver by adding or updating the session/security token.
+ *
+ *          The session token is only needed when you want to access a S3 bucket
+ *          using temporary security credentials.
```
jhendersonHDF, line 202: For the brief and details sections here, I'd update to just mention that they set the session/security token, since `H5Pset_fapl_ros3` is the API that actually modifies the FAPL to use the ROS3 VFD.

src/H5FDs3comms.c
```diff
@@ -1304,23 +1321,25 @@ H5FD_s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest)
...
+         * +  128 <max? len(signed_headers)>
+         * + 1024 <max? len(session_token)>
          */
-        char buffer1[512 + 1]; /* -> Canonical Request -> Signature */
         char buffer2[256 + 1]; /* -> String To Sign -> Credential */
         char iso8601now[ISO8601_SIZE];
-        char signed_headers[48 + 1];
+        buffer1        = (char *)H5MM_malloc(512 + H5FD_ROS3_MAX_SECRET_TOK_LEN +
+                                             1); /* -> Canonical Request -> Signature */
+        signed_headers = (char *)H5MM_malloc(48 + H5FD_ROS3_MAX_SECRET_KEY_LEN + 1);
```
jhendersonHDF, line 1323: Should add some error checking here to make sure `authorization`, `buffer1` and `signed_headers` were properly allocated before using them

src/H5FDros3.c
jhendersonHDF: I believe your implementation should be fine as is as long as everything works correctly. Arguments could probably made for either direction when it comes to storing the string as the whole value or a pointer to the value, but if everything works as expected without any memory issues, I guess I don't see any reason not use this implementation.

jhendersonHDF: Having said that, if you plan to use the implementation where the whole value is stored, you can probably just remove the `H5FD__ros3_str_token_close` and `H5FD__ros3_str_token_delete` callbacks and pass `NULL` for them in the call to `H5P_insert`.

tools/lib/h5tools_utils.c
```diff
@@ -1215,12 +1239,24 @@ h5tools_populate_ros3_fapl(H5FD_ros3_fapl_t *fa, const char **values)
...
             }
 
+            if (HDstrlen(values[3]) > H5FD_ROS3_MAX_SECRET_TOK_LEN) {
+                if (show_progress) {
+                    HDprintf("  ERROR: token value too long\n");
+                }
+                ret_value = 0;
+                goto done;
+            }
+            HDmemcpy(token, values[3], (HDstrlen(values[3]) + 1));
```
jhendersonHDF: In the case where `HDstrlen(values[3]) == H5FD_ROS3_MAX_SECRET_TOK_LEN`, it occurs to me that this might cause a buffer overflow if `token` is a buffer of size `H5FD_ROS3_MAX_SECRET_TOK_LEN`. It might also be reading past memory depending on the size of `values[3]`. I think the correct check above may be `if (HDstrlen(values[3]) > H5FD_ROS3_MAX_SECRET_TOK_LEN - 1) {`.

jhendersonHDF:
@jwsblokland Thanks for all your work on this PR! I think it's very close with just a couple minor cleanup tasks. In general, I try to avoid doing anything similar to `char token[H5FD_ROS3_MAX_SECRET_TOK_LEN];` in new code (instead, preferring dynamic allocation) to prevent warnings from cropping up, but I don't want to hold this PR for that since this is useful functionality and warnings can always be cleaned up later.

src/H5FDs3comms.c
jwsblokland: Ah, you are right. It turns out that I misunderstood the man page of `snprintf()`. I always thought that the specified maximum size `size` is the maximum number of characters that will be set based on the format string and specified input strings while it is actually `size - 1` and the last position is always reserved for the NULL-termination. Anyway, this is very useful to know.

I will fix this.

src/H5FDros3.h
jwsblokland: Ah, I finally understand what you are trying to say. It makes sense to. Let me see what I can do about it. Furthermore, if desired I can move the newly defined structure to, for example, `htools.h`  file.

src/H5FDs3comms.c
jwsblokland: This has been fixed.

src/H5FDros3.h
jwsblokland: I have moved the structure to `h5tools_utils.h`.

src/H5FDs3comms.c
jwsblokland: The additional error checkings have been implemented.

tools/lib/h5tools_utils.c
jwsblokland: This has been fixed by making use of the newly introduced structure `H5FD_ros3_fapl_ext_t`.

jwsblokland:
@jhendersonHDF I hope with my last commit I have addressed all outstanding tasks for this PR. Note, I have changed the implementation of setting and getting the token. Now it is more inline with your code snippets found on the forum with one small change. The token string is fully managed by the HDF5 library. This is inline what it is done for the secret_id and secret_key strings.

src/H5FDros3.c
```diff
@@ -599,6 +608,254 @@ H5FD__ros3_fapl_free(void *_fa)
...
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "failed to check if property token exists in plist")
+
+    if (token_exists) {
+        if (H5P_get(plist, ROS3_TOKEN_PROP_NAME, &token_src) < 0)
+            HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "unable to get token value")
+
+        HDmemcpy(token_src, token, HDstrlen(token) + 1);
+    }
+    else {
+        token_src = HDmalloc(sizeof(char) * (H5FD_ROS3_MAX_SECRET_TOK_LEN + 1));
```
jhendersonHDF: Should have a check here to make sure `token_src` isn't NULL afterwards.

jhendersonHDF:
Hi @jwsblokland, just a last minor comment about an error check and also this PR should add an entry under release_docs/RELEASE.txt under the "New Features" -> "Library" section for the feature (by convention, new entries usually go above any existing ones) and then I think this PR can be merged. Thanks again for all your work on this PR!

src/H5FDros3.c
jwsblokland: This error check has been implemented.

jwsblokland:
> Hi @jwsblokland, just a last minor comment about an error check and also this PR should add an entry under release_docs/RELEASE.txt under the "New Features" -> "Library" section for the feature (by convention, new entries usually go above any existing ones) and then I think this PR can be merged. Thanks again for all your work on this PR!

@jhendersonHDF Thanks. It was fun to extend the HDF5 library like this. It gives me a bit more understanding of the library itself. Furthermore, I improved release_docs/RELEASE.txt and removed the unused static variable `token` from `h5dump.c`.

jhendersonHDF (approved):

lrknox (approved):

PR #3035, opened 2023-06-02
CMake: (fix) Szip / libaec filter

jwsblokland:
- Fixed the cmake configuration for the filter Szip / libaec. Now, the decoding of this filter is properly configured and included in the HDF5 library.


lrknox (approved):

hyoklee (approved):

CMakeFilters.cmake
```diff
@@ -152,14 +152,14 @@ if (HDF5_ENABLE_SZIP_SUPPORT)
...
-        set (H5_HAVE_SZLIB_H 1)
-        set (H5_HAVE_LIBSZ 1)
-        set (SZIP_INCLUDE_DIR_GEN ${SZIP_INCLUDE_DIR})
-        set (SZIP_INCLUDE_DIRS ${SZIP_INCLUDE_DIRS} ${SZIP_INCLUDE_DIR})
-        set (LINK_COMP_LIBS ${LINK_COMP_LIBS} ${SZIP_LIBRARIES})
-      endif ()
+    endif ()
+    if (SZIP_FOUND)
+      set (H5_HAVE_FILTER_SZIP 1)
+      set (H5_HAVE_SZLIB_H 1)
```
byrnHDF: Actually these first three set routines should be collated into the last if (SZIP_FOUND) block to have one source of these settings.

jwsblokland: Yes, that is a good idea. I forced pushed this modification.

Would it be possible to test szip/libaec compression via Github actions on a more structure level? In this case, it was a bit of luck I catch this issue while working on improving the current ROS3 implementation by switching on both the encoding and decoding functionality of th szip/libaec filter.

byrnHDF:
Looks great!

byrnHDF:
For consistency - would you mind moving/collating the similar three lines in the zlib code (see lines 94-96)

jwsblokland:
> For consistency - would you mind moving/collating the similar three lines in the zlib code (see lines 94-96)

Sure no problem. I have pushed the changes. I noticed that lines 113-115 could potentially be moved to the bottom if statement if we can assume that `${ZLIB_INCLUDE_DIRS}` is properly defined for that case. Currently, I do not make that assumption.

byrnHDF (approved):

byrnHDF (approved):

PR #3038, opened 2023-06-03
Add casts to type conversion to fix long dbl bug

derobins:
If a user buffer is misaligned in the type conversion code, memcpy could fail if the call is SSE-optimized by the compiler. This change adds uint8_t * casts so the compiler won't make optimistic assumptions about buffer alignment.


derobins:
This only showed up in a different PR (#2382) where we replace the library re-implementation of memcpy with the one in the standard library, but in the past we've frequently had bugs with long doubles that were "fixed" by selecting lower optimization levels for that platform. In the case I tripped over, the problem only shows up at high optimization levels, where the compiler uses SSE to perform the memcpy (SSE buffers have to be aligned on 16 bytes).

qkoziol (approved):

PR #3058, opened 2023-06-06
Implementation for mpio driver with selection I/O.

vchoi-hdfgroup:
This is a draft for preliminary review.


src/H5FDint.c
```diff
@@ -1555,7 +1562,7 @@ H5FD__write_selection_translate(H5FD_t *file, H5FD_mem_t type, hid_t dxpl_id, ui
             io_len = MIN(file_len[file_seq_i], mem_len[mem_seq_i]);
 
             /* Check if we're using vector I/O */
-            if (use_vector) {
+            if (!skip_vector_cb && use_vector) {
```
fortnern: Should use skip_vector_cb to determine the value of use_vector on line 1460, then there will be no need to keep checking it

src/H5FDint.c
```diff
@@ -2392,6 +2434,431 @@ H5FD_sort_vector_io_req(hbool_t *vector_was_sorted, uint32_t _count, H5FD_mem_t
...
+ *              (also reference H5FD__mpio_vector_build_types() in src/H5FDmpio.c)
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ * Programmer:
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5FD_selection_build_types(hbool_t io_op_write, uint32_t num_pieces, H5S_t **file_spaces, H5S_t **mem_spaces,
```
fortnern: Having this function in a shared place is a little strange. I was originally thinking that the code in H5FDmpio.c and H5Dmpio.c shouldn't be shared. If it's working though it might be worth it. I'll continue thinking about a way to make this function more self-contained/generic/maintainable

fortnern: After further thought, I think we should move this function to H5FDmpio.c, and return this functionality to being integrated into link_piece_collective_io (with the frees moved to before the I/O as suggested in my email). I think it'd also be a good idea to move the check for num_pieces out of the function (and into the read/write callbacks), and remove the size_i  and mpi_buds_base parameters (calculate them in the read/write_callbacks).

vchoi-hdfgroup: I will move this function to H5FDmpio.c as indicated.
And I will restore link_piece_collective_io() back to its original state in H5Dmpio.c (with the frees moved to before the IO), is that correct?

fortnern: That's correct, thanks!

src/H5FDint.c
vchoi-hdfgroup: Fixed

src/H5FDint.c
vchoi-hdfgroup: Changes made as indicated for H5FD_selection_build_types().
However, I still need to pass num_pieces to the routine because it is needed to malloc the arrays of mtype, ftype and displacement etc.  I can calculate mpi_bufs_base in the read/write callbacks, but I still need to pass the calculated info to H5FD_selection_build_types() for calculating displacements of the pieces.

vchoi-hdfgroup:
I am going to create a new PR based on the latest "develop" branch and incorporate the mpio driver changes to there.  I think it's easier that way instead of resolving all the conflicts/CI failures etc.  Then I will close this PR.

vchoi-hdfgroup:
See PR #3222

PR #3071, opened 2023-06-07
Remove unused cruft from bin dir

derobins:


derobins:
Brings the 1.12 bin directory in line with develop

lrknox (approved):

PR #3077, opened 2023-06-07
Add MSVC warning pragmas

derobins:


derobins:
Argh. I'll fix this on a real Windows machine and recreate.

PR #3080, opened 2023-06-07
Minor CMake fixes

derobins:
* Switch a few MATCHES to STREQUAL
* Remove a few variable dereferences


config/cmake/HDFCompilerFlags.cmake
```diff
@@ -337,9 +337,9 @@ endif ()
 #option (HDF5_ENABLE_ASSERTS "Determines whether NDEBUG is defined to control assertions." OFF)
 set (HDF5_ENABLE_ASSERTS "OFF" CACHE STRING "Determines whether NDEBUG is defined to control assertions (OFF NO YES)")
 set_property (CACHE HDF5_ENABLE_ASSERTS PROPERTY STRINGS OFF NO YES)
-if (HDF5_ENABLE_ASSERTS MATCHES "YES")
+if (HDF5_ENABLE_ASSERTS STREQUAL "YES")
```
byrnHDF: I'm not sure about this - except that the set_property does specify the allowed values.

byrnHDF (approved):

config/cmake/HDFCompilerFlags.cmake
jhendersonHDF: I thought MATCHES is the appropriate call here when comparing against a variable directly and STREQUAL is when you do `"${HDF5_ENABLE_ASSERTS}"`

byrnHDF: MATCHES allow regex type compares.

byrnHDF: So YES and NO forces setting a define while OFF leaves the defines as is (whatever that setting may be). Maybe that should be documented in the comment that OFF == default?

derobins: Do we want to match "YES" or do we tolerate "yes", "Yes", etc.? We should decide on one and ensure that it's used everywhere we check for "YES" or "NO".

lrknox (approved):

config/cmake/HDFCompilerFlags.cmake
byrnHDF: By Convention I always use capitals for options (consistency).
In this instance, MATCHES (which allows regex) and STREQUAL behave identical.

byrnHDF: There is a difference between string options (this case) and boolean options. Strings must consider what we talked about. Booleans follow CMake rules.

PR #3081, opened 2023-06-07
Minor cleanup of H5detect.c

derobins:
* Align output
* Tidy
* Remove HD prefixes
* Remove author/date lines from output

### DIFF
--- a/src/H5detect.c
+++ b/src/H5detect.c
@@ -357,8 +357,8 @@ H5T__init_native(void)\n\
     dt->shared->u.atomic.prec = %d;\n\
     dt->shared->u.atomic.lsb_pad = H5T_PAD_ZERO;\n\
     dt->shared->u.atomic.msb_pad = H5T_PAD_ZERO;\n",
-                d[i].offset,        /* Offset */
-                d[i].precision);    /* Precision */
+                d[i].offset,     /* Offset */
+                d[i].precision); /* Precision */
 
         /* The part unique to floating point types */
         fprintf(rawoutstream, "\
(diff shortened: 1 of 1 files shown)


src/H5detect.c
```diff
@@ -174,7 +169,7 @@ precision(detected_t *d)
         }                                                                                                    \
         fix_order(sizeof(TYPE), _last, INFO.perm, (const char **)&_mesg);                                    \
                                                                                                              \
-        if (!HDstrcmp(_mesg, "VAX"))                                                                         \
+        if (!strcmp(_mesg, "VAX"))                                                                           \
```
hyoklee, line 171: Is there a reason that `strncmp()` is not used?

byrnHDF (approved):

src/H5detect.c
derobins: Age?

src/H5detect.c
```diff
@@ -22,28 +22,26 @@ static const char *FileHeader = "\n\
  * distribution tree, or in https://www.hdfgroup.org/licenses.               *\n\
  * If you do not have access to either file, you may request a copy from     *\n\
  * help@hdfgroup.org.                                                        *\n\
- * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *";
+ *                                                                           *\n\
```
qkoziol: Perhaps Github is presenting this in an unusual manner, but I'm not certain that you wanted two '*' on this line?

src/H5detect.c
```diff
@@ -722,21 +694,9 @@ find_bias(unsigned int epos, unsigned int esize, int *perm, void *_a)
...
-    const char *s;
-#ifdef H5_HAVE_GETPWUID
-    struct passwd *pwd = NULL;
-#else
-    int pwd      = 1;
-#endif
-    static const char *month_name[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun",
-                                       "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
-    static const char *purpose      = "\
+    int                i;
```
qkoziol: I'd like to leave the date+time, hostname, etc. information about where the file was generated.  It can be useful provenance information for archiving and debugging.

derobins: I disagree. It's a lot of platform-specific hax for a "feature" of very limited value. I would be very, very, very surprised if *anyone* in the past twenty years has tried to figure out when they configured HDF5 by peeking in generated H5T files. I can also imagine that some people are not going to be super excited to expose internal machines and usernames via HDF5 source code. And we don't do this in other generated files.

src/H5detect.c
derobins: I'll check it out. I recall it looking fine when I inspected the output.

qkoziol (approved):

PR #3086, opened 2023-06-09
Add support for authenticating access to S3 buckets using temporary credentials

aivazis:
This PR modifies the ROS3 VFD to provide support for presenting a `session_token` to AWS.

There is also some minor restructuring of the DEBUG macros in both `H5FDros3.c` and `H5FDs3comms.c` to help with API tracing and access pattern tracking.


jwsblokland:
Maybe worthwhile to take a look at PR #3030 - ROS3: (feature) Temporary security credentials and HDF forum page https://forum.hdfgroup.org/t/ros3-temporary-security-credentials/11135. Both of them is exactly about the same thing you want to achieve but taking into account not breaking the API.

aivazis:
i did. as it happens I'm on a webex with Aleksandar as i write this. the code here evolves the existing `fapl_t` by adding a new field for the `session_token` and bumping up its embedded version number. so it may be complementary to the approach in #3030

src/H5FDros3.h
```diff
@@ -70,18 +70,20 @@
...
+#define H5FD_ROS3_MAX_SECRET_KEY_LEN    128
+#define H5FD_ROS3_MAX_SESSION_TOKEN_LEN 4096
 
 typedef struct H5FD_ros3_fapl_t {
     int32_t version;
     hbool_t authenticate;
     char    aws_region[H5FD_ROS3_MAX_REGION_LEN + 1];
     char    secret_id[H5FD_ROS3_MAX_SECRET_ID_LEN + 1];
     char    secret_key[H5FD_ROS3_MAX_SECRET_KEY_LEN + 1];
+    char    session_token[H5FD_ROS3_MAX_SESSION_TOKEN_LEN + 1];
```
derobins, line 86: Changing a struct in a public header breaks binary compatibility so this change could not be merged downstream to 1.14, etc.

aivazis: the change comes with a bump in `H5FD_CURR_ROS3_FAPL_T_VERSION`, so it is easy to detect clients that have not been recompiled against the updated headers/library. we should be able to restore binary compatibility. shall we walk through the use cases, or are you guys happy to say this is a 1.15 fix?

jhendersonHDF:
Hi @aivazis, this looks fairly similar to @jwsblokland's PR mentioned previously, where we decided to add a new H5P routine so that the previous API doesn't change and the PR could be merged rather than waiting for a major release of HDF5. Does this PR include functionality that is different from that PR or is it the same session token functionality but added as part of the fapl structure?

aivazis:
well, i'm not particularly strongly invested in exactly how this capability hole is plugged. my users consider support for temporary credentials as critical, so i took what to me looked like the natural path: the deficient data structure has a built-in versioning mechanism, designed to handle the evolution of its schema, which is precisely what this is about. i added logic so that clients that have not been recompiled with the new header and present a version 1 structure to the api are supported by never dereferencing the missing field. this looks like "binary compatibility" to me. i've already successfully patched 1.10, 1.12, and 1.14, built the libraries and tools, and tested clients that had been compiled against libraries without the patch without noticing any bad behavior. but maybe i'm missing something.

i've added initializers for the new field to all the tools that touch it, and repaired all the test drivers, except for `tools/libtest/h5tools_test_utils.c` that looked like it deserved a bit more thinking.

in addition to support for the missing `session_token`, this PR has a bunch of fixes that remove core dumps when debugging and tracing is enabled. from trying to print the `purl` in the stats report after it has been freed, to bad format specifiers when generating reports and messages. also, the initialization of the stats bins now happens only the first time the VFD initializer is called. if you go another way with this PR, we should cherry pick these fixes and apply them to your branch.

I'm happy to help any which way you think is the right way to move forward.

lrknox:
PR #3030 has been merged to develop.  We can consider any parts of this PR that go beyond 3030 in a separate PR to develop.

PR #3087, opened 2023-06-09
speed-up building HDF5

brtnfld:
Don't build for external package actions more HDF5 options than what is needed.

### DIFF
--- a/.github/workflows/netcdf.yml
+++ b/.github/workflows/netcdf.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Install HDF5
       run: |
         ./autogen.sh
-        ./configure --prefix=/usr/local --disable-static --enable-shared --enable-hl --with-szlib
+        ./configure --prefix=/usr/local --disable-tests --disable-static --enable-shared --enable-hl --with-szlib
         make -j
         sudo make install -j
     - name: Checkout netCDF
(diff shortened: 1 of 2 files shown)


brtnfld:
I thought it was decided that external packages should not be built for each PR.

.github/workflows/netcdf.yml
```diff
@@ -2,22 +2,8 @@ name: netCDF
...
-      - 'ACKNOWLEDGEMENTS'
-      - 'COPYING**'
-      - '**.md'
-
-# Using concurrency to cancel any in-progress job or run
-concurrency:
-  group: ${{ github.workflow }}-${{ github.sha || github.event.pull_request.number }}
-  cancel-in-progress: true
+  schedule:
+    - cron: "6 0 * * *"
```
derobins: 1. If this is going to be tested nightly, we should probably also test HDF-EOS5 nightly
2. These integration tests take very little time to run. Why not run them on every commit? It's a LOT easier to block bad commits than it is to track people down to revert their changes if a nightly test fails.

glennsong09 (approved):

.github/workflows/netcdf.yml
brtnfld: Reverted to test with each PR.

derobins (approved):

PR #3088, opened 2023-06-09
Address memory issues when copying empty enums

derobins:
When copying an empty enum type (including implicitly, as when an enum is contained in a compound type), the library would allocate 0-size blocks of memory and attempt to memcpy 0 bytes from NULL pointers, which are undefined behavior. In debug mode, the library would raise an assert in H5MM.

The library now avoid undefined memory operations when copying empty enum types and a test that copies empty enums has been added.

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -3623,21 +3623,28 @@ H5T__complete_copy(H5T_t *new_dt, const H5T_t *old_dt, H5T_shared_t *reopened_fo
                  * of each new member with copied values. That is, H5T_copy() is a
                  * deep copy.
                  */
-                if (NULL == (new_dt->shared->u.enumer.name =
-                                 H5MM_malloc(new_dt->shared->u.enumer.nalloc * sizeof(char *))))
-                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTALLOC, FAIL, "enam name array memory allocation failed")
-                if (NULL == (new_dt->shared->u.enumer.value =
-                                 H5MM_malloc(new_dt->shared->u.enumer.nalloc * new_dt->shared->size)))
-                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTALLOC, FAIL,
-                                "enam value array memory allocation failed")
-                H5MM_memcpy(new_dt->shared->u.enumer.value, old_dt->shared->u.enumer.value,
-                            new_dt->shared->u.enumer.nmembs * new_dt->shared->size);
-                for (i = 0; i < new_dt->shared->u.enumer.nmembs; i++) {
-                    if (NULL == (s = H5MM_xstrdup(old_dt->shared->u.enumer.name[i])))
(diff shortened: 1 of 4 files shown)


src/H5T.c
```diff
@@ -3623,21 +3623,28 @@ H5T__complete_copy(H5T_t *new_dt, const H5T_t *old_dt, H5T_shared_t *reopened_fo
...
+                        if (NULL == (s = H5MM_xstrdup(old_dt->shared->u.enumer.name[i])))
+                            HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL,
+                                        "can't copy string for enum value's name")
+                        new_dt->shared->u.enumer.name[i] = s;
+                    }
+                }
+                else {
+                    /* Empty enum */
+                    HDmemset(&new_dt->shared->u.enumer, 0, sizeof(H5T_enum_t));
+                }
```
derobins, line 3647: This is the real change. We just memset the struct instead of performing memory options that the C standard says are 'undefined behavior'.

src/H5Tcompound.c
```diff
@@ -469,10 +469,12 @@ H5T__insert(H5T_t *parent, const char *name, size_t offset, const H5T_t *member)
...
     /* Add member to end of member array */
     idx                                       = parent->shared->u.compnd.nmembs;
-    parent->shared->u.compnd.memb[idx].name   = H5MM_xstrdup(name);
     parent->shared->u.compnd.memb[idx].offset = offset;
     parent->shared->u.compnd.memb[idx].size   = total_size;
-    parent->shared->u.compnd.memb[idx].type   = H5T_copy(member, H5T_COPY_ALL);
+    if (NULL == (parent->shared->u.compnd.memb[idx].name = H5MM_xstrdup(name)))
+        HGOTO_ERROR(H5E_DATATYPE, H5E_CANTALLOC, FAIL, "couldn't duplicate name string")
+    if (NULL == (parent->shared->u.compnd.memb[idx].type = H5T_copy(member, H5T_COPY_ALL)))
+        HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "couldn't copy datatype")
```
derobins, line 477: A little extra defensive programming here...

test/enum.c
derobins, line 1: This test was kind of a mess, so I cleaned it up.

test/enum.c
```diff
@@ -729,61 +708,114 @@ test_funcs(void)
...
- *              Tuesday, December 22, 1998
+ * Purpose:     Test that copying an empty enum works, including implicitly
+ *              when copying compound datatypes containing empty enums
  *
+ * Return:      Success:    0
+ *              Failure:    1
  *-------------------------------------------------------------------------
  */
+static int
+test_compound_insert_empty_enum(void)
```
derobins, line 728: This test ensures that copying empty enums works, whether directly or as a part of a compound type.

byrnHDF (approved):

fortnern (approved):

PR #3090, opened 2023-06-09
Build the ros3 VFD in GitHub actions

derobins:


.github/workflows/main.yml
```diff
@@ -109,6 +110,7 @@ jobs:
             parallel: OFF
             mirror_vfd: ON
             direct_vfd: ON
+            ros3_vfd: OFF
```
derobins, line 113: CMake has a CURL problem in the runner that I do not see when running locally

hyoklee (approved):

byrnHDF (approved):

PR #3094, opened 2023-06-10
Fix misc warnings on Windows

derobins:
* Debug functionality where pointers were munged into longs (which are only 32-bits on Windows)
* Fix a missing cast in Wstrcasestr_wrap()

### DIFF
--- a/src/H5Iint.c
+++ b/src/H5Iint.c
@@ -389,9 +389,9 @@ H5I__mark_node(void *_info, void H5_ATTR_UNUSED *key, void *_udata)
 #ifdef H5I_DEBUG
                     if (H5DEBUG(I)) {
                         HDfprintf(H5DEBUG(I),
-                                  "H5I: discard type=%d obj=0x%08lx "
+                                  "H5I: discard type=%d obj=%p "
                                   "failure ignored\n",
-                                  (int)udata->type_info->cls->type, (unsigned long)(info->object));
+                                  (int)udata->type_info->cls->type, info->object);
                     }
 #endif /* H5I_DEBUG */
 
@@ -412,9 +412,9 @@ H5I__mark_node(void *_info, void H5_ATTR_UNUSED *key, void *_udata)
 #ifdef H5I_DEBUG
                     if (H5DEBUG(I)) {
                         HDfprintf(H5DEBUG(I),
-                                  "H5I: free type=%d obj=0x%08lx "
+                                  "H5I: free type=%d obj=%p "
                                   "failure ignored\n",
-                                  (int)udata->type_info->cls->type, (unsigned long)(info->object));
+                                  (int)udata->type_info->cls->type, info->object);
                     }
 #endif /* H5I_DEBUG */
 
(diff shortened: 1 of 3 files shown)


src/H5Iint.c
```diff
@@ -389,9 +389,9 @@ H5I__mark_node(void *_info, void H5_ATTR_UNUSED *key, void *_udata)
 #ifdef H5I_DEBUG
                     if (H5DEBUG(I)) {
                         HDfprintf(H5DEBUG(I),
-                                  "H5I: discard type=%d obj=0x%08lx "
+                                  "H5I: discard type=%d obj=%p "
                                   "failure ignored\n",
-                                  (int)udata->type_info->cls->type, (unsigned long)(info->object));
+                                  (int)udata->type_info->cls->type, info->object);
```
derobins, line 394: unsigned long is 32-bits on Windows and can't hold an address. You can't use padding with %p, so I ripped it out.

src/H5T.c
```diff
@@ -1504,9 +1504,9 @@ H5T_top_term_package(void)
                         if (H5DEBUG(T)) {
                             HDfprintf(H5DEBUG(T),
                                       "H5T: conversion function "
-                                      "0x%08lx failed to free private data for "
+                                      "0x%016zx failed to free private data for "
                                       "%s (ignored)\n",
-                                      (unsigned long)(path->conv.u.app_func), path->name);
+                                      (size_t)path->conv.u.app_func, path->name);
```
derobins, line 1509: Another place where an address is munged into a long, which is too small on Windows. This case is trickier since you can't really print function pointers portably, though. The best I can come up with is casting to size_t and printing that as hex.

src/H5system.c
```diff
@@ -979,7 +979,7 @@ char *
 Wstrcasestr_wrap(const char *haystack, const char *needle)
 {
     if (needle && !*needle)
-        return haystack;
+        return (char *)haystack;
```
derobins, line 982: We're kind of cheating here, but this is what everyone else does when they implement this behavior.

src/H5T.c
jhendersonHDF: How much does portably printing matter here? Seems like using %p would be the most sane choice

byrnHDF: I would opt for %p.

byrnHDF (approved):

src/H5T.c
derobins: You can't use %p to print function pointers without warnings. %p expects a void pointer and a function pointer is not a void pointer since it doesn't point to an object.

jhendersonHDF: Ah I'd overlooked the _function_ part. Agreed on the approach here then. I'd suggest casting to uintptr_t instead of size_t, but uintptr_t is an optional type.

jhendersonHDF (approved):

src/H5T.c
derobins: Unfortunately, uintptr_t was similarly cranky

PR #3096, opened 2023-06-10
Make some changes to configure.ac to support MinGW

derobins:
Adds some H5pubconf.h entries and cpp flags for building on MinGW using the Autotools.

Also updates the Windows-related H5pubconf.h comments to be more accurate in CMake.

### DIFF
--- a/config/cmake/H5pubconf.h.in
+++ b/config/cmake/H5pubconf.h.in
@@ -14,13 +14,13 @@
 #ifndef H5_CONFIG_H_
 #define H5_CONFIG_H_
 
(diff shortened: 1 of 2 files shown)


derobins:
MinGW doesn't work yet since it'll die at the part where it runs H5detect, etc., but this brings it closer.

And yes, I'm still planning on retiring the Autotools at some point. I'm just investigating what MinGW does and I know the Autotools better. Might as well commit my changes as long as they are useful.

Still unresolved: getenv_s can't be found on a default install on Ubuntu. Maybe they don't build and install that?

byrnHDF (approved):

PR #3097, opened 2023-06-10
Remove long double conversion work-arounds

derobins:
Several options and public symbols that were provided to paper
      over non-standard long double conversions between signed/unsigned
      long and long long values were removed from the Autotools and
      CMake. These were added twenty years ago, when C99 and 64-bit
      platforms were less common and are no longer needed.

      Autotools:
        --enable-dconv-accuracy

      CMake:
        HDF5_WANT_DATA_ACCURACY

      H5pubconf.h symbols:
        H5_WANT_DATA_ACCURACY
        H5_LDOUBLE_TO_LONG_SPECIAL
        H5_LONG_TO_LDOUBLE_SPECIAL
        H5_LDOUBLE_TO_LLONG_ACCURATE
        H5_LLONG_TO_LDOUBLE_CORRECT
        H5_DISABLE_SOME_LDOUBLE_CONV


derobins:
Note that:

H5_LDOUBLE_TO_LLONG_ACCURATE
H5_LLONG_TO_LDOUBLE_CORRECT

Were the only symbols that had any effect on the library (along with the H5_WANT_DATA_ACCURACY symbol). The others just papered over tests in dt_arith.

derobins:
Looking at the comments, all of the platforms of concern are ancient (IRIX, HP-UX, macOS 10.4 (circa 2005), Power6). Given that long double is C99 (required for building HDF5), I would be surprised if these hacks were required anywhere anymore.

byrnHDF (approved):

bmribler (approved):

fortnern (approved):
Approved as long as it passes on all supported platforms

qkoziol (approved):
LGTM

qkoziol:
> Approved as long as it passes on all supported platforms

Same

PR #3098, opened 2023-06-10
Remove code paths where `long long` < 8 bytes

derobins:
The C99 standard says that a long long is 8+ bytes. This change removes a few parts of the code where we assume that a long long could be less than that.

### DIFF
--- a/test/big.c
+++ b/test/big.c
@@ -11,12 +11,6 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
 /*
- * Programmer:  Robb Matzke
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

bmribler (approved):

test/big.c
```diff
@@ -62,12 +56,6 @@
 
 #define MAX_TRIES 100
 
-#if H5_SIZEOF_LONG_LONG >= 8
```
fortnern, line 65: Maybe add a compile time assert?

derobins: No need. We require a C99 compiler and the standard requires this.

PR #3100, opened 2023-06-11
build(configure.ac): skip h5fuse.sh if subfiling VFD is not used (#3089)

### LINKED ISSUE #3089: Don't create `utils/subfiling_vfd/h5fuse.sh` if `--enable-subfiling-vfd` is not used.
**Describe the bug**

I see this message although I did not enable `subfiling_vfd`.

`config.status: creating utils/subfiling_vfd/h5fuse.sh`

**Expected behavior**

  I should not see the above message.

 - HDF5 version: Develop
 - OS and version: Ubuntu 22.04
 - Compiler and version: n/a
 - Build system: Autotools

hyoklee:
Close #3089.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4165,7 +4165,9 @@ AC_CONFIG_FILES([src/libhdf5.settings
                  hl/fortran/examples/Makefile
                  hl/fortran/examples/run-hlfortran-ex.sh])
 
-AC_CONFIG_FILES([utils/subfiling_vfd/h5fuse.sh], [chmod +x utils/subfiling_vfd/h5fuse.sh])
+if test "X$SUBFILING_VFD" = "Xyes"; then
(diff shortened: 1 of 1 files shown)


derobins (changes requested):
h5fuse isn't special. If we do one of these, we should do them all.

fortnern (approved):

hyoklee:
> h5fuse isn't special. If we do one of these, we should do them all.

Please create another issue. Let's merge this one first and finish the rest one by one.

jhendersonHDF:
Since the h5fuse script doesn't need the Subfiling VFD to be built and since I can definitely see use cases where a user would want to work with Subfiling VFD subfiles without having the Subfiling VFD available, I don't think we should make generating of the script dependent on whether the Subfiling VFD is built.

derobins:
This is a useful tool even without the subfiling VFD so we'll leave it in

PR #3103, opened 2023-06-11
Replace H5detect with run-time detection

derobins:
H5detect is a separate program that is run at build-time to determine floating-point characteristics of the target system, which are saved in a generated H5Tinit.c file that is then compiled into the library.

This program has historically been a huge impediment to cross-compiling and there is no real reason to not do these checks at library startup.


src/H5Tinit_float.c
derobins, line 1: I did not change these macros and functions very much. I just added normal HDF5 library boilerplate.

src/CMakeLists.txt
```diff
@@ -1036,8 +1037,7 @@ if (HDF5_GENERATE_HEADERS)
 endif ()
 
 #-----------------------------------------------------------------------------
-# Setup the H5detect utility which generates H5Tinit with platform
-# specific type checks inside
+# Set up source generators
 #-----------------------------------------------------------------------------
 if (HDF5_USE_PREGEN)
```
byrnHDF, line 1042: I think this option and related can also be removed. It was used to put a pre-gen H5Tinit in to skip H5detect.

byrnHDF: HDF5_GENERATED_SOURCE_DIR can be replaced with HDF5_SRC_BINARY_DIR.

derobins: I left anything in that looked like it could be needed for H5make_libsettings (which is next on my list, so it'll go away soon).

src/CMakeLists.txt
```diff
@@ -1221,12 +1126,12 @@ if (BUILD_SHARED_LIBS)
...
 #-----------------------------------------------------------------------------
-# Add H5Tinit source to build - generated by H5detect/CMake at configure time
+# Add generated source to build
 #-----------------------------------------------------------------------------
 if (BUILD_STATIC_LIBS)
-  set (gen_SRCS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c)
+  set (gen_SRCS ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c)
   add_custom_target (gen_${HDF5_LIB_TARGET} ALL
-    DEPENDS ${lib_prog_deps} ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1 ${HDF5_SRC_BINARY_DIR}/gen_SRCS.stamp2
+    DEPENDS ${lib_prog_deps} ${HDF5_SRC_BINARY_DIR}/gen_SRCS.stamp2
```
byrnHDF, line 1134: lib_prog_deps is just H5make_libsettings and stamp2 is just H5lib_settings.c
That would simplify this and the previous add_custom_commands.

byrnHDF:
Do a search for H5detect, there are two files in the bin dir that need changes. Also there is a comment in config/cmake/HDFMacros.cmake

byrnHDF:
knl_H5detect.sl.in.cmake  probably can just be deleted

byrnHDF:
Also in release_docs the files INSTALL_CMake.txt and README_HPC have instructions releated to HDF5_USE_PREGEN and HDF5_USE_PREGEN_DIR. Options that can be deleted with H5detect gone.

byrnHDF:
HDF5_BATCH_H5DETECT is also an option that only refs H5detect

release_docs/RELEASE.txt
```diff
@@ -47,6 +47,11 @@ New Features
 
     Configuration:
     -------------
+    - Running H5detect is no longer required for cross-compiling
+
+      The functionality of H5detect is now exercised at library startup,
+      so H5detect has been removed.
```
byrnHDF, line 53: Of course you should mention that this change also eliminates the three CMake variables: HDF5_USE_PREGEN, HDF5_USE_PREGEN_DIR, and HDF5_BATCH_H5DETECT

derobins:
Closing since #3104 gets rid of both programs

Issue #1948, opened 2022-07-29
cross-compiling from x86_64-linux-musl to  x86_64-linux-gnu


Alexander-Barth:
I am trying to provide a HDF5 binary for julia users. Julia's build tree uses cross-compilation from a  x86_64-linux-musl build environment. 

When compiling HDF5 1.12.2 or current develop branch (33f6ade6e77ed85442f001fa32924127455aca39) to target x86_64-linux-gnu (same architecture/OS but different libc):

```bash
./configure --prefix=/workspace/destdir --build=x86_64-linux-musl --host=x86_64-linux-gnu --enable-shared \
   --disable-static --with-zlib=/workspace/destdir --with-pthread=yes --enable-threadsafe \
   --enable-build-mode=production --enable-using-memchecker --enable-ros3-vfd \
   --enable-unsupported --enable-tests=no
```

I get this error:

```
checking maximum decimal precision for C... configure: error: in `/workspace/srcdir/hdf5-1.12.2':
configure: error: cannot run test program while cross compiling
```

Would it be possible to by-pass this test by e.g. setting an environment variable? I guess that this is the purpose of this patch from conda-forge:
https://github.com/conda-forge/hdf5-feedstock/blob/main/recipe/patches/osx_cross_configure.patch

Reference:
https://github.com/JuliaPackaging/Yggdrasil/issues/4511#issuecomment-1199281217

mkitti:
Once upon a time, @stevengj started a [fork to try to address cross compilation](https://github.com/stevengj/hdf5/commit/6b0545585282647bb8d7be8744d73745b000bdd3) (for Yggdrasil and Julia packaging).

His approach was to use autoconf's cross compilation features. Essentially, the compiler already knows the answers to many of the questions being asked during by these "test programs". Thus rather than executing the test programs, we just have to compile and inspect the resulting machine code.

The idea was to use the `AC_CHECK_SIZEOF` or `AC_COMPUTE_INT` macros to sort this out.
https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Generic-Compiler-Characteristics.html

I don't have much experience with this, but I would be interested in picking this up and pushing it further.

derobins:
Can this issue be closed?

* The long double checks are gone
* AC_RUN_IFELSE in the Autotools has defaults for everything that is left (just szip encoding and checking pthread_attr_setscope)
* H5detect is gone
* H5make_libsettings is gone

Cross compiling should basically just work now

[...]

Issue #1410, opened 2022-01-29
Support Android Targets


leleliu008:
https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.13/hdf5-1.13.0/src/hdf5-1.13.0.tar.bz2

```
cmake \
        -DHDF5_BUILD_DOC=OFF \
        -DBUILD_SHARED_LIBS=ON \
        -DBUILD_STATIC_LIBS=ON \
        -DBUILD_STATIC_EXECS=ON \
        -DTEST_LFS_WORKS_RUN=ON
        -S . \
        -B build.d
```

![27](https://user-images.githubusercontent.com/3040852/151649358-cb8667c9-2149-421b-b891-bd4a98a5a84f.PNG)

It seems that this project do not support cross-compiling. how can I skip these build time checks? Thanks.

byrnHDF:
You can pre-populate the file needed - you need to do a search. However we do have CMake support in the src/CMakeLists.txt file for using the file. About line 1028 HDF5_USE_PREGEN variable.

jebos:
@byrnHDF can you point to a documentation on how to 
I'm trying to just compile the current develop branch of hdf5 for a current android on windows using the latest NDK ect.

I'm a bit lost in regards of 
a) where to get the values for this file (how to generate)
and
b) how the file needs to look like (are there examples?)

Can you give me a hint into the correct direction. Thanks.

byrnHDF:
Some references first:
CMake toolchains: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html
aarch64 on Ubuntu: https://discourse.cmake.org/t/cross-compile-for-aarch64-on-ubuntu/2161/10
Using clang(includes an addendum for gcc) on windows to arm: https://stackoverflow.com/questions/54539682/how-to-set-up-cmake-to-cross-compile-with-clang-for-arm-embedded-on-windows?rq=1
CMake NDK: https://developer.android.com/ndk/guides/cmake
 

A note found during search:  make sure you set the CMAKE_FIND_ROOT_PATH variable to a path where you have an exact copy of the root filesystem you have on your target device (with libraries and binaries pre-compiled for the target processor).

byrnHDF:
Initial release_docs file: INSTALL_cross_compile.txt
`
//       HDF5 Build and Install Instructions for Cross Compiling       

Cross-compiling has several consequences for CMake:
    CMake cannot automatically detect the target platform.
[...]

Issue #1203, opened 2021-11-16
Cross-compilation support


musm:
Lack of cross-compilation from linux to other platforms would significantly help distribution of binaries in the Julia ecosystem. This has been a long-standing issue. See e.g. https://github.com/JuliaPackaging/Yggdrasil/pull/567, https://forum.hdfgroup.org/t/cross-compiling-for-windows/6735, to mention two threads. There have been several attempts at it including the fork https://github.com/stevengj/hdf5 and in addition Steven Varga has worked on looking at this in the past.

musm:
any status or thoughts?

gnuoyd:
On Mon, Feb 28, 2022 at 10:08:25AM -0800, Mustafa M wrote:
> any status or thoughts?

HDF5 is on the path to easier cross-compilation, but I cannot forecast
when the work will be completed.  I have a couple of outstanding PRs
that will help, [PR \#1400](https://github.com/HDFGroup/hdf5/pull/1400)
and [PR \#1426](https://github.com/HDFGroup/hdf5/pull/1426).

Dave

musm:
Thanks for the update, looks quite promising and glad to see progress

simonbyrne:
Any updates on this? @gnuoyd I noticed both PRs have now been merged: what other pieces need to get done?

byrnHDF:
#1410 also is related and I will document the CMake process there and then create a cross-compile document in the release_docs folder.

giordano:
@byrnHDF trying to build https://github.com/HDFGroup/hdf5/commit/eac2cd54e209cfa9556174f3fc1a592533aa64ad (most recent commit in `develop` as of now), I still get plenty of `TRY_RUN`s:
```
[23:55:44] CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:
[23:55:44]    TEST_LFS_WORKS_RUN (advanced)
[23:55:44] For details see /workspace/srcdir/hdf5/build/TryRunResults.cmake
[...]
[23:55:52] CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:
[23:55:52]    H5_LDOUBLE_TO_LONG_SPECIAL_RUN (advanced)
[23:55:52]    H5_LDOUBLE_TO_LONG_SPECIAL_RUN__TRYRUN_OUTPUT (advanced)
[...]

Issue #566, opened 2021-04-21
Something is still introspecting `sizeof (long double)` at compile time, breaking macOS Universal Binary builds


seanm:
Even after I made various fixes, I suspect something is still introspecting `sizeof (long double)` at compile time.

The problem occurs when I build HDF5 as a Universal Binary on macOS (that is, for both x86_64 and arm64 architectures).  On x86_64 `sizeof (long double)` is 16.  On arm64 it is 8.  If I build ***on*** and Intel Mac, then ***run*** the resulting library on a arm64 Mac, it fails at runtime here:

https://github.com/HDFGroup/hdf5/blob/3707675c0376e3866bc6eb7b926c9e7e2f9b3bec/src/H5Tconv.c#L735-L736

which evaluates to `4 != 4 || 16 != 8`.  i.e. `dt->shared->size` is 16 when it should be 8.  I suspect this is because the machine that ***built*** the library is x86_64 where `long double` is 16.

So far I haven't found where this could be happening... could anyone help point me where it might be coming from?

Thanks.

seanm:
I made some progress.  `H5detect` creates a file `H5Tinit.c` and its contents differ based on build machine introspection.

qkoziol:
> I made some progress. `H5detect` creates a file `H5Tinit.c` and its contents differ based on build machine introspection.

Yes, that is its purpose. :-)

seanm:
Indeed. :)

So I guess there are two ideas:

1. Re-evalutate the whole idea of introspecting the build machine. Such introspection of course makes cross-compilation hard/impossible (and Universal Binaries are basically cross-compiling). Perhaps build-time introspection can be replaced with compile-time instead.

2. Detect that there are two architectures listed in `CMAKE_OSX_ARCHITECTURES` and create two `H5Tinit.c` files. Or similarly create a single `H5Tinit.c` that has preprocessor switches that give a final result like:

```
#if __x86_64__
    dt->shared->type = H5T_FLOAT;
    dt->shared->size = 16;
#elif __arm64__
    dt->shared->type = H5T_FLOAT;
    dt->shared->size = 8;
#endif
```

My feeling is that 1 would be better, but 2 would be easier.  Thoughts?

seanm:
Is `H5detect` the only thing that introspects the building machine?

As a test, I modified `H5Tinit.c` by hand to have the `#if / #elif` as above, and now everything seems to be finally working for me as a Universal Binary. So this at least seems to be my last problem!

[...]

PR #3104, opened 2023-06-11
Remove H5detect and H5make_libsettings

derobins:
H5detect is a separate program that is run at build-time to determine floating-point characteristics of the target system, which are saved in a generated H5Tinit.c file that is then compiled into the library.

H5make_libsettings simply populated a library string with the contents of the libhdf5.settings file. This functionality has been replaced with a template file.

These programs have historically been a huge impediment to cross-compiling. There is no reason to not do the H5detect checks at library startup like the other datatypes and the config string is easily handled via a template.


derobins:
This should fix (or greatly mitigate) #566, #1203, #1410, and #1948. I intend for this to be available in 1.14.2, which will release in the fall. Not sure if we'll move it to 1.10 or 1.12.

derobins:
Fortran may still have problems cross-compiling as I think it still runs programs to determine behavior. I'll have to confer with @brtnfld.

CMakeLists.txt
```diff
@@ -1225,9 +1225,8 @@ endif ()
 configure_file (${HDF_RESOURCES_DIR}/H5pubconf.h.in ${HDF5_SRC_BINARY_DIR}/H5pubconf.h @ONLY)
 
 #-----------------------------------------------------------------------------
-# Options for use by cross compiling and toolchains
+# Generate the H5build_settings.c file
 #-----------------------------------------------------------------------------
-option (HDF5_USE_PREGEN "Use pre-generated Files" OFF)
-option (HDF5_BATCH_H5DETECT "Use a batch command for running h5detect" OFF)
+configure_file (${HDF5_SOURCE_DIR}/src/H5build_settings.cmake.c.in ${HDF5_SRC_BINARY_DIR}/H5build_settings.c @ONLY)
```
byrnHDF: Why not in the src/CMakeLists.txt file?

src/CMakeLists.txt
```diff
@@ -1036,14 +1038,9 @@ if (HDF5_GENERATE_HEADERS)
...
-# Setup the H5detect utility which generates H5Tinit with platform
-# specific type checks inside
+# Set up source generators
 #-----------------------------------------------------------------------------
-if (HDF5_USE_PREGEN)
-  set (HDF5_GENERATED_SOURCE_DIR ${HDF5_USE_PREGEN_DIR})
-else ()
-  set (HDF5_GENERATED_SOURCE_DIR ${HDF5_SRC_BINARY_DIR})
-endif ()
+set (HDF5_GENERATED_SOURCE_DIR ${HDF5_SRC_BINARY_DIR})
```
byrnHDF: HDF5_GENERATED_SOURCE_DIR can just be deleted because it is just HDF5_SRC_BINARY_DIR

src/CMakeLists.txt
```diff
@@ -1078,155 +1075,13 @@ if (LOCAL_BATCH_TEST)
...
-      ARGS       -E copy_if_different H5lib_settings.c shared/H5lib_settings.c
-      COMMAND    ${CMAKE_COMMAND}
-      ARGS       -E touch shared/shared_gen_SRCS.stamp2
-      DEPENDS H5make_libsettings gen_SRCS.stamp2
-      WORKING_DIRECTORY ${HDF5_SRC_BINARY_DIR}
-      COMMENT    "Copy H5lib_settings.c to shared folder"
-  )
-endif ()
-
 #-----------------------------------------------------------------------------
```
byrnHDF, line 1078: This is where the configure_file statement from /CMakeLists.txt:1230 should be located.

src/CMakeLists.txt
```diff
@@ -1078,155 +1075,13 @@ if (LOCAL_BATCH_TEST)
...
-# Add H5Tinit source to build - generated by H5detect/CMake at configure time
+# Set up library builds
 #-----------------------------------------------------------------------------
 if (BUILD_STATIC_LIBS)
-  set (gen_SRCS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c)
+  set (gen_SRCS ${HDF5_SRC_BINARY_DIR}/H5build_settings.c)
   add_custom_target (gen_${HDF5_LIB_TARGET} ALL
-    DEPENDS ${lib_prog_deps} ${HDF5_GENERATED_SOURCE_DIR}/gen_SRCS.stamp1 ${HDF5_SRC_BINARY_DIR}/gen_SRCS.stamp2
+    DEPENDS ${lib_prog_deps}
```
byrnHDF: lib_prog_deps can be deleted because it is now an empty set. Which means this add_custom_target can be deleted.

src/CMakeLists.txt
```diff
@@ -1264,9 +1119,9 @@ if (BUILD_STATIC_LIBS)
 endif ()
 
 if (BUILD_SHARED_LIBS)
-  set (shared_gen_SRCS ${HDF5_GENERATED_SOURCE_DIR}/shared/H5Tinit.c ${HDF5_SRC_BINARY_DIR}/shared/H5lib_settings.c)
+  set (shared_gen_SRCS ${HDF5_SRC_BINARY_DIR}/H5build_settings.c)
   add_custom_target (gen_${HDF5_LIBSH_TARGET} ALL
```
byrnHDF: add_custom_target can be deleted.

src/CMakeLists.txt
```diff
@@ -1264,9 +1119,9 @@ if (BUILD_STATIC_LIBS)
 endif ()
 
 if (BUILD_SHARED_LIBS)
-  set (shared_gen_SRCS ${HDF5_GENERATED_SOURCE_DIR}/shared/H5Tinit.c ${HDF5_SRC_BINARY_DIR}/shared/H5lib_settings.c)
+  set (shared_gen_SRCS ${HDF5_SRC_BINARY_DIR}/H5build_settings.c)
```
byrnHDF: This shared_gen_SRCS variable could also be deleted and just use ${HDF5_SRC_BINARY_DIR}/H5build_settings.c directly.

src/CMakeLists.txt
```diff
@@ -1078,155 +1075,13 @@ if (LOCAL_BATCH_TEST)
...
-  )
-endif ()
-
 #-----------------------------------------------------------------------------
-# Add H5Tinit source to build - generated by H5detect/CMake at configure time
+# Set up library builds
 #-----------------------------------------------------------------------------
 if (BUILD_STATIC_LIBS)
-  set (gen_SRCS ${HDF5_GENERATED_SOURCE_DIR}/H5Tinit.c ${HDF5_SRC_BINARY_DIR}/H5lib_settings.c)
+  set (gen_SRCS ${HDF5_SRC_BINARY_DIR}/H5build_settings.c)
```
byrnHDF: This shared_gen_SRCS variable could also be deleted and just use ${HDF5_SRC_BINARY_DIR}/H5build_settings.c directly.

byrnHDF (changes requested):
A few more removals and then it will be good.

byrnHDF (approved):

qkoziol:
These are two very different changes and should really be in separate PRs.

release_docs/RELEASE.txt
```diff
@@ -47,6 +47,24 @@ New Features
...
+
+      These were used to work around H5detect and H5make_libsettings and
+      are no longer required.
+
+    - Running H5make_libsettings is no longer required for cross-compiling
+
+      The functionality of H5make_libsettings is now handled via template files,
+      so H5make_libsettings has been removed.
+
+    - Running H5detect is no longer required for cross-compiling
```
qkoziol, line 63: There should be a specific callout for dropping VAX support.

fortnern:
Have we tested how much overhead this adds to library startup? Do we know of any users that would be particularly sensitive to startup time? MATLAB doesn't load HDF5 every time it eeds to read/write a MAT file or anything like that does it?

src/H5Tinit_float.c
```diff
@@ -0,0 +1,552 @@
...
+                _last         = _i;                                                                          \
+            }                                                                                                \
+        }                                                                                                    \
+        H5T__fix_order(sizeof(TYPE), _last, INFO.perm, &INFO.order);                                         \
+                                                                                                             \
+        /* Implicit mantissa bit */                                                                          \
+        _v1       = (TYPE)0.5L;                                                                              \
+        _v2       = (TYPE)1.0L;                                                                              \
+        INFO.imp  = H5T__imp_bit(sizeof(TYPE), INFO.perm, &_v1, &_v2, _pad_mask);                            \
+        INFO.norm = INFO.imp ? H5T_NORM_IMPLIED : H5T_NORM_NONE;                                             \
```
qkoziol, line 115: I don't see this line ("INFO.norm = ...") in the H5detect.c source - where did it come from?

src/H5Tinit_float.c
```diff
@@ -0,0 +1,552 @@
...
+        if ((aa = (unsigned char)(a[perm[i]] & pad_mask[perm[i]])) !=
+            (bb = (unsigned char)(b[perm[i]] & pad_mask[perm[i]]))) {
+            unsigned int j;
+
+            for (j = 0; j < 8; j++, aa >>= 1, bb >>= 1) {
+                if ((aa & 1) != (bb & 1))
+                    HGOTO_DONE(i * 8 + j);
+            }
+        }
+    }
```
qkoziol, line 273: There should be some new error detection in this routine, to bring the previous error checking in.

release_docs/RELEASE.txt
derobins: We never had "VAX support". That was just what we called "not LE, not BE". I left all that processing the same since ARM can also run in mixed endian mode.

qkoziol: Yeah, I was just considering deleting this comment, as I get further into the code...

derobins:
> Have we tested how much overhead this adds to library startup? Do we know of any users that would be particularly sensitive to startup time? MATLAB doesn't load HDF5 every time it eeds to read/write a MAT file or anything like that does it?

I have a hard time believing that doing a small amount of FP munging to guess parameters would have a significant effect on the startup time of the library. Definitely not enough to where I'm willing to sacrifice cross compiling on that altar.

src/H5Tinit_float.c
derobins: I added it because it was easier to set the norm here than do the string munging in H5detect.

derobins:
> These are two very different changes and should really be in separate PRs.

I originally did that, but the H5detect PR became a blocker on removing H5make_libsettings. Having two concurrent PRs in that touch the same files would have generated a TON of merge conflicts.

src/H5Tinit_float.c
derobins: I can put the asserts back in, but they didn't seem particularly helpful.

src/H5Tinit_float.c
```diff
@@ -0,0 +1,552 @@
...
+        for (i = 0; i < n; i++)
+            perm[i] = (n - 1) - i;
+    }
+    else {
+        /* Undetermined endianness - defaults to 'VAX' for historical
+         * reasons, but there are other mixed-endian systems (like ARM
+         * in rare cases)
+         */
+        assert(0 == n % 2);
+        *order = H5T_ORDER_VAX;
```
fortnern, line 327: Quincey's other comment implies this change drops VAX support, but this code appears to still support VAX (byte order at least). Does this code still do anything useful?

src/H5Tinit_float.c
qkoziol: Found it, resolving this comment.

src/H5Tinit_float.c
qkoziol: I think it would be better to return an error value and detect that in the calling code.

src/H5Tinit_float.c
qkoziol: Yes, no change in functionality, I was incorrect about my previous comment.

qkoziol:
> > Have we tested how much overhead this adds to library startup? Do we know of any users that would be particularly sensitive to startup time? MATLAB doesn't load HDF5 every time it eeds to read/write a MAT file or anything like that does it?
> 
> I have a hard time believing that doing a small amount of FP munging to guess parameters would have a significant effect on the startup time of the library. Definitely not enough to where I'm willing to sacrifice cross compiling on that altar.

Agree

qkoziol (changes requested):
I'm on board with everything, except I'd like to see the error checking that I mentioned.

qkoziol:
I'm looking forward to getting this change in, will be very good for long-term code maintenance!  👍

derobins:
> I'm on board with everything, except I'd like to see the error checking that I mentioned.

I'll add some error checking back in.

src/H5Tinit_float.c
derobins: Yes. I want to keep this in, even though VAX is defunct (we can change the H5T_ORDER_VAX name later). ARM allows mixed-endian floats and I would not be surprised to see other platforms do that as well in our heterogeneous future.

mkitti:
I discussed this with @derobins earlier. The one thing I wanted to offer is that some of what is being detected can be determined at compile time via a compile time error through static assertions. See https://github.com/HDFGroup/hdf5/issues/1203#issuecomment-1548606918 . 

In other words, you might not need to perform some of these runtime checks since the compiler itself already knows some of these values. This static assertions can be checked via [Godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGE6a4AMngMmAByPgBGmMQgZgDspAAOqAqETgwe3r7%2ByanpAsGhESzRsQm2mPaOAkIETMQEWT5%2BUpXVGXUNBEXhUTFxiQr1jc05bcPdvSVlgwCUtqhexMjsHOYAzCHI3lgmG25MCkqNAHQI%2B9gmGgCCm9u7mPtuVLSoTATnl9c3IQQA1CwmCEIAA3VB4dBzf4meJWW7/RH/SaOZAAfSOJwIEDSAC9MKgqNQ3h9oTCNgARfYUpG0un0%2BnSGFmMzmMxzfbwm60lF4dGYmLYgBigQAKmiAEo3CkASQAGvTqdSGSrVWZSMzWSyORsuTz6qiMcdBRAReKALI3MLi2UAcSRSspqudtLMTLZbJ1eqRvP5xsaprFaLtKsdNJdEYAbBqPdrOT99R8%2BUasYGLTKwmjsHLlLSwxHVQBaLhmACsMZZnvjCJ9BuTAoDZrR5ozaK4GizOfJVKdBZVhY2iU1Vd1CdrSb9qablrlndzDspyr7DJLAA4K1r2dXuePDQ3hUGZ22O9nc/nl4rLBt18O46Oa4jfSmTdOM6HF72VVxTlx4qXJAAnJIGyltgA6rkKG4jt6j51pOL6HjcCoMueKobKckgaGYq5mBskjxFcV4QVBd4wcicHPo2QbYMoQgyoEADyYR5h%2B4Zfj%2BAFmBonEARoYEaPEkG3lu953A%2B5ETpR2J4gSRLoEskT0NCqEXnSN6xiJZFPvuEAUhYgTNlaNoyvaiIqapiKlhsJGaWOsGSTpekGSGKGsRZdJcOWwlenZEl7v62JOc2ranixPZsapxZYVwNk%2BeJ2kBbp%2BnBZm7Zzt2S7uQOAmxduib%2BamQVHqFC7he5iLtm6uWifl9aJUVSHHul5mqRsGjqZWpG%2BQlhXJTO75lf837xAB8SRgBGxcHhq6RrhnkEdYbUdZucU7vZBUmkVb6uYNZinCypYCdeparod8RcVIYFLdVWkUY5yU0XRjHMaVS57SyGiSJIkafZxh2TZN/FcNG3nbr5xCYAQywMP8Ghg/EVK3BwCy0Jwpa8H4HBaKQqCcG41jWMiSwrJgzKTbwBCaMjCwANYgId%2BicJIGNUzjnC8AoIAaKQlNY8jpBwLASBoCwSR0DE5CUCLYv0LEfLIFxZhTdzWCgnymAAGp4JgADuDFJIwnA8HwdAEDEnMQJErORCEDQAJ5G7wNvMMQdsMZE2iYA4jukCLbCCAxDC0A7fOkFgkReMAhy0LQnPcLwWBAkY4ih/gEMOHgoKYHH2OYKoXteGbPt/FUrO0HgkTEPbHhYKzBDEHgLCOwsrxMMACha7r%2BuG/HMiCCIYjsG0/CCIoKjqKHugxQYRgoATlj6BXnOQAsqBJDUDBx7wqBZ8QDdYMvEALHYXsby4DDuJ4LR6EEIR9KUAwxSkaQb2MrR5C/GTTP0sQxSfGe1BGE0K%2B4x2in06EA7%2BD9f62CAW/PQkxGhQNmFwY%2BxNVgSBRmjFmodcYcH%2BKoGahZIySH%2BPLf4itvzoQ0P8CAuBCAkDJqgimVM5i03ptzVGHBmakExtjPBHMuY81YVgjgZgcH8PZsIvmbDSC7zSM4SQQA) across many compilers.


According to [section"5.2.4.2.2 Characteristics of floating types <float.h>" of the C99 standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)

> All integer values in the <float.h> header, except FLT_ROUNDS, shall be constant
expressions suitable for use in #if preprocessing directives; all floating values shall be
constant expressions. All except DECIMAL_DIG, FLT_EVAL_METHOD, FLT_RADIX,
and FLT_ROUNDS have separate names for all three floating-point types. The floatingpoint model representation is provided for all values except FLT_EVAL_METHOD and
FLT_ROUNDS.

The GCC documentation for the available values is below:
https://www.gnu.org/software/libc/manual/html_node/IEEE-Floating-Point.html

> A.5.3.3 IEEE Floating Point
> Here is an example showing how the floating type measurements come out for the most common floating point representation, specified by the IEEE Standard for Binary Floating Point Arithmetic (ANSI/IEEE Std 754-1985). Nearly all computers designed since the 1980s use this format.
>
> The IEEE single-precision float representation uses a base of 2. There is a sign bit, a mantissa with 23 bits plus one hidden bit (so the total precision is 24 base-2 digits), and an 8-bit exponent that can represent values in the range -125 to 128, inclusive.
>
> So, for an implementation that uses this representation for the float data type, appropriate values for the corresponding parameters are:
```
FLT_RADIX                             2
FLT_MANT_DIG                         24
FLT_DIG                               6
FLT_MIN_EXP                        -125
FLT_MIN_10_EXP                      -37
FLT_MAX_EXP                         128
FLT_MAX_10_EXP                      +38
FLT_MIN                 1.17549435E-38F
FLT_MAX                 3.40282347E+38F
FLT_EPSILON             1.19209290E-07F
```
> Here are the values for the double data type:
```
DBL_MANT_DIG                         53
DBL_DIG                              15
DBL_MIN_EXP                       -1021
DBL_MIN_10_EXP                     -307
DBL_MAX_EXP                        1024
DBL_MAX_10_EXP                      308
DBL_MAX         1.7976931348623157E+308
DBL_MIN         2.2250738585072014E-308
DBL_EPSILON     2.2204460492503131E-016
```

GCC's float.h is here:
https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/ginclude/float.h;h=45021e48b8f1cb6a641d00479054a7474d1a1c3d;hb=HEAD

LLVM's float.h is here:
https://github.com/llvm/llvm-project/blob/375d798067806de0d5323bae5341f5bad4b8c107/clang/lib/Headers/float.h#L2

This does not check endianness, so perhaps some of the runtime checks are necessary. Otherwise, many of them may not be necessary.

mkitti:
`long double` is not defined in IEEE 754 and differs with [MSVC as per Godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGE6a4AMngMmAByPgBGmMQSAGykAA6oCoRODB7evv5JKWkCwaERLNGxXAl2mA7pQgRMxASZPn5Stpj2jgK19QSF4VEx8bZ1DU3ZrQojvSH9JYPlAJS2qF7EyOwc5gDMIcjeWCZbbkwKSg0AdAiH2CYaAILbu/uYh25UtKhMBJfXtw9mOwYey8ByO70%2B3yuWxu9xCBAA1CwmCEIAA3VB4dALeEmADsVnu8KJ8MmXzwyAA%2BiczgQIKkAF6YVBUCAfIzw9ArSL0bGHAAi/OJQuFItF5VIOLMZnMZgWhwJdyFpMclOpMVpgT5FkCFIAsncwgAVCl8gCSAHFifzBaLbXahXFpJLpVK5VsFUq6iqqad1RBNdqTRbbdatnz7RHbVwABwSmUyt0e4nK8k%2Bmn%2BrU63WmsIU7AADWUQtD4cjkYAtOUttGuHGpQn5X9PWTVb6GhnA9nc1wNHnCziwzayxHy5IAJxbWvOhvupvJr2ptXtgNZu75vtFq2DsPDyNV6NO%2BOuxuE%2BcttN%2Bld6tcUnsbgcCne7iPjrZmOsu2UnxVn71LjWZte%2BYhtupZcOcMZjrik7jgArFssG4gCXC4nEsEaGYNyWK%2BZiBB%2BM5JkSKatumV5diBj6lls5xbHEZg9jBk5cGYGgTuhdFxNgo4Trh%2BHHrOp5EQuJGXoB2DKEIpqBAA8mExagUK4EaAeLH0UhkjRrB6GSJIWySBoXBcVwY54dO/EKnORLEJgBCrAw8IaN%2BeICvcHBLLQnCwbwfgcFopCoJwbjWNYJIrGsmCSpOvAEJoblLAA1iAsFxOcaExlpSFbGOE7QfonCSN5sX%2BZwvAKCAGikDFvluaQcCwEgaAsIkdAxOQlCNc19CxKiyCJIkFKosZFIGAQmCTBSqiOqQWCouSmAAGp4JgADu0mJIwnA8HwdCjcQZUQJERWRCE9QAJ6bbwx3MMQp3SZE2hVFVW2NWwgjSQwtDndV02YJEXjAMctC0GV3C8FgSJGOI334NZ1SomNRWYKoVReKNF3kII7RFbQeCRMQZ0eFgRUEMQeAsBdSzvEwwAKItK1rRtoMyIIIhiOwrT8IIigqOo326LWBhGCgwWWPouNlZASyoIknQMCDvCoPDxCk1gEsQEslTVM4ECuGMLSkEEMzFKUejJKkst66beSy30xvzG0HQ1FMlu1prsvdA0tsDGUww9C7vue0b3sSBrYXrCHeUcF5pA%2BX5AUcPCk2SIiCg9fCg1jucI1jQiEC4IQJCRVwCzRbFCxLAgmBMFgsTq6QiW6TRGlSJIZioRoGh0bikcFTHRXx6V5WVWX7mcGYhXfQPw/VeXpBK6kziSEAA%3D%3D) resulting in a compile time failure from the compiler.

The values for `long double` under MSVC can be checked with this [alternate set of static asserts](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGEgKykrgAyeAyYAHI%2BAEaYxCDSAA6oCoRODB7evgGkyamOAqHhUSyx8dJ2mA7pQgRMxASZPn5cgZXVArX1BEWRMXEJtnUNTdmtQ929JWUJAJS2qF7EyOwc5gDMYcjeWCbrbkwKSg0AdAh72CYaAIIbWzuYe25UtKhMBGcXV7c3YQQA1CwmGEIAA3VB4dCzf4mADsVhu/yR/wUdUcyAA%2BodjgQIKkAF6YVBUCCvIz/dCLaL0aF7AAi9ORTOZLKRAA5SDCzGZzGZZnsEdcmaj3nhMdi4rjgnSLMEMQBZa4RAAqGLpAEkAOLI%2BmM1n61n%2Bdac3m8/nrQXCtFirFHSUQaWytVa/W69Z0g2epnjLk87nmy3IkXo204h0yuXy9URDHYAAayiZbo9Xv1AFouBozFwTdyzQLvlbReK7Q1w06ozHM7GEzD3XrU6y0%2BsNLDc36%2BQXEUHrSWw47I9c4zXEzr6%2B7G/rM2ZpL78xbCz3i6H7QOFUOMdX44nk5P9S2OXP/V2hUuQxKy2vFXHXeOU1wTrCAJywgBsT/WXHWkjZr7Mn/8WFLksA923nQMkWDG0LylCMFWjW8GQnf4zBObl/FbdY2X8bDWzMDQpGwZsNEPU1jwXbtIN7FdLzg7BlCEdVggAeQiJM7yRVDuQ0SRJFfHinzMDDP0/IiCNfMDyMFRckWITACCWBh/g0E84TpDh5loTh/F4PwOC0UhUE4NxrGsFFFmWTAuU/XgCE0DT5gAaxAdZYUfV9nw0LDWifJ8cLMfROEkXT7MMzheAUEANFIOz9I00g4FgJA0BYRI6DichKBStL6HiUFkESRIMVBLgnwxdZ/AxVRX2kLBQTFTAADU8EwAB3ZjEkYTgeD4OgCDiSKIGiULojCeoAE9ut4UbmGIcbmOibQqlinqUrYQRmIYWhJri0gsGiLxgAOWhaEi7heCwIEjHEXb8Dk6pQUwM6DMwVQqi8fqpvIQRMC03baDwaJiAmjwsFCghiDwFgpvmF4mGABRmrajquvOmRBBEMR2CkdH5CUNRQt0HMDCMFBTMsfRAciyB5lQRICgYM7eFQR7iEhrBqYgeZ2gZlwGHcTxmj0EIwj6UoBhzPI0gEUYWlyFJpYYKZ%2BniHMeZqYZGkFsZbF%2B5aNcmUXpgliYRm1uWRQaZXxdV7mLJWCRNO0kLdqMjh/mqyRAQUfL/hKp8Tgq/4IFwQgSGsrhZls%2BzZnmBBMCYLB4i50hnLMNkTjZJ82TfdOMI8kj/ACv7gtIPSDLdiKopimOEpgRAQEWAhEg%2BzKIGy9LiAiVhVk973ff9tDX14TB8CIdm9H4DHRHEHGp7xlR1F2onSFa4HEhhwKOB0svQrd5iPpbgFiQ9mr%2B%2BQP3SsD/xg48VLO4jqOa7i2PU5c9ZA/WL/v5/7%2Bt9L8uzNwq2GrrFLQr8/pmBdhXYBYCHKkFZqkZwkggA%3D).

Is there a situation where the static assertions are true but the runtime checks disagree? Is it still worth checking these values if the values are contained in float.h?

derobins:
@mkitti - We should move the conversation about improvements to the Discussions section so it doesn't get lost. I'm looking to move this into the library ASAP so we can get to making cross-compiling work better (particularly for MinGW w/ GitHub actions), so any future changes would go in a separate PR.

derobins:
Also, before we retire any floating-point checks, we should think about how we're going to handle float16 values. There may be other floating-point schemes in the future that we'd need to handle.

https://en.wikipedia.org/wiki/Half-precision_floating-point_format

mkitti:
This pull request is a clear improvement, so please continue.

src/H5Tinit_float.c
```diff
@@ -0,0 +1,573 @@
...
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5T__bit_cmp(unsigned nbytes, int *perm, void *_a, void *_b, const unsigned char *pad_mask, unsigned *first)
+{
+    unsigned char *a = (unsigned char *)_a;
+    unsigned char *b = (unsigned char *)_b;
+    unsigned char  aa, bb;
+    herr_t         ret_value = SUCCEED;
+    ;
```
qkoziol: Nit: Bonus semicolon

qkoziol (approved):
Nice!

src/H5Tinit_float.c
derobins: Fixed

PR #3107, opened 2023-06-12
Fix the default for API_VERSION to only allow one default

byrnHDF:


hyoklee:
Do we have a GitHub Action to confirm the fix?

If not, please test your fix with this Action:

https://github.com/hyoklee/hdf5/blob/develop/.github/workflows/mac-cmake-offs.yml

lrknox (approved):

derobins (approved):

PR #3112, opened 2023-06-13
Fix a misc warning in test/vol.c

derobins:
The compiler complains about using integers instead of size_t for some sizes.

### DIFF
--- a/test/vol.c
+++ b/test/vol.c
@@ -573,9 +573,9 @@ reg_opt_datatype_get(void H5_ATTR_UNUSED *obj, H5VL_datatype_get_args_t *args, h
 static herr_t
 fake_vol_info_to_str(const void *info, char **str)
 {
-    herr_t    ret_value = SUCCEED; /* Return value */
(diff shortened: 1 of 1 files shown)


test/vol.c
```diff
@@ -573,9 +573,9 @@ reg_opt_datatype_get(void H5_ATTR_UNUSED *obj, H5VL_datatype_get_args_t *args, h
 static herr_t
 fake_vol_info_to_str(const void *info, char **str)
 {
-    herr_t    ret_value = SUCCEED; /* Return value */
-    const int val       = *(const int *)info;
-    const int str_size  = 16; /* The size of the string */
+    const int    val       = *(const int *)info;
+    const size_t str_size  = 16; /* The size of the string */
```
derobins, line 577: str_size needed to be a size_t, not an int. The rest is tidying.

byrnHDF (approved):

mattjala (approved):

PR #3118, opened 2023-06-13
Merges from develop/1.14

byrnHDF:
Fix release action that allows files to be uploaded.
Add autotools h5copy help test.
Update cmake tools.cmake file for clang-tidy.
CMake build is missing H5FDsubfiling.h
macros improved for configure output capture.
Remove obsolete files


config/cmake/HDFLibMacros.cmake
```diff
@@ -17,13 +17,14 @@ macro (ORIGINAL_ZLIB_LIBRARY compress_type)
...
         URL ${ZLIB_URL}
         URL_HASH ""
     )
   endif ()
   FetchContent_GetProperties(HDF5_ZLIB)
-  if(NOT zlib_POPULATED)
+  if(NOT hdf5_zlib_POPULATED)
     FetchContent_Populate(HDF5_ZLIB)
 
     # Copy an additional/replacement files into the populated source
```
lrknox, line 30:
```suggestion
    # Copy an additional/replacement file into the populated source
```

release_docs/RELEASE.txt
```diff
@@ -409,7 +409,15 @@ Bug Fixes since HDF5-1.12.2 release
 
     Java Library
     ------------
-    -
+    - Fixed switch case 'L' block missing a break statement.
+
+      The HDF5Array.arrayify method is missing a break statement in the case 'L': section
```
lrknox, line 414: This should probably be "was missing" and "caused" to agree with "Fixed" and "was fixed".

src/H5Rpublic.h
```diff
@@ -75,7 +75,7 @@ typedef haddr_t hobj_ref_t;
 /**
  * Buffer to store heap ID and index
  *
- * This needs to be large enough to store largest #haddr_t in a worst case
+ * This needs to be large enough to store largest haddr_t in a worst case
```
lrknox, line 78: large enough to store the largest possible haddr_t?

What's a worst case machine?

lrknox (approved):

config/cmake/HDFLibMacros.cmake
byrnHDF: This will need to be fixed in develop, then 1.14 first.

release_docs/RELEASE.txt
byrnHDF: Another item for the "fix the documentation" issue to fix in all branches.

src/H5Rpublic.h
byrnHDF: This change was only to fix doxygen warning.
Another item for the "fix the documentation" issue to fix in all branches.

PR #3119, opened 2023-06-13
Adds an optional version arg to bin/format_source

derobins:


derobins:
Our version of clang-format tends to lag the current one and it'd be nice to have an easy way to run clang-format-x on systems where the default clang version is different.

byrnHDF (approved):

byrnHDF (approved):

PR #3120, opened 2023-06-14
Remove unused HD macros

derobins:
The library prefixes most C and POSIX API calls with 'HD'. We are going to start removing these so the code looks like normal C.

This PR removes most of the unused HD markup macros.

### DIFF
--- a/config/cmake/ConfigureChecks.cmake
+++ b/config/cmake/ConfigureChecks.cmake
@@ -425,13 +425,6 @@ if (MINGW OR NOT WINDOWS)
   endif ()
   CHECK_SYMBOL_EXISTS (TIOCGWINSZ "sys/ioctl.h" ${HDF_PREFIX}_HAVE_TIOCGWINSZ)
   CHECK_SYMBOL_EXISTS (TIOCGETD   "sys/ioctl.h" ${HDF_PREFIX}_HAVE_TIOCGETD)
-
-  # ----------------------------------------------------------------------
-  # cygwin user credentials are different then on linux
-  #
-  if (NOT CYGWIN AND NOT MINGW)
-    CHECK_FUNCTION_EXISTS (getpwuid        ${HDF_PREFIX}_HAVE_GETPWUID)
-  endif ()
 endif ()
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


derobins:
Normally, I like to have 'complete' PRs that fully implement a thing. I feel like we're going to need to do the HD prefix removal in stages, though, since the changes will touch so much. Ditto converting hbool_t/TRUE/FALSE to bool/true/false.

derobins:
This doesn't remove all the unused HD macros. I tried to keep 'sets' of related functionality in case we needed to adjust to quiet warnings, etc. while the work takes place.

derobins:
The next two sets will probably target the network functionality in the mirror VFD and then I'll do the C99 API calls, which should be available everywhere and no longer need the prefixes. We can deal with the POSIX calls later as they are more complicated.

config/cmake/ConfigureChecks.cmake
```diff
@@ -451,10 +444,6 @@ CHECK_FUNCTION_EXISTS (rand_r            ${HDF_PREFIX}_HAVE_RAND_R)
 CHECK_FUNCTION_EXISTS (random            ${HDF_PREFIX}_HAVE_RANDOM)
 CHECK_FUNCTION_EXISTS (setsysinfo        ${HDF_PREFIX}_HAVE_SETSYSINFO)
 
-CHECK_FUNCTION_EXISTS (siglongjmp        ${HDF_PREFIX}_HAVE_SIGLONGJMP)
-CHECK_FUNCTION_EXISTS (sigsetjmp         ${HDF_PREFIX}_HAVE_SIGSETJMP)
-CHECK_FUNCTION_EXISTS (sigprocmask       ${HDF_PREFIX}_HAVE_SIGPROCMASK)
-
```
derobins, line 457: I think these were used for type checks that were removed some time ago

config/cmake/ConfigureChecks.cmake
```diff
@@ -425,13 +425,6 @@ if (MINGW OR NOT WINDOWS)
   endif ()
   CHECK_SYMBOL_EXISTS (TIOCGWINSZ "sys/ioctl.h" ${HDF_PREFIX}_HAVE_TIOCGWINSZ)
   CHECK_SYMBOL_EXISTS (TIOCGETD   "sys/ioctl.h" ${HDF_PREFIX}_HAVE_TIOCGETD)
-
-  # ----------------------------------------------------------------------
-  # cygwin user credentials are different then on linux
-  #
-  if (NOT CYGWIN AND NOT MINGW)
-    CHECK_FUNCTION_EXISTS (getpwuid        ${HDF_PREFIX}_HAVE_GETPWUID)
```
derobins, line 433: This was used in H5detect.c and H5make_libsettings.c and is unused now.

hyoklee (approved):

bmribler (approved):

byrnHDF (approved):

qkoziol (commented):
LGTM

PR #3121, opened 2023-06-14
Clean up mirror VFD code in utils dir

derobins:
* Remove dead code
* Replace mybzero with memset
* Replace hbool_t/TRUE/FALSE with bool/true/false


derobins:
This is the first place I'm ripping out hbool_t/TRUE/FALSE for replacement with bool/true/false

hyoklee (approved):

byrnHDF (approved):

PR #3124, opened 2023-06-14
Many fixes to various compiler warnings

seanm:


byrnHDF (approved):

seanm:
@byrnHDF thanks for reviewing.  I have one more commit to add since CI apparently uses -Werror for -Wunused-variable and the fixes from -Wmissing-variable-declarations (making variables static) allowed the compiler to realize some global variables are unused.

src/H5Fsfile.c
```diff
@@ -35,7 +35,7 @@ typedef struct H5F_sfile_node_t {
 H5FL_DEFINE_STATIC(H5F_sfile_node_t);
 
 /* Declare a local variable to track the shared file information */
-H5F_sfile_node_t *H5F_sfile_head_g = NULL;
+static H5F_sfile_node_t *H5F_sfile_head_g = NULL;
```
qkoziol: I'd rename this to H5F_sfile_head_s, if it's static

src/H5Iint.c
```diff
@@ -92,7 +92,7 @@ int              H5I_next_type_g = (int)H5I_NTYPES;
 H5FL_DEFINE_STATIC(H5I_id_info_t);
 
 /* Whether deletes are actually marks (for mark-and-sweep) */
-hbool_t H5I_marking_g = FALSE;
+static hbool_t H5I_marking_g = FALSE;
```
qkoziol: Same here:  H5I_marking_s

src/H5Fsfile.c
seanm: Does the `_g` mean global?  Because it's still global.  Isn't its globalness more interesting than it being static?

qkoziol: The _s / _g is an indication about the symbol's visibility.  Static symbols are not visible outside of their module, "global" (i.e. non-static) symbols are.

c++/examples/h5tutr_extend.cpp
```diff
@@ -113,6 +113,7 @@ main(void)
         // Get information to obtain memory dataspace.
         rank            = filespace->getSimpleExtentNdims();
         herr_t status_n = filespace->getSimpleExtentDims(dimsr);
```
qkoziol: This seems awkward.  How about eliminating the variable entirely and just casting the return value from the call to void?

c++/examples/readdata.cpp
```diff
@@ -85,6 +85,7 @@ main(void)
              */
             H5std_string order_string;
             H5T_order_t  order = intype.getOrder(order_string);
+            (void)order;
```
qkoziol: Same for the two in this file (remove variable and cast call's return value to void)

test/cache_tagging.c
```diff
@@ -551,7 +551,7 @@ check_file_open_tags(hid_t fcpl, int type)
 #endif                   /* NDEBUG */
     hid_t   fapl = -1;   /* File access prop list */
     haddr_t root_tag;    /* Root Group Tag */
-    haddr_t sbe_tag;     /* Sblock Extension Tag */
+    haddr_t sbe_tag = 0; /* Sblock Extension Tag */
```
qkoziol: This would be better as HADDR_UNDEF

qkoziol:
Why all the 'static' variable changes in the test files?   What compiler warning flag is complaining about those?

c++/examples/h5tutr_extend.cpp
seanm: Could do. As I said in the commit message though, it's sometimes nice in the debugger to be able to inspect the variable, which I assume is why the variable is there.  But I'm happy to remove it if you prefer.

seanm:
> Why all the 'static' variable changes in the test files? What compiler warning flag is complaining about those?

`-Wmissing-variable-declarations` (as per the commit message :))

derobins:
Which compiler and platform are emitting all these warnings?

test/testframe.c
```diff
@@ -221,7 +221,7 @@ TestParseCmdLine(int argc, char *argv[])
     hbool_t skipped_all = FALSE;
     int     ret_code;
 
-    while (argv++, --argc > 0) {
+    while ((void)argv++, --argc > 0) {
```
derobins, line 224: Making a mental note to replace this with getopt...

seanm:
> Which compiler and platform are emitting all these warnings?

I'm using clang (on macOS, but the OS won't matter).  I'm enabling non-default warnings. I'm trying to get HDF5 building warning-free with the same warning set that my other 3rd party dependencies already build cleanly with.

c++/examples/h5tutr_extend.cpp
derobins: I would also prefer casting to void as a library pattern. These are also examples, so debugging is less of an issue here.

qkoziol:
> > Why all the 'static' variable changes in the test files? What compiler warning flag is complaining about those?
> 
> `-Wmissing-variable-declarations` (as per the commit message :))

OK,missed that.  (In general, a less "omnibus" PR would be better)

seanm:
> OK,missed that. (In general, a less "omnibus" PR would be better)

okie dokie. Next time will do one PR per warning flag. These big ones are a lot to review I agree...

test/cache_tagging.c
seanm: Done.

c++/examples/readdata.cpp
seanm: Done.

c++/examples/h5tutr_extend.cpp
seanm: Done.

src/H5Iint.c
seanm: Done.

src/H5Fsfile.c
seanm: Done.

seanm:
@qkoziol @derobins If you'll forgive one more omnibusing :) I added 2 more commits, fixing 2 more warning options. They are very small though!

With this PR, with Xcode 14.2's clang, these now build cleanly with the Weverything & subtractive approach, at least with the build options I've been able to activate (which is as many as I could):

For C:
`-Weverything -Wno-padded -Wno-missing-prototypes -Wno-missing-noreturn -Wno-unused-macros -Wno-format-nonliteral -Wno-conversion -Wno-sign-compare -Wno-sign-conversion -Wno-switch-enum -Wno-unreachable-code -Wno-covered-switch-default -Wno-reserved-id-macro -Wno-disabled-macro-expansion -Wno-documentation -Wno-documentation-unknown-command -Wno-documentation-html -Wno-documentation-deprecated-sync -Wno-missing-variable-declarations -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-declaration-after-statement -Wno-reserved-identifier -Wno-extra-semi-stmt -Wno-unreachable-code-break -Wno-used-but-marked-unused -Wno-double-promotion -Wno-sign-compare -Wno-conditional-uninitialized -Wno-missing-variable-declarations -Wno-tautological-value-range-compare`

For C++:
`-Weverything -Wno-padded -Wno-missing-prototypes -Wno-missing-noreturn -Wno-unused-macros -Wno-format-nonliteral -Wno-conversion -Wno-sign-compare -Wno-sign-conversion -Wno-switch-enum -Wno-unreachable-code -Wno-covered-switch-default -Wno-reserved-id-macro -Wno-disabled-macro-expansion -Wno-documentation -Wno-documentation-unknown-command -Wno-documentation-html -Wno-documentation-deprecated-sync -Wno-missing-variable-declarations -Wno-unused-parameter -Wno-unused-but-set-variable -Wno-long-long -Wno-float-equal -Wno-c++98-compat-pedantic -Wno-exit-time-destructors -Wno-global-constructors -Wno-weak-vtables -Wno-extra-semi -Wno-old-style-cast -Wno-zero-as-null-pointer-constant -Wno-deprecated-dynamic-exception-spec -Wno-deprecated-copy-with-user-provided-dtor -Wno-unused-exception-parameter -Wno-header-hygiene`

Some of these suppressed warnings are worth fixing IMHO, but that can always be done piecemeal.  If your Mac CI could use these flags that would be great.  Let me know which Xcode version it is exactly and I'll check against that too.

derobins:
When this goes in, I'll update the clang flags in config/ to use the more restrictive set here. I'll also add in -Werror build check for clang as a GitHub action to make sure nothing regresses.

hl/src/H5DS.c
```diff
@@ -152,14 +152,14 @@ H5DSattach_scale(hid_t did, hid_t dsid, unsigned int idx)
     ds_list_t *dsbuf = NULL; /* array of attribute data in the DS pointing to the dataset */
     ds_list_t *dsbuf_w =
         NULL; /* array of "REFERENCE_LIST" attribute data to write when adding new reference to a dataset */
-    hobj_ref_t ref_to_ds; /* reference to the DS */
-    hobj_ref_t ref_j;     /* iterator reference */
+    hobj_ref_t ref_to_ds = 0; /* reference to the DS */
```
qkoziol: The invalid value for hobj_ref_t is HADDR_UNDEF

test/tattr.c
```diff
@@ -4542,7 +4543,7 @@ test_attr_corder_create_compact(hid_t fcpl, hid_t fapl)
 {
     hid_t    fid;                     /* HDF5 File ID            */
     hid_t    dset1, dset2, dset3;     /* Dataset IDs            */
-    hid_t    my_dataset;              /* Current dataset ID        */
+    hid_t    my_dataset = 0;          /* Current dataset ID        */
```
qkoziol: The invalid value for hid_t types is H5I_INVALID_HID

qkoziol:
Generally speaking, most of these changes look good.  I don't especially like making all the global variables in the tests (and tools, etc) have the 'static' keyword.  That makes them look "odd" to me.

test/genall5.c
```diff
@@ -1067,7 +1067,7 @@ os_grp_0(hid_t fid, const char *group_name)
 {
     hid_t        gid  = -1;
     hid_t        fapl = -1;
-    H5F_libver_t low, high;
+    H5F_libver_t low = 0, high = 0;
```
jhendersonHDF: Rather than initializing these to 0, I feel like using `H5F_LIBVER_EARLIEST` or `H5F_LIBVER_ERROR` here would be better.

jhendersonHDF:
> Generally speaking, most of these changes look good. I don't especially like making all the global variables in the tests (and tools, etc) have the 'static' keyword. That makes them look "odd" to me.

To me, adding static to global variables that are only accessed in the file they are declared in just seems like a good practice to stick to. Though, I generally prefer those variables to have names ending in "_g" rather than "_s", but that's a small detail.

jhendersonHDF (approved):
Looks like there are just a couple minor things to be addressed for this PR, but otherwise the changes look good to me.

hl/src/H5DS.c
seanm: Done.

test/tattr.c
seanm: Done.

test/genall5.c
seanm: Done with `H5F_LIBVER_ERROR`

derobins:
> > Generally speaking, most of these changes look good. I don't especially like making all the global variables in the tests (and tools, etc) have the 'static' keyword. That makes them look "odd" to me.
> 
> To me, adding static to global variables that are only accessed in the file they are declared in just seems like a good practice to stick to. Though, I generally prefer those variables to have names ending in "_g" rather than "_s", but that's a small detail.

Now that I've had some time to think about this, I agree that local-only globals with no matching `extern` should be static. I think the _s is overly pedantic, though. We should just use _g throughout. Let's do that in a separate PR, though.

derobins (approved):

PR #3129, opened 2023-06-15
chore: add Python for HDF-EOS zoo description

hyoklee:
Python is the 2nd most popular language for [HDF-EOS Zoo](https://hdfeos.org/zoo) example page after MATLAB.


derobins:
#3130 will fix the weird codespell issues

derobins:
Same here - I'll merge this since the unrelated spelling issues are fixed in develop

derobins (approved):

PR #3133, opened 2023-06-15
Revert long double checks

derobins:


byrnHDF (approved):

config/cmake/ConfigureChecks.cmake
```diff
@@ -847,3 +860,83 @@ if (HDF5_BUILD_FORTRAN)
...
+
+  endif ()
+endmacro ()
+
+#-----------------------------------------------------------------------------
+# Check various conversion capabilities
+#-----------------------------------------------------------------------------
+
+# ----------------------------------------------------------------------
+# Set the flag to indicate that the machine is using a special algorithm toconvert
```
lrknox, line 902:
```suggestion
# Set the flag to indicate that the machine is using a special algorithm to convert
```

config/cmake/ConfigureChecks.cmake
```diff
@@ -847,3 +860,83 @@ if (HDF5_BUILD_FORTRAN)
...
+# 'long double' to '(unsigned) long' values.  (This flag should only be set for
+# the IBM Power Linux.  When the bit sequence of long double is
+# 0x4351ccf385ebc8a0bfcc2a3c3d855620, the converted value of (unsigned)long
```
lrknox, line 905:
```suggestion
# 0x4351ccf385ebc8a0bfcc2a3c3d855620, the converted value of (unsigned) long
```

configure.ac
```diff
@@ -3402,6 +3423,158 @@ case "`uname`" in
...
+## The machine's conversion gets the correct value.  We define the macro and disable
+## this kind of test until we figure out what algorithm they use.
+##
+## CROSS-COMPILING: Assume 'no'
+AC_MSG_CHECKING([if using special algorithm to convert long double to (unsigned) long values])
+
+## NOTE: Place all configure test programs into cmake's source file, then use a preprocessor directive
+## to select the proper test program. This is done by echoing the #define and cat'ing the cmake
+## source file. (HDFFV-9467)
+
```
lrknox, line 3441: Should Autotools configure make this change to the cmake/ConversionTests.c file, should it do something else, or is this just leftover from CMake changes that shouldn't be here?

lrknox (commented):
Testing on IBM Power8

configure.ac
derobins: Dunno. This is just me putting back what was there before. Let's go ahead and merge this to stop the breakage and I'll come back to the issues you've raised here in a separate PR. That'll be easier to merge downstream.

PR #3139, opened 2023-06-15
Track s3 i/o when S3COMMS_DEBUG enabled

mattjala:
@ajelenak mentioned this would be useful as a general feature.

### DIFF
--- a/src/H5FDs3comms.c
+++ b/src/H5FDs3comms.c
@@ -1281,6 +1281,10 @@ H5FD_s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest)
             HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unable to format HTTP Range value");
     }
 
+#if S3COMMS_DEBUG
+    HDfprintf(stderr, "%s: Bytes %ld - %ld, Request Size: %ld\n", handle->httpverb, offset, offset + len, len);
+#endif
+
     /*******************
      * COMPILE REQUEST *
      *******************/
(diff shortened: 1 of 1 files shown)


byrnHDF:
Don't we already have a "Developer" options - maybe even just enabling the <package>_DEBUG settings?

hyoklee (approved):

derobins:
> Don't we already have a "Developer" options - maybe even just enabling the _DEBUG settings?

The DEBUG options here turn on logging / printf dumps, not extra checks, like in the library. We should probably rename the symbol to _LOGGING or similar.

src/H5FDs3comms.c
```diff
@@ -1281,6 +1281,11 @@ H5FD_s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest)
             HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "unable to format HTTP Range value");
     }
 
+#if S3COMMS_DEBUG
+    HDfprintf(stderr, "%s: Bytes %ld - %ld, Request Size: %ld\n", handle->httpverb, offset, offset + len,
+              len);
```
derobins: This needs to use PRIuHADDR for haddr_t values and zu for size_t.

longs are 32-bit on Windows so these would be wrong there.

derobins (changes requested):
Can go in when the format string is changed

byrnHDF:
Oh, this is a pre-existing define, too bad, the name, S3COMMS_TRACE, would be a closer to what it does.

byrnHDF (approved):

derobins (approved):

PR #3140, opened 2023-06-15
Fix path to libhdf5.settings in cmakehdf5

hyoklee:
This fixes the following error message:

```cat: libhdf5.settings: No such file or directory```


byrnHDF:
bin/cmakehdf5 should be deprecated!

derobins:
> bin/cmakehdf5 should be deprecated!

Okay, but until it's gone it should work :)

derobins (approved):

qkoziol (approved):

PR #3143, opened 2023-06-15
Removed .lnt linter files

derobins:
These were last usefully modified in 2004


derobins:
There have been changes since 2004, but only of the find-n-replace kind and they are clearly unmaintained.

hyoklee (approved):

qkoziol (approved):
Yes, we have better static analysis tools now

PR #3145, opened 2023-06-16
Remove hbool_t/TRUE/FALSE from java

derobins:
Replaces with bool/true/false


derobins:
I'll remove the HD prefixes in the next PR

hyoklee (approved):

byrnHDF (approved):

PR #3147, opened 2023-06-16
Define minimal permissions for new GitHub workflows

diogoteles08:
Hi!

I'm raising this PR with the exact same purpose described at the accepted issue #2973. I have previously fixed the issue by the PR #2980, but I noticed new GitHub workflows were created and currently lack minimal permissions, so I'm taking the liberty of raising this other PR.

Please let me know if for any reason the absence of permissions is intended.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ on:
         required: true
         type: string
 
-# Minimal permissions to be inherited by any job that don't declare it's own permissions
+# Minimal permissions to be inherited by any job that doesn't declare its own permissions
 permissions:
   contents: read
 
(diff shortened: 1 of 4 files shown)


byrnHDF:
Thanks for the help!  Yes, a couple of new ones snuck by.
I couldn't seem to get the daily-build write permissions correct, so thanks for fixing that.

byrnHDF (approved):

.github/workflows/release.yml
```diff
@@ -17,6 +17,10 @@ on:
         required: true
         type: string
 
+# Minimal permissions to be inherited by any job that don't declare it's own permissions
```
lrknox:
```suggestion
# Minimal permissions to be inherited by any job that doesn't declare its own permissions
```

lrknox (approved):

derobins (approved):

PR #3149, opened 2023-06-16
Removes the HD prefix from java C99 calls

derobins:
POSIX calls (HDstrndup, etc.) are unchanged


qkoziol (approved):
LGTM, although the printf() lines have some whitespace added to the end of the line

hyoklee (approved):

byrnHDF (approved):

derobins:
> LGTM, although the printf() lines have some whitespace added to the end of the line

It looks like those are in macros, so the extra space is just lining up the line continuation markers

PR #3160, opened 2023-06-20
Test doxygen examples

mkitti:
## Describe your changes

I added some of the doxygen examples to cmake in order to test if the examples can still be compiled. The doxygen examples were moved into `examples/doxygen` from `doxygen/examples` per @byrnHDF.

- [x] I made an entry in release_docs/RELEASE.txt (bug fixes, new features)


byrnHDF:
While this isn't a bad change, we need to do the examples properly as we have discussed.
Let's table this until we get through the immediate releases.
Then I will work on a solution that pulls in the Examples project.

mkitti:
OK. Let me know if therebis anything else productive that I can do here.

PR #3183, opened 2023-06-27
Remove HD/hbool_t from high-level lib

derobins:


byrnHDF (approved):

qkoziol (approved):

qkoziol:
Some of those large src files in the library would benefit from breaking up into smaller, more themed source modules.

bmribler (approved):

PR #3191, opened 2023-06-28
Rename HDassert() to assert()

derobins:


derobins:
Apologies for the size, but since it's just one function it should be easy to scroll through

hyoklee (approved):

byrnHDF (approved):

PR #3194, opened 2023-06-28
Rename HD(f)printf() to (f)printf()

derobins:


derobins:
Again, apologies for the size, but this should also be easy to scan since it's two closely-related functions.

hyoklee (approved):

byrnHDF (approved):

fortnern (approved):
Looks like there's a orphan HDqsort? Not sure how that happened, but looks fine otherwise

PR #3208, opened 2023-06-29
docs: remove redundancy in Data Transfer section of user guide

hyoklee:
image


byrnHDF (approved):

byrnHDF:
For those future investigators, names of refs are expected to be placed at the source of the ref definition and not after "\ref" usage.

lrknox (approved):

PR #3222, opened 2023-07-03
Implementation of the mpio driver with selection I/O.

vchoi-hdfgroup:
This also addresses the review comments of the previous PR #3058 which will be closed.


src/H5FDdevelop.h
```diff
@@ -196,12 +196,12 @@ typedef struct H5FD_class_t {
                           size_t sizes[], void *bufs[]);
     herr_t (*write_vector)(H5FD_t *file, hid_t dxpl, uint32_t count, H5FD_mem_t types[], haddr_t addrs[],
                            size_t sizes[], const void *bufs[]);
-    herr_t (*read_selection)(H5FD_t *file, H5FD_mem_t type, hid_t dxpl_id, size_t count, hid_t mem_spaces[],
```
fortnern: I understand we want to keep things consistent but if we're going to change the API to keep the count parameter the same type we should change it to size_t. I think we should just lave it alone for now though.

vchoi-hdfgroup: So I will keep <count > as size_t for read_selection() and write_selection()? But <count> as uint32_t for write_vector and read_vector?

fortnern: Yeah I don't think we should change it now, it's already part of the API

src/H5FDmpio.c
```diff
@@ -2716,6 +2733,924 @@ H5FD__mpio_write_vector(H5FD_t *_file, hid_t H5_ATTR_UNUSED dxpl_id, uint32_t co
...
+#endif
+        if (_file->base_addr > 0) {
+            /* Undo base address addition in internal routines before passing down to the mpio driver */
+            for (i = 0; i < count; i++) {
+                assert(offsets[i] >= _file->base_addr);
+                offsets[i] -= _file->base_addr;
+            }
+        }
+
+        if (H5FD_read_vector_from_selection(_file, type, count, mem_space_ids, file_space_ids, offsets,
```
fortnern: I think this should actually be H5FD_read_from_selection(). Same for write (I know my earlier email said vector)

src/H5FDdevelop.h
vchoi-hdfgroup: Just want to double-check:
<count> as size_t will be used for all APIs and the related intermediate routines:
H5FDread/write_selection(), H5FDread/write_vector_from_selection(), H5FDread/write_from_selection()

vchoi-hdfgroup: correction to previous comment: "count" as size_t

fortnern: Don't change any APIs right now, thanks, just revert the type change to the H5FD_class_t struct

fortnern (approved):

src/H5FDmpio.c
vchoi-hdfgroup: Done

src/H5FD.c
```diff
@@ -1820,14 +1820,408 @@ H5FDwrite_selection(H5FD_t *file, H5FD_mem_t type, hid_t dxpl_id, uint32_t count
...
+ *              It will translate the selection read to a vector read call
+ *              if vector reads are supported, or a series of scalar read
+ *              calls otherwise.
+ *
+ *              All reads are done according to the data transfer property
+ *              list dxpl_id (which may be the constant H5P_DEFAULT).
+ *
+ * Return:      Success:    SUCCEED
+ *                          All reads have completed successfully, and
+ *                          the results havce been into the supplied
```
lrknox:
```suggestion
 *                          the results have been written into the supplied
```

src/H5FD.c
```diff
@@ -1820,14 +1820,408 @@ H5FDwrite_selection(H5FD_t *file, H5FD_mem_t type, hid_t dxpl_id, uint32_t count
...
+ *
+ *              It will translate the selection read to a series of
+ *              scalar read calls.
+ *
+ *              All reads are done according to the data transfer property
+ *              list dxpl_id (which may be the constant H5P_DEFAULT).
+ *
+ * Return:      Success:    SUCCEED
+ *                          All reads have completed successfully, and
+ *                          the results havce been into the supplied
```
lrknox:
```suggestion
 *                          the results havce been into the supplied
```

src/H5FDint.c
```diff
@@ -1999,6 +2033,217 @@ H5FD_write_selection_id(H5FD_t *file, H5FD_mem_t type, uint32_t count, hid_t mem
...
+ *               It will translate the selection write to a series of
+ *               scalar write calls.
+ *
+ * Return:      Success:    SUCCEED
+ *                          All writes have completed successfully.
+ *
+ *              Failure:    FAIL
+ *                          One or more writes failed.
+ *
+ * Programmer:  NAF -- 5/19/21
```
lrknox: Are we removing "Programmer" references?

src/H5FDint.c
```diff
@@ -2378,6 +2671,187 @@ H5FD_sort_vector_io_req(hbool_t *vector_was_sorted, uint32_t _count, H5FD_mem_t
...
+ *
+ *              If it is not sorted, duplicate the mem_space_ids, file_space_ids,
+ *              offsets, element_sizes and bufs arrays, storing the base
+ *              addresses of the new arrays in *s_mem_space_ids_ptr,
+ *              s_file_space_ids_ptr, s_offsets_ptr, *s_element_sizes_ptr,
+ *              and s_bufs_ptr respectively.  Determine the sorted order
+ *              of the selection I/O request, and load it into the new
+ *              selections in sorted order.
+ *
+ *              Note that in this case, it is the callers responsibility
```
lrknox:
```suggestion
 *              Note that in this case, it is the caller's responsibility
```

lrknox (approved):

Issue #1320, opened 2021-12-20
Bug: HDF5 1.10.8/1.13.0 installs CMake packages in wrong location


ax3l:
At least for Unix, HDF5 1.10.8/1.13.0 installs the CMake packages in the wrong location.

Please see @jbigot report in https://github.com/spack/spack/issues/28074 for details.

Note that Spack installs HDF5 via its CMake installation scripts: https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/hdf5

ax3l:
Quoting @jbigot's Spack issue:

```console
spack install hdf5@1.10.8
ls ${SPACK_ROOT}/opt/spack/*/*/hdf5-*/share/cmake
```
output should be:
```
hdf5
```
as it is for 1.12.1 for example, but instead for 1.10.8, it is:
```
hdf5-config.cmake  hdf5-config-version.cmake  hdf5-targets.cmake  hdf5-targets-relwithdebinfo.cmake
```

As specified in cmake doc https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure
`<prefix>/share/cmake/` is not part of the Search Procedure, `<prefix>/share/cmake/<name>` is.

ax3l:
Tracked the PRs that caused this:
- `develop` & `1.13.0` release: #1041
- `1.10.8` release: #1043

cc @byrnHDF :)

ax3l:
This probably also affects the 1.13.0 release

byrnHDF:
<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
is part of the search path for *nix.

jbigot:
Dear @byrnHDF ,

in [CMake doc](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure), the 3 lines I see for Unix are:
```
<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
<prefix>/(lib/<arch>|lib*|share)/<name>*/                       (U)
<prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         (U)
```
plus 3 common lines for Win/Unix
```
<prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/         (W/U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/               (W/U)
[...]

PR #3223, opened 2023-07-04
Add /hdf5 subdir to HDF5_INSTALL_CMAKE_DIR

jschueller:
Usually cmake files are located in a subdirectory of PREFIX/lib/cmake/ named after the package, not directly at its root


hyoklee (approved):

byrnHDF:
This has been discussed and vetted many times.
From HISTORY-1_12_0-1_14_0.txt(line 1842):
    - Reworked corrected path searched by CMake find_package command

      The install path for cmake find_package files had been changed to use
        "share/cmake"
      for all platforms. However setting the HDF5_ROOT variable failed to locate
      the configuration files. The build variable HDF5_INSTALL_CMAKE_DIR is now
      set to the <INSTALL_DIR>/cmake folder. The location of the configuration 
      files can still be specified by the "HDF5_DIR" variable.

      (ADB - 2022/02/02)

(line 3518):
    - Corrected path searched by CMake find_package command

      The install path for cmake find_package files had been changed to use
        "share/cmake"
      for all platforms. However the trailing "hdf5" directory was not removed.
      This "hdf5" additional directory has been removed.

      (ADB - 2021/09/27)

See also:
#1041 fixed by #1414 in response to #1320 

From the last comment in #1320;   
The W/U paths all will be searched, just that they are all dirs (the * meaning is on the dir and not files as I thought)
the simple path /cmake will work.

bmribler (approved):

derobins:
As noted above, we've actually moved away from this since it caused problems. If you can point us to official recommendations to the contrary, please reopen with a link. Thanks!

PR #3224, opened 2023-07-05
Fixed check for a VOL's async compatibility  (#3164)

lrknox:
* cap flag fix in test

* added async comp. output


lrknox:
Fixes error in hdf5_1_14 daily test running "make check-passthrough-vol"  that was fixed for hdf5 develop on 6/22/2023.

byrnHDF (approved):

PR #3241, opened 2023-07-13
Subfiling VFD source cleanup

jhendersonHDF:
Fixes an issue with building the Subfiling VFD on macOS and cleans up unnecessary Mercury functionality in the utility files included for the VFD


qkoziol (approved):

qkoziol:
Why are we keeping part of the Mercury source code in the HDF5 distribution?   Seems fine for development of the subfiling VFD, but now should be moved to a git submodule or replaced with internal HDF5 code.

jhendersonHDF:
> Why are we keeping part of the Mercury source code in the HDF5 distribution? Seems fine for development of the subfiling VFD, but now should be moved to a git submodule or replaced with internal HDF5 code.

At some point during development of the VFD the decision was made to reuse Mercury's thread pool implementation rather than writing one or finding one from elsewhere. When I took over working on the Subfiling VFD, I just stripped away all the Mercury functionality that we didn't need (including more work on that here), which leaves us with just a couple header and source files that are mostly just wrappers around pthread calls and the Berkeley header-only queue implementation.

Moving Mercury to be a git submodule could be fine, though I'd rather not deal with building all of Mercury just to use a couple files out of it. Moving this functionality to something under H5TS is probably more appropriate, but in the interest of time this hasn't been done yet. We could refactor the code by moving it under H5TS and refactoring, but that would just be a minor cleanup thing, so I've left the code next to where it's directly used for now.

byrnHDF (approved):

PR #3244, opened 2023-07-13
install h5fuse.sh into the  bin directory for autotools

brtnfld:
CMake already does this.

### DIFF
--- a/utils/Makefile.am
+++ b/utils/Makefile.am
@@ -41,4 +41,6 @@ endif
 # All subdirectories
 SUBDIRS=$(MIRROR_VFD_DIR) $(TESTUTIL_DIR) $(TOOLS_DIR)
 
+bin_SCRIPTS = subfiling_vfd/h5fuse.sh
+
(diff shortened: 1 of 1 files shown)


utils/Makefile.am
```diff
@@ -41,4 +41,6 @@ endif
 # All subdirectories
 SUBDIRS=$(MIRROR_VFD_DIR) $(TESTUTIL_DIR) $(TOOLS_DIR)
 
+bin_SCRIPTS = subfiling_vfd/h5fuse.sh
```
jhendersonHDF, line 44: It looks like scripts dont get distributed by default, so this might need to be `dist_bin_SCRIPTS`:

https://www.gnu.org/software/automake/manual/html_node/Scripts.html

jhendersonHDF: Actually, since h5fuse is setup with AC_CONFIG_FILES, `bin_SCRIPTS` should work here

jhendersonHDF (approved):

fortnern (approved):

PR #3245, opened 2023-07-13
Switch parallel compression to use vector I/O

jhendersonHDF:
Updates parallel compression feature to use vector I/O instead of creating and passing down MPI derived types to VFD

### DIFF
--- a/src/H5Dmpio.c
+++ b/src/H5Dmpio.c
@@ -213,7 +213,7 @@ typedef struct H5D_chunk_index_info_t {
  * hh - A handle for hash tables provided by the uthash.h header
  *
  */
-typedef struct H5D_filtered_collective_io_info_t {
+typedef struct H5D_filtered_collective_chunk_info_t {
     H5D_chunk_index_info_t index_info;
 
     H5D_piece_info_t *chunk_info;
@@ -229,14 +229,14 @@ typedef struct H5D_filtered_collective_io_info_t {
     void             *buf;
 
     UT_hash_handle hh;
-} H5D_filtered_collective_io_info_t;
+} H5D_filtered_collective_chunk_info_t;
 
 /*
- * Top-level structure that contains an array of H5D_filtered_collective_chunk_list_t
+ * Top-level structure that contains an array of H5D_filtered_collective_chunk_info_t
  * chunk info structures for collective filtered I/O, as well as other useful information.
  * The struct's fields are as follows:
  *
- * chunk_infos - An array of H5D_filtered_collective_io_info_t structures that contain
+ * chunk_infos - An array of H5D_filtered_collective_chunk_info_t structures that contain
  *               information about a single chunk when performing collective filtered I/O.
  *
  * num_chunks_infos - The number of entries in the `chunk_infos` array.
@@ -246,7 +246,7 @@ typedef struct H5D_filtered_collective_io_info_t {
  *                      occur during dataset reads or during dataset writes when a chunk
  *                      has to go through a read - modify - write cycle). The value for
  *                      this field is based on a chunk's `need_read` field in its particular
- *                      H5D_filtered_collective_io_info_t structure, but may be adjusted
+ *                      H5D_filtered_collective_chunk_info_t structure, but may be adjusted
  *                      later depending on file space allocation timing and other factors.
  *
  *                      This field can be helpful to avoid needing to scan through the list
@@ -254,12 +254,12 @@ typedef struct H5D_filtered_collective_io_info_t {
  *                      allocate during read operations, as an example.
  *
  */
-typedef struct H5D_filtered_collective_chunk_list_t {
-    H5D_filtered_collective_io_info_t *chunk_infos;
-    H5D_filtered_collective_io_info_t *chunk_hash_table;
-    size_t                             num_chunk_infos;
-    size_t                             num_chunks_to_read;
-} H5D_filtered_collective_chunk_list_t;
+typedef struct H5D_filtered_collective_io_info_t {
+    H5D_filtered_collective_chunk_info_t *chunk_infos;
+    H5D_filtered_collective_chunk_info_t *chunk_hash_table;
+    size_t                                num_chunk_infos;
+    size_t                                num_chunks_to_read;
+} H5D_filtered_collective_io_info_t;
 
 /*
  * Information necessary for redistributing shared chunks during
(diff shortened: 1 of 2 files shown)


qkoziol:
I'll take a look at this later today

src/H5Dmpio.c
```diff
@@ -5601,256 +5350,128 @@ H5D__mpio_get_chunk_insert_info_types(MPI_Datatype *contig_type, hbool_t *contig
...
-
-        /*
-         * Determine number of chunks for I/O operation and
-         * setup for derived datatype creation if I/O operation
-         * includes multiple chunks
-         */
-        if (num_entries == 1) {
-            /* Set last valid index to 0 for contiguous datatype creation */
-            last_valid_idx = 0;
+        if (num_entries > UINT32_MAX)
```
qkoziol: This seems rare, but certainly could be addressed by performing vector I/O on UINT32_MAX entries at a time, until all the entries were processed.

src/H5Dmpio.c
```diff
@@ -5601,256 +5350,128 @@ H5D__mpio_get_chunk_insert_info_types(MPI_Datatype *contig_type, hbool_t *contig
...
-            else
-                chunk_count = chunk_list[0].need_read ? 1 : 0;
-        }
+        if (op_type == H5D_IO_OP_WRITE)
+            iovec_count = (uint32_t)num_entries;
         else {
-            MPI_Aint chunk_buf;
-            MPI_Aint base_buf;
-            haddr_t  base_offset = HADDR_UNDEF;
+            /* For read operations, determine number of chunks actually
```
qkoziol: Couldn't this loop be avoided by tracking the # of chunks being read while the list was being created?  (i.e. when the 'need_read' flag is set on each chunk)

src/H5Dmpio.c
```diff
@@ -5601,256 +5350,128 @@ H5D__mpio_get_chunk_insert_info_types(MPI_Datatype *contig_type, hbool_t *contig
...
-         * Create derived datatypes for the chunk list if this
-         * rank has any chunks to work on
+         * Since all I/O will be raw data, we can save on memory a bit by
+         * making use of H5FD_MEM_NOLIST to signal that all the memory types
+         * are the same across the I/O vectors
          */
-        if (chunk_count > 0) {
-            if (chunk_count == 1) {
-                int chunk_len;
+        if (NULL == (io_types = H5MM_malloc(2 * sizeof(*io_types))))
```
qkoziol: Why allocate this dynamically?   An array of size 2 is no problem on the stack.

qkoziol (approved):
I don't have any strong objection to the code the way it currently is, but I have suggested some small improvements that would be good to add.  If there's a strong reason not to update the code now, they can be deferred.

src/H5Dmpio.c
jhendersonHDF: Yes, this seemed like it would be rare enough that it wasn't worth putting the logic and testing in place at the current time. It would be nice if `H5FDwrite_vector`/`H5FDread_vector` had accepted a `uint64_t` for the count instead so this would be even less of an issue, but that's enshrined in the public API for now. If you think it's worth adding the logic in I can spend some time on it.

The implementation would likely be trivial, but testing I/O on > `UINT32_MAX` chunks may be a bit problematic.

src/H5Dmpio.c
jhendersonHDF: Sure, this should be fairly easy to implement. Though, there are a few places where this field may get updated after the list is constructed that would need to be accounted for, so this was the simplest way to deal with that.

src/H5Dmpio.c
jhendersonHDF: Just a leftover from dynamically allocating all the arrays to be `iovec_count` in size and then later remembering I don't need to for the `io_types` array. Will update.

src/H5Dmpio.c
qkoziol: Agree - this is the lowest priority of the suggestions I made.

src/H5Dmpio.c
jhendersonHDF: @qkoziol After thinking about this for a bit, it looks like the most reasonable way to handle this would be to add a `size_t` hint parameter ("no hint provided" represented as `SIZE_MAX`) to this function that lets the caller specify an expected number of entries for the I/O vectors and then check and grow the buffers if the caller was wrong with their hint. It's a little clunky, but shouldn't be bad to implement if it seems worth doing over another scan through the chunk list.

The issue is that in the current implementation there is no "top-level" data structure that gets passed between internal functions which could hold the expected number of entries. The main data structure is an array of one struct per chunk that has to be iterated through to determine how many chunks will be involved in the I/O.

qkoziol: I would suggest creating that 'top-level' struct, then you can put this counter in it, as well as a pointer to the array of chunks.  Then the # of parameters will stay the same, but you can pass around more information.

bmribler (approved):

lrknox (approved):

src/H5Dmpio.c
```diff
@@ -229,6 +229,36 @@ typedef struct H5D_filtered_collective_io_info_t {
...
 
     UT_hash_handle hh;
+} H5D_filtered_collective_chunk_info_t;
+
+/*
+ * Top-level structure that contains an array of H5D_filtered_collective_chunk_info_t
+ * chunk info structures for collective filtered I/O, as well as other useful information.
+ * The struct's fields are as follows:
+ *
+ * chunk_infos - An array of H5D_filtered_collective_chunk_info_t structures that contain
```
fortnern: Maybe "structures that each contain". Also, should probably mention the hash table field

src/H5Dmpio.c
```diff
@@ -1783,134 +1789,90 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, H5D_dset_io_info_
...
+        /* Perform vector I/O on chunks */
+        if (H5D__mpio_collective_filtered_vec_io(&chunk_list, io_info->f_sh, io_info->op_type) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "couldn't perform vector I/O on filtered chunks")
 
         /* Free up resources in anticipation of following collective operation */
-        for (i = 0; i < chunk_list_num_entries; i++) {
-            if (chunk_list[i].buf) {
-                H5MM_free(chunk_list[i].buf);
-                chunk_list[i].buf = NULL;
+        for (size_t i = 0; i < chunk_list.num_chunk_infos; i++) {
```
fortnern, line 1853: I was under the impression that our style conventions said not to place variable declarations in a loop statement like this. At least I haven't seen that anywhere else. If everyone wants to start doing this I don't have a problem with it , as long as there's not some obscure compiler we support that doesn't support this.

src/H5Dmpio.c
```diff
@@ -3271,6 +3210,20 @@ H5D__mpio_redistribute_shared_chunks(H5D_filtered_collective_io_info_t *chunk_li
                                                  io_info, mpi_rank, mpi_size) < 0)
         HGOTO_ERROR(H5E_DATASET, H5E_CANTREDISTRIBUTE, FAIL, "can't redistribute shared chunks")
 
+    /*
+     * Now that chunks have been redistributed, each rank must update
+     * their chunk list struct's `num_chunks_to_read` field since it
+     * may now be out of date.
+     */
+    for (i = 0; i < chunk_list->num_chunk_infos; i++) {
```
fortnern: It looks like this is redundant since this scan is also performed in H5D__mpio_redistribute_shared_chunks_int(). Am I missing something?

src/H5Dmpio.c
```diff
@@ -5578,274 +5420,146 @@ H5D__mpio_get_chunk_insert_info_types(MPI_Datatype *contig_type, hbool_t *contig
...
-    assert(file_type_derived);
+    assert(chunk_list);
+    assert(f_sh);
 
-    *mem_type_derived  = FALSE;
-    *file_type_derived = FALSE;
-    *new_mem_type      = MPI_BYTE;
-    *new_file_type     = MPI_BYTE;
+    if (op_type == H5D_IO_OP_WRITE)
+        iovec_count = (uint32_t)chunk_list->num_chunk_infos;
```
fortnern, line 5448: So if two processes write to a chunk, after chunks are assigned, the one that doesn't own it deletes the chunk from its chunk list?

src/H5Dmpio.c
```diff
@@ -5578,274 +5420,146 @@ H5D__mpio_get_chunk_insert_info_types(MPI_Datatype *contig_type, hbool_t *contig
...
-                if (MPI_SUCCESS != (mpi_code = MPI_Get_address(chunk_list[i].buf, &chunk_buf)))
-                    HMPI_GOTO_ERROR(FAIL, "MPI_Get_address failed", mpi_code)
+            /*
+             * Ensure the chunk list is sorted in ascending ordering of
+             * offset in the file. Note that we only compare that the
+             * current address is greater than the previous address and
+             * not equal to it; file addresses should only appear in the
+             * chunk list once.
+             */
+            if (vec_idx > 0)
```
fortnern, line 5512: I assume the compiler will generally optimize this out, but my preference would be to surround this in an #ifndef NDEBUG directive

fortnern:
Finished reviewing, looks good overall to me

src/H5Dmpio.c
jhendersonHDF: Thanks for the catch - I forgot to remove the old version after moving it down into the function

src/H5Dmpio.c
jhendersonHDF: Fixed

src/H5Dmpio.c
jhendersonHDF: Fixed

src/H5Dmpio.c
jhendersonHDF: That's correct and described in a little more detail in the header comment for H5D__mpio_share_chunk_modification_data().

fortnern (approved):

PR #3247, opened 2023-07-14
Fix some warnings in developer builds

jhendersonHDF:


src/H5.c
jhendersonHDF, line 1: The changes in this file fix Wjump-misses-init warnings by moving the initialization and termination sections into a new scope block. The malloc attribute also gets added to H5allocate_memory

testpar/API/t_coll_md_read.c
jhendersonHDF, line 1: No longer relevant with removal of the now unused H5D_ALL_CHUNK_ADDR_THRES_COL_NUM macro

testpar/t_coll_md.c
jhendersonHDF, line 1: No longer relevant with removal of the now unused H5D_ALL_CHUNK_ADDR_THRES_COL_NUM macro

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -197,6 +197,19 @@ else ()
   endif ()
 endif ()
 
+# Turn off -Winline warning for Developer builds as this flag
+# appears to conflict specifically with the -Og optimization
+# flag and will produce warnings about functions not being
+# considered for inlining under at least GNU compilers
+if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
```
byrnHDF: Why not just take it out of the gnu/developer-general file?

jhendersonHDF: It still seems like a good warning to have for -O1, -O2, -O3 builds where HDF5_ENABLE_DEV_WARNINGS is on. It might still be able to inform about why a function wasn't inlined and then someone can go inspect. Developer builds have HDF5_ENABLE_DEV_WARNINGS on by default, but if someone is building in RelWithDebInfo or similar with HDF5_ENABLE_DEV_WARNINGS on, then it makes sense to have -Winline there. It's just currently not compatible with what we probably think of as developer builds (developer or debug builds) since we add -Og in both cases.

byrnHDF: Secondly, it should probably be in the elseif block of line 187, which is where it would get added. Maybe take it out the developer-general file and only add it back in if supported? All in that elseif block.

jhendersonHDF: Moving into the elseif makes sense. As for taking the flag out and re-adding it later - that sort of seems to me like it defeats the purpose of having the warning flag files in the first place.

qkoziol (approved):

config/cmake/HDFCompilerFlags.cmake
byrnHDF: It depends on whether it is a usually used or a usually removed choice. Which one is more common?

jhendersonHDF: I suppose that depends on how many people usually turn on HDF5_ENABLE_DEV_WARNINGS

jhendersonHDF:
@byrnHDF I switched things over to adding -Winline rather than removing it. If that looks worse, I'll just revert the changes but at least move the cmake code into the if block

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -189,6 +189,19 @@ if (HDF5_ENABLE_DEV_WARNINGS)
   elseif (CMAKE_C_COMPILER_ID MATCHES "IntelLLVM" OR CMAKE_C_COMPILER_ID MATCHES "[Cc]lang")
     ADD_H5_FLAGS (H5_CFLAGS "${HDF5_SOURCE_DIR}/config/clang-warnings/developer-general")
   endif ()
+
+  # Turn on -Winline warnings now only for non-Debug and
+  # non-Developer builds. For at least GNU compilers this
+  # flag appears to conflict specifically with the -Og
+  # optimization flag and will produce warnings about functions
+  # not being considered for inlining
+  if (NOT ${HDF_CFG_NAME} MATCHES "Debug" AND NOT ${HDF_CFG_NAME} MATCHES "Developer")
```
byrnHDF, line 198: I have no problem with this arrangement.

Does release_docs/INSTALL_Warnings.txt section II. Default Warnings need changed?

byrnHDF (approved):

config/cmake/HDFCompilerFlags.cmake
jhendersonHDF: Looks like yes, but to be honest this looks like just another file that adds maintenance burden rather than directing people to the relevant files where they can see which warnings get added in

byrnHDF: Agree, just another issue to add to the list of documentation fixes.

src/H5Tref.c
```diff
@@ -1004,8 +1004,10 @@ H5T__ref_disk_write(H5VL_object_t H5_ATTR_UNUSED *src_file, const void *src_buf,
     src_size -= H5R_ENCODE_HEADER_SIZE;
 
 #ifndef NDEBUG
-    size_t buf_size_left = dst_size - sizeof(uint32_t);
-    assert(buf_size_left > sizeof(uint32_t));
+    {
+        size_t buf_size_left = dst_size - sizeof(uint32_t);
+        assert(buf_size_left > sizeof(uint32_t));
+    }
```
derobins, line 1010: C99 should be happy without the parens. Why add them?

derobins (approved):

PR #3254, opened 2023-07-18
changed the scope of #ifdef DOXYGEN to now include H5D multi-functions

brtnfld:


fortran/src/H5Dff.F90
```diff
@@ -2266,6 +2266,9 @@ SUBROUTINE h5dfill_char(fill_value, space_id, buf, hdferr)
     CALL h5dfill_ptr(f_ptr_fill_value, fill_type_id, f_ptr_buf, mem_type_id, space_id, hdferr)
 
   END SUBROUTINE h5dfill_char
+
+#endif
```
epourmal, line 2270: I have hard time finding where "if" begins. Would it be possible to add comments to such lines?

epourmal (approved):

lrknox (approved):

PR #3264, opened 2023-07-19
Another round of fixing -Wextra-semi-stmt warnings

seanm:


.clang-format
```diff
@@ -85,7 +85,6 @@ StatementMacros:
   - H5_GCC_DIAG_ON
   - H5_LEAVE
   - HDONE_ERROR
-  - HERROR
```
jhendersonHDF, line 88: Was this macro meant to be removed from StatementMacros even though this PR converts it to a statement like macro?

seanm: >Was this macro meant to be removed from StatementMacros

Yes.

>even though this PR converts it to a statement like macro?

This PR does some of that and some of the opposite. In the case of `HERROR` it was changed here to *require* a trailing semi-colon.

jhendersonHDF: Ah ok I was misinterpreting the purpose of StatementMacros. If my understanding is correct, does that mean HDONE_ERROR and HGOTO_DONE_TAG should also be removed from this file based on the changes in this PR?

seanm: For `HGOTO_DONE_TAG`, yes, you are correct. Fixed.

jhendersonHDF (approved):

.clang-format
jhendersonHDF: > For `HGOTO_DONE_TAG`, yes, you are correct. Fixed.

For my own education, what's different here between `HGOTO_DONE_TAG`/`HERROR` and `HDONE_ERROR`?

seanm: Ah, you are correct. I had double checked the 3 I did in the one commit, but `HDONE_ERROR` was in a different commit.  It should also be removed from the  `StatementMacros` list... will do that now...

(They all use the do-while(0) idiom, and such while statements have to have a trailing semicolon (which is not part of the macro) and so they are basically macros that behave like functions. Thus none of them should be in the StatementMacros list.)

jhendersonHDF: Gotcha, thanks for the clarification!

qkoziol (approved):

lrknox (approved):

PR #3306, opened 2023-07-28
Fixed -Wdeprecated-copy-dtor warnings by implementing a copy assignme…

seanm:
…nt operator

Example warning was:

warning: definition of implicit copy assignment operator for 'Group' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-dtor]

### DIFF
--- a/c++/src/H5Attribute.cpp
+++ b/c++/src/H5Attribute.cpp
@@ -605,4 +605,16 @@ Attribute::~Attribute()
     }
 }
 
+//--------------------------------------------------------------------------
+// Function:    Copy assignment operator
+Attribute &
+Attribute::operator=(const Attribute &original)
+{
+    if (&original != this) {
+        setId(original.id);
+    }
+
+    return *this;
+}
+
 } // namespace H5
--- a/c++/src/H5Group.cpp
+++ b/c++/src/H5Group.cpp
@@ -274,4 +274,16 @@ Group::~Group()
     }
 }
 
+//--------------------------------------------------------------------------
+// Function:    Copy assignment operator
+Group &
+Group::operator=(const Group &original)
+{
+    if (&original != this) {
+        setId(original.id);
+    }
+
+    return *this;
+}
+
 } // namespace H5
--- a/c++/src/H5Attribute.h
+++ b/c++/src/H5Attribute.h
@@ -78,6 +78,9 @@ class H5_DLLCPP Attribute : public AbstractDs, public H5Location {
     // Destructor: properly terminates access to this attribute.
     virtual ~Attribute() override;
 
+    // Copy assignment operator.
+    Attribute &operator=(const Attribute &original);
+
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
   protected:
     // Sets the attribute id.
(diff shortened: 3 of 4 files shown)


seanm:
@bmribler here is that WIP from the other PR...

c++/src/H5Attribute.cpp
```diff
@@ -607,7 +607,8 @@ Attribute::~Attribute()
 
 //--------------------------------------------------------------------------
 // Function:    Copy assignment operator
-Attribute & Attribute::operator= (const Attribute &other)
+Attribute &
+Attribute::operator=(const Attribute &other)
```
bmribler: Do you mind using "original" instead of "other" to keep it consistent with the existing code?

c++/src/H5Attribute.cpp
```diff
@@ -607,7 +607,8 @@ Attribute::~Attribute()
...
 {
     if (&other != this) {
     }
```
bmribler, line 615: setId(original.id); needed here so that both objects have the same hdf5 ID (see DataSet, DataType,...)

bmribler:
> @bmribler here is that WIP from the other PR...

Thanks, I got it.

c++/src/H5Attribute.cpp
seanm: Done.

c++/src/H5Attribute.cpp
seanm: Done. This is the crux of the WIP. I'm not sure what to actually put into these copy assignment operators...

derobins: I would say that a copy in this context is the equivalent of (re-)opening the thing in the same way, so you'd want a new, different ID. Unfortunately, we don't have a lightweight way of duplicating an ID, though that could be added pretty easily. There are calls that do what the library would do (e.g., H5Iobject_verify() and H5Iregister()), but we don't allow using those with library ID types (just user-defined ID types).

bmribler: In the HDF5 C++ API, the operator= uses the same ID as the original because if a different ID is created, it won't have the same characteristics as the original ID, specifically, if the rhs represents a named datatype, "this" would still be a transient datatype.  At least, that was my understanding at the time.  However, I just realized PropList didn't follow that concept but I don't remember why and I don't know if it needs to be fixed...  I'll have to look into that.

c++/src/H5Group.cpp
```diff
@@ -274,4 +274,15 @@ Group::~Group()
     }
 }
 
+//--------------------------------------------------------------------------
+// Function:    Copy assignment operator
+Group &
+Group::operator=(const Group &original)
+{
+    if (&original != this) {
+    }
```
bmribler, line 284: @seanm, I would do the same here as Attribute::operator=.

seanm: Done.

bmribler: Thanks!

bmribler (approved):

seanm:
@bmribler can you suggest / ping a second reviewer?

bmribler:
> @bmribler can you suggest / ping a second reviewer?

@seanm someone will review.

seanm:
@bmribler @lrknox @derobins friendly ping

derobins (approved):

PR #3309, opened 2023-07-29
Update so numbers according to API changes since version 1.14.1.

lrknox:


lrknox:
Numbers updated for code change and for only 2 added ros3 functions, no other public API changes since 1.14.1-2 release.

byrnHDF (approved):

PR #3313, opened 2023-07-31
Add test for H5Fget_obj_count with datatypes

mattjala:
The existing tests for this function are pretty sparse, especially with respect to datatypes.

This is expected to reveal an issue where, when counting the objects in all files, copies of transient datatypes are counted as open objects. Expected behavior is that only committed datatypes are counted as objects.

### DIFF
--- a/test/tfile.c
+++ b/test/tfile.c
@@ -178,6 +178,18 @@ const char *FILESPACE_NAME[] = {"tfilespace", NULL};
 #define DST_FILE "fill18_copy.h5"
 #define DSET_DS1 "DS1"
 
+/* Declarations for test_get_datatype_count() */
+#define DATATYPE_OBJ_COUNT_FILENAME_1 "datatype_obj_count_file1"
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

jhendersonHDF:
It's hard to tell exactly since the line numbers in CI builds don't seem to match up with the changes here, but based on the failure output it looks like this test is failing as expected. We should test this same change against 1.10 now to see if it fails in the same way.

mattjala:
> It's hard to tell exactly since the line numbers in CI builds don't seem to match up with the changes here, but based on the failure output it looks like this test is failing as expected. We should test this same change against 1.10 now to see if it fails in the same way.

I ran the same test on 1.10.10 locally, and it fails in the same way.

derobins:
* This requires a library fix. We can't check in tests that fail.
* This would change the semantics of the call since they are transient and not file objects.

PR #3314, opened 2023-08-01
Fix H5Otoken_to_str call in h5dump and other minor cleanup

jhendersonHDF:
This PR cleans up some errors h5dump outputs from running against the file in https://forum.hdfgroup.org/t/segmentation-fault-when-using-h5repack-on-netcdf-4-file/11297, but otherwise doesn't address the root cause of the issue.

### DIFF
--- a/src/H5Dint.c
+++ b/src/H5Dint.c
@@ -3540,11 +3540,11 @@ done:
 hid_t
 H5D_get_create_plist(const H5D_t *dset)
 {
-    H5P_genplist_t *dcpl_plist;    /* Dataset's DCPL */
-    H5P_genplist_t *new_plist;     /* Copy of dataset's DCPL */
-    H5O_layout_t    copied_layout; /* Layout to tweak */
-    H5O_fill_t      copied_fill;   /* Fill value to tweak */
-    H5O_efl_t       copied_efl;    /* External file list to tweak */
+    H5P_genplist_t *dcpl_plist;        /* Dataset's DCPL */
+    H5P_genplist_t *new_plist;         /* Copy of dataset's DCPL */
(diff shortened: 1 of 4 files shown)


tools/lib/h5tools_dump.c
jhendersonHDF, line 1: The code in this file should be refactored so it's easy to skip certain things when API calls fail while still printing as much as possible, but for now the changes in this file work around H5Dget_create_plist failed and also address a memory leak and three assertion failures.

byrnHDF (approved):

src/H5Dint.c
```diff
@@ -3697,11 +3697,15 @@ H5D_get_create_plist(const H5D_t *dset)
...
 
 done:
-    if (ret_value < 0)
+    if (ret_value < 0) {
         if (new_dcpl_id > 0)
             if (H5I_dec_app_ref(new_dcpl_id) < 0)
                 HDONE_ERROR(H5E_DATASET, H5E_CANTDEC, FAIL, "unable to close temporary object");
 
+        if (copied_fill.type && (H5T_close_real(copied_fill.type) < 0))
+            HDONE_ERROR(H5E_DATASET, H5E_CANTFREE, FAIL, "Can't free temporary datatype");
```
mattjala, line 3706: Is this error report redundant, since the call has already failed at this point?

mattjala (approved):

src/H5Dint.c
jhendersonHDF: One could make the argument that this pattern of reporting errors after the `done:` section in the library when a function has already failed is redundant, but sometimes the error messages can be useful and at least it's uniform so developers don't need to remember special rules about when to report an error or not.

Issue #3248, opened 2023-07-16
Loading plugin fails with missing directory


Grivisual:
While attempting to load a plugin the HDF5 library will fail if one of the directories in the plugin paths does not exist even if there are more paths to check.

File is HSPLpath.c in function HSPL__find_plugin_in_path_table. The guts of which are:

/* Loop over the paths in the table, checking for an appropriate plugin */
for (u = 0; u < H5PL_num_paths_g; u++) {

    /* Search for the plugin in this path */
    if (H5PL__find_plugin_in_path(search_params, found, H5PL_paths_g[u], plugin_info) < 0)
        HGOTO_ERROR(H5E_PLUGIN, H5E_CANTGET, FAIL, "search in path %s encountered an error",
                    H5PL_paths_g[u])

    /* Break out if found */
    if (*found) {
        if (!plugin_info)
            HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL, "plugin info should not be NULL")
        break;
    }
}
The call to HSPL__find_plugin_in_path will return -1 if the path does not exist, which causes a goto to ‘done’, even if all of HSPL_num_paths_g have yet to be checked.

Names change, directories move. If a path does not exist perhaps a warning could be generated but the rest of the paths still get checked.

PR #3315, opened 2023-08-01
Fix loading plugin fails with missing directory GH issue #3248

byrnHDF:
Fix issue #3248

### DIFF
--- a/src/H5PLpath.c
+++ b/src/H5PLpath.c
@@ -801,8 +801,7 @@ H5PL__find_plugin_in_path_table(const H5PL_search_params_t *search_params, hbool
 
         /* Search for the plugin in this path */
         if (H5PL__find_plugin_in_path(search_params, found, H5PL_paths_g[u], plugin_info) < 0)
-            HERROR(H5E_PLUGIN, H5E_CANTGET, "search in path %s encountered an error",
-                        H5PL_paths_g[u]);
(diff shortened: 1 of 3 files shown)


src/H5PLpath.c
```diff
@@ -801,8 +801,8 @@ H5PL__find_plugin_in_path_table(const H5PL_search_params_t *search_params, hbool
 
         /* Search for the plugin in this path */
         if (H5PL__find_plugin_in_path(search_params, found, H5PL_paths_g[u], plugin_info) < 0)
-            HGOTO_ERROR(H5E_PLUGIN, H5E_CANTGET, FAIL, "search in path %s encountered an error",
-                        H5PL_paths_g[u])
+            HDONE_ERROR(H5E_PLUGIN, H5E_CANTGET, FAIL, "search in path %s encountered an error",
+                        H5PL_paths_g[u]);
```
jhendersonHDF: With this change, you may want to consider clearing the error stack and setting `ret_value` to `SUCCEED` if the plugin gets found. Otherwise, I believe a missing directory will still fail this function and also print out the error stack for previous errors.

byrnHDF: Right, added test change and fix.

lrknox (approved):

jhendersonHDF (approved):

PR #3319, opened 2023-08-01
Fix for the bug exposed from running test/set_extent.c when selection…

vchoi-hdfgroup:
… I/O is enabled.

This is a fix from Neil.
The test/set_extent.c is modified to test for selection I/O enabled.


fortnern (approved):

fortnern:
Worth noting this change also enables in-place type conversion for the legacy (non-selection I/O) pathway. This was necessary to get some corner cases in the selection I/O pathway to work

fortnern:
We should remove the line "This is currently only used for selection I/O operations, which are used for collective I/O with type conversion." from the doxygen for H5Pset_modify_write_buf()

src/H5Dscatgath.c
```diff
@@ -490,46 +502,74 @@ H5D__scatgath_read(const H5D_io_info_t *io_info, const H5D_dset_io_info_t *dset_
...
-        /* Go figure out how many elements to read from the file */
         assert(H5S_SELECT_ITER_NELMTS(file_iter) == (dset_info->nelmts - smine_start));
-        smine_nelmts = (size_t)MIN(dset_info->type_info.request_nelmts, (dset_info->nelmts - smine_start));
+
+        /* Determine strip mine size. First check if we're doing in-place type conversion */
+        if (in_place_tconv) {
+            /* If this is not a selection I/O operation and there is a background buffer, we cannot exceed
+             * request_nelmts.  It could be part of a selection I/O operation if this read is used to fill in
+             * a nonexistent chunk */
+            if (dset_info->type_info.need_bkg &&
```
jhendersonHDF: I'm unsure of the differences between `H5T_BKG_YES` and `H5T_BKG_TEMP`. Should this be compared as `(H5T_BKG_YES == dset_info->type_info.need_bkg)` as below, rather than just checking if `need_bkg` isn't 0?

fortnern: Background buffer can be either not needed at all (need_bkg is 0), needed for conversion but the contents on entry don't matter (H5T_BKG_TEMP), or it needs to be initialized with the destination contents (H5T_BKG_YES). This code was put in to avoid running off the end of the background buffer during in-place type conversion. Normally in-place type conversion can use the whole width of the I/O since the type conversion buffer is the read buffer, but if the background buffer was allocated by the non-selection pathway (and is therefore limited in size) then we need to limit the strip mine size to the size of the background buffer.. Come to think of it the check for !H5D__SCATGATH_USE_CMPD_OPT_READ is redundant since if in_place_tconv is true, as is the condition to enter this code, that macro will always be false.

fortnern: So for this check it doesn't matter if the background buffer needs to be initialized or not, TEMP or YES should both trigger the limited strip mine size.

jhendersonHDF (approved):

src/H5Dscatgath.c
```diff
@@ -633,29 +679,57 @@ H5D__scatgath_write(const H5D_io_info_t *io_info, const H5D_dset_io_info_t *dset
...
-        n = H5D__gather_mem(buf, mem_iter, smine_nelmts, io_info->tconv_buf /*out*/);
-        if (n != smine_nelmts)
-            HGOTO_ERROR(H5E_IO, H5E_WRITEERROR, FAIL, "mem gather failed");
+        /* Determine strip mine size. First check if we're doing in-place type conversion */
+        if (in_place_tconv) {
+            /* If this is not a selection I/O operation and there is a background buffer, we cannot exceed
+             * request_nelmts.  It could be part of a selection I/O operation if this is used to write the
+             * fill value to a cached chunk that will immediately be evicted. */
+            if (dset_info->type_info.need_bkg &&
+                !H5D__SCATGATH_USE_CMPD_OPT_WRITE(dset_info, in_place_tconv) &&
```
fortnern: Vailin, can you  try changing this to:
`assert(!H5D__SCATGATH_USE_CMPD_OPT_WRITE(dset_info, in_place_tconv));`
`if (dset_info->type_info.need_bkg && (io_info->use_select_io != H5D_SELECTION_IO_MODE_ON))`

And the equivalent change on read?

vchoi-hdfgroup: Done

PR #3330, opened 2023-08-02
Fix CVE-2018-11202

derobins:
A malformed file could result in chunk index memory leaks. Under most conditions (i.e., when the --enable-using-memchecker option is NOT used), this would result in a small memory leak and and infinite loop and abort when shutting down the library. The infinite loop would be due to the "free list" package not being able to clear its resources so the library couldn't shut down. When the "using a memory checker" option is used, the free lists are disabled so there is just a memory leak with no abort on library shutdown.

The chunk index resources are now correctly cleaned up when reading misparsed files and valgrind confirms no memory leaks.

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -961,10 +961,13 @@ H5D__chunk_init(H5F_t *f, const H5D_t *const dset, hid_t dapl_id)
         HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "unable to set # of chunks for dataset");
 
 done:
-    if (FAIL == ret_value)
-        if (H5D__chunk_dest(dset) < 0)
-            HDONE_ERROR(H5E_DATASET, H5E_CANTFREE, FAIL,
-                        "unable to clean up chunk structures during error cleanup");
+    if (FAIL == ret_value) {
+        if (rdcc->slot)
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

src/H5Dchunk.c
```diff
@@ -961,6 +961,13 @@ H5D__chunk_init(H5F_t *f, const H5D_t *const dset, hid_t dapl_id)
         HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "unable to set # of chunks for dataset");
 
 done:
+    if (FAIL == ret_value) {
+        if (rdcc->slot)
+            rdcc->slot = H5FL_SEQ_FREE(H5D_rdcc_ent_ptr_t, rdcc->slot);
+
+        if (sc->ops->dest && (sc->ops->dest)(&idx_info) < 0)
```
fortnern: I believe we need to keep track of whether the init function was called. Using b-tree v1 as an example the dest callback appears to always call H5UC_DEC on the shared data, and H5UC_DEC assumes the shared data exists.

derobins: Okay, I'll add a flag when I get back from lunch.

jhendersonHDF (approved):

fortnern (approved):

PR #3336, opened 2023-08-02
Fix CVE-2018-13867

derobins:
A corrupt file containing an invalid local heap datablock address could trigger an assert failure when the metadata cache attempted to load the datablock from storage.

The local heap now verifies that the datablock address is valid when the local heap header information is parsed.

### DIFF
--- a/src/H5HLcache.c
+++ b/src/H5HLcache.c
@@ -194,6 +194,12 @@ H5HL__hdr_deserialize(H5HL_t *heap, const uint8_t *image, size_t len, H5HL_cache
         HGOTO_ERROR(H5E_HEAP, H5E_OVERFLOW, FAIL, "ran off end of input buffer while decoding");
     H5F_addr_decode_len(udata->sizeof_addr, &image, &(heap->dblk_addr));
 
+    /* Check that the datablock address is valid (might not be true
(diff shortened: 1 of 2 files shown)


src/H5HLcache.c
```diff
@@ -194,6 +194,12 @@ H5HL__hdr_deserialize(H5HL_t *heap, const uint8_t *image, size_t len, H5HL_cache
         HGOTO_ERROR(H5E_HEAP, H5E_OVERFLOW, FAIL, "ran off end of input buffer while decoding");
     H5F_addr_decode_len(udata->sizeof_addr, &image, &(heap->dblk_addr));
 
+    /* Check that the datablock address is valid (might not be true
+     * in a corrupt file)
+     */
+    if (!H5_addr_defined(heap->dblk_addr))
+        HGOTO_ERROR(H5E_HEAP, H5E_BADVALUE, FAIL, "bad datablock address");
```
derobins, line 201: This only checks for HADDR_UNDEF, which trips an assert later. Any other invalid address for the file will be reported via normal error handling when the cache tries to load the datablock.

jhendersonHDF (approved):

lrknox (approved):

PR #3341, opened 2023-08-03
Add Fortran ES module to deploy list

byrnHDF:


brtnfld (commented):
Is it not possible to use a list instead of a set? That way, we can have just one place to add module names and append the correct directory when needed.

byrnHDF:
Probably, just hadn't thought of it, I guess.

brtnfld (approved):

lrknox (approved):

PR #3343, opened 2023-08-03
Set version to 1.14.3-1.

lrknox:


lrknox:
Normally the RELEASE.txt file should be cleared out in the 1.14 develop branch when the release branch is created, but since we're still bringing changes with release notes into the hdf5_1_14 branch that we will want to take to the hdf5_1_14_2 branch for release, I'm deferring that task until the release branch is settled.

byrnHDF (approved):

lrknox:
clang_format check changed 3 lines and committed the changes.  All other tests passed.

PR #3352, opened 2023-08-04
Make h5dump spacing consistent when printing VLEN datatype (#3351)

jhendersonHDF:


tools/lib/h5tools_dump.c
```diff
@@ -2571,7 +2571,7 @@ h5tools_print_datatype(FILE *stream, h5tools_str_t *buffer, const h5tool_format_
             if (H5Tclose(super) < 0)
                 H5TOOLS_ERROR((-1), "H5Tclose failed");
 
-            h5tools_str_append(buffer, "%s", h5tools_dump_header_format->vlenblockend);
+            h5tools_str_append(buffer, " %s", h5tools_dump_header_format->vlenblockend);
```
byrnHDF, line 2574: This should have been changed where vlenblockend is defined.

byrnHDF: See line 159 of same file.

byrnHDF (changes requested):

tools/lib/h5tools_dump.c
byrnHDF: In fact, line 2567 is wrong and line 158 should be "H5T_VLEN { "

byrnHDF: Another one: line 2239 should change line 152 to be: "H5T_STRING {"

byrnHDF: Actually that whole function, h5tools_print_datatype, needs to fixed.

byrnHDF:
Wow, looking at again, the h5tools_dump_header_t define should not have included types like string and vlen and enum

byrnHDF:
I will fix this in develop.

byrnHDF:
This change is okay with #3353 created to fix issues

byrnHDF (approved):

jhendersonHDF:
I think the main issue here was just that the printing code isn't always consistent about who is responsible for adding spacing. It should be caller or callee, but not both. If the code in `h5tools_print_datatype` adds a space beforehand, like in "H5T_ARRAY { ", then it should also be responsible for adding the end spacing, as in " }" like it does now. Otherwise, it shouldn't add either space and let the actual datatype printing code do that. Otherwise you get these problems, especially when things are recursive.

byrnHDF:
I think that was the intention and the reason for my initial comments.
So there was two solutions, create PR #3353 and make everything use the formatting in the function
or add the missing types to dataformat struct and use the specific struct entry each with just a print "%s". Which would be correct. I went with the PR so as not to change the sorta-public API.

jhendersonHDF:
> I think that was the intention and the reason for my initial comments. So there was two solutions, create PR #3353 and make everything use the formatting in the function or add the missing types to dataformat struct and use the specific struct entry each with just a print "%s". Which would be correct. I went with the PR so as not to change the sorta-public API.

I think the struct types aren't meant to have any spacing in them. They're just for changing what the "begin block" and "end block" delimiters are and the formatting always needs to be down in either `h5tools_print_datatype` or whatever function that calls to print out a particular datatype. Seems to me like all the entries in the dataformat struct should just be a single character.

PR #3353, opened 2023-08-04
Add other types and full type to enum/str/vlen dataformat for structblock[begin/end]

byrnHDF:


tools/lib/h5tools_dump.c
```diff
@@ -2556,22 +2556,22 @@ h5tools_print_datatype(FILE *stream, h5tools_str_t *buffer, const h5tool_format_
...
-            h5tools_str_append(buffer, "H5T_VLEN %s ", h5tools_dump_header_format->vlenblockbegin);
+            h5tools_str_append(buffer, "H5T_VLEN %s ", h5tools_dump_header_format->structblockbegin);
 
             h5tools_print_datatype(stream, buffer, info, ctx, super, TRUE);
 
             if (H5Tclose(super) < 0)
                 H5TOOLS_ERROR((-1), "H5Tclose failed");
 
-            h5tools_str_append(buffer, " %s", h5tools_dump_header_format->vlenblockend);
+            h5tools_str_append(buffer, " %s", h5tools_dump_header_format->structblockend);
```
jhendersonHDF: Are these not defined exactly the same? Looks like the idea was to make it easy to change the graphical delimiter for different types. With this change it's all or nothing.

byrnHDF: Only three of the types are in the struct, this is more consistently applied.

jhendersonHDF: I'd assume that's because it doesn't make sense to have a begin and end block for building block types like `H5T_NATIVE_INT`. This may be consistent but it seems like it's taking away an easy way to change how the output looks.

byrnHDF: Yes, but only three datatypes are available [str, enum and vlen]!
The others must be changed in the function code.
So this makes the usage consistent. The better way would be to allow all datatypes to be changed. That will change program design, this PR doesn't .

byrnHDF: I'm not proposing the absolute way it should be, only making the current state consistent.

jhendersonHDF: > Yes, but only three datatypes are available [str, enum and vlen]!

There's also opaque and compound that currently use `structblockbegin` and `structblockend`. I think it's a little odd to use `structblockbegin` for opaque types, but it appears that's because the DDL for an opaque type is a pseudo-structure since it includes the opaque tag:

    DATATYPE  H5T_OPAQUE {
      OPAQUE_TAG "test opaque type";
    }

I would've at least have created an `opaqueblockbegin` and `opaqueblockend` rather than reusing `structblockbegin` since the "struct" there was clearing referring to use with compound datatypes, but it is what it is.

> The better way would be to allow all datatypes to be changed. That will change program design, this PR doesn't.

What I'm saying is there isn't anything to change for other datatypes because they aren't structured types, they're basic building block types. In the DDL for an 8-bit little-endian integer, you just get `DATATYPE  H5T_STD_I8LE`, which looks reasonable to me and makes sense why there isn't an `intblockbegin`/`intblockend`. I actually think moving all these types to using only `structblockbegin`/`structblockend` is actually less consistent since to me it again seems clear that "struct" refers to compound datatypes only.

If we're going to remove the other begin/end fields it would be nice to just remove them from the struct, but it's unfortunate that they're part of the public API in the tools library (though that seems to me like another argument for not changing which fields are being used, even if they're all just mapped to `{`/`}`).

jhendersonHDF (approved):

tools/lib/h5tools_dump.c
```diff
@@ -149,14 +149,14 @@ const h5tools_dump_header_t h5tools_standardformat = {
...
     "{",  /*udlinkblockbegin */
     "}",  /*udlinkblockend */
-    "{",  /*strblockbegin */
-    "}",  /*strblockend */
-    "{",  /*enumblockbegin */
-    "}",  /*enumblockend */
+    "{",  /*DO NOT USE strblockbegin */
+    "}",  /*DO NOT USE strblockend */
+    "{",  /*DO NOT USEenumblockbegin */
+    "}",  /*DO NOT USEenumblockend */
```
derobins: Should have a space after DO NOT USE

tools/lib/h5tools_dump.c
byrnHDF: sorry I was unclear, I am thinking about the datatypes in the function switch statements where the original vlen change was made. Each use of structblock is the same "H5T_XXXX %s" printf. I would move those to the struct definines so programs could change those instances.

tools/lib/h5tools_dump.c
byrnHDF: oh yes!

byrnHDF:
Changed to using the dataformat correctly. (Since we don't publish the header, yet)

derobins (approved):

PR #3368, opened 2023-08-10
Fix Heap-buffer-overflow WRITE in H5MM_memcpy

### LINKED ISSUE #3367: Out of bounds write in H5O__alloc_chunk
Found by fuzzing https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58658

Occures in
https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5Oalloc.c#L950-L951 when `p` incremented in the loop has no room to accommodate the data.

sashashura:
Closes #3367
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58658

The root cause is that [`H5MM_memcpy(p, ...)`](https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5Oalloc.c#L950-L951) is done in the loop [`for (u = 0, curr_msg = &oh->mesg[0]; u < oh->nmesgs; u++, curr_msg++)`](https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5Oalloc.c#L938). The [`p` is advanced forward and `size` is decreased](https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5Oalloc.c#L958-L959) on every copy, however there is no check that the `size` is still enough to accommodate the data.

### DIFF
--- a/src/H5Oalloc.c
+++ b/src/H5Oalloc.c
@@ -946,6 +946,9 @@ H5O__alloc_chunk(H5F_t *f, H5O_t *oh, size_t size, size_t found_null, const H5O_
                     else {
                         assert(curr_msg->type->id != H5O_CONT_ID);
 
+                        if (size < curr_msg->raw_size + (size_t)H5O_SIZEOF_MSGHDR_OH(oh))
+                            HGOTO_ERROR(H5E_OHDR, H5E_BADVALUE, FAIL, "invalid size");
+
                         /* Copy the raw data */
                         H5MM_memcpy(p, curr_msg->raw - (size_t)H5O_SIZEOF_MSGHDR_OH(oh),
                                     curr_msg->raw_size + (size_t)H5O_SIZEOF_MSGHDR_OH(oh));
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -523,6 +523,10 @@ Bug Fixes since HDF5-1.14.0 release
 
       Fixes Github issue #3034
 
+    - Fixed write buffer overflow in H5O__alloc_chunk
+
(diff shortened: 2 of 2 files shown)


sashashura:
Hi, do you have any questions? Are we waiting for someone?

jhendersonHDF:
> Hi, do you have any questions? Are we waiting for someone?

Hi @sashashura, I don't believe this is waiting on anyone and is currently just in the todo list among PRs to review. However, it would be helpful if you could add a short description of the issue and fix in release_docs/RELEASE.txt under the "Bug Fixes ... -> Library" section. You can see other entries there for a reference, but entries tend to be similar to:

    - Short description (with any applicable CVE numbers)

      Longer description

mattjala (approved):

jhendersonHDF (approved):

src/H5Oalloc.c
```diff
@@ -946,6 +946,9 @@ H5O__alloc_chunk(H5F_t *f, H5O_t *oh, size_t size, size_t found_null, const H5O_
                     else {
                         assert(curr_msg->type->id != H5O_CONT_ID);
 
+                        if (size < curr_msg->raw_size + (size_t)H5O_SIZEOF_MSGHDR_OH(oh))
+                            HGOTO_ERROR(H5E_OHDR, H5E_BADVALUE, FAIL, "invalid size");
+
```
derobins, line 951: We should probably make a sanitization pass over this whole function

derobins (approved):

PR #3373, opened 2023-08-10
Fix Use-of-uninitialized-value in H5FD__sec2_write

### LINKED ISSUE #3372: Use-of-uninitialized-value in H5FD__sec2_write
Found by fuzzing https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58644

An uninitialized memory potentially containing sensitive data is written into a file in https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5FDsec2.c#L794

sashashura:
Closes #3372
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58644

In the reproducer case it allocates 71 bytes for the buffer in https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5Centry.c#L548. Then it initializes only the first 49 bytes and the last 4 bytes in https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5FScache.c#L1233-L1237 and writes the whole 71 bytes buffer in https://github.com/HDFGroup/hdf5/blob/bf68e6eb3a2e9c209024769c1193fe31ffc39e23/src/H5FDsec2.c#L794

### DIFF
--- a/src/H5FScache.c
+++ b/src/H5FScache.c
@@ -1233,7 +1233,7 @@ H5FS__cache_sinfo_serialize(const H5F_t *f, void *_image, size_t len, void *_thi
     /* There may be empty space between entries and chksum */
     chksum_image    = (uint8_t *)(_image) + len - H5FS_SIZEOF_CHKSUM;
     /* Reset the unused space between entries and chksum */
-    memset(image, 0, len - H5FS_SIZEOF_CHKSUM - ((void*)image - _image));
+    memset(image, 0, len - H5FS_SIZEOF_CHKSUM - ((void *)image - _image));
     metadata_chksum = H5_checksum_metadata(_image, (size_t)(chksum_image - (uint8_t *)_image), 0);
     /* Metadata checksum */
     UINT32ENCODE(chksum_image, metadata_chksum);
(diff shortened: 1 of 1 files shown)


jhendersonHDF:
@sashashura, Apologies, it looks like I fixed this same issue in [31f4d5aa4e1f174cef8163e4af40f7bea728cd0c](https://github.com/HDFGroup/hdf5/commit/31f4d5aa4e1f174cef8163e4af40f7bea728cd0c) when I ran into it while working on a feature. Can you confirm that the latest develop fixes this for you?

src/H5FScache.c
```diff
@@ -1231,7 +1231,9 @@ H5FS__cache_sinfo_serialize(const H5F_t *f, void *_image, size_t len, void *_thi
     /* Compute checksum */
 
     /* There may be empty space between entries and chksum */
-    chksum_image    = (uint8_t *)(_image) + len - H5FS_SIZEOF_CHKSUM;
+    chksum_image = (uint8_t *)(_image) + len - H5FS_SIZEOF_CHKSUM;
+    /* Reset the unused space between entries and chksum */
+    memset(image, 0, len - H5FS_SIZEOF_CHKSUM - ((void *)image - _image));
```
derobins, line 1236: I'm not opposed to merging this in, but I feel like it's better to just memset the entire buffer and avoid the pointer math.

sashashura:
Yes, fixed.

derobins:
> Yes, fixed.

Great!

Thanks for your fixes, btw. Sorry it took us so long to get to them.

PR #3387, opened 2023-08-17
Fix assertion failure during file close on error

jhendersonHDF:
This PR fixes the assertion failure:

/src/H5MF.c:2585: H5MF_settle_raw_data_fsm: Assertion `f->shared->sblock' failed.

### DIFF
--- a/src/H5C.c
+++ b/src/H5C.c
@@ -416,7 +416,7 @@ H5C_prep_for_file_close(H5F_t *f)
 
 #ifdef H5_HAVE_PARALLEL
     if ((H5F_INTENT(f) & H5F_ACC_RDWR) && !image_generated && cache_ptr->aux_ptr != NULL &&
-        f->shared->fs_persist) {
+        f->shared->sblock && f->shared->fs_persist) {
         /* If persistent free space managers are enabled, flushing the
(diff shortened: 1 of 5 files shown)


src/H5C.c
```diff
@@ -416,7 +416,7 @@ H5C_prep_for_file_close(H5F_t *f)
 
 #ifdef H5_HAVE_PARALLEL
     if ((H5F_INTENT(f) & H5F_ACC_RDWR) && !image_generated && cache_ptr->aux_ptr != NULL &&
-        f->shared->fs_persist) {
+        f->shared->sblock && f->shared->fs_persist) {
```
jhendersonHDF, line 419: Don't try to serialize the cache if we failed to init the superblock, because `H5MF_settle_raw_data_fsm` assumes it's available and would crash otherwise. Also, for self-referential free space allocators, `H5MF_settle_meta_data_fsm` causes them to allocate space for themselves, which later leads to a failure in the assertion `H5AC_cache_is_clean(f, H5AC_RING_MDFSM)` in `H5F__dest`.

byrnHDF (approved):

lrknox (approved):

PR #3389, opened 2023-08-17
Fix valgrind warning about write of uninitialized bytes

jhendersonHDF:


byrnHDF (approved):

src/H5FScache.c
```diff
@@ -1231,7 +1232,17 @@ H5FS__cache_sinfo_serialize(const H5F_t *f, void *_image, size_t len, void *_thi
...
+    chksum_image = (uint8_t *)(_image) + len - H5FS_SIZEOF_CHKSUM;
+
+    /*
+     * If there is any empty space between the entries and
+     * checksum, make sure that the space is initialized
+     * before serializing it
+     */
+    gap_size = chksum_image - image;
+    if (gap_size > 0)
+        memset(image, 0, (size_t)gap_size);
```
lrknox, line 1244: I think the comment in line 1234 implies that there may be a buffer with data from image up to a gap followed by chksum_image.  If there is a gap, is the uninitialized memory at image or after the end of the image data?

jhendersonHDF: In this case, `image` is the pointer into the buffer to be serialized after all the processing has been done (adding in the info that's getting serialized). That pointer gets advanced in the buffer each time data is added. `chksum_image` is a pointer to the end of the whole allocated buffer, minus the size in bytes used to store the metadata checksum in the buffer. Sometimes `image == chksum_image` and sometimes `image < chksum_image`. In the case where `image` didn't cover the full allocated buffer (which can happen for different reasons like if the free space managers give this a free space block that's larger than needed), this clears the memory after `image` and up to `chksum_image` so that those last bytes in the buffer aren't random data that gets written to the file.

I'd expected that the checksum function would have complained about trying to use the uninitialized bytes, but there are comments in there about skipping the last block, so I'm guessing those bytes are in the block being skipped.

lrknox: Ok, that makes sense.  I was missing that the image pointer is pointing to the end of the data, not the beginning.

lrknox (approved):

PR #3407, opened 2023-08-22
Fix ph5diff tests for MPIEXEC_MAX_NUMPROCS=1

jhendersonHDF:


examples/CMakeTests.cmake
```diff
@@ -122,8 +122,8 @@ if (H5_HAVE_PARALLEL AND HDF5_TEST_PARALLEL AND NOT WIN32)
       add_test (NAME MPI_TEST_EXAMPLES-${parallel_example} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${NUMPROCS} ${MPIEXEC_PREFLAGS} $<TARGET_FILE:${parallel_example}> ${MPIEXEC_POSTFLAGS})
     else ()
       add_test (NAME MPI_TEST_EXAMPLES-${parallel_example} COMMAND "${CMAKE_COMMAND}"
-          -D "TEST_PROGRAM=${MPIEXEC_EXECUTABLE};${MPIEXEC_NUMPROC_FLAG};${NUMPROCS};${MPIEXEC_PREFLAGS};$<TARGET_FILE:${parallel_example}>;${MPIEXEC_POSTFLAGS}"
-          -D "TEST_ARGS:STRING="
+          -D "TEST_PROGRAM=${MPIEXEC_EXECUTABLE}"
+          -D "TEST_ARGS:STRING=${MPIEXEC_NUMPROC_FLAG};${NUMPROCS};${MPIEXEC_PREFLAGS};$<TARGET_FILE:${parallel_example}>;${MPIEXEC_POSTFLAGS}"
```
jhendersonHDF, line 126: Having these test arguments as part of TEST_PROGRAM was causing CMake to incorrectly interpret the test command. I'm not sure that the ph5diff tests were running correctly to begin with. Setting MPIEXEC_MAX_NUMPROCS=1 shows this issue by making all the ph5diff tests fail.

tools/src/h5diff/ph5diff_main.c
```diff
@@ -63,7 +63,7 @@ main(int argc, char *argv[])
     MPI_Comm_size(MPI_COMM_WORLD, &g_nTasks);
 
     if (g_nTasks == 1) {
-        printf("Only 1 task available...doing serial diff\n");
+        fprintf(stderr, "Only 1 task available...doing serial diff\n");
```
jhendersonHDF, line 66: This output statement makes it so the same output diff reference files can't be used between h5diff and ph5diff testing. Move this print statement to stderr since it doesn't seem that important anyway.

tools/src/h5diff/ph5diff_main.c
```diff
@@ -279,11 +279,12 @@ h5diff_exit(int status)
             phdiff_dismiss_workers();
             MPI_Barrier(MPI_COMM_WORLD);
         }
-        MPI_Finalize();
-        status = EXIT_SUCCESS; /* Reset exit status, since some mpiexec commands generate output on failure
-                                  status */
     }
 
+    MPI_Finalize();
```
jhendersonHDF, line 284: The call to MPI_Finalize() needed to be moved outside the above block. Otherwise it wouldn't be called when run with `mpiexec -np 1` and would cause the program to abort with an abnormal termination message.

tools/src/h5diff/ph5diff_main.c
```diff
@@ -279,11 +279,12 @@ h5diff_exit(int status)
...
+
+    status = EXIT_SUCCESS; /* Reset exit status, since some mpiexec commands generate output on failure status */
```
jhendersonHDF: Since the CMake testing can filter out stderr now, we likely should change this back to having ph5diff return the correct values instead of always returning 0, but it's simplest to leave this as is for now.

lrknox (approved):

derobins (approved):

PR #3412, opened 2023-08-23
Enable szip by default in Autotools

derobins:
Since libaec is so prevalent and BSD-licensed for both encode and decode, we build the szip filter by default when the szip or aec libraries are found.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -47,6 +47,12 @@ New Features
 
     Configuration:
     -------------
+    - Autotools builds now build the szip filter by default when an appropriate
+      library is found
+
+      Since libaec is prevalent and BSD-licensed for both encoding and
+      decoding, we build the szip filter by default now.
+
     - Removed CMake cross-compiling variables
 
       * HDF5_USE_PREGEN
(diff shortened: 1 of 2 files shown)


bmribler (approved):

release_docs/RELEASE.txt
```diff
@@ -47,6 +47,12 @@ New Features
 
     Configuration:
     -------------
+    - Autotools builds now build the szip filter by default when an appropriate
+      library is found
+
+      Since libaec is prevalent and BSD-licensed for both encoding and
+      decoding, we build the szip filter by default now.
+
```
lrknox, line 55: This is true for CMake, but looking at the code in configure.ac and configure I believe that for autotools using a found szip will require a prebuilt libsz.  If found it will build it into HDF5, but I don't see that autotools will build it from source.

byrnHDF (approved):

release_docs/RELEASE.txt
derobins: The Autotools can't build libraries from source like CMake

lrknox: That's what I thought.  Perhaps - 

    Autotools builds now include the szip filter by default when and appropriate library is found.

    Since libaec is prevalent and BSD-licensed for both encoding and decoding, the default behavior for both CMake 
    and Autotools builds will  include an appropriate and available szip library.  CMake will download and build the 
    libaec szip library by default.

Another question - what if HDF5 finds the old szip with license restrictions for encoding?

PR #3418, opened 2023-08-24
Revise file close assertion failure fix

jhendersonHDF:


jhendersonHDF:
@qkoziol Any thoughts on this PR? I originally created https://github.com/HDFGroup/hdf5/pull/3387 to fix an assertion failure during file close when a file fails to be created due to setting persistent, paged file free space management with too low of a library version bounds setting. However, that caused testphdf5 to start failing ONLY when free lists are enabled.

What happens is that the call to `H5MF_alloc` to allocate space in the file for the superblock causes the library to allocate memory for the free space manager and insert that entry into the cache. But `H5F__super_init` doesn't know that that happened and doesn't clean that up on failure, so the library fails in `H5_term_library -> H5SL_term_package -> H5FL_fac_term` because the free list entry for it is still allocated. When free lists are disabled, this just shows up as a memory leak and the library otherwise exits fine. Depending on the build mode, this library failure causes different behavior:

- Release mode: The library segfaults due to trying to print an error after the H5E package has already been brought down in `H5_term_library`. Since the H5FL package termination comes after H5E, this seems problematic in general, but that's a different discussion.
- Debug mode: The library fails with the assertion failure: `testphdf5: .../src/H5Eint.c:639: H5E_printf_stack: Assertion 'cls_id > 0' failed.` for the same reason above.
- Developer mode: Runs fine because free lists are disabled.

For the error case, the changes here add an `H5MF_xfree` call to free the file space that was allocated for the superblock and then a call to `H5MF_try_close` to try to shut down the free space managers before returning so that the metadata cache will be clean and `H5F__dest` will be happy later on. The call to `H5MF_try_close` is currently made if `non_default_fs_settings` is TRUE. That doesn't seem very ideal since the library defaults could change, but I couldn't think of a better way to have `H5F__super_init` determine whether it needs to try to close the free space managers.

src/H5C.c
```diff
@@ -416,7 +416,7 @@ H5C_prep_for_file_close(H5F_t *f)
 
 #ifdef H5_HAVE_PARALLEL
     if ((H5F_INTENT(f) & H5F_ACC_RDWR) && !image_generated && cache_ptr->aux_ptr != NULL &&
-        f->shared->sblock && f->shared->fs_persist) {
```
jhendersonHDF, line 419: This just goes back to the previous behavior and tries to close as much of the cache as possible, even if the superblock isn't available.

src/H5Fint.c
```diff
@@ -1624,8 +1624,7 @@ H5F__dest(H5F_t *f, hbool_t flush)
         HDONE_ERROR(H5E_FILE, H5E_CANTINIT, FAIL, "problems closing file");
     f->shared = NULL;
 
-    if (ret_value >= 0)
-        f = H5FL_FREE(H5F_t, f);
+    f = H5FL_FREE(H5F_t, f);
```
jhendersonHDF: Always free the H5F_t structure even if H5F__dest fails. Otherwise, this just leads to a memory leak.

src/H5MF.c
```diff
@@ -1058,7 +1058,6 @@ H5MF_xfree(H5F_t *f, H5FD_mem_t alloc_type, haddr_t addr, hsize_t size)
     assert(f);
     if (!H5_addr_defined(addr) || 0 == size)
         HGOTO_DONE(SUCCEED);
-    assert(addr != 0); /* Can't deallocate the superblock :-) */
```
jhendersonHDF: The intention here is clear, but it seems like there is a valid reason to call H5MF_xfree on the superblock for these changes. The alternative would be to create a counterpart function solely for the superblock, but that didn't really seem worth it.

src/H5MF.c
```diff
@@ -2581,16 +2580,14 @@ H5MF_settle_raw_data_fsm(H5F_t *f, hbool_t *fsm_settled)
...
-
         /* should only be called if file is opened R/W */
         assert(H5F_INTENT(f) & H5F_ACC_RDWR);
 
         /* shouldn't be called unless we have a superblock supporting the
          * superblock extension.
          */
-        assert(f->shared->sblock->super_vers >= HDF5_SUPERBLOCK_VERSION_2);
+        if (f->shared->sblock)
+            assert(f->shared->sblock->super_vers >= HDF5_SUPERBLOCK_VERSION_2);
```
jhendersonHDF, line 2591: In this and the below changes, try to close down as much as possible in the raw data FSM, even if the superblock isn't available. A bit dangerous if something else here uses the superblock, but I followed things far down and didn't really see anything else.

qkoziol:
I'll take a look tomorrow

src/H5Cimage.c
```diff
@@ -654,6 +654,11 @@ H5C__load_cache_image(H5F_t *f)
     } /* end if */
 
 done:
+    if (ret_value < 0) {
+        if (H5_addr_defined(cache_ptr->image_addr))
+            cache_ptr->image_buffer = H5MM_xfree(cache_ptr->image_buffer);
+    }
```
jhendersonHDF, line 660: Memory leak exposed by these file create failure changes. Free cache image pointer if this function fails.

src/H5Fint.c
jhendersonHDF: Revised this based on [this comment](https://github.com/HDFGroup/hdf5/blob/develop/hl/test/test_file_image.c#L368-L380). Always freeing the file structure was leading to some double frees when a file fails to be closed from H5Fclose and then the library later goes to free the structure to clear out IDs. Instead, only free the structure on failure if the call to `H5F__dest` came from `H5F_open`. Otherwise, it'll be freed later on.

src/H5MF.c
jhendersonHDF: Put this back, as the call to `H5MF_xfree` for freeing the file space allocated for the superblock was occasionally allocating a new sector in the file free space sections and causing more issues. Just calling `H5MF_try_close` appears to be enough to fix the issues here.

lrknox (approved):

byrnHDF (approved):

qkoziol:
Merged already, but this looks good

PR #3427, opened 2023-08-24
Fix for CVE-2018-15671

vchoi-hdfgroup:
h5stat -S $POC will result in a crash with s…egmenetation fault

It is because the object in the testfile points back to the root group. When the tool tries to traverse the object, it goes back to the root group and then back again. The fix is from Dana.  Because the fix changes the behavior of how objects are considered "visited" and "displayed", a fix to one of the expected h5dump output is followed.

### DIFF
--- a/src/H5Gint.c
+++ b/src/H5Gint.c
@@ -977,15 +977,13 @@ H5G__visit_cb(const H5O_link_t *lnk, void *_udata)
         /* Check if we've seen the object the link references before */
         if (NULL == H5SL_search(udata->visited, &obj_pos)) {
             H5O_type_t otype; /* Basic object type (group, dataset, etc.) */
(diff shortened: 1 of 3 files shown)


fortnern:
So the idea is the ref count on disk could say 1 even if it has more than one link pointing to it, correct? So we make H5G_visit always insert it into the skip list and always check for visited objects, and since it does that h5dump never needs to make the equivalent check, right? Sounds reasonable, though it could hurt performance on files with lots of objects. We could implement a hash table if this becomes an issue.

fortnern (approved):

byrnHDF (approved):

lrknox (approved):

lrknox (approved):

PR #3433, opened 2023-08-25
Add Intel oneAPI badges

derobins:


derobins:
These actions are currently failing (and we should fix that). It'd also be nice to combine the two actions into a single action.

hyoklee (approved):

PR #3435, opened 2023-08-26
Fix a memory leak when parsing corrupt files

derobins:
When parsing a file with a corrupt attribute object header message,
allocated memory would not be cleaned up. This could also trip
asserts in the memory sanity check code when that feature is enabled.

This fix ensures that memory allocated in the function is freed on
errors.

Fixes HDFFV-10774


derobins:
Allows closing HDFFV-10890, too

derobins:
Argh. The cleanup code differs between 1.10 and develop. Will re-create when I've done the surgery.

PR #3440, opened 2023-08-28
10 CVE fixes

derobins:


byrnHDF (approved):

src/H5Dchunk.c
```diff
@@ -700,9 +700,12 @@ H5D__chunk_set_info_real(H5O_layout_chunk_t *layout, unsigned ndims, const hsize
 
     /* Sanity checks */
     HDassert(layout);
-    HDassert(ndims > 0);
     HDassert(curr_dims);
 
+    /* Can happen when corrupt files are parsed */
+    if (ndims == 0)
```
fortnern, line 706: So chunked datasets cannot have a scalar dataspace? Is this something we explicitly forbade before? I see an error message in H5D__chunk_construct() for this case but the message isn't very useful ("no chunk information set?"). Assuming this isn't supported, we should document the restriction and change these two error messages to state that chunked datasets cannot have scalar dataspaces.

src/H5Fsuper_cache.c
```diff
@@ -441,85 +441,113 @@ H5F__cache_superblock_deserialize(const void *_image, size_t H5_ATTR_NDEBUG_UNUS
...
+        if (H5_IS_BUFFER_OVERFLOW(image, H5F_sizeof_addr(udata->f) * 4, end))
+            HGOTO_ERROR(H5E_FILE, H5E_OVERFLOW, NULL, "image pointer is out of bounds");
         H5F_addr_decode(udata->f, (const uint8_t **)&image, &sblock->base_addr /*out*/);
         H5F_addr_decode(udata->f, (const uint8_t **)&image, &sblock->ext_addr /*out*/);
         H5F_addr_decode(udata->f, (const uint8_t **)&image, &udata->stored_eof /*out*/);
         H5F_addr_decode(udata->f, (const uint8_t **)&image, &sblock->driver_addr /*out*/);
 
         /* Allocate space for the root group symbol table entry */
-        HDassert(!sblock->root_ent);
+        if (sblock->root_ent)
```
fortnern, line 542: I think this should remain as an assertion, I don't see how it could be triggered by anything external

src/H5Gint.c
```diff
@@ -940,15 +940,13 @@ H5G_visit_cb(const H5O_link_t *lnk, void *_udata)
         /* Check if we've seen the object the link references before */
         if (NULL == H5SL_search(udata->visited, &obj_pos)) {
             H5O_type_t otype; /* Basic object type (group, dataset, etc.) */
-            unsigned   rc;    /* Reference count of object    */
 
             /* Get the object's reference count and type */
```
fortnern, line 944: Change comment here since it's no longer getting the reference count

src/H5HLcache.c
```diff
@@ -555,11 +575,13 @@ H5HL__cache_prefix_serialize(const H5_ATTR_NDEBUG_UNUSED H5F_t *f, void *_image,
     heap = prfx->heap;
     HDassert(heap);
 
+#ifndef NDEBUG
     /* Compute the buffer size */
```
fortnern, line 579: Should we encose this in a {} block due to the local variable?

fortnern:
Didn't realize this was already merged, I think it was an old tab in my browser. My comments still stand.

derobins:
> Didn't realize this was already merged, I think it was an old tab in my browser. My comments still stand.

These are CVE fixes from develop. I didn't write new code.

PR #3449, opened 2023-08-30
Disable SZIP for Intel oneAPI Action

### LINKED ISSUE #3441: Intel CI failing
The two intel CI actions are failing because of szip - the actions should disable the filters since we are not testing the filters - the focus should be on the compiler and the library.

hyoklee:
This will make szip-related tests pass.
However, parallel and Fortran tests will still fail, which need further investigation in the future.

Close #3441.

### DIFF
--- a/.github/workflows/linux-icx-auto.yml
+++ b/.github/workflows/linux-icx-auto.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Install Dependencies
         run:  |
           sudo apt-get update
-          sudo apt-get install autoconf automake libtool libtool-bin
+          sudo apt-get install autoconf automake libtool libtool-bin libaec-dev
       - name: Add oneAPI to apt
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


byrnHDF:
How about a little divide and conquer strategy; Remove the install mpi line (in case there is an interaction with fortran)
"sudo apt install -y intel-oneapi-mpi-devel"

byrnHDF:
A couple of other items - can you merge the autotools jobs into here as it is triggered at the same time. 
Then add a job to do the parallel compile.
These can be another PR.

hyoklee:
> How about a little divide and conquer strategy; Remove the install mpi line (in case there is an interaction with fortran) "sudo apt install -y intel-oneapi-mpi-devel"

I think it's not necessary since this action worked fine until last month:

https://github.com/hdfeos/hdf5/actions/workflows/linux-icx.yml

I don't know what GitHub or HDF5 code change triggered Intel oneAPI failure.

byrnHDF:
I think we need to develop some guidelines for GH actions. Testing is as much Art as Science. One has to have a solid foundation (knowing the current state) to have predictability with the tests. I have found that breaking testing into limited and useful functionality allows you to limit the unintended interference of unknown interactions. One can always build up easily. With the concept of callable workflows in GH actions we should build our actions into a triangle, where combinations and more complicated builds are built upon more focused simpler actions at the bottom.

So here, I'm thinking we simplify the basic actions and then trigger ones that combine actions after the basic ones finish successfully. For example, maybe limit the basic action to just compile and minimal tests. Next trigger another with more tests and tools test. Finish up with binaries and examples. Etc.

derobins:
I agree with much of the sentiment expressed here, but I also think it's out of scope for this PR. Let's get this fixed ASAP so we have some basic checks on the oneAPI and we can look into expanding / simplifying it later. Ditto with the guidelines - 100% agree, but let's focus on making this critical GH action work and deal with other aspects in the next few weeks.

derobins (approved):

.github/workflows/linux-icx-auto.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Install Dependencies
         run:  |
           sudo apt-get update
-          sudo apt-get install autoconf automake libtool libtool-bin
+          sudo apt-get install autoconf automake libtool libtool-bin libaec-dev
```
lrknox, line 19: Just curious - why install libaec-dev while disabling SZIP?

lrknox (approved):

PR #3450, opened 2023-08-30
Update VOL CMake for REST VOL

mattjala:
CMakeVOL.cmake now sets some additional variables that would usually be set by `find_package(HDF5)`,  which the REST VOL's build process uses.

Also populated both docs fields in calls to `define_property`, because versions of CMake before 3.23 throw an error if not given both.

### DIFF
--- a/CMakeVOL.cmake
+++ b/CMakeVOL.cmake
@@ -34,30 +34,16 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
     message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' to allow building of external HDF5 VOL connectors")
   endif ()
 
-  set (HDF5_LIB_TARGETS "")
-  set (HDF5_HL_LIB_TARGETS "")
-
-  if (BUILD_STATIC_LIBS)
-    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIB_TARGET})
-    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIB_TARGET})
-  endif()
-
-  if (BUILD_SHARED_LIBS)
-    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIBSH_TARGET})
-    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIBSH_TARGET})
-  endif()
-
   # For compatibility, set some variables that projects would
   # typically look for after calling find_package(HDF5)
   set (HDF5_FOUND 1)
-  set (HDF5_LIBRARIES "${HDF5_LIB_TARGETS};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
+  set (HDF5_LIBRARIES "${HDF5_LIBSH_TARGET};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
   set (HDF5_INCLUDE_DIRS "${HDF5_SRC_INCLUDE_DIRS};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>")
-  set (HDF5_DIR "${HDF5_SOURCE_DIR}")
 
-  set (HDF5_C_LIBRARIES "${HDF5_LIB_TARGETS}")
-
+  set (HDF5_C_LIBRARIES "${HDF5_LIBRARIES}")
+  
   if (HDF5_BUILD_HL_LIB)
-    set (HDF5_C_HL_LIBRARIES "${HDF5_HL_LIB_TARGETS}")
+    set (HDF5_C_HL_LIBRARIES "${HDF5_HL_LIBSH_TARGET}")
   endif()
 
   set (HDF5_MAX_EXTERNAL_VOLS 10)
--- a/CMakeVOL.cmake
+++ b/CMakeVOL.cmake
@@ -34,11 +34,31 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
     message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' to allow building of external HDF5 VOL connectors")
   endif ()
 
+  set (HDF5_LIB_TARGETS "")
+  set (HDF5_HL_LIB_TARGETS "")
+
+  if (BUILD_STATIC_LIBS)
+    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIB_TARGET})
+    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIB_TARGET})
+  endif()
+
+  if (BUILD_SHARED_LIBS)
+    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIBSH_TARGET})
+    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIBSH_TARGET})
+  endif()
+
   # For compatibility, set some variables that projects would
   # typically look for after calling find_package(HDF5)
   set (HDF5_FOUND 1)
-  set (HDF5_LIBRARIES "${HDF5_LIBSH_TARGET};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
+  set (HDF5_LIBRARIES "${HDF5_LIB_TARGETS};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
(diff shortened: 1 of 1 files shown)


CMakeVOL.cmake
```diff
@@ -34,11 +34,31 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
     message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' to allow building of external HDF5 VOL connectors")
   endif ()
 
+  set (HDF5_LIB_TARGETS "")
+  set (HDF5_HL_LIB_TARGETS "")
+
+  if (BUILD_STATIC_LIBS)
```
byrnHDF: If both static and shared are enabled you will have all libraries in the link!
If VOL uses dynamic load function, then you can't use static HDF5 libs, anyway.

byrnHDF (changes requested):

CMakeVOL.cmake
```diff
@@ -34,11 +34,31 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
...
+  if (BUILD_SHARED_LIBS)
+    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIBSH_TARGET})
+    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIBSH_TARGET})
+  endif()
+
   # For compatibility, set some variables that projects would
   # typically look for after calling find_package(HDF5)
   set (HDF5_FOUND 1)
-  set (HDF5_LIBRARIES "${HDF5_LIBSH_TARGET};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
+  set (HDF5_LIBRARIES "${HDF5_LIB_TARGETS};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
```
byrnHDF, line 53: Where is the HDF5_LIBRARIES variable used? Is in the vol code?

CMakeVOL.cmake
```diff
@@ -34,11 +34,31 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
...
+  endif()
+
+  if (BUILD_SHARED_LIBS)
+    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIBSH_TARGET})
+    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIBSH_TARGET})
+  endif()
+
   # For compatibility, set some variables that projects would
   # typically look for after calling find_package(HDF5)
   set (HDF5_FOUND 1)
```
byrnHDF: This is where these vars should have a prefix like the plugins.

CMakeVOL.cmake
byrnHDF: Also, make sure these variables aren't used anywhere else, maybe?

CMakeVOL.cmake
mattjala: `HDF5_FOUND`, `HDF5_LIBRARIES`, and `HDF5_INCLUDE_DIRS`, are specifically set to match the expected output of `find_package(HDF5)`, which [would set these same variables](https://cmake.org/cmake/help/latest/module/FindHDF5.html#result-variables)

CMakeVOL.cmake
mattjala: Fixed in new commit to only add one or the other set of targets. These variable names also aren't used anywhere else in the library.

CMakeVOL.cmake
```diff
@@ -34,11 +34,28 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
     message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' to allow building of external HDF5 VOL connectors")
   endif ()
 
+  if (BUILD_SHARED_LIBS)
+    set(HDF5_LIB_TARGETS ${HDF5_LIBSH_TARGET})
+    set(HDF5_HL_LIB_TARGETS ${HDF5_HL_LIBSH_TARGET})
+  else ()
+    if (BUILD_STATIC_LIBS)
```
jhendersonHDF: As is, we shouldn't ever be trying to use static libraries when building VOL connectors, so I don't think we want these changes

CMakeVOL.cmake
jhendersonHDF: > This is where these vars should have a prefix like the plugins.

This code is specifically removing the `find_package(HDF5)` call in an external connector's CMake code and then setting the CMake variables that that call would have set if it had been made. Basically replacing the `find_package` functionality since HDF5 will get built at the time the connector does.

byrnHDF: Right. In plugins and others I do the same thing except with a prefix - of course it requires some control over the other package or we do this with zlib/szip - overwrite the CMakeLists.txt file that gets imported from the repo. See the config/cmake dirs LIBEAC and SZIP.

byrnHDF (approved):

CMakeVOL.cmake
jhendersonHDF: Here using a prefix would break the builds since we don't have any control over the external packages really and can't reliably predict what their CMake code would look like in order to try overwriting things to use a prefix. `find_package(HDF5 ...)` is pretty standardized, so we can get away with replacing that, but everything else is up in the air.

CMakeVOL.cmake
```diff
@@ -39,6 +39,13 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
   set (HDF5_FOUND 1)
   set (HDF5_LIBRARIES "${HDF5_LIBSH_TARGET};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
   set (HDF5_INCLUDE_DIRS "${HDF5_SRC_INCLUDE_DIRS};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>")
+  set (HDF5_DIR "${HDF5_SOURCE_DIR}")
+
+  set (HDF5_C_LIBRARIES "${HDF5_LIB_TARGETS}")
```
jhendersonHDF: Hmm, shouldn't this essentially be the same as `HDF5_LIBRARIES`?

mattjala: Fixed - this was initially to prevent spurious dependency creation in the REST VOL CMake, but I fixed that on the REST VOL's side.

CMakeVOL.cmake
```diff
@@ -39,6 +39,13 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
   set (HDF5_FOUND 1)
   set (HDF5_LIBRARIES "${HDF5_LIBSH_TARGET};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
   set (HDF5_INCLUDE_DIRS "${HDF5_SRC_INCLUDE_DIRS};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>")
+  set (HDF5_DIR "${HDF5_SOURCE_DIR}")
```
jhendersonHDF: I believe `HDF5_DIR` is supposed to point to an installed HDF5 directory location, so setting it to the source dir for HDF5 might cause issues. Might want to set it to something like `${CMAKE_INSTALL_PREFIX}`

byrnHDF: Do the VOLs actually use this HDF5_DIR variable - because I think it is only informational and shouldn't be used otherwise.

mattjala: It was used in the `vol-tests` repo/submodule, which is now deprecated. I replaced it in the REST VOL's CMake with a direct reference to `HDF5_SOURCE_DIR`, and removed this line.

jhendersonHDF: `HDF5_SOURCE_DIR` comes from the HDF5 project and so isn't likely to be a variable when building the REST VOL, except for possibly in the case of the prebuilt HDF5 notion that we're looking at removing now

mattjala: If the VOL is built as a subdirectory under the HDF5 library, then I think it should inherit HDF5's CMake variables, including this one.

byrnHDF: But it doesn't exist yet, until packaged?  HDF5_DIR is NOT HDF5_SOURCE_DIR. Need to look at the CMake FindHDF5.cmake module file.

byrnHDF: At least I think so.  How does the VOL actually use that variable?

mattjala: The VOL uses `HDF5_SOURCE_DIR` to locate the headers to include when the high level library is enabled (Specifically, to add `${HDF5_SOURCE_DIR}/hl/src` as an include directory).

I might be confused - I think `HDF5_SOURCE_DIR` is automatically created when `project(HDF5)` is invoked in HDF5's top level CMakeLists.txt, and because we're reworking the VOLs to be built as a CMake subdirectory of the HDF5 project, the CMakeLists.txt for the VOLs will have access to `HDF5_SOURCE_DIR`.

lrknox (approved):

PR #3453, opened 2023-08-31
Subfiling VFD - add note about MPICH_MAX_THREAD_SAFETY env. var.

jhendersonHDF:


byrnHDF (approved):

src/H5FDsubfiling/H5FDioc.c
```diff
@@ -235,10 +235,18 @@ H5FD_ioc_init(void)
                 /* If MPI is initialized, validate that it was initialized with MPI_THREAD_MULTIPLE */
                 if (MPI_SUCCESS != (mpi_code = MPI_Query_thread(&provided)))
                     H5_SUBFILING_MPI_GOTO_ERROR(H5I_INVALID_HID, "MPI_Query_thread failed", mpi_code);
-                if (provided != MPI_THREAD_MULTIPLE)
+                if (provided != MPI_THREAD_MULTIPLE) {
+#ifdef MPICH
```
qkoziol, line 239: We've tried to avoid these sorts of "platform" specific ifdef's in the library code.  This would be better a generic error message, with a pointer into the documentation, which could explain about the MPICH case.

qkoziol (changes requested):
I don't think that including a platform-specific error that is determined at compile time is the correct way to go here.  It's certainly possible that the library could be used with another MPI implementation at runtime (within the limits of ABI compatibility, of course).

src/H5FDsubfiling/H5FDioc.c
jhendersonHDF: I haven't really seen anything in the way of consistent documentation about this environment variable and if someone is using OpenMPI, then pointing them towards it seems like it would only serve to be confusing.

jhendersonHDF:
> I don't think that including a platform-specific error that is determined at compile time is the correct way to go here. It's certainly possible that the library could be used with another MPI implementation at runtime (within the limits of ABI compatibility, of course).

If you compile HDF5 with MPICH and then try to run with OpenMPI, you almost certainly will fail to run correctly, as I've seen many times in the past. I understand the sentiment about swapping out libraries at runtime, but I'm not sure that I've ever seen a case where this works correctly.

jhendersonHDF:
I suppose I could just put "If using MPICH" in the error message, though I wanted to be a bit more elegant than that.

qkoziol:
> > I don't think that including a platform-specific error that is determined at compile time is the correct way to go here. It's certainly possible that the library could be used with another MPI implementation at runtime (within the limits of ABI compatibility, of course).
> 
> If you compile HDF5 with MPICH and then try to run with OpenMPI, you almost certainly will fail to run correctly, as I've seen many times in the past. I understand the sentiment about swapping out libraries at runtime, but I'm not sure that I've ever seen a case where this works correctly.

This will happen, in the next few years (I'm one of the co-leads for the ABI working group on the MPI Forum).

qkoziol:
> I suppose I could just put "If using MPICH" in the error message, though I wanted to be a bit more elegant than that.

The discussion of "If using MPICH..." should be in the documentation, not an error message.  :-)

jhendersonHDF:
> > I suppose I could just put "If using MPICH" in the error message, though I wanted to be a bit more elegant than that.
> 
> The discussion of "If using MPICH..." should be in the documentation, not an error message. :-)

I agree somewhat, although I think it's much more likely to be missed when it isn't something that is immediately reported by the application. There's also the question of what "documentation" refers to here since it could be the Subfiling VFD API reference, HDF5 user's guide, etc.

jhendersonHDF:
Closing this for now as I don't think there's really consensus on it. I still think that an error message is far more helpful than something being buried in documentation. There would be awkwardness if someone compiles with a particular MPI implementation and then uses another one at runtime, but https://www.open-mpi.org/doc/v4.1/man3/MPI_Get_library_version.3.php looks promising, though may not be completely reliable.

PR #3455, opened 2023-08-31
Support CMake VOL builds with FetchContent from local directory

mattjala:
This is mostly to allow for rebuilding during VOL development without needed to push to an external repo each time.

### DIFF
--- a/CMakeVOL.cmake
+++ b/CMakeVOL.cmake
@@ -34,30 +34,17 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
     message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' to allow building of external HDF5 VOL connectors")
   endif ()
 
-  set (HDF5_LIB_TARGETS "")
-  set (HDF5_HL_LIB_TARGETS "")
-
-  if (BUILD_STATIC_LIBS)
-    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIB_TARGET})
-    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIB_TARGET})
-  endif()
-
-  if (BUILD_SHARED_LIBS)
-    list(APPEND HDF5_LIB_TARGETS ${HDF5_LIBSH_TARGET})
-    list(APPEND HDF5_HL_LIB_TARGETS ${HDF5_HL_LIBSH_TARGET})
-  endif()
-
   # For compatibility, set some variables that projects would
   # typically look for after calling find_package(HDF5)
   set (HDF5_FOUND 1)
-  set (HDF5_LIBRARIES "${HDF5_LIB_TARGETS};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
+  set (HDF5_LIBRARIES "${HDF5_LIBSH_TARGET};${LINK_LIBS};${LINK_COMP_LIBS};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>")
   set (HDF5_INCLUDE_DIRS "${HDF5_SRC_INCLUDE_DIRS};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>")
   set (HDF5_DIR "${HDF5_SOURCE_DIR}")
 
-  set (HDF5_C_LIBRARIES "${HDF5_LIB_TARGETS}")
-
+  set (HDF5_C_LIBRARIES "${HDF5_LIBRARIES}")
+  
   if (HDF5_BUILD_HL_LIB)
-    set (HDF5_C_HL_LIBRARIES "${HDF5_HL_LIB_TARGETS}")
+    set (HDF5_C_HL_LIBRARIES "${HDF5_HL_LIBSH_TARGET}")
   endif()
 
   set (HDF5_MAX_EXTERNAL_VOLS 10)
--- a/CMakeVOL.cmake
+++ b/CMakeVOL.cmake
@@ -30,20 +30,16 @@ endfunction ()
 option (HDF5_VOL_ALLOW_EXTERNAL "Allow building of external HDF5 VOL connectors with FetchContent" OFF)
 mark_as_advanced (HDF5_VOL_ALLOW_EXTERNAL)
 if (HDF5_VOL_ALLOW_EXTERNAL)
-  if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "NO" OR NOT HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT")
-    message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' or 'LOCAL_SOURCE_DIR' to allow building of external HDF5 VOL connectors")
-  endif ()
-
-  if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_SOURCE_DIR" AND NOT HDF5_VOL_PATH_TEMP)
-    message(FATAL_ERROR "HDF5_VOL_PATH_TEMP must have a path to VOL source when HDF5_ALLOW_EXTERNAL_SUPPORT = LOCAL_SOURCE_DIR")
-  endif ()
+  
+  if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "NO" OR (NOT HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT" AND NOT HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_DIR"))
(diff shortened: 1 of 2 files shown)


CMakeVOL.cmake
```diff
@@ -60,33 +60,55 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
...
-    mark_as_advanced (HDF5_VOL_URL${vol_idx_fixed})
+    if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT")
+      set (HDF5_VOL_URL${vol_idx_fixed} "" CACHE STRING "Git repository URL of an external HDF5 VOL connector to build")
+      mark_as_advanced (HDF5_VOL_URL${vol_idx_fixed})
+      set (HDF5_VOL_SOURCE "${HDF5_VOL_URL${vol_idx_fixed}}")
+    elseif(HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_DIR")
+      set (HDF5_VOL_PATH${vol_idx_fixed} "" CACHE STRING "Path to the source directory of an external HDF5 VOL connector to build")
+      mark_as_advanced (HDF5_VOL_PATH${vol_idx_fixed})
+      set (HDF5_VOL_SOURCE "${HDF5_VOL_PATH${vol_idx_fixed}}")
+    else()
```
jhendersonHDF: extra else here

CMakeVOL.cmake
```diff
@@ -60,33 +60,55 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
...
       # Extract the name of the VOL connector
-      string (FIND "${HDF5_VOL_URL${vol_idx_fixed}}" "/" hdf5_vol_name_pos REVERSE)
+      string (FIND "${HDF5_VOL_SOURCE}" "/" hdf5_vol_name_pos REVERSE)
       if (hdf5_vol_name_pos EQUAL -1)
-        message (SEND_ERROR "Invalid URL '${HDF5_VOL_URL${vol_idx_fixed}}' specified for HDF5_VOL_URL${vol_idx_fixed}")
+        if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT")
+          message (SEND_ERROR "Invalid URL '${HDF5_VOL_SOURCE}' specified for HDF5_VOL_URL${vol_idx_fixed}")
+        elseif (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_DIR")
+          message (SEND_ERROR "Invalid source path '${HDF5_VOL_SOURCE}' specified for HDF5_VOL_PATH${vol_idx_fixed}")
+        else()
```
jhendersonHDF: extra else

CMakeVOL.cmake
```diff
@@ -95,21 +117,38 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
...
+        FetchContent_Declare (HDF5_VOL_${hdf5_vol_name_lower}
+        GIT_REPOSITORY "${HDF5_VOL_SOURCE}"
         GIT_TAG "${HDF5_VOL_${hdf5_vol_name_upper}_BRANCH}"
-      )
+        )
+      elseif(HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_DIR")
+        FetchContent_Declare (HDF5_VOL_${hdf5_vol_name_lower}
+        SOURCE_DIR "${HDF5_VOL_SOURCE}"
+        )
+      else()
```
jhendersonHDF: extra else

CMakeVOL.cmake
```diff
@@ -95,21 +117,38 @@ if (HDF5_VOL_ALLOW_EXTERNAL)
...
       if (NOT hdf5_vol_${hdf5_vol_name_lower}_POPULATED)
         FetchContent_Populate(HDF5_VOL_${hdf5_vol_name_lower})
 
         if (NOT EXISTS "${hdf5_vol_${hdf5_vol_name_lower}_SOURCE_DIR}/CMakeLists.txt")
-          message (SEND_ERROR "The git repository branch '${HDF5_VOL_${hdf5_vol_name_upper}_BRANCH}' for VOL connector '${hdf5_vol_name}' does not appear to contain a CMakeLists.txt file")
+          if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT")
+            message (SEND_ERROR "The git repository branch '${HDF5_VOL_${hdf5_vol_name_upper}_BRANCH}' for VOL connector '${hdf5_vol_name}' does not appear to contain a CMakeLists.txt file")
+          elseif (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_DIR")
+            message(SEND_ERROR "The local directory '${HDF5_VOL_SOURCE}' for VOL connector '${hdf5_vol_name}' does not appear to contain a CMakeLists.txt file")
+          else()
```
jhendersonHDF: extra else

jhendersonHDF (approved):
Looks good to me, though I'll let @byrnHDF  comment on the CMake part

CMakeVOL.cmake
```diff
@@ -30,9 +30,9 @@ endfunction ()
 option (HDF5_VOL_ALLOW_EXTERNAL "Allow building of external HDF5 VOL connectors with FetchContent" OFF)
 mark_as_advanced (HDF5_VOL_ALLOW_EXTERNAL)
 if (HDF5_VOL_ALLOW_EXTERNAL)
-  if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "NO" OR NOT HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT")
-    message (FATAL_ERROR "HDF5_ALLOW_EXTERNAL_SUPPORT must be set to 'GIT' to allow building of external HDF5 VOL connectors")
-  endif ()
+  if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "NO" OR (NOT HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT" AND NOT HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "LOCAL_DIR"))
```
byrnHDF: LOCAL_DIR is not currently a valid choice - just TGZ

doc/cmake-vols-fetchcontent.md
```diff
@@ -49,26 +52,44 @@ For each URL specified, HDF5's CMake code will attempt to use CMake's
 [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html)
 functionality to retrieve the source code for a VOL connector pointed to by
 that URL and will try to build that VOL connector as part of the HDF5 library
-build process. The VOL connector must be able to be built by CMake and currently
+build process. 
+
+If `HDF5_ALLOW_EXTERNAL_SUPPORT` is set to "LOCAL_DIR", then the CMake cache
```
byrnHDF: Please see the existing use of HDF5_ALLOW_EXTERNAL_SUPPORT; these changes are overwriting current usage reqs.

CMakeVOL.cmake
mattjala: Would it be reasonable to separate the external support source used for VOLs and the external suppport source used elsewhere in the library? There could be a `HDF5_VOL_EXTERNAL_SUPPORT` for VOL sources, and `HDF5_ALLOW_EXTERNAL_SUPPORT` could be unchanged. This would avoid the current weirdness where wanting to get a VOL from github means you also need to get ZLIB and SZIP from github.

jhendersonHDF: For maintenance reasons, I didn't want to create a separate path just for support for VOL connectors, since external VFD plugins are another thing we'll have to think about at some point. There shouldn't be any need to grab ZLIB and SZIP from github just by turning on `HDF5_ALLOW_EXTERNAL_SUPPORT` though. I think that's just because the options for building SZIP and ZLIB are on by default.

byrnHDF: HDF5_ALLOW_EXTERNAL_SUPPORT is just three states NO: go find it, GIT: use a repo and TGZ: use a file.
Because at that time it was all do the same thing, compression, plugins, etc.
There is a supervariable per item: XXX_USE_EXTERNAL which determines whether to do a find_package or use the HDF5_ALLOW_EXTERNAL_SUPPORT option.  See the CMakeFilters.cmake file.
For the vol you could create a VOL_USE_EXTERNAL or even a new variable and leave the last option to follow the global HDF5_ALLOW_EXTERNAL_SUPPORT. 
Does this need refactored - yes, but maybe for version 2.0 of HDF5.

mattjala: There are a couple lines in `CMakeFilters.cmake` that tell CMake that ZLIB will come from github if "GIT" is set:
```
if (HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "GIT" OR HDF5_ALLOW_EXTERNAL_SUPPORT MATCHES "TGZ")
  set (ZLIB_USE_EXTERNAL "Use External Library Building for ZLIB" 1)
```

You can get around this by disabling ZLIB and SZIP when getting a VOL from Github, or just getting them from Github as well, but it's an unintuitive thing to encounter.

mattjala:
I moved the `LOCAL_DIR` option under the `HDF5_VOL_ALLOW_EXTERNAL` variable to prevent conflict with `HDF5_ALLOW_EXTERNAL_SUPPORT`. This also avoids the crossover between external VOL support and external filter support, and makes the two variables work similarly.

byrnHDF (approved):

PR #3456, opened 2023-08-31
Fix Subfiling VFD IOC assignment bug

jhendersonHDF:


src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -1614,6 +1614,10 @@ identify_ioc_ranks(sf_topology_t *app_topology, int rank_stride)
     assert(app_topology->app_layout->layout);
     assert(app_topology->app_layout->node_count > 0);
```
brtnfld, line 1615: Should we add an assert for the maxval of node_count? It was out-of-range in the issue.

jhendersonHDF: Is there a reasonable value we could use as a max for node count?

brtnfld: I don't think it can be greater than the number of ranks.

jhendersonHDF: Added

brtnfld (approved):
thanks!

derobins (approved):

PR #3491, opened 2023-09-02
Convert hbool_t --> bool in tools

derobins:


derobins:
Slowly purging hbool_t/TRUE/FALSE from the library...

I tried to be careful to not change output, which I left as "TRUE" and "FALSE" where we print property values, but @byrnHDF should confirm.

byrnHDF (approved):

lrknox (approved):

jhendersonHDF (approved):

lrknox (approved):

PR #3492, opened 2023-09-03
Convert hbool_t --> bool in examples

derobins:


byrnHDF (approved):
Looks like I will need to make these changes in the Examples repo too.

lrknox (approved):

PR #3494, opened 2023-09-03
Convert hbool_t --> bool in test

derobins:


byrnHDF (approved):

test/API/tselect.c
```diff
@@ -3189,7 +3189,7 @@ test_select_hyper_checker_board_dr__run_test(int test_num, const uint16_t *cube_
 
                     data_ok = test_select_hyper_checker_board_dr__verify_data(small_cube_buf_1, small_rank,
                                                                               edge_size, checker_edge_size,
-                                                                              expected_value, (hbool_t)TRUE);
+                                                                              expected_value, (bool)true);
```
jhendersonHDF, line 3192: Cast can probably be removed here. Not sure that any compiler cares though. Looks like there are several bool casts in this file after the changes

jhendersonHDF (approved):

test/API/tselect.c
derobins: Yeah, I thought about doing some additional cleanup work, but figured I'd leave that for a second pass later.

PR #3501, opened 2023-09-03
Add all float types to test

byrnHDF:
the 128 bit float test needs fixed

### DIFF
--- a/test/dsets.c
+++ b/test/dsets.c
@@ -5712,9 +5712,9 @@ test_floattypes(hid_t file)
         /* orig_data[] are initialized to be within the range that can be represented by
          * dataset datatype (no precision loss during datatype conversion)
          */
-        float  orig_data[2][5] = {{188384.0F, 19.103516F, -1.0831790e9F, -84.242188F, 5.2045898F},
+        float orig_data[2][5] = {{188384.0F, 19.103516F, -1.0831790e9F, -84.242188F, 5.2045898F},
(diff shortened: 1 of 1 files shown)


test/dsets.c
```diff
@@ -5683,6 +5683,306 @@ test_multiopen(hid_t file)
...
+         *----------------------------------------------------------------------
+         */
+        if (H5Tclose(datatype) < 0)
+            goto error;
+        if (H5Sclose(space) < 0)
+            goto error;
+        if (H5Dclose(dataset) < 0)
+            goto error;
+    }
+#if 0
```
derobins, line 5882: Why the `#if 0` ?

derobins: And does this pass on Power? We have frequent long double issues there.

test/dsets.c
```diff
@@ -5683,6 +5683,306 @@ test_multiopen(hid_t file)
...
+#if H5_SIZEOF_LONG_DOUBLE != H5_SIZEOF_DOUBLE
```
derobins, line 5883: I feel like we should run this even when long double and double are the same size.

byrnHDF: #if 0 because this is the test that needs fixed - the orig_data values need to be appropriate for long double (they are just copies of double from above).

byrnHDF:
Close in favor of #4060

PR #3505, opened 2023-09-05
Change type of `offset` arg in `H5Pset_external` to `HDoff_t`

phil-opp:
## Describe your changes

The `off_t` type is only 32-bit on Windows, which makes it impossible to link to higher offsets in large files.

The `H5O_efl_entry_t` struct defines its `offset` field already as `HDoff_t`, so no additional conversion is needed.,

https://github.com/HDFGroup/hdf5/issues/3506

- [ ] My code conforms to the guidelines in CONTRIBUTING.md
- [ ] I made an entry in release_docs/RELEASE.txt (bug fixes, new features)


mkitti:
I think we might need to version the function in order to change the API like this.

derobins (approved):

derobins:
Give us a little more time with this. We're thinking about ditching off_t entirely and moving to an unsigned integer.

phil-opp:
Sure, sounds good!

lrknox (approved):

derobins:
Update: I'm starting to go through the library to make sure HDoff_t is used throughout. I think there's a few other internal changes we should also make. We have a lot of projects in motion right now, but I'll try to get the underlying work done in October so we can make this change.

derobins:
Update: The internal changes should be complete now. We're discussing how versioning is going to work for this.

derobins:
This can probably go in at this point, but we still use `off_t` in places like the tools and tests. Those will have to get fixed in the near future.

derobins:
FWIW, this API call will NOT be versioned. If we didn't version API calls that use `hid_t` when we updated that from `int` to `int64`, we don't need to version this one.

jhendersonHDF (approved):
For going into develop this looks good. We'll likely want to think about how to make it so we don't accidentally sync this change back to the 1.14 branch.

PR #3509, opened 2023-09-06
Add NVHPC 23.7 GitHub Actions

hyoklee:
All tests pass with Fortran and parallel enabled for both Autotools and CMake.
Making NVHPC work with HDF5 is quite tricky.
These actions will help AI users who want to utilize NVDIA chips.

### DIFF
--- a/.github/workflows/linux-nvhpc-auto.yml
+++ b/.github/workflows/linux-nvhpc-auto.yml
@@ -1,9 +1,26 @@
 name: linux autotools nvhpc
 
 on:
+  workflow_dispatch:
   push:
-    branches:
-      - develop
+  pull_request:
+    branches: [ develop ]
+    paths-ignore:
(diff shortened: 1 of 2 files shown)


byrnHDF:
General comment about action yaml files - we need to always create them as callable workflows. And then we can have one yaml file that can be updated with new or different options.

lrknox:
Is this for nvhpc 27.3 or for 23.7?

.github/workflows/linux-nvhpc-auto.yml
```diff
@@ -0,0 +1,41 @@
+name: linux autotools nvhpc
+
+on:
+  push:
```
derobins, line 5: Should also do pull request for both?

derobins (approved):

lrknox (approved):

PR #3529, opened 2023-09-11
Convert main.yml CI into callable workflows

byrnHDF:
This is just a first pass breaking it into logical groups. Suggestions welcome on better flow and/or creating dependent workflows. (Like creating a set of tests workflows that depend on a build workflows using action/cache)


hyoklee:
`checkout@v3` -> `checkout@v4`

bmribler (approved):

lrknox (approved):

PR #3537, opened 2023-09-13
removed C_INT32_T from Fortran APIs

brtnfld:
* replaces the c_int32_t arguments in the Fortran APIs as it should not be needed, and the constants in H5_f.c are converting the c_int32_t arguments to ints.

* changed C types to uint32_t when the API calls for it.


epourmal (approved):
I agree with the change.

lrknox (approved):

PR #3540, opened 2023-09-15
Strip HD prefix from string/char C API calls

derobins:
* HD(f)(put|get)(s|c)
* HDstr*
* HDv*printf
* HD(s)(print|scan)f
* HDperror

But NOT:

* HDstrcase*
* HDvasprintf
* HDstrtok_r
* HDstrndup

As those are not C99 and have portability work-around implementations. They will be handled later.

### DIFF
--- a/c++/test/titerate.cpp
+++ b/c++/test/titerate.cpp
@@ -78,7 +78,7 @@ static void printelems(const Group &group, const H5std_string &dsname, const H5s
 static int
 iter_strcmp(const void *s1, const void *s2)
 {
-    return (HDstrcmp(*reinterpret_cast<const char *const *>(s1), *reinterpret_cast<const char *const *>(s2)));
+    return (strcmp(*reinterpret_cast<const char *const *>(s1), *reinterpret_cast<const char *const *>(s2)));
 }
 
 /*-------------------------------------------------------------------------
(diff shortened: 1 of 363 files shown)


derobins:
Looks like I messed up strcasestr for Windows. Will fix.

c++/test/titerate.cpp
```diff
@@ -162,18 +162,18 @@ test_iter_group(FileAccPropList &fapl)
...
-            check_values(lnames[i], "HDstrdup returns NULL", __LINE__, __FILE__);
+            lnames[i] = strdup(name);
+            check_values(lnames[i], "strdup returns NULL", __LINE__, __FILE__);
         }
 
         /* Create a group and named datatype under root group for testing */
         Group grp(file.createGroup(GROUP1, 0));
-        lnames[NDATASETS] = HDstrdup("grp");
-        check_values(lnames[NDATASETS], "HDstrdup returns NULL", __LINE__, __FILE__);
+        lnames[NDATASETS] = strdup("grp");
```
jhendersonHDF, line 171: strdup is only POSIX right? Do we want to be including that here?

derobins: I thought about it, but all C compilers I know of support it, so I'm letting it slide. I can switch it back if people feel strongly about it.

derobins: Though eventually it won't matter, we'll assume a POSIX environment and have quiet work-arounds for Windows (i.e., no HDfoo - just foo). We'll only use an H5_ prefix for calls that are quirky, like basename/dirname and getopt.

jhendersonHDF: Right, I'm not sure if there are any platforms where it would be a problem anymore or not. Definitely still issues for `strndup`, but not sure on `strdup`

jhendersonHDF (approved):

lrknox (approved):

PR #3547, opened 2023-09-19
Revert the removal of java class H5GroupInfo

byrnHDF:


lrknox (approved):

java/src/hdf/hdf5lib/HDF5GroupInfo.java
```diff
@@ -0,0 +1,188 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * Copyright by The HDF Group.                                               *
+ * Copyright by the Board of Trustees of the University of Illinois.         *
```
jhendersonHDF, line 3: IIRC, I think this was a newly-created file that may fall under the newer copyright, but I could be wrong on the timeline

byrnHDF: Fairly certain it was there for awhile - just we discontinued its usage? Maybe? Anyway it is the original file for a few releases.

jhendersonHDF (approved):

java/src/hdf/hdf5lib/HDF5GroupInfo.java
byrnHDF: It was there in 1.10.0

PR #3555, opened 2023-09-20
Hdf5 1 10 11

lrknox:
Initial updates to release branch for HDF5 1.10.11 release.


release_docs/RELEASE.txt
```diff
@@ -1,4 +1,4 @@
-HDF5 version 1.10.11-1 currently under development
+HDF5 version 1.10.11-2 currently under development
```
byrnHDF, line 1: So this stays under dev until you generate the tarballs?

byrnHDF (approved):

release_docs/RELEASE.txt
lrknox: Yes.  Running bin/release replaces "currently under development" with "released on <machine's current date>" in RELEASE.txt and README.md.

PR #3556, opened 2023-09-20
Check return values from HDF5 API calls.

hyoklee:
Remove unnecessary test line as well.

### DIFF
--- a/testpar/t_init_term.c
+++ b/testpar/t_init_term.c
@@ -39,9 +39,7 @@ main(int argc, char **argv)
 
     /* test if we can initialize the library with MPI being finalized
        and create a file serially */
-    if (H5open() < 0) {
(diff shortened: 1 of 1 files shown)


testpar/t_init_term.c
```diff
@@ -37,11 +37,11 @@ main(int argc, char **argv)
 
     MPI_Finalize();
 
-    nerrors += GetTestNumErrs();
-
     /* test if we can initialize the library with MPI being finalized
        and create a file serially */
-    H5open();
+    if (H5open() < 0) {
```
jhendersonHDF: We should use the VRFY macro for these checks since it's consistent and will increment nerrors on failure

jhendersonHDF (approved):

bmribler (approved):

PR #3561, opened 2023-09-20
Move tools directory into main repo

glennsong09:


byrnHDF:
Is it expected that the testfiles will be different here then in the tools/testfiles? If not, just copy the files from tools to the correct directories before tests are run. We could even sort the tools/testfiles folder into separate tool folders.

jhendersonHDF:
> Is it expected that the testfiles will be different here then in the tools/testfiles? If not, just copy the files from tools to the correct directories before tests are run. We could even sort the tools/testfiles folder into separate tool folders.

That's the plan at least. This is just initial work on a feature branch towards that so it's easy to compare testfiles and see what's different and the same.

jhendersonHDF:
After some discussion, we're going to do some cleanup of the tools test files directories in HDF5 first and then revisit this PR once we've had a chance to clean up a bit beforehand

Issue #3571, opened 2023-09-21
hdf5 fails to build with Xcode 15 / macOS Sonoma


fxcoudert:
**Describe the bug**

Compilation with Xcode 15 (on macOS Sonoma) leads to a build error:

```
  ld: unknown options: -commons 
  clang: error: linker command failed with exit code 1 (use -v to see invocation)
  make[1]: *** [H5detect] Error 1
```

Apple ships with Xcode 15 and macOS Sonoma a new linker, which does not accept the `-commons` option.

This commons from hdf5 configure at https://github.com/HDFGroup/hdf5/blob/95907ba548cd6d40eb625f7c79d9bc437df93c68/configure.ac#L1082

```
  ## tell libtool to do the right thing with COMMON symbols, this fixes
  ## corrupt values with COMMON and EQUIVALENCE when building shared
  ## Fortran libraries on OSX with gnu and Intel compilers (HDFFV-2772).
  case "`uname`" in
    Darwin*)
    H5_LDFLAGS="$H5_LDFLAGS -Wl,-commons,use_dylibs"
    ;;
  esac
```

**Expected behavior**
Successful install :)

**Platform (please complete the following information)**
 - HDF5 version: 1.14.2
 - OS and version: macOS 14.0 (Release Candidate)
 - Compiler and version: Xcode 15
 - Build system (e.g. CMake, Autotools) and version: autoconf 2.71, automake 1.66.5, libtool 2.4.7
 - Any configure options you specified: `./configure --disable-dependency-tracking --disable-silent-rules --enable-build-mode=production --enable-fortran --enable-cxx --prefix=/opt/homebrew/Cellar/hdf5/1.14.2 --with-szlib=/opt/homebrew/opt/libaec`
 - MPI library and version (parallel HDF5): MPI is not used

**Additional context**

fxcoudert:
I've filed the bug with Apple as FB13194355

Dave-Allured:
@fxcoudert, thank you for the report.  Macports users are now running into this, unknown `-commons`, after updating to Xcode 15 and/or Command Line Tools 15, since 15.0 release three days ago (2023 September 18).
https://trac.macports.org/ticket/68194

Dave-Allured:
Note:  `-commons` is a fortran-related option.  A Macports user has reported that HDF5 1.14.2 without the fortran API option builds fine with Xcode 15.0.

Dave-Allured:
[...]

PR #3581, opened 2023-09-22
Removed the use of -commons linking option on Darwin

brtnfld:
Removed the use of -commons linking option on Darwin as COMMON and EQUIVALENCE is no long used in the Fortran source. Addresses #3571


byrnHDF (approved):

derobins (approved):

derobins:
There's a bunch of other comments that reference Fortran linker issues in configure.ac, some of which are quite old (HDFFV-2772, for example). We might want to check to see how much of the Fortran hack code is still necessary before releasing 1.14.3. It'd be nice if we could remove the work-arounds.

derobins:
Should have a release note!

derobins (approved):

Dave-Allured:
Can you also fit this linker fix into the upcoming 1.10.11?  Thanks.

brtnfld:
We managed to squeeze the fix into 1.10.11.

PR #3582, opened 2023-09-23
Avoid  `unable to modify source` output message.

hyoklee:
I think the `|| die` message should appear only when there's a system error (e.g., disk full).
I think it's better to remove an old header before it's being (re)generated from a text file.


derobins:
I feel like the purpose of creating a backup of the original file is to check it for differences with what you just created. That's a pretty standard way of doing things in POSIX/Autotools-land, so I'm not sure this PR is moving the needle.

PR #3583, opened 2023-09-23
Remove NVHPC [set_but_not_used] warning.

hyoklee:
This PR will remove the following NVHPC warning:

```
warning: variable "env_value" was set but never used [set_but_not_used]
```

### DIFF
--- a/testpar/t_subfiling_vfd.c
+++ b/testpar/t_subfiling_vfd.c
@@ -2347,7 +2347,8 @@ main(int argc, char **argv)
         must_unset_ioc_thread_count_env = true;
     }
 
-    if (!(env_value = HDgetenv(H5FD_SUBFILING_CONFIG_FILE_PREFIX))) {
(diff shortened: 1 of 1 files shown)


testpar/t_subfiling_vfd.c
```diff
@@ -2347,7 +2347,8 @@ main(int argc, char **argv)
         must_unset_ioc_thread_count_env = true;
     }
 
-    if (!(env_value = HDgetenv(H5FD_SUBFILING_CONFIG_FILE_PREFIX))) {
```
byrnHDF, line 2350: If env_value is not used anywhere else just remove "env_value="  part?

jhendersonHDF: Yes, that would be the correct fix here.

jhendersonHDF: It also looks like `env_value` is no longer used outside of this one spot, so the variable can be removed.

hyoklee:
See #3590

PR #3589, opened 2023-09-25
Remove h5dwalk.1 man page.

hyoklee:
HDF5 doesn't support man page.


jhendersonHDF:
Agreed on removing this if nothing is currently installing the man page on the system, but it would be a good idea for us to support man pages for the HDF5 tools.

derobins:
This is pretty minimal so re-creating it when/if we have real man page support would not be that difficult.

Agree that HDF5 supporting man pages would be nice, but let's do that en masse later.

derobins (approved):

bmribler (approved):

PR #3590, opened 2023-09-25
Remove `sh` to run bash script.

hyoklee:
This will fix `h5fuse.sh: 23: function: not found` error.


testpar/t_subfiling_vfd.c
```diff
@@ -2347,7 +2345,7 @@ main(int argc, char **argv)
         must_unset_ioc_thread_count_env = true;
     }
 
-    if (!(env_value = HDgetenv(H5FD_SUBFILING_CONFIG_FILE_PREFIX))) {
+    if (NULL != HDgetenv(H5FD_SUBFILING_CONFIG_FILE_PREFIX)) {
```
jhendersonHDF: This is the opposite logic. It should be `if (!HDgetenv(...))`

jhendersonHDF (approved):

brtnfld (approved):

testpar/t_subfiling_vfd.c
brtnfld: @derobins, I think this was merged without this needed change.

jhendersonHDF: It looks like @hyoklee got the correct change in, just in a different form.

Issue #2658, opened 2023-04-03
Unexpected independent reads with chunked datasets


roblatham00:
**Describe the bug**
The Darshan tool shows us more independent reads than expected after setting collective metadata for reads.

When we use `H5Pset_coll_metadata_write` and `H5Pset_all_coll_metadata_ops` we see the expected decrease in metadata calls, except for one case:  when the dataset has an unlimited dimension:

Here's the h5dump of the header:
```
HDF5 "H4j.hdf5" {
GROUP "/" {
   DATASET "events" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 1780331, 10 ) / ( H5S_UNLIMITED, 10 ) }
      ATTRIBUTE "events" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SIMPLE { ( 10 ) / ( 10 ) }
      }
   }
   DATASET "init" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 10 ) / ( 10 ) }
      ATTRIBUTE "properties" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SIMPLE { ( 10 ) / ( 10 ) }
      }
   }
   DATASET "particles" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 12462317, 13 ) / ( H5S_UNLIMITED, 13 ) }
      ATTRIBUTE "properties" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SIMPLE { ( 13 ) / ( 13 ) }
      }
   }
   DATASET "procInfo" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 1, 6 ) / ( 1, 6 ) }
      ATTRIBUTE "properties" {
         DATATYPE  H5T_STRING {
[...]

PR #3591, opened 2023-09-25
Update parallel compression feature to support multi-dataset I/O

jhendersonHDF:


src/H5Dio.c
```diff
@@ -107,6 +107,17 @@ H5D__read(size_t count, H5D_dset_io_info_t *dset_info)
 
     FUNC_ENTER_NOAPI(FAIL)
 
+#ifdef H5_HAVE_PARALLEL
+    /* Reset the actual io mode properties to the default values in case
+     * the DXPL (if it's non-default) was previously used in a collective
+     * I/O operation.
+     */
+    if (!H5CX_is_def_dxpl()) {
```
jhendersonHDF, line 115: Here I needed to move the setting of the optimization method higher up so if something fails in this function the returned value is sane.

src/H5Dio.c
```diff
@@ -323,7 +338,11 @@ H5D__read(size_t count, H5D_dset_io_info_t *dset_info)
         }
 
         /* MDIO-specific second phase initialization */
-        for (i = 0; i < count; i++)
+        for (i = 0; i < count; i++) {
+            /* Check for skipped I/O */
+            if (dset_info[i].skip_io)
```
jhendersonHDF, line 344: Make sure we don't call the `mdio_init` callback for datasets where we're skipping I/O, because the callbacks may try to use fields that aren't initialized for the datasets

src/H5Dio.c
```diff
@@ -429,7 +449,7 @@ H5D__read(size_t count, H5D_dset_io_info_t *dset_info)
 
 done:
     /* Shut down the I/O op information */
-    for (i = 0; i < io_op_init; i++)
+    for (i = 0; i < count; i++)
```
jhendersonHDF: Here, we need to step through the full `count` number of dataset entries and call their `io_term` callbacks, instead of just `io_op_init` number of them. Otherwise, if the first `io_op_init` entries were skipped datasets, the latter entries won't get terminated and certain things are left around in memory, such as the cached dataset `sel_pieces` array. Unintended reuse of that cached field when doing an `H5Dread_multi` followed by `H5Dwrite_multi` gave me errors like:

```
HDF5-DIAG: Error detected in HDF5 (1.15.0) MPI-process 0:
  #000: /home/jhenderson/git/hdf5/src/H5D.c line 1446 in H5Dwrite_multi(): can't synchronously write data
    major: Dataset
    minor: Write failed
  #001: /home/jhenderson/git/hdf5/src/H5D.c line 1317 in H5D__write_api_common(): can't write data
    major: Dataset
    minor: Write failed
  #002: /home/jhenderson/git/hdf5/src/H5VLcallback.c line 2282 in H5VL_dataset_write_direct(): dataset write failed
    major: Virtual Object Layer
    minor: Write failed
  #003: /home/jhenderson/git/hdf5/src/H5VLcallback.c line 2237 in H5VL__dataset_write(): dataset write failed
    major: Virtual Object Layer
    minor: Write failed
  #004: /home/jhenderson/git/hdf5/src/H5VLnative_dataset.c line 420 in H5VL__native_dataset_write(): can't write data
    major: Dataset
    minor: Write failed
  #005: /home/jhenderson/git/hdf5/src/H5Dio.c line 732 in H5D__write(): can't initialize I/O info
    major: Dataset
    minor: Unable to initialize object
  #006: /home/jhenderson/git/hdf5/src/H5Dchunk.c line 1091 in H5D__chunk_io_init(): unable to create file and memory chunk selections
    major: Dataset
    minor: Unable to initialize object
  #007: /home/jhenderson/git/hdf5/src/H5Dchunk.c line 1241 in H5D__chunk_io_init_selections(): unable to create file chunk selections
    major: Dataset
    minor: Unable to initialize object
  #008: /home/jhenderson/git/hdf5/src/H5Dchunk.c line 1944 in H5D__create_piece_file_map_hyper(): can't insert piece into skip list
    major: Dataspace
    minor: Unable to insert object
  #009: /home/jhenderson/git/hdf5/src/H5SL.c line 1036 in H5SL_insert(): can't create new skip list node
    major: Skip Lists
    minor: Unable to insert object
  #010: /home/jhenderson/git/hdf5/src/H5SL.c line 709 in H5SL__insert_common(): can't insert duplicate key
    major: Skip Lists
    minor: Unable to insert object
```

src/H5Dpkg.h
```diff
@@ -254,6 +254,7 @@ typedef struct H5D_piece_info_t {
     unsigned mspace_shared;  /* Indicate that the memory space for a chunk is shared and shouldn't be freed */
     bool     in_place_tconv; /* Whether to perform type conversion in-place */
     size_t   buf_off;        /* Buffer offset for in-place type conversion */
+    bool     filtered_dset;  /* Whether the dataset this chunk is in has filters applied */
```
jhendersonHDF, line 257: For each `piece_info` struct, this just keeps track of whether the dataset is filtered for later use. We could retrieve it from the `dset_info` field, but this helps with lots of later pointer chasing by only doing it once during init.

src/H5Dpkg.h
```diff
@@ -292,26 +293,28 @@ typedef struct H5D_io_info_t {
...
-    H5_flexible_const_ptr_t base_maddr;          /* starting mem address */
-    H5D_selection_io_mode_t use_select_io;       /* Whether to use selection I/O */
-    uint8_t                *tconv_buf;           /* Datatype conv buffer */
-    bool                    tconv_buf_allocated; /* Whether the type conversion buffer was allocated */
-    size_t                  tconv_buf_size;      /* Size of type conversion buffer */
-    uint8_t                *bkg_buf;             /* Background buffer */
-    bool                    bkg_buf_allocated;   /* Whether the background buffer was allocated */
-    size_t                  bkg_buf_size;        /* Size of background buffer */
+    size_t                  count;          /* Number of datasets in I/O request */
+    size_t                  filtered_count; /* Number of datasets with filters applied in I/O request */
```
jhendersonHDF, line 297: Added this field to keep track of the number of datasets in the I/O request that are filtered so we can separate out the filtered and unfiltered datasets during I/O.

src/H5Dpkg.h
```diff
@@ -292,26 +293,28 @@ typedef struct H5D_io_info_t {
...
+    H5D_dset_io_info_t     *dsets_info;     /* dsets info where I/O is done to/from */
+    size_t                  piece_count;    /* Number of pieces in I/O request */
+    size_t                  pieces_added;   /* Number of pieces added so far to arrays */
+    size_t                  filtered_pieces_added; /* Number of filtered pieces in I/O request */
```
jhendersonHDF, line 301: Also added this field to keep track of the number of filtered vs. unfiltered chunks

src/H5FDmpio.c
jhendersonHDF, line 1: The changes in this file fix an invalid read issue where `s_bufs` only has one entry in it

src/H5VLnative_dataset.c
jhendersonHDF, line 1: Depending on the pathway taken through the dataset I/O code, lower level code may attempt to make use of several of these fields that were previously uninitialized. It seems best to just initialize them here so at least a segfault will happen and hopefully be caught quickly rather than having unpredictable behavior.

testpar/t_filters_parallel.c
```diff
@@ -385,8 +458,388 @@ verify_space_alloc_status(hid_t dset_id, hid_t dcpl_id, num_chunks_written_t chu
                 MPI_Abort(MPI_COMM_WORLD, 1);
         }
     }
+}
+
```
jhendersonHDF, line 462: The vast majority of changes in this file were just adapting the code to use new functions for creating/open/reading/writing multiple datasets at once rather than a single dataset. The new `verify_chunk_opt_status` function below tries to verify that the library is returning the correct value for whether linked-chunk, multi-chunk or non-optimized I/O was performed.

src/H5Dmpio.c
```diff
@@ -129,14 +118,21 @@ typedef struct H5D_chunk_alloc_info_t {
...
  * index entry for the chunk
  */
 typedef struct H5D_chunk_index_info_t {
-    hsize_t  chunk_idx;
+    /*
+     * These two fields must come in this order and next to
+     * each other for proper and efficient hashing
+     */
+    hsize_t chunk_idx;
+    haddr_t dset_oloc_addr;
```
jhendersonHDF, line 156: Since chunks being worked on could now have the same chunk index when they're from different datasets, the hash tables used by parallel compression make use of the chunk index and dataset object header address as the key. See https://troydhanson.github.io/uthash/userguide.html#_compound_keys for an example of using multiple struct fields as a compound key

src/H5Dmpio.c
```diff
@@ -636,8 +633,8 @@ H5D__mpio_opt_possible(H5D_io_info_t *io_info)
         }
 
         /* Check whether these are both simple or scalar dataspaces */
-        if (!((H5S_SIMPLE == H5S_GET_EXTENT_TYPE(mem_space) ||
-               H5S_SCALAR == H5S_GET_EXTENT_TYPE(mem_space)) &&
+        if (!((H5S_SIMPLE == H5S_GET_EXTENT_TYPE(mem_space) || H5S_SCALAR == H5S_GET_EXTENT_TYPE(mem_space) ||
+               H5S_NULL == H5S_GET_EXTENT_TYPE(mem_space)) &&
```
jhendersonHDF, line 706: From what I can tell, it should be fine to also perform optimized parallel I/O when the memory space is H5S_NULL. This would be the case when no elements are selected in the file space and a NULL dataspace or H5S_BLOCK was passed for the memory space. In either case, the intention would be to not contribute directly to the I/O, but still participate in collective operations.

src/H5Dmpio.c
```diff
@@ -1173,26 +1163,40 @@ H5D__piece_io(H5D_io_info_t *io_info)
...
+        }
 
+        /* Process all the unfiltered datasets */
+        if ((io_info->filtered_count == 0) || (io_info->filtered_count < io_info->count)) {
             /* Perform unfiltered link chunk collective IO */
             if (H5D__link_piece_collective_io(io_info, mpi_rank) < 0)
-                HGOTO_ERROR(H5E_IO, H5E_CANTGET, FAIL, "couldn't finish linked chunk MPI-IO");
+                HGOTO_ERROR(H5E_IO, (H5D_IO_OP_READ == io_info->op_type ? H5E_READERROR : H5E_WRITEERROR),
+                            FAIL, "couldn't finish linked chunk MPI-IO");
         }
```
jhendersonHDF, line 1261: Until further development, it seemed the best we can do right now is to separate out the filtered and unfiltered datasets during multi-dataset I/O and work on the two groups with two separate I/O calls. I updated the multi-dataset I/O code to handle skipping filtered datasets when doing I/O on unfiltered datasets and updated the parallel compression code for the opposite case.

src/H5Dmpio.c
```diff
@@ -1173,26 +1163,40 @@ H5D__piece_io(H5D_io_info_t *io_info)
...
     }
-
-    if (!use_multi_dset) {
+    else {
         /* Loop over datasets */
         for (i = 0; i < io_info->count; i++) {
+            if (io_info->dsets_info[i].skip_io)
+                continue;
```
jhendersonHDF, line 1267: Don't try to perform I/O on datasets where we're skipping I/O, as certain fields aren't initialized and will cause crashes.

doc/parallel-compression.md
```diff
@@ -76,6 +76,27 @@ Remember that the feature requires these writes to use collective
...
+   filtered vs. unfiltered datasets, etc.) to make two individual
+   multi-dataset I/O calls, one for the filtered datasets and one
+   for the unfiltered datasets. When performing writes to the datasets,
+   this would also allow independent write access to the unfiltered
+   datasets if desired, while still performing collective writes to
+   the filtered datasets.
+
 ## Incremental file space allocation support
 
 HDF5's [file space allocation time](https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALLOC_TIME)
```
bmribler: Could you please fix this link too?https://docs.hdfgroup.org/hdf5/develop/group___d_c_p_l.html#ga85faefca58387bba409b65c470d7d851

src/H5Dmpio.c
```diff
@@ -4084,20 +4277,86 @@ H5D__mpio_collective_filtered_chunk_read(H5D_filtered_collective_io_info_t *chun
     (void)mpi_rank;
 #endif
 
-    if (chunk_list->num_chunk_infos) {
-        /* Retrieve filter settings from API context */
-        if (H5CX_get_err_detect(&err_detect) < 0)
-            HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get error detection info");
-        if (H5CX_get_filter_cb(&filter_cb) < 0)
-            HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get I/O filter callback function");
+    /* Cache some info about each dataset for later use */
```
jhendersonHDF: In a few of the functions below, this is a common pattern. Whereas the parallel compression code previously worked on a single dataset at a time, the multi-dataset case means that the code needs a quick way to find the dataset that a particular chunk belongs to. If there's only one dataset in the I/O operation, this works almost exactly as before, with the exception of an extra small allocation to facilitate the multi-dataset I/O case. If multiple datasets are involved, a hash table is used to cache relevant info about each dataset, with the dataset info pointer or dataset's object header address (depending on the context) being used as the key.

src/H5Dmpio.c
```diff
@@ -5042,29 +5621,32 @@ H5D__mpio_get_chunk_redistribute_info_types(MPI_Datatype *contig_type, bool *con
...
     block_lengths[2] = 1;
     block_lengths[3] = 1;
     block_lengths[4] = 1;
+    block_lengths[5] = 1;
     displacements[0] = offsetof(H5D_chunk_redistribute_info_t, chunk_block);
     displacements[1] = offsetof(H5D_chunk_redistribute_info_t, chunk_idx);
-    displacements[2] = offsetof(H5D_chunk_redistribute_info_t, orig_owner);
-    displacements[3] = offsetof(H5D_chunk_redistribute_info_t, new_owner);
-    displacements[4] = offsetof(H5D_chunk_redistribute_info_t, num_writers);
+    displacements[2] = offsetof(H5D_chunk_redistribute_info_t, dset_oloc_addr);
```
jhendersonHDF, line 5508: When sending information about chunks between MPI ranks, the parallel compression code now also needs to send the dataset object header address for the chunk as well so that we can distinguish between two chunks that have the same chunk index value.

jhendersonHDF:
It looks like the autotools nvhpc action passes fine, but the cmake version segfaults in OpenMPI for some reason. I'm not sure why that is at first glance.

doc/parallel-compression.md
jhendersonHDF: Should we use docs.hdfgroup.org or hdfgroup.github.io for the links?

byrnHDF: I vote for github!!

jhendersonHDF: Updated all the links in the doc except for two portal links that seem to currently only exist there.

src/H5FDmpio.c
```diff
@@ -3099,7 +3099,7 @@ H5FD__mpio_read_selection(H5FD_t *_file, H5FD_mem_t type, hid_t H5_ATTR_UNUSED d
              */
             j = 0; /* guess at the index of the smallest value of s_bufs[] */
 
-            if (s_bufs[j + 1].vp != NULL) {
+            if ((count > 1) && (s_bufs[j + 1].vp != NULL)) {
```
fortnern: Is the check for s_bufs[j + 1].vp (which should always reduce to s_bufs[1].vp) actually necessary here? If not, can we just eliminate this if statement entirely since the for loop exit condition should cover it?

src/H5FDmpio.c
```diff
@@ -3467,7 +3467,7 @@ H5FD__mpio_write_selection(H5FD_t *_file, H5FD_mem_t type, hid_t H5_ATTR_UNUSED
              */
             j = 0; /* guess at the index of the smallest value of s_bufs[] */
 
-            if (s_bufs[j + 1].cvp != NULL) {
+            if ((count > 1) && (s_bufs[j + 1].cvp != NULL)) {
```
fortnern: Same here

src/H5FDmpio.c
fortnern: @vchoi-hdfgroup

jhendersonHDF: After looking at this, yes I believe the if is unnecessary due to the for loop conditions.

src/H5Dio.c
fortnern: Maybe put this in H5D__ioinfo_init() instead?

src/H5Dmpio.c
fortnern: Is there a test for this case?

fortnern: Also shouldn't this case be added to the file space as well?

jhendersonHDF: Since `H5S_BLOCK` is currently not allowed for the file space, I only modified the memory space side of things. I imagine this should also be allowed for the case where a NULL file space is explicitly passed to `H5Dwrite`/`H5Dread`, but I don't believe we test this and the parallel compression tests are currently only testing the case of a NULL space getting created for the memory space when `H5S_BLOCK` is passed and no elements are selected in the dataspaces.

src/H5Dmpio.c
```diff
@@ -2583,20 +2633,25 @@ H5D__cmp_filtered_collective_io_info_entry(const void *filtered_collective_io_in
...
-        hsize_t chunk_idx2 = entry2->index_info.chunk_idx;
+        haddr_t oloc_addr1 = entry1->index_info.dset_oloc_addr;
+        haddr_t oloc_addr2 = entry2->index_info.dset_oloc_addr;
+
+        if (0 == (ret_value = H5_addr_cmp(oloc_addr1, oloc_addr2))) {
+            hsize_t chunk_idx1 = entry1->index_info.chunk_idx;
+            hsize_t chunk_idx2 = entry2->index_info.chunk_idx;
 
-        ret_value = (chunk_idx1 > chunk_idx2) - (chunk_idx1 < chunk_idx2);
+            ret_value = (chunk_idx1 > chunk_idx2) - (chunk_idx1 < chunk_idx2);
```
fortnern, line 2776: Couldn't two chunks from different datasets have the same index?

src/H5Dmpio.c
```diff
@@ -2927,187 +3006,258 @@ H5D__obtain_mpio_mode(H5D_io_info_t *io_info, H5D_dset_io_info_t *di, uint8_t as
...
+                 * Check if the list is not in ascending order of offset in the file
+                 * or has unallocated chunks. In either case, the list should get
+                 * sorted.
                  */
-                if (H5D__chunk_is_partial_edge_chunk(di->dset->shared->ndims,
-                                                     di->dset->shared->layout.u.chunk.dim, chunk_info->scaled,
-                                                     di->dset->shared->curr_dims))
-                    local_info_array[i].skip_filter_pline = true;
-            }
+                if (buf_idx) {
```
fortnern: Might be worth changing the condition to if (!need_sort && buf_idx). Can't be certain without more investigation though (not sure how heavyweight the block is when compiled)

src/H5Dmpio.c
jhendersonHDF: Yes, but the logic above this should sort by dataset object header address first so that two chunks with the same index from different datasets will compare as different

src/H5Dmpio.c
jhendersonHDF: I'd guess it wouldn't be heavy enough to make much of a difference, but I think making that change is still a good idea.

src/H5Dmpio.c
fortnern: Maybe add this as a code comment?

fortnern: With some edits for context

src/H5Dmpio.c
fortnern: I've tried to figure this out on my own but it's taking too much time: are these structs always calloc'ed? Otherwise I'm concerned if there's a gap in the struct it could have garbage that would affect the key value.

src/H5Dmpio.c
```diff
@@ -4615,26 +5019,69 @@ H5D__mpio_collective_filtered_chunk_reallocate(H5D_filtered_collective_io_info_t
...
-    assert(idx_info->storage->idx_type != H5D_CHUNK_IDX_NONE);
+    assert(di);
 
 #ifdef H5Dmpio_DEBUG
     H5D_MPIO_TRACE_ENTER(mpi_rank);
     H5D_MPIO_TIME_START(mpi_rank, "Reallocation of chunk file space");
 #endif
 
+    /* Cache some info about each dataset for later use */
+    for (size_t dset_idx = 0; dset_idx < num_dset_infos; dset_idx++) {
```
fortnern: Would it be possible to build this hash table once and pass it between the different routines?

src/H5Dmpio.c
```diff
@@ -4952,10 +5500,18 @@ H5D__mpio_collective_filtered_chunk_reinsert(H5D_filtered_collective_io_info_t *
...
+            same_chunk = (0 == H5_addr_cmp(coll_entry->index_info.dset_oloc_addr,
+                                           chunk_list->chunk_infos[dbg_idx].index_info.dset_oloc_addr));
+            same_chunk = same_chunk && (coll_entry->index_info.chunk_idx ==
+                                        chunk_list->chunk_infos[dbg_idx].index_info.chunk_idx);
+
+            if (same_chunk) {
                 bool coords_match =
                     !memcmp(scaled_coords, chunk_list->chunk_infos[dbg_idx].chunk_info->scaled,
-                            di->dset->shared->ndims * sizeof(hsize_t));
+                            cached_dset_info->dset->shared->ndims * sizeof(hsize_t));
```
fortnern: If the dataset and index are the same is it still necessary to check the coordinates?

fortnern:
I've finished reviewing this, looks good aside from my comments/questions

src/H5Dmpio.c
jhendersonHDF: The structs are memset'ed during I/O setup at https://github.com/jhendersonHDF/hdf5/blob/parallel_comp_multi_dset/src/H5Dmpio.c#L3206-L3207. Then, the multi-field key approach is only used in one place in the new code for hashing and the code uses the structs directly (without them being copied into another temporary structure with possible garbage for example), so there shouldn't be any garbage in the structs during hashing. The same struct is nested inside the `H5D_chunk_insert_info_t` structure that is used for chunk index reinsertion (since we need the chunk index and other fields), so there could be garbage in that case, but that code uses a different hash table with a key based on a single field instead of the compound key approach, so we should be good.

src/H5Dmpio.c
fortnern: Oh wow not sure how I missed that

src/H5Dmpio.c
jhendersonHDF: I can probably refactor this for a bit of sharing to avoid the duplicated allocations, but since the code tries to send the minimal amount of data per chunk around through MPI, the hash tables are indexed in two different ways. The `chunk_read` and `chunk_update` functions both use a struct like:

```
    typedef struct {
        const H5D_dset_io_info_t *dset_io_info;
        H5D_fill_buf_info_t       fb_info;
        hsize_t                   file_chunk_size;
        H5S_t                    *fill_space;
        bool                      should_fill;
        bool                      fb_info_init;
        bool                      index_empty;

        UT_hash_handle hh;
    } per_dset_info;
```

and both use the `dset_io_info` field/pointer as a key, so those two functions can share a hash table. The `chunk_reallocate` and `chunk_reinsert` use structs that are close and could be shared if I update the second one to just include the (useless for its case) `dset` pointer:

```
    typedef struct {
        H5D_chk_idx_info_t chunk_idx_info;
        haddr_t            dset_oloc_addr;

        UT_hash_handle hh;
    } per_dset_info;
```

```
    typedef struct {
        H5D_chk_idx_info_t chunk_idx_info;
        haddr_t            dset_oloc_addr;
        H5D_t             *dset;

        UT_hash_handle hh;
    } per_dset_info;
```

Both functions hash based on the `dset_oloc_addr` field, so those two can shared. Let me look into this more, there may be a chance to reduce this down to only two hash table allocations, but maybe just one with some refactoring.

src/H5Dmpio.c
jhendersonHDF: Definitely not necessary, though it's very helpful to me in catching implementation bugs. I mostly added this check because of the code and note above where extensible arrays do not act predictably like other chunk index types:

```
             * Extensible arrays where the unlimited dimension is not
             * the slowest-changing dimension "swizzle" the coordinates
             * to move the unlimited dimension value to offset 0. Therefore,
             * we use the "swizzled" down chunks to calculate the "swizzled"
             * scaled coordinates and then we undo the "swizzle" operation.
             *
```

src/H5Dmpio.c
fortnern: OK sounds good, we might want to add comments about this though, since this would be very hard to debug if someone changes it so those conditions no longer hold true

src/H5Dmpio.c
fortnern: Should it be an assertion then?

jhendersonHDF: It is just below, though the GitHub view may not be great at showing that

src/H5FDmpio.c
jhendersonHDF: Actually, after more investigation it looks like the check is needed because there is some form of optimization being applied where entries in `s_bufs` are being left uninitialized and so the `if (s_bufs[j + 1].vp != NULL)` check is used to check for this optimization before checking any further into the array.

src/H5Dmpio.c
jhendersonHDF: @fortnern, I was able to collapse this down so that every function shares the same hash table that is constructed once on setup and it cleaned up the code a lot. However, this revealed some collective metadata issues where some ranks weren't correctly participating in collective metadata reads of the dataset chunk index where they should have been due to the `skip_io` approach in `H5D__read` causing them to skip over certain parts of the I/O code; this issue is somewhat related to #2658.

The most reasonable way to fix this seemed to push an initial collective read of the index into the `io_init` callback in `H5Dchunk.c`. However, this means that even when a dataset is being skipped for I/O, every rank still needs to call the `io_init` and `io_term` callbacks to ensure proper behavior. I could've pushed the logic further up into `H5D__read`/`H5D__write`, but decided not to since it's chunked dataset-specific. If this turns out to be a performance problem, I think it can be mitigated by updating `H5D__chunk_io_init` to skip certain portions of its code if a rank isn't really doing anything.

src/H5Dchunk.c
```diff
@@ -1114,6 +1114,31 @@ H5D__chunk_io_init(H5D_io_info_t *io_info, H5D_dset_io_info_t *dinfo)
...
+
+        /*
+         * TODO: Until the dataset chunk index callback structure has
+         * callbacks for checking if an index is opened and also for
+         * directly opening the index, the following fake chunk lookup
+         * serves the purpose of forcing a chunk index open operation
+         * on all ranks
+         */
+        if (H5D__chunk_lookup(dataset, scaled, &udata) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "unable to collectively open dataset chunk index");
```
jhendersonHDF, line 1138: I plan to address this properly when I address #2658, but for now this is about the only way to ensure that every rank has the dataset's chunk index opened and read in correctly.

src/H5Fmpi.c
jhendersonHDF, line 1: Created a "shared file pointer" version of `H5F_get_coll_metadata_reads` so that it can be used inside the chunk I/O init callback

src/H5Dmpio.c
```diff
@@ -1867,15 +1987,42 @@ H5D__link_chunk_filtered_collective_io(H5D_io_info_t *io_info, H5D_dset_io_info_
...
+
+            if (curr_dset_info->fb_info_init && H5D__fill_term(&curr_dset_info->fb_info) < 0)
+                HDONE_ERROR(H5E_DATASET, H5E_CANTFREE, FAIL, "can't release fill buffer info");
+            if (curr_dset_info->fill_space && H5S_close(curr_dset_info->fill_space) < 0)
+                HDONE_ERROR(H5E_DATASET, H5E_CLOSEERROR, FAIL, "can't close fill space");
+
+            H5MM_free(curr_dset_info);
+            curr_dset_info = NULL;
+        }
+    }
```
jhendersonHDF, line 2015: Cleanup for the dataset info hash table occurs at the top-level in the I/O routines now rather than in each sub-function that was creating/freeing the hash table each time

src/H5Dmpio.c
```diff
@@ -2927,187 +3134,352 @@ H5D__obtain_mpio_mode(H5D_io_info_t *io_info, H5D_dset_io_info_t *di, uint8_t as
...
+         * we can just use the info object directly if only one dataset is being
+         * worked on.
+         */
+        if (num_dset_infos > 1) {
+            HASH_ADD(hh, chunk_list->dset_info.dset_info_hash_table, dset_oloc_addr, sizeof(haddr_t),
+                     curr_dset_info);
+        }
+        else
+            chunk_list->dset_info.single_dset_info = curr_dset_info;
+        curr_dset_info = NULL;
```
jhendersonHDF, line 3296: All of the above logic was moved from each sub-function creating the hash table each time into this single spot in the I/O setup. All the functions now index into the hash table according to the dataset object header address so that the same struct can be shared among all the sub-functions

src/H5Dmpio.c
```diff
@@ -2927,187 +3134,352 @@ H5D__obtain_mpio_mode(H5D_io_info_t *io_info, H5D_dset_io_info_t *di, uint8_t as
...
-                local_info_array[i].index_info.chunk_idx = udata.chunk_idx;
-            else
-                local_info_array[i].index_info.chunk_idx = chunk_info->index;
-
-            local_info_array[i].index_info.filter_mask = udata.filter_mask;
-            local_info_array[i].index_info.need_insert = false;
-
-            chunk_node = H5SL_next(chunk_node);
+            if (H5D__chunk_lookup(di[dset_idx].dset, scaled, &udata) < 0)
+                HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "error looking up chunk address");
```
jhendersonHDF, line 3467: This fake lookup should no longer be necessary with the changes to the chunk `io_init` callback, but is probably better addressed with the solution to #2658

src/H5Dio.c
jhendersonHDF: Thinking about this more, I feel like this may even belong higher up in the library since `H5D__read` is called in a few other places. It also seems like good compartmentalization to just have `H5D__ioinfo_init` init the `io_info` structure without doing much else.

src/H5FDmpio.c
```diff
@@ -3097,9 +3097,9 @@ H5FD__mpio_read_selection(H5FD_t *_file, H5FD_mem_t type, hid_t H5_ATTR_UNUSED d
              * s_bufs[] to find the smallest value, and choose that for
              * mpi_bufs_base.
              */
-            j = 0; /* guess at the index of the smallest value of s_bufs[] */
 
-            if (s_bufs[j + 1].vp != NULL) {
+            j = 0; /* guess at the index of the smallest value of s_bufs[] */
+            if ((count > 1) && (s_bufs[1].vp != NULL)) {
```
fortnern, line 3102: Have you tried removing the if statement entirely?

jhendersonHDF: Yes, removing the if statement lead to warnings about use of uninitialized entries in `s_bufs` due to some optimization that is being applied in `H5FD_sort_selection_io_req` that appears to skip setting entries beyond `s_bufs[0]` and `s_bufs[1]` to anything.

jhendersonHDF:
Looks like nvhpc is also back to being unhappy sporadically. I'll still need to investigate that

src/H5Dmpio.c
```diff
@@ -3291,6 +3491,37 @@ H5D__mpio_collective_filtered_chunk_io_setup(const H5D_io_info_t *io_info, const
 
 done:
     if (ret_value < 0) {
+        /* Free temporary cached dataset info object */
```
fortnern, line 3489: Is this necessary? Seems like it's taken care of in the calling functions H5D__multi_chunk_filtered_collective_io and H5D__link_chunk_filtered_collective_io

jhendersonHDF: Not strictly necessary, just making sure things get cleaned up on failure in case it's called from elsewhere in the future. Although maybe I should move the freeing logic into a function to reduce duplication.

fortnern (approved):

jhendersonHDF:
@fortnern, you can ignore the commit adding MPI_Barrier calls to the tests, I just wanted to have those in there for future proofing reasons.

The real issue was that the "fill time never" and "undefined fill value" tests create datasets and then try to immediately read from them without first writing to them to check for bad behavior. Since those tests won't write fill values to the unfiltered chunks when testing the multi-dataset I/O APIs with a mixture of filtered and unfiltered datasets, this means that, in parallel, the file size wasn't getting updated unless the chunks get directly written to or the file is closed and reopened. This meant the read call was reading past the end of the file, which is usually fine, except that the MPI I/O driver tries to fill the rest of the application buffer with 0 at https://github.com/HDFGroup/hdf5/blob/develop/src/H5FDmpio.c#L1449-L1453. In the single dataset case this is fine, but in the multi-dataset case the application buffers aren't necessarily contiguous and so when the library tries to do the memset, it causes a heap buffer overflow. I'll try to fix this in the library at some point by skipping the memset if the memory datatype has strides in it or would otherwise cause this to be unsafe.

brtnfld (approved):

PR #3602, opened 2023-09-26
Fix potential uninitialized variable

derobins:
Moves a union initialization up a bit so it's performed before code that can jump to the cleanup target, where file descriptors could be checked without being initialized.

This could only happen in test code and only in an out-of-memory situation.

Fixes Coverity 1542254


derobins:
Apologies for the unimportant PR, but I wanted to go through the process of addressing a Coverity issue to see how things work.

byrnHDF (approved):

lrknox (approved):

PR #3603, opened 2023-09-27
Remove unnecessary assignment in test generator

derobins:
Fixes what looks like a copy/paste/modify error in the format convert test file generator, where an array element is assigned one value and them immediately overwritten by another value.

Fixes Coverity issue 1542285


derobins:
Another trivial Coverity fix

hyoklee (approved):

byrnHDF (approved):

PR #3606, opened 2023-09-27
Merge changes from 1.10 release file

byrnHDF:


release_docs/RELEASE.txt
```diff
@@ -232,41 +234,29 @@ New Features
...
-      so the library couldn't shut down. When the "using a memory checker"
-      option is used, the free lists are disabled so there is just a memory
-      leak with no abort on library shutdown.
+      Under most conditions (i.e., when the --enable-using-memchecker option is
+      NOT used), this would result in a small memory leak and and infinite loop
+      and abort when shutting down the library. The infinite loop would be due
+      to the "free list" package not being able to clear its resources so the
+      library couldn't shut down. When the "using a memory checker" option is
+      used, the free lists are disabled so there is just a memory leak with no
+      abort on library shutdown.
```
lrknox, line 255: Does the last sentence describe how things work after a change to fix the problem?  If so, would starting the sentence with "Now" , or inserting "now" in "the free lists are now disabled" make this clearer?

release_docs/RELEASE.txt
```diff
@@ -289,11 +279,42 @@ Bug Fixes since HDF5-1.12.2 release
...
+
+      The assert statement in question has been replaced with pointer checks
+      that don't raise errors. Since the function is in cleanup code, we do
+      our best to close and free things, even when presented with partially
+      initialized structs.
+
+      Fixes CVE-2016-4332 and HDFFV-9950 (confirmed via the cve_hdf5 repo)
+
+    - Fixed segfault on file close
+
```
lrknox, line 295: This entry is brought from HDF5 1.10.11 but is new to HDF5 1.12.3.  Was the corresponding code fix included in hdf5_1_12?

byrnHDF: Might have been accidental when moving the previous item?

lrknox: No, it was just moved up from line 387, which I didn't notice before.  It belongs in the file;  I'm not sure why the merge moved it.

release_docs/RELEASE.txt
byrnHDF: I think it's fine the way it is.

loricooperhdf (approved):

PR #3619, opened 2023-09-29
Fix Intel oneAPI icc warning

hyoklee:
This fixes the following warning:
```
/home/runner/work/hdf5/hdf5/test/h5test.c(484): warning #3179: deprecated conversion of string literal to char* (should be const char*)
                      suffix = nest_printf ? "-%%06d.h5" : "-%06d.h5";
                                                         ^
```


byrnHDF (approved):

derobins (approved):

derobins:
Weird that the constructs differ in const-ness. Probably a fragile fix, though, since surely someone will try to re-simplify that at some poing.

Issue #3613, opened 2023-09-28
Ensure static + threadsafe fails on Windows


derobins:
We can't clean up thread-specific values without DllMain, so this configuration needs to fail with Win32 in CMake.

derobins:
This is now enforced in CMake

PR #3622, opened 2023-10-03
Disable static + thread-safe on Windows w/ CMake

derobins:
The thread-safety feature on Windows requires a hook in DllMain() and thus is only available when HDF5 is built as a shared library.

This was previously a warning, but has now been elevated to a fatal error that cannot be overridden with ALLOW_UNSUPPORTED.

Fixes GitHub #3613


derobins:
*sigh* - codespell updated :(

fortnern (approved):

byrnHDF (approved):

PR #3625, opened 2023-10-03
Throw error in ID iteration for NULL object

mattjala:
Fixes a segfault that would occur when using a non-native VOL with an H5Iiterate callback that doesn't check for NULL itself (for example, `H5Z__check_unregister_dset_cb` as used in `H5Zunregister`).


mattjala:
This ended up causing more problem for the VOL connectors than it solved - a lot of API calls use `H5Iiterate` in a way that doesn't depend on the provided object pointer. A different workaround for the problem this was trying to solve (using `H5Zunregister` with VOL connectors) is provided in #3629.

PR #3629, opened 2023-10-03
Make filter unregister callbacks safe for VOL connectors

mattjala:
When using VOL connectors, the object pointer provided to a callback by `H5I_iterate` may not be valid, since it is either the `data` field of a `vol_obj` if the connector does not define a wrapper get object callback, or the return value of that get object callback if it is defined. Either way, it is unsafe to cast this pointer to `H5X_t*`. 

The callbacks now avoid the potentially unsafe pointer by using the handle id to get the same information.

### DIFF
--- a/src/H5Z.c
+++ b/src/H5Z.c
@@ -485,7 +485,7 @@ done:
  *-------------------------------------------------------------------------
  */
 static int
-H5Z__check_unregister_group_cb(void *obj_ptr, hid_t H5_ATTR_UNUSED obj_id, void *key)
+H5Z__check_unregister_group_cb(void H5_ATTR_UNUSED *obj_ptr, hid_t obj_id, void *key)
 {
     hid_t         ocpl_id         = -1;
     H5Z_object_t *object          = (H5Z_object_t *)key;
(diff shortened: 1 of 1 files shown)


src/H5Z.c
```diff
@@ -494,10 +494,8 @@ H5Z__check_unregister_group_cb(void *obj_ptr, hid_t H5_ATTR_UNUSED obj_id, void
 
     FUNC_ENTER_PACKAGE
 
-    assert(obj_ptr);
-
     /* Get the group creation property */
-    if ((ocpl_id = H5G_get_create_plist((H5G_t *)obj_ptr)) < 0)
+    if ((ocpl_id = H5Gget_create_plist(obj_id)) < 0)
```
derobins: Can't use public API calls in an internal API call (here and elsewhere).

fortnern (approved):

src/H5Z.c
```diff
@@ -581,26 +603,51 @@ H5Z__check_unregister_dset_cb(void *obj_ptr, hid_t H5_ATTR_UNUSED obj_id, void *
...
+    if (H5F_ACC_RDWR & intent) {
 
 #ifdef H5_HAVE_PARALLEL
+        H5G_loc_t loc;
+        H5F_t    *f = NULL;
+
         /* Check if MPIO driver is used */
+        if (H5G_loc(obj_id, &loc) < 0)
+            HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a location");
+        f = loc.oloc->file;
```
jhendersonHDF: This appears to have the same issue of assuming that we can cast things to an `H5F_t` and perform native operations on them

mattjala: You're right, `H5G_loc` does the same thing under the hood. Even if we didn't use it here, `H5F_mpi_retrieve_comm` invokes `H5G_loc` on the same object id anyway. Would it make more sense to just throw an error if we're using a non-native VOL and HDF5 was built parallel?

jhendersonHDF: Since the filter interface is not exactly exposed to non-native VOLs other than having the ability to make H5Z calls, I would probably check if the object is native and just skip this whole block if it's not, but don't throw an error. I would have done the same for the rest of the callbacks changed in this PR - just continue iteration if the object is non-native - but this PR is a good change towards being able to have non-native VOLs interact with the filter interface in the future.

jhendersonHDF (approved):

Issue #3587, opened 2023-09-25
1.12.2 and higher parallel build fails to compile with an MPI v2.2 implementation. Undefined reference to MPI_Type_create_hindexed_block 


corppeon:
Undefined reference to MPI_Type_create_hindexed_block link error with gcc and Platform MPI, HDF5_ENABLE_PARALLEL=True.

Platform MPI implements the MPI v2.2 API. There are references to the method MPI_Type_create_hindexed_block in H5Dchunk.c, H5D__chunk_collective_fill() method. These calls require the MPI v3.0 API or higher.

These H5Dchunk calls should be protected with a conditional define, similar to those used in H5Smpio.c (H5_CHECK_MPI_VERSION(3, 0)), and an alternative implementation, probably using MPI_Type_create_hindexed used for the older MPI API.

wreckdump:
What distribution and version of MPI are you using? I've had similar problem with OpenMPI 4.1.5. Try using older version of OpenMPI or try using MPICH, instead.

corppeon:
It's Platform MPI v9.01. We do use other MPIs as well. Other MPI distributions with a pre v3.0 API will see the same issue

brtnfld:
I have submitted a PR to resolve the compilation problem. However, there are still failed tests with MPI-2. Unfortunately, there is insufficient drive to continue supporting MPI-2 beyond this point. This is mainly due to the small number of users who use MPI-2 and the significant resources required to fix/maintain it. We would need to add CI tests and workarounds for MPI bugs that have since been fixed. As a result, there are few situations in which it is essential to support it.

As a result, we have decided to remove support for MPI-2 in the upcoming 1.14 release officially.

Out of curiosity, what 1.12 feature do you want over what I assume is HDF5 1.8 on the system?

corppeon:
we were previously using 1.12.1 which worked. Our motive for upgrading HDF5 to latest is to incorporate latest security fixes

derobins:
FWIW, 1.12.3 will be the last release of HDF5 1.12, after which that maintenance branch will be retired. I'd highly recommend moving to 1.14.

brtnfld:
We no longer support 1.12, and 1.14 does not support MPI 2.2, so I will close this issue.

PR #3641, opened 2023-10-05
HDF5 with MPI2 build fix

brtnfld:
Fixes https://github.com/HDFGroup/hdf5/issues/3587. However, the parallel tests t_2Gio, t_filter_parallel, t_bigio and testphdf5 will fail with MPI2 (mpich 1.5) and seem unrelated to this fix. So, it is still not recommended to use 1.12 with mpi2.


hyoklee (approved):

hyoklee:
BTW, there are many comment blocks that have "MPI-2":

```
src/H5Smpio.c:             * so we're using the MPI-2 version even though we don't need t\
he lb
src/H5FDmpio.h: * \param[in] comm MPI-2 communicator
src/H5FDmpio.h: * \param[in] info MPI-2 info object
src/H5FDmpio.h: *          in \c MPI_File_open of MPI-2. This function makes a duplicate \
of the
src/H5FDmpio.h: *          in MPI_File_open() of MPI-2. This function makes a duplicate c\
opy of
src/H5FDmpio.h: * \param[out] comm MPI-2 communicator
src/H5FDmpio.h: * \param[out] info MPI-2 info object
src/H5Fmodule.h: * are defined in MPI_FILE_OPEN of MPI-2.
src/H5FDmpio.c: * Purpose:     This is the MPI-2 I/O driver.
src/H5FDmpio.c: *              defined in MPI_FILE_OPEN of MPI-2. This function makes a
src/H5FDmpio.c: *              defined in MPI_FILE_OPEN of MPI-2. This function makes a
src/H5mpi.c:         * so we're using the MPI-2 version even though we don't need the lb
```

Can they be generalized as "MPI" by dropping "-2"?

byrnHDF (approved):

lrknox (approved):

PR #3642, opened 2023-10-06
Add autotools config files for NVIDIA compilers, always add -fPIC flag for nvfortran.

lrknox:
Note that nvcc compiler doesn't match compiler and version processing in nvidia-flags because the output of "nvcc -V" doesn't provide version information on the line beginning with "nvcc".  The only version information in the output is the Cuda version.  As it is now, nvcc will not get the -c99 flag from nvida-flags.  We could add different sed lines to get the compiler_vendor and compiler_version information with the Cuda version, but I'm told we should use nvc and not nvcc, so nvcc is not supported.


derobins (approved):
Looks good (as long as it's passing tests)

config/nvidia-fflags
```diff
@@ -0,0 +1,139 @@
...
+    #else
+    #    PROD_FCFLAGS="-O2 -s"
+    #fi
+    PROD_FCFLAGS="-fast -Mnoframe -fPIC"
+
+    #########
+    # Debug #
+    #########
+
+    DEBUG_FCFLAGS="-Mbounds -Mchkptr -Mdclchk -fPIC"
```
jhendersonHDF, line 86: I believe chkptr and dclchk are no longer an option for nvc. I see:

```
nvc --help | grep chk
-M[no]depchk        Check dependence relations for vector or parallel code
-Mchkstk            Check for sufficient stack space upon subprogram entry
```

jhendersonHDF (approved):

byrnHDF (approved):

config/cmake/HDFFortranCompilerFlags.cmake
```diff
@@ -57,6 +57,15 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "NAG")
     list (APPEND HDF5_CMAKE_Fortran_FLAGS "-ieee=full")
 endif ()
 
+if (CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC")
+  set (CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fPIC")
```
byrnHDF: This might not be needed as CMake should handle this flag.

config/cmake/HDFCXXCompilerFlags.cmake
```diff
@@ -49,6 +49,15 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL SunPro AND CMAKE_CXX_COMPILER_LOADED)
   endif ()
 endif ()
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "NVHPC")
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Minform=warn")
```
byrnHDF: c++11 flag should have a CMake option available.

byrnHDF: if (CMAKE_CXX_COMPILER_ID STREQUAL "NVHPC" AND CMAKE_CXX_COMPILER_LOADED)
  if (NOT DEFINED CMAKE_CXX${CMAKE_CXX_STANDARD}_STANDARD_COMPILE_OPTION)
      if (NOT CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD EQUAL 11)
        set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CMAKE_C11_STANDARD_COMPILE_OPTION}")
      endif ()
  endif ()
endif ()

lrknox: Added this code.

config/cmake/HDFFortranCompilerFlags.cmake
lrknox: Yes, it does

lrknox:
linux autotools nvhpc test fails when disk space is filled.  I tried it twice with the same results:

"H5Tdbg.c", line 365: warning: statement is unreachable [code_is_unreachable]
                  break;
                  ^
"H5VLnative_attr.c", line 79: warning: variable "plist" was set but never used [set_but_not_used]
Error: No space left on device : '/home/runner/runners/2.309.0/_diag/blocks/ff8f94e9-195c-4cef-9334-c4fd08e4cf2a_ca395085-040a-526b-2ce8-bdc85f692774.1'

I don't know what we can do about this.

Otherwise I think the PR is ready for final review.

byrnHDF (approved):

derobins:
Replaced with #3686

PR #3643, opened 2023-10-06
dropping MPI-2 support

brtnfld:


src/H5FDmpio.c
jhendersonHDF, line 1: Much nicer here!

jhendersonHDF (approved):

hyoklee (approved):

derobins (approved):
Approved, but this should not be merged to 1.14 until we go through the change process (i.e., it'll show up in 1.14.4).

brtnfld:
Technically this is not a change in functionality from a user's perspective since MPI-2 has not been supported/working since 12.1.

PR #3644, opened 2023-10-08
Switch IEEE flags for NAG Fortran

derobins:
Default is -ieee=stop, which causes problems when the H5T module performs floating-point type introspection.

The new mode is -ieee=full

### DIFF
--- a/config/linux-gnulibc1
+++ b/config/linux-gnulibc1
@@ -131,8 +131,6 @@ case $FC_BASENAME in
         H5_CFLAGS="$H5_CFLAGS"
 
         F9XSUFFIXFLAG=""
-# We force compiler to use upper case for external names
-# (just in case since this should be a default EIP)
         H5_FCFLAGS="$H5_FCFLAGS"
         FSEARCH_DIRS=""
 
@@ -162,9 +160,10 @@ case $FC_BASENAME in
     nagfor)
 
         F9XSUFFIXFLAG=""
-# We force compiler to use upper case for external names
-# (just in case since this should be a default EIP)
-        H5_FCFLAGS="$H5_FCFLAGS"
+        # NOTE: The default is -ieee=stop, which will cause problems
+        #       when the H5T module performs floating-point type
+        #       introspection
+        H5_FCFLAGS="$H5_FCFLAGS -ieee=full"
         FSEARCH_DIRS=""
 
         # Production


derobins:
@byrnHDF - How do we need to handle this in CMake?

derobins:
Current options docs:

https://support.nag.com/nagware/np/r70_doc/manual/compiler_2_4.html#OPTIONS

byrnHDF:
> @byrnHDF - How do we need to handle this in CMake?

Need to know what the compilerID is from the CMakeFiles dir and add it to the compiler flags file in config/cmake

qkoziol (approved):

config/linux-gnulibc1
```diff
@@ -162,9 +160,10 @@ case $FC_BASENAME in
     nagfor)
 
         F9XSUFFIXFLAG=""
-# We force compiler to use upper case for external names
-# (just in case since this should be a default EIP)
-        H5_FCFLAGS="$H5_FCFLAGS"
+        # NOTE: The default is -ieee=stop, which will cause problems
+        #       when the H5T module performs floating-point type
+        #       introspection
+        H5_FCFLAGS="$H5_FCFLAGS -ieee=full"
```
brtnfld, line 166: Would it not be better to check if the flag is supported with AC_COMPILE_IFELSE and throw an error if it is not?

untested:
```
AC_MSG_CHECKING([whether compile flag "-ieee" is supported])

H5_FCFLAGS="$H5_FCFLAGS -ieee=full"
AC_LANG_PUSH([Fortran])

AC_COMPILE_IFELSE([AC_LANG_PROGRAM([],
  [[END]])],
  [AC_MSG_RESULT([yes])],
  [AC_MSG_RESULT([no])
  AC_MSG_ERROR([nagfor requires compiler flag -ieee support])]
)

AC_LANG_POP([Fortran])
```

derobins: This is only for 1.14.3 and up and only for NAG Fortran. Also, we don't put autoconf macros in the compiler-specifc flags files. Those are just shell scripts that set flag variables. We could make -ieee=full dependent on the nagfor version, but is that really necessary?

derobins: fwiw, 6.1 supports -ieee=full and that's as far back as I can find documentation (circa 2017)

brtnfld: I think -ieee is only for 7+ versions, I don't think the 6 series supports it, and the 6 series is not that old.

Right, this is meant for an m4 module.

brtnfld: ok, I see it is supported in 6.1

brtnfld: It is probably okay then since I'm not sure if the earlier versions of NAG supported F2003, their documentation of the release history is not great

brtnfld (approved):

Issue #3620, opened 2023-10-02
Deprecated API functions can segfault with non-Native connectors


mattjala:
Old API functions  assume that the native VOL is being used, and cast the object data received from `H5VL_object_data` to a native type like `H5X_t` in a way that often causes invalid memory accesses later.

A specific example from [this issue encountered with the REST VOL](https://github.com/HDFGroup/vol-rest/pull/77#issuecomment-1741500166): when calling `H5Oget_info_by_name1`, a return of type `RV_object_t*` is later cast by `H5VL_native_get_file_struct` to `H5F_t*`, which causes a segfault on later attempts to access the shared pointer.

Since support for 1.10 is already dropped, and support for 1.12 is about to be dropped, this only needs to be dealt with in 1.14. There are a few possible solutions:
- Leave the deprecated functions as they are, because they shouldn't be used in 1.14+ anyway.
- Add checks that the native VOL is being used to any deprecated functions that work with objects.
- Remove the deprecated functions entirely.

fortnern:
`H5Oget_object_by_info1` doesn't exist, and `H5Oget_info1` doesn't have this problem as far as I can tell. What is an example of a function that shows this problem?

fortnern:
I think the issue is the call to `H5VL_native_token_to_addr()` in `H5O__get_info_old()`. We should fix this so it works properly with non-native VOLs.

PR #3647, opened 2023-10-09
Add native VOL checks to deprecated functions

mattjala:
Fix for #3620

Currently causes the HDF EOS tests to fail in `HE5_sw_writedata`, upon tyring to reclaim data from a variable length datatype during `HE5_PRwrite`.

### DIFF
--- a/src/H5Odeprec.c
+++ b/src/H5Odeprec.c
@@ -116,9 +116,11 @@ static herr_t
 H5O__iterate1_adapter(hid_t obj_id, const char *name, const H5O_info2_t *oinfo2, void *op_data)
 {
     H5O_visit1_adapter_t *shim_data = (H5O_visit1_adapter_t *)op_data;
-    H5O_info1_t           oinfo;                    /* Deprecated object info struct */
-    unsigned              dm_fields;                /* Fields for data model query */
-    unsigned              nat_fields;               /* Fields for native query */
+    H5O_info1_t           oinfo;      /* Deprecated object info struct */
+    unsigned              dm_fields;  /* Fields for data model query */
+    unsigned              nat_fields; /* Fields for native query */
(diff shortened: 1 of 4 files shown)


qkoziol (approved):

mattjala:
> Currently causes the HDF EOS tests to fail in `HE5_sw_writedata`, upon tyring to reclaim data from a variable length datatype during `HE5_PRwrite`.

Fixed by removing the native VOL check in `H5Dvlen_reclaim`, which turned out to be unnecessary anyway.

src/H5Odeprec.c
```diff
@@ -127,6 +129,19 @@ H5O__iterate1_adapter(hid_t obj_id, const char *name, const H5O_info2_t *oinfo2,
     assert(oinfo2);
     assert(op_data);
 
+    /* Get the VOL object */
+    if (NULL == (vol_obj = H5VL_vol_object(obj_id)))
+        HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "invalid location identifier");
+
+    /* Check if using native VOL connector */
+    if (H5VL_object_is_native(vol_obj, &is_native) < 0)
```
fortnern: This is probably fine, but I think it would be better to put this check at the top level, in HOvisit1(), etc. Even though it needs to be in more places I think it's better to catch it before it goes into the VOL connector at all and give a cleaner error message.

mattjala:
I moved the native VOL checks from the internal `H5O__iterate1_adapter`/`H5O__get_info_old` functions up into the API functions that use them internally. Also renamed the boolean `is_native -> is_native_vol_obj` for consistency within the file.

fortnern (approved):

jhendersonHDF (approved):

PR #3651, opened 2023-10-09
Use real URLs and updated names for plugins

byrnHDF:


glennsong09 (approved):

byrnHDF:
All the checks pass, but the daily_build CI will fail trying use the blosc filter. The blosc filter does work in a standalone plugins build, just not when built inline with HDF5.

lrknox (approved):

PR #3657, opened 2023-10-11
Add AOCC GitHub Action (#3504)

### LINKED ISSUE #3504: t_mpi test fails with OpenMPI 4.1.5, but passes with MPICH 4.1.2
**Describe the bug**
t_mpi test fails with OpenMPI 4.1.5, but passes with MPICH 4.1.2 (both compiled with AMD AOCC 4.1.0). And this affects the `netcdf` parallel related functions. For example, when `nc_open_par()` function from `netcdf` is called, it hangs without any activity.

**Expected behavior**
I expect the hdf5 parallel related tests and functions to work properly.

 - HDF5 version : 1.14.2
 - OS and version : Arch Linux 6.4.12-arch1-1
 - Compiler and version : AMD AOCC 4.1.0
 - Build system (e.g. CMake, Autotools) and version : GNU Make 4.4.1, Autoconf 2.71
 - Any configure options you specified : `--enable-parallel` during configuration stage
 - MPI library and version (parallel HDF5) : OpenMPI 4.1.5 (problematic), MPICH 4.1.2 (functional)

hyoklee:
close #3504

Note: HDF5 Fortran doesn't work with this action. Enabling it will fail the action.

### DIFF
--- a/.github/workflows/linux-auto-aocc-ompi.yml
+++ b/.github/workflows/linux-auto-aocc-ompi.yml
@@ -2,8 +2,26 @@ name: linux autotools aocc ompi
 
 on:
   push:
+  pull_request:  
     branches:
       - develop
+    paths-ignore:
+      - '.github/CODEOWNERS'
+      - '.github/FUNDING.yml'
+      - 'doc/**'
+      - 'release_docs/**'
+      - 'ACKNOWLEDGEMENTS'
+      - 'COPYING**'
+      - '**.md'
+
+# Using concurrency to cancel any in-progress job or run
+concurrency:
+  group: ${{ github.workflow }}-${{ github.sha || github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+      
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


derobins (approved):

derobins:
> close #3504
> 
> Note: HDF5 Fortran doesn't work with this action. Enabling it will fail the action.

Why does Fortran fail?

hyoklee:
> > close #3504
> > Note: HDF5 Fortran doesn't work with this action. Enabling it will fail the action.
> 
> Why does Fortran fail?

```
  FCLD     libhdf5_fortran.la
clang-16: error: unknown argument: '-soname'
clang-16: error: no such file or directory: 'libhdf5_fortran.so.1000'
```

bmribler (approved):

.github/workflows/linux-auto-aocc-ompi.yml
```diff
@@ -0,0 +1,48 @@
+name: linux autotools aocc ompi
+
+on:
+  push:
```
derobins, line 4: Needs to also be on PR

derobins (changes requested):
Should also be on PR

derobins:
I have a fix for AOCC w/ shared/Fortran/parallel. Let's get that in and then modify this PR to build + test more stuff.

derobins (approved):

PR #3660, opened 2023-10-11
1.14 sync with develop

jhendersonHDF:
Please look over this and let me know if there's anything that shouldn't be merged into the 1.14 branch.

@byrnHDF it was easiest to sync everything from develop, including the gzip and szip changes. If those weren't supposed to come to 1.14, please let me know so I can revert whichever commits are relevant.


CMakePresets.json
```diff
@@ -74,28 +70,28 @@
...
     },
     {
       "name": "ci-ExamplesVars",
       "hidden": true,
       "cacheVariables": {
-        "HDF5_EXAMPLES_COMPRESSED": {"type": "STRING", "value": "hdf5-examples-2.0.4.tar.gz"},
-        "HDF5_EXAMPLES_COMPRESSED_DIR": {"type": "STRING", "value": "${sourceParentDir}/temp"},
-        "EXAMPLES_TGZ_ORIGPATH": {"type": "STRING", "value": "https://github.com/HDFGroup/hdf5-examples/archive/refs/tags/"},
-        "EXAMPLES_TGZ_ORIGNAME": {"type": "STRING", "value": "2.0.4.tar.gz"}
+        "HDF5_EXAMPLES_COMPRESSED": {"type": "STRING", "value": "hdf5-examples-master.tar.gz"},
```
jhendersonHDF, line 91: Some of these examples links seemed suspect going from `2.0.4` to `master` so I didn't know if this was correct or not for 1.14 as opposed to develop

.github/workflows/h5py.yml
byrnHDF, line 1: This should not be added - cron type CI only works on default branch

.github/workflows/autotools.yml
jhendersonHDF, line 1: Looks like some of these synced workflow files need updating in this PR

.github/workflows/main-cmake.yml
```diff
@@ -0,0 +1,251 @@
+name: hdf5 dev CMake CI
```
byrnHDF: name should be hdf5 1.14 in all these yaml files

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '36 22 * * 4'
```
byrnHDF: Should just remove the schedule item and leave the push?

CMakeFilters.cmake
```diff
@@ -131,7 +119,7 @@ endif ()
 #-----------------------------------------------------------------------------
 # Option for SzLib support
 #-----------------------------------------------------------------------------
-option (HDF5_ENABLE_SZIP_SUPPORT "Use SZip Filter" OFF)
+option (HDF5_ENABLE_SZIP_SUPPORT "Use SZip Filter" ON)
```
byrnHDF, line 122: This should probably remain as previously

CMakePlugins.cmake
byrnHDF, line 1: This has an issue with blosc plugin - either the plugin stuff should be left as is or blosc will need to be excluded.

CMakePresets.json
byrnHDF: This will be gone when examples is moved in

.github/workflows/main-cmake.yml
jhendersonHDF: will do

.github/workflows/scorecard.yml
jhendersonHDF: yes I think that may have been how it was before. Will revert

CMakeFilters.cmake
jhendersonHDF: Was this change part of other changes or by itself? I can revert the commit or make the individual change; just concerned about getting confused between the state of this between develop and 1.14

config/cmake/ConfigureChecks.cmake
```diff
@@ -916,7 +909,7 @@ endmacro ()
 #-----------------------------------------------------------------------------
 
 # ----------------------------------------------------------------------
-# Set the flag to indicate that the machine is using a special algorithm to convert
+# Set the flag to indicate that the machine is using a special algorithm toconvert
```
byrnHDF: This went the wrong way!

CMakeFilters.cmake
byrnHDF: There were other changes

byrnHDF:
I'm not sure this is the way to do this - Reviewing will take a long time (it is slow on my computer!)
Maybe we should take it smaller bites?

config/cmake/ConfigureChecks.cmake
jhendersonHDF: Yes, looks like develop is wrong

jhendersonHDF:
> I'm not sure this is the way to do this - Reviewing will take a long time (it is slow on my computer!) Maybe we should take it smaller bites?

The best way of course would be to keep the two branches in sync regularly. Since there are apparently 540 commits not merged from develop, I don't know the best way to go about this. Smaller batches may help, but there are still a ton of changes. Cherry picking out the tools move PR may help, but that will make it harder to do a regular merge of the other files.

jhendersonHDF:
git merge-base shows that the last common commit between develop and the 1.14 branch is `b9f4c001ade4c230f8325080f3dd8181648324ab` from December 9. I could start there and pick up things in batches of a hundred commits or so, but the PRs that capture the tools move change and the `FALSE` -> `false`, `hbool_t` -> `bool` changes will still be huge.

lrknox (approved):

Issue #3275, opened 2023-07-26
core driver performance prohibitively slow on Windows


AlexKirko:
**Describe the bug**
We attempted to use the 'core' driver on Windows (in-memory), and quickly discovered that it was about 300x slower on Windows than on MacOS for a 22 GB file. The default driver works fine.

We use HDF5 to store a hierarchy of data points. When building the object, we stream from a binary, do some preprocessing, and then populate the HDF storage with more than a 100 data nodes, each containing about two dozen complex-valued numpy arrays.

**Expected behavior**
The use of the in-memory driver should be faster than the default (on-disk) one and the performance should be comparable between different OS.

**Platform (please complete the following information)**
 - HDF5 version: 1.14.1
 - OS and version: Windows 11 Pro
 - Compiler and version: installed from the official binary, hdf5-1.14.1-2-Std-win10_64-vs17.zip
 - Build system (e.g. CMake, Autotools): Visual Studio 17.6.5
 - Any configure options you specified: Nothing special

**Additional context**
This issue was originally reported on the h5py GitHub in 2021 here: https://github.com/h5py/h5py/issues/1827 , but I guess no one raised it with your group.

The problem seems to be that you rely on a `realloc()` call to get more memory regardless of the OS, and on Windows this often results in the dataset being copied on each additional memory request. In our situation, where we need to request thousands of extensions, hundreds larger than 1MB (default block_size), the storage creation slows down hundreds of times compared to Linux/MacOS because of the copy overhead (also, RAM usage spikes to 2x the size of the dataset on each copy).

[Here](https://blog.kowalczyk.info/article/2be/realloc-on-windows-vs.linux.html) is an explanation of the realloc differences on Windows and Linux.

You want to look [here](https://github.com/HDFGroup/hdf5/blob/85d216634aae94223ed5b75a77585f9fd60009e5/src/H5FDcore.c#L1364) in your code. What people typically do on Windows is request 1.3x the new dataset size. This requires 30% more RAM on Windows, but the performance gets much closer to `O(log(n))` the way it should be, instead of a copy on each increment.

derobins:
That's really interesting. I'll try to take a look at this after we get 1.14.2 out the door.

derobins:
[...]

PR #3661, opened 2023-10-12
Add Intel oneAPI and LLVM for Windows Action.

hyoklee:
This simple action may help Windows users (e.g., #3275 user).

 This action is also good for 1.14.3 release testing as well,
 since it uses the latest CMake 3.27.7 and Intel oneAPI for Windows.

 It has one failure:
```
99% tests passed, 1 tests failed out of 2199

Total Test time (real) = 563.81 sec

The following tests FAILED:
	1731 - H5DUMP-tfloatsattrs (Failed)
```

### DIFF
--- a/.github/workflows/win-ninja-icx.yml
+++ b/.github/workflows/win-ninja-icx.yml
@@ -4,7 +4,26 @@ on:
   push:
     branches:
       - develop
+  pull_request:  
+    branches:
+      - develop
+    paths-ignore:
+      - '.github/CODEOWNERS'
+      - '.github/FUNDING.yml'
+      - 'doc/**'
+      - 'release_docs/**'
+      - 'ACKNOWLEDGEMENTS'
+      - 'COPYING**'
+      - '**.md'
 
+# Using concurrency to cancel any in-progress job or run
+concurrency:
+  group: ${{ github.workflow }}-${{ github.sha || github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+      
 jobs:
   build:
     runs-on: windows-latest
--- a/.github/workflows/win-ninja-icx.yml
+++ b/.github/workflows/win-ninja-icx.yml
@@ -1,0 +1,39 @@
+name: windows ninja icx
+
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  build:
+    runs-on: windows-latest
+    steps:
+      - name: Install CMake and Ninja
+        uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: "3.27.7"
+      - name: Install LLVM and Clang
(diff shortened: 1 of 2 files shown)


byrnHDF:
Why is this not added to the intel-cmake.yml file?

byrnHDF (changes requested):
Integrate into intel-cmake.yml file

byrnHDF:
Also why is CMake 3.27.7 mentioned - the minimum for hdf5 is 3.18.

hyoklee:
> Also why is CMake 3.27.7 mentioned - the minimum for hdf5 is 3.18.

Spack is at 3.27.7.

https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/cmake/package.py

hyoklee:
> Why is this not added to the intel-cmake.yml file?

I want to have a separate badge. E.g., 

https://github.com/hdfeos/hdf5/blob/develop/WATCHME.md

byrnHDF:
We will need to do badges differently, Every callable workflow can return a status all the way back up the stack. One solution might be to use a json file.
Another possibility: https://github.com/marketplace/actions/dynamic-badges

Workflows need to be kept consistent.

.github/workflows/win-ninja-icx.yml
```diff
@@ -0,0 +1,39 @@
+name: windows ninja icx
+
+on:
+  push:
```
derobins, line 4: Should also be PR

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -426,7 +426,7 @@
     configure_file(${PROJECT_SOURCE_DIR}/exportfiles/tbinregR.exp ${PROJECT_BINARY_DIR}/testfiles/std/tbinregR.exp NEWLINE_STYLE CRLF)
     #file (READ ${PROJECT_SOURCE_DIR}/exportfiles/tbinregR.exp TEST_STREAM)
     #file (WRITE ${PROJECT_BINARY_DIR}/testfiles/std/tbinregR.exp "${TEST_STREAM}")
-    HDFTEST_COPY_FILE("${PROJECT_SOURCE_DIR}/expected/tfloatsattrs.wddl" "${PROJECT_BINARY_DIR}/testfiles/std/tfloatsattrs.ddl" "h5dump_std_files")
+    #HDFTEST_COPY_FILE("${PROJECT_SOURCE_DIR}/expected/tfloatsattrs.wddl" "${PROJECT_BINARY_DIR}/testfiles/std/tfloatsattrs.ddl" "h5dump_std_files")
```
byrnHDF, line 429: This change should not be changed in an action - this would need to be a different PR.

byrnHDF (changes requested):
This is not a solution

byrnHDF:
Resolved by #3836

Issue #2928, opened 2023-05-08
FORTRAN 2003 check... fails


hergljkds:
gfortran-13 (tried F90 and 2003, both failed)
x86_64

implemented -fdefault-real-16
trapping all
trying with openmp

..it compiles. checks are done in a sandbox...

check failures:

h5dcreate_f FAILED
h5dclose_f FAILED
h5dopen_f FAILED
H5Dget_create_plist_f FAILED
H5Pget_fill_value_f FAILED
ERROR: INCORRECT REAL VALIDATION ***ERROR: Returned wrong fill value
ERROR: INCORRECT REAL VALIDATION ***ERROR: Returned wrong fill value
***ERROR: Returned wrong fill value
h5dclose_f FAILED
H5Pclose_f FAILED
Testing filling functions                                                             *FAILED*

comments?

brtnfld:
I'm sorry this took so long. This is now fixed in develop, and soon, 1.14.3 will be released. Can you verify the issue is fixed? I tested it on Mac with gfortran 13, and it did not return the errors you reported but crashed when optimization flags were added.

PR #3662, opened 2023-10-12
Fixes test failure for gfortran -O2 and -O3, -fdefault-real-16

brtnfld:
Seems higher optimization causes corruption, fixes GH #2928 .


epourmal (approved):
Looks fine, but I am puzzled why it needs to be reset?

epourmal:
> Looks fine, but I am puzzled why it needs to be reset?
Never mind, I saw your comment.

byrnHDF (approved):

PR #3664, opened 2023-10-12
Initial merge of examples repo into library

byrnHDF:


HDF5Examples/COPYING
jhendersonHDF, line 1: Probably not needed since we have the root one?

HDF5Examples/JAVA/intro/testfiles/116/HDF5DatasetCreate.txt
jhendersonHDF, line 1: There seem to be several empty files like this under this folder

brtnfld:
Did we change that much between versions so we can't use the same test reference files? Can't we ignore inconsequential differences?

HDF5Examples/COPYING
byrnHDF: No, we need it for the binaries - I use the HDF5Example folder as a standalone build with the binaries.

HDF5Examples/JAVA/intro/testfiles/116/HDF5DatasetCreate.txt
byrnHDF: By-product of being able to use HDF5Examples as a general solution for hdf5 versions (including develop) and API restrictions.

byrnHDF:
> Did we change that much between versions so we can't use the same test reference files? Can't we ignore inconsequential differences?

Not without a lot of development time.

PR #3668, opened 2023-10-12
Update RELEASE.txt

loricooperhdf:
Minor edits largely on two new (I think) blurbs, plus an added comma here and there.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -47,12 +47,12 @@ New Features
 
     Configuration:
     -------------
-    - Thread-safety + static library disabled on Windows w/ CMake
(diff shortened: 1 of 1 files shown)


release_docs/RELEASE.txt
```diff
@@ -778,7 +781,7 @@ CMake vs. Autotools installations
 =================================
 While both build systems produce similar results, there are differences.
 Each system produces the same set of folders on linux (only CMake works
-on standard Windows); bin, include, lib and share. Autotools places the
+on standard Windows); bin, include, lib, and share. Autotools places the
```
byrnHDF, line 784: I agree with the intention but I was informed a decade ago that there is not a comma before an and in a list of items.

byrnHDF (approved):

PR #3673, opened 2023-10-13
sync a couple of misses from merge

byrnHDF:


jhendersonHDF (approved):
Git tried its best it seems

lrknox (approved):

Issue #3458, opened 2023-08-31
clang-16: error: unsupported option '--sort-common' '--as-needed'


wreckdump:
**Describe the bug**
I am getting the following compilartion error during the compile process.

```
clang-16: error: unsupported option '--sort-common'
clang-16: error: unsupported option '--as-needed'
```

**Expected behavior**

I expect the source code to compile without an error. Here is the configuration that I have used

```  
./configure \
    --prefix=/usr \
    --disable-static \
    --enable-parallel \
    --enable-hl \
    --enable-build-mode=production \
    --enable-fortran \
    --enable-tests \
    --with-zlib= \
    --with-szlib= \
    CC=mpicc CXX=mpic++ FC=mpifort CFLAGS="-O3 -march=znver2 -mcmodel=large" CXXFLAGS="-O3 -march=znver2 -mcmodel=large" FCFLAGS="-O3 -march=znver2 -mcmodel=large"
    make -j4
```

**Platform (please complete the following information)**
 - HDF5 version : 1.14.2
 - OS and version : Arch Linux 6.4.12-arch1-1
 - Compiler and version : AMD clang version 16.0.3 (CLANG: AOCC_4.1.0-Build#270 2023_07_10)
 - Build system (e.g. CMake, Autotools) and version : GNU Make 4.4.1, autoconf 2.71-4
 - Any configure options you specified : Specified above
 - MPI library and version (parallel HDF5) : OpenMPI 4.1.5 compiled with AMD AOCC 4.1.0

**Additional context**

I have tried to suppress the `--sort-common` and `--as-needed` linking flags by specifiying `LDFLAGS = ""`, but it fails to compile with the following error.

```
clang-16: error: unknown argument: '-soname'
clang-16: error: no such file or directory: 'libhdf5_fortran.so.310'
```

With AMD AOCC 4.0.0 it would compile without any flag issues.

derobins:
Why is the linker using unsupported flags? We don't hard-code --as-needed or --sort-common in any of the build files, so it's getting set automatically at configure time. Are you building with a different linker than what's being detected at configure time? Ditto -soname.

I can try building with aocc 4.1, but it'd be helpful if you could configure with --disable-silent-rules, see what your system is using (`which ld` and `ld --version`), and peek at config.log to see what configure is using. You might also want to try CMake to see if that builds correctly and what version(s) that is using.

derobins:
Also, if you are using Flang instead of gfortran, we don't support that yet.

[...]

Issue #3439, opened 2023-08-28
AOCC compiler cannot build HDF5


lcebaman:
**Describe the bug**
AOCC compiler is not able to build HDF5

**Platform (please complete the following information)**
-HDF5 develop
-AOCC 4.0
-OpenMPI 4.1.5

**How was configured**

```
export FC=mpif77
export F90=mpif90
export F77=mpif77
export CXX=mpic++
export LD=mpif90
export MPIF90=mpif90
export OMPI_CC=clang
export OMPI_FC=flang
export OMPI_CXX=clang++
./configure  --enable-parallel --enable-fortran --enable-shared --enable-static
```

**Error**
```
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
make[1]: *** [Makefile:1979: cache] Error 1
make[1]: *** Waiting for unfinished jobs....
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
make[1]: *** [Makefile:1987: cache_image] Error 1
make[1]: *** [Makefile:1983: cache_api] Error 1
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
make[1]: *** [Makefile:1995: cache_tagging] Error 1
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
make[1]: *** [Makefile:2239: lheap] Error 1
make[1]: *** [Makefile:2283: ohdr] Error 1
clang: error: no such file or directory: '../src/.libs/libhdf5.so'
make[1]: *** [Makefile:2319: stab] Error 1
```

hyoklee:
Hi, @lcebaman !

Thank you for testing AOCC with HDF5.
I could confirm the build failure.

However, if you disable shared  & Fortran & parallel, you can build and test HDF5 with the latest AOCC.
Here's an example GitHub Action result that passes:

https://github.com/hyoklee/actions/actions/runs/6014286859/job/16313640693 

Here's my test Action workflow for the above result: 
https://github.com/hyoklee/actions/blob/main/.github/workflows/aocc.yml

lcebaman:
Hi @hyoklee 

Just to confirm, you were able to build HDF5 with AOCC but with Fortran, parallel and shared options disabled.

hyoklee:
@lcebaman Yes.

derobins:
[...]

Issue #1588, opened 2022-04-06
Error while compiling HDF5 using AOCC compilers (error while loading shared libraries: libomp.so)


Enea2301:
Good evening. I'm trying to compile HDF5 using latest AMD AOCC compilers (3.2), but a problem arises constantly while executing the "make" command

**./H5_buildiface: error while loading shared libraries: libomp.so: cannot open shared object file: No such file or directory
make[3]: *** [Makefile:1513: H5_gen.F90] Error 127
**

The runtime command for configure is as continues: CC="$CC" CXX="$CXX" FC="$FC" CFLAGS="$CFLAGS" CXXFLAGS="$CXXFLAGS" FCFLAGS="$FCFLAGS"  LDFLAGS="$LDFLAGS" LD_LIBRARY_PATH="$LD_LIBRARY_PATH" LIBRARY_PATH="$LIBRARY_PATH" ./configure --with-zlib=$DIR/grib2 --prefix=$DIR/hdf5 --enable-hl --enable-fortran

The referenced bashrc variables:
export DIR=path/to/librariesupperfolder
export DIRAOCC=path/to/AOCCdir
export CC=$DIRAOCC/AOCC/bin/clang
export CXX=$DIRAOCC/AOCC/bin/clang++
export FC=$DIRAOCC/AOCC/bin/flang
export CFLAGS="-m64 -O3 -march=znver2 -mtune=znver2"
export CXXFLAGS="-m64 -O3 -march=znver2 -mtune=znver2"
export FCFLAGS="-m64 -O3 -march=znver2 -mtune=znver2"
export LDFLAGS="-L$DIRAOCC/AOCC/lib -L$DIRAOCC/AOCC/lib32"
export LD_LIBRARY_PATH=$DIRAOCC/AOCC/lib:$DIRAOCC/AOCC/lib32/:$DIRAOCC/AOCC/ompd:/usr/lib64:/usr/lib/x86_64-linux-gnu:/usr/lib:/usr/lib32
export LIBRARY_PATH=$DIRAOCC/AOCC/lib:$DIRAOCC/AOCC/lib32/:/usr/lib64:/usr/lib/x86_64-linux-gnu:/usr/lib:/usr/lib32

Any ideas on why it could be happening?
libomp.so does exist under both ../AOCC/lib and ../AOCC/lib32

derobins:
[...]

Issue #366, opened 2021-02-21
clang-11: error: unknown argument: '-soname'


VI-gha:
This error occurs when installing HDF5 with AOCC-2.3 (derived from clang/flang) 
It makes it impossible to enable-shared: how to sort this out?

VI-gha:
@lrknox Can anybody help with that?

lrknox:
@VI-gha do you have logs with the error that give more detailed errors?  Other build details would also be useful.  We'll see if we can set up or find AOCC-2.3 on a system available to us, try to reproduce and find the problem and a fix for it.

derobins:
This is/was a libtool error. The libtool script that is generated at compile time was missing the "-Wl," prefix necessary for linker options, so "-Wl,-soname" gets parsed as -soname, which fails.

As of October 2023, with AOCC 4.1, c/flang works out of the box w/ shared unless you are using an MPI compiler wrapper. We have a hack for that (at least in the Autotools), which will appear shortly.

This issue will be closed when the PR is merged to develop.

derobins:
Fixed w/ #3674

scemama:
Here is a patch for libtool that should fix the problem:
```
$ git diff 1ec8fa28dcb29500d485c136db28315671ec4c3b
diff --git a/m4/libtool.m4 b/m4/libtool.m4
index 79a2451e..df1a7860 100644
--- a/m4/libtool.m4
+++ b/m4/libtool.m4
@@ -4719,6 +4719,12 @@ m4_if([$1], [CXX], [
        _LT_TAGVAR(lt_prog_compiler_pic, $1)='-fPIC'
        _LT_TAGVAR(lt_prog_compiler_static, $1)='-static'
         ;;
+      # ARM flang
+      armflang*)
+        _LT_TAGVAR(lt_prog_compiler_wl, $1)='-Wl,'
+        _LT_TAGVAR(lt_prog_compiler_pic, $1)='-fPIC'
+        _LT_TAGVAR(lt_prog_compiler_static, $1)='-static'
+        ;;
       # Lahey Fortran 8.1.
       lf95*)
        _LT_TAGVAR(lt_prog_compiler_wl, $1)='-Wl,'
```
Maybe as the HDF5 group you will have more influence than me to make the libtool developers include it ;-).

PR #3674, opened 2023-10-14
Add support for AOCC & Flang w/ the Autotools

derobins:
* Adds a config/clang-fflags options file to support Flang
* Corrects missing "-Wl," from linker options in the libtool wrappers when using Flang, the MPI Fortran compiler wrappers, and building the shared library. This would often result in unrecognized options like -soname.
* Enable -nomp w/ Flang to avoid linking to the OpenMPI library.

CMake can build the parallel, shared library w/ Fortran using AOCC and Flang, so no changes were needed for that build system.

Fixes GitHub issues #3458, #3439, #1588, #366, #280

### DIFF
--- a/config/clang-fflags
+++ b/config/clang-fflags
@@ -1,0 +1,139 @@
+#                            -*- shell-script -*-
+#
+# Copyright by The HDF Group.
+# All rights reserved.
+#
+# This file is part of HDF5.  The full HDF5 copyright notice, including
+# terms governing use, modification, and redistribution, is contained in
+# the COPYING file, which can be found at the root of the source code
+# distribution tree, or in https://www.hdfgroup.org/licenses.
+# If you do not have access to either file, you may request a copy from
+# help@hdfgroup.org.
+
+
+# This file should be sourced into configure if the compiler is the
+# Clang Fortran compiler (flang) or a derivative.  It is careful not to do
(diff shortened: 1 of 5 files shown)


config/clang-fflags
```diff
@@ -0,0 +1,139 @@
...
+    # Warnings #
+    ############
+
+    ###########
+    # General #
+    ###########
+
+    # We don't use OpenMP, so don't link to it
+    H5_FCFLAGS="$H5_FCFLAGS -nomp"
+
```
brtnfld, line 117: I don't see the option -nomp for flang, but it seems to accept it. I only see -fno-openmp-extensions

brtnfld:
What happens for llvm flang? Does it also get identified as classic flang?

derobins:
> What happens for llvm flang? Does it also get identified as classic flang?

No idea. I only tested with AOCC.

config/clang-fflags
derobins: AOCC 4 lists it in the manual.

https://www.amd.com/content/dam/amd/en/documents/pdfs/developer/aocc/aocc-v4.0-ga-user-guide.pdf

hyoklee (approved):

lrknox (approved):

derobins:
I'm going to go ahead and merge this. We can tweak as necessary if we discover places where we need more support.

PR #3677, opened 2023-10-15
Fix a strncpy call to use dest size not src

derobins:
A strncpy call in a path construction call used the size of the src buffer instead of the dest buffer as the limit n.

This was switched to use the dest size and properly terminate the string if truncation occurs.

### DIFF
--- a/src/H5system.c
+++ b/src/H5system.c
@@ -620,7 +620,6 @@ H5_build_extpath(const char *name, char **extpath /*out*/)
 
     FUNC_ENTER_NOAPI_NOINIT
 
-    /* Sanity check */
     assert(name);
     assert(extpath);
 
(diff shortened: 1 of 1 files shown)


derobins:
From skybridge:

In file included from /usr/include/string.h:633,
                 from /gpfs/lknox/dt-sky/CI/hdf5-1.14.3-1/src/H5private.h:35,
                 from /gpfs/lknox/dt-sky/CI/hdf5-1.14.3-1/src/H5system.c:30:
/gpfs/lknox/dt-sky/CI/hdf5-1.14.3-1/src/H5system.c: In function ‘H5_build_extpath’:
[CTest: warning matched] /gpfs/lknox/dt-sky/CI/hdf5-1.14.3-1/src/H5system.c:687:13: warning: ‘__builtin_strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
  687 |             strncpy(full_path, cwdpath, cwdlen + 1);
      |             ^~~~~~~
[CTest: warning matched] /gpfs/lknox/dt-sky/CI/hdf5-1.14.3-1/src/H5system.c:680:22: note: length computed here
  680 |             cwdlen = strlen(cwdpath);
      |                      ^~~~~~~~~~~~~~~

byrnHDF (approved):

qkoziol (approved):
Good catch!

Issue #3585, opened 2023-09-25
hdf5 error when cross compiling for arm64-windows


jimwang118:
**Describe the bug**
I am the maintainer of vcpkg and when I use vcpkg to install hdf5 on arm64-windows triplet I get an error.
```
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   HAVE_IOEO_EXITCODE (advanced)
For details see F:/upvcpkg/buildtrees/hdf5/arm64-windows-dbg/TryRunResults.cmake
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   H5_LDOUBLE_TO_LONG_SPECIAL_RUN (advanced)
   H5_LDOUBLE_TO_LONG_SPECIAL_RUN__TRYRUN_OUTPUT (advanced)
For details see F:/upvcpkg/buildtrees/hdf5/arm64-windows-dbg/TryRunResults.cmake
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   H5_LONG_TO_LDOUBLE_SPECIAL_RUN (advanced)
   H5_LONG_TO_LDOUBLE_SPECIAL_RUN__TRYRUN_OUTPUT (advanced)
For details see F:/upvcpkg/buildtrees/hdf5/arm64-windows-dbg/TryRunResults.cmake
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   H5_LDOUBLE_TO_LLONG_ACCURATE_RUN (advanced)
   H5_LDOUBLE_TO_LLONG_ACCURATE_RUN__TRYRUN_OUTPUT (advanced)
For details see F:/upvcpkg/buildtrees/hdf5/arm64-windows-dbg/TryRunResults.cmake
CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
   H5_LLONG_TO_LDOUBLE_CORRECT_RUN (advanced)
   H5_LLONG_TO_LDOUBLE_CORRECT_RUN__TRYRUN_OUTPUT (advanced)
For details see F:/upvcpkg/buildtrees/hdf5/arm64-windows-dbg/TryRunResults.cmake
```

**Expected behavior**
Installation is normal

**Platform (please complete the following information)**
 - HDF5 version (if building from a maintenance branch, please include the commit hash) 1.14.0
 - OS and version windows 10
 - Compiler and version  msvc
[...]

PR #3683, opened 2023-10-15
Add CMake long double cross-compile defaults

derobins:
HDF5 performs a couple of checks at build time to see if long double values can be converted correctly (IBM's Power architecture uses a special format for long doubles). These checks are performed using TRY_RUN, which is a problem when cross-compiling.

These checks now use default values appropriate for most non-Power systems when cross-compiling. The cache values can be pre-set if necessary, which will preempt both the TRY_RUN and the default.

Affected values:
    H5_LDOUBLE_TO_LONG_SPECIAL      (default no)
    H5_LONG_TO_LDOUBLE_SPECIAL      (default no)
    H5_LDOUBLE_TO_LLONG_ACCURATE    (default yes)
    H5_LLONG_TO_LDOUBLE_CORRECT     (default yes)
    H5_DISABLE_SOME_LDOUBLE_CONV    (default no)

Fixes GitHub #3585


config/cmake/ConfigureChecks.cmake
```diff
@@ -872,35 +872,46 @@ if (HDF5_BUILD_FORTRAN)
...
     else ()
-      set (${TEST} "" CACHE INTERNAL ${msg})
-      message (VERBOSE "${msg}... no")
-      file (APPEND ${CMAKE_BINARY_DIR}/CMakeFiles/CMakeError.log
-          "Test ${TEST} Compile failed with the following output:\n ${OUTPUT}\n"
-      )
+      # Use the default if there's no cache variable and cross-compiling
+      if ($def)
+        message (VERBOSE "${msg}... yes (cross-compile default)")
+        set (${TEST} 1 CACHE INTERNAL ${msg})
```
byrnHDF, line 909: Shouldn't you use the value of def? (Which should be ${def})

byrnHDF: > Shouldn't you use the value of def? (Which should be ${def})

In line 907

derobins: Fixed

mattjala (approved):

nikitalita:
Yo, while this may not fix windows-x86_64 to arm64 cross-compiling, this DOES fix macos arm64-to-x86_64 and vice-versa.

test repo with vcpkg with a portfile pointed at this PR:  https://github.com/nikitalita/hdf5_test/actions/runs/6860017162/job/18653170927

bmribler (approved):

hyoklee (approved):

PR #3684, opened 2023-10-16
Remove ROBODoc.

hyoklee:
Make header comment consistent.


byrnHDF (approved):

brtnfld (changes requested):
This is a header topic in Robdoc, not the file's name in this case.

byrnHDF:
Do we still need Robodoc support?

brtnfld:
For now, until it is replaced with Doxygen.

PR #3686, opened 2023-10-16
Convert nv workflow to callable

byrnHDF:


jhendersonHDF:
Autotools looks to be failing in tests as expected. Should likely do `export RUNPARALLEL="mpiexec -np 2"` in the testing section and the equivalent for cmake.

config/nvidia-flags
```diff
@@ -0,0 +1,122 @@
...
+    # Use this for profiling with gprof
+    #PROFILE_CFLAGS="-pg"
+
+    ################
+    # Optimization #
+    ################
+
+    HIGH_OPT_CFLAGS="-O4"
+    DEBUG_OPT_CFLAGS="-gopt -O2"
+    NO_OPT_CFLAGS="-O0"
```
derobins, line 110: Change HIGH and DEBUG to -O1 and comment that -O2+ currently has test failures.

jhendersonHDF: IIRC, debug builds work fine. It was specifically when -O2 or greater appears alongside NDEBUG being defined

lrknox (approved):

bmribler (approved):

derobins (approved):

PR #3693, opened 2023-10-17
Fix hangs during collective I/O with independent metadata writes

jhendersonHDF:


vchoi-hdfgroup (approved):

derobins (approved):

src/H5Cmpio.c
```diff
@@ -367,6 +386,13 @@ H5C_apply_candidate_list(H5F_t *f, H5C_t *cache_ptr, unsigned num_candidates, ha
     if (H5C__flush_candidate_entries(f, entries_to_flush, entries_to_clear) < 0)
         HGOTO_ERROR(H5E_CACHE, H5E_CANTFLUSH, FAIL, "flush candidates failed");
 
+    /* Restore collective I/O if we temporarily disabled it */
+    if (restore_io_mode) {
```
fortnern, line 390: Is this block necessary? Won't it get restored in the done section?

jhendersonHDF: It would without this block, but I want to restore it as early as possible in case further collective I/O happens past this point in the future

fortnern (approved):

PR #3698, opened 2023-10-17
Restore floating-point suffixes in tests

derobins:
A prior commit removed too many F suffixes. This restores the suffixes for float variables.


derobins:
Fixes warnings like this (from HPC CDash):

[hl/test/test_table.c:235](https://github.com/HDFGroup/hdf5/blob/master/hl/test/test_table.c#L235):38: warning: implicit conversion between floating point types of different sizes [-Wimplicit-float-size-conversion]

jhendersonHDF (approved):

hyoklee (approved):

PR #3699, opened 2023-10-17
Suppress format string warnings in subfiling test

derobins:


derobins:
Spotted in HPC CDash. Suppresses warnings like this:

[testpar/t_subfiling_vfd.c:394](https://github.com/HDFGroup/hdf5/blob/master/testpar/t_subfiling_vfd.c#L394):30: warning: format string is not a string literal [-Wformat-nonliteral]

hyoklee (approved):

jhendersonHDF (approved):

PR #3703, opened 2023-10-18
Correct use of set() command with options (#3667)

jhendersonHDF:
Merging this back from 1.14 before syncing 1.14 with develop again


derobins (approved):

derobins:
Merged. This already lives in 1.14 so this is basically a reverse merge.

PR #3704, opened 2023-10-18
Sync changes that are only in 1.14 branch

jhendersonHDF:


fortran/src/H5Pff.F90
```diff
@@ -4680,7 +4680,7 @@ SUBROUTINE h5pget_fill_value_ptr(prp_id, type_id, fillvalue, hdferr)
     IMPLICIT NONE
     INTEGER(HID_T), INTENT(IN)  :: prp_id
     INTEGER(HID_T), INTENT(IN)  :: type_id
-    TYPE(C_PTR)                 :: fillvalue
+    TYPE(C_PTR)   , INTENT(IN)  :: fillvalue
```
jhendersonHDF: @brtnfld Is this the correct way to go with this? In develop there is no marked intent; only in the 1.14 branch, which appears to have happened during a previous merge

brtnfld: It should have no intent parameter

derobins (approved):

PR #3705, opened 2023-10-18
Sync 1.14 branch with develop

jhendersonHDF:


java/src/jni/h5aImp.c
```diff
@@ -1186,7 +1186,7 @@ Java_hdf_hdf5lib_H5_H5AwriteVL(JNIEnv *env, jclass clss, jlong attr_id, jlong me
     UNUSED(clss);
 
     if (NULL == buf)
-        H5_NULL_ARGUMENT_ERROR(ENVONLY, "H5AwriteVL: write buffer is NULL");
+        H5_NULL_ARGUMENT_ERROR(ENVONLY, "H5Awrite: write buffer is NULL");
```
jhendersonHDF, line 1189: This is actually the wrong way to go since the function is called H5AwriteVL, but there are several other messages in develop that need updating, so this just makes the two branches consistent for now

derobins (approved):

PR #3708, opened 2023-10-18
GH issue 3398 - option to disable doxygen warn as error

byrnHDF:


CMakeLists.txt
```diff
@@ -952,6 +952,13 @@ if (HDF5_BUILD_DOC AND EXISTS "${HDF5_DOXYGEN_DIR}" AND IS_DIRECTORY "${HDF5_DOX
 # check if Doxygen is installed
   find_package(Doxygen)
   if (DOXYGEN_FOUND)
+    option (HDF5_ENABLE_DOXY_WARNINGS "Enable fail if doxygen parsing has warnings." ON)
+    mark_as_advanced (HDF5_USE_FOLDERS)
```
jhendersonHDF: HDF5_ENABLE_DOXY_WARNINGS

configure.ac
jhendersonHDF, line 1: Looks like CMake can support FAIL_ON_WARNINGS vs. NO, while Autotols would be NO vs. YES. Is that a big deal? Also, is it a problem if Autotools specifies "no" and "yes" vs "NO" and "YES" to doxygen?

byrnHDF:
Dang, I need a break!

configure.ac
byrnHDF: No it needs the correct text

byrnHDF: Updated

jhendersonHDF (approved):

configure.ac
```diff
@@ -1207,6 +1207,30 @@ AC_ARG_ENABLE([doxygen],
...
 AC_MSG_RESULT([$HDF5_DOXYGEN])
 
+## Check if they would like to enable doxygen warnings as errors
+##
+
+## This needs to be exposed for the library info file.
+AC_SUBST([HDF5_DOXY_WARNINGS])
+
+## Default is to not build DOXYGEN
+HDF5_DOXY_WARNINGS=no
```
derobins: I think you want DOXY_ERR here. Also, the comment is misleading (copy-paste error?).

release_docs/RELEASE.txt
```diff
@@ -47,6 +47,18 @@ New Features
 
     Configuration:
     -------------
+    - Added new options for CMake and autotools to control the Doxygen
```
derobins: Autotools is capitalized

release_docs/RELEASE.txt
```diff
@@ -47,6 +47,18 @@ New Features
...
+      warnings as errors setting.
+
+        * HDF5_ENABLE_DOXY_WARNINGS: ON/OFF (Default: ON)
+        * doxygen-errors: enable/disable (Default: enable)
```
derobins: Should be --enable-doxygen-errors

configure.ac
byrnHDF: done

release_docs/RELEASE.txt
byrnHDF: done

release_docs/RELEASE.txt
byrnHDF: done

derobins (approved):

PR #3710, opened 2023-10-18
Work around Theta system issue failure in links test

jhendersonHDF:
When the Subfiling VFD is enabled, the links test may try to initialize the Subfiling VFD and call MPI_Init_thread. On Theta, this appears to have an issue that will cause the links test to fail. Reworked the test to check for the same conditions in a more roundabout way that doesn't involved initializing the Subfiling VFD

### DIFF
--- a/test/links.c
+++ b/test/links.c
@@ -9880,6 +9880,7 @@ external_set_elink_cb(hid_t fapl, bool new_format)
     set_elink_cb_t       op_data, *op_data_p;
     H5L_elink_traverse_t cb;
     char                 filename1[NAME_BUF_SIZE], filename2[NAME_BUF_SIZE];
+    bool                 driver_is_parallel;
     unsigned             flags;
 
     if (new_format)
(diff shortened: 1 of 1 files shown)


test/links.c
```diff
@@ -9890,16 +9891,21 @@ external_set_elink_cb(hid_t fapl, bool new_format)
...
        would report false, causing problems */
-    base_driver = H5Pget_driver(fapl);
-    op_data.base_fapl =
-        (base_driver == H5FD_FAMILY || base_driver == H5FD_MULTI || base_driver == H5FD_MPIO ||
-         base_driver == H5FD_CORE || base_driver == H5FD_DIRECT || base_driver == H5FD_SUBFILING)
-            ? H5P_DEFAULT
-            : fapl;
+    op_data.base_fapl = fapl;
+    if (base_driver == H5FD_CORE || base_driver == H5FD_FAMILY || base_driver == H5FD_MULTI ||
+            base_driver == H5FD_DIRECT || driver_is_parallel)
```
jhendersonHDF: With the rework of the library initialization, these VFD macros are mapped to an initialization call now instead of just an ID value. Because of that, this code initializes the Subfiling VFD when it's enabled and that seems to cause issues on Theta. This should resolve the failure at https://my.cdash.org/test/98835800 without causing an initialization of the VFD, while still basically checking for the same conditions as before (except that any future parallel VFDs may need an exception in this test, but it should probably be reworked anyway).

test/links.c
```diff
@@ -18434,14 +18440,12 @@ link_info_by_idx_old(hid_t fapl)
...
-    char        objname[NAME_BUF_SIZE];  /* Object name */
-    char        valname[NAME_BUF_SIZE];  /* Link value name */
-    char        filename[NAME_BUF_SIZE]; /* File name */
+    unsigned    hard_link;                                               /* Create hard or soft link? */
+    H5L_info2_t linfo;                                                   /* Link info struct */
+    char        objname[NAME_BUF_SIZE];                                  /* Object name */
+    char        valname[NAME_BUF_SIZE];                                  /* Link value name */
+    char        filename[NAME_BUF_SIZE];                                 /* File name */
     H5O_token_t objtoken[CORDER_NLINKS]; /* Tokens (Addresses) of the objects created */
-    void       *vol_obj_file = NULL;     /* Object of file_id */
```
jhendersonHDF, line 18444: nvhpc saw that `f` and `vol_obj_file` are unused in these two functions

brtnfld (approved):

derobins (approved):

test/links.c
derobins: We really need to rethink library initialization, especially when it comes to VFDs.

PR #3711, opened 2023-10-18
Add NV flags and update CI

byrnHDF:
Add flag files for nvhpc compiler
Convert nvhpc CI to callable workflows
Update Testing scripts


CTestConfig.cmake
```diff
@@ -34,7 +34,7 @@ else ()
   if (CDASH_LOCAL)
     set (CTEST_DROP_LOCATION "/submit.php?project=HDF5114")
   else ()
-    set (CTEST_DROP_LOCATION "/submit.php?project=HDF5")
+    set (CTEST_DROP_LOCATION "/submit.php?project=HDF5114")
```
jhendersonHDF, line 37: Looks like these two lines for setting `CTEST_DROP_LOCATION` are exactly the same?

jhendersonHDF (approved):

CTestConfig.cmake
derobins: Unclear why we do things differently w/ CDASH_LOCAL in the first place

PR #3714, opened 2023-10-19
Fixes and adjustments to t_filters_parallel

jhendersonHDF:
Broadcast number of datasets to create in multi-dataset I/O cases so that interference with random number generation doesn't cause mismatches between ranks

Set fill time for datasets to "never" by default and adjust on a per-test basis to avoid writing fill values to chunks when it's unnecessary

Reduce number of loops run in some tests when performing multi-dataset I/O

Fix an issue in the "fill time never" test where data verification could fill if file space reuse causes application buffers to be filled with chosen fill value when reading from datasets with uninitialized storage

Skip multi-chunk I/O test configurations for multi-dataset I/O configurations when the TestExpress level is > 1 since those tests can be more stressful on the file system

### DIFF
--- a/testpar/t_filters_parallel.c
+++ b/testpar/t_filters_parallel.c
@@ -576,11 +576,21 @@ create_datasets(hid_t parent_obj_id, const char *dset_name, hid_t type_id, hid_t
         case USE_MULTIPLE_DATASETS:
         case USE_MULTIPLE_DATASETS_MIXED_FILTERED:
             dset_name_ptr = dset_name_multi_buf;
-            n_dsets       = (rand() % (MAX_NUM_DSETS_MULTI - 1)) + 2;
+
+            if (MAINPROCESS)
+                n_dsets = (rand() % (MAX_NUM_DSETS_MULTI - 1)) + 2;
+
+            if (mpi_size > 1)
+                VRFY((MPI_SUCCESS == MPI_Bcast(&n_dsets, 1, MPI_INT, 0, comm)), "MPI_Bcast succeeded");
 
             /* Select between 1 and (n_dsets - 1) datasets to be unfiltered */
             if (test_mode == USE_MULTIPLE_DATASETS_MIXED_FILTERED) {
-                n_unfiltered = (rand() % (n_dsets - 1)) + 1;
+                if (MAINPROCESS)
+                    n_unfiltered = (rand() % (n_dsets - 1)) + 1;
+
+                if (mpi_size > 1)
+                    VRFY((MPI_SUCCESS == MPI_Bcast(&n_unfiltered, 1, MPI_INT, 0, comm)),
+                         "MPI_Bcast succeeded");
 
                 unfiltered_dcpl = H5Pcopy(dcpl_id);
                 VRFY((unfiltered_dcpl >= 0), "H5Pcopy succeeded");
@@ -621,7 +631,11 @@ create_datasets(hid_t parent_obj_id, const char *dset_name, hid_t type_id, hid_t
              * remaining datasets as unfiltered datasets. Otherwise,
              * randomly determine if a dataset will be unfiltered.
              */
-            unfiltered = ((size_t)n_unfiltered == dsets_left) || ((rand() % 2) == 0);
+            if (MAINPROCESS)
+                unfiltered = ((size_t)n_unfiltered == dsets_left) || ((rand() % 2) == 0);
+
+            if (mpi_size > 1)
+                VRFY((MPI_SUCCESS == MPI_Bcast(&unfiltered, 1, MPI_C_BOOL, 0, comm)), "MPI_Bcast succeeded");
 
             if (unfiltered) {
                 curr_dcpl = unfiltered_dcpl;
@@ -1201,6 +1215,12 @@ test_write_filtered_dataset_no_overlap_partial(const char *parent_group, H5Z_fil
     plist_id = H5Pcopy(dcpl_id);
     VRFY((plist_id >= 0), "DCPL copy succeeded");
 
+    /*
+     * Since we're only doing a partial write to the dataset, make
+     * sure the fill time is set appropriately
+     */
+    VRFY((H5Pset_fill_time(plist_id, H5D_FILL_TIME_IFSET) >= 0), "H5Pset_fill_time succeeded");
+
     VRFY((H5Pset_chunk(plist_id, WRITE_UNSHARED_FILTERED_CHUNKS_PARTIAL_DATASET_DIMS, chunk_dims) >= 0),
          "Chunk size set");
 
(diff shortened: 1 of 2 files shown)


testpar/t_filters_parallel.c
```diff
@@ -576,11 +576,22 @@ create_datasets(hid_t parent_obj_id, const char *dset_name, hid_t type_id, hid_t
         case USE_MULTIPLE_DATASETS:
         case USE_MULTIPLE_DATASETS_MIXED_FILTERED:
             dset_name_ptr = dset_name_multi_buf;
-            n_dsets       = (rand() % (MAX_NUM_DSETS_MULTI - 1)) + 2;
+
+            if (MAINPROCESS)
+                n_dsets = (rand() % (MAX_NUM_DSETS_MULTI - 1)) + 2;
```
jhendersonHDF, line 581: On some systems, other software in the stack appears to be causing rand() calls to get out of sync between ranks, which is responsible for hangs on those systems during the multi-dataset I/O tests. Broadcast these values from rank to ensure they're consistent across every rank

testpar/t_filters_parallel.c
```diff
@@ -1551,7 +1575,12 @@ test_write_filtered_dataset_single_unlim_dim_no_overlap(const char *parent_group
         read_bufs[dset_idx] = tmp_buf;
     }
 
-    for (size_t i = 0; i < (size_t)WRITE_UNSHARED_ONE_UNLIM_DIM_NLOOPS; i++) {
+    /* Determine number of loops to run through */
+    num_loops = WRITE_UNSHARED_ONE_UNLIM_DIM_NLOOPS;
+    if ((test_mode == USE_MULTIPLE_DATASETS) || (test_mode == USE_MULTIPLE_DATASETS_MIXED_FILTERED))
+        num_loops /= 2;
```
jhendersonHDF, line 1578: Reduce the number of loops run in some of these tests for the multi-dataset I/O case since it just amplifies the amount of testing for no particular reason

testpar/t_filters_parallel.c
```diff
@@ -5250,7 +5311,6 @@ test_read_filtered_dataset_all_no_selection(const char *parent_group, H5Z_filter
     void       *read_bufs[MAX_NUM_DSETS_MULTI]    = {0};
     hsize_t     dataset_dims[READ_ALL_NO_SELECTION_FILTERED_CHUNKS_DATASET_DIMS];
     hsize_t     chunk_dims[READ_ALL_NO_SELECTION_FILTERED_CHUNKS_DATASET_DIMS];
-    hsize_t     sel_dims[READ_ALL_NO_SELECTION_FILTERED_CHUNKS_DATASET_DIMS];
```
jhendersonHDF, line 5253: unused

testpar/t_filters_parallel.c
```diff
@@ -9609,34 +9692,6 @@ test_fill_time_never(const char *parent_group, H5Z_filter_t filter_id, hid_t fap
...
-                                  recvcounts, displs, C_DATATYPE_MPI, comm);
-        VRFY((MPI_SUCCESS == mpi_code), "MPI_Allgatherv succeeded");
-
-        /*
-         * It should be very unlikely for the dataset's random
-         * values to all be the fill value, so this should be
-         * a safe comparison in theory.
-         */
-        VRFY((0 != memcmp(read_bufs[dset_idx], fill_buf, read_buf_size)), "Data verification succeeded");
-    }
```
jhendersonHDF, line 9638: Remove two data verification sections from this particular test. They try to ensure that the chosen fill value isn't returned from reads of a dataset with uninitialized storage, but file space reuse means that there's no real guarantee what comes back from the reads and these checks were occasionally failing on some systems

testpar/t_filters_parallel.c
```diff
@@ -9956,6 +10008,18 @@ main(int argc, char **argv)
...
 
+                        /*
+                         * If TestExpress is > 1, only run the multi-chunk I/O
+                         * configuration tests for the 'USE_SINGLE_DATASET' case,
+                         * as the 'USE_MULTIPLE_DATASETS' and 'USE_MULTIPLE_DATASETS_MIXED_FILTERED'
+                         * cases are more stressful on the file system.
+                         */
+                        if (test_express_level_g > 1) {
+                            if (((test_mode == USE_MULTIPLE_DATASETS) || (test_mode == USE_MULTIPLE_DATASETS_MIXED_FILTERED))
+                                    && (chunk_opt != H5FD_MPIO_CHUNK_ONE_IO))
```
jhendersonHDF: When the TestExpress level is > 1, skip testing of multi-chunk I/O in the multi-dataset I/O case since it tends to just hammer the filesystem and the multi_dset test already does a bit of testing of this case.

testpar/t_filters_parallel.c
```diff
@@ -9992,6 +10056,15 @@ main(int argc, char **argv)
                         VRFY((H5Pset_dxpl_mpio_chunk_opt(dxpl_id, chunk_opt) >= 0),
                              "H5Pset_dxpl_mpio_chunk_opt succeeded");
 
+                        /*
+                         * Disable writing of fill values by default. Otherwise, a
+                         * lot of time may be spent writing fill values to chunks
+                         * when they're going to be fully overwritten anyway.
+                         * Individual tests will alter this behavior as necessary.
+                         */
+                        VRFY((H5Pset_fill_time(dcpl_id, H5D_FILL_TIME_NEVER) >= 0),
```
jhendersonHDF, line 10063: Set the fill time to "never" by default so that we can avoid needlessly writing fill values out to chunks in some cases

lrknox:
Tested on perlmutter:  passed in 13 seconds now.

lrknox (approved):

jhendersonHDF:
@lrknox Made one last small change to turn off persistent file free space management so that we don't run into the occasional infinite loop in the library's free space management code. If it's not too much trouble, could you try running on Perlmutter again just to make sure nothing's really changed there? Going to try to get a reproducer for the infinite loop issue.

lrknox:
> @lrknox Made one last small change to turn off persistent file free space management so that we don't run into the occasional infinite loop in the library's free space management code. If it's not too much trouble, could you try running on Perlmutter again just to make sure nothing's really changed there? Going to try to get a reproducer for the infinite loop issue.

Sure.

lrknox:
> @lrknox Made one last small change to turn off persistent file free space management so that we don't run into the occasional infinite loop in the library's free space management code. If it's not too much trouble, could you try running on Perlmutter again just to make sure nothing's really changed there? Going to try to get a reproducer for the infinite loop issue.

Done.  still good - ran 39 test configurations:

All Parallel Filters tests passed - total test time was 10.150296 seconds
<end of output>
Test time =  44.65 sec
----------------------------------------------------------
Test Passed.
"MPI_TEST_t_filters_parallel" end time: Oct 19 14:42 PDT
"MPI_TEST_t_filters_parallel" time elapsed: 00:00:44
----------------------------------------------------------

derobins (approved):

PR #3716, opened 2023-10-19
Fix some issues with collective metadata reads for chunked datasets

jhendersonHDF:
Add functions/callbacks for explicit control over chunk index open/close

Add functions/callbacks to check if chunk index is open or not so that it can be opened if necessary before temporarily disabling collective metadata reads in the library

Add functions/callbacks for requesting loading of additional chunk index metadata beyond the chunk index itself


jhendersonHDF:
This PR is meant to address the issues found and reported on in section 3.2 of 
[HDF5_Collective_IO_Performance_Investigation.pdf](https://github.com/HDFGroup/hdf5/files/13038066/HDF5_Collective_IO_Performance_Investigation.pdf) and also reported in #2658. It's unclear if the independent reads in #2658 are specifically from the extensible array structure, header and/or index block, but these changes fix collective reading of all those, so it should at least help there.

src/H5Dbtree.c
```diff
@@ -959,14 +1031,34 @@ H5D__btree_idx_get_addr(const H5D_chk_idx_info_t *idx_info, H5D_chunk_ud_t *udat
...
+ *
+ * Purpose:     Load additional chunk index metadata beyond the chunk index
+ *              itself. Currently a no-op.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5D__btree_idx_load_metadata(const H5D_chk_idx_info_t H5_ATTR_UNUSED *idx_info)
```
fortnern, line 1046: Should this load the root node?

src/H5Dbtree2.c
```diff
@@ -1003,16 +1076,36 @@ H5D__bt2_idx_get_addr(const H5D_chk_idx_info_t *idx_info, H5D_chunk_ud_t *udata)
...
+ * Purpose:     Load additional chunk index metadata beyond the chunk index
+ *              itself. Currently a no-op.
  *
- * Purpose:	Translate the B-tree specific chunk record into a generic
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5D__bt2_idx_load_metadata(const H5D_chk_idx_info_t H5_ATTR_UNUSED *idx_info)
```
fortnern, line 1083: Same comment about loading root note

src/H5Dbtree2.c
```diff
@@ -854,7 +920,10 @@ H5D__bt2_idx_insert(const H5D_chk_idx_info_t *idx_info, H5D_chunk_ud_t *udata,
     assert(H5_addr_defined(udata->chunk_block.offset));
 
     /* Check if the v2 B-tree is open yet */
-    if (NULL == idx_info->storage->u.btree2.bt2) {
+    if (H5D__bt2_idx_is_open(idx_info, &index_open) < 0)
```
fortnern: I'm a little nervous about potential function call overhead here (and other places that call this function within this file). It might get inlined, but maybe we should create a macro here to be safe?

src/H5Dbtree2.c
jhendersonHDF: Hmm, yes looks like I should probably add a lookup here for the root node.

src/H5Dbtree.c
jhendersonHDF: For v1 btrees, it looks like the search starts from metadata that will already be in the cache as a consequence of opening the index.

src/H5Dbtree2.c
jhendersonHDF: I'd be very surprised if there was any overhead at all, but I could create a macro that the is_open function calls as well.

src/H5Dearray.c
```diff
@@ -953,7 +1016,10 @@ H5D__earray_idx_insert(const H5D_chk_idx_info_t *idx_info, H5D_chunk_ud_t *udata
     assert(udata);
 
     /* Check if the extensible array is open yet */
-    if (NULL == idx_info->storage->u.earray.ea) {
+    if (H5D__earray_idx_is_open(idx_info, &index_open) < 0)
```
fortnern: Same comment about function call overhead

src/H5Dfarray.c
```diff
@@ -901,7 +963,10 @@ H5D__farray_idx_insert(const H5D_chk_idx_info_t *idx_info, H5D_chunk_ud_t *udata
     assert(udata);
 
     /* Check if the fixed array is open yet */
-    if (NULL == idx_info->storage->u.farray.fa) {
+    if (H5D__farray_idx_is_open(idx_info, &index_open) < 0)
```
fortnern: Same comment about function call overhead

src/H5Dbtree.c
fortnern: So the root node is loaded collectively?

jhendersonHDF: Yes, at least for v1 btrees the metadata needed should already be loaded collectively after the "index open" call

src/H5Dbtree2.c
fortnern: It might take too much time to get in for 1.14.3 since I think you'd need to add a H5B2 function, but something to keep in mind for later

jhendersonHDF: It looks like performing a bt2 `get_addr` call will have the same effect as earray and farray where in this case a fake chunk lookup will make sure the root node is read in collectively. We should eventually update the interfaces behind the chunk indices to have a more explicit way of controlling this in the future though.

fortnern:
Other than the is_open macros this looks good to me. We should keep in mind the idea of collective pre-loading of the root node as a potential performance enhancement

src/H5Dbtree.c
fortnern: H5D__btree_idx_open() is a no-op. Does it happen in some other function?

fortnern (approved):

src/H5Dbtree.c
jhendersonHDF: Ah yes sorry, on second look it doesn't occur as part of the index open call, it seems that the v1 B-tree root node exists directly at the file address for the chunk index and so no open really occurs. This may need to be revisited, but v1 btrees seem to be less straightforward than other index types

src/H5Dbtree2.c
fortnern: This will also open other some internal nodes and the leaf node for chunk 0, which is honestly fine.

fortnern: Though since there's multiple steps (need to read each node to get the address of the node further down) it could be wasted overhead if no process actually needs any data on that "side" of the dataset. But again, something to keep in mind for later.

src/H5Dbtree.c
fortnern: I think you could  implement it the same way as the other indices with an address query of chunk 0 (though see my notes on the b-tree 2 implementation)

src/H5Dbtree2.c
fortnern: Taking this further, we could, after collectively loading only the root node, collectively load every node at the next level down in a single operation, then again at the next level, and so on (up to a limit that maybe could be set depending on the size of the metadata cache and the average node size)

fortnern: To prime the metadata cache in the most efficient way possible

jhendersonHDF: Agreed that without having explicit control over loading the root node, there may be some wasted effort in loading the data for that chunk if no one needs it. I suppose the hope is that if the user has collective metadata reads enabled, the benefit of having the root node loaded at scale will be more worth it. That and the ranks that are interested in the chunk won't end up reading it independently.

derobins (approved):

PR #3723, opened 2023-10-19
Test scripts now execute in-source with creation of tmp dir

byrnHDF:
The tool scripts create a tmp folder for testing and then can delete it without worry since all the testfiles are copies.


jhendersonHDF:
There still seem to be some

`ls: cannot access '/home/runner/work/hdf5/hdf5/tools/test/h5dump/../../test/h5dump/testfiles/pbits': No such file or directory`

lines in the Autotools workflows

byrnHDF:
Found issue with path warnings - fixed.

derobins (approved):

jhendersonHDF (approved):

PR #3728, opened 2023-10-19
Suppress cast-qual warning in H5TB Fortran wrapper

derobins:
This interface is fundamentally broken, const-wise.

### DIFF
--- a/hl/fortran/src/H5TBfc.c
+++ b/hl/fortran/src/H5TBfc.c
@@ -92,10 +92,12 @@ h5tbmake_table_c(size_t_f *namelen1, _fcd name1, hid_t_f *loc_id, size_t_f *name
     /*
      * call H5TBmake_table function.
      */
+    H5_GCC_CLANG_DIAG_OFF("cast-qual")
     if (H5TBmake_table(c_name1, (hid_t)*loc_id, c_name, c_nfields, (hsize_t)*nrecords, (size_t)*type_size,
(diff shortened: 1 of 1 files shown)


derobins:
https://stackoverflow.com/questions/14562845/why-does-passing-char-as-const-char-generate-a-warning

byrnHDF (approved):

jhendersonHDF (approved):
Fine for now, though we should eventually go fix all these. The interface is correct; the code isn't technically

brtnfld:
It's probably best to call the C API from Fortran and eliminate the broken code.

jhendersonHDF:
> It's probably best to call the C API from Fortran and eliminate the broken code.

The code can also be fixed easily, the const interface just makes it a little more awkward to interface with and be const-correct

brtnfld:
Are we talking about the  C API  sig. itself or the use of it in the C wrapper?

jhendersonHDF:
> Are we talking about the C API sig. itself or the use of it in the C wrapper?

The code that uses it in the wrapper. The C API signature is correct as far as const goes. We're passing in an array of non-modifiable pointers. In fact, `const char * const *field_names` would be an even better signature since the function doesn't need to modify the entries in `field_names` either, but that'd be even more awkward to deal with.

derobins:
Yeah, I'm not super happy with this, but we have a couple of places where we use `const foo**` and it causes problems and this is how we paper over it. We should definitely tweak the API's consts in a future major release.

PR #3732, opened 2023-10-20
Add testing of Subfiling VFD with zlib compression

hyoklee:
1. Help user to find the directory when test fails.
2. Test zlib compression (OESS-388).

Fixes OESS-388

### DIFF
--- a/testpar/t_subfiling_vfd.c
+++ b/testpar/t_subfiling_vfd.c
@@ -100,9 +100,7 @@ static void test_subfiling_precreate_rank_0(void);
 static void test_subfiling_write_many_read_one(void);
 static void test_subfiling_write_many_read_few(void);
 static void test_subfiling_h5fuse(void);
-#ifdef H5_HAVE_FILTER_DEFLATE
 static void t_zlib(void);
-#endif
 
 static test_func tests[] = {test_create_and_close,
                             test_ioc_only_fail,
@@ -113,12 +111,8 @@ static test_func tests[] = {test_create_and_close,
                             test_subfiling_precreate_rank_0,
                             test_subfiling_write_many_read_one,
                             test_subfiling_write_many_read_few,
-                            test_subfiling_h5fuse
-#ifdef H5_HAVE_FILTER_DEFLATE
-                            ,
-                            t_zlib
-#endif
-};
+                            test_subfiling_h5fuse,
+                            t_zlib};
 
 /* ---------------------------------------------------------------------------
  * Function:    create_subfiling_ioc_fapl
@@ -2121,7 +2115,6 @@ parse_subfiling_env_vars(void)
     }
 }
 
-#ifdef H5_HAVE_FILTER_DEFLATE
 /*
  * Test zlib.
  */
@@ -2131,6 +2124,10 @@ parse_subfiling_env_vars(void)
 static void
 t_zlib(void)
 {
+    if (MAINPROCESS)
+        TESTING_2("write zlib & read w/ single MPI rank");
+
+#ifdef H5_HAVE_FILTER_DEFLATE
     hsize_t start[1];
     hsize_t count[1];
     hsize_t dset_dims[1];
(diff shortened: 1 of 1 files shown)


testpar/t_subfiling_vfd.c
```diff
@@ -100,18 +100,24 @@ static void test_subfiling_precreate_rank_0(void);
...
+                            test_ioc_only_fail,
+                            test_config_file,
+                            test_stripe_sizes,
+                            test_selection_strategies,
+                            test_read_different_stripe_size,
+                            test_subfiling_precreate_rank_0,
+                            test_subfiling_write_many_read_one,
+                            test_subfiling_write_many_read_few,
+                            test_subfiling_h5fuse
+#ifdef H5_HAVE_FILTER_DEFLATE
```
byrnHDF: Just add the test to list, and check in the test proper and print "SKIPPED deflate filter not available"

hyoklee: I want to reduce output message from CMake. I also want to skip building it to speed up compilation.

byrnHDF: Testing requires consistency in order to compare results across a variety of configurations and platforms. 
I doubt adding the build of a single output statement is going to take that much time.

#ifdef has deflate
 full test
#else
 print SKIPPED - blah blah
#endif

testpar/t_subfiling_vfd.c
```diff
@@ -2116,6 +2121,159 @@ parse_subfiling_env_vars(void)
...
+    curr_nerrors = nerrors;
+
+    dxpl_id = H5Pcreate(H5P_DATASET_XFER);
+    VRFY((dxpl_id >= 0), "DXPL creation succeeded");
+    VRFY((H5Pset_dxpl_mpio(dxpl_id, H5FD_MPIO_COLLECTIVE) >= 0), "H5Pset_dxpl_mpio() succeeded");
+
+    /* Set selection I/O mode on DXPL */
+    VRFY((H5Pset_selection_io(dxpl_id, H5D_SELECTION_IO_MODE_ON) >= 0), "H5Pset_selection_io succeeded");
+
+    if (MAINPROCESS)
```
byrnHDF: Why isn't this first to be executed?

byrnHDF (changes requested):

testpar/t_subfiling_vfd.c
hyoklee: Good catch - I don't know.
I just copied from the existing function.
@jhendersonHDF , do you know why?

jhendersonHDF: The above code was added by someone else. Probably best to move the MAINPROCESS prints up everywhere. I can do that after this is merged.

testpar/t_subfiling_vfd.c
```diff
@@ -100,19 +100,19 @@ static void test_subfiling_precreate_rank_0(void);
...
+                            test_ioc_only_fail,
+                            test_config_file,
+                            test_stripe_sizes,
+                            test_selection_strategies,
+                            test_read_different_stripe_size,
+                            test_subfiling_precreate_rank_0,
+                            test_subfiling_write_many_read_one,
+                            test_subfiling_write_many_read_few,
+                            test_subfiling_h5fuse,
+                            t_zlib};
```
brtnfld: should be test_zlib for consistency

testpar/t_subfiling_vfd.c
```diff
@@ -2116,6 +2115,162 @@ parse_subfiling_env_vars(void)
...
+    buf = NULL;
+
+    VRFY((H5Pclose(plist_id) >= 0), "H5Pclose() succeeded");
+    VRFY((H5Dclose(dset_id) >= 0), "Dataset close succeeded");
+    VRFY((H5Fclose(file_id) >= 0), "File close succeeded");
+
+    mpi_code_g = MPI_Barrier(comm_g);
+    VRFY((mpi_code_g == MPI_SUCCESS), "MPI_Barrier succeeded");
+
+    if (MAINPROCESS) {
```
brtnfld, line 2221: Can you also add a test for reading from all the ranks? I think you can just add it to this test.

brtnfld:
Can we get an update on this PR? Where are we at getting it into 1.14.3?

brtnfld:
Adding an extra parameter to each function that turns on/off chunking/compression would be better than copying one of the previous test cases. That way compression is tested for all the cases currently tested for.

testpar/t_subfiling_vfd.c
hyoklee: I added it.

hyoklee:
> Adding an extra parameter to each function that turns on/off chunking/compression would be better than copying one of the previous test cases. That way compression is tested for all the cases currently tested for.

Not for 1.14.3 but for 1.14.4.

brtnfld:
Replaced by #3769

PR #3738, opened 2023-10-20
Add intel oneapi windows build to CI CMake

byrnHDF:
OneAPI on windows works by disabling Fortran and excluding a test with rounding issues for floats.


byrnHDF:
Only the nvhpc check fails here - probably doesn't have the latest nvhpc changes?

jhendersonHDF:
> Only the nvhpc check fails here - probably doesn't have the latest nvhpc changes?

Seems to be due to something else. Missing symbols at linking time

Issue #3144, opened 2023-06-16
ICX command line warning for hdf5


HathewayWill:
icx: command line warning #10430: Unsupported command line options encountered
These options as listed are not supported.
For more information, use '-qnextgen-diag'.
option list: 
	-Wp64

see attached files

`  ############################# HDF5 library for NetCDF4 & parallel functionality ############################

  cd $WRF_FOLDER/Downloads
  tar -xvzf hdf5-1_14_1-2.tar.gz
  cd hdf5-hdf5-1_14_1-2

  CC=$MPICC FC=$MPIFC CXX=$MPICXX F90=$MPIF90 F77=$MPIF77 CFLAGS=$CFLAGS ./configure --prefix=$DIR/grib2 --with-zlib=$DIR/grib2 --enable-hl --enable-fortran --enable-parallel

  make -j $CPU_HALF_EVEN 2>&1 | tee hdf5.make.log
  make -j $CPU_HALF_EVEN install 2>&1 | tee hdf5.makeinstall.log

  # other libraries below need these variables to be set
  export HDF5=$DIR/grib2
  export LD_LIBRARY_PATH=$DIR/grib2/lib:$LD_LIBRARY_PATH
  export PATH=$HDF5/bin:$PATH
  export PHDF5=$DIR/grib2

  echo " "

`

[config.log](https://github.com/HDFGroup/hdf5/files/11764309/config.log)
[hdf5.make.log](https://github.com/HDFGroup/hdf5/files/11764310/hdf5.make.log)
[hdf5.makeinstall.log](https://github.com/HDFGroup/hdf5/files/11764311/hdf5.makeinstall.log)

byrnHDF:
The Intel warnings files have not been updated since version 18 (5 years ago) and before Intel moved to OneAPI.
The CompilerFlags files also are likely to need updates and maybe distinguish between intel classic and IntelLLVM compilers.

derobins:
This should be fixed w/ #3751, which splits the Intel configuration files into 'classic' and oneAPI.

PR #3751, opened 2023-10-22
Update Autotools to correctly configure oneAPI

derobins:
Splits the Intel config files under the Autotools into 'classic' Intel and oneAPI versions, fixing 'unsupported option' messages.

Also turns off `-check uninit` (new in 2023) in Fortran, which kills the H5_buildiface program due to false positives.

### DIFF
--- a/config/apple
+++ b/config/apple
@@ -55,29 +55,18 @@ fi
 # Figure out C compiler flags
 . $srcdir/config/gnu-flags
 . $srcdir/config/clang-flags
+. $srcdir/config/oneapi-flags
 . $srcdir/config/intel-flags
-
-# temp patch: if GCC 4.2.1 is used in Lion or Mountain Lion systems, do not
-# use -O option as it causes failures in test/dt_arith.
-case "$host_os" in
-    darwin1[12].*)   # lion & mountain lion
-        #echo cc_vendor=$cc_vendor'-'cc_version=$cc_version
-        case "$cc_vendor-$cc_version" in
-            gcc-4.2.1)
-                # Remove any -O flags
-                #echo PROD_CFLAGS=$PROD_CFLAGS
-                PROD_CFLAGS="`echo $PROD_CFLAGS | sed -e 's/-O[0-3]*//'`"
-                #echo new PROD_CFLAGS=$PROD_CFLAGS
-                ;;
-        esac
-        ;;
-esac
 
 if test "X-" = "X-$FC"; then
     case $CC_BASENAME in
         gcc*)
             FC=gfortran
             FC_BASENAME=gfortran
+            ;;
+        icx*)
+            FC=ifx
+            FC_BASENAME=ifx
             ;;
         icc*)
             FC=ifort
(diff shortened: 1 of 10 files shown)


derobins:
Fixes #3144

derobins:
I tried getting Fortran to work, but there are linker errors (it works fine locally).

Maybe we need to use something like this?

https://github.com/fortran-lang/setup-fortran

config/apple
```diff
@@ -55,30 +55,19 @@ fi
...
-        case "$cc_vendor-$cc_version" in
-            gcc-4.2.1)
-                # Remove any -O flags
-                #echo PROD_CFLAGS=$PROD_CFLAGS
-                PROD_CFLAGS="`echo $PROD_CFLAGS | sed -e 's/-O[0-3]*//'`"
-                #echo new PROD_CFLAGS=$PROD_CFLAGS
-                ;;
-        esac
-        ;;
-esac
```
derobins, line 74: Not only is this ancient, we recently fixed the underlying problem.

lrknox (approved):

config/oneapi-fflags
```diff
@@ -0,0 +1,145 @@
...
+    ##############
+
+    PROD_FCFLAGS=
+
+    #########
+    # Debug #
+    #########
+
+    # Don't use -check uninit or you'll get false positives from H5_buildiface
+    DEBUG_FCFLAGS="-check all,nouninit"
```
byrnHDF, line 95: Will this need to be added in the CMake config file, too?

byrnHDF (approved):

brtnfld (approved):

config/freebsd
```diff
@@ -57,8 +64,11 @@ fi
 # Figure out FORTRAN compiler flags
 . $srcdir/config/gnu-fflags
 
-# Figure out Intel F90 compiler flags
-. $srcdir/config/intel-fflags
+# Figure out Intel oneAPI FC compiler flags
+. $srcdir/config/oneapi-fflags
+
+# Figure out Intel classic FC compiler flags
+. $srcdir/config/classic-fflags
```
balay, line 71: This file doesn't exist in repo
```
balay@pj01:~/tmp/hdf5$ git co -q a6d1bda6d2d 
balay@pj01:~/tmp/hdf5$ git ls-files |grep classic-fflags
balay@pj01:~/tmp/hdf5$ 
```
On fbsd with `hdf5-1.14.3`
```
checking for config freebsd12.1... no
checking for config freebsd... found
compiler '/home/svcpetsc/petsc-hash-pkgs/39f577/bin/mpicc' is GNU gcc-9.2.0
compiler '/home/svcpetsc/petsc-hash-pkgs/39f577/bin/mpif90' is GNU gfortran-9.2.0.: cannot open ./config/classic-fflags: No such file or directory

                    Output in config.log for HDF5: This file contains any messages produced by compilers while
running configure, to aid debugging if configure makes a mistake.
```

balay: Perhaps the following fix? 

https://github.com/HDFGroup/hdf5/pull/3883

Issue #3381, opened 2023-08-14
Reduce number of S3 requests on file open in ros3 VFD


ajelenak:
When opening HDF5 files with the ros3 VFD, the VFD always makes something like the four separate S3 requests below for tiny amount of bytes at the beginning of the file:

```
GET: Bytes 0 - 7, Request Size: 8
GET: Bytes 0 - 15, Request Size: 16
GET: Bytes 16 - 95, Request Size: 80
GET: Bytes 96 - 607, Request Size: 512
```

Would it be possible to have ros3 VFD make a single request for a larger amount of bytes on file open and cache it somehow to satisfy multiple libhdf5 data reads? The request size could be configurable. Something like 16 MiB default could be really helpful because a lot of useful information can be packed at the beginning of an HDF5 file (thinking about cloud-optimized HDF5 files).

derobins:
Fixed in #3753, but we may want to expand on that in the future.

PR #3753, opened 2023-10-23
Add a simple cache to the ros3 VFD

derobins:
Adds a small cache of the first N bytes of a file opened with the read-only S3 (ros3) VFD, where N is 4kiB or the size of the file, whichever is smaller. This avoids a lot of small I/O operations on file open.

Addresses GitHub issue #3381

### DIFF
--- a/src/H5FDros3.c
+++ b/src/H5FDros3.c
@@ -43,6 +43,9 @@
  */
 #define ROS3_STATS 0
 
+/* Max size of the cache, in bytes */
+#define ROS3_MAX_CACHE_SIZE 4096
+
 /* The driver identification number, initialized at runtime
  */
 static hid_t H5FD_ROS3_g = 0;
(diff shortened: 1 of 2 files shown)


src/H5FDros3.c
```diff
@@ -1666,41 +1686,50 @@ H5FD__ros3_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_UNU
...
+        if (H5FD_s3comms_s3r_read(file->s3r_handle, addr, size, buf) == FAIL)
+            HGOTO_ERROR(H5E_VFL, H5E_READERROR, FAIL, "unable to execute read");
 
 #if ROS3_STATS
 
-    /* Find which "bin" this read fits in. Can be "overflow" bin.  */
-    for (bin_i = 0; bin_i < ROS3_STATS_BIN_COUNT; bin_i++)
-        if ((unsigned long long)size < ros3_stats_boundaries[bin_i])
-            break;
-    bin = (type == H5FD_MEM_DRAW) ? &file->raw[bin_i] : &file->meta[bin_i];
```
derobins, line 1687: This stuff just got moved over because it's in an else clause now. I figured we only wanted to track stats when we did an actual read.

jhendersonHDF (approved):

jhendersonHDF:
It would be good to address #3381's point about allowing the cache size to be configured at runtime at some point in the future before closing that issue.

mattjala (approved):

Issue #3025, opened 2023-05-28
MPI I/O hints are not preserved after closing the file access property list


wkliao:
MPI Info object was populated by the underneath MPI-IO library in PR #1636.
However, the hints are not preserved after the file access property list is closed.
Calling `H5Fget_access_plist()` and `H5Pget_fapl_mpio()` returns MPI_INFO_NULL.
Below is a test program. It is tested using the develop branch and 1.14.1-2.

```
#include <stdio.h>
#include <stdlib.h>
#include <mpi.h>
#include <hdf5.h>

int main (int argc, char **argv)
{
    herr_t err;
    hid_t file_id, fapl_id;
    MPI_Info info, info_used;

    MPI_Init(&argc, &argv);
    MPI_Info_create(&info);
    MPI_Info_set(info, "romio_no_indep_rw", "true");
 
    fapl_id = H5Pcreate(H5P_FILE_ACCESS);
    H5Pset_fapl_mpio(fapl_id, MPI_COMM_WORLD, info);
    file_id = H5Fcreate("dummy.h5", H5F_ACC_TRUNC, H5P_DEFAULT, fapl_id);
    H5Pclose(fapl_id);

    fapl_id = H5Fget_access_plist(file_id);
    H5Pget_fapl_mpio(fapl_id, NULL, &info_used);

    if (info_used == MPI_INFO_NULL) printf("Error: info is MPI_INFO_NULL\n");

    H5Pclose(fapl_id);
    H5Fclose(file_id);
    MPI_Info_free(&info);

    MPI_Finalize();
    return 0;
}
```

hyoklee:
@derobins merged #1636.

fortnern:
Summary: 1636 changed the library to retrieve MPI Info from MPI after opening the file, which may be different from the Info passe to MPI. This retrieved info is, however, only stored on the FAPL, so it is lost when the FAPL is closed. We should either modify the library to cache this info on the file struct or modify H5Fget_access_plist to retrieve the Info from MPI

PR #3755, opened 2023-10-23
Preserve MPI-I/O file hints when fapl is closed

vchoi-hdfgroup:
Save the MPI info in the file struct so H5Fget_access_plist() can retrieve it from there.

Fixes GitHub #3025


src/H5Fint.c
jhendersonHDF, line 1: Since the library calls `H5_mpi_info_dup` when getting MPI info from a FAPL, I believe this means the library will need to close the `MPI_Info` struct with `H5_mpi_info_free` when closing an `H5F_shared_t` structure now.

src/H5Fint.c
```diff
@@ -1414,6 +1413,14 @@ H5F__dest(H5F_t *f, bool flush, bool free_on_failure)
             f->shared->efc = NULL;
         } /* end if */
 
+#ifdef H5_HAVE_PARALLEL
+        if (f->shared->mpi_info != MPI_INFO_NULL) {
```
jhendersonHDF, line 1423: I would add a line in `H5F__new` to explicitly set `f->shared->mpi_info` to `MPI_INFO_NULL` when the shared structure is allocated, just in case we fail before the added line that sets this field and we try to call `H5_mpi_info_free` on uninitialized memory. Otherwise, this PR looks good to me.

vchoi-hdfgroup: done

jhendersonHDF (approved):

testpar/t_file.c
```diff
@@ -1119,3 +1119,63 @@ test_evict_on_close_parallel_unsupp(void)
...
+    herr_t      ret;     /* Generic return value */
+    int         mpi_ret; /* MPI return value */
+
+    filename = (const char *)GetTestParameters();
+
+    /* set up MPI parameters */
+    mpi_ret = MPI_Info_create(&info);
+    VRFY((mpi_ret >= 0), "MPI_Info_create succeeded");
+
+    mpi_ret = MPI_Info_set(info, "hdf_info_fapl", "true");
```
fortnern: Would you be able to test MPI_Info_get() after retrieving the info to make sure it still has the same value?

fortnern: Other than that, looks good to me. I'm fine with merging it to get it into 1.14 and adding this line to the test after if needed.

fortnern (approved):

testpar/t_file.c
jhendersonHDF: Agreed on checking to make sure that the hint is still the same after retrieving the prop list

testpar/t_file.c
```diff
@@ -1165,6 +1173,21 @@ test_fapl_preserve_hints(void)
...
+    VRFY((mpi_ret == MPI_SUCCESS), "MPI_Info_get_nthkey succeeded");
+
+    mpi_ret = MPI_Info_get(info_used, key_used, 20, value_used, &flag);
+    VRFY((mpi_ret == MPI_SUCCESS), "MPI_Info_get succeeded");
+
+    ret = strcmp(key_used, key);
+    VRFY(ret == 0, "strcmp key");
+
+    ret = strcmp(value_used, value);
+    VRFY(ret == 0, "strcmp value");
```
jhendersonHDF: I believe these changes are fine, but there's a possibility this test may fail depending on whether MPI keys come exactly in the order they were set on the Info object. Otherwise, the MPI I/O VFD has code to get any info set by the MPI implementation after an MPI_File_open call and then set that back on the FAPL (https://github.com/HDFGroup/hdf5/blob/develop/src/H5FDmpio.c#L892-L934)

jhendersonHDF: It may be best to call MPI_Info_get_nkeys and then loop through all the keys, checking to see if any of them match the hint we set here and then failing if none match

vchoi-hdfgroup: I will change that based on MPI_Info_get_nkeys.

vchoi-hdfgroup: Done

PR #3772, opened 2023-10-25
Cancel running GitHub workflows on push to same PR

jhendersonHDF:
Uses github.ref for the concurrency group for workflows when run as part of a PR (github.head_ref is available) so that previously-triggered workflows are cancelled when a new commit is pushed to a PR. Otherwise, uses github.run_id when commits are pushed to develop so that this shouldn't cancel things when multiple commits are merged to develop at close times.

### DIFF
--- a/.github/workflows/cve.yml
+++ b/.github/workflows/cve.yml
@@ -16,7 +16,7 @@ on:
 
 # Using concurrency to cancel any in-progress job or run
 concurrency:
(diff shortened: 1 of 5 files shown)


bmribler (approved):

brtnfld (approved):

jhendersonHDF:
I believe that this is the direction we want to go with the current setup we have for workflows, but if this starts cancelling things in develop again I'll need to revisit. The goal is to make it so that repeated pushes to a PR cancel previous workflows so that we don't starve resources from other PRs.

PR #3776, opened 2023-10-25
Fix a format string warning in the C++ examples

derobins:


derobins:
Shows up in HPC CDash output

hl/c++/examples/ptExampleFL.cpp
```diff
@@ -72,7 +72,7 @@ main(void)
     if (err < 0)
         fprintf(stderr, "Error getting packet count.");
 
-    printf("Number of packets in packet table after five appends: %llu\n", count);
+    printf("Number of packets in packet table after five appends: %" PRIdHSIZE "\n", count);
```
derobins: count is hsize_t and should use the format specifier in H5public.h

jhendersonHDF: Should this be PRIuHSIZE instead?

derobins: Yes. Good catch!

jhendersonHDF (approved):
We should merge this after https://github.com/HDFGroup/hdf5/actions/runs/6646446878/job/18059890906 finishes to see if the next push to develop picks up the openmpi cache

lrknox (approved):

PR #3781, opened 2023-10-26
Don't build util tests when HDF5_EXTERNALLY_CONFIGURED=ON

### LINKED ISSUE #3780: Can't compile with `HDF5_EXTERNALLY_CONFIGURED=ON` (cmake)
**Describe the bug**
Compilation fails when using `HDF5_EXTERNALLY_CONFIGURED=ON` due to the `test` subdirectory being correctly excluded, but `utils/test` being incorrectly included. `swmr_check_compat_vfd.c` includes `h5test.h` and thus depends on the `test` directory having been added.

**Additional context**
A workaround is to explicitly set `BUILD_TESTING` to `OFF`. However, given that the mechanism with `HDF5_EXTERNALLY_CONFIGURED` exists, it should be taken into account at this point as well.

Krzmbrzl:
Fixes #3780

### DIFF
--- a/utils/CMakeLists.txt
+++ b/utils/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.18)
 project (HDF5_UTILS C)
 
-if (BUILD_TESTING)
+if (NOT HDF5_EXTERNALLY_CONFIGURED AND BUILD_TESTING)
   add_subdirectory (test)
 endif ()
 
--- a/c++/CMakeLists.txt
+++ b/c++/CMakeLists.txt
@@ -13,6 +13,6 @@ endif ()
 #-----------------------------------------------------------------------------
 # Build the CPP unit tests
 #-----------------------------------------------------------------------------
-if (BUILD_TESTING)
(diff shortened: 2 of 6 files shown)


utils/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.18)
 project (HDF5_UTILS C)
 
-if (BUILD_TESTING)
+if (NOT HDF5_EXTERNALLY_CONFIGURED AND BUILD_TESTING)
   add_subdirectory (test)
 endif ()
```
Krzmbrzl, line 6: I think one could also think of only building the tests if the utils are built in the first place. Building only the tests doesn't seem to make much sense :thinking: :eyes:

byrnHDF (changes requested):
We should guard all the add_subdirectory test folders, not just an arbitrary instance.

Krzmbrzl:
I think I added guards for all other tests as well :eyes:

byrnHDF (approved):

lrknox (approved):

jhendersonHDF (approved):

brtnfld (approved):

PR #3782, opened 2023-10-27
Rework MPI Info FAPL preserve PR to use VFD 'ctl' operations

jhendersonHDF:
This doesn't fix the test failure with nvhpc and intel classic, but brings the changes more in line with how we retrieve the MPI_Comm object currently by using the VFD 'ctl' callback. The test failure with nvhpc goes away with an optimization level of -O1 or lower, so I still suspect that's due to an nvhpc problem and not something in the library. It appears that Nvidia also does or was building HDF5 with -O1 from https://forums.developer.nvidia.com/t/help-configuring-hdf5-with-nvhpc-no-version-information-available/183413.

### DIFF
--- a/src/H5FDmpi.c
+++ b/src/H5FDmpi.c
@@ -143,7 +143,7 @@ done:
 } /* end H5FD_mpi_get_comm() */
 
 /*-------------------------------------------------------------------------
- * Function:    H5FD_mpi_get_comm
+ * Function:    H5FD_mpi_get_info
  *
  * Purpose:     Retrieves the file's MPI_Info info object
  *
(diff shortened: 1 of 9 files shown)


src/H5FDmpi.c
```diff
@@ -143,6 +142,45 @@ H5FD_mpi_get_comm(H5FD_t *file)
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5FD_mpi_get_comm() */
 
+/*-------------------------------------------------------------------------
+ * Function:    H5FD_mpi_get_comm
+ *
```
vchoi-hdfgroup, line 147: H5FD_mpi_get_info

byrnHDF (approved):

fortnern (approved):

PR #3785, opened 2023-10-27
Print some messages in parallel tests on MPI rank 0 only

jhendersonHDF:


brtnfld:
sameshape also has an issue with all ranks printing "Test filenames are:"

The "For help use:" message gets print on all ranks.

t_mpi also has all ranks print, "MPIO GB file write test" and read test, each also prints the hostname, not sure that is needed.

jhendersonHDF:
> sameshape also has an issue with all ranks printing "Test filenames are:"
> 
> The "For help use:" message gets print on all ranks.
> 
> t_mpi also has all ranks print, "MPIO GB file write test" and read test, each also prints the hostname, not sure that is needed.

Some of the output is difficult to control because the parallel framework uses stuff that was only designed for serial, but I should be able to fix these

brtnfld (approved):

derobins (approved):

PR #3787, opened 2023-10-27
Fix misc. warnings from GCC when compiling with -fsanitize=undefined

jhendersonHDF:


tools/src/misc/h5repart.c
```diff
@@ -227,6 +227,10 @@ main(int argc, char *argv[])
     if (argno >= argc)
         usage(prog_name);
     src_gen_name = argv[argno++];
+    if (!src_gen_name) {
+        fprintf(stderr, "invalid source file name pointer");
+        exit(EXIT_FAILURE);
+    }
```
derobins, line 233: Even though we are just exiting anyway, valgrind is going to complain we leak the name memory (allocated above) if we ever thoroughly test the command-ilne options options.

derobins (approved):

byrnHDF (approved):

lrknox (approved):

PR #3793, opened 2023-10-28
Don't install h5tools_test_utils test program on system

jhendersonHDF:
This is just a test program for some tools utility functions and shouldn't be installed


derobins (approved):

derobins:
If this was publicly deployed, we should include a release note about how this was in error and it's gone now.

jhendersonHDF:
> If this was publicly deployed, we should include a release note about how this was in error and it's gone now.

Will do

byrnHDF (approved):

PR #3800, opened 2023-10-30
Set NVHPC maximum optimization level to -O1 for now

jhendersonHDF:
Compiling HDF5 with NVHPC 23.5 - 23.9 results in test failures in 4 different test files that need to be resolved. Since those tests pass with an optimization level of -O1 (and -O0) and it is currently unclear whether the test failures are due to issues in HDF5 or issues in the 'nvc' compiler, set the maximum optimization level for NVHPC to -O1 until the test failures are resolved.

### DIFF
--- a/config/cmake/HDFCompilerFlags.cmake
+++ b/config/cmake/HDFCompilerFlags.cmake
@@ -56,6 +56,44 @@ if (CMAKE_C_COMPILER_ID STREQUAL "NVHPC" )
   else ()
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Mbounds -g")
   endif ()
+
+  # With at least NVHPC 23.5 - 23.9, compiling with -O2 or higher and -DNDEBUG
+  # appears to have issues that manifest in the tests as incorrect metadata
+  # checksums being read or memory being corrupted. Compiling without -DNDEBUG
+  # does not appear to have these issues, but is not ideal due to compiling in
+  # asserts and other library debug code. Compiling with -O1 also does not appear
+  # to have these issues, so set maximum optimization level to -O1 for now until
+  # it can be determined whether these issues are compiler-specific or issues
+  # in the library.
+  set (cmake_c_flags_minsizerel_edited "${CMAKE_C_FLAGS_MINSIZEREL}")
+  string (REPLACE "-O2" "" cmake_c_flags_minsizerel_edited "${cmake_c_flags_minsizerel_edited}")
+  string (REPLACE "-O3" "" cmake_c_flags_minsizerel_edited "${cmake_c_flags_minsizerel_edited}")
+  string (REPLACE "-O4" "" cmake_c_flags_minsizerel_edited "${cmake_c_flags_minsizerel_edited}")
+  string (REPLACE "-Ofast" "" cmake_c_flags_minsizerel_edited "${cmake_c_flags_minsizerel_edited}")
+  string (REPLACE "-fast" "" cmake_c_flags_minsizerel_edited "${cmake_c_flags_minsizerel_edited}")
+  string (STRIP "${cmake_c_flags_minsizerel_edited}" cmake_c_flags_minsizerel_edited)
+  string (PREPEND cmake_c_flags_minsizerel_edited "-O1 ")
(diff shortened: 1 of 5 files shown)


jhendersonHDF:
The MPI I/O FAPL preserve PR test failure still exists on some systems until I can figure out what's going on there, but these changes pass locally with NVHPC 23.5 - 23.9 and on at least one system that isn't having the FAPL preserve test failure

config/nvidia-flags
```diff
@@ -106,7 +107,8 @@ if test "X-nvc" = "X-$cc_vendor" -o "X-nvcc" = "X-$cc_vendor"; then
     ################
 
     HIGH_OPT_CFLAGS="-O1"         # -O2+ currently has test failures.
```
jhendersonHDF, line 109: Looks like we already did this previously for Autotools builds, which is why the NVHPC action passes tests for Autotools builds, but the tests were disabled for CMake builds

byrnHDF (approved):

jhendersonHDF:
Looks like this is hitting a separate failure in testphdf5 and also a new one on JUnit-TestH5D. More investigation needed..

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -56,6 +56,44 @@ if (CMAKE_C_COMPILER_ID STREQUAL "NVHPC" )
...
   endif ()
+
+  # With at least NVHPC 23.5 - 23.9, compiling with -O2 or higher and -DNDEBUG
+  # appears to have issues that manifest in the tests as incorrect metadata
+  # checksums being read or memory being corrupted. Compiling without -DNDEBUG
+  # does not appear to have these issues, but is not ideal due to compiling in
+  # asserts and other library debug code. Compiling with -O1 also does not appear
+  # to have these issues, so set maximum optimization level to -O1 for now until
+  # it can be determined whether these issues are compiler-specific or issues
+  # in the library.
```
brtnfld, line 67: Do we have a keyword that we can add to this to indicate we should revisit the issue at a later date?

brtnfld (approved):
lgtm

config/cmake/HDFCompilerFlags.cmake
jhendersonHDF: I thought about creating an issue to track this, but open to other suggestions as well.

brtnfld: @derobins , we probably need to develop a procedure indicating workarounds that should be revisited at a future date. Otherwise, they will be forgotten about.

jhendersonHDF:
Actually, the Autotools NVHPC workflow doesn't seem to be getting any flags from `nvidia-flags` at all:

```
Languages:
----------
                              C: yes
                     C Compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/23.7/comm_libs/openmpi4/bin/mpicc
                       CPPFLAGS: 
                    H5_CPPFLAGS: -D_GNU_SOURCE -D_POSIX_C_SOURCE=200809L   -DNDEBUG -UH5_DEBUG_API -I/home/runner/work/hdf5/hdf5/src/H5FDsubfiling
                    AM_CPPFLAGS: 
                        C Flags: 
                     H5 C Flags:       
                     AM C Flags: 
               Shared C Library: yes
               Static C Library: yes
                        Fortran: yes
               Fortran Compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/23.7/comm_libs/openmpi4/bin/mpifort
                  Fortran Flags: -fPIC -fortranlibs
               H5 Fortran Flags:      
               AM Fortran Flags: 
         Shared Fortran Library: yes
         Static Fortran Library: yes
               Module Directory: ${includedir}
                            C++: no
                           Java: no
```

jhendersonHDF:
The testphdf5 failure has been fixed now, but there's still a segfault in JUnit-TestH5D that I haven't seen anywhere else

lrknox:
> The testphdf5 failure has been fixed now, but there's still a segfault in JUnit-TestH5D that I haven't seen anywhere else

Java fails in nvhpc-cmake.yml test.  The nvhpc-auto.yml test doesn't enable java.  We suggest disabling java for the cmake test also and creating an issue for java with nvhpc.  The failing test is the java test of H5Pset_fapl_multi (muti).

lrknox (approved):

PR #3804, opened 2023-10-30
Add processing of NVHPC flags in linux-gnulibc1 file

jhendersonHDF:
Pull in NVHPC flags from `nvidia-flags`, `nvidia-fflags` and `nvidia-cxxflags`

### DIFF
--- a/.github/workflows/nvhpc-auto.yml
+++ b/.github/workflows/nvhpc-auto.yml
@@ -67,11 +67,11 @@ jobs:
 
       # RUN TESTS
       # NORMAL
-      - name: Autotools Run Tests
-        run: |
-          export PATH=/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/comm_libs/openmpi4/bin:/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/compilers/bin:$PATH
-          make check -j
-        working-directory: ${{ runner.workspace }}/build
+#      - name: Autotools Run Tests
+#        run: |
+#          export PATH=/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/comm_libs/openmpi4/bin:/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/compilers/bin:$PATH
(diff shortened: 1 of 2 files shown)


jhendersonHDF:
I'm expecting this will fail in the Autotools NVHPC action since, without #3800, it will add in `-fast` to the compile options from `nvidia-flags`, which implies `-O2`.

With these changes, I can see that Autotools builds with NVHPC now get `-c99 -Minform=warn -fast  -s   -O1` in `H5_CFLAGS`

jhendersonHDF:
Looks like this is indeed now failing in the same places as CMake with the flags being pulled in correctly. I've disabled the testing in this PR and then will add re-enabling it back to #3800, though it appears that needs more work to pass testing.

.github/workflows/nvhpc-auto.yml
```diff
@@ -67,11 +67,11 @@ jobs:
...
-      - name: Autotools Run Tests
-        run: |
-          export PATH=/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/comm_libs/openmpi4/bin:/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/compilers/bin:$PATH
-          make check -j
-        working-directory: ${{ runner.workspace }}/build
+#      - name: Autotools Run Tests
+#        run: |
+#          export PATH=/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/comm_libs/openmpi4/bin:/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/compilers/bin:$PATH
+#          make check -j
+#        working-directory: ${{ runner.workspace }}/build
```
derobins, line 74: Why are the checks commented out?

byrnHDF (approved):

.github/workflows/nvhpc-auto.yml
lrknox: See previous comment.

lrknox (approved):

PR #3806, opened 2023-10-31
Fix memory corruption in 'MPI I/O FAPL preserve' test

jhendersonHDF:
Change `key_used` variable to be size `MPI_MAX_INFO_KEY + 1` (extra byte just in case, since the OpenMPI standard seems a bit unclear on the true max size) to fix possible buffer overflow due to it being smaller than `MPI_MAX_INFO_KEY`. Change `value_used` variable to be size `MPI_MAX_INFO_VAL + 1` and allocate it on the heap since `MPI_MAX_INFO_VAL` could be large.

### DIFF
--- a/testpar/t_file.c
+++ b/testpar/t_file.c
@@ -1128,27 +1128,26 @@ test_evict_on_close_parallel_unsupp(void)
 void
 test_fapl_preserve_hints(void)
 {
-    hid_t       fid     = H5I_INVALID_HID; /* HDF5 file ID */
-    hid_t       fapl_id = H5I_INVALID_HID; /* File access plist */
     const char *filename;
-
-    int  nkeys_used;
-    bool same = false;
-
-    MPI_Info    info  = MPI_INFO_NULL;
-    const char *key   = "hdf_info_fapl";
-    const char *value = "xyz";
-
-    MPI_Info info_used = MPI_INFO_NULL;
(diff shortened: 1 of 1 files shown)


testpar/t_file.c
```diff
@@ -1184,16 +1183,15 @@ test_fapl_preserve_hints(void)
...
+        memset(value_used, 0, MPI_MAX_INFO_VAL + 1);
 
         /* Get the nth key */
         mpi_ret = MPI_Info_get_nthkey(info_used, i, key_used);
         VRFY((mpi_ret == MPI_SUCCESS), "MPI_Info_get_nthkey succeeded");
 
         if (!strcmp(key_used, key)) {
-
-            mpi_ret = MPI_Info_get(info_used, key_used, 20, value_used, &flag);
+            mpi_ret = MPI_Info_get(info_used, key_used, MPI_MAX_INFO_VAL, value_used, &flag);
```
jhendersonHDF, line 1194: The standard says the value length specified here should be 1 less than the number of bytes allocated to make room for the NUL terminator

jhendersonHDF:
This fixes the test issue on Summit and I'm hoping it will fix the same issue on other systems that are currently failing.

derobins (approved):

lrknox (approved):

PR #3821, opened 2023-11-01
Merge CMake updates from 1.12 development branch

byrnHDF:


CTestConfig.cmake
```diff
@@ -34,7 +34,7 @@ else ()
   if (CDASH_LOCAL)
     set (CTEST_DROP_LOCATION "/submit.php?project=HDF5112")
   else ()
-    set (CTEST_DROP_LOCATION "/submit.php?project=HDF5")
+    set (CTEST_DROP_LOCATION "/submit.php?project=HDF5112")
```
lrknox, line 37: Test scripts on HPC machines sending results to my.cdash.org have to override this (HDF5112 to HDF5), otherwise the results just disappear.  It's annoying, but we can live with it.

lrknox (approved):

PR #3825, opened 2023-11-02
Create test for H5Pget_dxpl_mpio

glennsong09:
Created a test for H5Pget_dxpl_mpio in testpar to ensure method works correctly.

### DIFF
--- a/testpar/t_ph5basic.c
+++ b/testpar/t_ph5basic.c
@@ -192,14 +192,14 @@ test_fapl_mpio_dup(void)
 void
 test_get_dxpl_mpio(void)
 {
-    hid_t            fid           = H5I_INVALID_HID;
-    hid_t            sid           = H5I_INVALID_HID;
-    hid_t            did           = H5I_INVALID_HID;
-    hid_t            dxpl_id       = H5I_INVALID_HID;
-    hsize_t          dims[2]       = {100, 100};
+    hid_t            fid     = H5I_INVALID_HID;
+    hid_t            sid     = H5I_INVALID_HID;
+    hid_t            did     = H5I_INVALID_HID;
+    hid_t            dxpl_id = H5I_INVALID_HID;
+    hsize_t          dims[2] = {100, 100};
     hsize_t          i, j;
     H5FD_mpio_xfer_t xfer_mode;
-    int              *data_g       = NULL;
+    int             *data_g = NULL;
 
     if (VERBOSE_MED)
         printf("Verify get_fxpl_mpio correctly gets the data transfer mode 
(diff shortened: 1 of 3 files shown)


testpar/t_ph5basic.c
```diff
@@ -177,3 +177,119 @@ test_fapl_mpio_dup(void)
         VRFY((mrc == MPI_SUCCESS), "MPI_Info_free");
     }
 } /* end test_fapl_mpio_dup() */
+
+/*-------------------------------------------------------------------------
+ * Function:    test_get_dxpl_mpio
+ *
+ * Purpose:     Test that H5Pget_fxpl_mpio will properly return the data
```
jhendersonHDF: H5Pget_fxpl_mpio -> H5Pget_dxpl_mpio

testpar/t_ph5basic.c
```diff
@@ -177,3 +177,119 @@ test_fapl_mpio_dup(void)
...
+    H5FD_mpio_xfer_t xfer_mode;
+    hsize_t          dims[2] = {100, 100};
+    hsize_t          i, j;
+    int             *data = NULL;
+    int              mpi_rank, mpi_size;
+    const char      *filename;
+    herr_t           ret;
+
+    if (VERBOSE_MED)
+        printf("Verify get_fxpl_mpio correctly gets the data transfer mode"
```
jhendersonHDF: H5Pget_fxpl_mpio -> H5Pget_dxpl_mpio

testpar/t_ph5basic.c
```diff
@@ -177,3 +177,119 @@ test_fapl_mpio_dup(void)
...
+    ret = H5Pclose(fapl);
+    VRFY((ret >= 0), "H5Pclose succeeded");
+
+    ret = H5Pclose(dxpl);
+    VRFY((ret >= 0), "H5Pclose succeeded");
+
+    ret = H5Dclose(did);
+    VRFY((ret >= 0), "H5Dclose succeeded");
+
+    ret = H5Sclose(sid) < 0;
```
jhendersonHDF: Looks like the comparison to 0 here and below might be accidental leftovers

jhendersonHDF:
It would be nice to also check the value after a call to H5Dread in both cases (collective vs. independent), but otherwise this looks reasonable

testpar/t_ph5basic.c
glennsong09: Oh, let me fix that.

testpar/t_ph5basic.c
glennsong09: Let me change these.

glennsong09:
> It would be nice to also check the value after a call to H5Dread in both cases (collective vs. independent), but otherwise this looks reasonable

I can add that.

jhendersonHDF (approved):

brtnfld (approved):
lgtm

PR #3826, opened 2023-11-02
Implement selection vector I/O with collective chunk filling

vchoi-hdfgroup:
Also fix a bug in H5FD__mpio_write_vector() to account for fixed size optimization when computing max address.

Addresses ECP-344


src/H5Dchunk.c
```diff
@@ -5763,48 +5609,24 @@ H5D__chunk_collective_fill(const H5D_t *dset, H5D_chunk_coll_fill_info_t *chunk_
...
-    if (MPI_BYTE != mem_type)
-        if (MPI_SUCCESS != (mpi_code = MPI_Type_free(&mem_type)))
-            HMPI_DONE_ERROR(FAIL, "MPI_Type_free failed", mpi_code)
-    H5MM_xfree(chunk_disp_array);
-    H5MM_xfree(block_disps);
-    H5MM_xfree(block_lens);
+    H5MM_xfree(io_addrs);
+    H5MM_xfree(io_wbufs);
+    if (!all_same_block_len)
+        H5MM_xfree(io_sizes);
```
fortnern: Minor comment: I'd prefer to just always call H5MM_xfree here to make the code more robust against future changes. It's a no-op if io_sizes in NULL so this won't cause problems

src/H5FDmpio.c
```diff
@@ -2473,8 +2473,13 @@ H5FD__mpio_write_vector(H5FD_t *_file, hid_t H5_ATTR_UNUSED dxpl_id, uint32_t co
             HGOTO_ERROR(H5E_VFL, H5E_CANTGET, FAIL, "can't build MPI datatypes for I/O");
 
         /* Compute max address written to */
-        if (count > 0)
-            max_addr = s_addrs[count - 1] + (haddr_t)(s_sizes[count - 1]);
+        /* Bug: need to account for fixed size optimization */
+        if (count > 0) {
```
fortnern: A few things here:
- If count == 1 this reads off the end of the s_sizes array (maybe just change the condition to count > 1)
- The sizes can actually end anywhere in the array, not just at index 1. I'd recommend checking the last element to see if it's 0, then scanning from index 1 and up until you see a 0. So maybe something like:
`if (count > 1) {`
`    uint32_t idx = count - 1;`
`    if (s_sizes[idx] == 0)`
`        for (idx = 0; idx < count - 1; idx++)`
`            if (s_sizes[idx + 1] == 0)`
`                 break;`
`    max_addr = s_addrs[count - 1] + (haddr_t)(s_sizes[idx])`
`}`

fortnern: Also the comment shouldn't proclaim it as a bug, just explain what it's doing

fortnern: Actually my code is wrong too, I'll try to fix it later tonight. We might need to change the H5FD__mpio_vector_build_types function

vchoi-hdfgroup: I think <s_sizes> returned from H5FD__mpio_vector_build_types() can be (1) or (2) as follows: 

1. <count> entries of sizes
2. Two entries for same size optimization: entry 0 is the size and entry 1 is 0
Maybe the better way is to change H5FD__mpio_vector_build_types().

fortnern: It can be any number of entries between 1 (with a 0 terminator if count > 1) and  count

fortnern: If we have time it would be great to add a test for this case. We would have to use the public H5FD interface

src/H5Dchunk.c
derobins: The C standard explicitly says that calling free() on NULL pointers is acceptable.

vchoi-hdfgroup: Fixed

src/H5FDmpio.c
vchoi-hdfgroup: Applied the patch from Neil.  
I will work on adding a test for this case.

fortnern (approved):

jhendersonHDF:
The code looks fine and is much simpler, but it does remove the part where the old code divided up the chunks among all the MPI ranks. The current approach will have every MPI rank writing all of the chunks out. Some MPI implementations may optimize for this, but I think we ran into the case where this doesn't necessarily happen and so implemented the rank 0 bcast feature. I'd suggest we should either have only rank 0 write out the chunks and all the other ranks should set their I/O sizes to 0 (and maybe set the bufs to NULL and addrs to HADDR_UNDEF) or try to put back the logic of dividing up the work. Ideally the latter since rank 0 only writes may not scale well.

jhendersonHDF (changes requested):
See comment for discussion

vchoi-hdfgroup:
> .

I will put back the logic of dividing the work like the original routine.

vchoi-hdfgroup:
I will work on putting back the logic of dividing up the work like the original H5D__chunk_collective_fill() routine

src/H5Dchunk.c
```diff
@@ -5603,156 +5607,73 @@ H5D__chunk_collective_fill(const H5D_t *dset, H5D_chunk_coll_fill_info_t *chunk_
 
     /* Check if we have any chunks to write on this rank */
     if (num_blocks > 0 || (leftover && leftover > mpi_rank)) {
```
fortnern: Minor nitpick with pre-existing code - the first leftover is redundant since mpi_rank will never be less than 0. This could be `if (num_blocks > 0 || leftover > mpi_rank) {`

src/H5Dchunk.c
```diff
@@ -5603,156 +5607,73 @@ H5D__chunk_collective_fill(const H5D_t *dset, H5D_chunk_coll_fill_info_t *chunk_
...
-        if (!all_same_block_len)
-            if (NULL == (block_lens = (int *)H5MM_malloc((size_t)(blocks + 1) * sizeof(int))))
-                HGOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "couldn't allocate chunk lengths buffer");
+        if (chunk_fill_info->chunk_info[i].chunk_size != chunk_fill_info->chunk_info[i - 1].chunk_size)
+            all_same_block_len = false;
+    }
 
-        for (i = 0; i < (size_t)blocks; i++) {
-            size_t idx = i + (size_t)(mpi_rank * blocks);
+    if (need_sort)
```
fortnern, line 5638: Could you add some more comments to the blocks that don't have them, explaining what the code is doing?

fortnern: Also, for this block in particular, add something like this note (I just discussed this with Scot):

Note that we sort all of the chunks here, and not just a subset corresponding to this rank. We do this since we have found MPI I/O to work better when each rank writes blocks that are contiguous in the file, and by sorting the full list we maximize the chance of that happening.

testpar/t_filters_parallel.c
```diff
@@ -556,10 +556,20 @@ verify_chunk_opt_status(size_t num_dsets, test_mode_t test_mode, bool any_io, bo
              * elements (because no elements were raw data), which is what happens when performing I/O on a
              * filtered dataset with no selection. Vector I/O does report an I/O call was made if passed a raw
              * data element of size 0, so this is consistent. */
+            /* Changes: for allocation, the reported I/O is vector I/O */
```
fortnern: No need to call out "changes" in the code comment here. Just explain why we expect the different modes under each condition as  the code stands with this PR

testpar/t_filters_parallel.c
```diff
@@ -592,15 +602,14 @@ verify_chunk_opt_status(size_t num_dsets, test_mode_t test_mode, bool any_io, bo
                          * should be scalar I/O for allocation in addition to vector I/O for the actual data.
                          * If we're reading from an unallocated dataset then there should be no actual I/O.
                          * Otherwise there should only be vector I/O. */
-                        if (did_alloc)
-                            VRFY((H5D_SCALAR_IO | H5D_VECTOR_IO) == actual_sel_io_mode_reduced,
-                                 "verified actual selection I/O mode was scalar and vector I/O");
-                        else if (unalloc_read)
+                        /* Changes: for allocation, the reported I/O is vector I/O */
```
fortnern: Same here. Explain that both collective filling/allocation and filtered I/O use vector I/O so only that is reported whether or not allocation happened.

src/H5Dchunk.c
jhendersonHDF: It seems like we could potentially be making extra work for ourselves now by sorting here since the MPI I/O VFD is already scanning the chunk list to check for this anyway and we don't necessarily need to impose the monotonically non-decreasing file displacement requirement on other parallel VFDs. That said, since the sort has to happen at some point anyway if the addresses are out of order, I'm not sure if the above note is really needed here.

jhendersonHDF:
@vchoi-hdfgroup Thanks for putting pack the logic for dividing up the chunks among the ranks. This looks good to me and I don't really have anything to add to @fortnern's comments.

jhendersonHDF (approved):

src/H5Dchunk.c
vchoi-hdfgroup: Done with this and subsequent comments.

fortnern (approved):

fortnern:
Looks good to me.

PR #3835, opened 2023-11-07
HDF5 API test updates

jhendersonHDF:
Removed test duplication from bringing API tests
back into the library from external VOL tests
repo

Synced changes between API tests and library's
tests

Updated API tests CMake code to directly use and
install testhdf5, testphdf5, etc. instead of
creating duplicate binaries

Added new h5_using_native_vol() test function to
determine whether the VOL connector being used
is (or the VOL connector stack being used resolves to)
the native VOL connector


jhendersonHDF:
This PR removes the remaining test duplication from bringing the external VOL tests back into the library. I'm working on GitHub actions to test the external connectors, but since there are so many changes for this PR, I'd like to get these in separately so it's easier to review the changes for the GitHub actions later.

test/API/CMakeLists.txt
```diff
@@ -161,109 +169,6 @@ if (HDF5_ENABLE_FORMATTERS)
   clang_format (HDF5_TEST_h5_api_test_FORMAT h5_api_test)
 endif ()
 
-if (HDF5_TEST_API_INSTALL)
```
jhendersonHDF, line 164: Moved the install section to the bottom of this file for consistency with elsewhere

test/API/CMakeLists.txt
```diff
@@ -161,109 +169,6 @@ if (HDF5_ENABLE_FORMATTERS)
...
-      OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
-    COMPONENT
-      tests
-  )
-endif ()
-
-#-----------------------------------------------------------------------------
-# Build the ported HDF5 test executables
-#-----------------------------------------------------------------------------
-foreach (api_test_extra ${HDF5_API_TESTS_EXTRA})
```
jhendersonHDF, line 182: This whole block can be removed since we no longer build testhdf5, etc. twice and instead use the existing executable

test/API/CMakeLists.txt
```diff
@@ -332,93 +227,99 @@ if (HDF5_TEST_SERIAL)
       )
     endforeach ()
 
-    # Add tests for each external VOL connector that was built
```
jhendersonHDF, line 335: Everything here just shifted over once inside the `if (BUILD_SHARED_LIBS)` block

test/API/CMakeLists.txt
```diff
@@ -309,17 +215,6 @@ if (HDF5_TEST_SERIAL)
       set (last_api_test "h5_api_test_${api_test}")
     endforeach ()
 
-    foreach (hdf5_test ${HDF5_API_TESTS_EXTRA})
```
jhendersonHDF, line 312: Only run testhdf5, etc. with VOL connectors. Don't run it twice with the native connector

test/API/CMakeLists.txt
```diff
@@ -431,70 +332,117 @@ if (HDF5_TEST_SERIAL)
...
-
-      get_target_property (vol_lib_targets "${external_vol_tgt}" HDF5_VOL_TARGETS)
-      foreach (lib_target ${vol_lib_targets})
-        get_target_property (lib_target_output_dir "${lib_target}" LIBRARY_OUTPUT_DIRECTORY)
-        if (NOT "${lib_target_output_dir}" STREQUAL "lib_target_output_dir-NOTFOUND"
-            AND NOT "${lib_target_output_dir}" STREQUAL ""
-            AND NOT "${lib_target_output_dir}" STREQUAL "${CMAKE_BINARY_DIR}/${HDF5_INSTALL_BIN_DIR}")
-          set (vol_plugin_paths "${vol_plugin_paths}${CMAKE_SEP}${lib_target_output_dir}")
-        endif ()
+    if (BUILD_SHARED_LIBS)
```
jhendersonHDF, line 343: Same here about everything shifting over

byrnHDF (commented):
CMake looks good

testpar/API/CMakeLists.txt
jhendersonHDF, line 1: The same comments for `test/API/CMakeLists.txt` generally apply in this file as well

test/tmisc.c
```diff
@@ -1255,36 +1267,23 @@ test_misc8(void)
     hsize_t storage_size; /* Number of bytes of raw data storage used */
     int    *wdata;        /* Data to write */
     int    *tdata;        /* Temporary pointer to data write */
-#ifdef VERIFY_DATA
```
jhendersonHDF, line 1258: VERIFY_DATA is never defined anywhere, so I re-enabled this part of the test

testpar/t_filter_read.c
```diff
@@ -194,9 +205,8 @@ test_filter_read(void)
     unsigned      disable_partial_chunk_filters; /* Whether filters are disabled on partial chunks */
     herr_t        hrc;
     const char   *filename;
-#ifdef H5_HAVE_FILTER_FLETCHER32
```
jhendersonHDF, line 197: H5_HAVE_FILTER_FLETCHER32 is never defined anywhere, so I re-enabled this part of the test

jhendersonHDF:
Looks like there may be a few changes I made after these changes that may need to come over as well

test/tattr.c
```diff
@@ -6978,11 +7286,22 @@ attr_iterate_check(hid_t fid, const char *dsetname, hid_t obj_id, H5_index_t idx
...
     int    old_nerrs; /* Number of errors when entering this check */
-    herr_t ret;       /* Generic return value */
+    bool   vol_is_native;
+    herr_t ret; /* Generic return value */
 
     /* Retrieve the current # of reported errors */
     old_nerrs = GetTestNumErrs();
 
+    if (!(vol_cap_flags_g & H5VL_CAP_FLAG_ITERATE)) {
+        SKIPPED();
```
mattjala, line 7296: There are a few instances of `SKIPPED()` - should these be `MESSAGE(5, (" -- SKIPPED --\n"));` for consistency?

jhendersonHDF: In general I tried to use MESSAGE only if there is any output at the start of a test function which is also printed out with MESSAGE. That way the SKIPPED message is only printed if output for the test is printed. In this case, it probably even makes more sense to just return without printing skipped since there isn't any output nearby. I don't believe these capability flag changes were ever checked with a VOL connector that would fail at this point in order to see what the output looks like. That's a good todo item.

mattjala (approved):
LGTM

brtnfld:
h5test.c line 1282, var should be env?

#ifdef H5_HAVE_PARALLEL
 * Function:  getenv_all
 *
 * Purpose:  Used to get the environment that the root MPI task has.
 *     name specifies which environment variable to look for
 ***     var ....**

jhendersonHDF:
> h5test.c line 1282, var should be env?
> 
> #ifdef H5_HAVE_PARALLEL /*-------------------------------------------------------------------------
> 
>     * Function:  getenv_all
> 
>     * Purpose:  Used to get the environment that the root MPI task has.
> 
>     * ```
>       name specifies which environment variable to look for
>       ```
> 
> 
> *** var ....**

`var` or `val`? Are you referring to the line that says "val is the string to which the value of that environment variable will be copied."? It looks like that's out of date since there's no other parameter to the function and the value is instead returned directly from the function.

brtnfld (approved):
LGTM

brtnfld:
I was referring to var description.

fortnern (approved):

PR #3837, opened 2023-11-07
Disable FP exceptions in H5T init code

### LINKED ISSUE #3831: `DETECT_F` causes an invalid FP operation
**Describe the bug**
Calling `h5open_f` (Fortran) leads to an invalid floating point exception. This is problematic when catching FPEs, as the application crashes when initializing HDF5.

With the following `prgm.f90` program:

```fortran
program prgm
    use hdf5
    implicit none
    integer :: ierr
    call h5open_f(ierr)
    if (ierr /= 0) then
        error stop "error open"
    endif
    call h5close_f(ierr)
    if (ierr /= 0) then
        error stop "error close"
    endif
    print "(a)", "all good"
end program
```

We have the following issue:

```sh
$ h5pfc -showme
gfortran -I/usr/include -L/usr/lib -lhdf5hl_fortran -lhdf5_hl -lhdf5_fortran -lhdf5 -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto=auto -lsz -lz -ldl -lm -Wl,-rpath -Wl,/usr/lib

$ h5pfc prgm.f90 -ffpe-trap=invalid

$ gdb a.out
# ...
(gdb) run
# ...
Program received signal SIGFPE, Arithmetic exception.
0x00007ffff7cd5072 in H5T__init_native_float_types () at /usr/src/debug/hdf5-openmpi/hdf5-1.14.3/src/H5Tinit_float.c:531
531         DETECT_F(long double, LDOUBLE, det);
```

**Expected behavior**
No FPE should be raised when initializing HDF5.

 - HDF5 version: 1.14.3
 - OS and version: arch linux, up-to-date at the time of writing
 - Compiler and version: gfortran 13.2.1
 - Build system (e.g. CMake, Autotools) and version: NA
 - Any configure options you specified: NA
 - MPI library and version (parallel HDF5): NA

**Additional context**
We do not encounter the issue with hdf5 1.12 on other machines.

derobins:
The H5T floating-point datatype initialization code can raise exceptions when handling signaling NaNs. This change disables FE_INVALID exceptions during initialization.

Also removes the -ieee=full change for NAG Fortran as that shouldn't be necessary anymore.

Fixes #3831

### DIFF
--- a/src/H5Tinit_float.c
+++ b/src/H5Tinit_float.c
@@ -51,19 +51,23 @@
  * Function:    DETECT_F
  *
  * Purpose:     This macro takes a floating point type like `double' and
- *              a base name like `natd' and detects byte order, mantissa
- *              location, exponent location, sign bit location, presence or
- *              absence of implicit mantissa bit, and exponent bias and
- *              initializes a detected_t structure with those properties.
+ *              and detects byte order, mantissa location, exponent location,
+ *              sign bit location, presence or absence of implicit mantissa
+ *              bit, and exponent bias and initializes a detected_t structure
+ *              with those properties.
+ *
+ *              Note that these operations can raise floating-point
+ *              exceptions and building with some compiler options
+ *              (especially Fortran) can cause problems.
  *-------------------------------------------------------------------------
  */
-#define DETECT_F(TYPE, VAR, INFO)                                                                            \
+#define DETECT_F(TYPE, INFO)                                                                                 \
     do {                                                                                                     \
-        TYPE          _v1, _v2, _v3;                                                                         \
-        unsigned char _buf1[sizeof(TYPE)], _buf3[sizeof(TYPE)];                                              \
-        unsigned char _pad_mask[sizeof(TYPE)];                                                               \
-        unsigned char _byte_mask;                                                                            \
-        int           _i, _j, _last = (-1);                                                                  \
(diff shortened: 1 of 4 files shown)


brtnfld:
We'll also need to remove the NAG know issues blurb in Release.txt.

derobins:
Stupid GNU extensions. Will fix.

brtnfld (approved):
lgtm

src/H5Tinit_float.c
```diff
@@ -51,19 +51,23 @@
...
+ *
+ *              Note that these operations can raise floating-point
+ *              exceptions and building with some compiler options
+ *              (especially Fortran) can cause problems.
  *-------------------------------------------------------------------------
  */
-#define DETECT_F(TYPE, VAR, INFO)                                                                            \
+#define DETECT_F(TYPE, INFO)                                                                                 \
     do {                                                                                                     \
-        TYPE          _v1, _v2, _v3;                                                                         \
```
fortnern, line 62: Is there trailing whitespace on these lines?

fortnern (approved):

src/H5Tinit_float.c
derobins: No. This is macro code so the line extends to the line continuation mark.

PR #3836, opened 2023-11-07
Add intel oneapi windows build to CI CMake

byrnHDF:
Replaces #3738

PR #3839, opened 2023-11-09
Fixes Fortran compilation  on Windows

brtnfld:
Fixes unresolved symbols with Fortran libraries on Windows with Intel OneAPI.


brtnfld:
Need to resolve subfiling DLLs when parallel is enabled.

hyoklee (approved):

fortran/src/CMakeLists.txt
```diff
@@ -358,7 +359,10 @@ if (BUILD_STATIC_LIBS)
 endif ()
 
 if (BUILD_SHARED_LIBS)
-  add_library (${HDF5_F90_LIBSH_TARGET} SHARED ${f90_F_SOURCES_SHARED})
+  if(MSVC)
+      set(DLLDEF ${HDF5_F90_SRC_BINARY_DIR}/hdf5_fortrandll.def)
+  endif()
+  add_library (${HDF5_F90_LIBSH_TARGET} SHARED ${DLLDEF} ${f90_F_SOURCES_SHARED})
```
byrnHDF, line 367: This relies on "DLLDEF" not being defined for non-msvc. I would prefer that we use an if/else here and in the other file. The other question and maybe why the other form did not work - should we check for it being on windows "if (WIN32) " instead of "if (MSVC)"  - because those are not the same thing. (Another choice is "if (WINDOWS)")

fortran/src/CMakeLists.txt
```diff
@@ -374,13 +378,6 @@ if (BUILD_SHARED_LIBS)
       PUBLIC ${HDF5_F90_C_LIBSH_TARGET}
       PRIVATE ${LINK_Fortran_LIBS} "$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_Fortran_LIBRARIES}>"
   )
-#  set_property(TARGET ${HDF5_F90_LIBSH_TARGET} APPEND PROPERTY LINK_FLAGS "$<$<STREQUAL:x${CMAKE_Fortran_SIMULATE_ID},xMSVC>:-SUBSYSTEM:CONSOLE>")
-#  set_property(TARGET ${HDF5_F90_LIBSH_TARGET} APPEND PROPERTY LINK_FLAGS "$<$<STREQUAL:x${CMAKE_Fortran_SIMULATE_ID},xMSVC>:${WIN_LINK_FLAGS}>")
-#  set_property(TARGET ${HDF5_F90_LIBSH_TARGET} APPEND PROPERTY LINK_FLAGS "$<$<STREQUAL:x${CMAKE_Fortran_SIMULATE_ID},xMSVC>:-DLL>")
-#  set_property(TARGET ${HDF5_F90_LIBSH_TARGET} APPEND PROPERTY LINK_FLAGS "$<$<STREQUAL:x${CMAKE_Fortran_SIMULATE_ID},xMSVC>:-DEF:${HDF5_F90_SRC_BINARY_DIR}/hdf5_fortrandll.def>")
-  if(MSVC)
```
byrnHDF, line 381: Maybe MSVC should be WIN32? OR maybe the commented out property commands might work now?

byrnHDF (changes requested):
The add_library changes are the only ones that need to be fixed.

brtnfld:
Replaced with PR #3836

Issue #3840, opened 2023-11-09
Compound conversion with user callback segfaults


mattjala:
**Describe the bug**
When the user defines a conversion function between two compound types that doesn't provide a value for `cdata->private`, the library will segfault while checking if the two compound types are subsets of one another in `H5D__typeinfo_init`.

**Expected behavior**
Type conversion and dataset write should complete without errors. `cdata->priv` should not be used by the library when a user-defined callback function is specified.

**Platform (please complete the following information)**
 - HDF5 version: Develop (commit hash `5e787741cd`)
 - Ubuntu debian 22.04
 - gcc 11.3.0
 - CMake 3.22.1
 - HDF5 built in Release mode

**Additional context**
Replication program:

```
#include "hdf5.h"
#include <stdlib.h>

typedef struct src_t {
     uint32_t a;
     float b;
  } src_t;

typedef struct dst_t {
     float b;
  } dst_t;

herr_t convert(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
    size_t nelmts, size_t buf_stride, size_t bkg_stride, void *buf,
    void *bkg, hid_t dxpl)
{
  herr_t retval = EXIT_SUCCESS;
  switch (cdata->command)
  {
  case H5T_CONV_INIT:
    printf("Initializing conversion function...\n");
    break;
  case H5T_CONV_CONV:
    printf("Converting...\n");
    for (size_t i = 0; i < nelmts; ++i)
      ((dst_t*) buf)[i].b = ((src_t*) buf)[i].b;
    break;
  case H5T_CONV_FREE:
    printf("Finalizing conversion function...\n");
    break;
  default:
    break;
  }
  return retval;
}

int main() {
  hid_t src = H5I_INVALID_HID;
  hid_t dst = H5I_INVALID_HID;
  hid_t file_id = H5I_INVALID_HID;
  hid_t space_id = H5I_INVALID_HID;
  hid_t dset_id = H5I_INVALID_HID;
  
  if ((src = H5Tcreate(H5T_COMPOUND, sizeof(struct src_t))) < 0) {
    printf("src dtype creation failed\n");
  }

  if (H5Tinsert(src, "a", HOFFSET(struct src_t, a), H5T_NATIVE_UINT32) < 0) {
    printf("dtype insertion failed\n");
  }

  if (H5Tinsert(src, "b", HOFFSET(struct src_t, b), H5T_NATIVE_FLOAT) < 0) {
[...]

PR #3842, opened 2023-11-10
Fix segfault on user compound dtype conversion callback

mattjala:
Fix for #3840, with associated test.

The bad memory access occurred while checking if the two datatypes were compound subsets of one another. This check is done using information in `cdata->priv`, which is never set when a user conversion callback is in use. This fixes the issue by returning NULL, which is already checked for in the macros that use the compound subset information (`H5D__SCATGATH_USE_CMPD_OPT_READ`, `H5D__SCATGATH_USE_CMPD_OPT_WRITE`).

### DIFF
--- a/test/dtypes.c
+++ b/test/dtypes.c
@@ -100,6 +100,15 @@ typedef enum dtype_t {
     OTHER
 } dtype_t; /* This doesn't seem to be used anywhere... -BMR */
 
+typedef struct src_cmpd_t {
+    uint32_t a;
+    float    b;
+} src_cmpd_t;
+
+typedef struct dst_cmpd_t {
+    float b;
+} dst_cmpd_t;
+
 typedef enum { E1_RED, E1_GREEN, E1_BLUE, E1_ORANGE, E1_YELLOW } color_t;
 
 /* Constant for size of conversion buffer for int <-> float exception test */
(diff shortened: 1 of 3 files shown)


fortnern:
I'm concerned that a user conversion function might use the priv field for its own purposes, and then still trigger the bug because the library code only checks to see if priv is NULL. Instead, we could add a check in H5T_path_compound_subset() to see if it's using an app conversion function (and maybe check to see if the function is H5T__conv_struct() instead of checking for are_compounds) before calling H5T__conv_struct_subset(). H5T_path_compound_subset() is the only place that calls H5T__conv_struct_subset().

fortnern (approved):

test/dtypes.c
```diff
@@ -3788,6 +3819,109 @@ test_compound_18(void)
...
+    assert(space_id > 0);
+
+    /* Create a dataset with the destination compound datatype */
+    if ((dset_id = H5Dcreate2(file_id, "dataset", dst, space_id, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT)) <
+        0) {
+        FAIL_PUTS_ERROR("couldn't create dataset with compound datatype");
+    }
+
+    if (H5Dwrite(dset_id, src, space_id, H5S_ALL, H5P_DEFAULT, buf) < 0) {
+        FAIL_PUTS_ERROR("couldn't write to dataset with user compound conversion function");
```
derobins: These should probably all just be TEST_ERROR. FAIL_STACK_ERROR is rarely needed and FAIL_PUTS_ERROR is usually only used when erroring on non-HDF5 errors. The convention is also to not use brackets around the error lines.

lrknox (approved):

PR #3845, opened 2023-11-13
cmakedefine macro uses C comments with "#undef"

byrnHDF:


lrknox:
Change fixed compile failure on vortex (ppc64le @ sandia) with clang 14.0.4

brtnfld (approved):
It looks fine, but I don't think we use C_SIZEOF anymore. I can address that in a separate PR.

lrknox (approved):

PR #3859, opened 2023-11-16
Add HDF5_DISABLE_TESTS_REGEX option to skip tests

byrnHDF:


lrknox (approved):

tools/test/h5repack/CMakeTests.cmake
```diff
@@ -314,7 +323,7 @@
             NAME H5REPACK-${testname}
             COMMAND ${CMAKE_COMMAND} -E echo "SKIP ${ARGN} ${PROJECT_BINARY_DIR}/testfiles/${testfile} ${PROJECT_BINARY_DIR}/testfiles/out-${testname}.${testfile}"
         )
-        set_property(TEST H5REPACK-${testname} PROPERTY DISABLED)
+        set_property(TEST H5REPACK-${testname} PROPERTY DISABLED true)
```
lrknox, line 326: Is this missing if ("H5REPACK-${testname}" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")?  When I tested, not specifying H5REPACK nor H5IMPORT, there were 25 H5REPACK tests and 2 H5IMPORT tests that were disabled.

lrknox (changes requested):

lrknox (approved):
Those skips were caused by non-functioning Zlib and Szip tar.gz files.  Running it again with updated tar.gz file skipped  only the designated ph5diff and t_bigio tests.

brtnfld:
Will you create an issue for autotools to have the same functionality?

CMakeTests.cmake
```diff
@@ -23,6 +23,9 @@
   math (EXPR CTEST_LONG_TIMEOUT "${DART_TESTING_TIMEOUT} * 2")
   math (EXPR CTEST_VERY_LONG_TIMEOUT "${DART_TESTING_TIMEOUT} * 3")
 
+  option (HDF5_DISABLE_TESTS_REGEX "Pattern to set tests to DISABLED" "")
```
brtnfld: Maybe "exclude" would be more precise than "disable", "Pattern used to exclude specific tests"

byrnHDF: #3863 created

byrnHDF:
> Will you create an issue for autotools to have the same functionality?
#3864 created.

CMakeTests.cmake
byrnHDF: Not really, the tests are being skipped by disabling the execution of the tests. Besides the CMake property is DISABLED

brtnfld: I think making things clear for the user should be a priority. So far, we have used "ENABLE" and "DISABLE" as boolean values.

at least: "Pattern used to mark specific tests as DISABLED"

byrnHDF: Changed option text

lrknox: @brtnfld, is this good now?

PR #3861, opened 2023-11-20
Add MSYS2 action

hyoklee:
Some tests fail on all 4 environments.
However, compilation works fine.
clang17 environment needs extra definitions.

### DIFF
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -1,0 +1,57 @@
+name: msys2
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+    branches: [ develop ]
+    paths-ignore:
+      - '.github/CODEOWNERS'
+      - '.github/FUNDING.yml'
+      - 'doc/**'
+      - 'release_docs/**'
+      - 'ACKNOWLEDGEMENTS'
+      - 'COPYING**'
+      - '**.md'
+
+permissions:
+  contents: read
+
+# Using concurrency to cancel any in-progress job or run
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        sys:
+          - mingw64
+          - mingw32
+          - ucrt64
+          - clang64
+    runs-on: windows-latest
+    steps:
+      - name: Setup MSYS2
+        uses: msys2/setup-msys2@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/msys2.yml
```diff
@@ -0,0 +1,57 @@
...
+            ninja:p
+      - name: Test HDF5
+        shell: msys2 {0}
+        run: |
+          uname -a
+          git clone https://github.com/HDFgroup/hdf5.git
+          cd hdf5
+          mkdir build
+          cd build
+          cmake -DH5_HAVE_ALARM:INTERNAL=0 -DH5_HAVE_ASPRINTF:INTERNAL=0 -DH5_HAVE_VASPRINTF:INTERNAL=0 ..
```
derobins, line 55: We should probably move the alarm, asprintf, and vasprintf checks in CMake so they check on Windows (I'm assuming they are not checked on Win32).

derobins: Interestingly, they are run all the time. We'll need to figure out why they are not being detected on Windows w/ MSYS2.

derobins (approved):

.github/workflows/msys2.yml
```diff
@@ -0,0 +1,57 @@
...
+            cmake:p
+            ninja:p
+      - name: Test HDF5
+        shell: msys2 {0}
+        run: |
+          uname -a
+          git clone https://github.com/HDFgroup/hdf5.git
+          cd hdf5
+          mkdir build
+          cd build
```
byrnHDF, line 54: Please add steps between configure, build and test

.github/workflows/msys2.yml
```diff
@@ -0,0 +1,57 @@
+name: msys2
```
byrnHDF, line 1: Change name to match other yml files and add job name with matrix.sys variable

.github/workflows/msys2.yml
```diff
@@ -0,0 +1,57 @@
...
+
+on:
+  workflow_dispatch:
+  push:
```
byrnHDF, line 5: At only 12 mins of time we should make this a callable workflow from the cmake.yml.

.github/workflows/msys2.yml
byrnHDF, line 55: Is there is a reason to not use the same configure / build / test steps we use in other CMake yml files?

byrnHDF:
We should fix the issues raised here and then create a solution that addresses all the issues for MSYS testing.

byrnHDF:
Superseded by #3938

PR #3868, opened 2023-11-22
Complete the `if command line option` sentence.

hyoklee:
Make comment clear and consistent.

From:
```
   /* enable error reporting if command line option */
```

To:
```
  /* Enable error reporting if -E command line option is specified */
```

### DIFF
--- a/tools/src/h5dump/h5dump.c
+++ b/tools/src/h5dump/h5dump.c
@@ -1357,7 +1357,7 @@ main(int argc, char *argv[])
         goto done;
     }
 
-    /* enable error reporting if command line option */
+    /* Enable error reporting if -E command line option is specified */
     h5tools_error_report();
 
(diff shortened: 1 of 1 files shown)


tools/src/h5dump/h5dump.c
```diff
@@ -1357,7 +1357,7 @@ main(int argc, char *argv[])
         goto done;
     }
 
-    /* enable error reporting if command line option */
+    /* Enable error reporting if -E command line option is specified */
```
byrnHDF: While technically correct, we are discouraging the use of the short names. Use "--enable-error-stack"

byrnHDF (changes requested):

byrnHDF (approved):

bmribler (approved):

lrknox (approved):

PR #3874, opened 2023-11-28
Update examples with 1.6 api and static libs

byrnHDF:


lrknox (approved):
Fixes new configure error on theta

jhendersonHDF (approved):

PR #3877, opened 2023-11-30
Report error for h5dump subfiling vfd option if mpi is unavailable

hyoklee:
This PR will help user to understand why subfiling vfd failed for h5dump.

This PR also allows `--vfd-value=12` option for h5dump. 
12 means subfiling.

### DIFF
--- a/tools/lib/h5tools.c
+++ b/tools/lib/h5tools.c
@@ -597,6 +597,9 @@ h5tools_set_fapl_vfd(hid_t fapl_id, h5tools_vfd_info_t *vfd_info)
                     if (H5Pset_fapl_subfiling(fapl_id, (const H5FD_subfiling_config_t *)vfd_info->info) < 0)
                         H5TOOLS_GOTO_ERROR(FAIL, "H5Pset_fapl_subfiling() failed");
                 }
+                else {
+                    H5TOOLS_GOTO_ERROR(FAIL, "MPI is not available for Subfiling VFD");
+                }
 #else
                 H5TOOLS_GOTO_ERROR(FAIL, "The Subfiling VFD is not enabled");
 #endif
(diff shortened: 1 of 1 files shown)


tools/lib/h5tools.c
```diff
@@ -626,6 +629,12 @@ h5tools_set_fapl_vfd(hid_t fapl_id, h5tools_vfd_info_t *vfd_info)
              *
              * Currently, driver configuration strings are unsupported.
              */
+#if defined(H5_HAVE_PARALLEL) && defined(H5_HAVE_SUBFILING_VFD)
+            if (vfd_info->u.value == H5_VFD_SUBFILING) {
+                if (H5Pset_fapl_subfiling(fapl_id, (const H5FD_subfiling_config_t *)vfd_info->info) < 0)
+                    H5TOOLS_GOTO_ERROR(FAIL, "H5Pset_fapl_subfiling() failed");
+            }
```
jhendersonHDF: We should also have the section from above:

```
#else
                H5TOOLS_GOTO_ERROR(FAIL, "The Subfiling VFD is not enabled");
#endif
```

tools/lib/h5tools.c
```diff
@@ -597,6 +597,9 @@ h5tools_set_fapl_vfd(hid_t fapl_id, h5tools_vfd_info_t *vfd_info)
                     if (H5Pset_fapl_subfiling(fapl_id, (const H5FD_subfiling_config_t *)vfd_info->info) < 0)
                         H5TOOLS_GOTO_ERROR(FAIL, "H5Pset_fapl_subfiling() failed");
                 }
+                else {
+                    H5TOOLS_GOTO_ERROR(FAIL, "MPI is not available for Subfiling VFD");
```
jhendersonHDF: See my comment on #3880; this may not be needed if we decide to remove this check.

tools/lib/h5tools.c
```diff
@@ -626,6 +618,16 @@ h5tools_set_fapl_vfd(hid_t fapl_id, h5tools_vfd_info_t *vfd_info)
...
+
+            if (vfd_info->u.value == H5_VFD_SUBFILING) {
+#if defined(H5_HAVE_PARALLEL) && defined(H5_HAVE_SUBFILING_VFD)
+                if (H5Pset_fapl_subfiling(fapl_id, (const H5FD_subfiling_config_t *)vfd_info->info) < 0)
+                    H5TOOLS_GOTO_ERROR(FAIL, "H5Pset_fapl_subfiling() failed");
+#else
+                H5TOOLS_GOTO_ERROR(FAIL, "The Subfiling VFD is not enabled");
+#endif
+            }
+
```
jhendersonHDF: The call to `H5Pset_driver_by_value` below should be in an else statement so that it doesn't try to override the call to `H5Pset_fapl_subfiling` added here

jhendersonHDF (approved):

bmribler (approved):

PR #3878, opened 2023-11-30
Add subfiling for h5dump filedriver option help message

hyoklee:
`subfiling` is missing in h5dump help message as an available option value.
This PR will fix it.

### DIFF
--- a/tools/src/h5dump/h5dump.c
+++ b/tools/src/h5dump/h5dump.c
@@ -330,7 +330,7 @@ usage(const char *prog)
     PRINTVALSTREAM(
         rawoutstream,
         "      \"sec2\", \"family\", \"split\", \"multi\", \"direct\", \"stream\", and \"subfiling\".\n");
-    PRINTVALSTREAM(rawoutstream, "      Without the file driver flag, the file will be opened with each driver in\n");    
+    PRINTVALSTREAM(rawoutstream, "      Without the file driver flag, the file will be opened with each driver in\n");
     PRINTVALSTREAM(rawoutstream, "      turn and in the order specified above until one driver succeeds\n");
     PRINTVALSTREAM(rawoutstream, "      in opening the file.\n");
     PRINTVALSTREAM(rawoutstream,
--- a/tools/src/h5dump/h5dump.c
+++ b/tools/src/h5dump/h5dump.c
@@ -326,11 +326,13 @@ usage(const char *prog)
                    "      (Alternate compact form of subsetting is described in the Reference Manual)\n");
     PRINTVALSTREAM(rawoutstream, "\n");
     PRINTVALSTREAM(rawoutstream, "--------------- Option Argument Conventions ---------------\n");
-    PRINTVALSTREAM(rawoutstream,
-                   "  D - is the file driver to use in opening the file. Acceptable values are\n");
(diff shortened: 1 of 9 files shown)


tools/src/h5dump/h5dump.c
```diff
@@ -326,11 +326,13 @@ usage(const char *prog)
...
     PRINTVALSTREAM(rawoutstream, "\n");
     PRINTVALSTREAM(rawoutstream, "--------------- Option Argument Conventions ---------------\n");
-    PRINTVALSTREAM(rawoutstream, "  D - is the file driver to use in opening the file. Acceptable values\n");
+    PRINTVALSTREAM(rawoutstream,
+                   "  D - is the file driver to use in opening the file. Acceptable values are\n");
     PRINTVALSTREAM(
         rawoutstream,
-        "      are \"sec2\", \"family\", \"split\", \"multi\", \"direct\", and \"stream\". Without\n");
-    PRINTVALSTREAM(rawoutstream, "      the file driver flag, the file will be opened with each driver in\n");
+        "      \"sec2\", \"family\", \"split\", \"multi\", \"direct\", \"stream\", and \"subfiling\".\n");
```
jhendersonHDF: I think rather than adding individual items each time a new driver is "supported", we should just have a documented place we can point to with the names to specify for each VFD so that we don't have this churn of having to update testing files. Other VFD names beyond those listed here can already be specified (and I'm not sure that "stream" is actually valid) here and for other tools, so having this in one place in either the doxygen or somewhere else would be more maintainable.

tools/src/h5dump/h5dump.c
```diff
@@ -326,11 +326,15 @@ usage(const char *prog)
...
-    PRINTVALSTREAM(rawoutstream, "  D - is the file driver to use in opening the file. Acceptable values\n");
     PRINTVALSTREAM(
         rawoutstream,
-        "      are \"sec2\", \"family\", \"split\", \"multi\", \"direct\", and \"stream\". Without\n");
-    PRINTVALSTREAM(rawoutstream, "      the file driver flag, the file will be opened with each driver in\n");
+        "  D - is the file driver to use in opening the file. Acceptable values are available from\n");
+    PRINTVALSTREAM(
+        rawoutstream,
+        "      "
+        "https://portal.hdfgroup.org/documentation/hdf5-docs/registered_virtual_file_drivers_vfds.html.\n");
```
lrknox, line 335: @bmribler @loricooperhdf  Are the portal.hdfgroup.org urls in this PR going to remain, or will they need to be replaced?

lrknox: They include "Hosted on GitHub Pages", so I assume they're probably going to be around.

bmribler: > @bmribler @loricooperhdf Are the portal.hdfgroup.org urls in this PR going to remain, or will they need to be replaced?

This url, https://portal.hdfgroup.org/documentation/hdf5-docs/registered_virtual_file_drivers_vfds.html, is new and I assume it'll stay.  Lori can confirm.

lrknox (approved):

glennsong09 (approved):

derobins (approved):

PR #3880, opened 2023-12-01
Support h5dump command line options for subfiling VFD

hyoklee:
The three h5dump options 

* --filedriver=subfiling
* --vfd-name=subfiling
*  --vfd-value=12

do not work for HDF5 stub file.

This PR will allow users to specify one of them to dump HDF5 stub file created by subfiling VFD without relying on setting `HDF5_DRIVER` environment variable.

See also #3879 about the test stub file.
See also PR #3877. It is required to make `--vfd-value=12` work.

### DIFF
--- a/tools/src/h5dump/h5dump.c
+++ b/tools/src/h5dump/h5dump.c
@@ -149,6 +149,29 @@ static struct h5_long_options l_opts[] = {{"attribute", require_arg, 'a'},
                                           {"vfd-info", require_arg, '6'},
                                           {NULL, 0, '\0'}};
 
+#if defined(H5_HAVE_PARALLEL) && defined(H5_HAVE_SUBFILING_VFD)
+static int mpi_code_g          = MPI_ERR_OTHER;
+static int use_subfiling_vfd_g = false;
+static int
+set_mpi(int argc, char *argv[])
+{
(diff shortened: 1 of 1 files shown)


tools/src/h5dump/h5dump.c
```diff
@@ -149,6 +149,29 @@ static struct h5_long_options l_opts[] = {{"attribute", require_arg, 'a'},
                                           {"vfd-info", require_arg, '6'},
                                           {NULL, 0, '\0'}};
 
+#if defined(H5_HAVE_PARALLEL) && defined(H5_HAVE_SUBFILING_VFD)
```
jhendersonHDF, line 152: Rather than needing to call MPI_Init in every serial tool that wants to support parallel VFDs/VOL connectors, I think it makes more sense to remove the check for whether MPI is initialized in the tools library and then document that a parallel VFD or VOL connector should initialize MPI itself if it needs it and it wasn't already initialized. However, this assumes that systems which were previously broken in this regard (Theta IIRC) are no longer problematic due to the tools trying to open a file with any available VFD when the first attempt fails. In fact, it may even be appropriate to remove that functionality from the tools at this point.

Regardless, we shouldn't generally have functionality in the tools with an #ifdef for a particular VFD.

PR #3886, opened 2023-12-04
autogen.sh: check directly for libtoolize instead of libtool [as libtool  is not used here]

balay:
On ubuntu - libtoolize can be installed without libtool. So best to check directly for libtoolize - instead of indirectly via  libtool

```
$ ./autogen.sh

**************************
* HDF5 autogen.sh script * **************************

Running trace script:
Finished processing HDF5 API calls:
	instrumented 0 API functions and 0 argument lists

Running error generation script:
Generating 'H5Epubgen.h'
Generating 'H5Einit.h'
Generating 'H5Eterm.h'
Generating 'H5Edefin.h'

Running API version generation script:
Generating 'H5version.h'

Running overflow macro generation script:
Generating 'H5overflow.h'

./libtoolize --copy --force
./autogen.sh: 202: ./libtoolize: not found
```

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -44,7 +44,7 @@
 #   HDF5_AUTOHEADER
 #   HDF5_AUTOMAKE
 #   HDF5_AUTOCONF
-#   HDF5_LIBTOOL
+#   HDF5_LIBTOOLIZE
 #   HDF5_M4
 #
 # Note that aclocal will attempt to include libtool's share/aclocal
@@ -111,14 +111,14 @@ fi
 if test -z "${HDF5_ACLOCAL}"; then
     HDF5_ACLOCAL="$(command -v aclocal)"
 fi
-if test -z "${HDF5_LIBTOOL}"; then
+if test -z "${HDF5_LIBTOOLIZE}"; then
     case "$(uname)" in
     Darwin*)
         # libtool on OS-X is non-gnu
-        HDF5_LIBTOOL="$(command -v glibtool)"
+        HDF5_LIBTOOLIZE="$(command -v glibtoolize)"
         ;;
     *)
-        HDF5_LIBTOOL="$(command -v libtool)"
+        HDF5_LIBTOOLIZE="$(command -v libtoolize)"
         ;;
     esac
 fi
@@ -129,25 +129,9 @@ fi
 
 # Make sure that these versions of the autotools are in the path
 AUTOCONF_DIR=$(dirname "${HDF5_AUTOCONF}")
-LIBTOOL_DIR=$(dirname "${HDF5_LIBTOOL}")
+LIBTOOL_DIR=$(dirname "${HDF5_LIBTOOLIZE}")
 M4_DIR=$(dirname "${HDF5_M4}")
 PATH=${AUTOCONF_DIR}:${LIBTOOL_DIR}:${M4_DIR}:$PATH
-
-# Make libtoolize match the specified libtool
-case "$(uname)" in
-Darwin*)
-    # On OS X, libtoolize could be named glibtoolize or
-    # libtoolize. Try the former first, then fall back
-    # to the latter if it's not found.
-    HDF5_LIBTOOLIZE="${LIBTOOL_DIR}/glibtoolize"
-    if [ ! -f "$HDF5_LIBTOOLIZE" ] ; then
-        HDF5_LIBTOOLIZE="${LIBTOOL_DIR}/libtoolize"
-    fi
-    ;;
-*)
-    HDF5_LIBTOOLIZE="${LIBTOOL_DIR}/libtoolize"
-    ;;
-esac
 
 # Run scripts that process source.
 #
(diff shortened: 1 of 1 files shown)


autogen.sh
```diff
@@ -129,26 +129,10 @@ fi
...
-    # to the latter if it's not found.
-    HDF5_LIBTOOLIZE="${LIBTOOL_DIR}/glibtoolize"
-    if [ ! -f "$HDF5_LIBTOOLIZE" ] ; then
-        HDF5_LIBTOOLIZE="${LIBTOOL_DIR}/libtoolize"
-    fi
-    ;;
-*)
-    HDF5_LIBTOOLIZE="${LIBTOOL_DIR}/libtoolize"
-    ;;
-esac
```
derobins, line 150: Why are you removing the MacOS hack?

balay: Not sure I understand.

MacOS hack is already at line 116 above.

My change is removing the redirection from "libtool" to "libtoolize/glibtoolize" -  i.e why not look directly for `libtoolize/glibtoolize` [as libtool is not really used in this script - only `libtoolize/glibtoolize` is used]?

My build on MacOS goes through  fine  with this PR change
```
balay@mpro git.hdf5 % sw_vers              
ProductName:		macOS
ProductVersion:		14.1.2
BuildVersion:		23B92
balay@mpro git.hdf5 % which libtool        
/usr/bin/libtool
balay@mpro git.hdf5 % which libtoolize     
libtoolize not found
balay@mpro git.hdf5 % which glibtool       
/opt/homebrew/bin/glibtool
balay@mpro git.hdf5 % which glibtoolize    
/opt/homebrew/bin/glibtoolize
balay@mpro git.hdf5 % git show -q --oneline
d86cb0da314 (HEAD) autogen.sh: check directly for libtoolize instead of libtool [as libtool is not used here]
balay@mpro git.hdf5 % ./autogen.sh 

**************************
* HDF5 autogen.sh script *
**************************

Running trace script:
Finished processing HDF5 API calls:
	instrumented 0 API functions and 0 argument lists

Running error generation script:
Generating 'H5Epubgen.h'
Generating 'H5Einit.h'
Generating 'H5Eterm.h'
Generating 'H5Edefin.h'

Running API version generation script:
Generating 'H5version.h'

Running overflow macro generation script:
Generating 'H5overflow.h'

/opt/homebrew/bin/glibtoolize --copy --force
glibtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'bin'.
glibtoolize: copying file 'bin/ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
glibtoolize: copying file 'm4/libtool.m4'
glibtoolize: copying file 'm4/ltoptions.m4'
glibtoolize: copying file 'm4/ltsugar.m4'
glibtoolize: copying file 'm4/ltversion.m4'
glibtoolize: copying file 'm4/lt~obsolete.m4'
glibtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.

NOTE: You can ignore the warning about adding -I m4.
      We already do this in an included file.

/opt/homebrew/bin/aclocal --force -I m4 -I /opt/homebrew/bin/../share/aclocal

/opt/homebrew/bin/autoheader --force

/opt/homebrew/bin/automake --copy --add-missing --force-missing
configure.ac:506: installing 'bin/compile'
configure.ac:136: installing 'bin/config.guess'
configure.ac:136: installing 'bin/config.sub'
configure.ac:35: installing 'bin/install-sh'
configure.ac:35: installing 'bin/missing'
parallel-tests: installing 'bin/test-driver'
c++/src/Makefile.am: installing 'bin/depcomp'

/opt/homebrew/bin/autoconf --force --warnings=no-obsolete

*** SUCCESS ***

balay@mpro git.hdf5 % 


```

Or perhaps I don't know the use case - where the build breaks without this extra re-direction [from libtool to libtoolize]

balay: Checking if there are any issues to resolve before this PR can be merged!

derobins: Sorry - I didn't see that since I was looking at the PR on my phone.

derobins (approved):

autogen.sh
qkoziol: The previous MacOS code would check for both glibtoolize and libtoolize, but this change makes it only check for glibtoolize.   If it's working, great, let's simplify, but it might need the previous hack if other devs need libtoolize instead of glibtoolize on the Mac.

qkoziol (changes requested):
Might need to check for both libtoolize and glibtoolize on the Mac.  I suggest this code on all platforms, instead of the `case`:

```
    HDF5_LIBTOOLIZE="$(command -v glibtoolize)"
    if [ ! -f "$HDF5_LIBTOOLIZE" ] ; then
        HDF5_LIBTOOLIZE="$(command -v libtoolize)"
    fi
```

balay:
> Might need to check for both libtoolize and glibtoolize on the Mac. I suggest this code on all platforms, instead of the `case`:

thanks, pushed!

qkoziol (approved):

PR #3890, opened 2023-12-12
Correct urls and handling of external git repos

byrnHDF:
JPEG and LXF plugins do not have git repos
libaec and bzip2 must use a specific branch
Add missing cache entries for zlib, libeac and plugins


hyoklee (approved):
It would be nice if there's a corresponding GH action to test these.

lrknox (approved):

PR #3896, opened 2023-12-15
Implement optimized support for vector I/O in Subfiling VFD

jhendersonHDF:
Vector I/O requests are now processed within a single set of I/O call batches, rather than each I/O vector entry (tuple constructed from the types, addrs, sizes and bufs arrays) being processed individually. This allows I/O to be more efficiently parallelized among the I/O concentrator processes during large I/O requests.

### DIFF
--- a/src/H5FDsubfiling/H5FDioc.c
+++ b/src/H5FDsubfiling/H5FDioc.c
@@ -1610,12 +1610,14 @@ H5FD__ioc_write_vector_internal(H5FD_t *_file, uint32_t count, H5FD_mem_t H5_ATT
     H5FD_ioc_t          *file_ptr      = (H5FD_ioc_t *)_file;
     io_req_t           **sf_io_reqs    = NULL;
     int64_t              sf_context_id = -1;
+    size_t               io_size       = 0;
+    bool                 extend_sizes  = false;
     herr_t               ret_value     = SUCCEED;
 
     assert(_file);
-    assert(addrs);
-    assert(sizes);
-    assert(bufs);
+    assert((addrs) || (count == 0));
+    assert((sizes) || (count == 0));
+    assert((bufs) || (count == 0));
 
     if (count == 0)
         H5_SUBFILING_GOTO_DONE(SUCCEED);
@@ -1648,12 +1650,22 @@ H5FD__ioc_write_vector_internal(H5FD_t *_file, uint32_t count, H5FD_mem_t H5_ATT
     for (size_t i = 0; i < (size_t)count; i++) {
         herr_t write_status;
 
-        if (sizes[i] == 0)
+        if (!extend_sizes) {
+            if ((i > 0) && (sizes[i] == 0)) {
+                extend_sizes = true;
+                io_size      = sizes[i - 1];
+            }
+            else {
+                io_size = sizes[i];
+            }
+        }
+
+        if (io_size == 0)
             H5_SUBFILING_GOTO_ERROR(H5E_IO, H5E_WRITEERROR, FAIL, "invalid size argument of 0");
 
         H5_CHECK_OVERFLOW(addrs[i], haddr_t, int64_t);
-        H5_CHECK_OVERFLOW(sizes[i], size_t, int64_t);
-        write_status = ioc__write_independent_async(sf_context_id, (int64_t)addrs[i], (int64_t)sizes[i],
+        H5_CHECK_OVERFLOW(io_size, size_t, int64_t);
+        write_status = ioc__write_independent_async(sf_context_id, (int64_t)addrs[i], (int64_t)io_size,
                                                     bufs[i], &sf_io_reqs[i]);
 
         if (write_status < 0)
@@ -1691,12 +1703,14 @@ H5FD__ioc_read_vector_internal(H5FD_t *_file, uint32_t count, haddr_t addrs[], s
     H5FD_ioc_t          *file_ptr      = (H5FD_ioc_t *)_file;
     io_req_t           **sf_io_reqs    = NULL;
     int64_t              sf_context_id = -1;
+    size_t               io_size       = 0;
+    bool                 extend_sizes  = false;
     herr_t               ret_value     = SUCCEED;
 
     assert(_file);
-    assert(addrs);
-    assert(sizes);
-    assert(bufs);
+    assert((addrs) || (count == 0));
+    assert((sizes) || (count == 0));
+    assert((bufs) || (count == 0));
 
     if (count == 0)
         H5_SUBFILING_GOTO_DONE(SUCCEED);
@@ -1720,12 +1734,22 @@ H5FD__ioc_read_vector_internal(H5FD_t *_file, uint32_t count, haddr_t addrs[], s
         H5_SUBFILING_GOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "can't allocate MPI request array");
 
     for (size_t i = 0; i < (size_t)count; i++) {
-        int read_status;
+        herr_t read_status;
+
+        if (!extend_sizes) {
+            if ((i > 0) && (sizes[i] == 0)) {
+                extend_sizes = true;
+                io_size      = sizes[i - 1];
+            }
+            else {
+                io_size = sizes[i];
+            }
+        }
 
         H5_CHECK_OVERFLOW(addrs[i], haddr_t, int64_t);
-        H5_CHECK_OVERFLOW(sizes[i], size_t, int64_t);
-        read_status = ioc__read_independent_async(sf_context_id, (int64_t)addrs[i], (int64_t)sizes[i],
-                                                  bufs[i], &sf_io_reqs[i]);
+        H5_CHECK_OVERFLOW(io_size, size_t, int64_t);
+        read_status = ioc__read_independent_async(sf_context_id, (int64_t)addrs[i], (int64_t)io_size, bufs[i],
+                                                  &sf_io_reqs[i]);
 
         if (read_status < 0)
             H5_SUBFILING_GOTO_ERROR(H5E_IO, H5E_READERROR, FAIL, "couldn't queue read operation");
--- a/src/H5FDsubfiling/H5FDioc_int.c
+++ b/src/H5FDsubfiling/H5FDioc_int.c
@@ -297,9 +297,13 @@ ioc__read_independent_async(int64_t context_id, int64_t offset, int64_t elements
      * unpredictable order. However, if some IOCs own more than
      * 1 subfile, we need to associate each read with a unique
      * message tag to make sure the data is received in the
-     * correct order.
+     * correct order. We also need a unique message tag in the
+     * case where only 1 subfile is used in total. In this case,
+     * vector I/O calls are passed directly down to this VFD without
+     * being split up into multiple I/O requests, so we need the
+     * tag to distinguish each I/O request.
      */
-    need_data_tag = num_subfiles != num_io_concentrators;
+    need_data_tag = (num_subfiles == 1) || (num_subfiles != num_io_concentrators);
     if (!need_data_tag)
         data_tag = READ_INDEP_DATA;
 
--- a/src/H5FDsubfiling/H5FDioc_threads.c
+++ b/src/H5FDsubfiling/H5FDioc_threads.c
@@ -456,8 +456,9 @@ translate_opcode(io_op_t op)
         case LOGGING_OP:
             return "LOGGING_OP";
             break;
+        default:
+            return "unknown";
     }
-    return "unknown";
 }
 #endif
 
@@ -873,9 +874,14 @@ ioc_file_queue_read_indep(sf_work_request_t *msg, int ioc_idx, int source, MPI_C
      * unpredictable order. However, if some IOCs own more than
      * 1 subfile, we need to associate each read with a unique
      * message tag to make sure the data is received in the
-     * correct order.
+     * correct order. We also need a unique message tag in the
+     * case where only 1 subfile is used in total. In this case,
+     * vector I/O calls are passed directly down to this VFD without
+     * being split up into multiple I/O requests, so we need the
+     * tag to distinguish each I/O request.
      */
-    need_data_tag = sf_context->sf_num_subfiles != sf_context->topology->n_io_concentrators;
+    need_data_tag = (sf_context->sf_num_subfiles == 1) ||
+                    (sf_context->sf_num_subfiles != sf_context->topology->n_io_concentrators);
     if (!need_data_tag)
         send_tag = READ_INDEP_DATA;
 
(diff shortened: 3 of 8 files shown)


src/H5FDsubfiling/H5FDioc.c
jhendersonHDF, line 1: The changes in this file mostly just add the size-extending feature for vector I/O to the IOC VFD. The `types` parameter is not used here, so that extending feature wasn't added.

src/H5FDsubfiling/H5FDioc_int.c
jhendersonHDF, line 1: Now that vector I/O is supported, vector I/O requests could get passed directly to the IOC VFD if there is only 1 subfile since it doesn't make since to split up the I/O requests across subfiles in that case. The change in this file addresses a new problem that came up with the new support for vector I/O.

src/H5FDsubfiling/H5FDioc_threads.c
jhendersonHDF, line 1: Same comment here about addressing vector I/O requests getting passed directly down to the IOC VFD

testpar/t_subfiling_vfd.c
```diff
@@ -40,7 +40,7 @@
 #define PATH_MAX 4096
 #endif
 
-#define DEFAULT_DEFLATE_LEVEL 9
+#define DEFAULT_DEFLATE_LEVEL 4
```
jhendersonHDF, line 43: Change the compression level from 9 to something a bit less aggressive since the tests were spending ~60% of the time just doing compression and that isn't the main focus for the tests.

testpar/t_subfiling_vfd.c
```diff
@@ -2360,11 +2360,33 @@ main(int argc, char **argv)
     if (MAINPROCESS)
         puts("");
 
+    if (MAINPROCESS)
```
jhendersonHDF, line 3056: Move the block that sets compression up so that the tests run once with compression before setting environment variables and once after setting them. Previously, the compression tests were running twice only after the environment variables had been set.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -1535,199 +1565,16 @@ static herr_t
...
+    H5FD_subfiling_t *file_ptr  = (H5FD_subfiling_t *)_file;
+    herr_t            ret_value = SUCCEED;
 
-    assert(file_ptr && file_ptr->pub.cls);
+    assert(file_ptr);
+    assert(file_ptr->pub.driver_id == H5FD_SUBFILING);
     assert(buf);
 
-    /* Check for overflow conditions */
-    if (!H5_addr_defined(addr))
```
jhendersonHDF, line 1555: All the logic between read and write was mostly duplicated, so I moved it into `H5FD__subfiling_io_helper`

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2086,98 +1709,27 @@ H5FD__subfiling_read_vector(H5FD_t *_file, hid_t dxpl_id, uint32_t count, H5FD_m
     /* Set DXPL for operation */
     H5CX_set_dxpl(dxpl_id);
 
-    /* TODO: setup real support for vector I/O */
     if (file_ptr->fa.require_ioc) {
-
-        bool       extend_sizes = false;
```
jhendersonHDF, line 2092: All this logic is now moved into the part of `H5FD__subfiling_io_helper` that handles the vector I/O requests.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2581,138 +2043,913 @@ H5FD__subfiling_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void
...
+
+        /*************************************
+         * Striping across multiple subfiles *
+         *************************************/
+
+        /*
+         * Generate the types, addrs, sizes and bufs I/O vectors for
+         * this I/O request.
+         */
+        status = generate_io_vectors(
```
jhendersonHDF, line 2200: The old `init_indep_io` function has been replaced by `generate_io_vectors`, which can now handle translating more than one (offset, I/O size, type, buffer) tuple into a set of I/O vectors that spans the subfiles. The new function is also a bit more efficient in that it directly generates the I/O vectors rather than generating lists of offset, size, types, buffer arrays that are used to populate I/O vectors later like before.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2581,138 +2043,913 @@ H5FD__subfiling_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void
...
+ *
+ *              io_bufs (OUT)
+ *                - pointer to the I/O buffers I/O vector to populate
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+translate_io_req_to_iovec(subfiling_context_t *sf_context, size_t iovec_idx, size_t iovec_len,
```
jhendersonHDF, line 2921: Much of the logic in this function is unchanged from the previous function it was in (`init_indep_io`). The main difference is that the buffer indexing had to change since the I/O vectors used for the final I/O requests are populated directly here rather than generating arrays that the vectors would have later been populated with.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -3020,20 +3311,21 @@ init_indep_io(subfiling_context_t *sf_context, int64_t file_offset, size_t io_ne
  *-------------------------------------------------------------------------
  */
 static herr_t
-iovec_fill_first(subfiling_context_t *sf_context, int64_t iovec_depth, int64_t target_datasize,
-                 int64_t start_mem_offset, int64_t start_file_offset, int64_t first_io_len,
-                 int64_t *mem_offset_out, int64_t *target_file_offset_out, int64_t *io_block_len_out)
+iovec_fill_first(subfiling_context_t *sf_context, size_t iovec_len, int64_t cur_iovec_depth,
```
jhendersonHDF, line 3338: The changes in the `iovec_fill_` functions below are just changes to the buffer indexing for generating I/O vectors directly.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2581,138 +2043,913 @@ H5FD__subfiling_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void
...
+                        extend_sizes = true;
+                        size         = sizes[i - 1];
+                    }
+                    else {
+                        size = sizes[i];
+                    }
+                }
+
+                H5_CHECK_OVERFLOW(size, size_t, int);
+                if (MPI_SUCCESS != MPI_Bcast(bufs[i].vp, (int)size, MPI_BYTE, 0, file_ptr->comm))
```
fortnern, line 2326: Maybe not the focus of this PR but how hard would it be to accumulate these into a single bcast?

jhendersonHDF: Indeed, this was mostly just copied over from the previous code, but it could probably be handled better with a derived type instead of multiple bcasts.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2581,138 +2043,913 @@ H5FD__subfiling_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void
...
+            H5_SUBFILING_GOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL,
+                                    "can't allocate copy of I/O buffers array");
+
+        extend_types = false;
+        extend_sizes = false;
+        for (size_t i = 0; i < count; i++) {
+            if (!extend_types) {
+                if ((i > 0) && (types[i] == H5FD_MEM_NOLIST)) {
+                    extend_types = true;
+                    type         = types[i - 1];
```
fortnern, line 2441: Seems silly to use the type variable for type here when it can only be one thing, though I guess it makes the code more similar to other places. Not that it makes a huge difference

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2581,138 +2043,913 @@ H5FD__subfiling_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void
...
+    /* Only mirror I/O from MPI rank 0 */
+    if (file_ptr->mpi_rank != 0)
+        H5_SUBFILING_GOTO_DONE(SUCCEED);
+
+    if (count == 0)
+        H5_SUBFILING_GOTO_DONE(SUCCEED);
+
+    for (size_t i = 0; i < count; i++) {
+        if (!extend_types) {
+            if ((i > 0) && (types[i] == H5FD_MEM_NOLIST)) {
```
fortnern, line 2390: I don't think we need to keep scanning the list after we find H5FD_MEM_NOLIST

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2581,138 +2043,913 @@ H5FD__subfiling_ctl(H5FD_t *_file, uint64_t op_code, uint64_t flags, const void
...
+        else if (data_size < stripe_size) {
+            /*
+             * I/O of a size smaller than the stripe size could
+             * touch one or two subfiles at most, depending on
+             * the file offset.
+             */
+            cur_max_num_subfiles = 2;
+        }
+        else {
+            cur_max_num_subfiles = data_size / stripe_size;
```
fortnern: I'm not sure this calculation is correct. For example if the block starts at a strip boundary, the stripe size is 1000, and the data size is 2001, this will return 2 subfiles, when it really touches 3. And I think it needs to take into account the actual starting position, not just whether it's on a boundary, to avoid wasting space.

I think something like `(((cur_file_offset % stripe_size) + data_size - 1) / stripe_size) + 1` could work

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2792,212 +3027,268 @@ init_indep_io(subfiling_context_t *sf_context, int64_t file_offset, size_t io_ne
...
-                    if (num_full_stripes)
-                        num_full_stripes--;
-                }
+                thin_uniform_section = (subfile_idx > first_subfile_idx) && (subfile_idx > last_subfile_idx);
             }
 
-            /* Account for subfiles with uniform segments */
-            if (!is_first && !is_last) {
-                bool thin_uniform_section = false;
+            if (last_subfile_idx < first_subfile_idx) {
```
fortnern: Should this be else if? I don't think both could be true.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2792,212 +3027,268 @@ init_indep_io(subfiling_context_t *sf_context, int64_t file_offset, size_t io_ne
...
+                 * than both the starting subfile and ending subfile
+                 * indices, it is a "thinner" section with a smaller
+                 * I/O vector depth.
                  */
-                if (final_length < stripe_size) {
-                    subfile_bytes += final_length;
-                    if (num_full_stripes)
-                        num_full_stripes--;
-                }
+                thin_uniform_section = (subfile_idx > first_subfile_idx) && (subfile_idx > last_subfile_idx);
```
fortnern: Wouldn't this also be the case if it was below both? Also don't you only need to check if it's greater than last_subfile_idx? (And if it's less than first_subfile_idx in the case I mentioned)

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -2792,212 +3027,268 @@ init_indep_io(subfiling_context_t *sf_context, int64_t file_offset, size_t io_ne
...
-            if (!is_first && !is_last) {
-                bool thin_uniform_section = false;
+            if (last_subfile_idx < first_subfile_idx) {
+                /*
+                 * This can also happen when the subfile with the final
+                 * data segment has a smaller subfile index than the
+                 * subfile with the first data segment and the current
+                 * subfile index falls between the two.
+                 */
+                thin_uniform_section = thin_uniform_section || ((last_subfile_idx < subfile_idx) &&
```
fortnern: Related to the above comment - I don't think you need the || here

src/H5FDsubfiling/H5FDsubfiling.c
jhendersonHDF: The below `if (cur_file_offset % stripe_size)` is responsible for adding 1 to the value when the offset is aligned to a stripe boundary. I was less concerned about wasting space since an extra entry in the arrays would make very little difference.

src/H5FDsubfiling/H5FDsubfiling.c
jhendersonHDF: Yes (or just an `else`), it was just easier for me to read it this way at the time of writing.

src/H5FDsubfiling/H5FDsubfiling.c
jhendersonHDF: Agreed. This was likely a leftover from fixing a bug a long time ago.

src/H5FDsubfiling/H5FDsubfiling.c
fortnern: In the case above this code results in an array that's too small

jhendersonHDF: I see that now. I couldn't immediately find a case where the above suggestion is problematic, so I'll switch to that instead.

src/H5FDsubfiling/H5FDsubfiling.c
jhendersonHDF: > Wouldn't this also be the case if it was below both?

After thinking about this for a while and drawing some diagrams, I believe this is the case, yes, though it should just populate the I/O vectors with data that isn't used. Let me write a test for this case and verify, then I'll fix this.

> Also don't you only need to check if it's greater than last_subfile_idx?

Ah, yes that follows from the logic of `if (last_subfile_idx >= first_subfile_idx)`. Good catch.

fortnern:
Done reviewing, looks good aside from my comments. We probably don't need to implement the more efficient rank 0 bcast right now but we should make a note of it.

src/H5FDsubfiling/H5FDsubfiling.c
jhendersonHDF: For now, I've just left this the same since it matches other places and was easier for me to reason about.

jhendersonHDF:
@fortnern I believe I've addressed the most important comments from your review. I've added test cases that test for all the cases discussed in the comments; the newly-added max number of subfiles calculation was indeed problematic, but the test passes after your suggested change. The "thin uniform section" calculation wasn't exactly problematic, but I added a test for it anyway just to be sure.

I haven't implemented the performance improvement for the rank0 bcast strategy yet, but I can add a comment to the code about that or document it somewhere else if you think that's better.

fortnern:
I think we can just create an issue for the rank 0 bcast improvement for now

fortnern (approved):

lrknox (approved):

PR #3900, opened 2023-12-20
Disable doxygen as errors for netcdf

byrnHDF:
Do not build doxygen in CI.  NetCDF doxygen has a doxygen warning treated as error, which causes this particular test to fail.  Not having found an option to disable the error for a warning, we won't enable doxygen in the NetCDF build and leave testing it to NetCDF.


byrnHDF:
This needs to be a netcdf option, not a hdf5 option.

lrknox (approved):

hyoklee (approved):

PR #3909, opened 2023-12-23
Add Doxygen documentation for H5FD module and h5tools.

hyoklee:
This will help tool users by exposing h5dump VFD options.

20231222_h5tools_doxygen

20231222_h5tools_doxygen_drivernames

Once this PR is merged, add the link to https://portal.hdfgroup.org/documentation/hdf5-docs/registered_virtual_file_drivers_vfds.html for #3878.

P.S. H5FD was missing so I added it as well.


byrnHDF (changes requested):
This basically exposes the tools API, which we have avoided because of a lack of design and resources. This "publish the API" needs a company directive.
 In addition this adds a "C" source file to the doxygen. Also, we should not add just the C library APIs for FD package without the adding the other languages.

lrknox:
> This basically exposes the tools API, which we have avoided because of a lack of design and resources. This "publish the API" needs a company directive. 
@derobins What do you think?

jhendersonHDF:
I think we definitely need something like this, but also think it should just be a table in the H5FD module file doxygen. I don't think it makes sense to tie it to the tools, especially since these changes are just exposing an implementation detail.

derobins:
We should expose the names, but not in this way. I'm going to go ahead and close this.

@hyoklee - Please create an issue for this and add it to the 1.14.4 milestone.

PR #3915, opened 2024-01-02
CMake: Disable UNITY_BUILD

jschueller:
Useful to avoid compilation errors from a top-level project that uses UNITY_BUILD, eg vtk.
UNITY_BUILD seems error-prone to support properly here because of the macros, lets disable it for most targets.


byrnHDF:
This might be better to disable it globally instead of for individual targets;
If no explicit UNITY_BUILD_MODE has been specified, CMake will default to BATCH. However;
CMAKE_UNITY_BUILD variable is used to initialize the UNITY_BUILD property of targets when they are created. By default, this variable is not set, which will result in unity builds being disabled.

I'm guessing that if hdf5 is within another project that sets UNITY_BUILD is the reason for this. I think we need to use the HDF5_EXTERNALLY_CONFIGURED variable perhaps to disable it for hdf5?

jschueller:
no need, it can be disabled inconditionnally until someone wants to add support for it;
if you set CMAKE_UNITY_BUILD global variable even inside hdf5 the build will fail too

byrnHDF:
Creating different PR#4515 to handle this globally.

PR #3920, opened 2024-01-03
Fix building VOLs under library with CMake <3.24

mattjala:
CMake versions before 3.24 [don't have](https://cmake.org/cmake/help/latest/module/FetchContent.html#commands) the `FIND_PACKAGE_ARGS` option for `FetchContent_Declare`, and this line is only used for CMake <3.24.


jhendersonHDF (approved):

CMakeVOL.cmake
```diff
@@ -183,7 +183,6 @@ if (HDF5_VOL_ALLOW_EXTERNAL MATCHES "GIT" OR HDF5_VOL_ALLOW_EXTERNAL MATCHES "LO
           FetchContent_Declare (${hdf5_vol_depname}
             GIT_REPOSITORY "${HDF5_VOL_SOURCE}"
             GIT_TAG "${HDF5_VOL_${hdf5_vol_name_upper}_BRANCH}"
-            FIND_PACKAGE_ARGS NAMES ${hdf5_vol_name_lower}
```
byrnHDF, line 186: Could you just create a var that is empty for < 3.24 and has content otherwise?

mattjala: Done

byrnHDF (approved):

PR #3927, opened 2024-01-05
Add citation file to specify how this repository wants to be cited

senden9:
Hi!

I created a `CITATION.cff` to make it more obvious how this software wants to be cited. The `.cff` format is understood by GitHub (show how to cite on the sidebar of the repo), Zenodo, and Zotero.

The information in that file is based on http://web.archive.org/web/20230610185232/https://portal.hdfgroup.org/display/knowledge/How+do+I+properly+cite+HDF5%2C+HDFView%2C+and+HSDS+in+a+paper which is the lastest source of information I found.
The file itself is created by filling that information into https://citation-file-format.github.io/cff-initializer-javascript/.

Currently, the file contains only the website as a source, which is ok! A lot of scientific software wants to be cited with a specific paper combined. By having this file here anyone who wants to give credit to HDF5 in a publication knows that the preferred way is by linking to the project website.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,12 @@
+cff-version: 1.2.0
+title: 'Hierarchical Data Format, version 5'
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - name: The HDF Group
+    website: 'https://www.hdfgroup.org'
+repository-code: 'https://github.com/HDFGroup/hdf5'
+url: 'https://www.hdfgroup.org/HDF5/'
+repository-artifact: 'https://www.hdfgroup.org/downloads/hdf5/'


senden9:
If you click on my fork you can see how this file is rendered by GitHub
grafik

If you do not like the `.cff` stuff alternatively a BibTex file would be doable.

caiohamamura:
> If you do not like the `.cff` stuff alternatively a BibTex file would be doable.

I think this is the preferred way, since GitHub will already provide the BibTeX file derived from the CITATION.cff.

loricooperhdf (approved):

jhendersonHDF (approved):

senden9:
So I have two approvals (from @jhendersonHDF and @loricooperhdf) and one change request by @caiohamamura. I personally do not care in which format the information is provided as long as it is clear.

Pro BibTeX: The format the most people want to use in the end I assume.
Pro .cff: GitHub renders this file directly; it is easy parsable; GitHub generates a BibTeX format automatically from that file; can be converted in different formats automatically.

In the end it comes down to what the where the people find the code I think. If we see GitHub as the target platform then the cff file wins (in my opinion). If we see just the git repo without GitHub I think a BibTex file would be better since people could use it directly. In my opinion people will use GitHub to interact with a codebase that they do not contribute to, so cff wins.

caiohamamura:
Not actually, I'm not asking for change, I call for keeping cff, the author itself said it could be otherwise, but I said we should stick with cff since the GitHub platform will also provide the conversion

senden9:
Ah, ok. Then I misunderstood your comment 👌

glennsong09 (approved):

nikosarcevic:
beautiful that this was merged 9 hours before I was trying to find how to cite :)

Issue #3316, opened 2023-08-01
H5Fget_obj_count for all files counts copy of transient datatypes


mattjala:
**Describe the bug**
If a copy of a transient datatype exists when H5F_get_obj_count is called with file_id = H5F_OBJ_ALL, H5Fget_obj_count will count that datatype as an opened object.

**Expected behavior**
H5Fget_obj_count should only count committed datatypes that are currently open. 

**Platform (please complete the following information)**
 - HDF5 version 1.14.1, 1.10.10
 - Ubuntu 22.04
 - gcc 11.3.0
 - CMake 3.22.1
 - Serial

**Additional context**
Issue elicited with the test added in #3313, specifically:
```
#define DATATYPE_OBJ_COUNT_TRANSIENT_DATATYPE H5T_NATIVE_INT
    ...
    type_id3 = H5Tcopy(DATATYPE_OBJ_COUNT_TRANSIENT_DATATYPE);
    ...
    obj_count = H5Fget_obj_count(H5F_OBJ_ALL, H5F_OBJ_DATATYPE);
    VERIFY(obj_count, 2, "H5Fget_obj_count");

    /* Create dataset with copy of transient datatype */
    dset_id4 = H5Dcreate2(file_id1, DATATYPE_OBJ_COUNT_DATASET_NAME_4, type_id3, dspace_id, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);

     ...
    obj_count = H5Fget_obj_count(H5F_OBJ_ALL, H5F_OBJ_DATATYPE);
    VERIFY(obj_count, 2, "H5Fget_obj_count");

    obj_count = H5Fget_obj_count(H5F_OBJ_ALL, H5F_OBJ_ALL);
    VERIFY(obj_count, 8, "H5Fget_obj_count");
```

Which result in the errors:
```
Testing  -- Low-Level File I/O (file) 
*** UNEXPECTED VALUE from H5Fget_obj_count should be 2, but is 3 at line 1318 in /home/matthewlarson/Downloads/hdf5/test/tfile.c
*** UNEXPECTED VALUE from H5Fget_obj_count should be 2, but is 3 at line 1328 in /home/matthewlarson/Downloads/hdf5/test/tfile.c
*** UNEXPECTED VALUE from H5Fget_obj_count should be 8, but is 9 at line 1331 in /home/matthewlarson/Downloads/hdf5/test/tfile.c
```

derobins:
[...]

PR #3928, opened 2024-01-05
H5Fget_obj_ids/count no longer count copies of transient types

mattjala:
When using `H5Fget_obj_ids`/`H5Fget_obj_count` over all files, copies of transient datatypes are considered open object identifiers. Transient datatypes don't exist on any file, don't need to be closed, and the documentation specifically states that only committed datatypes are returned,  so this behavior doesn't make sense.

The Fortran mounting test previously used copies of transient datatypes to test object counting in mounted files. It now uses groups for that purpose. 

This could be considered a change to the behavior of API functions, so I added a note to RELEASE.txt.

Addresses #3316

### DIFF
--- a/src/H5F.c
+++ b/src/H5F.c
@@ -200,12 +200,13 @@ H5F__get_all_count_cb(void H5_ATTR_UNUSED *obj_ptr, hid_t obj_id, void *key)
 {
     H5F_trav_obj_cnt_t *udata        = (H5F_trav_obj_cnt_t *)key;
     H5I_type_t          obj_type     = H5I_UNINIT;
+    H5T_t              *type         = NULL;
     htri_t              is_committed = FAIL;
     int                 ret_value    = H5_ITER_CONT; /* Return value */
 
     FUNC_ENTER_PACKAGE
 
-    if ((obj_type = H5Iget_type(obj_id)) < 0)
+    if ((obj_type = H5I_get_type(obj_id)) < 0)
         HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, H5_ITER_ERROR, "object has invalid id type");
 
     if ((obj_type != H5I_DATATYPE)) {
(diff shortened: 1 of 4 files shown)


src/H5F.c
```diff
@@ -196,15 +196,31 @@ H5Fget_access_plist(hid_t file_id)
...
-    int                 ret_value = H5_ITER_CONT; /* Return value */
+    H5F_trav_obj_cnt_t *udata        = (H5F_trav_obj_cnt_t *)key;
+    H5I_type_t          obj_type     = H5I_UNINIT;
+    htri_t              is_committed = FAIL;
+    int                 ret_value    = H5_ITER_CONT; /* Return value */
 
-    FUNC_ENTER_PACKAGE_NOERR
+    FUNC_ENTER_PACKAGE
+
+    if ((obj_type = H5Iget_type(obj_id)) < 0)
```
qkoziol: I think this is a bug and should be corrected.  However, the library internal routines should never (with *very* few exceptions) call API routines.  I'll take a look at which internal routines to use later tonight.  (If someone else doesn't get to it before me :-) ).

qkoziol (changes requested):
Shouldn't call API routines from internal library routines.

src/H5F.c
mattjala: Replaced `H5Iget_type` -> `H5I_get_type`,  `H5Tcommitted` -> `H5I_object_verify` + `H5T_is_named`

brtnfld:
It seems inconsistent that an object is not considered "technically" open (and thus not counted), yet it can still be closed using the H5Tclose function. Additionally, a copied type should be closed even if not created.

This is similar to the debate on whether the objects opened through an external link should be counted, as they are also considered "transient." ref. HDFFV-11231.

Regardless, I'm assuming most applications use H5Fget_obj_count to get a count of objects that need to be closed, including H5Tcopy ids. Besides, we currently don't provide a way to detect these types of open objects ( property lists, datatypes, etc...) otherwise, which is another issue.

mattjala:
> Regardless, I'm assuming most applications use H5Fget_obj_count to get a count of objects that need to be closed, including H5Tcopy ids. Besides, we currently don't provide a way to detect these types of open objects ( property lists, datatypes, etc...) otherwise, which is another issue.

This is true - these changes should be paired with some kind of API addition that does return all ids that need to be closed, whether that's a new API call or a new option for these calls. That would be for the next release, since it's definitely an API change. For now, I'll mark this as draft and update the current documentation for `H5Fget_obj_ids/H5Fget_obj_count` to reflect how they currently work (#3932).

mattjala:
For now, this work is being moved to the [h5f_get_obj_rework feature branch](/HDFGroup/hdf5/tree/feature/h5f_get_obj_rework).

PR #3935, opened 2024-01-08
Fix for HDFFV-9174

vchoi-hdfgroup:


brtnfld:
Why bother with new fixes for 1.10 if it is no longer maintained?

derobins:
We no longer maintain 1.10

PR #3939, opened 2024-01-10
Remove incorrect include of sz header

byrnHDF:


lrknox (approved):
Build errors on vortex at Sandia are fixed by this change.

jhendersonHDF (approved):

PR #3940, opened 2024-01-11
Sync API tests with vol-tests

mattjala:
Includes changes from the following PRs:

HDFGroup/vol-tests#58 (H5Ovisit tests)
HDFGroup/vol-tests#60 (Fill value tests)
HDFGroup/vol-tests#62 (int-to-real conversion tests)
HDFGroup/vol-tests#63 (Additional fill value test)
HDFGroup/vol-tests#64 (String encoding tests)
HDFGroup/vol-tests#65 (user-defined filter test)


test/API/H5_api_dataset_test.h
```diff
@@ -23,6 +23,13 @@ int H5_api_dataset_test(void);
  *                                              *
  ************************************************/
 
+/* Check if the relative difference is smaller than a
+ *  predefined value M.  See if two values are relatively equal.
+ *  It's the developer's responsibility not to pass in the value 0, which
+ *  may cause the equation to fail.
+ */
+#define VL_DBL_REL_EQUAL(X, Y, M) (fabs(((Y) - (X)) / (X)) < (M))
```
jhendersonHDF: Now that these tests will be back in the library, you should be able to use the equivalent `H5_DBL_REL_EQUAL` macro from H5private.h rather than creating a new one here.

test/API/H5_api_dataset_test.c
```diff
@@ -2711,19 +3127,23 @@ test_create_dataset_creation_properties(void)
...
-        H5Tclose(compact_dtype);
-        H5Tclose(dset_dtype);
-        H5Dclose(dset_id);
-        H5Pclose(dcpl_id);
-        H5Gclose(group_id);
-        H5Gclose(container_group);
-        H5Fclose(file_id);
-    }
-    H5E_END_TRY
+        if (read_buf) {
```
jhendersonHDF, line 3130: Looks like an extra `{` got inserted here, causing all the close calls below to only be called if read_buf is non-NULL here.

test/API/H5_api_dataset_test.c
```diff
@@ -4487,8 +4907,8 @@ test_read_multi_dataset_small_all(void)
     for (i = 0; i < DATASET_MULTI_COUNT; i++) {
         char dset_name[DSET_NAME_BUF_SIZE];
 
-        if (DSET_NAME_BUF_SIZE <= snprintf(dset_name, DSET_NAME_BUF_SIZE, "%s%zu%c",
-                                           DATASET_SMALL_READ_TEST_ALL_DSET_NAME, i, '\0'))
+        if (DSET_NAME_BUF_SIZE <= (unsigned int)snprintf(dset_name, DSET_NAME_BUF_SIZE, "%s%zu%c",
```
jhendersonHDF: It doesn't seem like a good idea in general to cast the return value of snprintf here to an unsigned since snprintf could return a negative value and at least in theory could end up not causing an error. I see that `DSET_NAME_BUF_SIZE` is defined as 64u, which I'm guessing is the reason for the case here, but it's better to just have `#define DSET_NAME_BUF_SIZE 64` so we don't need the cast here, or at least assign the return value of snprintf to an int and check for a negative value before casting to an unsigned type.

test/API/H5_api_dataset_test.c
```diff
@@ -4632,8 +5052,9 @@ test_read_multi_dataset_small_hyperslab(void)
     for (i = 0; i < DATASET_MULTI_COUNT; i++) {
         char dset_name[DSET_NAME_BUF_SIZE];
 
-        if (DSET_NAME_BUF_SIZE <= snprintf(dset_name, DSET_NAME_BUF_SIZE, "%s%zu%c",
-                                           DATASET_SMALL_READ_TEST_HYPERSLAB_DSET_NAME, i, '\0'))
+        if (DSET_NAME_BUF_SIZE <= (unsigned int)snprintf(dset_name, DSET_NAME_BUF_SIZE, "%s%zu%c",
```
jhendersonHDF: Same comment here about the unsigned cast (and probably other places in the code as well).

jhendersonHDF:
Just a couple small issues, but otherwise looks fine

jhendersonHDF (approved):

lrknox (approved):

PR #3942, opened 2024-01-17
Remove cached datatype conversion path table entries on file close

### LINKED ISSUE #1256: h5py (+PyTorch) memory leak since hdf5 1.12.1
In h5py, the Python memory explodes when an array is loaded from an h5py object and converted to a PyTorch tensor, but only if you also retrieve an attribute. See the original issue: h5py/h5py#2010.

I am not entirely sure if this is an h5py issue or a problem in libhdf5, but it is connected to this change in hdf5: 98a27f1fe4f0cb96ce7fdeee2e95079ae19db5cd (between libhdf5 1.12.0 and 1.12.1). Before this change, the leak does not appear, after, it does.

Example
---------
On my system, I can produce this error by running `pip install h5py numpy torch` and then running this script:
```python
import h5py
import numpy as np
import torch

with h5py.File('test.h5', 'w') as ds:
    ds.create_dataset('scan', data=np.ones(shape=(1000, 1000)))
    ds['scan'].attrs['example'] = 'abcdefg'

def load_leaky():
    with h5py.File('test.h5', 'r') as ds:
        x_i = ds['scan'][:]
        ds['scan'].attrs['example']
    torch.tensor(x_i)

while True:
    load_leaky()
```
Running this will cause the Python memory to explode. All three steps are necessary (loading an array + reading an attribute + converting the array to PyTorch) to create this issue.

Bisection result
---------
I repeated this with self-compiled h5py 3.6.0 with libdfh5 compiled from git between 1.12.0 and 1.12.1. The problem depends on the hdf5 version: 1.12.0 works, 1.12.1 doesn't, and 98a27f1fe4f0cb96ce7fdeee2e95079ae19db5cd is the first commit where it goes wrong.

Is this an hdf5 issue or is this something in h5py?

cc @tacaswell

jhendersonHDF:
When performing datatype conversions during I/O, the library checks to see whether it can re-use a cached datatype conversion pathway by performing comparisons between the source and destination datatypes of the current operation and the source and destination datatypes associated with each cached datatype conversion pathway. For variable-length and reference datatypes, a comparison is made between the VOL object for the file associated with these datatypes, which may change as a file is closed and reopened. In workflows involving a loop that opens a file, performs I/O on an object with a variable-length or reference datatype and then closes the file, this can lead to constant memory usage growth as the library compares the file VOL objects between the datatypes as different and adds a new cached conversion pathway entry on each iteration during I/O. This is now fixed by clearing out any cached conversion pathway entries for variable-length or reference datatypes associated with a particular file when that file is closed.

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -343,7 +343,6 @@ typedef H5T_t *(*H5T_copy_func_t)(H5T_t *old_dt);
 static herr_t H5T__register_int(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst,
                                 H5T_lib_conv_t func);
 static herr_t H5T__register(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_conv_func_t *conv);
-static herr_t H5T__unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_conv_t func);
 static htri_t H5T__compiler_conv(H5T_t *src, H5T_t *dst);
 static herr_t H5T__set_size(H5T_t *dt, size_t size);
 static herr_t H5T__close_cb(H5T_t *dt, void **request);
@@ -2672,7 +2671,7 @@ done:
 } /* end H5Tregister() */
 
 /*-------------------------------------------------------------------------
- * Function:   H5T__unregister
+ * Function:   H5T_unregister
  *
  * Purpose:    Removes conversion paths that match the specified criteria.
  *        All arguments are optional. Missing arguments are wild cards.
@@ -2683,15 +2682,16 @@ done:
  *
  *-------------------------------------------------------------------------
  */
-static herr_t
-H5T__unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_conv_t func)
+herr_t
+H5T_unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5VL_object_t *owned_vol_obj,
+               H5T_conv_t func)
 {
     H5T_path_t *path   = NULL; /*conversion path             */
     H5T_soft_t *soft   = NULL; /*soft conversion information */
     int         nprint = 0;    /*number of paths shut down   */
     int         i;             /*counter                     */
 
-    FUNC_ENTER_PACKAGE_NOERR
+    FUNC_ENTER_NOAPI_NOERR
 
     /* Remove matching entries from the soft list */
     if (H5T_PERS_DONTCARE == pers || H5T_PERS_SOFT == pers) {
@@ -2704,6 +2704,8 @@ H5T__unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_c
                 continue;
             if (dst && dst->shared->type != soft->dst)
                 continue;
+            if (owned_vol_obj)
+                continue;
             if (func && func != soft->conv.u.app_func)
                 continue;
 
@@ -2714,13 +2716,20 @@ H5T__unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_c
 
     /* Remove matching conversion paths, except no-op path */
     for (i = H5T_g.npaths - 1; i > 0; --i) {
+        bool nomatch;
+
         path = H5T_g.path[i];
         assert(path);
 
+        nomatch = ((H5T_PERS_SOFT == pers && path->is_hard) || (H5T_PERS_HARD == pers && !path->is_hard)) ||
+                  (name && *name && strcmp(name, path->name) != 0) ||
+                  (src && H5T_cmp(src, path->src, false)) || (dst && H5T_cmp(dst, path->dst, false)) ||
+                  (owned_vol_obj && (owned_vol_obj != path->src->shared->owned_vol_obj) &&
+                   (owned_vol_obj != path->dst->shared->owned_vol_obj)) ||
+                  (func && func != path->conv.u.app_func);
+
         /* Not a match */
-        if (((H5T_PERS_SOFT == pers && path->is_hard) || (H5T_PERS_HARD == pers && !path->is_hard)) ||
-            (name && *name && strcmp(name, path->name) != 0) || (src && H5T_cmp(src, path->src, false)) ||
-            (dst && H5T_cmp(dst, path->dst, false)) || (func && func != path->conv.u.app_func)) {
+        if (nomatch) {
             /*
              * Notify all other functions to recalculate private data since some
              * functions might cache a list of conversion functions.  For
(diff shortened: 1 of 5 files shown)


jhendersonHDF:
This PR adds a simple fix for the problem described in #1256. The main problem behind the issue is that variable-length and reference datatypes are compared according to their associated file VOL object, which may change as a file is closed and reopened. When attempting to reuse type conversion pathways in an application that has a file open -> I/O -> file close loop, the library will then compare datatypes in the current conversion operation as different from the datatypes in the cached pathway table, causing the caching table to continually grow as new entries are added until the H5T package is torn down at library termination. Further, since variable-length and reference datatypes hold a reference to their associated file's VOL object after https://github.com/HDFGroup/hdf5/pull/274, the memory for the VOL object will also be held around until the library terminates. Together, this can eventually lead to a drastic explosion of memory usage.

Several approaches for fixing this were proposed, including having a special conversion path table just for variable-length and reference datatypes, but this fix seems the most straightforward with existing library infrastructure. The downside is that it adds a scan of the cached conversion pathway table on every file close operation. If this still seems workable, I'll add a RELEASE note entry for the fix; Otherwise, we should discuss a better solution.

jhendersonHDF:
Example program that demonstrates the issue:

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#include "hdf5.h"

int
main(int argc, char **argv)
{
    const char *buf[] = {"attr_value"};
    hid_t       file_id  = H5I_INVALID_HID;
    hid_t       attr_id  = H5I_INVALID_HID;
    hid_t       str_type = H5I_INVALID_HID;
    hid_t       space_id = H5I_INVALID_HID;
    int        *irbuf    = NULL;
    char       **rbuf    = NULL;

    printf("Test memory leak\n");

    file_id = H5Fcreate("file.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
    if (file_id == H5I_INVALID_HID) {
        printf("Couldn't create file\n");
        exit(EXIT_FAILURE);
    }

    str_type = H5Tcopy(H5T_C_S1);
    if (str_type == H5I_INVALID_HID) {
        printf("Couldn't create string type\n");
        exit(EXIT_FAILURE);
    }

    if (H5Tset_size(str_type, H5T_VARIABLE) < 0) {
        printf("Couldn't set string size\n");
        exit(EXIT_FAILURE);
    }

    space_id = H5Screate(H5S_SCALAR);
    if (space_id == H5I_INVALID_HID) {
        printf("Couldn't create dataspace\n");
        exit(EXIT_FAILURE);
    }

    attr_id = H5Acreate2(file_id, "attribute", str_type, space_id, H5P_DEFAULT, H5P_DEFAULT);
    if (attr_id == H5I_INVALID_HID) {
        printf("Couldn't create attribute\n");
        exit(EXIT_FAILURE);
    }

    if (H5Awrite(attr_id, str_type, buf) < 0) {
        printf("Couldn't write to attribute\n");
        exit(EXIT_FAILURE);
    }

    H5Aclose(attr_id);
    H5Fclose(file_id);

    irbuf = malloc(1000 * 1000 * sizeof(int));
    rbuf  = malloc(sizeof(char *));

    /* for (size_t i = 0; i < 10; i++) { */
    while (true) {
        file_id = H5Fopen("file.h5", H5F_ACC_RDONLY, H5P_DEFAULT);
        if (file_id == H5I_INVALID_HID) {
            printf("Couldn't open file\n");
            exit(EXIT_FAILURE);
        }

        attr_id = H5Aopen(file_id, "attribute", H5P_DEFAULT);
        if (attr_id == H5I_INVALID_HID) {
            printf("Couldn't open attribute\n");
            exit(EXIT_FAILURE);
        }

        if (H5Aread(attr_id, str_type, rbuf) < 0) {
            printf("Couldn't read from attribute\n");
            exit(EXIT_FAILURE);
        }

        H5Treclaim(str_type, space_id, H5P_DEFAULT, rbuf);

        H5Aclose(attr_id);
        H5Fclose(file_id);
    }

    free(irbuf);
    free(rbuf);

    H5Tclose(str_type);
    H5Sclose(space_id);

    return 0;
}
```

src/H5T.c
```diff
@@ -2704,6 +2704,8 @@ H5T__unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_c
                 continue;
             if (dst && dst->shared->type != soft->dst)
                 continue;
+            if (owned_vol_obj)
```
jhendersonHDF: We don't want to remove the vlen conversion from the soft conversions table here, so always continue on if the `owned_vol_obj` parameter is specified. Ideally, the logic for removing the soft conversion functions would be separate from removing the cached conversion pathway table entries, but this seems minimally invasive.

qkoziol: The entire for loop can be skipped if owned_vol_obj is non-NULL.  (So you could add that check to the surrounding if statement, which will reduce the performance impact a little)

src/H5T.c
```diff
@@ -2714,13 +2716,20 @@ H5T__unregister(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_c
 
     /* Remove matching conversion paths, except no-op path */
     for (i = H5T_g.npaths - 1; i > 0; --i) {
+        bool nomatch;
+
         path = H5T_g.path[i];
         assert(path);
 
+        nomatch = ((H5T_PERS_SOFT == pers && path->is_hard) || (H5T_PERS_HARD == pers && !path->is_hard)) ||
```
qkoziol: This has become increasingly unwieldy - a small "match" routine is justified?

qkoziol (approved):
I'm clicking 'approve', but believe that if you make the two small changes I've suggested, it would be better.  :-)

src/H5T.c
jhendersonHDF: Very true; will do

src/H5T.c
jhendersonHDF: Agreed that it's become unwieldy, though is the point of a "match" routine just to move the visual ugliness out of this routine?

qkoziol: Mostly, and possibly to re-use

qkoziol (commented):
Super, thanks for updating!   :-)

derobins (approved):

tacaswell:
Even with this merged, is it worth sorting out how to manually clear this memory in h5py?

jhendersonHDF:
> Even with this merged, is it worth sorting out how to manually clear this memory in h5py?

I think I need to better understand the proxy buffering mechanism in h5py first, but I believe that this is still a problem (at least for non-string variable-length types) since I can't see anywhere else in the h5py codebase that `H5Dvlen_reclaim`/`H5Treclaim` is called after a read operation. Since this PR should fix the original issue with memory growth and since the h5py docs seem to suggest avoiding variable-length types where possible, it seems like a lower priority issue, but I think it can still be important. I quickly threw together https://github.com/jhendersonHDF/h5py/commit/90048215f535810cdf55aa7f21c28d6bc60636ca as a start, but it seemed a bit unwieldy to me and I was also still working on creating an h5py example for the issue.

PR #3944, opened 2024-01-18
Replace off_t with HDoff_t internally

derobins:
off_t is a 32-bit signed value on Windows, so we should use HDoff_t (which is __int64 on Windows) internally instead.

Also defines HDftell on Windows to be _ftelli64().


derobins:
MSDN reference:

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/ftell-ftelli64?view=msvc-170

jhendersonHDF (approved):

mattjala (approved):

Issue #1682, opened 2022-04-24
Documentation - variable-length datatype description seems incomplete (doesn't mention global heap)


gringer:
I'm working on trying to decode an HDF5 file (attached) manually using [the specification](https://docs.hdfgroup.org/hdf5/develop/_f_m_t3.html#DataObject) (for the purpose of writing rust code for a decoder that quickly extracts raw signal data from nanopore FAST5 files), and am getting tripped up by the definition of variable-length data. I'm currently looking at the attribute message starting at location 0x14e62 in the file, which I've expanded out here in 8-byte chunks for clarity:

```
00014e62  0c 00 48 00 04 00 00 00  01 00 0d 00 14 00 08 00  ## attribute header / lengths
00014e72  66 69 6c 65 5f 76 65 72  73 69 6f 6e 00 00 00 00  ## attribute name
00014e82  19 01 01 00 10 00 00 00  10 00 00 00 01 00 00 00  ## datatype
00014e92  00 00 08 00 00 00 00 00  01 00 00 00 00 00 00 00  ## datatype / dataspace
00014ea2  03 00 00 00 00 08 00 00  00 00 00 00 73 00 00 00  ## dataspace
```
I can understand the first two lines of this:
 * Attribute message; 0x48 bytes excluding header; not shared; version 1; 0xd bytes for name; 0x14 bytes for datatype; 0x08 bytes for dataspace
 * Name [file_version\0] + padding to 8-byte boundary

The first 4 bytes of the third line define a v1 variable-length datatype; type string; null-terminated; with UTF-8 encoding. Beyond this I think the specification indicates a length should follow, so that's another 0x10 bytes for the datatype (consistent with the attribute information)... but then I get lost.

The datatype message specification suggests the next information that follows is "Properties", and for variable-length datatypes it states, "Each variable-length type is based on some parent type. The information for that parent type is described recursively by this field." Unfortunately, I don't understand what this means. The next four bytes in the datatype definition are `10 00 00 00`, and I can't find anything in the specification to help decode them. If I assume this is a variable class/version information segment, then I would expect it to decode to v1, fixed point. If I assume this is the property of a string, the specification tells me, "There are no properties defined for the string class." If I assume this is the property of an array, I end up with a dimensionality of 16, and there's not enough space in the datatype to define 16 dimensions.

[...]

PR #3950, opened 2024-01-22
Clarify vlen string message documentation

mattjala:
Clarifies the issues raised in #1682

### DIFF
--- a/doxygen/examples/H5.format.html
+++ b/doxygen/examples/H5.format.html
@@ -10780,6 +10780,12 @@ within the embedded dataspace]<br />
               <a name="ClassReference"></a>
                 <p>Class specific information for the Reference class (Class 7):</p>
 
+                <p>Note that for region references, the stored data is
+                   a <a href="#GlobalHeapID">Global Heap ID</a> pointing to information
+                   about the region stored in the global heap.
+                </p>
+
+
                 <div align="center">
                   <table class="desc">
                     <caption>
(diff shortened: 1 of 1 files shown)


derobins (approved):

glennsong09 (approved):

doxygen/examples/H5.format.html
```diff
@@ -11276,7 +11287,7 @@ <h4><a name="DatatypeMessage">IV.A.2.d. The Datatype Message</a></h4>
                         </tr>
 
                         <tr>
-                          <td colspan="4"><br />Base Type<br /><br /></td>
+                          <td colspan="4"><br />Parent Type Message<br /><br /></td>
```
jhendersonHDF, line 11290: To me, "base type" seems more appropriate since a variable-length type is similar to a container around some other type. That is to say, it seems to me to be more appropriate to call the variable-length type the parent type and the more fundamental type should be referred to as the base type.

mattjala: The documentation generally seems to use 'base type' and 'parent type' interchangeably. Enums and arrays are also described as having 'parent types' - would it be best to standardize all of these cases to refer to 'base types'? The field in the shared datatype structure is named `parent`, but that may not matter since it's internal.

jhendersonHDF: I definitely think it would be good to standardize on a term, whether that's 'base type' or 'parent type'. The library definitely uses the two interchangeably fairly often, but I've personally always found 'parent type' to be confusing when used to refer to the type that a container type (vlen, array, compound) contains. The header comment for `H5Tvlen_create` even says "Create a new variable-length datatype based on the specified BASE_TYPE." and the parameter is `base_id`.

@derobins @fortnern Any opinions on this?

PR #3954, opened 2024-01-23
h5 compiler wrappers now pass all arguments passed to it to the compile line

brtnfld:
The issue was that the "allargs" variable was not being used in the final command of the compiler wrapper. Any entries containing an escaped quote (\", \') or other non-matching argument (*) would not be passed to the compile line. I have fixed this problem by ensuring all arguments passed to the compiler wrapper are now included in the compile line.

Addresses: HDFFV-8630, h5pfc doesn't pass quotes in preprocessor macros to preprocessor


derobins:
Can we add a test to make sure this doesn't have problems in the future?

lrknox (approved):

lrknox (approved):

byrnHDF (approved):

PR #3956, opened 2024-01-24
msvc: don't declare `HAVE_TIMEZONE` for older MSVC

mathstuf:
It was introduced in MSVC 15 (Visual Studio 2017).


mathstuf:
Ah…Intel OneAPI definitely needs considered here. Not sure what's up with the other failure off-hand.

jhendersonHDF:
It looks like the vast majority of these failures are probably because of the #ifdef structure in https://github.com/HDFGroup/hdf5/blob/develop/src/H5system.c#L207 and because of `HAVE_TIMEZONE` not getting defined for some reason. Though, according to the CMake docs, `MSVC_VERSION` is preferred over `MSVC14`, so you might try something like:

```Cmake
if (MSVC_VERSION GREATER_EQUAL 1910)
```

based on https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html#variable:MSVC_VERSION

mathstuf:
OneAPI probably needs some way to detect what backing runtime/headers are available too.

byrnHDF:
> It looks like the vast majority of these failures are probably because of the #ifdef structure in https://github.com/HDFGroup/hdf5/blob/develop/src/H5system.c#L207 and because of `HAVE_TIMEZONE` not getting defined for some reason. Though, according to the CMake docs, `MSVC_VERSION` is preferred over `MSVC14`, so you might try something like:
> 
> ```cmake
> if (MSVC_VERSION GREATER_EQUAL 1910)
> ```
> 
> based on https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html#variable:MSVC_VERSION

This, MSVC_VERSION, is the preferred format .

byrnHDF (changes requested):
Use MSVC_VERSION

derobins (approved):

byrnHDF (approved):

Issue #3951, opened 2024-01-22
mpi_f08 error freeing mpi_info type returned by h5pget_fapl_mpi_f


michalakes:
**Describe the bug**
The mpi_info descriptor returned by h5pget_fapl_mpi_f (https://docs.hdfgroup.org/hdf5/develop/group___f_h5_p.html#ga7e27d00061a651c885c6b69a86892303) is an older MPI-1 style integer descriptor that is not being freed correctly by the MPI-2 MPI_INFO_FREE routine (we have `use mpi_f08` in our Fortran code). 

The Fortran 2008 interface to MPI_INFO_FREE specifies type(MPI_INFO)  (https://www.open-mpi.org/doc/v3.0/man3/MPI_Info_free.3.php#sect8) rather than an integer argument so that passing the integer descriptor returned by h5pget_fapl_mpi_f generates a compile-time error:

> `There is no matching specific subroutine for this generic subroutine call.   [MPI_INFO_FREE]`

If we declare a variable mpiinfo of type(MPI_INFO), call MPI_INFO_CREATE(mpiinfo), then assign the value of the descriptor returned by h5pget_fapl_mpi_f to mpiinfo%mpi_val, we can then pass mpiinfo to MPI_INFO_FREE and the code compiles. However, the MPI_INFO_FREE command returns an non-zero error code indicating that MPI_INFO_FREE failed.

We note that the page for H5Pget_fapl_mpio (https://docs.hdfgroup.org/hdf5/develop/group___f_a_p_l.html#ga5d9e4b9f27791ccf19484a898f3fd6fa) states explicitly that it is our responsibility to free the duplicate MPI communicator and mpi_info objects returned by H5Pget_fapl_mpio:

> "Since the MPI communicator and Info object are duplicates of the stored information, future modifications to the access property list will not affect them. It is the responsibility of the application to free these objects."

We have had problems at run-time when the system runs out of handles for unfreed MPI communicators. We wish to avoid any unexpected issues with handles for unfreed MPI_INFO descriptors as well.  For MPI communicators, setting comm%mpi_val with the integer handle returned by H5Pget_fapl_mpio_f and calling MPI_COMM_FREE to free the comm object does work. That is, it returns zero for error code. It just doesn't work for MPI_INFO handles.

**Expected behavior**
MPI_INFO_FREE should return a zero error code when called with an MPI_INFO handle returned by H5Pget_fapl_mpio_f

**Platform (please complete the following information)**
[...]

PR #3959, opened 2024-01-24
Add API support for Fortran MPI_F08 module definitions.

brtnfld:
Adds support for MPI's MPI_F08 module datatypes: type(MPI_COMM) and type(MPI_INFO) for HDF5 APIs:

H5PSET_FAPL_MPIO_F, H5PGET_FAPL_MPIO_F,  H5PSET_MPI_PARAMS_F, H5PGET_MPI_PARAMS_F

Includes tests and updated documentation.

See: #3951


epourmal (approved):
Looks good to me. Question: is there any way to avoid code duplication in mpi_para.F90 using ifdef? The only difference is the type definition for info and comm variables, rigth?

brtnfld:
My assumption that TYPE(MPI*) is C interpretable is wrong, as it is for  MPICH but not OpenMPI; I need to instead go back to using  MPI_Comm_f2c, and I'll rework the test to simplify it.

ref. https://github.com/open-mpi/ompi/issues/12274

epourmal:
> My assumption that TYPE(MPI*) is C interpretable is wrong, as it is for MPICH but not OpenMPI; I need to instead go back to using MPI_Comm_f2c, and I'll rework the test to simplify it.
> 
> ref. [open-mpi/ompi#12274](https://github.com/open-mpi/ompi/issues/12274)

I see. It would be really nice to get rid of the wrappers, but I guess, it will take time...

brtnfld:
It's more straightforward to rewrite the test if I get into generic types (type(*)) or use classes; I'm not sure I want to introduce that complexity/Fortran standards into the tests.

byrnHDF (approved):

PR #3963, opened 2024-01-25
Document that fileno is same for all handles

mattjala:


byrnHDF (approved):

src/H5Fpublic.h
```diff
@@ -741,6 +741,11 @@ H5_DLL herr_t H5Fget_intent(hid_t file_id, unsigned *intent);
  *          file identifier \p file_id and the pointer \p fnumber to the file
  *          number.
  *
+ *          This file number is the same for all open instances of the same
+ *          file. If the active VFD does not enforce this, then it is the
+ *          application's responsibility to avoid opening multiple handles into the
+ *          same file, which results in undefined behavior.
+ *
```
derobins, line 750: Should also mention that, in addition to VFD 'cmp' callback support as you have here, we also need to be able to determine that the file has been opened multiple times, which might not always work depending on the file system.

derobins (approved):

PR #3966, opened 2024-01-26
Added a GitHub Codespaces configuration.

gheber:
This configuration has everything installed to do development, fix an issue, or do debugging.

### DIFF
--- a/.devcontainer/noop.txt
+++ b/.devcontainer/noop.txt
@@ -1,0 +1,3 @@
+This file copied into the container along with environment.yml* from the parent
+folder. This file is included to prevents the Dockerfile COPY instruction from 
+failing if no environment.yml is found.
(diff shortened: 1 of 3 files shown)


.devcontainer/noop.txt
```diff
@@ -0,0 +1,3 @@
+This file copied into the container along with environment.yml* from the parent
+folder. This file is included to prevents the Dockerfile COPY instruction from 
+failing if no environment.yml is found.
```
bmribler: Suggest small change:
This file is copied into the container along with environment.yml* from the parent
folder.  It is included to prevent the Dockerfile COPY instruction from 
failing if no environment.yml is found.

gheber: Done.

bmribler (approved):

derobins (approved):

Issue #2414, opened 2023-01-20
Segfault when copying dataset with attributes


takluyver:
**Describe the bug**

We've come across a case where `H5Ocopy` causes a segfault when copying an object with attributes. We originally came across this via h5py, but I can replicate it with the h5copy command:

```
h5copy -i repacked2.h5 -o test2.h5 -s value -d value
```

Here is the input file to reproduce this (zipped to please Github): [repacked2.zip](https://github.com/HDFGroup/hdf5/files/10467499/repacked2.zip)

I suspect that this is related to dense attribute storage - when deleting attributes, I could reproduce it with 6 attributes, but not with 4 (6 is the default `min_dense`, so below this it will switch back to compound storage). Passing `-f noattr` to h5copy avoids the error.

I haven't been yet been able to reproduce it with a newly created file - the example file above is a heavily cut down version of a file from the European XFEL data acquisition system. I believe it's a valid HDF5 file - I can read all of the attributes with no errors. And ideally even if the file is corrupt, we'd like HDF5 to catch that without segfaulting.

We saw a segfault here on various HDF5 versions (1.10.6, 1.12.1, 1.12.2, 1.14.0), but I've focused on 1.14.0 for investigation.

<details>
<summary>GDB Backtrace from h5copy 1.14.0</summary>

```
#0  0x00002aaaab271ec4 in H5T__conv_vlen (src_id=216172782113784125, dst_id=216172782113784126, cdata=0x699580, 
    nelmts=1, buf_stride=0, bkg_stride=0, buf=0x69ae78, bkg=0x0) at H5Tconv.c:3296
#1  0x00002aaaab256178 in H5T_convert (tpath=0x699510, src_id=216172782113784125, dst_id=216172782113784126, 
    nelmts=1, buf_stride=0, bkg_stride=0, buf=0x69ae78, bkg=0x0) at H5T.c:5449
#2  0x00002aaaaad5cc27 in H5A__attr_copy_file (attr_src=0x697830, file_dst=0x682710, 
    recompute_size=0x7fffffffd2ff, cpy_info=0x7fffffffd5a0) at H5Aint.c:2367
#3  0x00002aaaaad5dbbe in H5A__dense_post_copy_file_cb (attr_src=0x697830, _udata=0x7fffffffd310) at H5Aint.c:2556
#4  0x00002aaaaad4c930 in H5A__dense_iterate_bt2_cb (_record=0x696ac8, _bt2_udata=0x7fffffffd210)
[...]

PR #3967, opened 2024-01-29
Fix for github issue #2414: segfault when copying dataset with attrib…

### LINKED ISSUE #3214: Segfault when copying dataset
**Describe the bug**
I have created an HDF5 file with h5py that contains a compound dataset which cannot be copied (not using `h5py.h5o.copy` and not using the `h5copy` program). I get a Segmentation fault.

The interesting point here is, that I can create a new HDF5 file `input2.h5` with an identical dataset which can be copied. When I look at the two input files `input.h5` and `input2.h5` with HDFView, the only difference I see is the object ref.

Here are the data files (`output.h5` is just an empty file for your convenience).
[h5copy_compound.zip](https://github.com/HDFGroup/hdf5/files/11914414/h5copy_compound.zip)

```bash
# This works:
$ h5copy -i input.h5 -o output.h5 -s "/events/frame" -d "/frame"
# This causes a segmentation fault:
$ h5copy -i input.h5 -o output.h5 -s "/tables/cytoshot_monitor" -d "/monitor"
Segmentation fault
# But this works with a dataset containing the same `cytoshot_monitor` information as input.h5:
$ h5copy -i input2.h5 -o output.h5 -s "/tables/cytoshot_monitor" -d "/monitor"
```

Possibly related issue: https://github.com/HDFGroup/hdf5/issues/2414 (But here it is not related to attributes I think).

**Expected behavior**
Performing h5copy does not segfault.

 - HDF5 version:
```
Summary of the h5py configuration
---------------------------------

h5py    3.8.0
HDF5    1.12.2
Python  3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.23.5
cython (built with) 0.29.33
numpy (built against) 1.21.6
HDF5 (built against) 1.12.2
```
 - OS and version Ubuntu 22.04 LTS

vchoi-hdfgroup:
…utes.

This also fixes github issue #3241: segfault when copying dataset. Need to set the location via H5T_set_loc() of the src datatype when copying dense attributes.
Otherwise the vlen callbacks are not set up therefore causing seg fault when doing H5T_convert() -> H5T__conv_vlen().


mattjala:
I think issue #3214 is the one you meant to link - 3241 seems unrelated.

src/H5Aint.c
```diff
@@ -2456,6 +2456,10 @@ H5A__dense_post_copy_file_cb(const H5A_t *attr_src, void *_udata)
     assert(udata->file);
     assert(udata->cpy_info);
 
+    /* Set the location of the src datatype */
+    if (H5T_set_loc(attr_src->shared->dt, H5F_VOL_OBJ(udata->oloc_src->file), H5T_LOC_DISK) < 0)
+        HGOTO_ERROR(H5E_DATATYPE, H5E_CANTINIT, FAIL, "cannot mark datatype on disk");
```
jhendersonHDF: Even though they're the same value, looks like this should probably be `H5_ITER_ERROR` rather than `FAIL`, along with the other `FAIL`s in this function. Otherwise, looks good to me.

vchoi-hdfgroup:
> I think issue #3214 is the one you meant to link - 3241 seems unrelated.

You are right!

src/H5Aint.c
vchoi-hdfgroup: Fixed

jhendersonHDF (approved):

lrknox (approved):

PR #3968, opened 2024-01-29
Fixed XL and gfortran errors

brtnfld:
Corrected type and intent.


epourmal (approved):
Looks good to me.

brtnfld:
This fixes the compilation errors, but with the XL compilers, the Fortran test returns garbage for the **len** array. This is what the C API is passing back. I've tried passing a pointer to the array to C, which returns the same garbage values. The offset array has correct values. We need a C test that checks the values in len and offset.

derobins (approved):

PR #3976, opened 2024-01-31
append '-WF,' when passing C preprocessor directives to the xlf compiler

brtnfld:
fixes fortran wrapper compiler (h5*fc) test when using the xlf compiler.


epourmal (approved):
Looks good.

lrknox (approved):

PR #3980, opened 2024-02-01
Add callable ABI checker workflow GitHub Action.

hyoklee:
You can generate C/Java ABI reports with this callable workflow.

Sample usage:

```
name: Check C/Java ABI

on:
  push:
    branches:
      - develop

permissions:
  contents: read

jobs:
  call-workflow-abi_c:
    uses: ./.github/workflows/abi_c.yml
    with:
      y0: '14'
      y1: '15'
      z0: '3'
      z1: '4'
```

### DIFF
--- a/.github/workflows/abi_c.yml
+++ b/.github/workflows/abi_c.yml
@@ -1,0 +1,87 @@
+name: Check Application Binary Interface (ABI)
+
+on:
+  workflow_call:
+    inputs:
+      y0:
+        description: 'Release Minor Version'
+        type: string
+        required: true
+        default: '14'
+      y1:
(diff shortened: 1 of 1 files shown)


byrnHDF:
Looks good.
As for inputs, we should think where this will get called; 
1. during daily-build on a schedule, it will use the snapshot binary vs the last release binary
2. during actual release it will be the new release vs the last release.
So maybe the inputs can be simpler, maybe something like the inputs for the release-files workflow?

hyoklee:
This is no longer necessary.

See https://github.com/HDFGroup/hdf5/pull/4006

PR #3988, opened 2024-02-04
Make platform specific test presets for windows and macs

byrnHDF:
Set mac tests to "-j 2" instead of four
Set windows Intel tests to exclude tfloatattrs test.


hyoklee:
I could confirm that this patch is necessary for Intel oneAPI 2024.0.2 on Windows:

https://my.cdash.org/test/120199759

hyoklee (approved):

lrknox (approved):

PR #3994, opened 2024-02-07
Remove oneapi/clang compiler printf() type warning.

hyoklee:
This PR fixes the following warning:
```
/.../hdfsrc/HDF5Examples/C/H5T/h5ex_t_convert.c:123:33: warning: format specifies type 'int' but the argument has type 'hsize_t' (aka 'unsigned long long') [-Wformat]

        printf("sensor[%d]:\n", i);
                       ~~       ^
                       %llu
1 warning generated.
```

### DIFF
--- a/HDF5Examples/C/H5T/h5ex_t_convert.c
+++ b/HDF5Examples/C/H5T/h5ex_t_convert.c
@@ -120,7 +120,7 @@ main(void)
      * Output the data to the screen.
      */
     for (i = 0; i < DIM0; i++) {
-        printf("sensor[" PRIuHSIZE "]:\n", i);
+        printf("sensor[%" PRIuHSIZE "]:\n", i);
         printf("Serial number   : %d\n", sensor[i].serial_no);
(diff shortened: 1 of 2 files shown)


HDF5Examples/C/H5T/h5ex_t_convert.c
```diff
@@ -120,7 +120,7 @@ main(void)
      * Output the data to the screen.
      */
     for (i = 0; i < DIM0; i++) {
-        printf("sensor[%d]:\n", i);
+        printf("sensor[%llu]:\n", i);
```
jhendersonHDF: Should use `%" PRIuHSIZE "...` rather than `%llu`

jhendersonHDF:
Missing a "%", but it looks like the "build only" workflows are still running tests. @byrnHDF I don't think that's intended.

jhendersonHDF (approved):

byrnHDF (approved):

Issue #1824, opened 2022-06-23
Replace all sprintf with safer snprintf


seanm:
Even after https://github.com/HDFGroup/hdf5/pull/1815 there are 16 remaining uses of HDsprintf() left. They were a bit too tricky for me to confidently change as I don't know the codebase well.

seanm:
@jhendersonHDF

seanm:
I took another stab at this, and was disappointed to see there are *more* sprintf than there used to be.

Anyway, I fixed more of them. After [this](https://github.com/HDFGroup/hdf5/pull/4003) there is only 1 left!

jhendersonHDF:
Thanks @seanm, I'd forgotten about this; will review.

derobins:
All done. Yay!

PR #4001, opened 2024-02-09
Replace sprintf() with snprintf() to remove Xcode/clang warning.

hyoklee:
Here's a typical warning message:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
188 | __deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
```

PR #4003, opened 2024-02-11
Issue #1824: Replaced most remaining sprintf with safer snprint

seanm:


byrnHDF (approved):

seanm:
@byrnHDF thanks for your speedy review!

Be sure the check the non-trivial ones carefully, they can be tricky...

test/filter_plugin.c
jhendersonHDF, line 1: For the changes here, see #4001. Seems the two calls can simply be combined into a single `snprintf` call

seanm: Oh funny timing that @hyoklee was also addressing some of these.  I'll integrate that here.

test/filter_plugin.c
```diff
@@ -847,10 +847,7 @@ test_creating_groups_using_plugins(hid_t fid)
 
     /* Create multiple groups under the top-level group */
     for (i = 0; i < N_SUBGROUPS; i++) {
-        char *sp = subgroup_name;
```
jhendersonHDF, line 850: Will need to remove the references to `sp` as well since that var. just goes away

seanm: oh whoops I had that staged but for got to commit it...

seanm:
For the last remaining one, I have a starting point: https://github.com/HDFGroup/hdf5/pull/4007

src/H5FDsplitter.c
```diff
@@ -532,27 +532,28 @@ H5FD__splitter_get_default_wo_path(char *new_path, size_t new_path_len, const ch
...
     /* Determine if filename contains a ".h5" extension. */
-    if ((file_extension = strstr(base_filename, ".h5"))) {
+    file_extension = strstr(base_filename, ".h5");
+    if (file_extension) {
         /* Insert the suffix between the filename and ".h5" extension. */
-        strcpy(new_path, base_filename);
-        file_extension = strstr(new_path, ".h5");
-        sprintf(file_extension, "%s%s", suffix, ".h5");
+        intptr_t beginningLength = file_extension - base_filename;
+        snprintf(new_path, new_path_len, "%.*s%s%s", (int)beginningLength, base_filename, suffix, ".h5");
```
jhendersonHDF, line 539: Ah clever and always nicer to do things in a single `snprintf` call!

jhendersonHDF (approved):

PR #4008, opened 2024-02-12
Fix some output issues with ph5diff

jhendersonHDF:
`printf` calls in parallel `h5diff` should be using the `parallel_print` wrapper function, since it does some buffering into a temporary file and bare `printf` calls can cause some interactions where output gets split between standard out and the buffered file.

### DIFF
--- a/tools/src/h5diff/ph5diff_main.c
+++ b/tools/src/h5diff/ph5diff_main.c
@@ -85,8 +85,9 @@ main(int argc, char *argv[])
 
             MPI_Barrier(MPI_COMM_WORLD);
 
-            print_info(&opts);
             print_manager_output();
+
+            print_info(&opts);
         }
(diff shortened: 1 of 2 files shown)


tools/src/h5diff/ph5diff_main.c
```diff
@@ -85,8 +85,9 @@ main(int argc, char *argv[])
 
             MPI_Barrier(MPI_COMM_WORLD);
 
-            print_info(&opts);
             print_manager_output();
+
+            print_info(&opts);
```
jhendersonHDF, line 90: Make the lines printed by `print_info` come at the bottom of output, similar to serial `h5diff`

byrnHDF (approved):

lrknox (approved):

Issue #3084, opened 2023-06-08
Address performance when flushing a large number of datasets individually


bmribler:
Flushing datasets individually via H5Dflush() scales quadratically with the number of datasets. Flushing the entire file scales  linearly, as expected.

It would be useful to investigate this behavior to see if there's a better way to track dirty metadata that scales linearly when a large number of datasets are flushed individually.

derobins:
Closed with #4017

PR #4017, opened 2024-02-14
Improve performance of flushing single objects

fortnern:
Previously the algorithm for flushing a single object was:

1. Use a hash table to quickly look up cache entries with the matching tag, and set the "flush_marker" field in the entry
2. Iterate over ALL cache entries, adding ALL dirty entries to the (initially empty) main skip list
3. Iterate over the skip list, flushing entries that have the flush_marker field set
4. Destroy the skip list

This causes obvious performance problems when there are many cache entries, with multiple iterations over all entries (instead of only the tagged entries), and an O(N log N) build of the skip list, where N is the total number of entries. This PR changes the algorithm to:

1. Use a hash table to quickly look up cache entries with the matching tag, adding each to the main skip list
2. Iterate over the skip list, flushing all entries
3. Destroy the skip list

Since this algorithm quickly finds only the tagged cache entries and does not operate on other entries it is much faster (verified in benchmarks). With this change, flush markers are no longer used, so they have been removed, simplifying the code in some places (especially in the cache test).

This should have a minor change in behavior in that, if a new cache entry is created during the flush, it will be flushed. Previously this would not happen because the flush marker was not set. I believe this is closer to the intent of single object flushes.

It should be possible to further improve performance by using a data structure other than a skip list here, since it is only used to sort entries so they are flushed in increasing address order, and there are more efficient ways to sort an array, but I believe this should improve performance enough for now, though replacing the skip list would also improve the performance of full file flushes (as on file close).


qkoziol:
Sounds very cool!  I'll take a look at the code tomorrow (Thurs) morning.

fortnern:
Addresses #3084

qkoziol (approved):
Nice optimization & code cleanup!

src/H5Centry.c
```diff
@@ -503,36 +503,24 @@ H5C__flush_single_entry(H5F_t *f, H5C_cache_entry_t *entry_ptr, unsigned flags)
...
         if (entry_ptr->in_slist) {
             assert(entry_ptr->is_dirty);
-            if (entry_ptr->flush_marker && !entry_ptr->is_dirty)
+            if (!entry_ptr->is_dirty)
                 HGOTO_ERROR(H5E_CACHE, H5E_SYSTEM, FAIL, "entry in slist failed sanity checks");
         } /* end if */
-        else {
-            assert(!entry_ptr->is_dirty);
-            assert(!entry_ptr->flush_marker);
-            if (entry_ptr->is_dirty || entry_ptr->flush_marker)
```
jhendersonHDF, line 512: Should part of this sanity check have been kept to ensure that there aren't entries which are NOT in the slist but ARE dirty, or is that no longer relevant?

fortnern: With this change it is possible for entries to be dirty but not in the skip list. This happens when flushing a single object (tag) while there are dirty pieces of metadata not associated with that tag.

jhendersonHDF (approved):

fortnern:
I also removed the "clear_slist" parameter from H5C_set_slist_enabled(). This used to be needed for the old way of flushing tagged entries, and was also used for H5C_evict(), but it turns out it isn't necessary in the latter case since, while pinned entries aren't evicted, they're still flushed and therefore removed from the skiplist. It's also apparently not necessary in the place in the test where it was used.

PR #4021, opened 2024-02-15
Fix memory leak in H5LTopen_file_image when H5LT_FILE_IMAGE_DONT_COPY flag is used

### LINKED ISSUE #827: Memory leak in H5LTopen_file_image with H5LT_FILE_IMAGE_DONT_COPY
I observe the following relevant steps:

In `H5LTopen_file_image`:
1. `udata` is allocated
2. `ref_count` and `vfd_ref_count` are set to 1 and 0, respectively.
3. `udata_copy` is called, incrementing `ref_count` to 2.
4. `image_malloc` is called with `H5FD_FILE_IMAGE_OP_FILE_OPEN`, incrementing `vfd_ref_count` to 1.
5. `udata_free` is called, decrementing `ref_count` to 1, not freeing `udatata` because `ref_count` and `vfd_ref_count` are not 0.

In `H5Fclose` on the returned file id:
1. `image_free` is called with `H5FD_FILE_IMAGE_OP_FILE_CLOSE`, decrementing `vfd_ref_count` to 0.

At this point, another call to `udata_free` would finally set `ref_count` to 0 and deallocate `udata`, but I couldn't find a way to make this happen. Should `udata_free` have been called somewhere in `H5Fclose`? Is there a way to call it manually after `image_free` or `H5Fclose`?

I see the same behavior on HDF5 versions 1.10.0-patch1 and 1.12.0 (the latter from h5py).

jhendersonHDF:
When the H5LT_FILE_IMAGE_DONT_COPY flag is passed to H5LTopen_file_image, the internally-allocated udata structure gets leaked as the core file driver doesn't have a way to determine when or if it needs to call the 'udata_free' callback. This has been fixed by freeing the udata structure when the 'image_free' callback gets made during file close, where the file is holding the last reference to the udata structure.

Fixes #827

### DIFF
--- a/hl/src/H5LT.c
+++ b/hl/src/H5LT.c
@@ -286,14 +286,33 @@ image_realloc(void *ptr, size_t size, H5FD_file_image_op_t file_image_op, void *
         goto out;
 
     if (file_image_op == H5FD_FILE_IMAGE_OP_FILE_RESIZE) {
+        void *tmp_realloc;
+
         if (udata->vfd_image_ptr != ptr)
             goto out;
 
         if (udata->vfd_ref_count != 1)
             goto out;
 
-        if (NULL == (udata->vfd_image_ptr = realloc(ptr, size)))
+        /* Make sure all the udata structure image pointers
+         * match each other before we update them
+         */
+        assert(udata->vfd_image_ptr == udata->app_image_ptr);
+        assert(udata->vfd_image_ptr == udata->fapl_image_ptr);
+
+        tmp_realloc = realloc(ptr, size);
+        if (tmp_realloc) {
+            udata->vfd_image_ptr  = tmp_realloc;
+            udata->app_image_ptr  = udata->vfd_image_ptr;
+            udata->fapl_image_ptr = udata->vfd_image_ptr;
+        }
+        else {
+            free(ptr);
+            udata->vfd_image_ptr  = NULL;
+            udata->app_image_ptr  = NULL;
+            udata->fapl_image_ptr = NULL;
             goto out;
+        }
 
         udata->vfd_image_size = size;
         return_value          = udata->vfd_image_ptr;
@@ -359,11 +378,20 @@ image_free(void *ptr, H5FD_file_image_op_t file_image_op, void *_udata)
              * references */
             if (udata->fapl_ref_count == 0 && udata->vfd_ref_count == 0 &&
                 !(udata->flags & H5LT_FILE_IMAGE_DONT_RELEASE)) {
+                /* Make sure we aren't going to leak memory elsewhere */
+                assert(udata->app_image_ptr == udata->vfd_image_ptr || udata->app_image_ptr == NULL);
+                assert(udata->fapl_image_ptr == udata->vfd_image_ptr || udata->fapl_image_ptr == NULL);
+
                 free(udata->vfd_image_ptr);
                 udata->app_image_ptr  = NULL;
                 udata->fapl_image_ptr = NULL;
                 udata->vfd_image_ptr  = NULL;
-            } /* end if */
+            }
+
+            /* release reference to udata structure */
+            if (udata_free(udata) < 0)
+                goto out;
+
             break;
 
         /* added unused labels to keep the compiler quite */
@@ -437,9 +465,15 @@ udata_free(void *_udata)
 
     udata->ref_count--;
 
-    /* checks that there are no references outstanding before deallocating udata */
-    if (udata->ref_count == 0 && udata->fapl_ref_count == 0 && udata->vfd_ref_count == 0)
+    if (udata->ref_count == 0) {
+        /* There should not be any outstanding references
+         * to the udata structure at this point.
+         */
+        assert(udata->fapl_ref_count == 0);
+        assert(udata->vfd_ref_count == 0);
+
         free(udata);
+    }
 
     return (SUCCEED);
 
@@ -728,13 +762,13 @@ out:
 hid_t
 H5LTopen_file_image(void *buf_ptr, size_t buf_size, unsigned flags)
 {
-    hid_t    fapl = -1, file_id = -1; /* HDF5 identifiers */
-    unsigned file_open_flags;         /* Flags for image open */
-    char     file_name[64];           /* Filename buffer */
-    size_t   alloc_incr;              /* Buffer allocation increment */
-    size_t   min_incr  = 65536;       /* Minimum buffer increment */
-    double   buf_prcnt = 0.1;         /* Percentage of buffer size to set
-                                          as increment */
+    H5LT_file_image_ud_t       *udata = NULL;            /* Pointer to udata structure */
+    hid_t                       fapl = -1, file_id = -1; /* HDF5 identifiers */
+    unsigned                    file_open_flags;         /* Flags for image open */
+    char                        file_name[64];           /* Filename buffer */
+    size_t                      alloc_incr;              /* Buffer allocation increment */
+    size_t                      min_incr  = 65536;       /* Minimum buffer increment */
+    double                      buf_prcnt = 0.1;         /* Percentage of buffer size to set as increment */
     static long                 file_name_counter;
     H5FD_file_image_callbacks_t callbacks = {&image_malloc, &image_memcpy, &image_realloc, &image_free,
                                              &udata_copy,   &udata_free,   (void *)NULL};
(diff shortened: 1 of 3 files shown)


fortnern:
Why isn't udata_free called when the fapl is shut down via H5P__facc_file_image_info_close()->H5P__file_image_info_free()?

jhendersonHDF:
> Why isn't udata_free called when the fapl is shut down via H5P__facc_file_image_info_close()->H5P__file_image_info_free()?

It's called once there and decrements `ref_count` to 1 since `udata_copy` got called during `H5Pset_file_image_callbacks` in `H5LTopen_file_image` when the `H5LT_FILE_IMAGE_DONT_COPY` flag is given. At that point when the FAPL is closed, the udata structure is essentially owned by the file driver rather than the FAPL. Since there seems to be no good way for a file driver to know when or if it needs to call the `udata_free` callback (it can't introspect the udata structure) and since the allocation of the udata structure is unique to the callbacks set by `H5LTopen_file_image`, those callbacks are responsible for freeing it at file close time when the file has the last reference.

fortnern:
Is there a need to use the udata after the FAPL has been shut down?

jhendersonHDF:
> Is there a need to use the udata after the FAPL has been shut down?

Based on https://docs.hdfgroup.org/hdf5/develop/group___h5_l_t.html#ga07fcf2af194e87a3e45252544ebe4aa4, it looks like the `H5LT_FILE_IMAGE_DONT_COPY` flag is an optimization and the udata structure just manages the reference count so the file image can be opened multiple items without being copied and associated with a FAPL.

hl/src/H5LT.c
```diff
@@ -787,11 +798,9 @@ H5LTopen_file_image(void *buf_ptr, size_t buf_size, unsigned flags)
         callbacks.udata = (void *)udata;
```
fortnern, line 834: I think the comment above "corresponding to the first FAPL" should be changed to something like "the file driver holds a reference to udata, which is released when image_free is called"

hl/src/H5LT.c
```diff
@@ -355,15 +355,28 @@ image_free(void *ptr, H5FD_file_image_op_t file_image_op, void *_udata)
...
-            /* release the shared buffer only if indicated by the respective flag and there are no outstanding
-             * references */
-            if (udata->fapl_ref_count == 0 && udata->vfd_ref_count == 0 &&
-                !(udata->flags & H5LT_FILE_IMAGE_DONT_RELEASE)) {
-                free(udata->vfd_image_ptr);
-                udata->app_image_ptr  = NULL;
-                udata->fapl_image_ptr = NULL;
-                udata->vfd_image_ptr  = NULL;
-            } /* end if */
+            if (udata->fapl_ref_count == 0 && udata->vfd_ref_count == 0) {
```
fortnern: I think this function should always decrement udata->ref_count and free udata if it drops to zero (i.e. the standard way of releasing a reference to  a reference counted object). Is there a possibility that fapl_ref_count or vfd_ref_count could be nonzero if udata->ref_count is zero? If not, we should assert as much. It it is possible, we need to investigate more.

hl/src/H5LT.c
```diff
@@ -355,15 +355,28 @@ image_free(void *ptr, H5FD_file_image_op_t file_image_op, void *_udata)
...
+                /* release the shared buffer only if indicated by the respective
+                 * flag and there are no outstanding references
+                 */
+                if (!(udata->flags & H5LT_FILE_IMAGE_DONT_RELEASE)) {
+                    free(udata->vfd_image_ptr);
+                    udata->app_image_ptr  = NULL;
+                    udata->fapl_image_ptr = NULL;
+                    udata->vfd_image_ptr  = NULL;
```
fortnern: Should we assert here that these pointers point to the same location?

fortnern:
I think the way the reference count is handled here in general is problematic but I don't think we can fix it properly without changing the API. This should be good enough for now, once the comments are addressed.

hl/src/H5LT.c
jhendersonHDF: I can add that assert if you think it's useful. I believe that depending on whether one actually opens the file image or just calls `H5Pset_file_image` to set a file image on a FAPL without ever opening a file (odd but could happen), then you could either have `udata->app_image_ptr == udata->fapl_image_ptr` and `udata->vfd_image_ptr == NULL` OR `udata->app_image_ptr == udata->fapl_image_ptr == udata->vfd_image_ptr`. In this particular block, I believe the latter should always be true, but I haven't verified that.

hl/src/H5LT.c
jhendersonHDF: The default `udata_free` callback supplied by H5LT.c wraps around this and always decrements `udata->ref_count` and then only frees the structure when it drops to 0. Changing this code to always call `udata_free` here appears to pass the file image tests, though those aren't too extensive; I'm not sure yet if there are cases where that would be problematic. It seems one could unknowingly create a different memory leak with `H5Pget_file_image_callbacks` since it calls `udata_copy` without a chance for an `image_free` call somewhere, but this shouldn't happen with the default callbacks at least.

In my current understanding of the code, I don't believe there should be a case where `udata->ref_count == 0` and `udata->fapl_ref_count != 0 || udata->vfd_ref_count != 0`. The udata structure should always be owned by either a FAPL or the VFD, so we should always at least have `udata->ref_count == 1 && udata->fapl_ref_count == 1` OR `udata->ref_count == 1 && udata->vfd_ref_count == 1` here before everything gets closed and cleaned up.

jhendersonHDF:
> I think the way the reference count is handled here in general is problematic but I don't think we can fix it properly without changing the API. This should be good enough for now, once the comments are addressed.

Agreed, the reference counting is problematic here. This memory leak is inherent to the library's implementations of the file image callbacks, but it also seems to point to the interface needing fixing so users could have similar "nocopy" implementations without creating memory leaks.

hl/src/H5LT.c
fortnern: I guess we should assert that everything is either the same or NULL, we don't want to leak more memory

hl/src/H5LT.c
fortnern: Ok I guess my suggestion would be to always call udata_free, after the `if (udata->fapl_ref_count == 0 && udata->vfd_ref_count == 0) {` block, and assert in udata_free that all the ref counts are 0 if udata->ref_count is 0

hl/src/H5LT.c
fortnern:
```
assert(udata->app_image_ptr == udata->vfd_image_ptr || udata->app_image_ptr == NULL);
assert(udata->fapl_image_ptr == udata->vfd_image_ptr || udata->fapl_image_ptr == NULL);
```

hl/src/H5LT.c
```diff
@@ -286,14 +286,33 @@ image_realloc(void *ptr, size_t size, H5FD_file_image_op_t file_image_op, void *
...
+         * match each other before we update them
+         */
+        assert(udata->vfd_image_ptr == udata->app_image_ptr);
+        assert(udata->vfd_image_ptr == udata->fapl_image_ptr);
+
+        tmp_realloc = realloc(ptr, size);
+        if (tmp_realloc) {
+            udata->vfd_image_ptr  = tmp_realloc;
+            udata->app_image_ptr  = udata->vfd_image_ptr;
+            udata->fapl_image_ptr = udata->vfd_image_ptr;
```
jhendersonHDF, line 307: Since the `H5FD_FILE_IMAGE_OP_FILE_RESIZE` operation will realloc the file image buffer, I had to make sure this code updates all the pointers before we get to the newly-added asserts checking for pointers being equivalent before freeing `udata->vfd_image_ptr`.

src/H5Pfapl.c
```diff
@@ -3209,11 +3210,18 @@ H5Pset_file_image_callbacks(hid_t fapl_id, H5FD_file_image_callbacks_t *callback
...
 
+    copied_udata = true;
+
     /* Set values */
     if (H5P_poke(fapl, H5F_ACS_FILE_IMAGE_INFO_NAME, &info) < 0)
         HGOTO_ERROR(H5E_PLIST, H5E_CANTSET, FAIL, "can't set file image info");
 
 done:
+    if (ret_value < 0) {
+        if (copied_udata && (callbacks_ptr->udata_free(info.callbacks.udata) < 0))
```
jhendersonHDF, line 3221: Since this function calls `udata_copy` and will increment the ref. count on the udata structure, make sure we call `udata_free` if this fails.

hl/src/H5LT.c
```diff
@@ -781,17 +813,30 @@ H5LTopen_file_image(void *buf_ptr, size_t buf_size, unsigned flags)
...
+         * Initially, the FAPL created in this function holds a reference
+         * to the udata structure allocated above. The call to
+         * H5Pset_file_image_callbacks below will associate the udata
+         * structure with the FAPL and cause it to hold another reference
+         * in preparation for transfer of ownership to the file driver.
+         * Once the file has been opened with this FAPL and the FAPL is
+         * closed, the reference held by the FAPL is released and ownership
+         * is transferred to the file driver. The udata structure will then
+         * be freed when the image_free callback is made and the file driver
+         * releases its reference to the structure.
```
fortnern: So this reference added here (the "1") is really always the file driver's reference right? After the call to H5Pset_file_image_callbacks the rc is 2, one for the FAPL one for the FD right? Sorry if this seems pedantic but I think it's important to always explicitly who owns a reference.

fortnern (approved):
Looks good aside from the minor nitpick about the comment

hl/src/H5LT.c
jhendersonHDF: Yes, wording this is a bit tricky because in actuality no one owns that reference to the udata structure initially. The call to `H5Pset_file_image_callbacks` comes first and associates the udata structure with the FAPL, at which point the RC is 2 and yet still no one owns that initial reference. It isn't until the H5Fopen call that the FD actually owns the reference, which isn't a particularly great design since it makes cleaning things up tricky if `H5LTopen_file_image` fails. I'll reword a bit to make this explicit.

derobins (approved):

PR #4023, opened 2024-02-16
Relax CMake version compatibility

ZedThree:
Consider the following minimal `CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.14)
project(test)

find_package(HDF5 1.8.10 REQUIRED CONFIG)
```

and building like `cmake . -B build -DHDF5_ROOT=/path/to/installed/hdf5`.

The current CMake version compatibility is `SameMinorVersion`, which causes `find_package` to fail if the installed HDF5 version doesn't match exactly (ignoring the patch number):

```
CMake Error at CMakeLists.txt:4 (find_package):
  Could not find a configuration file for package "HDF5" that is compatible
  with requested version "1.8.10".

  The following configuration files were considered but not accepted:

    /path/to/installed/hdf5/cmake/hdf5-config.cmake, version: 1.15.0
```

This PR changes the version compatibility to `SameMajorVersion`, which allows the above CMake test to configure correctly.

This also brings the CMake config file version compatibility behaviour in line with the CMake-bundled `FindHDF5` -- notice that removing `CONFIG` from the `find_package` call works fine without this fix, as `FindHDF5` is used instead.

Semi-related, I've also deleted some commented out code and the unused version config file that was replaced by the call to `write_basic_package_version_file`


derobins:
Historically, we have used the semantic versioning minor version as our "major version" so this change would break binary compatibility. This will be changed in a future major release of the library (probably the next time we have to change the file format), when we release HDF5 2.0.0.

ZedThree:
> this change would break binary compatibility

I'm not sure I follow, sorry. The intention here is for API compatibility, and should only affect calls to `find_package`. Currently, it's not possible to request "at least 1.10", say, and use the CMake config file for a later version. This is in contrast to using the `FindHDF5` module with an autotools build of HDF5, for example.

PR #4029, opened 2024-02-19
Check argument for CMake REGEX FCMangle.h.

hyoklee:
See @imikejackson [forum](https://forum.hdfgroup.org/t/configuring-1-8-18-on-macos-sierra-with-gfortran-6-3-0/3999) message.
 
 Also, prevent using incompatible c/Fortran compiler pairs in advance before it reaches REGEX checking.

### DIFF
--- a/config/cmake/HDFUseFortran.cmake
+++ b/config/cmake/HDFUseFortran.cmake
@@ -32,6 +32,9 @@ endif ()
 #-----------------------------------------------------------------------------
 include (FortranCInterface)
 
+#-----------------------------------------------------------------------------
+# Verify that the Fortran and C/C++ compilers work together
+#-----------------------------------------------------------------------------
 FortranCInterface_VERIFY()
 
 FortranCInterface_HEADER (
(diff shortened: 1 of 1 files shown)


config/cmake/HDFUseFortran.cmake
```diff
@@ -31,19 +31,27 @@ endif ()
...
 )
 
 file (STRINGS ${CMAKE_BINARY_DIR}/FCMangle.h CONTENTS REGEX "H5_FC_GLOBAL\\(.*,.*\\) +(.*)")
-string (REGEX MATCH "H5_FC_GLOBAL\\(.*,.*\\) +(.*)" RESULT ${CONTENTS})
-set (H5_FC_FUNC "H5_FC_FUNC(name,NAME) ${CMAKE_MATCH_1}")
+if (CONTENTS)
+  message(${CONTENTS})
+  string (REGEX MATCH "H5_FC_GLOBAL\\(.*,.*\\) +(.*)" RESULT  ${CONTENTS})
+  set (H5_FC_FUNC "H5_FC_FUNC(name,NAME) ${CMAKE_MATCH_1}")
+endif()
```
byrnHDF: should there be an else block? (and below, too)  If neither REGEX is found won't H5_FC_FUNC be undefined?

hyoklee: FortranCInterface_VERIFY() should fail for else case.

byrnHDF: Then why is the if needed?

hyoklee: Because I wasn't sure what's the right approach:

1. Use F..._VERIFY()
2. Safeguard REGEX with if ().

E.g., approach 2. was used in many places in HDF5 CMake codes.

byrnHDF: Add a comment to the FortranCInterface_VERIFY() call that it will fail if the compilers are incompatible the remove the if/endifs as they are just confusing.

byrnHDF (approved):

bmribler (approved):

PR #4030, opened 2024-02-20
Fix H5F_get_access_plist to copy file locking settings

jhendersonHDF:
H5F_get_access_plist previously did not copy over the file locking settings from a file into the new File Access Property List that it creates. This would make it difficult to match the file locking settings between an external file and its parent file.

### DIFF
--- a/src/H5Fint.c
+++ b/src/H5Fint.c
@@ -79,7 +79,8 @@ static int    H5F__get_objects_cb(void *obj_ptr, hid_t obj_id, void *key);
 static herr_t H5F__build_name(const char *prefix, const char *file_name, char **full_name /*out*/);
 static char  *H5F__getenv_prefix_name(char **env_prefix /*in,out*/);
 static H5F_t *H5F__new(H5F_shared_t *shared, unsigned flags, hid_t fcpl_id, hid_t fapl_id, H5FD_t *lf);
-static herr_t H5F__check_if_using_file_locks(H5P_genplist_t *fapl, bool *use_file_locking);
+static herr_t H5F__check_if_using_file_locks(H5P_genplist_t *fapl, bool *use_file_locking,
+                                             bool *ignore_disabled_locks);
 static herr_t H5F__dest(H5F_t *f, bool flush, bool free_on_failure);
 static herr_t H5F__build_actual_name(const H5F_t *f, const H5P_genplist_t *fapl, const char *name,
                                      char ** /*out*/ actual_name);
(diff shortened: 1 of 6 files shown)


src/H5Fint.c
```diff
@@ -1951,14 +1982,19 @@ H5F_open(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id)
...
-        file->shared->use_file_locking = use_file_locking;
-    else if (shared->nrefs > 1)
+    if (shared->nrefs == 1) {
+        file->shared->use_file_locking      = use_file_locking;
+        file->shared->ignore_disabled_locks = ignore_disabled_locks;
+    }
+    else if (shared->nrefs > 1) {
         if (file->shared->use_file_locking != use_file_locking)
             HGOTO_ERROR(H5E_FILE, H5E_CANTINIT, NULL, "file locking flag values don't match");
+        if (file->shared->use_file_locking && (file->shared->ignore_disabled_locks != ignore_disabled_locks))
```
jhendersonHDF, line 1997: I wasn't quite sure on this addition, but it seems reasonable that, similar to how we check the file locking setting, we would also want to check that a file which is opened multiple times has the same "ignore disabled file locks" setting each time.

src/H5Fint.c
```diff
@@ -79,7 +79,8 @@ static int    H5F__get_objects_cb(void *obj_ptr, hid_t obj_id, void *key);
 static herr_t H5F__build_name(const char *prefix, const char *file_name, char **full_name /*out*/);
 static char  *H5F__getenv_prefix_name(char **env_prefix /*in,out*/);
 static H5F_t *H5F__new(H5F_shared_t *shared, unsigned flags, hid_t fcpl_id, hid_t fapl_id, H5FD_t *lf);
-static herr_t H5F__check_if_using_file_locks(H5P_genplist_t *fapl, bool *use_file_locking);
+static herr_t H5F__check_if_using_file_locks(H5P_genplist_t *fapl, bool *use_file_locking,
+                                             bool *ignore_disabled_locks);
```
lrknox, line 83: Will adding the extra parameter break binary compatibility if merged to hdf5_1_14?

jhendersonHDF: This is just a static function, so changing the parameters shouldn't be a problem. Nothing directly in the public API is touched here.

lrknox: I see, good

lrknox (approved):

byrnHDF (approved):

PR #4031, opened 2024-02-20
Replace deprecated Fortran 'include mpif.h' with 'USE mpi'

cniethammer:
With MPI 4.1 the use of the mpif.h include file has been deprecated. Codes should transition to USE mpi or USE mpi_f08.


brtnfld (approved):
Looks good, thanks.

lrknox (approved):

PR #4038, opened 2024-02-22
Handle certain empty subfiling environment variables 

### LINKED ISSUE #3978: subfiling should gracefully handle enviroment varibles that are not set to anything
subfiling environment variables   that are defined but not set to any values could cause issues (e.g., 

export H5FD_SUBFILING_IOC_SELECTION_CRITERIA=
export H5FD_SUBFILING_SUBFILE_PREFIX=, etc...) 

with the HDF5 library, it should either fail gracefully or assume the default values for those environment variables.

glennsong09:
Handles the case of subfiling environment variables such as H5FD_SUBFILING_IOC_SELECTION_CRITERIA and H5FD_SUBFILING_SUBFILE_PREFIX being set to the empty string, which caused MPI_TEST_t_subfiling_vfd and MPI_TEST_t_vfd to fail.

Fixes #3978.

### DIFF
--- a/src/H5FDsubfiling/H5FDioc.c
+++ b/src/H5FDsubfiling/H5FDioc.c
@@ -1498,7 +1498,7 @@ H5FD__ioc_del(const char *name, hid_t fapl)
         /* TODO: No support for subfile directory prefix currently */
         /* TODO: Possibly try loading config file prefix from file before deleting */
         snprintf(tmp_filename, PATH_MAX, "%s/" H5FD_SUBFILING_CONFIG_FILENAME_TEMPLATE,
-                 prefix_env ? prefix_env : file_dirname, base_filename, (uint64_t)st.st_ino);
+                 prefix_env && (strlen(prefix_env) > 0) ? prefix_env : file_dirname, base_filename, (uint64_t)st.st_ino);
 
         if (NULL == (config_file = fopen(tmp_filename, "r"))) {
             if (ENOENT == errno) {
--- a/src/H5FDsubfiling/H5FDioc.c
+++ b/src/H5FDsubfiling/H5FDioc.c
@@ -224,7 +224,7 @@ H5FD_ioc_init(void)
 
         /* Check if IOC VFD has been loaded dynamically */
         env_var = getenv(HDF5_DRIVER);
-        if (env_var && !strcmp(env_var, H5FD_IOC_NAME)) {
+        if ((env_var && !strcmp(env_var, H5FD_IOC_NAME) && (strlen(env_value) > 0))) {
             int mpi_initialized = 0;
             int provided        = 0;
 
@@ -1497,8 +1497,10 @@ H5FD__ioc_del(const char *name, hid_t fapl)
 
         /* TODO: No support for subfile directory prefix currently */
         /* TODO: Possibly try loading config file prefix from file before deleting */
-        snprintf(tmp_filename, PATH_MAX, "%s/" H5FD_SUBFILING_CONFIG_FILENAME_TEMPLATE,
+        if (prefix_env && (strlen(prefix_env) > 0)) {
+            snprintf(tmp_filename, PATH_MAX, "%s/" H5FD_SUBFILING_CONFIG_FILENAME_TEMPLATE,
                  prefix_env ? prefix_env : file_dirname, base_filename, (uint64_t)st.st_ino);
+        }
 
         if (NULL == (config_file = fopen(tmp_filename, "r"))) {
             if (ENOENT == errno) {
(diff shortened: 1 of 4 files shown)


brtnfld (commented):
I would add a test where all the env. var. are empty to ensure it still regresses to the defaults.

src/H5FDsubfiling/H5FDioc.c
```diff
@@ -1497,8 +1497,10 @@ H5FD__ioc_del(const char *name, hid_t fapl)
 
         /* TODO: No support for subfile directory prefix currently */
         /* TODO: Possibly try loading config file prefix from file before deleting */
-        snprintf(tmp_filename, PATH_MAX, "%s/" H5FD_SUBFILING_CONFIG_FILENAME_TEMPLATE,
-                 prefix_env ? prefix_env : file_dirname, base_filename, (uint64_t)st.st_ino);
+        if (prefix_env && (strlen(prefix_env) > 0)) {
+            snprintf(tmp_filename, PATH_MAX, "%s/" H5FD_SUBFILING_CONFIG_FILENAME_TEMPLATE,
+                     prefix_env ? prefix_env : file_dirname, base_filename, (uint64_t)st.st_ino);
```
jhendersonHDF: This `snprintf` call should be ok as it was. This line checks `prefix_env` and adds that in for the first `%s` specifier if the environment variable was set. Otherwise, `file_dirname` is added in. With these changes, not having the environment variable set means nothing will get set in `tmp_filename`, so the open below should always end up failing.

src/H5FDsubfiling/H5FDioc.c
```diff
@@ -224,7 +224,7 @@ H5FD_ioc_init(void)
 
         /* Check if IOC VFD has been loaded dynamically */
         env_var = getenv(HDF5_DRIVER);
-        if (env_var && !strcmp(env_var, H5FD_IOC_NAME)) {
+        if ((env_var && !strcmp(env_var, H5FD_IOC_NAME) && (strlen(env_value) > 0))) {
```
jhendersonHDF: Not strictly necessary, but it may be reasonable to have the `strlen` check in between the `env_var` and `!strcmp(env_var, H5FD_IOC_NAME)` checks. If `!strcmp(env_var, H5FD_IOC_NAME)` is true, there's not really a need to check the length of `env_var` at that point. Checking the string length before the `!strcmp(env_var, H5FD_IOC_NAME)` would also save you a string comparison when the environment variable is empty, though that's not really a big deal.

src/H5FDsubfiling/H5FDioc.c
jhendersonHDF: My mistake, I was only partially correct here. The `prefix_env ? prefix_env : file_dirname` part inside the snprintf call actually needs to be changed to `prefix_env && (strlen(prefix_env) > 0) ? prefix_env : file_dirname` like you've been doing elsewhere. That way an empty prefix doesn't cause the pathname to be modified to "/original_path" and likely cause failures.

jhendersonHDF (approved):

jhendersonHDF:
@glennsong09 and I discovered some testing failures with MPICH that are not related to this PR. Will fix and setup a daily (for now) Subfiling testing action with MPICH and OpenMPI and will eventually merge with the coming VFD CI action.

testpar/t_vfd.c
```diff
@@ -6250,6 +6255,68 @@ main(int argc, char **argv)
 
     test_vector_io(mpi_rank, mpi_size);
 
+#ifdef H5_HAVE_SUBFILING_VFD
+
+    if (mpi_rank == 0)
+        printf("\n --- TESTING SUBFILING VFD: environment variables set to empty --- \n");
+
+    subfiling_subfile_prefix_saved         = getenv("H5FD_SUBFILING_SUBFILE_PREFIX");
```
qkoziol: I believe that you need to strdup these

jhendersonHDF: True, although in this case we probably don't even need to try to save the values. @glennsong09 I think we can update this to just have the setenv/unsetenv pairs.

glennsong09: Okay, I removed the value saving code.

testpar/t_subfiling_vfd.c
```diff
@@ -2892,6 +2892,11 @@ main(int argc, char **argv)
     bool     must_unset_config_dir_env       = false;
     int      required                        = MPI_THREAD_MULTIPLE;
     int      provided                        = 0;
+    char    *subfiling_subfile_prefix_saved;
```
qkoziol: Same fix here?  (Either not saving the values or strdup'ing them)

glennsong09: Do you think saving the values is necessary? I wasn't sure. If you don't think so, I can just remove it there too.

jhendersonHDF: For the time being, we don't really need to save the values since this is the last set of tests that run. If another set of tests is interested in the values from the environment in the future we may need to save the values, but otherwise it's fine to remove them for now.

glennsong09: Okay, I can remove it from this test too.

jhendersonHDF (approved):

qkoziol (approved):

PR #4039, opened 2024-02-23
Fix VOL-compatibility issues in External Link API test 

mattjala:
* Fixed some external link tests not prepending the API  test prefix to the filename.
* Modified link tests that use `H5Iget_name` to accept any of the valid names for the target object, since [the API](https://docs.hdfgroup.org/hdf5/develop/group___h5_i.html#title6) does not specify which of an object's possible names will be returned.
* Fixed link iteration test counting skips as failures due to doing the capability flag checks from within the link callbacks. Now, the capability flags are checked before the link iteration is performed.

### DIFF
--- a/test/API/H5_api_link_test.c
+++ b/test/API/H5_api_link_test.c
@@ -900,7 +900,7 @@ test_create_hard_link_invalid_params(void)
         {
             TESTING_2("H5Lcreate_hard across files");
 
-            snprintf(ext_link_filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s", EXTERNAL_LINK_TEST_FILE_NAME);
+            snprintf(ext_link_filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s%s", test_path_prefix, EXTERNAL_LINK_TEST_FILE_NAME);
 
             if ((ext_file_id = H5Fcreate(ext_link_filename, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT)) < 0) {
                 H5_FAILED();
@@ -1999,7 +1999,7 @@ test_create_external_link(void)
         return 0;
     }
 
-    snprintf(ext_link_filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s", EXTERNAL_LINK_TEST_FILE_NAME);
+    snprintf(ext_link_filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s%s", test_path_prefix, EXTERNAL_LINK_TEST_FILE_NAME);
 
     if ((file_id = H5Fcreate(ext_link_filename, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT)) < 0) {
         H5_FAILED();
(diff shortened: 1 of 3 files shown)


test/API/H5_api_link_test.c
```diff
@@ -505,16 +506,32 @@ test_create_hard_link_many(void)
...
+
     /* Check name */
     if (H5Iget_name(group_id2, objname, (size_t)HARD_LINK_TEST_GROUP_MANY_NAME_BUF_SIZE) < 0) {
         H5_FAILED();
         printf("    couldn't get the name of the object '%s'\n", HARD_LINK_TEST_GROUP_MANY_FINAL_NAME);
         goto error;
     }
 
-    if (strcmp(objname, "/" LINK_TEST_GROUP_NAME "/" HARD_LINK_TEST_GROUP_MANY_NAME "/hard21")) {
+    for (size_t i = 1; i < HARD_LINK_TEST_GROUP_MANY_NUM_HARD_LINKS + 1; i++) {
```
fortnern, line 517: Might as well add a short circuit success here. For the test after the loop you could just switch the order to valid_name_matched || !strcmp(...

test/API/H5_api_link_test.c
```diff
@@ -1709,9 +1728,24 @@ test_create_soft_link_many(void)
         goto error;
     }
 
-    if (strcmp(objname, "/" LINK_TEST_GROUP_NAME "/" SOFT_LINK_TEST_GROUP_MANY_NAME "/soft16")) {
+    for (size_t i = 1; i < SOFT_LINK_TEST_GROUP_MANY_NAME_SOFT_LINK_COUNT + 1; i++) {
```
fortnern, line 1731: Same

test/API/H5_api_link_test.c
```diff
@@ -1999,7 +2033,8 @@ test_create_external_link(void)
         return 0;
     }
 
-    snprintf(ext_link_filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s", EXTERNAL_LINK_TEST_FILE_NAME);
+    snprintf(ext_link_filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s%s", test_path_prefix,
```
fortnern, line 2036: Seems that testing absolute and relative path resolutions are two different things. What is the motivation for changing it? Should we test both?

mattjala: This isn't related to path resolution within a file. Some VOLs need to place their test files within a folder (e.g. the REST VOL creates its test files under `/home/test_user1`). Other places in the API tests that use a filename apply the prefix, but a few of these external link tests didn't get updated. For the native vol, `test_path_prefix` is empty.

test/API/H5_api_link_test.c
```diff
@@ -20843,6 +21011,15 @@ test_link_iterate_mixed_links(void)
         {
             TESTING_2("H5Literate_by_name2 by link name in increasing order");
 
+            if (!(vol_cap_flags_g & H5VL_CAP_FLAG_CREATION_ORDER) ||
```
fortnern: Does this test and the next really need creation order?

test/API/H5_api_link_test.c
```diff
@@ -2795,9 +2852,20 @@ test_create_external_link_ping_pong(void)
...
             }
 
-            if (strcmp(objname, "/final/new_group")) {
+            for (size_t i = 1; i < EXTERNAL_LINK_TEST_PING_PONG_NUM_LINKS + 1; i++) {
+                char name_possibility[H5_API_TEST_FILENAME_MAX_LENGTH];
+
+                snprintf(name_possibility, H5_API_TEST_FILENAME_MAX_LENGTH, "%s%zu%s", "/link", i,
+                         "/new_group");
+
+                valid_name_matched = !strcmp(objname, name_possibility) || valid_name_matched;
```
fortnern, line 2861: (Minor) maybe add short circuit here

test/API/H5_api_link_test.c
mattjala: No, support for link name ordering shouldn't depend on creation order support. I've removed a few unnecessary checks for this flag.

fortnern (approved):

jhendersonHDF (approved):

PR #4040, opened 2024-02-23
Add upddated cmake tools from source location

byrnHDF:


config/sanitizer/README.md
```diff
@@ -45,15 +31,9 @@ This is a collection of quite useful scripts that expand the possibilities for b
...
-
-These obviously force the standard to be required, and also disables compiler-specific extensions, ie `--std=gnu++11`. This helps to prevent fragmenting the code base with items not available elsewhere, adhering to the agreed C++ standards only.
-
 ## Sanitizer Builds [`sanitizers.cmake`](sanitizers.cmake)
 
-Sanitizers are tools that perform checks during a program's runtime and return issues, and as such, along with unit testing, code coverage and static analysis, are another tool to add to the programmer's toolbox. And, of course, like the previous tools, they are simple to add to any project using CMake, allowing any project and developer to quickly and easily use them.
+Sanitizers are tools that perform checks during a program’s runtime and returns issues, and as such, along with unit testing, code coverage and static analysis, is another tool to add to the programmers toolbox. And of course, like the previous tools, are tragically simple to add into any project using CMake, allowing any project and developer to quickly and easily use.
```
lrknox, line 36: return and are were correct;  the subject  is either tools or sanitizers, and both are plural.  Also "tragically simple" seems a little odd, maybe "trivially simple" or just simple would be better? 
Also, unless there's a reason I missed, it might be best to more clearly identify the previous tools that are simple to add into any project.

config/sanitizer/README.md
```diff
@@ -72,16 +52,18 @@ A quick rundown of the tools available, and what they do:
     - Division by zero
     - Unreachable code
 - [MemorySanitizer](https://clang.llvm.org/docs/MemorySanitizer.html) detects uninitialized reads.
+- [Control Flow Integrity](https://clang.llvm.org/docs/ControlFlowIntegrity.html) is designed to detect certain forms of undefined behaviour that can potentially allow attackers to subvert the program's control flow.
 
-These are used by declaring the `USE_SANITIZER` CMake variable as one of:
+These are used by declaring the `USE_SANITIZER` CMake variable as string containing any of:
```
lrknox, line 57: "a string"?

byrnHDF:
The text is original from the source - I just removed what we don't use.

lrknox (approved):

lrknox:
> The text is original from the source - I just removed what we don't use.

All right, then.  I don't know why they changed the grammar from correct to incorrect in the README.md file.

jhendersonHDF (approved):

PR #4045, opened 2024-02-25
Remove check for gettimeofday + tz in CMake

derobins:
This is not used in the library


derobins:
All these CMake 'run a program' checks need to go. They make cross-compiling more complicated.

byrnHDF (approved):

byrnHDF (approved):

lrknox (approved):

PR #4046, opened 2024-02-25
Add options to allow tools type selection and naming

byrnHDF:


lrknox (approved):

CMakeLists.txt
```diff
@@ -492,6 +494,15 @@ if (ONLY_SHARED_LIBS)
   set (H5_ENABLE_STATIC_LIB NO)
   set (BUILD_SHARED_LIBS ON CACHE BOOL "Build Shared Libraries" FORCE)
   set (BUILD_STATIC_LIBS OFF CACHE BOOL "Build Static Libraries" FORCE)
+  if (HDF5_BUILD_STATIC_TOOLS)
+    message (WARNING "Cannot build static tools without static libraries. Building shared tools.")
+  endif ()
+  set (HDF5_BUILD_STATIC_TOOLS OFF CACHE BOOL "Build Static Tools NOT Shared Tools" FORCE)
+endif ()
+
+if (NOT BUILD_SHARED_LIBS AND NOT HDF5_BUILD_STATIC_TOOLS)
```
jhendersonHDF, line 503: Since the default is `OFF` for `HDF5_BUILD_STATIC_TOOLS`, it seems like anyone who builds only a static library by turning off `BUILD_SHARED_LIBS` is going to get this warning by default, which could be either annoying or surprising/confusing. Since the only case I can think of where someone will want to set the `HDF5_BUILD_STATIC_TOOLS` option to on is the case where both static and shared are built but they want static tools instead, I think we probably don't need this warning.

CMakeLists.txt
```diff
@@ -492,6 +494,15 @@ if (ONLY_SHARED_LIBS)
...
+  message (WARNING "Cannot build shared tools without shared libraries. Building static tools.")
```
jhendersonHDF: Since the default is `OFF` for `HDF5_BUILD_STATIC_TOOLS`, it seems like anyone who builds only a static library by turning off `BUILD_SHARED_LIBS` is going to get this warning by default, which could be either annoying or surprising/confusing. Since the only case I can think of where someone will want to set the `HDF5_BUILD_STATIC_TOOLS` option to on is the case where both static and shared are built but they want static tools instead, I think we probably don't need this warning.

config/sanitizer/README.md
jhendersonHDF, line 1: Change in this file seems unrelated. Was `img/code-cov.png` just missing or something?

release_docs/RELEASE.txt
jhendersonHDF, line 1: I would also have a sentence here mentioning that we're removing the "-shared" named tools and only have the regularly named tools, which are either linked to the shared or static library.

tools/src/misc/CMakeLists.txt
```diff
@@ -5,122 +5,88 @@ project (HDF5_TOOLS_SRC_MISC C)
...
-
-  set (H5_DEP_EXECUTABLES ${H5_DEP_EXECUTABLES}
-      h5debug-shared
-      h5repart-shared
-      h5mkgrp-shared
-      h5clear-shared
-      h5delete-shared
-  )
-endif ()
+set (H5_DEP_EXECUTABLES
```
jhendersonHDF, line 73: In the old static block, I see:

```
  set (H5_DEP_EXECUTABLES
      h5debug
      h5repart
      h5mkgrp
      h5clear
      h5delete
  )
```

but in the old shared block I see:

```
  set (H5_DEP_EXECUTABLES ${H5_DEP_EXECUTABLES}
      h5debug-shared
      h5repart-shared
      h5mkgrp-shared
      h5clear-shared
      h5delete-shared
  )
```

Should here and everywhere else that's setting `H5_DEP_EXECUTABLES` in other files actually be using the latter version to append to the list rather than overwriting it?

CMakeLists.txt
byrnHDF: Should it be changed to just a STATUS message? OR maybe VERBOSE and only show if that level is requested?

config/sanitizer/README.md
byrnHDF: So I removed the tools we don't use and I did not bring over any images that were unrelated to our needs in a different PR.  This may be just an accidental commit

tools/src/misc/CMakeLists.txt
byrnHDF: That should have been deleted.

CMakeLists.txt
byrnHDF: Used VERBOSE, which is a better choice.

tools/src/misc/CMakeLists.txt
jhendersonHDF: > That should have been deleted.

Do you mean `H5_DEP_EXECUTABLES` should have been deleted everywhere? Otherwise, the old static block is deleted in your changes, but where it gets set throughout different CMake files alternates between `set (H5_DEP_EXECUTABLES ...)` and `set (H5_DEP_EXECUTABLES ${H5_DEP_EXECUTABLES} ...)`. Just wondering if it's always supposed to be `set (H5_DEP_EXECUTABLES ${H5_DEP_EXECUTABLES} ...)` everywhere.

byrnHDF: No the variable is reset for each tool

byrnHDF: > > That should have been deleted.
> 
> Do you mean `H5_DEP_EXECUTABLES` should have been deleted everywhere? Otherwise, the old static block is deleted in your changes, but where it gets set throughout different CMake files alternates between `set (H5_DEP_EXECUTABLES ...)` and `set (H5_DEP_EXECUTABLES ${H5_DEP_EXECUTABLES} ...)`. Just wondering if it's always supposed to be `set (H5_DEP_EXECUTABLES ${H5_DEP_EXECUTABLES} ...)` everywhere.

Sorry - I thought you were referring to code that was still in the file.

jhendersonHDF: Oh I see, it's just collecting a list of targets for the `install` command and some files only need to set it once. That file was previously appending to the list in case both static and shared tools were built. Makes sense.

jhendersonHDF (approved):

Issue #3985, opened 2024-02-02
Enable the subfiling VFD by default


brtnfld:
If --enable-parallel (or equiv. in CMake) is specified, then the subfiling VFD should be enabled. Currently, it is disabled by default.

PR #4048, opened 2024-02-25
Automatically turn on Subfiling VFD if parallel is enabled in build options

glennsong09:
Automatically turn on Subfiling VFD if parallel is enabled in build options in both Autotools and CMake. 

Addresses #3985.


derobins:
Let's return to this in 1.14.5

PR #4057, opened 2024-03-04
Do not enable szip for sanitizer runs

byrnHDF:


jhendersonHDF:
Is this due to problems on the library side that can be fixed or issues in the szip filter itself?

byrnHDF:
The issues are in libaec.

derobins (approved):

jhendersonHDF (approved):

PR #4060, opened 2024-03-04
Update long double test with correct values

byrnHDF:
This replaces #3501

### DIFF
--- a/test/dsets.c
+++ b/test/dsets.c
@@ -5793,12 +5793,11 @@ test_floattypes(hid_t file)
 
     /* double */
     {
-        double orig_data[2][5] = {{(double)1.6081706885101836e+60L, (double)-255.32099170994480,
-                                   (double)1.2677579992621376e-61L, (double)64568.289448797700,
-                                   (double)-1.0619721778839084e-75L},
-                                  {(double)2.1499497833454840e+56L, (double)6.6562295504670740e-3,
-                                   (double)-1.5747263393432150, (double)1.0711093225222612,
-                                   (double)-9.8971679387636870e-1}};
+        double orig_data[2][5] = {
+            {(double)1.6081706885101836e+60L, (double)-255.32099170994480, (double)1.2677579992621376e-61L,
+             (double)64568.289448797700, (double)-1.0619721778839084e-75L},
+            {(double)2.1499497833454840e+56L, (double)6.6562295504670740e-3, (double)-1.5747263393432150,
+             (double)1.0711093225222612, (double)-9.8971679387636870e-1}};
         double new_data[2][5];
 
         TESTING("    double (setup)");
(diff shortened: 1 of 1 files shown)


test/dsets.c
```diff
@@ -5681,6 +5681,308 @@ test_multiopen(hid_t file)
...
+             (long double)-1.5747263393432150995747263393432150,
+             (long double)1.0711093225222612990711093225222612,
+             (long double)-9.8971679387636870998971679387636870e-1}};
+        long double new_data[2][5];
+
+        TESTING("    long double (setup)");
+
+        /* Define user-defined quad-precision floating-point type for dataset */
+        datatype  = H5Tcopy(H5T_NATIVE_LDOUBLE);
+        precision = 128;
```
jhendersonHDF, line 5898: The precision is actually 123 bits here. I'm surprised setting it to 128 doesn't cause conversion accuracy issues, but we should still set it to 123 bits here so we aren't lying.

test/dsets.c
```diff
@@ -5681,6 +5681,308 @@ test_multiopen(hid_t file)
...
+         *----------------------------------------------------------------------
+         */
+        TESTING("    long double (read)");
+
+        /* Read the dataset back */
+        if (H5Dread(dataset, H5T_NATIVE_LDOUBLE, H5S_ALL, H5S_ALL, H5P_DEFAULT, new_data) < 0)
+            goto error;
+
+        /* Check that the values read are the same as the values written
+         * Assume size of long long = size of double
```
jhendersonHDF: I'm not sure where this comment about long long = size of double came from. I don't see any long longs in this test.

test/dsets.c
```diff
@@ -5681,6 +5681,306 @@ test_multiopen(hid_t file)
...
+    {
+        long double orig_data[2][5] = {
+            {(long double)1.6081706885101836e+600L, (long double)-255.3209917099448032099170994480,
+             (long double)1.2677579992621376e-610L, (long double)64568.289448797700289448797700,
+             (long double)-1.0619721778839084e-750L},
+            {(long double)2.1499497833454840991499497833454840e+560L,
+             (long double)6.6562295504670740996562295504670740e-3,
+             (long double)-1.5747263393432150995747263393432150,
+             (long double)1.0711093225222612990711093225222612,
+             (long double)-9.8971679387636870998971679387636870e-1}};
```
brtnfld, line 5891: why hard code the values as opposed to using a math function? If not, how were these values chosen? They can go much bigger/smaller than this.

test/dsets.c
```diff
@@ -5681,6 +5681,306 @@ test_multiopen(hid_t file)
...
+        /* Read the dataset back */
+        if (H5Dread(dataset, H5T_NATIVE_LDOUBLE, H5S_ALL, H5S_ALL, H5P_DEFAULT, new_data) < 0)
+            goto error;
+
+        /* Check that the values read are the same as the values written */
+        for (i = 0; i < (size_t)size[0]; i++) {
+            for (j = 0; j < (size_t)size[1]; j++) {
+                if (isnan(orig_data[i][j]))
+                    continue; /* skip if value is NaN */
+                if (!H5_LDBL_ABS_EQUAL(new_data[i][j], orig_data[i][j])) {
```
brtnfld, line 5947: Are we sure that is checking the complete precision? I would think it would fail for the cases where the complete precision was not initialized. Also, do we expect NaN values? I would think that would be an error.

test/dsets.c
byrnHDF: It was just to test values that would be past the double range of values.
A programmatic way would be useful in the future.

test/dsets.c
byrnHDF: I could try, but I don't remember if I had problems if it didn't match the sign pos. Maybe because of the offset?

test/dsets.c
byrnHDF: This just follows the other float and double parts. 
These would be issues that should be noted.

jhendersonHDF: These macros just get the difference of the two float, double, or long double values and then take the absolute value and compare it to the relevant epsilon value. It should be fine for checking the data.

derobins (approved):

jhendersonHDF (approved):

PR #4062, opened 2024-03-04
New option for building with static CRT in Windows

### LINKED ISSUE #3984: Windows_MT.cmake has no effect anymore
Hello,
I was building latest static hdf5 1.14.3 libs on Windows targeting static MSVCRT libs via cmake. I found out about Windows_MT.cmake from readme. So I disabled build of shared libs, enabled usage of static msvcrt libs and included mentioned Windows_MT.cmake to UserMacros.cmake. To be sure I added few message prints to mentioned cmake file to be super duper sure that the code gets executed during the configure step.

Windows_MT.cmake does simple replace of /MD to /MT for the compiler and /libs:dll to /libs:static for the linker. However, the code is not working properly, since there is nothing to be replaced in vanilla state (freshly generated cmake project). Following is the printout from inside of the cmake loops:

```CMAKE_C_FLAGS=   /DWIN32 /D_WINDOWS -wd5105
CMAKE_C_FLAGS_DEBUG= /Zi /Ob0 /Od /RTC1
CMAKE_C_FLAGS_RELEASE= /O2 /Ob2 /DNDEBUG
CMAKE_C_FLAGS_MINSIZEREL= /O1 /Ob1 /DNDEBUG
CMAKE_C_FLAGS_RELWITHDEBINFO= /Zi /O2 /Ob1 /DNDEBUG
CMAKE_CXX_FLAGS= /DWIN32 /D_WINDOWS /GR /EHsc
CMAKE_CXX_FLAGS_DEBUG= /Zi /Ob0 /Od /RTC1
CMAKE_CXX_FLAGS_RELEASE= /O2 /Ob2 /DNDEBUG
CMAKE_CXX_FLAGS_MINSIZEREL= /O1 /Ob1 /DNDEBUG
CMAKE_CXX_FLAGS_RELWITHDEBINFO= /Zi /O2 /Ob1 /DNDEBUG
CMAKE_Fortran_FLAGS=
CMAKE_Fortran_FLAGS_DEBUG=
CMAKE_Fortran_FLAGS_RELEASE=
CMAKE_Fortran_FLAGS_MINSIZEREL=
CMAKE_Fortran_FLAGS_RELWITHDEBINFO=
```

You see that there are not /MD or /libs flags to be replaced. So I just added the else branch in order to append the flags if there is nothing to replace. That also didn't work properly. I could now see proper flags being appended  (/MT and /libs:static), but in my case generated Visual Studio solution was still showing me that MultiThreaded DLL library is used as Runtime library in project properties. After building many tests failed. I think the reason is the combination of (implicit) /MD and /NODEFAULTLIB:MSVCRT options. Only after adding 

```set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")```

the generated Visual Studio solution was showing me that static runtime lib is used. Then all compiled and all tests passed. Strangely enough I compiled hdf5 1.12.3 earlier that day with exactly the same tools and there just including Windows_MT.cmake was enough.

Can someone please check what changed in hdf5 cmake tooling so this feature got broken? Also can you please consider adding option `BUILD_STATIC_CRT_LIBS` to be a regular hdf cmake option and include Windows_MT.cmake by default? I see no reason why this option should be removed by default 🤔 

I tested the same behaviour on two computers (home and at work):
- latest up-to-date Visual Studio 2022 + 2022 compiler (v143)
- latest up-to-date Visual Studio 2022 + 2017 compiler (v141)
- cmake 3.28.x

### PR DESCRIPTION
**Pull request by byrnHDF**
Closes #3984

### CODE REVIEW

**Review by hyoklee — APPROVED** [review]

File: CMakeLists.txt [hunk 1]
```diff
@@ -942,6 +942,15 @@ if (HDF5_ENABLE_MAP_API)
   set (H5_HAVE_MAP_API 1)
 endif ()
 
+#-----------------------------------------------------------------------------
+# Option to Build with Static CRT libraries on Windows
+#-----------------------------------------------------------------------------
+option (HDF5_BUILD_STATIC_CRT_LIBS "Build With Static Windows CRT Libraries" OFF)
```
jhendersonHDF: Seems like this option is something you only want to bring in when building under Windows. Otherwise it's just adding more options to the space that aren't going to get used.

derobins (approved):

PR #4064, opened 2024-03-04
Increased H5FD_ROS3_MAX_SECRET_TOK_LEN to 4096 to accommodate long AWS secret tokens

### LINKED ISSUE #4063: The value of H5FD_ROS3_MAX_SECRET_TOK_LEN is too small
**Describe the bug**
The AWS STS routinely issues secret tokens that are longer than 1024 bytes. The value of H5FD_ROS3_MAX_SECRET_TOK_LEN is set at 1024, hence long secret tokens do not get added to the authorization headers.
 
The limitation is resent in the HEAD of the develop branch as 2024/03/04

aivazis:
Increased H5FD_ROS3_MAX_SECRET_TOK_LEN to 4096 to allow long AWS secret tokens to be included in the authorization headers.

While walking through the code in `H5FDros3` and `H5FDs3comms`, i bumped into some stuff, so i made the following tiny improvements:

- use the macro H5FD_ROS3_MAX_SECRET_TOK_LEN when allocating temporary buffers for constructing the headers, rather than a 1024 literal
- corrected the use of a dangling pointer when printing out host stats
- fixed some bad format specifiers
- stratified DEBUG_LEVEL into multiple values to allow for finer control over what gets printed out during debugging

Closes #4063

### DIFF
--- a/src/H5FDs3comms.c
+++ b/src/H5FDs3comms.c
@@ -47,9 +47,14 @@
 
 #ifdef H5_HAVE_ROS3_VFD
 
-/* toggle debugging (enable with 1)
+/* toggle debugging: pick a level
  */
-#define S3COMMS_DEBUG 0
+#define S3COMMS_DEBUG_NONE           0
+#define S3COMMS_DEBUG_REQUESTS       1
+#define S3COMMS_DEBUG_TRACE_API      2
+#define S3COMMS_DEBUG_TRACE_INTERNAL 3
+#define S3COMMS_DEBUG_HEADERS        4
+#define S3COMMS_DEBUG                S3COMMS_DEBUG_REQUESTS
 
 /* manipulate verbosity of CURL output
  * operates separately from S3COMMS_DEBUG
@@ -213,7 +218,7 @@ H5FD_s3comms_hrb_node_set(hrb_node_t **L, const char *name, const char *value)
 
     FUNC_ENTER_NOAPI_NOINIT
 
-#if S3COMMS_DEBUG
+#if S3COMMS_DEBUG >= S3COMMS_DEBUG_HEADERS
     fprintf(stdout, "called H5FD_s3comms_hrb_node_set.");
     printf("NAME: %s\n", name);
     printf("VALUE: %s\n", value);
@@ -292,7 +297,7 @@ H5FD_s3comms_hrb_node_set(hrb_node_t **L, const char *name, const char *value)
         if (value == NULL)
             HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "trying to remove node from empty list");
         else {
-#if S3COMMS_DEBUG
+#if S3COMMS_DEBUG >= S3COMMS_DEBUG_HEADERS
             printf("CREATE NEW\n");
             fflush(stdout);
 #endif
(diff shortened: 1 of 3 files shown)


mattjala (approved):

src/H5FDs3comms.c
```diff
@@ -47,9 +47,14 @@
...
-/* toggle debugging (enable with 1)
+/* toggle debugging: pick a level
  */
-#define S3COMMS_DEBUG 0
+#define S3COMMS_DEBUG_NONE           0
+#define S3COMMS_DEBUG_REQUESTS       1
+#define S3COMMS_DEBUG_TRACE_API      2
+#define S3COMMS_DEBUG_TRACE_INTERNAL 3
+#define S3COMMS_DEBUG_HEADERS        4
+#define S3COMMS_DEBUG                S3COMMS_DEBUG_REQUESTS
```
jhendersonHDF: Was the intention here to turn on comms debugging by default? We probably want all debugging off by default otherwise.

jhendersonHDF:
The changes look fine to me and are a definite improvement, but I believe this might cause some large stack frame warnings in -Werror builds due to the usage of `H5FD_ROS3_MAX_SECRET_TOK_LEN` in a struct in `tools/lib/h5tools_utils.h`. I don't believe we currently have any -Werror builds where we are enabling support for the ROS3 VFD, so we may want to check this.

src/H5FDs3comms.c
aivazis: no. oversight on my part. i'll push a fix asap

aivazis: thanks for catching it

aivazis:
> The changes look fine to me and are a definite improvement, but I believe this might cause some large stack frame warnings in -Werror builds due to the usage of `H5FD_ROS3_MAX_SECRET_TOK_LEN` in a struct in `tools/lib/h5tools_utils.h`. I don't believe we currently have any -Werror builds where we are enabling support for the ROS3 VFD, so we may want to check this.

we may be able to get away with a smaller value for the maximum token length. i'll make another pass at the AWS documentation for a tighter limit.

jhendersonHDF:
> we may be able to get away with a smaller value for the maximum token length. i'll make another pass at the AWS documentation for a tighter limit.

We should probably check first to see if this will even be an issue, but I'm also not a big fan of "large" arrays in structures anyway. We should probably rewrite the code a bit so we don't end up with that array on the stack.

src/H5FDs3comms.c
```diff
@@ -1645,7 +1654,7 @@ H5FD_s3comms_aws_canonical_request(char *canonical_request_dest, int _cr_size, c
     size_t      sh_size      = (size_t)_sh_size;
     size_t      cr_len       = 0; /* working length of canonical request str */
     size_t      sh_len       = 0; /* working length of signed headers str */
-    char        tmpstr[1024];
+    char        tmpstr[H5FD_ROS3_MAX_SECRET_TOK_LEN];
```
derobins, line 1657: I would allocate this dynamically.  4096 is probably big enough to start tripping our stack size warnings and we endeavor to be warning-free.

derobins:
> > The changes look fine to me and are a definite improvement, but I believe this might cause some large stack frame warnings in -Werror builds due to the usage of `H5FD_ROS3_MAX_SECRET_TOK_LEN` in a struct in `tools/lib/h5tools_utils.h`. I don't believe we currently have any -Werror builds where we are enabling support for the ROS3 VFD, so we may want to check this.
> 
> we may be able to get away with a smaller value for the maximum token length. i'll make another pass at the AWS documentation for a tighter limit.

I would just allocate it dynamically and then it's a non-problem.

derobins (approved):

PR #4066, opened 2024-03-05
Ignore UserPresets and Use only C compiler for sanitizers

byrnHDF:


.gitignore
jhendersonHDF, line 1: Unintentional change?

byrnHDF: No, there are two possible locations and neither file should be checked in

jhendersonHDF (approved):

hyoklee (approved):

derobins (approved):

PR #4070, opened 2024-03-05
Use H5T_STD_I32LE to create datatype in vds examples

byrnHDF:


brtnfld:
As these examples are provided, it is important to explain why H5T_NATIVE_INT should not be used for them. This would not be an appropriate example for someone using a BE machine. It would be preferable if the examples were not altered for testing.

derobins:
> As these examples are provided, it is important to explain why H5T_NATIVE_INT should not be used for them. This would not be an appropriate example for someone using a BE machine. It would be preferable if the examples were not altered for testing.

It's fine on a BE machine since it's the file type. I personally use precise types for creating datasets and native types for memory. If you wrote a middleware layer that used native type for the file type, you'd wind up with a mix of big- and little-endian files, which is probably not what people want when they standardize on HDF5.

derobins (approved):

brtnfld:
It is generally recommended to use native data types instead of specific types whenever possible. This is because specific types may not be the best choice when transferring code to a different machine. However, since it appears that we are not consistent with the examples currently, the PR is acceptable.

derobins:
In the March 7th working group, we decided that this was acceptable as a quick fix and that we should go over the examples for uniformity, etc. in the future (good potential HUG hackathon topic).

bmribler (approved):

PR #4071, opened 2024-03-05
Fix a memory leak in the cmpd_dset test

derobins:
This was due to not freeing a test buffer. It was not a core library memory leak.

### DIFF
--- a/test/cmpd_dset.c
+++ b/test/cmpd_dset.c
@@ -1115,6 +1115,7 @@ test_select_compound(char *fname, hid_t fapl, hid_t in_dxpl, unsigned set_fillva
 
     /* Release buffers */
     free(s1);
+    free(save_s1);
     free(s3);
     free(save_s3);
     free(rbuf1);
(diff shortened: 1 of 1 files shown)


test/cmpd_dset.c
```diff
@@ -1145,30 +1146,19 @@ test_select_compound(char *fname, hid_t fapl, hid_t in_dxpl, unsigned set_fillva
...
+    free(save_s3);
+    free(rbuf1);
+    free(rbuf3);
+    free(s7);
+    free(save_s7);
+    free(s8);
+    free(save_s8);
+    free(rbuf7);
+    free(rbuf8);
+    free(save_rbuf8);
```
derobins, line 1161: You can free a NULL pointer

derobins:
There is still a write of uninitialized bytes in the core library, but I'll fix that in a separate PR.

byrnHDF (approved):

jhendersonHDF (approved):

PR #4072, opened 2024-03-05
Fix uninitialized bytes in cmpd_dset test

derobins:
Compound fill values were set to the integer -1, causing valgrind to flag 'uninitialized bytes' errors.

This is just a problem with the cmpd_dset test and not a core library problem.

### DIFF
--- a/test/cmpd_dset.c
+++ b/test/cmpd_dset.c
@@ -450,8 +450,11 @@ test_select_src_subset(char *fname, hid_t fapl, hid_t in_dxpl, unsigned set_fill
     hsize_t        dims[2]       = {NX, NY};
     hsize_t        chunk_dims[2] = {NX / 10, NY / 10};
     unsigned char *rew_buf = NULL, *save_rew_buf = NULL, *rbuf = NULL;
-    int            fillvalue = (-1);
+    stype1         fillvalue;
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

jhendersonHDF (approved):

mattjala (approved):

fortnern:
Are you sure we're not losing something by changing the fill value from -1 to 0, which is more likely to be there "accidentally" if the library didn't fill the dataset due to a bug. I see using an int fill value for a compound type is wrong, but the fact that it was initialized to -1 implies it cares about the fill value being something other than 0. Maybe the code was copied from a different test where it matters more?

derobins:
> Are you sure we're not losing something by changing the fill value from -1 to 0, which is more likely to be there "accidentally" if the library didn't fill the dataset due to a bug. I see using an int fill value for a compound type is wrong, but the fact that it was initialized to -1 implies it cares about the fill value being something other than 0. Maybe the code was copied from a different test where it matters more?

It looks like a copy-paste error. I thought about initializing everything to -1, but that looked messy and the test clearly doesn't care about the fill value since it was initialized to "-1 + garbage bytes".

PR #4073, opened 2024-03-06
Fix uninitialized bytes in selection I/O test

derobins:
This was due to a complex type fill value being set to -1 instead of a proper complex value. This was a test problem and not a core library issue.


derobins:
Same problem as cmpd_dset.c

byrnHDF (approved):

jhendersonHDF (approved):

PR #4074, opened 2024-03-06
Fix bin/trace script w/ out params

### LINKED ISSUE #3733: bin/trace does not work properly with /*out*/ comments
Some or all API functions that have arguments with `\*out*\` comments, commonly used to indicate that an argument functions as an output, are given incorrect trace macros. Most or all of these arguments are mapped to `x` in the trace macro, no matter the actual type of the argument. Since `x` is only supposed to map to `void` (not `void *`) or `va_arg`, this is clearly a bug.

derobins:
The bin/trace script adds TRACE macros to public API calls in the main C library. This script had a parsing bug that caused functions that were annotated with /*out*/, etc. to be labeled as void pointers instead of typed pointers.

This is mainly a developer feature and not visible to consumers of the public API.

The bin/trace script now annotates public API calls properly.

Fixes GH #3733

### DIFF
--- a/bin/trace
+++ b/bin/trace
@@ -318,7 +318,7 @@ sub rewrite_func ($$$$$) {
     $args =~ s/\/\*\s*in\s*\*\///g;  # Get rid of /*in*/
     $args =~ s/\/\*\s*out\s*\*\///g; # Get rid of /*out*/
     $args =~ s/\/\*\s*in,\s*out\s*\*\///g; # Get rid of /*in,out*/
-    $args =~ s/\/\*\s*in_out\s*\*\///g; # Get rid of /*in,out*/
+    $args =~ s/\/\*\s*in_out\s*\*\///g; # Get rid of /*in_out*/
     $args =~ s/H5FL_TRACK_PARAMS//g; # Remove free list macro
     $args =~ s/\n#.*?\n/\n/g;        # Remove lines beginning with '#'
     my @args = split /,[\s\n]*/, $args;
--- a/bin/trace
+++ b/bin/trace
@@ -75,6 +75,7 @@ $Source = "";
                "H5F_file_space_type_t"      => "Ft",
                "H5F_libver_t"               => "Fv",
                "H5G_iterate_t"              => "Gi",
+               "H5G_info_t"                 => "GI",
                "H5G_obj_t"                  => "Go",
                "H5G_stat_t"                 => "Gs",
                "hsize_t"                    => "h",
--- a/bin/trace
+++ b/bin/trace
@@ -31,6 +31,7 @@ $Source = "";
                "H5A_operator1_t"            => "Ao",
                "H5A_operator2_t"            => "AO",
                "hbool_t"                    => "b",
+               "bool"                       => "b",
                "H5AC_cache_config_t"        => "Cc",
                "H5AC_cache_image_config_t"  => "CC",
                "double"                     => "d",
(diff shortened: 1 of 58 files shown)


bin/trace
```diff
@@ -320,8 +328,7 @@ sub rewrite_func ($$$$$) {
         next;
       }
       unless ($arg=~/^((\s*[a-z_A-Z](\w|\*)*\s+)+(\s*\*\s*|\s*const\s*|\s*volatile\s*)*)
-              ([a-z_A-Z]\w*)(\[.*?\])?
-              (\s*\/\*\s*(in|out|in_out)\s*\*\/)?\s*$/x) {
+              ([a-z_A-Z]\w*)(\[.*?\])?\s*$/x) {
```
derobins, line 333: Now we just strip out the comments, so there's no need to check for them in the `unless` regex

derobins:
Crack open bin/trace if you want to spot check any of these. There's a big map with all the assignments at the top of the file.

derobins:
Don't merge this yet. There are more missing types.

derobins:
In the March 7th working group, we decided to merge this PR, tag the commit as "removed_bin_trace" or similar, and then remove the bin/trace script and associated code. This will give us improved code if we decide to resurrect this scheme.

bin/trace
```diff
@@ -305,10 +307,18 @@ sub rewrite_func ($$$$$) {
...
+    # First remove:
+    #   * /*in*/, /*out*/, /*in_out*/, and /*in,out*/ comments
+    #   * preprocessor lines that start with #
+    #   * H5FL_TRACKING_PARAMS macros (free list code only)
+    #
+    # then split the function arguments on commas
+    $args =~ s/\/\*\s*in\s*\*\///g;  # Get rid of /*in*/
+    $args =~ s/\/\*\s*out\s*\*\///g; # Get rid of /*out*/
+    $args =~ s/\/\*\s*in,\s*out\s*\*\///g; # Get rid of /*in,out*/
+    $args =~ s/\/\*\s*in_out\s*\*\///g; # Get rid of /*in,out*/
```
jhendersonHDF: Very minor nitpick that doesn't really matter: `in,out` -> `in_out` in the comment

derobins: No problem. Will fix.

src/H5Shyper.c
```diff
@@ -4821,7 +4821,7 @@ H5Sget_select_hyper_blocklist(hid_t spaceid, hsize_t startblock, hsize_t numbloc
     herr_t ret_value; /* return value */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE4("e", "ihhx", spaceid, startblock, numblocks, buf);
+    H5TRACE4("e", "ihh*[a2]h", spaceid, startblock, numblocks, buf);
```
jhendersonHDF, line 4824: Interesting syntax here. Is this intended?

derobins: Ha. No idea. I didn't add anything like that, so I'd have to check to see what's going on in the trace code in H5trace.c.

src/H5Spoint.c
```diff
@@ -1625,7 +1625,7 @@ H5Sget_select_elem_pointlist(hid_t spaceid, hsize_t startpoint, hsize_t numpoint
     herr_t ret_value; /* return value */
 
     FUNC_ENTER_API(FAIL)
-    H5TRACE4("e", "ihhx", spaceid, startpoint, numpoints, buf);
+    H5TRACE4("e", "ihh*[a2]h", spaceid, startpoint, numpoints, buf);
```
jhendersonHDF, line 1628: Another one here as well if this isn't expected

byrnHDF (approved):

lrknox (approved):

qkoziol:
I'm a little late to this, but the change looks good.

PR #4078, opened 2024-03-07
Add support for the new MSVC preprocessor

### LINKED ISSUE #2515: Support of new MSVC preprocessor
**Describe the bug**

Compilation failure using MSVC compiler with `/Zc:preprocessor` option enabled.

**Expected behavior**

It should compile without issue.

 - HDF5 version: develop (7b833f04b5146bdad339ff10d42aadc416fb2f00)
 - OS and version: Windows 11
 - Compiler and version: MSVC 19.34
 - Build system and version: Cmake 3.26

**Additional context**

A new fully conformant preprocessor has been added recently to MSVC. In particular, it changes the way trailing commas in variadic macro are handled, which is described [here](https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170#comma-elision-in-variadic-macros). As a result, enabling this new preprocessor leads to compilation failures because of the way variadic macro are defined on Windows, e.g. [here](https://github.com/HDFGroup/hdf5/blob/7b833f04b5146bdad339ff10d42aadc416fb2f00/src/H5win32defs.h#L73). The use of the new preprocessor can be detected using defines, as explained [here](https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170#new-predefined-macro). It would be nice to update HDF5 in accordance.

derobins:
Microsoft added support for a new, standards-conformant preprocessor
to MSVC, which can be enabled with the /Zc:preprocessor option. This
preprocessor would trip over our HDopen() variadic function-like
macro, which, on Windows w/ MSVC, uses a quirk that only works
with the legacy MSVC preprocessor.

ifdefs have been added that select the correct HDopen() form and
allow building HDF5 with the /Zc:preprocessor option.

Fixes #2515

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -47,6 +47,21 @@ New Features
 
     Configuration:
     -------------
+    - Added support for the new MSVC preprocessor
+
+      Microsoft added support for a new, standards-conformant preprocessor
+      to MSVC, which can be enabled with the /Zc:preprocessor option. This
+      preprocessor would trip over our HDopen() variadic function-like
+      macro, which uses a feature that only works with the legacy preprocessor.
+
+      ifdefs have been added that select the correct HDopen() form and
+      allow building HDF5 with the /Zc:preprocessor option.
+
+      The HDopen() macro is located in an internal header file and only
+      affects building the HDF5 library from source.
+
+      Fixes GitHub #2515
+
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

glennsong09 (approved):

duburcqa:
Thank you for this !

PR #4081, opened 2024-03-07
Fixed asserts due to H5Pset_est_link_info() values

### LINKED ISSUE #1632: Assertion with certain H5Pset_est_link_info() parameters on group flush / close
Certain combinations of the est_num_entries and est_name_len parameter values can cause an assertion to be raised in debug builds of the HDF5 library when the group creation property list is used to create groups. The assertion arises when the group's object header is flushed to the disk. We are still characterizing this bug and determining its origin and effects in release builds.

This bug is present in the 1.8, 1.10, 1.12, and develop branches. We're still working to determine the exact version where it was introduced.

Reproducer:

```
int main() {
        hid_t file_id, group_id, plist_id, flist_id;
        herr_t res;

        flist_id = H5Pcreate(H5P_FILE_ACCESS);
        assert(flist_id != H5I_INVALID_HID);
        res = H5Pset_libver_bounds(flist_id, H5F_LIBVER_LATEST, H5F_LIBVER_LATEST);
        assert(res >= 0);

        file_id = H5Fcreate("test.h5", H5F_ACC_TRUNC, H5P_DEFAULT, flist_id);
        assert(file_id != H5I_INVALID_HID);

        plist_id = H5Pcreate(H5P_GROUP_CREATE);
        assert(plist_id != H5I_INVALID_HID);
        res = H5Pset_est_link_info(plist_id, 570, 100);
        assert(res >= 0);

        group_id = H5Gcreate(file_id, "g", H5P_DEFAULT, plist_id, H5P_DEFAULT);
        assert(group_id != H5I_INVALID_HID);

        H5Gclose(group_id);
        H5Pclose(plist_id);
        H5Fclose(file_id);
        H5Pclose(flist_id);
}
```

derobins:
If large values for est_num_entries and/or est_name_len were passed to H5Pset_est_link_info(), the library would attempt to create an object header NIL message to reserve enough space to hold the links in compact form (i.e., concatenated), which could exceed allowable object header message size limits and trip asserts in the library.

This bug only occurred when using the HDF5 1.8 file format or later and required the product of the two values to be ~64k more than the size of any links written to the group, which would cause the library to write out a too-large NIL spacer message to reserve the space for the unwritten links.

The library now inspects the phase change values to see if the dataset is likely to be compact and checks the size to ensure any NIL spacer messages won't be larger than the library allows.

Fixes #1632

### DIFF
--- a/src/H5Gobj.c
+++ b/src/H5Gobj.c
@@ -218,7 +218,25 @@ H5G__obj_create_real(H5F_t *f, const H5O_ginfo_t *ginfo, const H5O_linfo_t *linf
         assert(link_size);
 
         /* Compute size of header to use for creation */
-        hdr_size = linfo_size + ginfo_size + pline_size + (ginfo->est_num_entries * link_size);
+
+        /* Basic header size */
+        hdr_size = linfo_size + ginfo_size + pline_size;
+
+        /* If this is likely to be a compact group, add space for the link
+         * messages, unless the size of the link messages is greater than
+         * the largest allowable object header message size, since the size
+         * of the link messages is the size of the NIL spacer message that
+         * would have to be written out to reserve enough space to hold the
+         * links if the group were left empty.
+         */
+        bool compact = ginfo->est_num_entries <= ginfo->max_compact;
+        if (compact) {
+
+            size_t size_of_links = ginfo->est_num_entries * link_size;
+
+            if (size_of_links < H5O_MESG_MAX_SIZE)
+                hdr_size += size_of_links;
+        }
     } /* end if */
     else
         hdr_size = (size_t)(4 + 2 * H5F_SIZEOF_ADDR(f));
--- a/test/tmisc.c
+++ b/test/tmisc.c
@@ -338,6 +338,7 @@ typedef struct {
 #define CVE_2020_10812_FILENAME "cve_2020_10812.h5"
 
 #define MISC38_FILE "type_conversion_path_table_issue.h5"
+#define MISC39_FILE "set_est_link_info.h5"
 
 /****************************************************************
 **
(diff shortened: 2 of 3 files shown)


src/H5Gobj.c
```diff
@@ -218,7 +218,25 @@ H5G__obj_create_real(H5F_t *f, const H5O_ginfo_t *ginfo, const H5O_linfo_t *linf
...
+        hdr_size = linfo_size + ginfo_size + pline_size;
+
+        /* If this is likely to be a compact group, add space for the link
+         * messages, unless the size of the link messages is greater than
+         * the largest allowable object header message size, since the size
+         * of the link messages is the size of the NIL spacer message that
+         * would have to be written out to reserve enough space to hold the
+         * links if the group were left empty.
+         */
+        bool compact = ginfo->est_num_entries <= ginfo->max_compact;
```
qkoziol, line 232: We don't normally declare variables inline like this.  Why not just put the comparison in the 'if' statement?

qkoziol (approved):

src/H5Gobj.c
derobins: It's significantly uglier

test/tmisc.c
```diff
@@ -6445,6 +6446,103 @@ test_misc38(void)
...
+    ret = H5Fclose(fid);
+    CHECK(ret, FAIL, "H5Fclose");
+
+    /* Close the property lists */
+    ret = H5Pclose(fapl);
+    CHECK(ret, FAIL, "H5Pclose");
+    ret = H5Pclose(gcpl);
+    CHECK(ret, FAIL, "H5Pclose");
+
+} /* end test_misc25c() */
```
jhendersonHDF: "test_misc39", though it's probably high time we stop adding comments to the end of functions

src/H5Gobj.c
```diff
@@ -218,7 +218,25 @@ H5G__obj_create_real(H5F_t *f, const H5O_ginfo_t *ginfo, const H5O_linfo_t *linf
...
+        if (compact) {
+
+            size_t size_of_links = ginfo->est_num_entries * link_size;
+
+            if (size_of_links < H5O_MESG_MAX_SIZE)
```
jhendersonHDF, line 237: Don't you want to compare `hdr_size + size_of_links < H5O_MESG_MAX_SIZE` here or are the links the only thing that goes into the message?

test/tmisc.c
```diff
@@ -6514,6 +6612,7 @@ test_misc(void)
     test_misc36(); /* Exercise H5atclose and H5is_library_terminating */
     test_misc37(); /* Test for seg fault failure at file close */
     test_misc38(); /* Test for type conversion path table issue */
+    test_misc39(); /* Test for type conversion path table issue */
```
jhendersonHDF: Also copy pasted comment

src/H5Gobj.c
derobins: Just the links. The other stuff goes in other messages.

jhendersonHDF (approved):

PR #4084, opened 2024-03-07
Added macOS support statement

brtnfld:
This would mean we test starting with macOS 12 (Monterey), released October 25, 2021. The best estimates have older versions accounting for less than 3% of the install versions.

Also, do you know if the list of platform testing is correct, and is this maintainable going forward with all the different testing going on between GitHub, BB, and daily tests?

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -1288,6 +1288,12 @@ Bug Fixes since HDF5-1.14.0 release
 Platforms Tested
 ===================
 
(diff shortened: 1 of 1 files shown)


derobins (approved):

byrnHDF (commented):
The same statement could be said for Visual Studio and Ububtu/linux LTS versions, etc

jhendersonHDF (approved):

derobins:
> The same statement could be said for Visual Studio and Ububtu/linux LTS versions, etc

Yes. We should do that.

PR #4085, opened 2024-03-07
Expand check for variable-length or reference types when clearing datatype conversion paths

jhendersonHDF:
When clearing out datatype conversion paths involving variable-length or reference datatypes on file close, also check for these datatypes inside compound, variable-length or array datatypes.

#3942 accounted for conversion paths involving variable-length or reference datatypes directly, but did not account for paths involving these datatypes inside a container type, like a compound, variable-length or array datatype.

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -350,12 +350,13 @@ static H5T_path_t *H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const
                                        H5T_conv_func_t *conv);
 static bool        H5T_path_match(H5T_path_t *path, H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst,
                                   H5VL_object_t *owned_vol_obj, H5T_conv_t func);
-static bool        H5T__detect_vlen_ref(const H5T_t *dt);
-static H5T_t      *H5T__initiate_copy(const H5T_t *old_dt);
-static H5T_t      *H5T__copy_transient(H5T_t *old_dt);
-static H5T_t      *H5T__copy_all(H5T_t *old_dt);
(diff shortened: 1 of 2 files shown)


src/H5T.c
```diff
@@ -5214,6 +5216,65 @@ H5T_path_match(H5T_path_t *path, H5T_pers_t pers, const char *name, H5T_t *src,
...
+                break;
+
+            case H5T_INTEGER:
+            case H5T_FLOAT:
+            case H5T_TIME:
+            case H5T_STRING:
+            case H5T_BITFIELD:
+            case H5T_OPAQUE:
+            case H5T_REFERENCE: /* Should have been determined by above check */
+            case H5T_ENUM:
```
qkoziol, line 5271: I would think that the ENUM & VLEN parent (base) types should also be checked, why would they be covered by the earlier check?

jhendersonHDF: For enum, the base type can only be an integer type, so we shouldn't need to check that since only vlen and reference types will have a VOL object field set. But good point on the vlen of vlen; I wasn't thinking recursively enough!

qkoziol: Is it possible that the base type of the enum is a VOL object?  (I think the answer is yes, but haven't looked in detail)

jhendersonHDF: > Is it possible that the base type of the enum is a VOL object? (I think the answer is yes, but haven't looked in detail)

I suppose with a committed datatype you might have an enum where the base type is wrapped in a VOL object, but this function is only about checking the `owned_vol_obj` field in a datatype's shared structure, which only ever gets set for variable-length and reference datatypes. Adding a check against the base type of an enum shouldn't really hurt anything, but the enum type and its base type should never have the `owned_vol_obj` field set.

mattjala (approved):

derobins (approved):

PR #4087, opened 2024-03-08
Initialize selection type in both if/else sections

### LINKED ISSUE #3969: Potential "use before initialization" bug
**Describe the bug**
I ran the valgrind analyser and it showed a "use before initialization" here https://github.com/HDFGroup/hdf5/blob/develop/src/H5Dchunk.c#L3290

The field `fsel_type` was marked as uninitialized by valgrind. The field was created here https://github.com/HDFGroup/hdf5/blob/develop/src/H5Dchunk.c#L1047, after I initialized the field manually the warning disappeared.

Tested on 1.14.3

glennsong09:
The field fsel_type was never initialized in the if section of the if/else initialization code. 

Fixes #3969.

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -1206,8 +1206,8 @@ H5D__chunk_io_init_selections(H5D_io_info_t *io_info, H5D_dset_io_info_t *dinfo)
         fm->use_single = true;
 
         /* Initialize selection type in memory and file */
-	fm->msel_type = H5S_SEL_NONE;
-	fm->fsel_type = H5S_SEL_NONE;
+        fm->msel_type = H5S_SEL_NONE;
+        fm->fsel_type = H5S_SEL_NONE;
 
         /* Initialize single chunk dataspace */
         if (NULL == dataset->shared->cache.chunk.single_space) {
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -1062,8 +1062,8 @@ H5D__chunk_io_init(H5D_io_info_t *io_info, H5D_dset_io_info_t *dinfo)
     fm->single_piece_info = NULL;
 
     /* Initialize selection type in memory and file */
-    fm->msel_type         = H5S_SEL_ERROR;
-    fm->fsel_type         = H5S_SEL_ERROR;
+    fm->msel_type = H5S_SEL_ERROR;
+    fm->fsel_type = H5S_SEL_ERROR;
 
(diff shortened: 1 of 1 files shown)


derobins (changes requested):
I would probably initialize the chunk map in H5D__chunk_io_init() so the struct is fully initialized to sane values in a single location.

You might also change the allocation to use H5FL_CALLOC to avoid future uninitialized fields.

It might also be worth considering initializing the selection type fields to H5S_SEL_ERROR to further avoid uninitialized use.

src/H5Dchunk.c
```diff
@@ -1205,6 +1205,10 @@ H5D__chunk_io_init_selections(H5D_io_info_t *io_info, H5D_dset_io_info_t *dinfo)
         /* Initialize skip list for chunk selections */
         fm->use_single = true;
 
+        /* Initialize selection type in memory and file */
```
qkoziol: I suggest that you put this in H5D__chunk_io_init() instead, so that it doesn't get missed for another case in the future.

qkoziol (changes requested):

src/H5Dchunk.c
derobins: I see we are both working this evening :)

qkoziol:
LOL - @derobins and I with the same comment, simultaneously

jhendersonHDF:
> You might also change the allocation to use H5FL_CALLOC to avoid future uninitialized fields.

I believe the chunk map used to be allocated with calloc and it was switched back to malloc because the allocation overhead of calloc in a tight loop from a Mathworks example was proving to be a bit much.

derobins:
> > You might also change the allocation to use H5FL_CALLOC to avoid future uninitialized fields.
> 
> I believe the chunk map used to be allocated with calloc and it was switched back to malloc because the allocation overhead of calloc in a tight loop from a Mathworks example was proving to be a bit much.

That's fine. Just setting the two values in that function would be a good fix.

glennsong09:
> > > You might also change the allocation to use H5FL_CALLOC to avoid future uninitialized fields.
> > 
> > 
> > I believe the chunk map used to be allocated with calloc and it was switched back to malloc because the allocation overhead of calloc in a tight loop from a Mathworks example was proving to be a bit much.
> 
> That's fine. Just setting the two values in that function would be a good fix.

Okay, I updated the initialization location and values.

jhendersonHDF (approved):

derobins (approved):

qkoziol (approved):

PR #4089, opened 2024-03-08
Remove H5B debug checks

derobins:
The H5B (version 1 B-tree) package would add some computationally expensive integrity checks when H5B_DEBUG was defined. Due to their negative effects on performance, this option was rarely turned on, making the H5B__assert() check function stale, if not dead, code.

This change:

* Builds H5B__assert() when NDEBUG is not defined (the function relies on assert()) so it gets compiled more often.
* Removes some printf debugging statements in the B-tree code
* Removes all H5B "extra debug" checks that are leftover from past debugging sessions. Maintainers can add H5B__assert() selectively to perform integrity checks when debugging.
* Removes the HDF5_ENABLE_DEBUG_H5B CMake option

H5B_DEBUG now has no effect


byrnHDF (approved):

src/H5B.c
```diff
@@ -944,14 +915,8 @@ H5B__insert_helper(H5F_t *f, H5B_ins_ud_t *bt_ud, const H5B_class_t *type, uint8
 #endif /* H5_STRICT_FORMAT_CHECKS */
     }
     else if (cmp) {
-        /*
-         * We couldn't figure out which branch to follow out of this node. THIS
-         * IS A MAJOR PROBLEM THAT NEEDS TO BE FIXED --rpm.
-         */
-        assert("INTERNAL HDF5 ERROR (contact rpm)" && 0);
-#ifdef NDEBUG
-        HDabort();
```
jhendersonHDF, line 953: While the library shouldn't be aborting, removing this without having a fix here seems like a bad idea for release builds.

src/H5B.c
```diff
@@ -944,14 +915,8 @@ H5B__insert_helper(H5F_t *f, H5B_ins_ud_t *bt_ud, const H5B_class_t *type, uint8
...
-#endif /* NDEBUG */
+        /* We couldn't figure out which branch to follow out of this node */
+        assert("INTERNAL HDF5 ERROR" && 0);
```
derobins: I'll turn this into a real error handler

src/H5B.c
derobins: Fixed below

jhendersonHDF (approved):

PR #4097, opened 2024-03-09
Change how stats are printed in H5Z

derobins:
H5Z used the soon-to-be-removed HDEBUG macro to decide if stats would be dumped and to what stream. This is now handled by a DUMP_DEBUG_STATS_g variable and the output is always sent to stdout.

This is an internal change, not normally visible to users.

### DIFF
--- a/src/H5Z.c
+++ b/src/H5Z.c
@@ -29,6 +29,11 @@
 #include "szlib.h"
 #endif
 
+/* This module can collect and optionally dump some simple filter statistics
+ * to stdout. If you want to do this, you need to define H5Z_DEBUG and set
+ * the DUMP_DEBUG_STATS_g flag to true.
+ */
+
 /* Local typedefs */
 #ifdef H5Z_DEBUG
 typedef struct H5Z_stats_t {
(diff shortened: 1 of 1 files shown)


derobins:
I'm on the fence about removing this completely. This change basically leaves the existing scheme in place, without creating dead code.

byrnHDF (approved):

jhendersonHDF:
> H5Z used the soon-to-be-removed HDEBUG macro to decide if stats would be dumped and to what stream. This is now handled by a DUMP_DEBUG_STATS_g variable and the output is always sent to stdout.
> 
> This is an internal change, not normally visible to users.

What's the motivation for removing the whole HDEBUG scheme? I use it from time to time and it's nice to be able to turn individual package debugging on and off quickly with an environment variable. Seems we could add a single flag to the setting for HDF5_DEBUG that determines whether or not you want print output from any of the packages rather than having these new package-individual schemes.

qkoziol:
I don't think these stats are useful, they are just leftovers from early days of the feature.

qkoziol (approved):

PR #4102, opened 2024-03-09
long double tests has problems setting precision with offset

byrnHDF:
Needs testing on problem machines.

### DIFF
--- a/test/dsets.c
+++ b/test/dsets.c
@@ -5898,9 +5898,9 @@ test_floattypes(hid_t file)
         precision = 128;
         if (H5Tset_precision(datatype, precision) < 0)
             goto error;
-        if (H5Tset_fields(datatype, (size_t)127, (size_t)112, (size_t)15, (size_t)5, (size_t)107) < 0)
+        if (H5Tset_fields(datatype, (size_t)127, (size_t)112, (size_t)15, (size_t)0, (size_t)112) < 0)
             goto error;
-        offset = 5;
+        offset = 0;
         if (H5Tset_offset(datatype, offset) < 0)
             goto error;
         if (H5Tset_size(datatype, (size_t)16) < 0)
--- a/test/dsets.c
+++ b/test/dsets.c
@@ -5877,6 +5877,7 @@ test_floattypes(hid_t file)
         if (H5Dclose(dataset) < 0)
             goto error;
     }
+#if 0
 #if H5_SIZEOF_LONG_DOUBLE != H5_SIZEOF_DOUBLE
     /* long double */
     {
@@ -5966,6 +5967,7 @@ test_floattypes(hid_t file)
         if (H5Dclose(dataset) < 0)
             goto error;
     }
+#endif
 #endif
 
     return SUCCEED;


lrknox:
testing on echidna;  will follow with test on vortex

From: Allen Byrne ***@***.***>
Sent: Saturday, March 9, 2024 10:29 AM
To: HDFGroup/hdf5 ***@***.***>
Cc: Larry Knox ***@***.***>; Review requested ***@***.***>
Subject: [HDFGroup/hdf5] long double tests has problems setting precision with offset (PR #4102)


Needs testing on problem machines.

You can view, comment on, or merge this pull request online at:

  https://github.com/HDFGroup/hdf5/pull/4102

Commit Summary

  *   c20021c<https://github.com/HDFGroup/hdf5/pull/4102/commits/c20021cce8d618911590cb73133c6a304d4a33b6> long double tests has problems setting precision with offset

File Changes

(1 file<https://github.com/HDFGroup/hdf5/pull/4102/files>)

  *   M test/dsets.c<https://github.com/HDFGroup/hdf5/pull/4102/files#diff-25dbef713138b9ba29006d404ed7967a82bf3687560b84b6aee5748468040199> (4)

Patch Links:

  *   https://github.com/HDFGroup/hdf5/pull/4102.patch
  *   https://github.com/HDFGroup/hdf5/pull/4102.diff

test/dsets.c
```diff
@@ -5898,9 +5898,9 @@ test_floattypes(hid_t file)
         precision = 128;
         if (H5Tset_precision(datatype, precision) < 0)
             goto error;
-        if (H5Tset_fields(datatype, (size_t)127, (size_t)112, (size_t)15, (size_t)5, (size_t)107) < 0)
+        if (H5Tset_fields(datatype, (size_t)127, (size_t)112, (size_t)15, (size_t)0, (size_t)112) < 0)
             goto error;
-        offset = 5;
+        offset = 0;
```
jhendersonHDF, line 5904: Have you tried setting the precision to 123 instead of changing the offset? Setting the precision to 128 with an offset of 5 is lying to the library.

lrknox:
still fails on echidna byrnHDF<https://github.com/byrnHDF/hdf5/tree/develop-ld-dsets>:<https://github.com/byrnHDF/hdf5/tree/develop-ld-dsets>develop-ld-dset<https://github.com/byrnHDF/hdf5/tree/develop-ld-dsets>s

jhendersonHDF:
> still fails on echidna

What's the error? If echidna has true 128-bit floats, then setting the precision will have to come _after_ the call to `H5Tset_fields` since the fields have to be adjusted before precision can be reduced. On many machines long double has 80 bits of precision, so `H5Tset_precision` can be called before `H5Tset_fields` since it would be increasing the precision. If this is the problem, we'll need to check the precision of the long double copy type before deciding the order to call the functions.

lrknox:
All I see in dsets.txt is
Testing     long double (read)                                        *FAILED*
    Read different values than written.
    At index 0,0
dsets.txt.err is empty

jhendersonHDF:
> All I see in dsets.txt is
> Testing     long double (read)                                        *FAILED*
>    Read different values than written.
>    At index 0,0
> dsets.txt.err is empty

It might be that echidna's using a 64-bit long double format. Passing the "-mlong-double-64" flag to GCC shows the test failure on my "typical" X86 little-endian system.

jhendersonHDF:
You can see this if you look at the build warnings on Lassen, for example: https://my.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=2516582. The floating-point constants chosen are outside of the range of `long double` on these machines and causing the values to change. Passing the "-mlong-double-64" flag to GCC shows the warning:

```
warning: floating constant truncated to zero [-Woverflow]
 5885 |              (long double)1.2677579992621376e-610L, (long double)64568.289448797700289448797700,
      |              ^

```

byrnHDF:
I tried 123 and it failed  - complained about sign pos

byrnHDF:
Looks like this needs some library level work (library developer skill) to get this right.

Should I just disable it?

jhendersonHDF:
> I tried 123 and it failed  - complained about sign pos

We should still figure out why that's failing and fix it. 128 is just wrong for the precision when the offset is 5. 128 is fine if this PR changes the offset to 0, but it's good to test with types with offsets here.

> Looks like this needs some library level work (library developer skill) to get this right.

Seems like all you need to do is pick smaller constants that could fit inside a 64-bit long double value, since that appears to be the smallest representation I can find across most architectures.

byrnHDF:
> > I tried 123 and it failed  - complained about sign pos
> 
> We should still figure out why that's failing and fix it. 128 is just wrong for the precision when the offset is 5. 128 is fine if this PR changes the offset to 0, but it's good to test with types with offsets here.
> 
> > Looks like this needs some library level work (library developer skill) to get this right.
> 
> Seems like all you need to do is pick smaller constants that could fit inside a 64-bit long double value, since that appears to be the smallest representation I can find across most architectures.

 Would a 64-bit long double value be any different then the double? Could we still use the same long double values? 
So set fields would be, 63, 53, 10, 5, 48 ? precision = 58   ?

lrknox:
with "setenv CFLAGS -mlong-double-64" on echidna, all tests passed, though 26 H5IMPORT and H5REPACK tests didn't run.

jhendersonHDF:
> Would a 64-bit long double value be any different then the double? Could we still use the same long double values? So set fields would be, 63, 53, 10, 5, 48 ? precision = 58 ?

Yes, I believe the standard just says there is an ordering from float -> double -> long double so double could be smaller than a 64-bit long double in theory. Try these changes though (it's a patch file but I had to name it .txt to upload):

[long_double.txt](https://github.com/HDFGroup/hdf5/files/14547379/long_double.txt)

bmribler (approved):

test/dsets.c
byrnHDF: It would not compile with 123 - complaint about sign position.

lrknox (approved):

PR #4103, opened 2024-03-09
Remove H5DEBUG() calls from H5Dmpio.c

derobins:
Just use stdout when a stream is needed.

### DIFF
--- a/src/H5Dmpio.c
+++ b/src/H5Dmpio.c
@@ -614,8 +614,7 @@ H5D__mpio_debug_init(void)
     if (debug_str)
         H5D__mpio_parse_debug_str(debug_str);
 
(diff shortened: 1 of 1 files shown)


derobins:
@jhendersonHDF - Is the printf debugging scheme in this file still useful?

jhendersonHDF:
Very useful, yes. Though it's probably fine to just use stdout here. As long as the `debug_stream` assignment on ~line 1190 stays in. It's useful to pipe parallel debugging output to separate files based on MPI rank number; otherwise, all the output gets interleaved and can be next to useless.

jhendersonHDF (approved):

byrnHDF (approved):

qkoziol (approved):

PR #4104, opened 2024-03-10
Refactor datatype conversion code to use pointers rather than IDs

jhendersonHDF:
The datatype conversion code previously used IDs for the source and destination datatypes rather than pointers to the internal structures for those datatypes. This was mostly due to the need for an ID for these datatypes that can be passed to an application-registered datatype conversion function or datatype conversion exception function. However, using IDs internally caused a lot of unnecessary ID lookups and hurt performance of datatype conversions in general. This was especially problematic for compound datatype conversions, where the ID lookups were occuring on every member of every compound element of a dataset. The code has now been refactored to use pointers internally and only create IDs for datatypes when necessary.

Fixed a test issue in dt_arith where a library datatype conversion function was being cast to an application conversion function. Since the two have different prototypes, this started failing after the parameters for a library conversion function changed from hid_t to H5T_t * and an extra parameter was added. This appears to have worked coincidentally in the past since the only different between a library conversion function and application conversion function was an extra DXPL parameter at the end of an application conversion function.

Fixed an issue where memory wasn't being freed in the h5fc_chk_idx test program. Even though the program exits quickly after allocating the memory, it still causes failures when testing with -fsanitize=address.

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -179,16 +179,15 @@ typedef struct H5D_chunk_it_ud3_t {
     bool                  do_convert;   /* Whether to perform type conversions */
 
     /* needed for converting variable-length data */
-    hid_t        tid_src;          /* Datatype ID for source datatype */
-    hid_t        tid_dst;          /* Datatype ID for destination datatype */
-    hid_t        tid_mem;          /* Datatype ID for memory datatype */
-    const H5T_t *dt_src;           /* Source datatype */
-    H5T_path_t  *tpath_src_mem;    /* Datatype conversion path from source file to memory */
-    H5T_path_t  *tpath_mem_dst;    /* Datatype conversion path from memory to dest. file */
-    void        *reclaim_buf;      /* Buffer for reclaiming data */
-    size_t       reclaim_buf_size; /* Reclaim buffer size */
-    uint32_t     nelmts;           /* Number of elements in buffer */
-    H5S_t       *buf_space;        /* Dataspace describing buffer */
+    H5T_t      *dt_src;           /* Source datatype */
+    H5T_t      *dt_dst;           /* Destination datatype */
+    H5T_t      *dt_mem;           /* Memory datatype */
+    H5T_path_t *tpath_src_mem;    /* Datatype conversion path from source file to memory */
+    H5T_path_t *tpath_mem_dst;    /* Datatype conversion path from memory to dest. file */
+    void       *reclaim_buf;      /* Buffer for reclaiming data */
+    size_t      reclaim_buf_size; /* Reclaim buffer size */
+    uint32_t    nelmts;           /* Number of elements in buffer */
+    H5S_t      *buf_space;        /* Dataspace describing buffer */
 
     /* needed for compressed variable-length data */
     const H5O_pline_t *pline;      /* Filter pipeline */
@@ -297,9 +296,9 @@ static herr_t   H5D__create_piece_file_map_all(H5D_dset_io_info_t *di, H5D_io_in
 static herr_t   H5D__create_piece_file_map_hyper(H5D_dset_io_info_t *di, H5D_io_info_t *io_info);
 static herr_t   H5D__create_piece_mem_map_1d(const H5D_dset_io_info_t *di);
 static herr_t   H5D__create_piece_mem_map_hyper(const H5D_dset_io_info_t *di);
-static herr_t   H5D__piece_file_cb(void *elem, const H5T_t *type, unsigned ndims, const hsize_t *coords,
+static herr_t   H5D__piece_file_cb(void *elem, H5T_t *type, unsigned ndims, const hsize_t *coords,
                                    void *_opdata);
-static herr_t   H5D__piece_mem_cb(void *elem, const H5T_t *type, unsigned ndims, const hsize_t *coords,
+static herr_t   H5D__piece_mem_cb(void *elem, H5T_t *type, unsigned ndims, const hsize_t *coords,
                                   void *_opdata);
 static herr_t   H5D__chunk_may_use_select_io(H5D_io_info_t *io_info, const H5D_dset_io_info_t *dset_info);
 static unsigned H5D__chunk_hash_val(const H5D_shared_t *shared, const hsize_t *scaled);
@@ -2234,7 +2233,7 @@ done:
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5D__piece_file_cb(void H5_ATTR_UNUSED *elem, const H5T_t H5_ATTR_UNUSED *type, unsigned ndims,
+H5D__piece_file_cb(void H5_ATTR_UNUSED *elem, H5T_t H5_ATTR_UNUSED *type, unsigned ndims,
                    const hsize_t *coords, void *_opdata)
 {
     H5D_io_info_wrap_t *opdata  = (H5D_io_info_wrap_t *)_opdata;
@@ -2360,7 +2359,7 @@ done:
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5D__piece_mem_cb(void H5_ATTR_UNUSED *elem, const H5T_t H5_ATTR_UNUSED *type, unsigned ndims,
+H5D__piece_mem_cb(void H5_ATTR_UNUSED *elem, H5T_t H5_ATTR_UNUSED *type, unsigned ndims,
                   const hsize_t *coords, void *_opdata)
 {
     H5D_io_info_wrap_t *opdata = (H5D_io_info_wrap_t *)_opdata;
@@ -4590,8 +4589,8 @@ H5D__chunk_lock(const H5D_io_info_t H5_ATTR_NDEBUG_UNUSED *io_info, const H5D_ds
                     /* Initialize the fill value buffer */
                     /* (use the compact dataset storage buffer as the fill value buffer) */
                     if (H5D__fill_init(&fb_info, chunk, NULL, NULL, NULL, NULL,
-                                       &dset->shared->dcpl_cache.fill, dset->shared->type,
-                                       dset->shared->type_id, (size_t)0, chunk_size) < 0)
+                                       &dset->shared->dcpl_cache.fill, dset->shared->type, (size_t)0,
+                                       chunk_size) < 0)
                         HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, NULL, "can't initialize fill buffer info");
                     fb_info_init = true;
 
@@ -5020,8 +5019,8 @@ H5D__chunk_allocate(const H5D_t *dset, bool full_overwrite, const hsize_t old_di
         /* Initialize the fill value buffer */
         /* (delay allocating fill buffer for VL datatypes until refilling) */
         if (H5D__fill_init(&fb_info, NULL, H5D__chunk_mem_alloc, pline, H5D__chunk_mem_free, pline,
-                           &dset->shared->dcpl_cache.fill, dset->shared->type, dset->shared->type_id,
-                           (size_t)0, orig_chunk_size) < 0)
+                           &dset->shared->dcpl_cache.fill, dset->shared->type, (size_t)0,
+                           orig_chunk_size) < 0)
             HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "can't initialize fill buffer info");
         fb_info_init = true;
 
@@ -5793,8 +5792,7 @@ H5D__chunk_prune_fill(H5D_chunk_it_ud1_t *udata, bool new_unfilt_chunk)
     if (!udata->fb_info_init) {
         H5_CHECK_OVERFLOW(udata->elmts_per_chunk, uint32_t, size_t);
         if (H5D__fill_init(&udata->fb_info, NULL, NULL, NULL, NULL, NULL, &dset->shared->dcpl_cache.fill,
-                           dset->shared->type, dset->shared->type_id, (size_t)udata->elmts_per_chunk,
-                           chunk_size) < 0)
+                           dset->shared->type, (size_t)udata->elmts_per_chunk, chunk_size) < 0)
             HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "can't initialize fill buffer info");
         udata->fb_info_init = true;
     } /* end if */
@@ -6576,6 +6574,7 @@ H5D__chunk_copy_cb(const H5D_chunk_rec_t *chunk_rec, void *_udata)
     bool                need_insert = false; /* Whether the chunk needs to be inserted into the index */
 
     /* General information about chunk copy */
+    H5T_t             *dt_src   = udata->dt_src;
     void              *bkg      = udata->bkg;      /* Background buffer for datatype conversion */
     void              *buf      = udata->buf;      /* Chunk buffer for I/O & datatype conversions */
     size_t             buf_size = udata->buf_size; /* Size of chunk buffer */
@@ -6608,9 +6607,9 @@ H5D__chunk_copy_cb(const H5D_chunk_rec_t *chunk_rec, void *_udata)
 
     /* Check parameter for type conversion */
     if (udata->do_convert) {
-        if (H5T_detect_class(udata->dt_src, H5T_VLEN, false) > 0)
+        if (H5T_detect_class(dt_src, H5T_VLEN, false) > 0)
             is_vlen = true;
-        else if ((H5T_get_class(udata->dt_src, false) == H5T_REFERENCE) &&
+        else if ((H5T_get_class(dt_src, false) == H5T_REFERENCE) &&
                  (udata->file_src != udata->idx_info_dst->f))
             fix_ref = true;
         else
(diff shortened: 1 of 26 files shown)


jhendersonHDF:
There are some other changes I'd like to make in the H5T code, but I wanted to keep this as small as possible and get it out of the way first. I have not yet added a RELEASE.txt entry because this change doesn't really fit under new feature or bugs fixed. Maybe we should have a "performance enhancements" section?

Some quick testing shows that this gives at least a 3x improvement in simple cases of compound conversions, but I'm still profiling to find other spots that can be improved on.

src/H5Dchunk.c
```diff
@@ -297,9 +296,9 @@ static herr_t   H5D__create_piece_file_map_all(H5D_dset_io_info_t *di, H5D_io_in
 static herr_t   H5D__create_piece_file_map_hyper(H5D_dset_io_info_t *di, H5D_io_info_t *io_info);
 static herr_t   H5D__create_piece_mem_map_1d(const H5D_dset_io_info_t *di);
 static herr_t   H5D__create_piece_mem_map_hyper(const H5D_dset_io_info_t *di);
-static herr_t   H5D__piece_file_cb(void *elem, const H5T_t *type, unsigned ndims, const hsize_t *coords,
```
jhendersonHDF, line 300: There are a lot of `const`s that had to be removed in this PR, mostly because the `H5T_t *` for the datatypes aren't being hidden behind an `hid_t` anymore and are used directly, where they end up getting passed to either `H5T_close` or `H5T_convert`, both of which would require casting const away. If we're able to mark the src and dst datatype parameters to `H5T_convert` as `const`, a good chunk of these can probably come back, but I wasn't sure if that was reasonable to do. Usage of `H5T_close` is more problematic.

src/H5Ofill.c
```diff
@@ -713,32 +707,22 @@ H5O__fill_old_size(const H5F_t H5_ATTR_UNUSED *f, const void *_fill)
...
 
     assert(fill);
 
     if (fill->buf) {
         if (fill->type && H5T_detect_class(fill->type, H5T_VLEN, false) > 0) {
-            H5T_t *fill_type;  /* Copy of fill value datatype */
             H5S_t *fill_space; /* Scalar dataspace for fill value element */
 
-            /* Copy the fill value datatype and get an ID for it */
-            if (NULL == (fill_type = H5T_copy(fill->type, H5T_COPY_TRANSIENT)))
```
jhendersonHDF, line 729: No longer need to copy this datatype since we're not creating an ID and `H5Treclaim` won't modify it

src/H5Tpkg.h
```diff
@@ -150,9 +150,17 @@ struct H5T_stats_t {
     H5_timevals_t times;  /*total time for conversion	     */
 };
 
+/* Context struct for information used during datatype conversions */
+typedef struct H5T_conv_ctx_t {
```
jhendersonHDF, line 154: New structure that holds relevant information during conversion to make conversions on compound and other container types faster. Similar to `H5T_cdata_t`, but that structure is public since it gets passed to an app conversion callback and the information we want here is partly private.

src/H5Tpkg.h
```diff
@@ -479,357 +487,472 @@ H5_DLL herr_t H5T__commit_named(const H5G_loc_t *loc, const char *name, H5T_t *d
...
-                                     size_t buf_stride, size_t bkg_stride, void *buf, void *bkg);
-H5_DLL herr_t H5T__conv_ldouble_long(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts,
-                                     size_t buf_stride, size_t bkg_stride, void *buf, void *bkg);
-H5_DLL herr_t H5T__conv_ldouble_ulong(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts,
-                                      size_t buf_stride, size_t bkg_stride, void *buf, void *bkg);
-H5_DLL herr_t H5T__conv_ldouble_llong(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts,
-                                      size_t buf_stride, size_t bkg_stride, void *buf, void *bkg);
-H5_DLL herr_t H5T__conv_ldouble_ullong(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t nelmts,
-                                       size_t buf_stride, size_t bkg_stride, void *buf, void *bkg);
+H5_DLL herr_t H5T__conv_noop(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t *conv_ctx,
```
jhendersonHDF, line 496: All the conversion function prototypes here switched from `hid_t src_id, hid_t dst_id` -> `H5T_t *src, H5T_t *dst` and got an extra `const H5T_conv_ctx_t *conv_ctx` parameter.

src/H5Tpkg.h
```diff
@@ -479,357 +487,472 @@ H5_DLL herr_t H5T__commit_named(const H5G_loc_t *loc, const char *name, H5T_t *d
 H5_DLL H5T_t *H5T__open_name(const H5G_loc_t *loc, const char *name);
 H5_DLL hid_t  H5T__get_create_plist(const H5T_t *type);
 
+/* Helper function for H5T_convert that accepts a pointer to a H5T_conv_ctx_t structure */
+H5_DLL herr_t H5T_convert_with_ctx(H5T_path_t *tpath, H5T_t *src_type, H5T_t *dst_type,
```
jhendersonHDF, line 491: This is a new helper function for `H5T_convert` that helps avoid a severe amount of overhead during compound and other container datatype conversions where `H5T_convert` is called for every single member of every element. With the changes in this PR, the `H5T_conv_ctx_t` structure gets initialized in `H5T_convert`, meaning that any conversion function calling `H5T_convert` in a loop would initialize a new structure for every member of every element. In the case where an application-registered conversion exception function is being used, this would cause the conversion function to register and then tear down an ID for the source and destination datatypes each time, which was very expensive. Container type conversion functions can instead copy the current `H5T_conv_ctx_t` structure into a temporary one, update its `src_type_id` and `dst_type_id` fields with the datatype IDs for the current member being converted, and then call `H5T_convert_with_ctx` to avoid all the overhead.

test/dt_arith.c
```diff
@@ -511,20 +521,21 @@ test_hard_query(void)
         goto error;
     }
 
-    /* Unregister the hard conversion from int to float.  Verify the conversion
-     * is a soft conversion. */
-    H5Tunregister(H5T_PERS_HARD, NULL, H5T_NATIVE_INT, H5T_NATIVE_FLOAT,
-                  (H5T_conv_t)((void (*)(void))H5T__conv_int_float));
```
jhendersonHDF, line 517: This test was previously casting `H5T__conv_int_float` from a library conversion function into an application conversion function. Since the two have different prototypes, the changes in this PR starting tripping over this. This appears to have worked in the past since the difference between application conversion functions and library conversion functions is that application functions just have a DXPL parameter at the very end of the list.

tools/test/h5format_convert/h5fc_chk_idx.c
jhendersonHDF, line 1: The changes here just fix an issue when testing with `-fsanitize=address`. Not exactly a problematic memory leak since the program exits quickly, but still causes problems.

src/H5Dint.c
```diff
@@ -2729,9 +2729,8 @@ H5D__vlen_get_buf_size(H5D_t *dset, hid_t type_id, hid_t space_id, hsize_t *size
     vlen_bufsize.common.size = 0;
 
     /* Call H5S_select_iterate with args, etc. */
-    dset_op.op_type          = H5S_SEL_ITER_OP_APP;
-    dset_op.u.app_op.op      = H5D__vlen_get_buf_size_cb;
-    dset_op.u.app_op.type_id = type_id;
+    dset_op.op_type  = H5S_SEL_ITER_OP_LIB;
+    dset_op.u.lib_op = H5D__vlen_get_buf_size_cb;
```
jhendersonHDF, line 2733: It didn't look like there was a need for the callback here to be an application iter op, as opposed to a library one. If I left this as an application op, the passed in `type_id` would have to be unwrapped for every single element so that the `H5T_t *` can be assigned to `dset_info.mem_type` in the changes above to `H5D__vlen_get_buf_size_cb`. Switching this to a library iter op passes an `H5T_t *` to the callback which is much simpler.

src/H5T.c
```diff
@@ -2523,27 +2526,36 @@ H5T__register(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_con
...
+            if (NULL == (tmp_dtype = H5T_copy(old_path->dst, H5T_COPY_ALL)))
+                HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "unable to copy dst datatype");
+
             memset(&cdata, 0, sizeof cdata);
             cdata.command = H5T_CONV_INIT;
             if (conv->is_app) {
+                if ((tmp_sid = H5I_register(H5I_DATATYPE, tmp_stype, false)) < 0)
+                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTREGISTER, FAIL,
+                                "unable to register ID for source datatype");
+                if ((tmp_did = H5I_register(H5I_DATATYPE, tmp_dtype, false)) < 0)
```
jhendersonHDF, line 2540: Only register an ID for the types here if we have to pass them to an application conversion function

src/H5Dchunk.c
qkoziol: I don't think that those H5T_t's to H5T_convert can be const, since their state gets updated for certain conversions.  (I'm thinking vlen & ref conversions)

qkoziol: No problem removing const here...

src/H5Dint.c
qkoziol: Agree, good call

src/H5T.c
```diff
@@ -2523,27 +2526,36 @@ H5T__register(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_con
...
+                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTREGISTER, FAIL,
+                                "unable to register ID for destination datatype");
+
                 if ((conv->u.app_func)(tmp_sid, tmp_did, &cdata, (size_t)0, (size_t)0, (size_t)0, NULL, NULL,
                                        H5CX_get_dxpl()) < 0) {
                     H5I_dec_ref(tmp_sid);
```
qkoziol, line 2546: I know this is legacy code, but adding checks for error returns would be good throughout this section of code.

src/H5Dchunk.c
jhendersonHDF: At least from a quick scan it doesn't look like the direct contents of the pointed to structs get modified, but deeper parts of the pointed to struct, like the contents of the file pointer for vlen and ref types, might get modified. Though, depending on what's actually modified we could get away with it, as long as the direct contents of the pointed to H5T_t aren't modified. Would have to dig deeper to see if there would be problems, but could be interesting.

src/H5T.c
```diff
@@ -4993,28 +5043,53 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
...
-            HGOTO_ERROR(H5E_DATATYPE, H5E_CANTREGISTER, NULL,
-                        "unable to register destination conversion type for query");
+        if (path->src && (NULL == (tmp_stype = H5T_copy(path->src, H5T_COPY_ALL))))
+            HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, NULL, "unable to copy source datatype");
+        if (path->dst && (NULL == (tmp_dtype = H5T_copy(path->dst, H5T_COPY_ALL))))
+            HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, NULL, "unable to copy destination datatype");
+
         path->cdata.command = H5T_CONV_INIT;
         if (conv->is_app) {
+            if (tmp_stype && ((src_id = H5I_register(H5I_DATATYPE, tmp_stype, false)) < 0))
```
qkoziol, line 5053: How would tmp_stype or tmp_dtype be NULL here?  Shouldn't you always register src_id and dst_id?

src/H5Ofill.c
jhendersonHDF: I think there are also several more opportunities for removing type copying calls since a lot of them only appear to have been done so that when the ID they were wrapped in is closed, it doesn't free the original datatype. But I didn't want to go down that rabbit hole yet since it could get complicated, especially in the case where a copy of a vlen or reference type is made and the "location" of the copied type is then changed.

src/H5T.c
jhendersonHDF: Will do.

src/H5Tconv.c
```diff
@@ -2032,17 +2042,27 @@ H5T__conv_struct_init(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata)
...
 
-                type = H5T_copy(src->shared->u.compnd.memb[i].type, H5T_COPY_ALL);
-                tid  = H5I_register(H5I_DATATYPE, type, false);
-                assert(tid >= 0);
+                if (NULL == (type = H5T_copy(src->shared->u.compnd.memb[i].type, H5T_COPY_ALL)))
+                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL,
+                                "can't copy source compound member datatype");
+                priv->src_memb[i] = type;
+
+                if ((tid = H5I_register(H5I_DATATYPE, type, false)) < 0)
```
qkoziol, line 2053: Why do we need IDs for these types still?

src/H5T.c
jhendersonHDF: Based on the above code that was previously only copying the type and wrapping it in an ID if `path->src` and `path->dst` aren't NULL, I kept that same logic assuming that there's a case where the copying could get skipped. That said, if it made it into one of the following blocks, it would pass NULL or H5I_INVALID_HID depending on the type of conversion function, but after studying the code for a while I couldn't really determine when the case would be that `path->src` or `path->dst` are NULL, so I tried to leave the functionality the same.

src/H5Tconv.c
```diff
@@ -3024,23 +3066,21 @@ H5T__conv_enum_numeric(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata, size_t ne
...
-                HGOTO_ERROR(H5E_DATASET, H5E_UNSUPPORTED, FAIL,
+                HGOTO_ERROR(H5E_DATATYPE, H5E_UNSUPPORTED, FAIL,
                             "unable to convert between src and dest datatype");
             }
             else if (!H5T_path_noop(tpath)) {
-                if ((src_parent_id = H5I_register(H5I_DATATYPE, H5T_copy(src_parent, H5T_COPY_ALL), false)) <
-                    0)
-                    HGOTO_ERROR(H5E_DATASET, H5E_CANTREGISTER, FAIL,
-                                "unable to register types for conversion");
+                if (NULL == (tmp_type = H5T_copy(src_parent, H5T_COPY_ALL)))
```
qkoziol, line 3079: Why does this type need to be copied?   Same question for *tons* of other copies - I think the only reason 90% of them were originally copied was because they needed an ID and decrementing the ID would delete the type object.   Since we don't need the IDs, we shouldn't need the copies.

src/H5Tconv.c
jhendersonHDF: Still need them for passing to the lower members functions in case a conversion exception function is being used. I'd like to only conditionally allocate the arrays and create the IDs, but there currently isn't anything passed to the conversion function that could tell me at init time whether I need to. The new conversion context is only available at conversion time, not init time. I suppose we could add a field to the `H5T_cdata_t` structure, though since it's a public one I've tried to avoid that. It also wouldn't really be of interest to an application conversion callback either.

src/H5Tconv.c
jhendersonHDF: Agreed (and added a comment above after thinking for a bit). My goal would be to tackle that directly after this PR, I just didn't want to dive into that yet since it can be tricky and this is also already a fairly large set of changes.

qkoziol (approved):
Super!   Long overdue!

src/H5Ofill.c
qkoziol: Agree, I mentioned this in another comment, and I think there's good opportunity for more performance improvements from removing the unnecessary copy & free calls.

src/H5T.c
qkoziol: Wouldn't the IDs be INVALID then?

src/H5Tconv.c
qkoziol: Couldn't you call H5T_convert here, instead of H5T_convert_ctx, and let it sort things out?

qkoziol: Or add a helper routine (maybe in the _init() call) that determined which ones needed an ID and which could avoid making one?

src/H5T.c
jhendersonHDF: Yes. That part of the code probably needs to be studied more since previously AND currently, if it hit that particular spot it would pass invalid stuff to the conversion functions.

qkoziol: Sorry, re-read your earlier comment and you said it would pass INVALID. :-)

qkoziol: I think the earlier version of the code was buggy and you can assume that the tmp_stype & tmp_dtype are non-NULL.

src/H5Tconv.c
jhendersonHDF: The ctx version of convert was actually created for this specific reason since calling H5T_convert would cause an enormous amount of overhead as it would be creating and destroying IDs for every member of every compound element when a conversion exception function is used. That said, maybe I can provide the context object at init time with a field that lets the function know if it needs these IDs and then separate which fields of the context will be valid based on the conversion command type specified in the cdata.

qkoziol: Yes, that's the direction I was thinking.

src/H5T.c
```diff
@@ -1498,8 +1498,8 @@ H5T_top_term_package(void)
                     }                          /* end if */
                 }                              /* end if */
                 else {
-                    if ((path->conv.u.lib_func)((hid_t)FAIL, (hid_t)FAIL, &(path->cdata), (size_t)0,
-                                                (size_t)0, (size_t)0, NULL, NULL) < 0) {
+                    if ((path->conv.u.lib_func)(NULL, NULL, &(path->cdata), NULL, (size_t)0, (size_t)0,
```
derobins, line 1501: Don't need noise casts for small integers

derobins (approved):

src/H5T.c
jhendersonHDF: I may leave this like it is for now. That function is fairly complicated, but it looks like the datatypes in question could be NULL for the library's internal no-op function, as well as an application-registered soft conversion no-op function. If that's the case it would cause an assertion in the library when trying to copy a NULL type, but these functions seem to happily accept NULL or H5I_INVALID_HID for the datatypes as they're just ignored by the function. It seems like a contrived case but I can't easily predict what people may do. The library's no-op conversion function should never hit that part of the code as it is currently, but an application function might.

PR #4108, opened 2024-03-10
Move common variables out of cache test blocks

derobins:
This function has stack size warnings and needs some fixin'

### DIFF
--- a/test/cache.c
+++ b/test/cache.c
@@ -4145,6 +4145,16 @@ check_flush_cache__flush_ops(H5F_t *file_ptr)
     struct fo_flush_entry_check     *checks        = NULL;
     H5C_t                           *cache_ptr     = file_ptr->shared->cache;
 
+    /* Per-test variables. Each test sets these. */
+    int          test_num;
+    unsigned int flush_flags;
+    int          spec_size;
+    int          check_size;
+    unsigned     init_expected_index_len;
+    size_t       init_expected_index_size;
+    unsigned     expected_index_len;
+    size_t       expected_index_size;
+
     if (cache_ptr == NULL) {
 
         pass         = false;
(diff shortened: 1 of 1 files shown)


test/cache.c
```diff
@@ -4145,6 +4145,16 @@ check_flush_cache__flush_ops(H5F_t *file_ptr)
...
 
+    /* Per-test variables. Each test sets these. */
+    int          test_num;
+    unsigned int flush_flags;
+    int          spec_size;
+    int          check_size;
+    unsigned     init_expected_index_len;
+    size_t       init_expected_index_size;
+    unsigned     expected_index_len;
+    size_t       expected_index_size;
```
derobins, line 4156: There are so many tests w/ repeated block variables in this function, it generates a stack size warning w/ XCode's clang

derobins:
This still has stack size warnings w/ clang. I'm going to have to separate the giant function into separate tests :(

byrnHDF (approved):

jhendersonHDF:
> This still has stack size warnings w/ clang. I'm going to have to separate the giant function into separate tests :(

Are Clang and GCC using the same flags? I see the `stack-usage=8192` flag for GCC, but doesn't look like we set anything for Clang.

jhendersonHDF (approved):

qkoziol (approved):

derobins:
> > This still has stack size warnings w/ clang. I'm going to have to separate the giant function into separate tests :(
> 
> Are Clang and GCC using the same flags? I see the `stack-usage=8192` flag for GCC, but doesn't look like we set anything for Clang.

We should probably do a pass to see if they are in sync. I noticed it on MacOS.

PR #4109, opened 2024-03-10
Convert H5B__assert to use error checks

derobins:
Switches assert() calls to HGOTO_ERROR in H5B__assert() so it can be used in production mode. Also renames it to H5B__verify_structure() to better reflect what it checks.

### DIFF
--- a/src/H5Bdbg.c
+++ b/src/H5Bdbg.c
@@ -53,9 +53,7 @@ H5B_debug(H5F_t *f, haddr_t addr, FILE *stream, int indent, int fwidth, const H5
 
     FUNC_ENTER_NOAPI(FAIL)
 
-    /*
-     * Check arguments.
-     */
+    /* Check arguments */
     assert(f);
     assert(H5_addr_defined(addr));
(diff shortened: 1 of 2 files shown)


src/H5Bdbg.c
```diff
@@ -163,91 +156,106 @@ H5B__assert(H5F_t *f, haddr_t addr, const H5B_class_t *type, void *udata)
...
     bt = NULL; /* Make certain future references will be caught */
 
-    /*
-     * Do a breadth-first search of the tree.  New nodes are added to the end
+    /* Do a breadth-first search of the tree.  New nodes are added to the end
      * of the queue as the `cur' pointer is advanced toward the end.  We don't
      * remove any nodes from the queue because we need them in the uniqueness
      * test.
      */
-    for (ncell = 0; cur; ncell++) {
```
derobins, line 193: clang complains that ncell isn't used

byrnHDF (approved):

qkoziol (approved):

PR #4113, opened 2024-03-11
Implement ID creation optimization for container datatype conversions

jhendersonHDF:
Makes the datatype conversion context object available during both the initialization and conversion processes for a datatype conversion function, allowing the compound, variable-length and array datatype conversion functions to avoid creating IDs for the datatypes when they aren't necessary

Adds internal H5CX_pushed routine to determine if an API context is available to retrieve values from

Also adds error checking to several places in H5T.c and H5Tconv.c where the code had previously assumed object close operations would succeed

### DIFF
--- a/src/H5CX.c
+++ b/src/H5CX.c
@@ -760,6 +760,28 @@ H5CX__get_context(void)
 #endif /* H5_HAVE_THREADSAFE */
 
 /*-------------------------------------------------------------------------
+ * Function:    H5CX_pushed
+ *
+ * Purpose:     Returns whether or not an API context has been pushed.
+ *
+ * Return:      true/false
+ *
+ *-------------------------------------------------------------------------
+ */
+bool
+H5CX_pushed(void)
+{
+    H5CX_node_t **head = NULL; /* Pointer to head of API context list */
+
+    FUNC_ENTER_NOAPI_NOERR
+
+    head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
+    assert(head);
+
+    FUNC_LEAVE_NOAPI(*head != NULL);
+}
+
+/*-------------------------------------------------------------------------
  * Function:    H5CX__push_common
  *
  * Purpose:     Internal routine to push a context for an API call.
--- a/src/H5Tconv.c
+++ b/src/H5Tconv.c
@@ -185,9 +185,9 @@
 #define H5T_CONV_Xx_CORE(STYPE, DTYPE, S, D, ST, DT, D_MIN, D_MAX)                                           \
     {                                                                                                        \
         if (*(S) > (ST)(D_MAX)) {                                                                            \
-            H5T_conv_ret_t except_ret =                                                                      \
-                (conv_ctx->cb_struct.func)(H5T_CONV_EXCEPT_RANGE_HI, conv_ctx->src_type_id,                  \
-                                           conv_ctx->dst_type_id, S, D, conv_ctx->cb_struct.user_data);      \
+            H5T_conv_ret_t except_ret = (conv_ctx->u.conv.cb_struct.func)(                                   \
+                H5T_CONV_EXCEPT_RANGE_HI, conv_ctx->u.conv.src_type_id, conv_ctx->u.conv.dst_type_id, S, D,  \
+                conv_ctx->u.conv.cb_struct.user_data);                                                       \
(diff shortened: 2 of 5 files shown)


src/H5CX.c
```diff
@@ -759,6 +759,28 @@ H5CX__get_context(void)
...
+ * Function:    H5CX_pushed
+ *
+ * Purpose:     Returns whether or not an API context has been pushed.
+ *
+ * Return:      true/false
+ *
+ *-------------------------------------------------------------------------
+ */
+bool
+H5CX_pushed(void)
```
jhendersonHDF, line 772: This is a bit of a hack, but see the changes in H5T.c for why this is needed.

src/H5Tpkg.h
```diff
@@ -152,17 +152,35 @@ struct H5T_stats_t {
 
 /* Context struct for information used during datatype conversions */
 typedef struct H5T_conv_ctx_t {
-    H5T_conv_cb_t cb_struct;
-    hid_t         dxpl_id;
-    hid_t         src_type_id;
-    hid_t         dst_type_id;
+    union {
```
jhendersonHDF, line 155: Most of the changes in this PR related to turning this struct's fields into a union so we can distinguish which fields are valid to access depending on which conversion command is being performed.

jhendersonHDF:
This PR is a partial follow-up on the changes in https://github.com/HDFGroup/hdf5/pull/4104. The other major part of that PR to address is the removal of many `H5T_copy` calls now that they likely aren't needed, but I don't believe that can be safely done without adding `const` to `H5T_convert`'s `src` and `dst` parameters. I think this can reasonably be done since a conversion function shouldn't be modifying the original datatypes anyway and if they really need to, they can make a temporary internal copy of the datatype (like the variable-length and array conversion functions are already doing).

src/H5Tconv.c
```diff
@@ -2005,13 +2015,31 @@ H5T__conv_struct_init(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata)
...
+            HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL,
+                        "couldn't allocate source to destination member mapping array");
+        if (NULL == (priv->src_memb = (H5T_t **)H5MM_malloc(src_nmembs * sizeof(H5T_t *))))
+            HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL,
+                        "couldn't allocate source compound member datatype array");
+        if (NULL == (priv->dst_memb = (H5T_t **)H5MM_malloc(dst_nmembs * sizeof(H5T_t *))))
+            HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL,
+                        "couldn't allocate destination compound member datatype array");
+
+        priv->need_ids = (cdata->command == H5T_CONV_INIT && conv_ctx->u.init.cb_struct.func) ||
```
qkoziol, line 2030: When is this false?

jhendersonHDF: This will be false when `cb_struct.func` is NULL. That's the case either during library initialization of the H5T package or when a conversion exception function hasn't been set and isn't being used.

qkoziol (approved):

src/H5Tconv.c
qkoziol: Ah, missed the dependency on the function pointer being NULL.

qkoziol:
Nice!  Now the extra copies?

jhendersonHDF:
> Nice! Now the extra copies?

Yes, looking at this next after fixing an inefficiency with compounds that have array members due to the recursive part of that causing the array conversion function to register IDs repeatedly. Variable-length members of compounds have the same issue, but it doesn't even show up in profiling because the variable-length callback operations like `is_null`, `write` and `read` overwhelm the profile anyway (I'll still try to fix the ID registration though).

To fix the overhead of the datatype copies, I think `src` and `dst` in `H5T_convert` need to become `const` to safely do this so I can assume any datatype passed in is not going to be changed. The variable-length conversion code is already making internal temporary copies of datatypes it might modify, so seems like marking the originals as `const` should be fine. I just don't want to, for example, pass an attribute's datatype to `H5T_convert` directly and be surprised later on when someone's new conversion function modifies the attribute's datatype unknowingly. If that new conversion needs to make changes it can just copy the type temporarily during init.

derobins (approved):

PR #4114, opened 2024-03-11
Fix up dsets test for some platforms with different long double format

jhendersonHDF:


jhendersonHDF:
There are apparently still some issues with this when cross-compiling on ARM, so we may have to look into that more.

qkoziol (approved):

byrnHDF (approved):

PR #4116, opened 2024-03-11
Handle IBM long double issues in dsets.c test_floattypes test

jhendersonHDF:


jhendersonHDF:
Tested on Echidna and a couple other machines, but we should test elsewhere. Rather than trying to write to a dataset with a custom floating-point type that may have less precision than the native long double type, check to see how many bits of the native type are used and if it's less than 128, create a custom 128-bit floating-point type where the mantissa takes up the extra bits so we don't lose precision between write and read. Otherwise, just write and read using the native long double type rather than trying to guess and promote to a wider type.

test/dsets.c
```diff
@@ -5891,41 +5891,35 @@ test_floattypes(hid_t file)
...
             goto error;
-        if (old_precision < precision) {
+
+        /* Get the layout of the native long double type */
+        if (H5Tget_fields(datatype, &ld_spos, &ld_epos, &ld_esize, &ld_mpos, &ld_msize) < 0)
+            goto error;
+
+        /* Check if all "tgt_precision"+ bits are already used. If not, define
+         * a custom floating-point type where the mantissa takes up the extra
+         * extra bits. Otherwise, just write and read using the native long
```
hyoklee: `the extra extra bits` -> `the extra bits`?

lrknox (approved):
Test passes on echidna.

byrnHDF (approved):

PR #4119, opened 2024-03-12
Use AC_SYS_LARGEFILE on Autotools

derobins:
We previously used a hack introduced in 1.8.5 to paper over differences in off_t and off64_t when determining the type sizes. We no longer explicitly support off64_t in the library and AC_SYS_LARGEFILE works fine.


configure.ac
```diff
@@ -1460,26 +1460,9 @@ if test "X${enable_shared}" = "Xyes"; then
...
-  [Number of bits in a file offset, on hosts where this is settable.],
-  [_AC_SYS_LARGEFILE_TEST_INCLUDES])
-
-## Check for _LARGE_FILES
-if test "$ac_cv_sys_file_offset_bits" = unknown; then
-  _AC_SYS_LARGEFILE_MACRO_VALUE([_LARGE_FILES], [1],
-    [ac_cv_sys_large_files],
-    [Define for large files, on AIX-style hosts.],
-    [_AC_SYS_LARGEFILE_TEST_INCLUDES])
-fi
```
derobins, line 1482: We only use off_t in the library, so this hack is no longer necessary. We just need to get the size of off_t so H5_SIZEOF_OFF_T is set correctly and AC_SYS_LARGEFILE is fine for that.

byrnHDF (approved):

fortnern (approved):
I'm not an expert on this, but as long as it works on all supported platforms this simplification looks good.

Issue #1917, opened 2022-07-21
H5Gget_objtype_by_idx returns the linked object type instead of H5G_LINK


wenjuno:
I tested it with HDF5 1.10.7, 1.12.1 and 1.13.1. With 1.10.6 the function returns H5G_LINK for soft link, hard link and external link but with 1.12.1 and 1.13.1 it returns the type of the linked object (e.g. H5G_DATASET if the link points to a dataset). Is it an intentional change or a regression? Thanks.

cronhan:
I also ran into this now after upgrading from 1.10 -> 1.12, albeit "indirectly" via `H5Location::getObjTypeByIdx` which now throws an exception for dangling links instead of just returning the type as it should. Not sure how to work around the problem :(

bmribler:
@cronhan @wenjuno Do you happen to have the minimal repro code?

wenjuno:
Hi @bmribler, I don't remember where I saved my code. I think I just made a file with different link types and wrote a simple program to print the output from the API.

bmribler:
@wenjuno @cronhan 

I'm sorry to let you know that we've decided not to fix this issue.

Returning H5G_LINK was an incorrect behavior of H5Gget_objtype_by_idx().  The function was to return the "object type" and H5G_link was not a type of an object.  That was the reason that H5Gget_objtype_by_idx() was changed in the first place.  The incorrect behavior will not be put back into this deprecated function.  Applications, which use H5Lget_info_by_idx() to get link type, will need to be modified to use [H5Lget_info_by_idx()](https://support.hdfgroup.org/releases/hdf5/v1_14/v1_14_5/documentation/doxygen/group___h5_l.html#ga4db00b8b944eae68233438165c784b67).  Please also see the note in [H5Gget_objtype_by_idx()](https://support.hdfgroup.org/releases/hdf5/v1_14/v1_14_5/documentation/doxygen/group___h5_g.html#gab1383b8cca3fa99410ad36427059c5a7).

wenjuno:
Thanks for the update.

PR #4120, opened 2024-03-12
Add clarification for current behavior of H5Get_objtype_by_idx()

glennsong09:
It seems to be an unintentional change, but the current behavior is not documented anywhere. 

Addresses #1917.

### DIFF
--- a/src/H5Gpublic.h
+++ b/src/H5Gpublic.h
@@ -1197,12 +1197,12 @@ H5_DLL ssize_t H5Gget_objname_by_idx(hid_t loc_id, hsize_t idx, char *name, size
  *          \p idx is the transient index used to iterate through the objects in
  *          the group. This parameter is described in more detail in the
  *          discussion of H5Gget_objname_by_idx().
- * 
+ *
  * \note    As of 1.12.0, H5Get_objtype_by_idx() returns the type of the object
  *          that the link points to, but it has been deprecated for H5Oget_info().
- *          Previous behavior for this function returned H5G_LINK for any link type. 
- *          To get the link type, an application may use H5Lget_info_by_idx() 
- *          instead. 
+ *          Previous behavior for this function returned H5G_LINK for any link type.
+ *          To get the link type, an application may use H5Lget_info_by_idx()
+ *          instead.
  *
  * \version 1.8.0 Function deprecated in this release.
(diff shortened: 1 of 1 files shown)


bmribler (approved):

derobins:
Documenting the change in behavior is great, but we shouldn't close the issue until we understand why the change occurred and whether we're happy with the current behavior.

jhendersonHDF:
> Documenting the change in behavior is great, but we shouldn't close the issue until we understand why the change occurred and whether we're happy with the current behavior.

Bisected to https://github.com/HDFGroup/hdf5/commit/ca1b9025fe8be1eab6bd76d58ab78bb9450c41f1

glennsong09:
> Documenting the change in behavior is great, but we shouldn't close the issue until we understand why the change occurred and whether we're happy with the current behavior.

If I delete the line in my PR, does it stop it from automatically closing the issue?

jhendersonHDF:
> > Documenting the change in behavior is great, but we shouldn't close the issue until we understand why the change occurred and whether we're happy with the current behavior.
> 
> If I delete the line in my PR, does it stop it from automatically closing the issue?

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

"Addresses" isn't one of GitHub's keywords, so this PR shouldn't close the issue when it's merged. You can also look in the Development section in the bottom right to see if the issue is linked to the PR.

src/H5Gpublic.h
```diff
@@ -1198,6 +1198,12 @@ H5_DLL ssize_t H5Gget_objname_by_idx(hid_t loc_id, hsize_t idx, char *name, size
  *          the group. This parameter is described in more detail in the
  *          discussion of H5Gget_objname_by_idx().
  *
+ * \note    As of 1.12.0, H5Get_objtype_by_idx() returns the type of the object
```
mattjala: Typo: `H5Get_objtype_by_idx` -> `H5Gget_objtype_by_idx`

mattjala (changes requested):
Looks good besides the typo

src/H5Gpublic.h
glennsong09: Fixed.

jhendersonHDF (approved):

mattjala (approved):

PR #4130, opened 2024-03-13
Fix segfault in vlen io API test

mattjala:
- Fixed an issue where skipping the test could lead to uninitialized memory in `rbuf`/`wbuf` being freed
- Fixed using `H5Gopen` macro instead of `H5Gopen2`
- Added missing sub-test headers for selections
- If the test fails, only attempt to close datasets if they have valid handles

This should fix the Cache VOL's segfault in the daily CI.

### DIFF
--- a/test/API/H5_api_dataset_test.c
+++ b/test/API/H5_api_dataset_test.c
@@ -9853,6 +9853,9 @@ test_dataset_vlen_io(void)
     hvl_t wbuf[DATASET_VLEN_IO_DSET_DIMS];
     hvl_t rbuf[DATASET_VLEN_IO_DSET_DIMS];
 
+    memset(wbuf, 0, sizeof(hvl_t) * DATASET_VLEN_IO_DSET_DIMS);
+    memset(rbuf, 0, sizeof(hvl_t) * DATASET_VLEN_IO_DSET_DIMS);
+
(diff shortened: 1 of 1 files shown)


test/API/H5_api_dataset_test.c
```diff
@@ -10453,9 +10459,12 @@ test_dataset_vlen_io(void)
...
         H5Gclose(container_group);
-        H5Dclose(dset_int);
-        H5Dclose(dset_float);
-        H5Dclose(dset_string);
+        if (dset_int != H5I_INVALID_HID)
+            H5Dclose(dset_int);
+        if (dset_float != H5I_INVALID_HID)
+            H5Dclose(dset_float);
+        if (dset_string != H5I_INVALID_HID)
+            H5Dclose(dset_string);
```
jhendersonHDF, line 10467: Since these IDs were initialized to `H5I_INVALID_HID` and the purpose of the BEGIN_TRY and END_TRY blocks is to hide any errors while we try to close anything not yet closed while possibly closing stuff that IS closed, you shouldn't need to check for `H5I_INVALID_HID` here. That said, it doesn't hurt anything.

jhendersonHDF (approved):

qkoziol (approved):

PR #4135, opened 2024-03-14
Skip vlen IO API test for cache VOL

mattjala:
The segfault occurred during the closing of the float/string datasets. The test now calls `H5Dflush` and `H5Dclose` in each subtest to make sure the vlen buffers are flushed to disk before they are freed.

### DIFF
--- a/test/API/H5_api_dataset_test.c
+++ b/test/API/H5_api_dataset_test.c
@@ -9861,13 +9861,16 @@ test_dataset_vlen_io(void)
 
     /* Make sure the connector supports the API functions being tested */
     if (!(vol_cap_flags_g & H5VL_CAP_FLAG_FILE_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_GROUP_BASIC) ||
-        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE) ||
-        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE)) {
+        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE)) {
         SKIPPED();
         printf("    API functions for basic file, group, or dataset aren't supported with this "
                "connector\n");
         return 0;
     }
+
+    /* Skipped for now due to segfault with the Cache VOL */
+    SKIPPED();
+    return 0;
 
     TESTING_2("test setup");
 
@@ -10511,10 +10514,6 @@ test_dataset_set_extent_chunked_unlimited(void)
     hid_t   fspace_id  = H5I_INVALID_HID;
 
     TESTING("H5Dset_extent on chunked dataset with unlimited dimensions");
-
-    /* Skipped for now due to segfault with the Cache VOL */
-    SKIPPED();
-    return 0;
 
     /* Make sure the connector supports the API functions being tested */
     if (!(vol_cap_flags_g & H5VL_CAP_FLAG_FILE_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_GROUP_BASIC) ||
--- a/test/API/H5_api_dataset_test.c
+++ b/test/API/H5_api_dataset_test.c
@@ -10443,23 +10443,6 @@ test_dataset_vlen_io(void)
 
     if (H5Sclose(space_id) < 0)
         TEST_ERROR;
-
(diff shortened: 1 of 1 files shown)


test/API/H5_api_dataset_test.c
```diff
@@ -10251,6 +10265,9 @@ test_dataset_vlen_io(void)
             if ((H5Dwrite(dset_int, vlen_int, space_id, space_id, H5P_DEFAULT, (const void *)wbuf)) < 0)
                 PART_TEST_ERROR(rw_point_selection);
 
+            if (H5Dflush(dset_int) < 0)
```
jhendersonHDF, line 10271: Hmm I'm surprised adding a flush here would fix the problem since I'd assume that would just move the segfault to here instead. Plus, `H5Dflush` is mostly used for SWMR and I wouldn't think it would be needed here. Since the native VOL was passing without this call, I wonder if this is either a problem in the cache VOL or a problem in the VOL interface instead.

mattjala: The segfault seems to be due to the cache vol still using the dynamically allocated heap buffers when they're freed after the reads/writes. As it turns out, this doesn't resolve the issue either - it cropped up again in the CI [here](https://github.com/mattjala/hdf5/actions/runs/8295314503). The cache VOL docs say that H5Dwrite/H5Dclose should be semi-blocking, so this may be an issue on the cache VOL.

qkoziol:
Agree - seems likely to be in the cache connector.

mattjala:
I've set this test to be skipped for now so that CI works. Ideally we'd be able to use the VOL connector capability flags to do this, but 1. the Cache VOL directly inherits the native VOL's capability flags, and 2. There's no capability flag corresponding to variable length type support - the closest one is for stored datatypes. I've opened hpc-io/vol-cache#26 on the cache VOL to document this, though the odds that it gets fixed in the near future are low.

test/API/H5_api_dataset_test.c
```diff
@@ -9861,7 +9861,8 @@ test_dataset_vlen_io(void)
 
     /* Make sure the connector supports the API functions being tested */
     if (!(vol_cap_flags_g & H5VL_CAP_FLAG_FILE_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_GROUP_BASIC) ||
-        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE)) {
+        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE) ||
+        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE)) {
```
jhendersonHDF: Duplicate check for `H5VL_CAP_FLAG_DATASET_MORE` flag here

test/API/H5_api_dataset_test.c
```diff
@@ -10511,9 +10512,14 @@ test_dataset_set_extent_chunked_unlimited(void)
...
 
+    /* Skipped for now due to segfault with the Cache VOL */
+    SKIPPED();
+    return 0;
+
     /* Make sure the connector supports the API functions being tested */
     if (!(vol_cap_flags_g & H5VL_CAP_FLAG_FILE_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_GROUP_BASIC) ||
-        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE)) {
+        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_BASIC) || !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE) ||
+        !(vol_cap_flags_g & H5VL_CAP_FLAG_DATASET_MORE)) {
```
jhendersonHDF, line 10521: Duplicate check for `H5VL_CAP_FLAG_DATASET_MORE` flag here

test/API/H5_api_dataset_test.c
```diff
@@ -10511,9 +10512,14 @@ test_dataset_set_extent_chunked_unlimited(void)
 
     TESTING("H5Dset_extent on chunked dataset with unlimited dimensions");
 
+    /* Skipped for now due to segfault with the Cache VOL */
+    SKIPPED();
```
jhendersonHDF: Looks like this SKIPPED is in the wrong place

bmribler (approved):
The flushes/closing steps are reasonable even we're skipping the test.

jhendersonHDF:
> The flushes/closing steps are reasonable even we're skipping the test.

Agreed. They aren't really hurting anything.

jhendersonHDF (approved):

PR #4049, opened 2024-02-26
Added new H5E with tests.

brtnfld:
- Added Fortran H5E APIs:
      h5eregister_class_f, h5eunregister_class_f, h5ecreate_msg_f, h5eclose_msg_f
      h5eget_msg_f, h5epush_f, h5eget_num_f, h5ewalk_f, h5eget_class_name_f,
      h5eappend_stack_f, h5eget_current_stack_f, h5eset_current_stack_f, h5ecreate_stack_f,
      h5eclose_stack_f, h5epop_f, h5eprint_f (C h5eprint v2 signature)

-Addresses Issue #3987

PR #4138, opened 2024-03-14
Fixed misc. H5E fortran failures due to previous PR

brtnfld:
fixed promotion of integers and reals tests and check-passthrough-vol failure, introduced by PR #4049


derobins (approved):

lrknox (approved):
This may only fix one of the errors

PR #4139, opened 2024-03-14
Propagate group creation properties to intermediate groups

### LINKED ISSUE #3945: Group creation properties are not propagated to the intermediate groups
**Describe the bug**
When creating the group with link property set to create the intermediate groups, one would assume that the group creation properties would be propagated to all the created groups, but it is not so. This is the minimal reproduction example.

```c
#include <assert.h>
#include "hdf5.h"

int main(void)
{
        hid_t linkProps, groupProps, file, group;
        unsigned crOrder;
 
        linkProps = H5Pcreate(H5P_LINK_CREATE);
        groupProps = H5Pcreate(H5P_GROUP_CREATE);
        
        H5Pset_create_intermediate_group(linkProps, 1);
        H5Pset_attr_creation_order(groupProps, H5P_CRT_ORDER_TRACKED);
        
        file = H5Fcreate("test.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
        
        H5Gcreate2(file, "test/test", linkProps, groupProps, H5P_DEFAULT);
        
        group = H5Gopen2(file, "test", H5P_DEFAULT);
        groupProps = H5Gget_create_plist(group);
        H5Pget_attr_creation_order(groupProps, &crOrder); 
        
        /* this assert will fail */
        assert(crOrder == H5P_CRT_ORDER_TRACKED);

        return 0;
}
```

**Expected behavior**
The intermadiate groups being created with the same creation properties.

vchoi-hdfgroup:
Propagate object creation properties to intermediate groups.

Fixes #3945


test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    hid_t    dcpl        = H5I_INVALID_HID;
+    hid_t    fid         = H5I_INVALID_HID;
+    hid_t    gid         = H5I_INVALID_HID;
+    hid_t    gid2        = H5I_INVALID_HID;
+    hid_t    def_gid     = H5I_INVALID_HID;
+    hid_t    did         = H5I_INVALID_HID;
+    hid_t    sid         = H5I_INVALID_HID;
+    hsize_t  dims[1]     = {10};
+    unsigned cr_order    = 0;
+    hbool_t  track_times = false;
```
derobins: Use bool instead of hbool_t

src/H5Gtraverse.c
```diff
@@ -666,8 +669,14 @@ H5G__traverse_real(const H5G_loc_t *_loc, const char *name, unsigned target, H5G
...
 
                 /* Create the intermediate group */
-                /* XXX: Should we allow user to control the group creation params here? -QAK */
-                gcrt_info.gcpl_id    = H5P_GROUP_CREATE_DEFAULT;
+                gcrt_info.gcpl_id = H5P_GROUP_CREATE_DEFAULT;
+                if ((ocrt_info = ((H5L_trav_cr_t *)op_data)->ocrt_info) != NULL) {
+                    if (ocrt_info->obj_type == H5O_TYPE_GROUP)
+                        gcrt_info.gcpl_id = ((H5G_obj_create_t *)ocrt_info->crt_info)->gcpl_id;
+                    else if (ocrt_info->obj_type == H5O_TYPE_DATASET)
+                        gcrt_info.gcpl_id = ((H5D_obj_create_t *)ocrt_info->crt_info)->dcpl_id;
```
qkoziol: This seems like a poor reason to include the H5Dpkg header.  Please add a private 'getter' routine to the H5D interface instead.

qkoziol: Actually, it looks like there should be 3 new private routines: a 'getter' in the H5L interface (so the H5L_trav_cr_t struct stays hidden from other packages), and a 'getter' for the H5G and H5D interfaces that retrieve the GCPL & DCPL IDs, respectively.

qkoziol (changes requested):
Adding 'getter' routines will avoid coupling these packages together.

src/H5Gtraverse.c
vchoi-hdfgroup: I implemented the 3 getters for H5L, H5G, and H5D interfaces.
However, I got the following CI build errors: 
 /usr/bin/ld: ../src/.libs/libhdf5.so: undefined reference to `H5D_OBJ_ID'

I couldn't figure out what went wrong for the macro H5D_OBJ_ID in the H5D interface.  Anyone knows what might be the problem?

qkoziol (approved):

src/H5Gtraverse.c
qkoziol: Not certain about this, since H5Gtraverse included H5Dprivate.h. Maybe delete your build directory and re-run autogen.sh?

vchoi-hdfgroup: The failures are from the CI actions. Not from my own build on jelly.

qkoziol (approved):
Nice!

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    fid = H5Fcreate(MISC40_FILE, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+    CHECK(fid, FAIL, "H5Fcreate");
+
+    /*
+     * Verify default creation properties when creating
+     * group with intermdiate group
+     */
+
+    /* Create groups that use default creation properties */
+    def_gid = H5Gcreate2(fid, "def_group1/def_group2", lcpl, H5P_DEFAULT, H5P_DEFAULT);
```
bmribler, line 6687: Is it better to use constants for these names?

src/H5Gtraverse.c
```diff
@@ -665,8 +667,15 @@ H5G__traverse_real(const H5G_loc_t *_loc, const char *name, unsigned target, H5G
...
                 /* Create the intermediate group */
-                /* XXX: Should we allow user to control the group creation params here? -QAK */
-                gcrt_info.gcpl_id    = H5P_GROUP_CREATE_DEFAULT;
+                gcrt_info.gcpl_id = H5P_GROUP_CREATE_DEFAULT;
+                /* Propagate the object creation properties when creating intermedidate groups */
+                if ((target & H5G_CRT_OBJ) && (ocrt_info = H5L_OCRT_INFO(op_data)) != NULL) {
+                    if (ocrt_info->obj_type == H5O_TYPE_GROUP)
+                        gcrt_info.gcpl_id = H5G_OBJ_ID(ocrt_info->crt_info);
+                    else if (ocrt_info->obj_type == H5O_TYPE_DATASET)
+                        gcrt_info.gcpl_id = H5D_OBJ_ID(ocrt_info->crt_info);
```
jhendersonHDF, line 676: Seems like this may be problematic to assign a DCPL ID to a GCPL ID. They're both subclasses of H5P_CLS_OBJECT_CREATE_g, but if the code ever needs to retrieve some group-specific properties (maybe in the future), it seems like that may fail. Should `gcrt_info` have an `ocpl_id` field rather than a `gcpl_id` field to try to make it clear that only generic OCPL properties should be retrieved from it?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+
+    def_gcpl = H5Gget_create_plist(def_gid);
+    CHECK(def_gcpl, FAIL, "H5Gget_create_plist");
+
+    /* Default is true */
+    status = H5Pget_obj_track_times(def_gcpl, &track_times);
+    CHECK(status, FAIL, "H5Gclose");
+    VERIFY(track_times, true, "H5Pget_obj_track_times");
+
+    /* Default is false */
```
bmribler, line 6704: Default is 0?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    status = H5Pget_attr_creation_order(def_gcpl, &cr_order);
+    CHECK(status, FAIL, "H5Gclose");
+    VERIFY(cr_order, 0, "H5Pget_attr_creation_order");
+
+    status = H5Gclose(def_gid);
+    CHECK(status, FAIL, "H5Gclose");
+
+    status = H5Pclose(def_gcpl);
+    CHECK(status, FAIL, "H5Gclose");
```
bmribler: H5Pclose?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+
+    /*
+     * Verify non-default creation properties when creating
+     * group with intermediate groups
+     */
+    gcpl = H5Pcreate(H5P_GROUP_CREATE);
+    CHECK(gcpl, FAIL, "H5Fcreate");
```
bmribler: H5Pcreate?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    status = H5Pget_attr_creation_order(gcpl, &cr_order);
+    CHECK(status, FAIL, "H5Pget_attr_creation_order");
+    VERIFY(cr_order, H5P_CRT_ORDER_TRACKED, "H5Pget_attr_creation_order");
+
+    status = H5Pget_obj_track_times(gcpl, &track_times);
+    CHECK(status, FAIL, "H5Pget_obj_track_times");
+    VERIFY(track_times, false, "H5Pget_obj_track_times");
+
+    status = H5Gclose(gid);
+    CHECK(status, FAIL, "H5Pclose");
```
bmribler: H5Gclose?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    status = H5Pclose(gcpl);
+    CHECK(status, FAIL, "H5Pclose");
+
+    /*
+     * Verify non-default creation properties when creating
+     * dataset with intermediate group
+     */
+    dcpl = H5Pcreate(H5P_DATASET_CREATE);
+    CHECK(dcpl, FAIL, "H5Fcreate");
```
bmribler: H5Pcreate?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    CHECK(status, FAIL, "H5Pget_attr_creation_order");
+    VERIFY(cr_order, H5P_CRT_ORDER_TRACKED, "H5Pget_attr_creation_order");
+
+    status = H5Pget_obj_track_times(gcpl, &track_times);
+    CHECK(status, FAIL, "H5Pget_obj_track_times");
+    VERIFY(track_times, true, "H5Pget_obj_track_times");
+
+    /* Verify dname */
+    did = H5Dopen2(gid, "dname", H5P_DEFAULT);
+    CHECK(did, FAIL, "H5Gopen2");
```
bmribler: H5Dopen2?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+
+    status = H5Pget_attr_creation_order(dcpl, &cr_order);
+    VERIFY(cr_order, H5P_CRT_ORDER_TRACKED, "H5Pget_attr_creation_order");
+
+    status = H5Pget_obj_track_times(dcpl, &track_times);
+    CHECK(status, FAIL, "H5Pget_obj_track_times");
+    VERIFY(track_times, true, "H5Pget_obj_track_times");
+
+    status = H5Dclose(did);
+    CHECK(status, FAIL, "H5Pclose");
```
bmribler: H5Dclose?

test/tmisc.c
```diff
@@ -6640,6 +6641,223 @@ test_misc39(void)
...
+    status = H5Pclose(dcpl);
+    CHECK(status, FAIL, "H5Pclose");
+
+    status = H5Gclose(gid);
+    CHECK(status, FAIL, "H5Pclose");
```
bmribler: H5Gclose

bmribler (commented):
Minor typos

src/H5Gtraverse.c
qkoziol: > Seems like this may be problematic to assign a DCPL ID to a GCPL ID. They're both subclasses of H5P_CLS_OBJECT_CREATE_g, but if the code ever needs to retrieve some group-specific properties (maybe in the future), it seems like that may fail. Should `gcrt_info` have an `ocpl_id` field rather than a `gcpl_id` field to try to make it clear that only generic OCPL properties should be retrieved from it?

Agree

vchoi-hdfgroup: I have couple questions about changing gcpl_id to ocpl_id:
--Changing gcpl_id field in gcrt_info to ocpl_id would lead to changes in quite a few files which uses gcrt_info.gpcl_id.
 --If I change gcpl_id in gcrt_info which is H5G_obj_create_t struct, should I also change dcpl_id field in H5D_obj_create_t struct to ocpl_id?

jhendersonHDF: > --Changing gcpl_id field in gcrt_info to ocpl_id would lead to changes in quite a few files which uses gcrt_info.gpcl_id.

It's probably fine to put this off to a separate focused PR since the code appears to currently only retrieve generic properties.

> --If I change gcpl_id in gcrt_info which is H5G_obj_create_t struct, should I also change dcpl_id field in H5D_obj_create_t struct to ocpl_id?

As far as I can tell, it looks like the only place that sets up a `H5D_obj_create_t` struct sets its `dcpl_id` field to a DCPL, so that one shouldn't need to be changed.

jhendersonHDF (approved):

PR #4140, opened 2024-03-14
Fix error when overwriting an indirectly nested vlen with a shorter sequence

fortnern:
The code for freeing nested vlens didn't take into account the possibility that the nested vlen might be inside a compound or array, and inappropriately assumed the buffer pointed directly to a serialized vlen. Modified to handle compounds and arrays correctly.

The library still does not free all space used by triply nested vlens when overwriting (even when not shrinking), nor for references nested in vlens, and has these same problems when shrinking datasets. Chunks that are entirely deleted (due to shrinking or the dataset being deleted) I believe do not have vlen or reference data freed at all. None of these cases should cause errors though, just wasted space.


derobins:
I fixed the easy 1.6 compatibility issues, but the code raises warnings which trip the -Werror checks and those need to be fixed

fortnern:
Checks passed, not sure why they're not showing up on this page

jhendersonHDF (approved):

jhendersonHDF:
I can see the checks passed in your branch. I'm guessing the clang-format action caused the checks to not show up here.

qkoziol (approved):

Issue #3872, opened 2023-11-27
Building error on HDF5 1.12.3 - configure with autoconf 2.71 fails with Cygwin


lrknox:
HDF Forum post from marco.atzeri:

Building the cygwin package for 1.12.3 I found 2 surprises that caused me puzzling build messages and errors:

Comparing with 1.12.2 I found

— hdf5-1.12.2-1.x86_64/origsrc/hdf5-1.12.2/configure.ac 2022-04-19 18:44:03.000000000 +0200
+++ hdf5-1.12.3-1.x86_64/origsrc/hdf5-1.12.3/configure.ac 2023-11-02 04:43:13.000000000 +
…

-AC_PREREQ([2.69])
+AC_PREREQ([2.71])

the change for AC is wrong as 2.71 complains a lot about obsolete declarations and 2.69 works fine. Reverting to 2.69 cut all the misleading Autoconf messages.

…

+## Strip out -Werror from CFLAGS since that can cause checks to fail when
+## compiling the test program fails due to warnings
+CFLAGS=“echo $CFLAGS | sed -e 's/-Werror//g'”
+CXXFLAGS=“echo $CXXFLAGS | sed -e 's/-Werror//g'”
+FCFLAGS=“echo $FCFLAGS | sed -e 's/-Werror//g'”
+JAVACFLAGS=“echo $JAVACFLAGS | sed -e 's/-Werror//g'”
+CPPFLAGS=“echo $CPPFLAGS | sed -e 's/-Werror//g'”

This implementation is error prone.On Cygwin build system the variables are crippled

CFLAGS = -ggdb -O2 -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 …

and stripping “-Werror” leave “=format-security” alone causing configure to estimate
that the C compiler fails to compile.

**Platform (please complete the following information)**
 - HDF5 version 1.12.3
 - OS Cygwin
 - Compiler and version
 - Build system Autotools - Autoconf 2.71

derobins:
Stripping out -Werror was necessary to get the warning checks working in CI because otherwise configure dies. We'll re-engineer. Autoconf 2.71 is required now due to bugs in 2.69 that cause problems with Intel's new compilers.

derobins:
The -Werror issue was fixed in #4144.

Unfortunately, we're going to have to stick with autoconf 2.71. Current Cygwin should work fine with this (we have an Autotools Cygwin CI action to confirm).

PR #4144, opened 2024-03-15
Fix Autotools -Werror cleanup

derobins:
The Autotools temporarily scrub -Werror(=whatever) from CFLAGS, etc.
  so configure checks don't trip over warnings generated by configure
  check programs. The sed line originally only scrubbed -Werror but not
  -Werror=something, which would cause errors when the '=something' was
  left behind in CFLAGS.

  The sed line has been updated to handle -Werror=something lines.

  Fixes one issue raised in #3872


derobins:
I'll manually close #3872. I need to add a comment about using autoconf 2.71.

derobins:
You can check my sed expression here:

https://sed.js.org/

byrnHDF (approved):

lrknox (approved):

PR #4145, opened 2024-03-15
Remove useless headers

derobins:
Removes unnecessary headers from C library source files.


derobins:
I have a script that identifies these. I haven't run it in a long time, so it's time to take out the trash.

byrnHDF (approved):

derobins:
Also possibly of interest:

https://github.com/include-what-you-use/include-what-you-use

byrnHDF:
> Also possibly of interest:
> 
> https://github.com/include-what-you-use/include-what-you-use

We have that as a macro pair in the config/sanitizer/tools.cmake file

jhendersonHDF (approved):

PR #4148, opened 2024-03-15
Remove useless headers from HDF5 headers

derobins:


derobins:
As before, I have a script that highlights problems.

derobins:
Wow, that's a LOT of failures that I don't see locally. Will address later.

PR #4152, opened 2024-03-15
Allow H5Soffset_simple to accept NULL offsets

derobins:
The reference manual states that the offset parameter of H5Soffset_simple()
  can be set to NULL to reset the offset of a simple dataspace to 0. This
  has never been true, and passing NULL was regarded as an error.

  The library will now accept NULL for the offset parameter and will
  correctly set the offset to zero.

  Fixes HDFFV-9299


src/H5Stest.c
```diff
@@ -367,3 +362,40 @@ H5S__internal_consistency_test(hid_t space_id)
...
+ USAGE
+    herr_t H5S__verify_offsets(hid_t space_id)
+        hid_t space_id;          IN:  dataspace id
+        const hssize_t *offset;  IN: Offset to position the selection at
+ RETURNS
+    Non-negative true/false on success, negative on failure
+ DESCRIPTION
+    This function is necessary because there is no public API call
+    that lets you get the offsets
+--------------------------------------------------------------------------*/
```
derobins, line 380: Strangely, there's no getter for a selection's offsets

src/H5Sselect.c
```diff
@@ -95,10 +90,14 @@ H5S_select_offset(H5S_t *space, const hssize_t *offset)
...
+    /* Copy the offset over. As a special case, when offset is NULL, we
+     * reset all dimensions to zero.
+     */
+    if (offset)
+        H5MM_memcpy(space->select.offset, offset, sizeof(hssize_t) * space->extent.rank);
+    else
+        memset(space->select.offset, 0, sizeof(hssize_t) * space->extent.rank);
 
     /* Indicate that the offset was changed */
     space->select.offset_changed = true;
```
fortnern, line 103: should offset_changed be set to false if NULL was passed?

derobins: We don't do that when you pass in all zeroes, so that would be something slightly new under the hood, but it could be done. We'd probably want to do that in the all zeroes case, too.

jhendersonHDF: Wouldn't we want to set it to false only when the offset was already 0?

derobins: I think the offset is only non-zero when you've moved the selection around.

qkoziol (approved):

byrnHDF (approved):

PR #4162, opened 2024-03-16
CMake CHECK_FUNCTION_EXISTS --> CHECK_SYMBOL_EXISTS

derobins:
Recommended as per CMake docs:

https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html


derobins:
Well that opened a can of worms. I'll fix this later.

PR #4164, opened 2024-03-16
Remove some H5T_copy calls that are now unnecessary

jhendersonHDF:
Removes some datatype copying calls that are now unnecessary after refactoring the datatype conversion code to use pointers internally rather than IDs

Rewrites the enum conversion function so that it uses cached copies of the source and destination datatypes in order to avoid modifying the datatypes passed in

Adds a 'recursive' field to the datatype conversion context which allows the conversion functions for members of a container datatype to skip unnecessary repetitive conversion setup code

Changes internal datatype conversion callback functions so that the source and destination datatype structure pointers are const

Removes some unused and unnecessary internal IDs registered with H5I_register

### DIFF
--- a/src/H5Aint.c
+++ b/src/H5Aint.c
@@ -681,9 +681,7 @@ H5A__read(const H5A_t *attr, const H5T_t *mem_type, void *buf)
     uint8_t    *bkg_buf   = NULL; /* background buffer */
     hssize_t    snelmts;          /* elements in attribute */
     size_t      nelmts;           /* elements in attribute*/
-    H5T_path_t *tpath    = NULL;  /* type conversion info    */
-    H5T_t      *src_type = NULL;  /* temporary datatype */
-    H5T_t      *dst_type = NULL;  /* temporary datatype */
+    H5T_path_t *tpath = NULL;     /* type conversion info    */
     size_t      src_type_size;    /* size of source type     */
     size_t      dst_type_size;    /* size of destination type */
     size_t      buf_size;         /* desired buffer size    */
@@ -723,11 +721,6 @@ H5A__read(const H5A_t *attr, const H5T_t *mem_type, void *buf)
             if (!H5T_path_noop(tpath)) {
                 H5T_bkg_t need_bkg; /* Background buffer type */
 
-                if (NULL == (src_type = H5T_copy(attr->shared->dt, H5T_COPY_ALL)))
-                    HGOTO_ERROR(H5E_ATTR, H5E_CANTCOPY, FAIL, "unable to copy attribute datatype");
-                if (NULL == (dst_type = H5T_copy(mem_type, H5T_COPY_ALL)))
-                    HGOTO_ERROR(H5E_ATTR, H5E_CANTCOPY, FAIL, "unable to copy memory datatype");
-
                 /* Get the maximum buffer size needed and allocate it */
                 buf_size = nelmts * MAX(src_type_size, dst_type_size);
                 if (NULL == (tconv_buf = H5FL_BLK_MALLOC(attr_buf, buf_size)))
@@ -752,8 +745,8 @@ H5A__read(const H5A_t *attr, const H5T_t *mem_type, void *buf)
                 }
 
                 /* Perform datatype conversion.  */
-                if (H5T_convert(tpath, src_type, dst_type, nelmts, (size_t)0, (size_t)0, tconv_buf, bkg_buf) <
-                    0)
+                if (H5T_convert(tpath, attr->shared->dt, mem_type, nelmts, (size_t)0, (size_t)0, tconv_buf,
+                                bkg_buf) < 0)
                     HGOTO_ERROR(H5E_ATTR, H5E_CANTCONVERT, FAIL, "datatype conversion failed");
 
                 /* Copy the converted data into the user's buffer */
@@ -771,10 +764,6 @@ H5A__read(const H5A_t *attr, const H5T_t *mem_type, void *buf)
 
 done:
     /* Release resources */
-    if (src_type && H5T_close(src_type) < 0)
-        HDONE_ERROR(H5E_ATTR, H5E_CANTCLOSEOBJ, FAIL, "unable to close temporary datatype");
-    if (dst_type && H5T_close(dst_type) < 0)
-        HDONE_ERROR(H5E_ATTR, H5E_CANTCLOSEOBJ, FAIL, "unable to close temporary datatype");
     if (tconv_buf)
         tconv_buf = H5FL_BLK_FREE(attr_buf, tconv_buf);
     if (bkg_buf)
@@ -805,9 +794,7 @@ H5A__write(H5A_t *attr, const H5T_t *mem_type, const void *buf)
     uint8_t    *bkg_buf   = NULL; /* temp conversion buffer */
     hssize_t    snelmts;          /* elements in attribute */
     size_t      nelmts;           /* elements in attribute */
-    H5T_path_t *tpath    = NULL;  /* conversion information*/
-    H5T_t      *src_type = NULL;  /* temporary datatype */
-    H5T_t      *dst_type = NULL;  /* temporary datatype */
+    H5T_path_t *tpath = NULL;     /* conversion information*/
     size_t      src_type_size;    /* size of source type    */
     size_t      dst_type_size;    /* size of destination type*/
     size_t      buf_size;         /* desired buffer size    */
@@ -843,11 +830,6 @@ H5A__write(H5A_t *attr, const H5T_t *mem_type, const void *buf)
         if (!H5T_path_noop(tpath)) {
             H5T_bkg_t need_bkg; /* Background buffer type */
 
-            if (NULL == (src_type = H5T_copy(mem_type, H5T_COPY_ALL)))
-                HGOTO_ERROR(H5E_ATTR, H5E_CANTCOPY, FAIL, "unable to copy memory datatype");
-            if (NULL == (dst_type = H5T_copy(attr->shared->dt, H5T_COPY_ALL)))
-                HGOTO_ERROR(H5E_ATTR, H5E_CANTCOPY, FAIL, "unable to copy attribute datatype");
-
             /* Get the maximum buffer size needed and allocate it */
             buf_size = nelmts * MAX(src_type_size, dst_type_size);
             if (NULL == (tconv_buf = H5FL_BLK_MALLOC(attr_buf, buf_size)))
@@ -880,7 +862,8 @@ H5A__write(H5A_t *attr, const H5T_t *mem_type, const void *buf)
             }
 
             /* Perform datatype conversion */
-            if (H5T_convert(tpath, src_type, dst_type, nelmts, (size_t)0, (size_t)0, tconv_buf, bkg_buf) < 0)
+            if (H5T_convert(tpath, mem_type, attr->shared->dt, nelmts, (size_t)0, (size_t)0, tconv_buf,
+                            bkg_buf) < 0)
                 HGOTO_ERROR(H5E_ATTR, H5E_CANTCONVERT, FAIL, "datatype conversion failed");
 
             /* Free the previous attribute data buffer, if there is one */
@@ -911,10 +894,6 @@ H5A__write(H5A_t *attr, const H5T_t *mem_type, const void *buf)
 
 done:
     /* Release resources */
-    if (src_type && H5T_close(src_type) < 0)
-        HDONE_ERROR(H5E_ATTR, H5E_CANTCLOSEOBJ, FAIL, "unable to close temporary datatype");
-    if (dst_type && H5T_close(dst_type) < 0)
-        HDONE_ERROR(H5E_ATTR, H5E_CANTCLOSEOBJ, FAIL, "unable to close temporary datatype");
     if (tconv_buf)
         tconv_buf = H5FL_BLK_FREE(attr_buf, tconv_buf);
     if (bkg_buf)
@@ -2084,10 +2063,10 @@ H5A__attr_copy_file(const H5A_t *attr_src, H5F_t *file_dst, bool *recompute_size
 {
     H5A_t   *attr_dst    = NULL; /* Destination attribute */
     H5T_t   *dt_mem      = NULL; /* Memory datatype */
+    H5S_t   *buf_space   = NULL; /* Dataspace describing buffer */
     void    *buf         = NULL; /* Buffer for copying data */
     void    *reclaim_buf = NULL; /* Buffer for reclaiming data */
     void    *bkg_buf     = NULL; /* Background buffer */
-    hid_t    buf_sid     = -1;   /* ID for buffer dataspace */
     hssize_t sdst_nelmts;        /* # of elements in destination attribute (signed) */
     size_t   dst_nelmts;         /* # of elements in destination attribute */
     size_t   dst_dt_size;        /* Size of destination attribute datatype */
@@ -2200,7 +2179,6 @@ H5A__attr_copy_file(const H5A_t *attr_src, H5F_t *file_dst, bool *recompute_size
             size_t      src_dt_size;                   /* Source datatype size */
             size_t      tmp_dt_size;                   /* Temp. datatype size */
             size_t      max_dt_size;                   /* Max atatype size */
-            H5S_t      *buf_space;                     /* Dataspace describing buffer */
             hsize_t     buf_dim;                       /* Dimension for buffer */
             size_t      nelmts;                        /* Number of elements in buffer */
             size_t      buf_size;                      /* Size of copy buffer */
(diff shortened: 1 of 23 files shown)


src/H5Aint.c
```diff
@@ -2241,12 +2219,6 @@ H5A__attr_copy_file(const H5A_t *attr_src, H5F_t *file_dst, bool *recompute_size
             if (NULL == (buf_space = H5S_create_simple((unsigned)1, &buf_dim, NULL)))
                 HGOTO_ERROR(H5E_DATASPACE, H5E_CANTCREATE, NULL, "can't create simple dataspace");
 
-            /* Register */
-            if ((buf_sid = H5I_register(H5I_DATASPACE, buf_space, false)) < 0) {
```
jhendersonHDF, line 2245: This ID was only being used as a wrapper for closing `buf_space`

src/H5Dchunk.c
```diff
@@ -296,9 +296,9 @@ static herr_t   H5D__create_piece_file_map_all(H5D_dset_io_info_t *di, H5D_io_in
 static herr_t   H5D__create_piece_file_map_hyper(H5D_dset_io_info_t *di, H5D_io_info_t *io_info);
 static herr_t   H5D__create_piece_mem_map_1d(const H5D_dset_io_info_t *di);
 static herr_t   H5D__create_piece_mem_map_hyper(const H5D_dset_io_info_t *di);
-static herr_t   H5D__piece_file_cb(void *elem, H5T_t *type, unsigned ndims, const hsize_t *coords,
+static herr_t   H5D__piece_file_cb(void *elem, const H5T_t *type, unsigned ndims, const hsize_t *coords,
```
jhendersonHDF, line 299: Putting `const` back in a lot of places it was previously removed from now that `H5T_convert`'s datatype parameters are marked `const`

src/H5Dchunk.c
```diff
@@ -6939,12 +6928,6 @@ H5D__chunk_copy(H5F_t *f_src, H5O_storage_chunk_t *storage_src, H5O_layout_chunk
         if (NULL == (buf_space = H5S_create_simple((unsigned)1, &buf_dim, NULL)))
             HGOTO_ERROR(H5E_DATASPACE, H5E_CANTCREATE, FAIL, "can't create simple dataspace");
 
-        /* Register */
-        if ((sid_buf = H5I_register(H5I_DATASPACE, buf_space, false)) < 0) {
```
jhendersonHDF, line 6943: This ID was only being used as a wrapper for closing `buf_space`

src/H5Dcompact.c
```diff
@@ -548,12 +547,6 @@ H5D__compact_copy(H5F_t *f_src, H5O_storage_compact_t *_storage_src, H5F_t *f_ds
         if (NULL == (buf_space = H5S_create_simple((unsigned)1, &buf_dim, NULL)))
             HGOTO_ERROR(H5E_DATASPACE, H5E_CANTCREATE, FAIL, "can't create simple dataspace");
 
-        /* Register */
-        if ((buf_sid = H5I_register(H5I_DATASPACE, buf_space, false)) < 0) {
```
jhendersonHDF, line 552: This ID was only being used as a wrapper for closing `buf_space`

src/H5Dcontig.c
```diff
@@ -1689,12 +1688,6 @@ H5D__contig_copy(H5F_t *f_src, const H5O_storage_contig_t *storage_src, H5F_t *f
         if (NULL == (buf_space = H5S_create_simple((unsigned)1, buf_dim, NULL)))
             HGOTO_ERROR(H5E_DATASPACE, H5E_CANTCREATE, FAIL, "can't create simple dataspace");
 
-        /* Register */
-        if ((buf_sid = H5I_register(H5I_DATASPACE, buf_space, false)) < 0) {
```
jhendersonHDF, line 1693: This ID was only being used as a wrapper for closing `buf_space`

src/H5Tpkg.h
```diff
@@ -506,469 +516,531 @@ H5_DLL H5T_t *H5T__open_name(const H5G_loc_t *loc, const char *name);
...
-                                  size_t nelmts, size_t buf_stride, size_t bkg_stride, void *_buf, void *bkg);
-H5_DLL herr_t H5T__conv_struct(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t *conv_ctx,
-                               size_t nelmts, size_t buf_stride, size_t bkg_stride, void *_buf, void *bkg);
-H5_DLL herr_t H5T__conv_struct_opt(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t *conv_ctx,
-                                   size_t nelmts, size_t buf_stride, size_t bkg_stride, void *_buf,
-                                   void *bkg);
-H5_DLL herr_t H5T__conv_enum(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t *conv_ctx,
-                             size_t nelmts, size_t buf_stride, size_t bkg_stride, void *buf, void *bkg);
-H5_DLL herr_t H5T__conv_enum_numeric(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata,
+H5_DLL herr_t H5T__conv_noop(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata,
```
jhendersonHDF, line 524: Most of the changes in this file are just putting `const` on `src` and `dst`

src/H5T.c
```diff
@@ -2537,14 +2537,14 @@ H5T__register(H5T_pers_t pers, const char *name, H5T_t *src, H5T_t *dst, H5T_con
...
-                HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "unable to copy src datatype");
-            if (NULL == (tmp_dtype = H5T_copy(old_path->dst, H5T_COPY_ALL)))
-                HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "unable to copy dst datatype");
-
             memset(&cdata, 0, sizeof cdata);
             cdata.command = H5T_CONV_INIT;
             if (conv->is_app) {
+                if (NULL == (tmp_stype = H5T_copy(old_path->src, H5T_COPY_ALL)))
+                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "unable to copy src datatype");
+                if (NULL == (tmp_dtype = H5T_copy(old_path->dst, H5T_COPY_ALL)))
```
jhendersonHDF, line 2545: Only need to make a copy of the datatypes if we're going to register an ID for them and pass them to the application conversion function

src/H5T.c
```diff
@@ -5288,14 +5261,6 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
         table          = path;
     } /* end else-if */
 
-    /* Set the flag to indicate both source and destination types are compound types
-     * for the optimization of data reading (in H5Dio.c).
-     * Make sure that path->are_compounds is only true for compound types.
-     */
-    path->are_compounds = false;
```
jhendersonHDF, line 5295: This field wasn't used anywhere and is trivial to add back in the future if needed

src/H5Dint.c
```diff
@@ -3650,19 +3650,22 @@ H5D_get_create_plist(const H5D_t *dset)
             uint8_t *bkg_buf = NULL; /* Background conversion buffer */
             size_t   bkg_size;       /* Size of background buffer */
 
-            if (NULL == (src_type = H5T_copy(dset->shared->type, H5T_COPY_ALL)))
-                HGOTO_ERROR(H5E_DATASET, H5E_CANTCOPY, FAIL, "unable to copy dataset's datatype");
-            if (NULL == (dst_type = H5T_copy(copied_fill.type, H5T_COPY_TRANSIENT)))
```
jhendersonHDF, line 3655: Any copy call that used `H5T_COPY_ALL` was able to be removed, but the `H5T_COPY_TRANSIENT` ones were left in for now as the code seems to want to set the location of the datatype such that it's a memory type. Since this is only needed for variable-length and reference types, I updated this to only make transient copies of the type if it's one of those two classes. We can probably optimize this further by pushing the copy calls down into the conversion routine and caching those copies.

src/H5Tconv.c
```diff
@@ -2753,131 +2795,205 @@ H5T__conv_struct_opt(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5T__conv_enum_init(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata)
```
jhendersonHDF, line 2756: The enum conversion routine had to be rewritten slightly so that it makes a copy of the source and destination datatypes, caches them and works directly on those, rather than modifying the original datatypes when it sorts the enum members by name or value. Otherwise, removing the datatype copies higher up in the library would cause the datatype an object was created with to be surprisingly modified. The enum members would then be out of order in the file and cause `h5dump` tests to fail when they expected the members to be in the order they were inserted into the enum type. Likely not a big deal, but I wouldn't generally expect the datatypes passed to `H5Tconvert`/`H5T_convert`, or at least the source datatype, to be modified.

While this was rewritten mostly from a functionality perspective, some synthetic benchmarking of converting 100,000,000 enum elements shows very little change in conversion performance (usually in favor of this update), to the point that it's just noise.

src/H5Tconv.c
```diff
@@ -2753,131 +2795,205 @@ H5T__conv_struct_opt(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_
...
+{
+    H5T_conv_enum_t *priv          = NULL; /* Private conversion data */
+    int             *map           = NULL; /* Map from src value to dst idx */
+    bool             rebuild_cache = false;
+    herr_t           ret_value     = SUCCEED;
 
     FUNC_ENTER_PACKAGE
 
     cdata->need_bkg = H5T_BKG_NO;
-    if (NULL == (priv = (H5T_enum_struct_t *)(cdata->priv = H5MM_calloc(sizeof(*priv)))))
```
jhendersonHDF, line 2769: The large block of logic below was essentially just moved into the `if (rebuild_cache) {` block

jhendersonHDF:
@qkoziol I wanted to avoid making copies of the datatypes by just making a mapping from `src`->`dst`, but I couldn't see a good way of knowing when we need to invalidate the mapping without having a copy of `src` and `dst` to compare against when hitting the `H5T_CONV_CONV` step after `H5T_CONV_INIT`. Sometimes the datatypes differed between the two steps (probably due to some other datatype copying higher up), so this seemed necessary. It was far easier to do it this way and doesn't really appear to affect conversion performance since the cache is generally reused after being rebuilt when a mismatch of datatypes is found.

jhendersonHDF:
I will have a separate PR with a RELEASE.txt note covering the recent changes in the H5T code, including this change.

mattjala (approved):

src/H5Tconv.c
```diff
@@ -2966,18 +3079,18 @@ H5T__conv_enum(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t
...
-                d         = buf + (nelmts - 1) * dst->shared->size;
+                H5_CHECK_OVERFLOW(src_sh->size, size_t, ssize_t);
+                H5_CHECK_OVERFLOW(dst_sh->size, size_t, ssize_t);
+                src_delta = -(ssize_t)src_sh->size;
+                dst_delta = -(ssize_t)dst_sh->size;
+                s         = buf + (nelmts - 1) * src_sh->size;
+                d         = buf + (nelmts - 1) * dst_sh->size;
             }
 
             for (i = 0; i < nelmts; i++, s += src_delta, d += dst_delta) {
```
qkoziol: Small optimization:  I suggest hoisting the if/else outside of the loop (duplicating the loop in each branch), since 'priv->length' is constant for all elements.

src/H5Tconv.c
```diff
@@ -2989,9 +3102,9 @@ H5T__conv_enum(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t
...
                      * SLU - 2011/5/24)
                      */
-                    if (1 == src->shared->size)
+                    if (1 == src_sh->size)
                         n = *((signed char *)s);
-                    else if (sizeof(short) == src->shared->size)
+                    else if (sizeof(short) == src_sh->size)
                         n = *((short *)((void *)s));
                     else
                         n = *((int *)((void *)s));
```
qkoziol, line 3106: Can an enum's base type be a long?

src/H5Tconv.c
```diff
@@ -3006,28 +3119,28 @@ H5T__conv_enum(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t
                                 conv_ctx->u.conv.dst_type_id, s, d, conv_ctx->u.conv.cb_struct.user_data);
```
qkoziol, line 3115: I believe that this can underflow as well as overflow, in which case it should throw a RANGE_LO exception.

src/H5Tconv.c
```diff
@@ -2753,131 +2795,205 @@ H5T__conv_struct_opt(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_
...
-        if (j >= dst->shared->u.enumer.nmembs)
-            HGOTO_ERROR(H5E_DATATYPE, H5E_UNSUPPORTED, FAIL,
-                        "source type is not a subset of destination type");
-        priv->src2dst[i] = (int)j;
-    } /* end for */
+    priv = (H5T_conv_enum_t *)(cdata->priv);
+    if (!priv) {
+        if (NULL == (priv = (H5T_conv_enum_t *)(cdata->priv = calloc(1, sizeof(*priv)))))
+            HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "memory allocation failed");
+        if (0 == src->shared->u.enumer.nmembs)
```
qkoziol: This 'if' (on the source # of members) can be removed, since it's covered in the 'if(rebuil_cache)' case.

qkoziol (approved):
Small tweaks suggested, but definitely good to go now.

src/H5Tconv.c
jhendersonHDF: Yes, I don't think this optimization ever took that into account

src/H5Tconv.c
jhendersonHDF: I wouldn't be surprised if there are bugs with conversions on enums. There are old issues still open specifically around enums. I was trying to avoid making any changes to functionality.

src/H5Tconv.c
jhendersonHDF: True. Will remove.

src/H5Tconv.c
jhendersonHDF: I'd be surprised if most compilers don't optimize that with branch prediction, but I can do that.

qkoziol:
Looking good  👍

src/H5Tconv.c
```diff
@@ -2753,131 +2795,201 @@ H5T__conv_struct_opt(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_
...
+    else {
+        /* Check if we need to rebuild our cache. For now, treat
+         * enums as different even if one is just a subset of the
+         * other
+         */
+        if (cdata->command == H5T_CONV_CONV && conv_ctx->u.conv.recursive)
+            /* Recursive conversion; we can reuse the cache */
+            rebuild_cache = false;
+        else {
+            if (0 != H5T_cmp(src, priv->src_copy, false) || 0 != H5T_cmp(dst, priv->dst_copy, false))
```
fortnern, line 2824: How does this not cause problems if there's an enum inside a compound? Wouldn't the compound code mark it as recursive then this enum code would fail to calculate the mapping?

src/H5Tconv.c
```diff
@@ -2950,13 +3060,12 @@ H5T__conv_enum(H5T_t *src, H5T_t *dst, H5T_cdata_t *cdata, const H5T_conv_ctx_t
...
 
-            /* priv->src2dst map was computed for certain sort keys. Make sure those same
-             * sort keys are used here during conversion. See H5T__conv_enum_init(). But
-             * we actually don't care about the source type's order when doing the O(1)
-             * conversion algorithm, which is turned on by non-zero priv->length */
-            H5T__sort_name(dst, NULL);
-            if (!priv->length)
-                H5T__sort_value(src, NULL);
+            /* Reuse cache if possible, rebuild otherwise */
+            if (H5T__conv_enum_init(src, dst, cdata, conv_ctx) < 0)
```
fortnern, line 3064: Why the change here?

fortnern:
Looks good aside from my questions

src/H5Tconv.c
jhendersonHDF: At this point, the compound init should have called `H5T_path_find` for all the member types, initing the enum path in the process and building/rebuilding the cache as necessary. At the point we get here from conversion command with this marked as recursive, we should have a valid cache that we can just use for each of the compound elements being converted rather than checking if we need to rebuild the cache for every element.

src/H5Tconv.c
jhendersonHDF: Part of the cache building code in init does the same sorting here, so whether we're reusing the cache or rebuilding it, we should have the same sorted order as expected here.

PR #4165, opened 2024-03-16
Fix some minor warnings

jhendersonHDF:


derobins (approved):

HDF5Examples/C/H5T/h5ex_t_opaqueatt.c
```diff
@@ -121,7 +121,7 @@ main(void)
      */
     printf("Datatype tag for %s is: \"%s\"\n", ATTRIBUTE, tag);
     for (i = 0; i < dims[0]; i++) {
-        printf("%s[%llu]: ", ATTRIBUTE, i);
+        printf("%s[%" PRIuHSIZE "]: ", ATTRIBUTE, i);
```
byrnHDF, line 124: is PRIuHSIZE available afterwards when using a binary to build the examples?

byrnHDF (changes requested):
The examples have to be built in line with the library and after an installation.

HDF5Examples/C/H5T/h5ex_t_opaqueatt.c
derobins: Yes, It's in H5public.h.

byrnHDF (approved):

PR #4167, opened 2024-03-17
Bump the size of the mirror VFD IP field

derobins:
The IP address string isn't big enought to hold an IPv4-mapped IPv6 address.


derobins:
From @hyoklee

https://stackoverflow.com/questions/1076714/max-length-for-client-ip-address

byrnHDF (approved):

qkoziol (approved):

PR #4172, opened 2024-03-18
Address code page issues w/ Windows file paths

derobins:
On Windows, HDF5 attempted to convert file paths passed to open() and remove() to UTF-16 in order to handle Unicode file paths. This scheme does not work when the system uses code pages to handle non-ASCII file names.

As suggested in the forum post below, we now also try to see if we can open the file with open(), which should handle systems where non-ASCII code pages are in use.

https://forum.hdfgroup.org/t/open-create-hdf5-files-with-non-utf8-chars-such-as-shift-jis/11785

### DIFF
--- a/src/H5system.c
+++ b/src/H5system.c
@@ -592,7 +592,7 @@ Wremove(const char *path)
     wchar_t *wpath = NULL; /* UTF-16 version of the path */
     int      ret   = -1;
 
-    /* First try opening the file with the normal POSIX open() call.
+    /* First try removing the file with the normal POSIX remove() call.
      * This will handle ASCII without additional processing as well as
      * systems where code pages are being used instead of true Unicode.
      */
--- a/src/H5system.c
+++ b/src/H5system.c
@@ -514,16 +514,17 @@ error:
 } /* end H5_get_utf16_str() */
 
 /*-------------------------------------------------------------------------
- * Function:     Wopen_utf8
+ * Function:     Wopen
  *
- * Purpose:      UTF-8 equivalent of open(2) for use on Windows
+ * Purpose:      Equivalent of open(2) for use on Windows. Necessary to
+ *               handle code pages and Unicode on that platform.
  *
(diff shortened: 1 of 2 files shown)


qkoziol (approved):

byrnHDF (approved):

emmenlau:
Dear @derobins and all! Thanks a lot for this effort, its appreciated! However the current state of UTF8-support in hdf5 is actually causing me some trouble :-(

The new behavior of hdf5 is to try and interpret a file name as ASCII first, and only if that fails, interpret as UTF8 (via the conversion to UTF16 for MS Windows file open methods). This logic is not only applied when opening or removing an existing file, but also when creating a new file. However, I would consider it rather unpredictable whether a file path can be created, when interpreting the string as ASCII?

For example, the file name "äöüㅜ☺㋡🌀.h5" can be successfully created on our Windows 11 test system (English locale with a fallback to German), but the actual file on disk will be named "\xC3\xA4\xC3\xB6\xC3\xBC\xE3\x85\x9C\xE2\x98\xBA\xE3\x8B\xA1\xF0\x9F\x8C\x80.h5"!

I guess that there are combinations of UTF8-characters that are actually illegal in ASCII - but there are seemingly also a lot of UTF8-characters that are just garbled ASCII, but still valid file names :-( :-(

Personally I would much prefer if hdf5 would give concrete user control over this. For example, our application is using exclusively `UTF8` in `std::string` on all platforms. And looking at https://utf8everywhere.org/ and related efforts, this standard is becoming more and more adopted in the C++ world.

It would be great if there where a compiler define or file-open-parameter that allows to enforce this standard.

Thanks a lot for your consideration!

takluyver:
Yeah, as @emmenlau says, this has largely broken using UTF-8 filenames; h5py tests failed on Windows as soon as we updated the version of HDF5 in use.

I'll open a new issue about this.

takluyver:
#5037 is the new issue

PR #4176, opened 2024-03-18
Add initial skeleton file for daily VFD CI workflow

jhendersonHDF:
Adds testing of Subfiling VFD

### DIFF
--- a/.github/workflows/vfd-main.yml
+++ b/.github/workflows/vfd-main.yml
@@ -1,0 +1,66 @@
+name: hdf5 VFD CI main
+
+on:
+  workflow_call:
+    inputs:
+      build_sys:
+        description: "build system used"
+        required: true
+        type: string
+      build_mode:
+        description: "release vs. debug build"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  # Test HDF5 VFDs that are always built by default
+  #hdf5_vfd_standard:
+  #  uses: ./.github/workflows/vfd-standard.yml
+  #  with:
+  #    build_sys: "${{ inputs.build_sys }}"
+  #    build_mode: "${{ inputs.build_mode }}"
+
+  # Test HDF5 MPI I/O VFD
+  #hdf5_vfd_mpiio:
+  #  uses: ./.github/workflows/vfd-mpiio.yml
+  #  with:
+  #    build_sys: "${{ inputs.build_sys }}"
+  #    build_mode: "${{ inputs.build_mode }}"
+
+  # Test HDF5 Direct VFD
+  #hdf5_vfd_direct:
+  #  uses: ./.github/workflows/vfd-direct.yml
+  #  with:
+  #    build_sys: "${{ inputs.build_sys }}"
+  #    build_mode: "${{ inputs.build_mode }}"
+
+  #Test HDF5 Mirror VFD
+  #hdf5_vfd_mirror:
+  #  uses: ./.github/workflows/vfd-mirror.yml
+  #  with:
+  #    build_sys: "${{ inputs.build_sys }}"
+  #    build_mode: "${{ inputs.build_mode }}"
+
+  #Test HDF5 ROS3 VFD
+  #hdf5_vfd_ros3:
+  #  uses: ./.github/workflows/vfd-ros3.yml
+  #  with:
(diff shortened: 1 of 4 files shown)


.github/workflows/vfd-main.yml
```diff
@@ -0,0 +1,66 @@
...
+        description: "release vs. debug build"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  # Test HDF5 VFDs that are always built by default
+  #hdf5_vfd_standard:
```
jhendersonHDF, line 20: These can be filled in afterward, but I listed the ones I know about here. It would be nice for all of the "standard" VFDs to all use the same build of HDF5 instead of all being a separate build, but it would also be nice to be able to test the different VFDs separately to better parallelize.

.github/workflows/vfd-subfiling.yml
```diff
@@ -0,0 +1,136 @@
...
+  contents: read
+
+jobs:
+  build_and_test:
+    strategy:
+      # Let jobs run to completion even if one fails
+      fail-fast: false
+      matrix:
+        os_name: ["Ubuntu"]
+#        os_name: ["Ubuntu", "MacOS"]
```
jhendersonHDF, line 25: Commented out some things here that need more work. MacOS testing needs added; MPICH testing will require building from source since the default version from APT on Ubuntu hangs almost immediately in Subfiling. I'd like to create separate actions for building and then caching OpenMPI and MPICH so that all of our actions can reuse that rather than constantly rebuilding it or installing a likely old version from apt

.github/workflows/vfd.yml
```diff
@@ -0,0 +1,47 @@
...
+          - build_sys: "CMake"
+            build_mode: "Release"
+          - build_sys: "CMake"
+            build_mode: "Debug"
+          - build_sys: "Autotools"
+            build_mode: "production"
+          - build_sys: "Autotools"
+            build_mode: "debug"
+        # Exclude mismatched configurations
+        exclude:
```
jhendersonHDF, line 28: Seems to need both the include and exclude here, otherwise GitHub either excludes all the configurations or includes all the configurations

jhendersonHDF:
The workflows were organized in this way because not all VFDs will be testable across all the three main OSes. It will be nice to reuse builds of HDF5 where we can, but this will probably only be possible for the more standard VFDs.

glennsong09 (approved):

.github/workflows/vfd.yml
byrnHDF: You could just split this into autotools and cmake. It may be better for clarity.

byrnHDF:
I don't have a strong issue against this implementation, however maybe we should consider separating autotools vs CMake into separate files - this would make it easier for future devs to understand with minor duplication.

jhendersonHDF:
> I don't have a strong issue against this implementation, however maybe we should consider separating autotools vs CMake into separate files - this would make it easier for future devs to understand with minor duplication.

I went that way at first, but not only did the files look nearly identical, it also seemed to just be adding extra levels of indirection for no real good reason, which means you have to keep digging to find the real files to edit. Since most of the VFDs other than the "standard" ones are likely going to have to build the library differently from each other anyway, it's probably easiest if they all have their Autotools and CMake logic in one separate file each. At that point, it seems the only difference between CMake and Autotools is whether the workflow file above them calls those workflows with the string "Autotools" or "CMake". Happy for any suggestions on how to split this though; it just mostly seemed unnecessary.

byrnHDF:
Yeah, since you tried it already, this is fine. It's kind of the same with our CI, too. But it helps when looking at the reports to see the groupings. And sometimes autotools/CMake needs a special change for one job that doesn't need to be everywhere. Also I like not having to decipher the if conditions to follow the flow.
But really, no strong feeling either way.

byrnHDF (approved):

jhendersonHDF:
> Yeah, since you tried it already, this is fine. It's kind of the same with our CI, too. But it helps when looking at the reports to see the groupings. And sometimes autotools/CMake needs a special change for one job that doesn't need to be everywhere. Also I like not having to decipher the if conditions to follow the flow. But really, no strong feeling either way.

Sure, and we can always change this once we at least have something up and running.

PR #4190, opened 2024-03-20
Fixes several MinGW + Autotools issues

derobins:
* Fixes detection of various Windows libraries, etc.
* Corrects alarm(2) configure checks
* Uses Win32 threads by default w/ Pthreads override, if desired
* Set _WIN32_WINNT correctly for MinGW
* Fix setenv(3) wrapper for MinGW, which does not have getenv_s()

MinGW Autotools support is still not Amazing, but this at least allows the library and tools build and is better about thread-safety

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1544,10 +1544,10 @@ case $host_os in
 esac
 
 ## Windows
-case "`uname`" in
-  MINGW*)
+case "$host_os" in
+  *mingw*)
     # The Winsock library
-    AC_CHECK_LIB([ws2_32], [GetUserName])
+    AC_CHECK_LIB([ws2_32], [htons])
     ;;
 esac
 
@@ -1562,9 +1562,9 @@ AC_CHECK_HEADERS([arpa/inet.h netdb.h netinet/in.h sys/socket.h])
 ##
 if test "X${enable_shared}" = "Xyes"; then
   AC_MSG_CHECKING([if libtool needs -no-undefined flag to build shared libraries])
-  case "`uname`" in
-    CYGWIN*|MINGW*|AIX*)
-      ## Add in the -no-undefined flag to LDFLAGS for libtool.
+  case "$$host_os" in
+    *cygwin*|*mingw*|*aix*)
+      ## Add in the -no-undefined flag to LDFLAGS for libtool
       AC_MSG_RESULT([yes])
       H5_LDFLAGS="$H5_LDFLAGS -no-undefined"
       ;;
@@ -1583,7 +1583,7 @@ AC_SYS_LARGEFILE
 ## ----------------------------------------------------------------------
 ## Add necessary defines for Linux Systems.
 ##
-case "$host_cpu-$host_vendor-$host_os" in
+case "$host_os" in
   *linux*)
     ## Add POSIX support on Linux systems, so <features.h> defines
     ## __USE_POSIX, which is required to get the prototype for fdopen
@@ -1613,6 +1613,7 @@ case "$host_cpu-$host_vendor-$host_os" in
     H5_CPPFLAGS="-D_GNU_SOURCE $H5_CPPFLAGS"
     ;;
   *mingw*)
+    HDF_MINGW="yes"
     AC_DEFINE([HAVE_WINDOWS], [1], [Define if this is a Windows machine])
     AC_DEFINE([HAVE_WIN32_API], [1], [Define if on the Windows platform using the Win32 API])
     AC_DEFINE([HAVE_MINGW], [1], [Define if using MinGW])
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -1544,10 +1544,10 @@ case $host_os in
 esac
 
 ## Windows
-case "`uname`" in
-  MINGW*)
+case "$host_os" in
+  *mingw*)
     # The Winsock library
-    AC_CHECK_LIB([ws2_32], [GetUserName])
+    AC_CHECK_LIB([ws2_32], [htons])
```
derobins, line 1550: GetUserName() doesn't really appear in Winsock2 (it's an alias for GetUserNameA or GetUserNameEx) and isn't used in the library anyway.

configure.ac
```diff
@@ -1562,9 +1562,9 @@ AC_CHECK_HEADERS([arpa/inet.h netdb.h netinet/in.h sys/socket.h])
 ##
 if test "X${enable_shared}" = "Xyes"; then
   AC_MSG_CHECKING([if libtool needs -no-undefined flag to build shared libraries])
-  case "`uname`" in
-    CYGWIN*|MINGW*|AIX*)
-      ## Add in the -no-undefined flag to LDFLAGS for libtool.
+  case "$$host_os" in
+    *cygwin*|*mingw*|*aix*)
+      ## Add in the -no-undefined flag to LDFLAGS for libtool
```
derobins, line 1567: All these uname tests fail to detect MinGW, etc. $host_os is more accurate and works correctly.

configure.ac
```diff
@@ -2026,101 +2027,118 @@ if test "X$THREADSAFE" = "Xyes"; then
     ## and/or a library path.  If the library path is specified then it must
     ## be preceded by a comma.
     ##
-    ## Thread-safety in HDF5 only uses Pthreads via configure, so the
-    ## default is "check", though this only has an effect when
-    ## --enable-threadsafe is specified.
+    ## The default is to use Pthreads when building with the Autotools, unless
+    ## we're building w/ MinGW.
```
derobins, line 2031: Since MinGW produces Win32 executables, it makes more sense to prefer Win32 threads.

configure.ac
```diff
@@ -2026,101 +2027,118 @@ if test "X$THREADSAFE" = "Xyes"; then
...
+            mingw_use_win32_threads="yes"
+            unset HAVE_PTHREAD
+            AC_DEFINE([HAVE_WIN_THREADS], [1], [Define to 1 if you have win32 threads])
         fi
-        ;;
-      no)
-        AC_MSG_ERROR([Must use Pthreads with thread safety])
-        ;;
-      *)
+    fi
```
derobins, line 2052: The above tests are new, and check to see if we're on MinGW and if someone specified --enable-pthreads=(yes or paths).

Everything below is just tabbing the existing code over by one.

configure.ac
```diff
@@ -2226,12 +2244,23 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
...
+case "$host_os" in
+  *mingw*)
+    # alarm(2) support is spotty in MinGW, so assume it doesn't exist
+    #
+    # https://lists.gnu.org/archive/html/bug-gnulib/2013-03/msg00040.html
+    ;;
+  *)
+    AC_CHECK_FUNCS([alarm])
+    ;;
+esac
```
derobins, line 2262: Alarm is flaky on MinGW and sometimes gets picked up via AC_CHECK_FUNCS, even when it doesn't work.

src/H5private.h
```diff
@@ -121,6 +121,14 @@
 #define WIN32_LEAN_AND_MEAN /* Exclude rarely-used stuff from Windows headers */
 #define NOGDI               /* Exclude Graphic Display Interface macros */
 
+/* InitOnceExecuteOnce() requires 0x0600 to work on MinGW w/ Win32 threads */
+#if defined(H5_HAVE_MINGW) && defined(H5_HAVE_THREADSAFE)
+#if !defined(_WIN32_WINNT) || (_WIN32_WINNT < 0x0600)
+#undef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600
+#endif
+#endif
```
derobins, line 130: It turns out MinGW sets this value *very* conservatively for maximum compatibility (sometimes Windows NT4!). We bump it here in order to get the threading functionality we need.

src/H5system.c
```diff
@@ -339,13 +339,20 @@ Wsetenv(const char *name, const char *value, int overwrite)
...
         if (err || bufsize)
             return (int)err;
-    } /* end if */
+#else
+        /* MinGW doesn't have getenv_s() */
+        char *test = getenv(name);
+        if (*test)
+            return FAIL;
+#endif
+    }
```
derobins, line 355: Need to hack around MinGW's spotty support of the _s functions.

derobins:
We're still not doing a spectacular job at supporting MinGW via the Autotools. This does do the right thing when it comes to Win32 vs. Pthreads, though, and I can build static executables and run them via Wine. That's better than before.

byrnHDF (approved):

derobins:
This is for cross-compiling using MinGW on Linux, btw.

test/testframe.c
```diff
@@ -637,7 +637,8 @@ SetTest(const char *testname, int action)
 /* Enable a test timer that will kill long-running tests, the time is configurable
  * via an environment variable.
  *
- * Only useful on POSIX systems where alarm(2) is present.
+ * Only useful on POSIX systems where alarm(2) is present. This does not include
+ * MinGW builds, which will often incorrectly decide that alarm(2) exists.
```
jhendersonHDF, line 641: Is this a side effect of needing some compilation flag for alarm to link/work correctly? I wouldn't be surprised if `AC_CHECK_FUNCS` is picking up the symbol from a header file, but it depends on something being #defined for it to actually be available.

derobins: Apparently it's an optional thing to build in and the header and actual symbol aren't always in sync.

jhendersonHDF (approved):

PR #4191, opened 2024-03-20
h5diff compares attribute data like dataset data

byrnHDF:


tools/src/h5diff/h5diff_common.c
```diff
@@ -816,7 +816,7 @@ usage(void)
     PRINTVALSTREAM(rawoutstream, "  2) Datasets\n");
     PRINTVALSTREAM(rawoutstream,
                    "      Array rank and dimensions, datatypes, and data values are compared.\n");
-    PRINTVALSTREAM(rawoutstream, "  3) Datatypes\n");
+    PRINTVALSTREAM(rawoutstream, "  3) Attributes and Datatypes\n");
```
jhendersonHDF: Should this have gone on "Datasets" above?

mattjala: Should this group attributes and datasets instead of attributes and datatypes?

byrnHDF: yes!!

jhendersonHDF (approved):

derobins (approved):

PR #4194, opened 2024-03-20
Fix issue with Subfiling VFD and multiple opens of same file

jhendersonHDF:


jhendersonHDF:
This also fixes a few VFD test failures with subfiling related to testing multiple file opens.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -188,23 +187,27 @@ H5_get_subfiling_object(int64_t object_id)
             assert((size_t)obj_index < sf_context_cache_size);
         }
 
-        /*
-         * Since this cache currently just keeps all entries until
-         * application exit, context entry indices should just be
-         * consecutive
-         */
```
jhendersonHDF, line 195: After #4159, this cache invariant is no longer true in the case of multiple file opens and/or files closed out of order.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -310,13 +313,21 @@ H5_free_subfiling_object(int64_t object_id)
             H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTGET, FAIL,
                                     "couldn't get subfiling context for subfiling object ID");
 
-        if (H5_free_subfiling_object_int(sf_context) < 0)
-            H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "couldn't free subfiling context object");
+        if (sf_context->file_ref == 0 || --sf_context->file_ref == 0) {
```
jhendersonHDF, line 317: Check for a ref. of 0 here since we can end up failing during context object creation before it's been associated with a file.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -610,32 +623,28 @@ H5_open_subfiling_stub_file(const char *name, unsigned flags, MPI_Comm file_comm
...
- *              storage arrays can be stack based rather than explicitly
- *              allocated and freed.
- *
- *              The Internal function is responsible for sending all IOC
- *              instances, the (sub)file open requests.
- *
- *              Prior to calling the internal open function, we initialize
- *              a new subfiling context that contains topology info and
- *              new MPI communicators that facilitate messaging between
- *              HDF5 clients and the IOCs.
```
jhendersonHDF, line 626: Most of this comment hasn't been relevant in a long time. Updated a bit to reflect the current behavior of `H5_open_subfiles`

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -648,21 +657,40 @@ H5_open_subfiles(const char *base_filename, uint64_t file_id, H5FD_subfiling_par
...
+        if (init_subfiling(base_filename, file_id, subfiling_config, file_acc_flags, file_comm, &context_id) <
+            0)
+            H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL, "couldn't initialize subfiling context");
 
-    /*
-     * If we're actually using the IOCs, we will
-     * start the service threads on the identified
-     * ranks as part of the subfile opening.
-     */
-    if (open_subfile_with_context(sf_context, file_acc_flags) < 0)
```
jhendersonHDF, line 664: This function's no longer needed since this PR hoists the call to `record_fid_to_subfile` up here so that we increment the file ref. count on the context object when it gets reused by an extra open of a file. The only other thing the old function did was call `ioc_open_files` which was also hoisted up to here.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -1799,6 +1839,8 @@ init_subfiling_context(subfiling_context_t *sf_context, const char *base_filenam
     assert(MPI_COMM_NULL != file_comm);
 
     sf_context->h5_file_id      = file_id;
+    sf_context->threads_inited  = false;
+    sf_context->file_ref        = 0;
```
jhendersonHDF, line 1848: Start with a ref. of 0 since the context won't have an associated file until a bit later.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -2088,24 +2082,40 @@ record_fid_to_subfile(uint64_t file_id, int64_t subfile_context_id, int *next_in
...
  *
- * Return:      None
- * Errors:      Cannot fail.
+ * Return:      Non-negative on success/Negative on failure
  *
  *-------------------------------------------------------------------------
  */
-static void
+static herr_t
 clear_fid_map_entry(uint64_t file_id, int64_t sf_context_id)
```
jhendersonHDF, line 2113: Rewrote this function a bit for clarity and added a check of the context's file reference count before clearing a mapping entry for a file to make sure we only do that when the last file hold a reference to the context is closed.

testpar/t_vfd.c
```diff
@@ -29,9 +29,9 @@
 static MPI_Comm comm = MPI_COMM_WORLD;
 static MPI_Info info = MPI_INFO_NULL;
 
-bool        pass               = true; /* set to false on error */
-bool        disp_failure_mssgs = true; /* global force display of failure messages */
-const char *failure_mssg       = NULL;
+static bool        pass               = true; /* set to false on error */
+static bool        disp_failure_mssgs = true; /* global force display of failure messages */
+static const char *failure_mssg       = NULL;
```
jhendersonHDF, line 34: Added static here since AddressSanitizer complains about these being defined in `test/cache_common.h` as well.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -2088,24 +2082,40 @@ record_fid_to_subfile(uint64_t file_id, int64_t subfile_context_id, int *next_in
...
+        /* Only clear map entry if this is the last file
+         * holding a reference to the context
+         */
+        if (NULL == (sf_context = H5_get_subfiling_object(sf_context_id)))
+            H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTGET, FAIL, "couldn't get subfiling context object");
+
+        if (sf_context->file_ref == 0 || sf_context->file_ref == 1) {
+            sf_open_file_map[i].file_id       = UINT64_MAX;
+            sf_open_file_map[i].sf_context_id = -1;
         }
```
bmribler, line 2135: So, this could potentially happen twice, when file_ref ==1 then later file_ref == 0?  Just curious...

jhendersonHDF: In the current code with this PR, this function's only called in two places: either when opening a file fails where the object will have a ref. of 0 or during file close when the object will have a ref. of 1 and be freed shortly after calling this function. Code changes in the future could change things, but in general we shouldn't ever arrive here twice for the same context object. Usually you'd start a reference count on an object at 1, but dealing with proper reference counting in failure states can then be difficult.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -188,23 +187,27 @@ H5_get_subfiling_object(int64_t object_id)
...
-            ret_value = sf_context_cache[obj_index];
-        else {
-            assert(!sf_context_cache[sf_context_cache_num_entries]);
+        ret_value = sf_context_cache[obj_index];
+        if (!ret_value) {
+            size_t next_idx;
 
             /* Allocate a new subfiling context object */
-            if (NULL == (ret_value = calloc(1, sizeof(subfiling_context_t))))
+            if (NULL == (sf_context_cache[obj_index] = calloc(1, sizeof(subfiling_context_t))))
```
bmribler, line 196: What's the benefit of swapping out ret_value here?

jhendersonHDF: No benefit, it's just a preference. I prefer to assign the return value of malloc/calloc to something first, then set ret_value rather than the other way around.

bmribler (approved):

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -648,21 +657,40 @@ H5_open_subfiles(const char *base_filename, uint64_t file_id, H5FD_subfiling_par
     if (!context_id_out)
         H5_SUBFILING_GOTO_ERROR(H5E_FILE, H5E_CANTOPENFILE, FAIL, "invalid subfiling context ID pointer");
 
-    /* Initialize new subfiling context ID based on configuration information */
-    if (init_subfiling(base_filename, file_id, subfiling_config, file_acc_flags, file_comm, &context_id) < 0)
-        H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL, "couldn't initialize subfiling context");
+    /* Check if this file is already open */
+    H5E_BEGIN_TRY
+    {
+        context_id = H5_subfile_fid_to_context(file_id);
```
qkoziol: Could you switch this routine to pass a pointer to the context_id as a parameter, to leave the return value solely for errors, instead of overloading the return value with both?   Another "code smell" that I've been gradually cleaning up when I get a chance, and it would be good to avoid spreading it further here.  :-)   (Ideally, we can get all internal routines away from the return value overloading)

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -2074,8 +2063,13 @@ record_fid_to_subfile(uint64_t file_id, int64_t subfile_context_id, int *next_in
             *next_index = index;
         }
 
-        sf_open_file_map[index].file_id         = file_id;
-        sf_open_file_map[index++].sf_context_id = subfile_context_id;
+        sf_open_file_map[index].file_id       = file_id;
+        sf_open_file_map[index].sf_context_id = subfile_context_id;
```
qkoziol, line 2090: Was the previous 'index++' a bug?

src/H5FDsubfiling/H5subfiling_common.c
jhendersonHDF: Sure, I can do that. Agreed that return values should generally be reserved for errors.

src/H5FDsubfiling/H5subfiling_common.c
jhendersonHDF: It was just unnecessary since it's at the end of the function. Likely I copy-pasted from a loop at some point.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -648,21 +658,40 @@ H5_open_subfiles(const char *base_filename, uint64_t file_id, H5FD_subfiling_par
     if (!context_id_out)
         H5_SUBFILING_GOTO_ERROR(H5E_FILE, H5E_CANTOPENFILE, FAIL, "invalid subfiling context ID pointer");
 
-    /* Initialize new subfiling context ID based on configuration information */
-    if (init_subfiling(base_filename, file_id, subfiling_config, file_acc_flags, file_comm, &context_id) < 0)
-        H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL, "couldn't initialize subfiling context");
+    /* Check if this file is already open */
+    H5E_BEGIN_TRY
+    {
+        status = H5_subfile_fid_to_context(file_id, &context_id);
```
qkoziol: I would think that the H5E_BEGIN_TRY/H5E_END_TRY would no longer be necessary, now that only "real" errors are thrown from this routine.  Can this be switched to just a normal:

if (func() < 0)
   <goto error>

src/H5FDsubfiling/H5subfiling_common.c
qkoziol: Cool, thanks.  (See comment below about the H5E_BEGIN_TRY/H5E_END_TRY)

src/H5FDsubfiling/H5subfiling_common.c
jhendersonHDF: If it's the very first file open with the VFD, the opened file map won't be allocated yet and cause an error. Not a great design but the allocation of the map resides in the code that writes to it, which made decent sense at the time. Probably worth a slight refactoring.

qkoziol: > If it's the very first file open with the VFD, the opened file map won't be allocated yet and cause an error. Not a great design but the allocation of the map resides in the code that writes to it, which made decent sense at the time. Probably worth a slight refactoring.

Yeah, I see that now.   Could you fix it by extracting this block from record_fid_to_subfile() into a separate helper routine, called by that routine and H5_subfile_fid_to_context():

```
   if (!sf_open_file_map) { 
        if (NULL == (sf_open_file_map = malloc((size_t)DEFAULT_FILE_MAP_ENTRIES *
            H5_SUBFILING_GOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "couldn't

        sf_file_map_size = DEFAULT_FILE_MAP_ENTRIES;
        for (int i = 0; i < sf_file_map_size; i++) {
            sf_open_file_map[i].file_id       = UINT64_MAX;
            sf_open_file_map[i].sf_context_id = -1;
        }
    }
```

Because, if it's not an error, then H5_subfile_fid_to_context() should not return an error.  :-)

qkoziol (approved):

PR #4204, opened 2024-03-21
Suggested header footer for NEWSLETTER

byrnHDF:


mattjala (approved):

hyoklee (approved):

release_docs/NEWSLETTER.txt
```diff
@@ -1,3 +1,8 @@
+Release of HDF5 1.14.4 Library and Tools is now available from the HDF5 Releases page.
```
jhendersonHDF, line 1: Is there a header and footer in the NEWSLETTER file for HDF4? This seems like just another thing that has to be updated with new release version branches.

byrnHDF: This copies what is on the NEWSLETTER website.

byrnHDF: It is simple to add it to the list of files to change in the vers script.

jhendersonHDF: I figured the idea was to have a file that's empty other than the entries just so there's a quick reference to pull from. But, no complaints about the header and footer; just wanting to avoid putting version numbers anywhere they aren't exactly needed since it just adds to the maintenance burden. If `h5vers` can handle it though, great!

jhendersonHDF (approved):

PR #4205, opened 2024-03-21
Use tar.gz extension for ABI reports

hyoklee:
Users can't open `.reports` file easily without proper file extension after downloading it.


bmribler (approved):

byrnHDF (changes requested):
Extension needs added in the release-files and remove-files yaml files too

byrnHDF (approved):

derobins (approved):

glennsong09 (approved):

PR #4209, opened 2024-03-21
Addressed Fortran  issues with promoted integers and reals via compilation flags

brtnfld:
(1)  change builds so the test programs that detect the native integer and reals are compiled with user-supplied FCFLAGS.
(2)  Correct passing of the correct integer or real datatypes in the source and tests

Should address issues with xlf compilers with -q32, -q64, with and without -qrealsize=8 -qintsize=8


brtnfld:
It passes fortran tests for mpich build both with and without  -fdefault-integer-8 -fdefault-real-8

byrnHDF (approved):

epourmal (approved):
Looks like very good changes to me.

lrknox (approved):

PR #4211, opened 2024-03-22
Fix 'make check-vfd' target for Autotools

jhendersonHDF:
Changes Autotools testing to use HDF5_TEST_DRIVER environment variable to avoid running tests that don't work well with several VFDs

Restores old h5_get_vfd_fapl() testing function to setup a FAPL with a particular VFD


config/cmake/HDF5Macros.cmake
```diff
@@ -70,7 +70,13 @@ macro (H5_SET_VFD_LIST)
       split
       multi
       family
-      splitter
+      # Splitter VFD currently can't be tested with the h5_fileaccess()
```
jhendersonHDF, line 73: The splitter VFD was previously being tested just with the environment variable approach, where it would add "_wo" to the end of filenames to create the name for the W/O file, which worked well in testing. With the changes in this PR though, any tests that use `h5_fileaccess()` would setup a FAPL with a single W/O path that gets used for any file create or open calls that use that FAPL, leading to failures due to multiple locks on the same file. We should revisit this at some point.

config/cmake/HDF5Macros.cmake
```diff
@@ -82,16 +88,21 @@ macro (H5_SET_VFD_LIST)
     # list (APPEND VFD_LIST mpio)
   endif ()
   if (H5_HAVE_MIRROR_VFD)
-    list (APPEND VFD_LIST mirror)
+    # Mirror VFD needs network configuration, etc. and isn't easy to set
+    # reasonable defaults for that info.
```
jhendersonHDF, line 95: Matching what's in Autotools

src/H5FDtest.c
```diff
@@ -89,7 +89,10 @@ H5FD__supports_swmr_test(const char *vfd_name)
 
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
-    if (!vfd_name || !strcmp(vfd_name, "") || !strcmp(vfd_name, "nomatch"))
```
jhendersonHDF, line 92: No need for the "nomatch" case. It was only checked in a few places where we can just as easily return "sec2" for the same behavior.

test/CMakeVFDTests.cmake
```diff
@@ -71,6 +71,9 @@ add_custom_target(HDF5_VFDTEST_LIB_files ALL COMMENT "Copying files needed by HD
       links_env
       external_env
       vds_env
+      mirror_vfd
+      ros3
+      hdfs
```
jhendersonHDF, line 76: Exclude VFD-specific tests from CMake's VFD testing, as it often doesn't make sense to run these tests with other VFDs.

test/h5test.c
```diff
@@ -772,6 +784,242 @@ h5_fileaccess_flags(unsigned flags)
...
+ * Purpose:     Sets the file driver for a FAPL according to the value
+ *              specified in the constant "HDF5_DRIVER" or environment
+ *              variable "HDF5_DRIVER" or "HDF5_TEST_DRIVER".
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+h5_get_vfd_fapl(hid_t fapl)
```
jhendersonHDF, line 794: Restored old `h5_get_vfd_fapl()` function and added a few VFDs to it.

byrnHDF (approved):

test/h5test.c
```diff
@@ -772,6 +779,241 @@ h5_fileaccess_flags(unsigned flags)
     return H5I_INVALID_HID;
 } /* end h5_fileaccess_flags() */
 
+/*-------------------------------------------------------------------------
+ * Function:    h5_get_vfd_fapl
+ *
+ * Purpose:     Sets the file driver for a FAPL according to the value
+ *              specified in the constant "HDF5_DRIVER" or environment
```
qkoziol: Is this still supposed to listen to the HDF5_DRIVER macro?

test/h5test.c
```diff
@@ -2193,12 +2435,38 @@ h5_using_native_vol(hid_t fapl_id, hid_t obj_id, bool *is_native_vol)
...
+ *-------------------------------------------------------------------------
+ */
+const char *
+h5_get_test_driver_name(void)
+{
+    char *envval;
+
+    assert(H5_DEFAULT_VFD == H5FD_SEC2);
+
+    if ((envval = getenv(HDF5_DRIVER)))
```
qkoziol, line 2455: This seems to be mixing the HDF5_DRIVER source macro and the environment variable "HDF5_DRIVER".

test/h5test.c
jhendersonHDF: No I missed updating that comment

test/h5test.c
```diff
@@ -2212,10 +2480,10 @@ h5_using_default_driver(const char *drv_name)
     assert(H5_DEFAULT_VFD == H5FD_SEC2);
 
     if (!drv_name)
-        drv_name = getenv(HDF5_DRIVER);
+        drv_name = h5_get_test_driver_name();
 
     if (drv_name)
-        return (!strcmp(drv_name, "sec2") || !strcmp(drv_name, "nomatch"));
+        return !strcmp(drv_name, "sec2");
```
qkoziol: It would be nice if this used the same symbol as the library to determine the default VFD, instead of hard-coding the "sec2" string here.

test/h5test.c
jhendersonHDF: On advice from someone else, I started defining macros that map to the name of the environment variable for maintenance. `HDF5_DRIVER` = "HDF5_DRIVER". We haven't really had an `HDF5_DRIVER` source macro for a while and I don't think we ever tested that that approach was working correctly.

test/h5test.c
jhendersonHDF: Seems easy enough to add a `H5_DEFAULT_VFD_NAME` macro if that's what you mean

test/h5test.c
qkoziol: I agree - I think the HDF5_DRIVER macro should be removed.  That would clean up the code in this area quite a bit.

qkoziol: If it's your new macro, you should have one for the HDF5_TEST_DRIVER environment variable also.

qkoziol: I would also suggest putting 'ENV' somewhere in their names, so that it's a little more clear what their purpose is.

qkoziol:
Generally looking good, especially if we can resolve the HDF5_DRIVER / "HDF5_TEST_DRIVER" asymmetry.

lrknox (approved):

test/h5test.c
jhendersonHDF: I'm not sure I understand. As far as I remember, `HDF5_DRIVER` never existed anywhere in the library or configure code until it was added as `HDF5_DRIVER` = "HDF5_DRIVER". One could have passed it as a compile option, but I don't think it was documented and we definitely didn't have a test for that at the time I worked on this previously. So it seems there's not much to remove here other than the one `#ifdef HDF5_DRIVER` I removed in this PR. `HDF5_DRIVER` is also in H5public.h so I'd prefer not to change it, but open to any discussion there.

No problem adding a macro for "HDF5_TEST_DRIVER", though my intention for this is for it to be a temporary environment variable, so that one would probably go in a private header somewhere. Using "HDF5_TEST_DRIVER" is just a way for CMake and Autotools to play nicely together since Autotools doesn't pick and choose which tests to run with all the VFDs and `HDF5_DRIVER` works similar to how `HDF5_VOL_CONNECTOR` does for setting something on a FAPL. Because of that, `make check-vfd` causes Autotools to run every single test with all the VFDs and fails in a lot of places while CMake is conservative about what it tests. It was an oversight at the time I implemented the VFD plugins feature.

jhendersonHDF: I should clarify that by temporary, I mean that in the future we need to refactor what tests get run for VFD testing so Autotools can test more things. But until then, "HDF5_TEST_DRIVER" is a stopgap that lets CMake keep generally testing the same things it was with VFDs, while preventing Autotools from running too many tests against VFDs.

test/h5test.c
jhendersonHDF: Added this and updated a few occurrences that seemed reasonable to change.

jhendersonHDF:
@qkoziol Merging this so that we have VFD testing back in a working state, but happy to follow up on any discussion here in a separate PR if needed.

qkoziol:
> @qkoziol Merging this so that we have VFD testing back in a working state, but happy to follow up on any discussion here in a separate PR if needed.

Sorry, couldn't get back to this yesterday - yes, I understand your comments and agree that the way you have it now (as a temporary solution, hopefully) is a good way forward.

PR #4218, opened 2024-03-22
Fix problems with background buffers and array datatypes

fortnern:
This patch addresses two problems with array datatype conversions:

1. The conversion code did not advance in the background buffer when converting each constituent element. Addressed this.
2. The conversion code always reported, in the INIT phase, that it did not require a background buffer, even if the parent did require one. While in the CONV phase the converison routine would initialize the parent conversion path and allocate a background buffer if necessary, it did not fill the background buffer with the destination contents. Changed this code to initialize the parent conversion in the INIT phase, report the background buffer necessity value from that path, and save a pointer to the conversion path for use in the CONV phase.

No testing yet, will add Monday


fortnern:
I'm able to reproduce the h5dumpgentest failure and I'll look into it Monday

jhendersonHDF (approved):

qkoziol (approved):

PR #4222, opened 2024-03-23
Fix an issue with type size assumptions in h5dumpgentest

jhendersonHDF:


tools/test/h5dump/h5dumpgentest.c
jhendersonHDF, line 1: This test uses the STD fixed-width datatypes for compound fields while assuming the sizes chosen match up with 16-bit short, 32-bit int and 64-bit long

lrknox (approved):

byrnHDF (approved):

PR #4223, opened 2024-03-23
Fix issue with -Werror cleanup sed command in configure.ac

jhendersonHDF:


jhendersonHDF:
This sed command in configure.ac is generally failing and causing passed in CFLAGS to be wiped out, which messes up several other things further down in configure.ac.

configure.ac
```diff
@@ -198,21 +198,36 @@ saved_user_CPPFLAGS="$CPPFLAGS"
...
 ##
-WERROR_SED= "sed -e 's/-Werror\(=[^[:space:]]\+\)\?//g'"
-
-CFLAGS="`echo $CFLAGS | $WERROR_SED`"
-CXXFLAGS="`echo $CXXFLAGS | $WERROR_SED`"
-FCFLAGS="`echo $FCFLAGS | $WERROR_SED`"
-JAVACFLAGS="`echo $JAVACFLAGS | $WERROR_SED`"
-CPPFLAGS="`echo $CPPFLAGS | $WERROR_SED`"
+## Note that the outer pair of '[]' ends up getting removed
+WERROR_SED='s/-Werror\(=[[^[:space:]]]\+\)\?//g'
```
jhendersonHDF, line 210: This was about the only reasonable way I could get this to work with the sed string inside a variable. Also, regardless of whether this is single or double-quoted, something appears to always be eating the outer pair of `[]`, so this adds an extra pair to be eaten. You can see this in any of the autotools runners, for example:

`/home/runner/work/hdf5/hdf5/configure: line 4807: sed -e 's/-Werror\(=^[:space:]\+\)\?//g': No such file or directory`

I also had to move the sed command out of this string to make things work.

configure.ac
```diff
@@ -198,21 +198,36 @@ saved_user_CPPFLAGS="$CPPFLAGS"
...
+CFLAGS_SED="`echo $CFLAGS | sed -e $WERROR_SED`"
+if test $? -eq 0; then
```
jhendersonHDF, line 212: Check to make sure the sed command didn't fail before assigning the result back to `CFLAGS`, etc. This may cause "-Werror" to not be removed on some systems if they end up not eating the extra outer pair of `[]`, but at least it won't wipe out `CFLAGS` and others

lrknox (approved):

hyoklee (approved):
This is a good fix for RISC-V Autotools.

PR #4225, opened 2024-03-24
Rewrite H5T__path_find_real for clarity

jhendersonHDF:


jhendersonHDF:
I personally find `H5T__path_find_real` a bit hard to parse through at times, so I rewrote it a bit in a way that I believe is a lot easier to work through.

src/H5T.c
```diff
@@ -5056,6 +5065,60 @@ H5T_cmp(const H5T_t *dt1, const H5T_t *dt2, bool superset)
...
+ *             sure to increment the index value by 1 if the value of the
+ *             last comparison is > 0.
+ *
+ * Return:    Success:    Pointer to the path in the path table
+ *            Failure:    NULL if no matching path is found in the table
+ *
+ *-------------------------------------------------------------------------
+ */
+static void *
+H5T__bsearch_path_table(const H5T_t *src, const H5T_t *dst, int *last_cmp, int *idx)
```
jhendersonHDF, line 5003: Moved the binary search logic into this helper function. I wanted to just implement this using `bsearch`, but its interface is a bit awkward to use for the things we're interested in knowing about like the last comparison value and the index into the table where we landed.

src/H5T.c
```diff
@@ -5157,80 +5223,40 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
     if (H5CX_pushed() && (H5CX_get_dt_conv_cb(&tmp_ctx.u.init.cb_struct) < 0))
         HGOTO_ERROR(H5E_DATATYPE, H5E_CANTGET, NULL, "unable to get conversion exception callback");
 
-    /*
-     * Make sure the first entry in the table is the no-op conversion path.
-     */
-    if (0 == H5T_g.npaths) {
```
jhendersonHDF, line 5163: Moved the allocation and initialization of the path table into a helper function

src/H5T.c
```diff
@@ -5245,67 +5271,245 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
             HGOTO_ERROR(H5E_DATATYPE, H5E_CANTINIT, NULL, "unable to copy datatype for conversion path");
     } /* end if */
     else
-        path = table;
+        path = matched_path;
+
+    /* Initialize the path if it's a new path */
+    if (new_path && H5T__path_find_init_new_path(path, src, dst, conv, &tmp_ctx) < 0)
```
jhendersonHDF, line 5190: Moved the two large sections that initialize a new path with either the specified hard conversion function or an existing soft function into the new `H5T__path_find_init_new_path` helper function, making the purpose of those sections clearer and localizing the temporary datatypes and IDs for those types used when initializing the conversion function

src/H5T.c
```diff
@@ -5157,80 +5223,40 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
...
 
     /* Keep a record of the number of paths in the table, in case one of the
      * initialization calls below (hard or soft) causes more entries to be
      * added to the table - QAK, 1/26/02
      */
     old_npaths = H5T_g.npaths;
 
+    /* Set a few convenience variables */
+    new_api_hard_func = (matched_path && conv->is_app && conv->u.app_func);
+    new_lib_hard_func = (matched_path && !conv->is_app && conv->u.lib_func);
```
qkoziol: Suggest using 'new_lib_func' for a variable name, since it's not guaranteed to be a 'hard' conversion.    (Else, we wouldn't need the '!matched->is_hard' test, below)

src/H5T.c
```diff
@@ -5245,67 +5271,245 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
...
+
+    /* Replace an existing table entry or add a new entry */
+    if (matched_path && new_path) {
+        herr_t status;
+        int    nprint = 0;
+
+        assert(matched_path == H5T_g.path[path_idx]);
+        H5T__print_stats(matched_path, &nprint);
+
+        /* Free the old conversion path table entry */
```
qkoziol: This block of code, to free a conversion path entry, is replicated at least 4 times in this file (search for H5T_CONV_FREE) and would be good to move to a helper routine also.

qkoziol (approved):
Solid improvement, even without the suggestions I made.

src/H5T.c
jhendersonHDF: In the current implementation I think it is guaranteed, in that `H5T__path_find_real` is currently only called in two places in the library: from the `H5T_path_find` function, in which case `app_func/lib_func` is always NULL, and from `H5T__register`, in which case `H5T__path_find_real` is only called when the passed in function is specified as `H5T_PERS_HARD`. That said, seems this could change in the future, so no problem renaming these to just "new_xxx_func".

byrnHDF (approved):

PR #4227, opened 2024-03-24
Rework snapshots/release workflows for consistent args

byrnHDF:


.github/workflows/cmake-ctest.yml
```diff
@@ -46,11 +51,11 @@ jobs:
...
           FILE_NAME_BASE=$(echo "${{ inputs.file_base }}")
           echo "FILE_BASE=$FILE_NAME_BASE" >> $GITHUB_OUTPUT
-          if [[ '${{ inputs.use_environ }}' == 'snapshots' ]]
+          if [[ '${{ inputs.use_environ }}' == 'release' ]]
           then
-            SOURCE_NAME_BASE=$(echo "hdfsrc")
+            SOURCE_NAME_BASE=$(echo "${{ inputs.snap_name }}")
           else
-            SOURCE_NAME_BASE=$(echo "$FILE_NAME_BASE")
+            SOURCE_NAME_BASE=$(echo "hdfsrc")
```
jhendersonHDF, line 58: Since you set the default for "snap_name" to be hdfsrc above, can't you always just use `SOURCE_NAME_BASE=$(echo "${{ inputs.snap_name }}")` rather than needing the conditionals here?

byrnHDF: Possibly - that is likely what I was thinking - but after fighting this all weekend I didn't want to break it. At least until after the release.

jhendersonHDF (approved):

derobins (approved):

PR #4228, opened 2024-03-24
Set DXPL in API context for native VOL attribute I/O calls

jhendersonHDF:


jhendersonHDF:
This fixes testing of the LOG VOL. After https://github.com/HDFGroup/hdf5/pull/4113, the library started needing a valid DXPL to be set in the API context at all times when `H5T__path_find_real` is called and an API context was pushed. Previously, attribute writes and reads in the native VOL weren't setting that.

For future consideration around similar problems that may occur, in the LOG VOL's tests, there is a complex interaction going on in the LOG VOL that's something like the following:

At some point an `H5Awrite` call is made where the LOG VOL determines it should flush file metadata that it's been buffering up internally -> The `H5Awrite` call pushes an API context and heads down to the LOG VOL -> while flushing this metadata, the LOG VOL makes a few dataset optional calls down to the native VOL corresponding to `H5Dget_offset` for internal use, passing a new DXPL ID that it creates internally for each call -> each time those optional calls are made, they call `H5CX_set_dxpl` and overwrite the DXPL ID set in the API context -> we finally reach the part of `H5Awrite` that involves the native VOL, but `H5T__path_find_real` fails because it tries to get the property list for the API context's DXPL ID, which the LOG VOL already closed between the `H5Dget_offset` calls and getting down to the native VOL.

We probably need to make sure to call `H5CX_set_dxpl` anywhere where a DXPL ID is passed in to a native VOL callback function.

qkoziol:
I don't have a chance to look at this right now, but it strikes me that the FUNC_ENTER_REENTER fix I've got in the work may address it.

jhendersonHDF:
> I don't have a chance to look at this right now, but it strikes me that the FUNC_ENTER_REENTER fix I've got in the work may address it.

Sure, I think a lot of this needs to be refactored to get a better handle on what's interacting with the API context and when, but for now this just matches what we do in the dataset I/O calls and seems like it was just missed.

derobins (approved):

qkoziol (approved):

PR #4229, opened 2024-03-24
Fix Java JNI warnings

jhendersonHDF:


jhendersonHDF:
Mostly fixes warnings around casts from `jbyte *` to another pointer type with larger alignment requirements.

java/src/jni/h5rImp.c
```diff
@@ -204,6 +204,12 @@ Java_hdf_hdf5lib_H5_H5Rdestroy(JNIEnv *env, jclass clss, jbyteArray ref)
     if ((status = H5Rdestroy(&loc_ref)) < 0)
         H5_LIBRARY_ERROR(ENVONLY);
 
+    /* Reset buffer to 0 to be safe. H5Rdestroy does this, but
+     * only on our temporary reference object that we copied
+     * into for alignment reasons.
+     */
+    memset(refBuf, 0, H5R_REF_BUF_SIZE);
```
jhendersonHDF, line 211: @byrnHDF This PR may be related to the issue you've seen with references inside container types. After fixing the logic in nativeData.c to use pointers with the right alignment, I got a test failure in the Java region reference example from HDF5Examples due to a double free from `H5Rdestroy`.

This memset does what `H5Rdestroy` does above, except the call above was only doing it on the temporary reference variable that gets memcpy'd into, so we weren't clearing the passed in reference buffer after `H5Rdestroy`. Not a big deal as long as you don't try to call `H5Rdestroy` on the same buffer twice in Java, but the fact the example passed before tells me we were likely working with incorrectly-aligned pointers in this case.

java/src/jni/h5util.c
```diff
@@ -82,10 +82,10 @@ void          translate_atomic_wbuf(JNIEnv *env, jobject in_obj, jlong mem_type_
                                     void *raw_buf);
 
 /* Strings for output */
-#define H5_TOOLS_GROUP     "GROUP"
-#define H5_TOOLS_DATASET   "DATASET"
-#define H5_TOOLS_DATATYPE  "DATATYPE"
-#define H5_TOOLS_ATTRIBUTE "ATTRIBUTE"
```
jhendersonHDF, line 88: ATTRIBUTE is currently never used.

test/h5test.c
jhendersonHDF, line 1: Miscellaneous warning fix in this file

java/src/jni/nativeData.c
```diff
@@ -93,71 +92,18 @@ Java_hdf_hdf5lib_HDFNativeData_byteToInt___3B(JNIEnv *env, jclass clss, jbyteArr
     return rarray;
 } /* end Java_hdf_hdf5lib_HDFNativeData_byteToInt___3B */
 
-/* returns float [] */
-JNIEXPORT jfloatArray JNICALL
-Java_hdf_hdf5lib_HDFNativeData_byteToFloat___3B(JNIEnv *env, jclass clss,
```
jhendersonHDF, line 98: Fixed this function and moved it down next to byteToDouble

jhendersonHDF:
CI is currently failing in Windows actions because https://community.chocolatey.org/api/v2/ is down.

byrnHDF (approved):
Great!

derobins (approved):

derobins (approved):

PR #4230, opened 2024-03-24
Update Windows CI to not install ninja

jhendersonHDF:
We aren't using ninja in the workflow and it's causing CI actions to fail since https://community.chocolatey.org/api/v2/ is currently down. The OneAPI actions do use Ninja however, so those will likely still fail for now.

### DIFF
--- a/.github/workflows/main-cmake.yml
+++ b/.github/workflows/main-cmake.yml
@@ -135,10 +135,6 @@ jobs:
 
       # CMake gets libaec from fetchcontent
 
-      - name: Install Dependencies (Windows)
-        run: choco install ninja
-        if: matrix.os == 'windows-latest'
-
       - name: Install Dependencies (macOS)
         run: brew install ninja
         if: matrix.os == 'macos-13'


.github/workflows/main-cmake.yml
```diff
@@ -135,10 +135,6 @@ jobs:
 
       # CMake gets libaec from fetchcontent
 
-      - name: Install Dependencies (Windows)
```
byrnHDF, line 138: Maybe - Ninja will be needed if we want to use clang (like OneAPI, it is needed for alternative compilers on windows).  Note that ninja is required for the binary tests.

jhendersonHDF: I think we need to be cautious about what checks are required for PRs if they're going to go out and grab things from the internet outside of what's the default on runners. Testing alternative compilers on Windows is great, but probably doesn't need to be a required check.

byrnHDF: Clang and OneAPI on windows are just as important as those compilers for linusx, right?
Unfortunately GH actions for testing are problematic at any time and chocaoately isn't necessarily worse then anything else being used.
The more important metric is time for PRs S.B. less than 25 mins for a workflow.

byrnHDF: Maybe main-cmake.yml would be better named main-cmake-vs.yml or vs-cmake.yml?

jhendersonHDF: At the very least we should be able to keep a cached installation of ninja and other required things around when we've gotten them once. It also doesn't seem like Ninja is used anywhere for the binary tests. It seems like just the OneAPI Windows action is using it.

jhendersonHDF: We may also look at https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html. Seems like we can use clang and possibly OneAPI as well with VS generators so that we don't need ninja in the first place.

byrnHDF: bintest definitely uses ninja as the generator.

lrknox (approved):

byrnHDF (approved):

Issue #3762, opened 2023-10-24
H5Centry.c:1232: H5C__load_entry: Assertion `entry->size < H5C_MAX_ENTRY_SIZE' failed.


gheber:
Creating a large number of small datasets fails with the assertion:
```
H5Centry.c:1232: H5C__load_entry: Assertion `entry->size < H5C_MAX_ENTRY_SIZE' failed.
```
The error occurs with the current development branch under Debian 12 x86_64 with GCC 13.2.0.

A reproducer is shown below:
```
#include "hdf5.h"

int main()
{
    hid_t file = H5Fcreate("why.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
    hsize_t dims[3] = {18, 18, 15};
    hid_t fspace = H5Screate_simple(3, dims, NULL);

    float data[18][18][15];
    for (size_t i = 0; i < 18; i++)
    {
        for (size_t j = 0; j < 18; j++)
        {
            for (size_t k = 0; k < 15; k++)
            {
                data[i][j][k] = (float) (i * j * k);
            }
        }
    }

#ifdef USE_SUB_GROUP
    hid_t group = H5Gcreate(file, "group", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
#endif

    char name[20];
    /*
     * 1441790 -> file created successfully, h5stat is happy
     *
     * 1441791 -> file created successfully, h5stat barfs:
     *            h5stat: H5Centry.c:1232: H5C__load_entry: Assertion `entry->size < H5C_MAX_ENTRY_SIZE' failed.
     * 
     * 1441792 -> dataset creation fails with:
     *            a.out:  H5Centry.c:1232: H5C__load_entry: Assertion `entry->size < H5C_MAX_ENTRY_SIZE' failed.
     */
    for (size_t i = 0; i < 1441791; ++i)
    {
        sprintf(name, "data%06lu", i);

        // Using a subgroup doesn't save our bacon. Same error behavior.
#ifdef USE_SUB_GROUP
        hid_t dset = H5Dcreate(group, name, H5T_NATIVE_FLOAT, fspace, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
#else        
        hid_t dset = H5Dcreate(file, name, H5T_NATIVE_FLOAT, fspace, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
#endif
[...]

PR #4231, opened 2024-03-25
Fixed a cache assert with too-large metadata objects

derobins:
If the library tries to load a metadata object that is above the library's hard-coded limits, the size will trip an assert in debug builds. In HDF5 1.14.4, this can happen if you create a very large number of links in an old-style group that uses local heaps.

The library will now emit a normal error when it tries to load a metadata object that is too large.

Partially addresses GitHub #3762

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -693,6 +693,18 @@ Bug Fixes since HDF5-1.14.0 release
 
     Library
     -------
+    - Fixed a cache assert with too-large metadata objects
+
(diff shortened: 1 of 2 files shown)


derobins:
#3762 should stay open until we fix the underlying problem (hopefully in 1.14.5). In the meantime, this fix will prevent bad behavior for that bug as well as any future size oversights.

qkoziol (approved):

hyoklee (approved):

PR #4234, opened 2024-03-25
File format security issues

derobins:
Addresses file format security issues detected via fuzzing.

Credit: Amazon Web Services and @qkoziol


derobins:
This needs RELEASE.txt info, which we'll add shortly. Let's get this review started so we can get this into develop and hdf5_1_14 ASAP.

src/H5Spoint.c
```diff
@@ -576,8 +576,14 @@ H5S__point_add(H5S_t *space, H5S_seloper_t op, size_t num_elem, const hsize_t *c
...
 
+        /* The following allocation relies on the size of an hcoords_t being
+         * the same as an 'H5S_pnt_node_t *', so fail now if that's not true
+         */
+        HDcompile_assert(sizeof(hcoords_t) >= sizeof(H5S_pnt_node_t *));
+
         /* Allocate space for the new node */
-        if (NULL == (new_node = (H5S_pnt_node_t *)H5FL_ARR_MALLOC(hcoords_t, space->extent.rank)))
+        /* Note: allocating "rank + 1" to allow for 'next' pointer */
+        if (NULL == (new_node = (H5S_pnt_node_t *)H5FL_ARR_MALLOC(hcoords_t, space->extent.rank + 1)))
```
mattjala, line 586: How is this allocation not causing constant memory errors in the current version of the library due to being too small?

jhendersonHDF: It's likely due to memory that was already allocated on the free list being reused here

derobins:
I have a set of files to add to https://github.com/HDFGroup/cve_hdf5, which I will do after we merge these changes to hdf5_1_14. If I do it now, the CI will break.

src/H5Odtype.c
```diff
@@ -452,6 +487,13 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
                 }
                 if (temp_type->shared->size == 0)
                     HGOTO_ERROR(H5E_DATATYPE, H5E_CANTDECODE, FAIL, "type size can't be zero");
+                if ((dt->shared->u.compnd.memb[dt->shared->u.compnd.nmembs].offset +
```
jhendersonHDF, line 492: Seems like this should be `dt->shared->u.compnd.nmembs - 1`

src/H5Olinfo.c
```diff
@@ -140,6 +140,9 @@ H5O__linfo_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUS
         if (H5_IS_BUFFER_OVERFLOW(p, 8, p_end))
             HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
         INT64DECODE(p, linfo->max_corder);
+        if (linfo->max_corder < 0)
+            HGOTO_ERROR(H5E_OHDR, H5E_BADVALUE, NULL, "invalid max creation order value for message: %lld",
+                        (long long)linfo->max_corder);
```
jhendersonHDF: Can be printed with `message: %" PRId64` here rather than casting

mattjala (approved):

test/tmisc.c
```diff
@@ -337,9 +337,19 @@ typedef struct {
...
 
-#define MISC38_FILE "type_conversion_path_table_issue.h5"
-#define MISC39_FILE "set_est_link_info.h5"
-#define MISC40_FILE "obj_props_intermediate.h5"
+/* Definitions for misc. test #38 */
+#define MISC38A_FILE     "tmisc38a.h5"
+#define MISC38A_DSETNAME "Fletcher_float_data_be"
+#define MISC38B_FILE     "tmisc38b.h5"
+#define MISC38B_DSETNAME "unusual_datatype"
+#define MISC38C_FILE     "tmisc38b.h5"
```
jhendersonHDF: Should probably use "tmisc38c.h5" for this filename just to avoid any test weirdness in the future

byrnHDF: Oh, it will overwrite the B file and not create a C, right?

tools/test/h5dump/expected/tnbit.ddl
```diff
@@ -1,10 +1,10 @@
 HDF5 "tfilters.h5" {
 DATASET "nbit" {
-   DATATYPE  32-bit little-endian integer 3-bit precision
+   DATATYPE  32-bit little-endian integer 17-bit precision
    DATASPACE  SIMPLE { ( 20, 10 ) / ( 20, 10 ) }
    STORAGE_LAYOUT {
       CHUNKED ( 10, 5 )
-      SIZE XXXX (10.XXX:1 COMPRESSION)
```
jhendersonHDF, line 7: Likely still want to leave the size and compression ratio masked out here

test/tmisc.c
jhendersonHDF: Looks like it from the test

src/H5Pfapl.c
```diff
@@ -6264,3 +6276,72 @@ H5P__facc_vol_close(const char H5_ATTR_UNUSED *name, size_t H5_ATTR_UNUSED size,
...
+H5Pget_relax_file_integrity_checks(hid_t plist_id, uint64_t *flags /*out*/)
+{
+    H5P_genplist_t *plist;               /* Property list pointer */
+    herr_t          ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_API(FAIL)
+    H5TRACE2("e", "i*UL", plist_id, flags);
+
+    /* Get the property list structure */
+    if (NULL == (plist = H5P_object_verify(plist_id, H5P_FILE_ACCESS)))
```
jhendersonHDF, line 6340: Not a necessity, but it'd be nice to account for `H5P_DEFAULT` here

src/H5Ppublic.h
```diff
@@ -5634,6 +5634,79 @@ H5_DLL herr_t H5Pset_mdc_image_config(hid_t plist_id, H5AC_cache_image_config_t
...
+ *          files created with these configurations.
+ *
+ *          A more complete solution that avoids errors on both object creation
+ *          and access is to use the H5Pset_libver_bounds routine with a low
+ *          bound of at least #H5F_LIBVER_V18 when creating objects with these
+ *          configurations.  This will cause the library to checksum a file's
+ *          metadata, allowing accidental data corruption to be correctly
+ *          detected and errors correctly issued without ambiguity.
+ *
+ * \since 1.14.3
```
jhendersonHDF: 1.14.4

src/H5Ppublic.h
```diff
@@ -5634,6 +5634,79 @@ H5_DLL herr_t H5Pset_mdc_image_config(hid_t plist_id, H5AC_cache_image_config_t
...
+ * \fapl_id{plist_id}
+ * \param[out] flags  Relaxed file integrity check flags
+ *
+ * \return \herr_t
+ *
+ * \details H5Pget_relax_file_integrity_checks() retrieves the relaxed file
+ *          integrity check value into \p flags for the file access property
+ *          list specified in \p plist_id.
+ *
+ * \since 1.14.3
```
jhendersonHDF: Also 1.14.4

test/tmisc.c
```diff
@@ -6264,14 +6287,230 @@ test_misc37(void)
...
+        /* Create a committed datatype with the unusual datatype */
+        H5E_BEGIN_TRY
+        {
+            ret = H5Tcommit2(fid, MISC38C_TYPENAME, tid, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+        }
+        H5E_END_TRY
+        if (u > 0) {
+            CHECK(ret, FAIL, "H5Tcommit2");
+
+            ret = H5Tclose(tid);
```
jhendersonHDF: Should move this outside the if - else block so we close `tid` even if `H5Tcommit2` fails above

qkoziol (approved):

src/H5Tcommit.c
```diff
@@ -510,23 +526,39 @@ H5T__commit(H5F_t *file, H5T_t *type, hid_t tcpl_id)
...
+            else
+                oloc_ptr = &(type->oloc);
+            if (H5O_dec_rc_by_loc(oloc_ptr) < 0)
                 HDONE_ERROR(H5E_DATATYPE, H5E_CANTDEC, FAIL,
                             "unable to decrement refcount on newly created object");
-            if (H5O_close(&(type->oloc), NULL) < 0)
+            if (H5O_close(oloc_ptr, NULL) < 0)
                 HDONE_ERROR(H5E_DATATYPE, H5E_CLOSEERROR, FAIL, "unable to release object header");
-            if (H5O_delete(file, type->sh_loc.u.loc.oh_addr) < 0)
+            if (H5O_delete(file, oloc_ptr->addr) < 0)
```
bmribler, line 547: Can oloc_ptr->addr still be used after H5O_close(oloc_ptr, NULL)?

src/H5Tconv.c
```diff
@@ -3801,6 +3818,57 @@ H5T__conv_vlen(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, const H5T
...
+                dest_count += elmtno; /* Include partial iteration in first pass, for forward conversions */
+            }
+            else
+                d = (uint8_t *)buf + (nelmts * orig_d_stride);
+
+            /* Destroy references that have already been converted */
+            while (dest_count > 0) {
+                H5_GCC_DIAG_OFF("discarded-qualifiers")
+                H5_CLANG_DIAG_OFF("incompatible-pointer-types-discards-qualifiers")
+                H5T_vlen_reclaim_elmt(d, dst); /* Ignore errors at this point */
```
jhendersonHDF, line 3839: Rather than hiding the const warnings here due to `H5T_vlen_reclaim_elmt` taking a `H5T_t *` parameter, I think we can just update `H5T_vlen_reclaim_elmt` to take a `const H5T_t *` instead. All it does is pass that parameter to `H5T__vlen_reclaim`, which takes a `const H5T_t *`. Would be much cleaner.

src/H5Tconv.c
```diff
@@ -3801,6 +3818,57 @@ H5T__conv_vlen(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, const H5T
...
+                d = (uint8_t *)buf + ((nelmts - safe) * orig_d_stride);
+            else
+                d = (uint8_t *)buf + ((nelmts - elmtno) * orig_d_stride);
+
+            /* Destroy references that have already been converted */
+            while (dest_count > 0) {
+                H5_GCC_DIAG_OFF("discarded-qualifiers")
+                H5_CLANG_DIAG_OFF("incompatible-pointer-types-discards-qualifiers")
+                H5T_vlen_reclaim_elmt(d, dst); /* Ignore errors at this point */
+                H5_GCC_DIAG_ON("discarded-qualifiers")
```
jhendersonHDF: Same comment here about `H5T_vlen_reclaim_elmt`

src/H5Tconv.c
```diff
@@ -4221,6 +4303,49 @@ H5T__conv_ref(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata,
     } /* end switch */
 
 done:
+    /* Release converted elements on error */
+    if (ret_value < 0 && conversions_made) {
```
jhendersonHDF, line 4299: After looking at this and the above side by side, it looks like the only difference is whether we call `H5T_vlen_reclaim_elmt` or `H5Rdestroy` on the element. Would be much nicer inside a helper function.

src/H5Tconv.c
```diff
@@ -3801,6 +3818,57 @@ H5T__conv_vlen(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, const H5T
...
+
+            /* Set pointer to correct location, based on direction chosen */
+            if (convert_forward) {
+                d = (uint8_t *)buf;
+                dest_count += elmtno; /* Include partial iteration in first pass, for forward conversions */
+            }
+            else
+                d = (uint8_t *)buf + (nelmts * orig_d_stride);
+
+            /* Destroy references that have already been converted */
```
jhendersonHDF: Nitpick: this is for freeing variable-length elements rather than destroying references. Same for below

src/H5private.h
```diff
@@ -338,7 +347,10 @@
 /* Check if a read of size bytes starting at ptr would overflow past
  * the last valid byte, pointed to by buffer_end.
  */
-#define H5_IS_BUFFER_OVERFLOW(ptr, size, buffer_end) (((ptr) + (size)-1) > (buffer_end))
+#define H5_IS_BUFFER_OVERFLOW(ptr, size, buffer_end)                                                         \
+    (((ptr) > (buffer_end)) ||                             /* Bad precondition */                            \
+     ((ptrdiff_t)(size) > (((buffer_end) - (ptr)) + 1)) || /* Typical overflow */                            \
+     ((intptr_t)(size) < 0))                               /* Negative 'size' would wrap 'ptr' */
```
jhendersonHDF: This is the only part of the PR I keep getting hung up on. If a large enough value is passed in for `size`, won't this be implementation-defined what happens if it overflows when casting to `intptr_t`? It seems like we need something like:

`((size <= (size_t)INTPTR_MAX) && ((intptr_t)(size) < 0))`

though, I'm not quite sure that's right either in the case where `INTPTR_MAX` is larger than `SIZE_MAX`, causing wraparound when casting `INTPTR_MAX` to `size_t`

src/H5T.c
```diff
@@ -6638,3 +6639,43 @@ H5T__get_path_table_npaths(void)
...
+ *              bitfield) with an unusual # of unused bits.  This means
+ *              that the precision (i.e. the # of bits used) is less than
+ *              the size of the datatype, at power-of-two boundaries.
+ *
+ * Return:      true/false on success, can't fail
+ *
+ *-------------------------------------------------------------------------
+ */
+bool
+H5T_is_numeric_with_unusual_unused_bits(const H5T_t *dt)
```
jhendersonHDF, line 6656: Testing on PPC64LE by @hyoklee shows that this probably won't work correctly for the IBM long double type. Suggest either:

```C
#if LDBL_MANT_DIG == 106
    /* This won't work for the IBM long double type */
    HGOTO_DONE(false);
#else
    ...
#endif
```

or some other solution, until we can maybe address issues with the datatype.

src/H5Ppublic.h
derobins: Fixed

src/H5Ppublic.h
derobins: Fixed

tools/test/h5dump/expected/tnbit.ddl
derobins: Fixed

src/H5Olinfo.c
derobins: Fixed

src/H5Tconv.c
derobins: Fixed

src/H5Tconv.c
derobins: Fixed

src/H5Tconv.c
derobins: Fixed

src/H5Tconv.c
derobins: Agree, but I'll leave this for later work. Right now I just want to get this in.

src/H5Tcommit.c
derobins: No, it points to temp_oloc, which is local and cleaned up immediately after this.

test/tmisc.c
derobins: Done

test/tmisc.c
derobins: Fixed

src/H5private.h
fortnern: So I think it's technically possible to have a valid buffer so large that ptrdiff_t wraps to negative when computing the size. This could be an issue on 32 bit systems that are pushing the limits of their architecture. So size could be negative when cast to ptrdiff_t and still be valid. I don't know if the rest of the library would handle this case correctly.

fortnern: Maybe something like `(((ptr) > (buffer_end)) || (((ptr) + (size) - 1) > (buffer_end)) || (((ptr) + (size)) < (ptr)))`

We might want to cast the pointers to `const uint8_t *` as well, and size to `size_t`.

src/H5Odtype.c
jhendersonHDF: Actually this one's fine. `dt->shared->u.compnd.nmembs` is the loop index here so this should be valid.

src/H5private.h
jhendersonHDF: Hmm yes I think this is a bit more in line with what I was thinking.

jhendersonHDF: After a little research though, it seems some compilers often remove a `ptr + len < ptr` type check since it's UB anyway.

qkoziol: I suggest getting my notes for files that tripped overflows from Dana and making certain that any change to what I worked out last year are still protecting from the related corrupted files.  Some of them were pretty obscure... :-/

jhendersonHDF: Definitely plan to test against the CVE files. I think it's mostly a matter of needing to check `size` against the max of the signed type it's being cast into first, and then casting it and checking against < 0 before then doing the pointer subtraction and comparison. Unfortunately it's difficult to do this in a nice way since the macro has no type information and it's not always a `size_t` value that's being passed in.

src/H5encode.h
```diff
@@ -224,7 +224,7 @@
         n = 0;                                                                                               \
         (p) += 8;                                                                                            \
         for (_i = 0; _i < sizeof(uint64_t); _i++)                                                            \
-            n = (n << 8) | *(--p);                                                                           \
+            n = (uint64_t)(((uint64_t)n << 8) | *(--p));                                                     \
```
jhendersonHDF, line 227: Casting this back to a `uint64_t` seems to be showing a warning:

```
conversion from ‘long unsigned int’ to ‘uint32_t’ {aka ‘unsigned int’} may change value [-Warith-conversion]
  227 |             n = (uint64_t)(((uint64_t)n << 8) | *(--p));                                                     \
      |                 ^
src/H5Olayout.c:398:33: note: in expansion of macro ‘H5F_DECODE_LENGTH’
  398 |                                 H5F_DECODE_LENGTH(f, p, mesg->storage.u.chunk.u.single.nbytes);
      |                                 ^~~~~~~~~~~~~~~~~
```

Seems like we could end up trying to UINT64DECODE into a `uint32_t` value

src/H5private.h
jhendersonHDF: Rewrote this slightly. It's basically the same thing, but using `ptrdiff_t` instead of `intptr_t`, checking against `PTRDIFF_MAX` before casting and also doing the check for negative values before the typical overflow check. Passes all current CVE tests, including the ones specifically around the changes in this PR.

Because of the "type-limits" warnings that had to be suppressed in this PR, I'm tempted to just move the macro into an inline function, but for the sake of getting these changes in, I don't think needs to happen right now.

src/H5encode.h
jhendersonHDF: Updated this to decode into a uint64_t and then checked_assign it to the uint32_t field.

jhendersonHDF:
CMake seems to be failing in the `H5DUMP-tnbit` test because the compression ratio line isn't getting masked out for some reason. Not sure on netcdf failures, but investigating.

jhendersonHDF:
Updating the `H5_IS_KNOWN_BUFFER_OVERFLOW` macro to use `H5_IS_BUFFER_OVERFLOW` when `skip` is false caused the netCDF tests to start failing in one test case when trying to open a file after creating a dataset with an opaque type in it. This fails both before and after my changes to `H5_IS_BUFFER_OVERFLOW`. More investigation is likely needed there, but I'll put the change back for now.

src/H5Odtype.c
```diff
@@ -176,6 +176,14 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
                 HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, FAIL, "ran off end of input buffer while decoding");
             UINT16DECODE(*pp, dt->shared->u.atomic.offset);
             UINT16DECODE(*pp, dt->shared->u.atomic.prec);
+
+            /* Sanity checks */
+            if (dt->shared->u.atomic.offset >= (dt->shared->size * 8))
+                HGOTO_ERROR(H5E_DATATYPE, H5E_BADRANGE, FAIL, "integer offset out of bounds");
+            if (0 == dt->shared->u.atomic.prec)
```
fortnern, line 183: Could we support this as a no-op?

src/H5Odtype.c
```diff
@@ -176,6 +176,14 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
...
+                HGOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "precision is zero");
+            if (((dt->shared->u.atomic.offset + dt->shared->u.atomic.prec) - 1) >= (dt->shared->size * 8))
```
fortnern, line 185: Could we get rid of the -1 and change the >= to a > here (and in other places in this function)?

src/H5Odtype.c
fortnern, line 492: I think we should either always use `offset + size > dt->shared->size` or `offset + size - 1 >= dt->shared->size`

src/H5Oattribute.c
```diff
@@ -211,6 +211,13 @@ H5O__attr_create(const H5O_loc_t *loc, H5A_t *attr)
     if (NULL == (oh = H5O_pin(loc)))
         HGOTO_ERROR(H5E_ATTR, H5E_CANTPIN, FAIL, "unable to pin object header");
 
+    /* Check for creating attribute with unusual datatype */
+    if (H5T_is_numeric_with_unusual_unused_bits(attr->shared->dt) &&
```
fortnern, line 215: Maybe move the checksum part up so we can short circuit the more expensive function call

src/H5Odtype.c
```diff
@@ -770,6 +812,19 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
             HGOTO_ERROR(H5E_DATATYPE, H5E_UNSUPPORTED, FAIL, "unknown datatype class found");
     }
 
+    /* Check for numeric type w/unusual # of unused bits */
+    if (H5T_is_numeric_with_unusual_unused_bits(dt))
```
fortnern, line 818: This doesn't seem to suppress the error when checksumming is turned on

derobins:
> CMake seems to be failing in the `H5DUMP-tnbit` test because the compression ratio line isn't getting masked out for some reason. Not sure on netcdf failures, but investigating.

In the Autotools, I had to change the lead digit of a macro, since it goes from 10 to 1.

jhendersonHDF:
Will address the H5_IS_BUFFER_OVERFLOW issue and @fortnern's comments in a separate PR so as to not hold up this PR and create a bunch of extra testing work.

PR #4238, opened 2024-03-25
Fix uninitialized variable

byrnHDF:


c++/test/tattr.cpp
```diff
@@ -1542,7 +1542,7 @@ test_attr_dtype_shared(FileAccPropList &fapl)
         if (is_default_vfd_compat) {
             // Check size of file
             filesize = h5_get_file_size(FILE_DTYPE.c_str(), H5P_DEFAULT);
-            verify_val(static_cast<long>(filesize), static_cast<long>(empty_filesize), "Checking file size",
+            verify_val(static_cast<long>(filesize), static_cast<long>(filesize), "Checking file size",
```
jhendersonHDF, line 1545: Should be fixed in #4232. This is trying to compare the current file size against the empty file size retrieved earlier.

byrnHDF: Okay - looks like the failure I was looking at was at the same time as 4232 was merged. I'll close this.

PR #4240, opened 2024-03-25
Rework Fortran macros to use the proper code.

byrnHDF:


config/cmake/HDFUseFortran.cmake
```diff
@@ -109,6 +87,8 @@ set (REALISNOTDOUBLE_CODE
        END PROGRAM main
   "
 )
+check_fortran_source_compiles (${REALISNOTDOUBLE_CODE} ${HDF_PREFIX}_FORTRAN_DEFAULT_REAL_NOT_DOUBLE SRC_EXT f90)
```
brtnfld: This is not needed anymore since we now generate all the interfaces.

lrknox (approved):

brtnfld (approved):

PR #4243, opened 2024-03-26
Turn on -Werror for Java in GitHub -Werror workflows

jhendersonHDF:


config/gnu-warnings/error-8
```diff
@@ -7,3 +7,14 @@
...
+# it normally wouldn't (because those platforms don't require alignment).
+# While this flag doesn't follow the -Werror format like above, it's
+# placed here to make sure that it comes after the -Werror=cast-align
+# line from error-general in the list of flags. Otherwise, if the
+# '--enable-warnings-as-errors' flag isn't passed at configure time,
+# the logic in config/gnu-flags that demotes these errors to their
+# normal warning form will cause -Wcast-align to come after
+# -Wcast-align=strict in the list of flags, causing it to take
+# precedence and mask cast-align warnings from GCC on certain platforms.
+-Wcast-align=strict
```
jhendersonHDF, line 20: While clang always seems to warn about cast-align problems, GCC won't on some platforms like x86 and others unless you specify cast-align=strict (because those platforms don't require alignment). We were adding this in `config/gnu-warnings/8`, but the way the flags ended up being ordered was causing the '-Werror=cast-align' line from `config/gnu-warnings/error-general` to come later than the strict version, causing it to take precedence and hide these warnings with GCC.

derobins (approved):

lrknox (approved):

byrnHDF (approved):

PR #4250, opened 2024-03-26
Fix fatal error C1083: Cannot open include file

hyoklee:
See https://cdash.hdfgroup.org/viewBuildError.php?buildid=125323

### DIFF
--- a/src/H5Zdeflate.c
+++ b/src/H5Zdeflate.c
@@ -23,7 +23,7 @@
 #define H5_ZLIB_HEADER "zlib.h"
 #endif
 #if defined(H5_ZLIB_HEADER)
-#include H5_ZLIB_HEADER /* "zlib.h" */
+#include H5_ZLIB_HEADER "zlib.h"
 #endif
 
 /* Local function prototypes */


src/H5Zdeflate.c
```diff
@@ -23,7 +23,7 @@
 #define H5_ZLIB_HEADER "zlib.h"
 #endif
 #if defined(H5_ZLIB_HEADER)
-#include H5_ZLIB_HEADER /* "zlib.h" */
+#include H5_ZLIB_HEADER "zlib.h"
```
byrnHDF, line 26: I don't think you want both, and why does H5_ZLIB_HEADER fail since it should be exactly the same as the text you added?

byrnHDF: direct_chunk.c has the same include block - does it fail?

byrnHDF: The reason it is not hard-coded is that the header file could be overridden with something different.

byrnHDF (changes requested):
The question should be why is the H5_ZLIB_HEADER not defined correctly?

hyoklee:
@byrnHDF, I set `-DHDF5_ENABLE_Z_LIB_SUPPORT:BOOL=OFF`.
Build failed on win10 + VS2022 + oneAPI mpicc.

tbeu:
By the way, why is it called `HDF5_ENABLE_Z_LIB_SUPPORT` and not `HDF5_ENABLE_ZLIB_SUPPORT`? Would make dev live easier when filtering for zlib in CMake GUI.

grafik

src/H5Zdeflate.c
hyoklee: > direct_chunk.c has the same include block - does it fail?

No. (I guess build process failed before it reached direct_chunk.c).

hyoklee: > I don't think you want both, and why does H5_ZLIB_HEADER fail since it should be exactly the same as the text you added?

I think the problem is escape character: `\"zlib.h\"`. How can we fix it?

byrnHDF: That was passed in as a define so that needs fixed

PR #4254, opened 2024-03-26
Remove an error check regarding large cache objects

derobins:
In PR#4231 an assert() call was converted to a normal HDF5 error check. It turns out that the original assert() was added by a developer as a way of being alerted that large cache objects existed instead of as a guard against incorrect behavior, making it unnecessary in either debug or release builds.

The error check has been removed.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -700,10 +700,7 @@ Bug Fixes since HDF5-1.14.0 release
       builds. In HDF5 1.14.4, this can happen if you create a very large
       number of links in an old-style group that uses local heaps.
 
-      The library will now emit a normal error when it tries to load a
-      metadata object that is too large.
-
-      Partially addresses GitHub #3762
+      Fixes GitHub #3762
 
     - Fixed an issue with the Subfiling VFD and multiple opens of a
       file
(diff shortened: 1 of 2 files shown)


release_docs/RELEASE.txt
```diff
@@ -700,10 +700,7 @@ Bug Fixes since HDF5-1.14.0 release
       builds. In HDF5 1.14.4, this can happen if you create a very large
       number of links in an old-style group that uses local heaps.
 
-      The library will now emit a normal error when it tries to load a
```
lrknox, line 703: The paragraph above from the original release note says that a too-large size object will trip an assert, but It seems that assert is the original error check that has been removed,  Probably that statement is now incorrect?

lrknox (changes requested):

release_docs/RELEASE.txt
derobins: The cache will hold the object. The assert is unnecessary. John just added it as a way of being alerted that we'd created a large object. It's not a guard against incorrect behavior.

qkoziol (approved):
Is there an actual problem still?

lrknox (approved):

PR #4264, opened 2024-03-27
Prevent stack overflows in H5E__push_stack

jhendersonHDF:


jhendersonHDF:
The C++ testhdf5 test is currently failing in Release builds with the family VFD due to a stack overflow in `H5E_printf_stack() -> H5E__push_stack()`. I'm not certain yet why the call to `H5E_printf_stack()` initially happens, but the stack overflow happens when `H5E__push_stack()` can't increment the reference count on an error class/major/minor ID and then `H5I_inc_ref()` calls `HGOTO_ERROR() -> H5E_printf_stack()` and goes on. The `H5MM_xstrdup` call also needs fixed in `H5E__push_stack()`, but I wanted to see if this is a reasonable approach first.

derobins (approved):

qkoziol (approved):

PR #4270, opened 2024-03-28
Clean up test/cmpd_dtransform.c

derobins:
* Fix uninitialized memory warning from sanitizers
* FAIL_STACK_ERROR --> TEST_ERROR
* Emit output
* Delete test file when done

### DIFF
--- a/test/cmpd_dtransform.c
+++ b/test/cmpd_dtransform.c
@@ -132,14 +132,6 @@ main(void)
         if (data[idx] != data_res[idx])
             TEST_ERROR;
 
-    free(atts);
-    free(atts_res);
-    free(data);
-    free(data_res);
-
-    if (H5Fdelete(FILENAME, H5P_DEFAULT) < 0)
-        TEST_ERROR;
-
     /* Close all identifiers. */
     if (H5Pclose(dxpl_id) < 0)
         TEST_ERROR;
@@ -157,6 +149,13 @@ main(void)
         TEST_ERROR;
     if (H5Tclose(str_dtyp_id) < 0)
         TEST_ERROR;
+
+    free(atts);
+    free(atts_res);
+    free(data);
+    free(data_res);
+
+    HDremove(FILENAME);
 
     PASSED();
     return EXIT_SUCCESS;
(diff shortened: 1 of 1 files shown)


test/cmpd_dtransform.c
```diff
@@ -43,54 +44,58 @@ main(void)
     att_t      *atts        = NULL;
     att_t      *atts_res    = NULL;
 
+    printf("Testing writing compound attributes followed by data w/ transform.\n");
+
+    TESTING("data types are reset properly");
+
     /* Compound datatype */
-    if (NULL == (atts = malloc(sizeof(att_t))))
+    if (NULL == (atts = (att_t *)calloc(1, sizeof(att_t))))
```
derobins, line 52: This raised uninitialized memory sanitizer warnings

test/cmpd_dtransform.c
```diff
@@ -132,25 +137,29 @@ main(void)
     free(data);
     free(data_res);
 
+    if (H5Fdelete(FILENAME, H5P_DEFAULT) < 0)
+        TEST_ERROR;
```
jhendersonHDF: May want to set any freed buffers above to NULL so a failure here doesn't double-free.

test/cmpd_dtransform.c
```diff
@@ -43,54 +44,58 @@ main(void)
     att_t      *atts        = NULL;
     att_t      *atts_res    = NULL;
 
+    printf("Testing writing compound attributes followed by data w/ transform.\n");
+
+    TESTING("data types are reset properly");
```
jhendersonHDF, line 49: Does this need to be TESTING_2 if the printf above is a sort of test heading?

jhendersonHDF (approved):
Minor comments, but looks good

hyoklee (approved):

test/cmpd_dtransform.c
derobins: That's how we enter every test. It's just glommed in with the test output itself since this is a one-off written by a non-HDFG person.

test/cmpd_dtransform.c
derobins: Fixed via reordering

PR #4272, opened 2024-03-28
In CMAKE replace deprecated CMAKE_COMPILER_IS_GNU**

byrnHDF:


derobins (approved):

derobins:
I feel like this probably should have been two PRs since there's two things going on here

byrnHDF:
> I feel like this probably should have been two PRs since there's two things going on here
Well, when I added the sanitzer CXX checks is when I discovered the deprecation. The sanitizer change can be separated.

jhendersonHDF (approved):

PR #4273, opened 2024-03-28
Merge julia workflows into standard ci format

byrnHDF:


.github/workflows/julia-auto.yml
```diff
@@ -0,0 +1,80 @@
...
+          sudo apt install -y m4 flex bison cmake libzip-dev openssl build-essential
+
+      - name: Autotools Configure
+        shell: bash
+        run: |
+          sh ./autogen.sh
+          mkdir "${{ runner.workspace }}/build"
+          cd "${{ runner.workspace }}/build"
+          $GITHUB_WORKSPACE/configure \
+            FCFLAGS="-fPIC -fortranlibs" \
```
jhendersonHDF: Doesn't seem like we need to set FCFLAGS here since we disable fortran below

.github/workflows/mingw-cmake.yml
jhendersonHDF, line 1: Not used anymore?

.github/workflows/julia-auto.yml
derobins: Agree. Should remove this line.

.github/workflows/mingw-cmake.yml
byrnHDF: Nope, accident to include it

jhendersonHDF (approved):

derobins (approved):

PR #4276, opened 2024-03-28
Addresses configuration fortran testing flags

brtnfld:
Turns warnings into errors for intel in the Fortran logical configuration test.
Tested on Perlmutter with Intel compilers.


epourmal (approved):
Looks good but I am wondering if this change should be in the Intel config file in the config directory?

lrknox (approved):

brtnfld:
The issue is that we now have multiple checks with Intel for classic and one-API, and some of the variables needed are not global. It seems more straightforward to get the information from the compiler output.

derobins (approved):

byrnHDF (approved):

PR #4277, opened 2024-03-28
Add configure options for disabling extension features

jhendersonHDF:
Add configure option to enable or disable extension features in general

Add configure option to enable or disable _Float16 support


config/cmake/ConfigureChecks.cmake
```diff
@@ -872,93 +872,113 @@ H5ConversionTests (${HDF_PREFIX}_LLONG_TO_LDOUBLE_CORRECT TRUE "Checking IF corr
...
-      # would cause any assignments to a _Float16 variable to be elided,
-      # whereas Debug builds would perform incorrect hardware conversions by
-      # simply chopping off all the bytes of the value except for the first 2.
-      # These tests pass on MacOS 14, so let's perform a quick test to check
-      # if the hardware conversion is done correctly.
+# Option to enable or disable all non-standard features. Specific features can
+# be enabled or disabled with their respective options below
+option (HDF5_ENABLE_NONSTANDARD_FEATURES "Enable support for non-standard programming language features" ON)
+# Options for enabling or disabling individual features
+option (HDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16 "Enable support for _Float16 C datatype" ${HDF5_ENABLE_NONSTANDARD_FEATURES})
```
jhendersonHDF, line 886: @byrnHDF I couldn't figure out a nice way to make it so that turning off `HDF5_ENABLE_NONSTANDARD_FEATURES` in a GUI will also turn off `HDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16` initially and vice versa, but for most cases we probably care about, this tends to just work.

derobins:
I feel like we should add this to the libhdf5.settings file and H5build_settings*.

jhendersonHDF:
> I feel like we should add this to the libhdf5.settings file and H5build_settings*.

Seems reasonable. Thinking about it, I probably also need to add the CMake options to the hdf5-config.cmake file

config/cmake/ConfigureChecks.cmake
byrnHDF: I think this is a really good solution - can still be changed by command line.

byrnHDF (approved):

lrknox (approved):

PR #4278, opened 2024-03-29
Fix range check in H5_addr_overlap

jhendersonHDF:
When the H5_addr_overlap macro was updated to use H5_RANGE_OVERLAP, it failed to take into account that H5_RANGE_OVERLAP expects the range to be inclusive. This lead to an assertion failure in H5MM_memcpy due to a memcpy operation on overlapping memory. This has been fixed by subtracting 1 from the calculated high bound values passed to H5_RANGE_OVERLAP.


jhendersonHDF:
Testing by @derobins on the fheap test with a test express setting of 0 showed this started causing problems after #4234.

derobins (approved):

lrknox (approved):

qkoziol:
Good fix, I should have caught that!

PR #4281, opened 2024-03-29
Add TestExpress=0 workflow

byrnHDF:


.github/workflows/testxpr.yml
```diff
@@ -0,0 +1,137 @@
...
+    - cron: "0 9 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  build_and_test:
+    strategy:
+      matrix:
+        build_sys: ["CMake"]
```
jhendersonHDF, line 16: Will this just be Cmake for now?

.github/workflows/testxpr.yml
```diff
@@ -0,0 +1,137 @@
...
+                $GITHUB_WORKSPACE
+        if: ${{ matrix.build_sys == 'CMake' }}
+
+      - name: CMake Build
+        run: cmake --build . --parallel 3 --config ${{ matrix.build_mode }}
+        working-directory: ${{ runner.workspace }}/build
+        if: ${{ matrix.build_sys == 'CMake' }}
+
+      - name: CMake Run Tests
+        run: ctest . --parallel 2 -C ${{ matrix.build_mode }} -V -R H5TESTXPR
```
jhendersonHDF, line 101: We may have to see how well these tests run and then set this back to serial testing if necessary. At express level 0 some of these tests are so comprehensive that running two different ones at the same time will cause them both to take an order of magnitude longer to run.

.github/workflows/testxpr.yml
```diff
@@ -0,0 +1,137 @@
+name: hdf5 TestExpress CI
+
+# Run TestExpress CI daily at 04:00 CDT (9:00 UTC)
+on:
+  workflow_dispatch:
+  schedule:
```
jhendersonHDF, line 6: Since this will mostly run on a schedule, we may want to add a status link in the README or somewhere else so we can monitor it.

.github/workflows/testxpr.yml
byrnHDF: Autotools changes in pipeline once CMake works

.github/workflows/testxpr.yml
byrnHDF: That's kinda the plan - see how it works first

.github/workflows/testxpr.yml
byrnHDF: Maybe, sounds good.

jhendersonHDF (approved):

lrknox (approved):

PR #4283, opened 2024-03-29
Fixes for file format security issues

jhendersonHDF:
Addresses file format security issues detected via fuzzing.

Credit: Amazon Web Services and @qkoziol

### DIFF
--- a/src/H5Glink.c
+++ b/src/H5Glink.c
@@ -176,70 +176,6 @@ H5G__link_cmp_corder_dec(const void *lnk1, const void *lnk2)
 } /* end H5G__link_cmp_corder_dec() */
 
 /*-------------------------------------------------------------------------
- * Function:	H5G__ent_to_link
- *
- * Purpose:     Convert a symbol table entry to a link
- *
- * Return:	Non-negative on success/Negative on failure
- *
- *-------------------------------------------------------------------------
- */
-herr_t
-H5G__ent_to_link(H5O_link_t *lnk, const H5HL_t *heap, const H5G_entry_t *ent, const char *name)
-{
-    bool   dup_soft  = false;   /* xstrdup the symbolic link name or not */
-    herr_t ret_value = SUCCEED; /* Return value */
-
-    FUNC_ENTER_PACKAGE
-
-    /* check arguments */
-    assert(lnk);
-    assert(heap);
-    assert(ent);
-    assert(name);
-
-    /* Set (default) common info for link */
-    lnk->cset         = H5F_DEFAULT_CSET;
-    lnk->corder       = 0;
-    lnk->corder_valid = false; /* Creation order not valid for this link */
-    if ((lnk->name = H5MM_xstrdup(name)) == NULL)
-        HGOTO_ERROR(H5E_LINK, H5E_CANTGET, FAIL, "unable to duplicate link name");
-
-    /* Object is a symbolic or hard link */
-    if (ent->type == H5G_CACHED_SLINK) {
-        const char *s; /* Pointer to link value */
-
-        if ((s = (const char *)H5HL_offset_into(heap, ent->cache.slink.lval_offset)) == NULL)
-            HGOTO_ERROR(H5E_LINK, H5E_CANTGET, FAIL, "unable to get symbolic link name");
-
-        /* Copy the link value */
-        if ((lnk->u.soft.name = H5MM_xstrdup(s)) == NULL)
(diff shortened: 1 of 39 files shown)


src/H5T.c
```diff
@@ -5265,26 +5285,28 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
...
-H5T__path_find_init_path_table(void)
+H5T__init_path_table(void)
 {
     herr_t ret_value = SUCCEED;
 
     FUNC_ENTER_PACKAGE
 
-    assert(0 == H5T_g.npaths);
+    /* Sanity check */
+    if (0 != H5T_g.npaths)
```
jhendersonHDF, line 5295: The changes in this file mostly come from merging my cleanup with the same cleanup from @qkoziol. This appears to be the main change that was in this file.

src/H5T.c
```diff
@@ -5674,6 +5698,47 @@ H5T_path_noop(const H5T_path_t *p)
...
+ * Function:    H5T_noop_conv
+ *
+ * Purpose:     Check if a conversion between two dataypes will be a no-op
+ *
+ * Return:      true / false (can't fail)
+ *
+ *-------------------------------------------------------------------------
+ */
+bool
+H5T_noop_conv(const H5T_t *src, const H5T_t *dst)
```
jhendersonHDF, line 5711: The changes in this file mostly come from merging my cleanup with the same cleanup from @qkoziol. This appears to be the main change that was in this file.

tools/lib/h5tools_utils.c
```diff
@@ -589,12 +589,12 @@ dump_table(hid_t fid, char *tablename, table_t *table)
 
     PRINTSTREAM(rawoutstream, "%s: # of entries = %d\n", tablename, table->nobjs);
     for (u = 0; u < table->nobjs; u++) {
-        H5VLconnector_token_to_str(fid, table->objs[u].obj_token, &obj_tok_str);
```
jhendersonHDF, line 592: This function is inside a debugging #ifdef, so looks like it was just never updated.

src/H5Odtype.c
```diff
@@ -388,6 +389,13 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
                 if (H5_IS_KNOWN_BUFFER_OVERFLOW(skip, *pp, actual_name_length, p_end))
                     HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, FAIL, "ran off end of input buffer while decoding");
 
+                /* Check for duplicated field name */
+                for (memb_idx = 0; memb_idx < dt->shared->u.compnd.nmembs; memb_idx++)
```
jhendersonHDF, line 393: I wonder if we can do this check at the end of decoding a compound datatype and in some efficient way rather than doing it for every single member. That seems like a real performance killer for large compounds.

src/H5Glink.c
```diff
@@ -175,70 +175,6 @@ H5G__link_cmp_corder_dec(const void *lnk1, const void *lnk2)
...
- * Function:	H5G__ent_to_link
- *
- * Purpose:     Convert a symbol table entry to a link
- *
- * Return:	Non-negative on success/Negative on failure
- *
- *-------------------------------------------------------------------------
- */
-herr_t
-H5G__ent_to_link(H5O_link_t *lnk, const H5HL_t *heap, const H5G_entry_t *ent, const char *name)
```
jhendersonHDF, line 188: This function appears to have been moved to `H5Gent.c`, while `H5G__ent_convert()` from `H5Gent.c` was renamed to `H5G__link_to_ent()` and moved into this file.

derobins (approved):

src/H5Odtype.c
qkoziol: I believe that it'll be O(n^2) either way...  :-/

src/H5Glink.c
qkoziol: Yes, made more senes with what the routines were doing

qkoziol (approved):

src/H5Odtype.c
jhendersonHDF: What if we build a temporary hashmap indexed by hashing the member name and use it for lookups?

jhendersonHDF: Not something that really needs to hold this PR up, but there are probably some ways we can improve on this.

PR #4285, opened 2024-03-29
Fix example ref files path

byrnHDF:


doxygen/dox/ExamplesAPI.dox
```diff
@@ -726,7 +726,7 @@ FORTRAN
   MATLAB  PyHigh  PyLow
 </td>
 <td>h5ex_t_vlstring.h5</td>
-<td><a href="https://github.com/HDFGroup/hdf5/blob/develop/HDF5Examples/C/H5T/tfiles/18/h5ex_t_vlstring.tst">h5ex_t_vlstring.tst</a></td>
+<td><a href="https://github.com/HDFGroup/hdf5/blob/develop/HDF5Examples/C/H5T/tfiles/16/h5ex_t_vlstring.tst">h5ex_t_vlstring.tst</a></td>
 <td><a href="https://github.com/HDFGroup/hdf5/blob/develop/HDF5Examples/C/H5T/tfiles/18/h5ex_t_vlstring.ddl">h5ex_t_vlstring.ddl</a></td>
```
jhendersonHDF, line 730: Why use different versions of things in all these different places? Such as 16 for the .tst file and 18 for the .ddl file?

byrnHDF: Because there is a difference between versions for results: 114 needs to point to the correct files. All the *.tst files are the same since 1.6. DDLs are almost the same since 1.8, references are the same since 112. And of course your change for VLEN is since 1.14.

jhendersonHDF (approved):

lrknox (approved):

PR #4298, opened 2024-04-01
Fix nvc warnings for unreachable break

hyoklee:
```
"/pscratch/sd/l/lrknox/dt-perlm/CI/hdf5-1.15.0/src/H5trace.c", line 2414: warning: statement is unreachable [code_is_unreachable]                           break;                           ^  [CTest: warning matched] Remark: individual 

"/pscratch/sd/l/lrknox/dt-perlm/CI/hdf5-1.15.0/src/H5trace.c", line 2430: warning: statement is unreachable [code_is_unreachable]                           break;                           ^
```

### DIFF
--- a/src/H5trace.c
+++ b/src/H5trace.c
@@ -2411,7 +2411,6 @@ H5_trace_args(H5RS_str_t *rs, const char *type, va_list ap)
                             H5RS_acat(rs, "Reference Region");
                             goto error;
                         } /* end block */
-                        break;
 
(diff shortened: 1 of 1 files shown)


src/H5trace.c
jhendersonHDF, line 1: We should probably make use of any warning suppression options that NVHPC has rather than just removing the break `statements`. Otherwise, if this code's refactored in the future someone may forget to add a `break` back.

PR #4303, opened 2024-04-02
PR merge workflows

byrnHDF:
Merge the Test Express workflows into the PR CI
Split merge request triggers into autotools vs cmake


jhendersonHDF:
Why merge the TestExpress workflows into the CI rather than leaving it as a daily action like before? If it's because they don't take much time, then we don't need a separate workflow to begin with. But the reason to have them as a daily action is because certain VFDs are definitely slow with a TestExpress setting of 0, so having it be a daily action makes it easy to isolate that testing.

byrnHDF:
> But the reason to have them as a daily action is because certain VFDs are definitely slow with a TestExpress setting of 0, so having it be a daily action makes it easy to isolate that testing.

They can be easily called from a vfd workflow as a callable workflow. And we can add an optional parameter to set a vfd env variable.

.github/workflows/testxpr-auto.yml
```diff
@@ -0,0 +1,71 @@
+name: hdf5 TestExpress Autotools CI
+
+# Run TestExpress CI daily at 04:00 CDT (9:00 UTC)
```
jhendersonHDF: Not relevant anymore

.github/workflows/testxpr-cmake.yml
```diff
@@ -0,0 +1,69 @@
+name: hdf5 TestExpress CMake CI
+
+# Run TestExpress CI daily at 04:00 CDT (9:00 UTC)
```
jhendersonHDF: Same here

jhendersonHDF (approved):

lrknox (approved):

PR #4304, opened 2024-04-02
Report build options of VFDs

brtnfld:


CMakeLists.txt
```diff
@@ -666,6 +666,7 @@ set (CMAKE_MODULE_PATH ${HDF_RESOURCES_DIR} ${CMAKE_MODULE_PATH})
 # Option to Enable HDFS
 #-----------------------------------------------------------------------------
 option (HDF5_ENABLE_HDFS "Enable HDFS" OFF)
+set(H5_HAVE_LIBHDFS_SETTING OFF)
```
jhendersonHDF: Rather than adding a bunch of new variables for all the settings that could be ON or OFF, we should probably just force the option variable values to OFF if the feature can't be enabled and then just use the value of the option variables.

byrnHDF: Agree, options are ON/OFF and should be used in the settings file (if the option is requested and fails to be allowed the option should be set to OFF and a message displayed). H5_* variable should be only for code.

byrnHDF: If there is not an option for a setting variable then we can create a temp at the time of creating the settings file.

brtnfld: I'll update it.

brtnfld:
So would it be something like this to reset it if the option fails?

set (HDF5_ENABLE_MIRROR_VFD OFF CACHE BOOL "Build the Mirror Virtual File Driver" FORCE)

jhendersonHDF:
> So would it be something like this to reset it if the option fails?
> 
> set (HDF5_ENABLE_MIRROR_VFD OFF CACHE BOOL "Build the Mirror Virtual File Driver" FORCE)

Yes, we'd need to add lines like that in all the places where a feature gets disabled for some reason. Some refactoring of the CMake logic may be needed to make that not too ugly.

hyoklee (approved):

lrknox (approved):

byrnHDF (approved):

PR #4309, opened 2024-04-02
Simplify check for datatypes with unusual number of unused bits

jhendersonHDF:
Avoids potential undefined behavior in H5T_is_numeric_with_unusual_unused_bits

Fixes #4293

PR #4315, opened 2024-04-03
Add check for decoded datatype precision overflow

jhendersonHDF:
Adds a check for the case where a decoded datatype's precision could overflow SIZE_MAX due to the size of a datatype being larger than SIZE_MAX / 8

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -6261,10 +6261,11 @@ H5T_set_loc(H5T_t *dt, H5VL_object_t *file, H5T_loc_t loc)
         /* Check the datatype of this element */
         switch (dt->shared->type) {
             case H5T_ARRAY: /* Recurse on VL, compound and array base element type */
-                /* Recurse if it's VL, compound, enum or array */
+                /* Recurse if it's VL, compound, enum, array or reference */
                 /* (If the force_conv flag is _not_ set, the type cannot change in size, so don't recurse) */
                 if (dt->shared->parent->shared->force_conv &&
-                    H5T_IS_COMPLEX(dt->shared->parent->shared->type)) {
+                    (H5T_IS_COMPLEX(dt->shared->parent->shared->type) ||
+                     H5T_IS_REF(dt->shared->parent->shared))) {
                     /* Keep the old base element size for later */
                     old_size = dt->shared->parent->shared->size;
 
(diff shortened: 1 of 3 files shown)


jhendersonHDF:
Related to #4309. Adds a check to make sure that `dt->shared->size * 8` doesn't overflow `SIZE_MAX` before we set a decoded datatype's precision or use the previous calculation later on.

src/H5Tpkg.h
```diff
@@ -45,8 +45,7 @@
 #define H5T_NAMELEN 32
 
 /* Macro to ease detecting "complex" datatypes (i.e. those with base types or fields) */
-#define H5T_IS_COMPLEX(t)                                                                                    \
-    ((t) == H5T_COMPOUND || (t) == H5T_ENUM || (t) == H5T_VLEN || (t) == H5T_ARRAY || (t) == H5T_REFERENCE)
+#define H5T_IS_COMPLEX(t) ((t) == H5T_COMPOUND || (t) == H5T_ENUM || (t) == H5T_VLEN || (t) == H5T_ARRAY)
```
jhendersonHDF, line 48: I believe `H5T_REFERENCE` was only added to this macro to make the logic in `H5T_set_loc` work correctly for reference types. Reference types are treated as atomic types, but would be excluded when checking a datatype with the `H5T_IS_ATOMIC` macro due to the `!H5T_IS_COMPLEX` check.

src/H5T.c
```diff
@@ -6347,8 +6349,7 @@ H5T_set_loc(H5T_t *dt, H5VL_object_t *file, H5T_loc_t loc)
                  * them as part of the same blob)*/
                 /* (If the force_conv flag is _not_ set, the type cannot change in size, so don't recurse) */
                 if (dt->shared->parent->shared->force_conv &&
-                    H5T_IS_COMPLEX(dt->shared->parent->shared->type) &&
-                    (dt->shared->parent->shared->type != H5T_REFERENCE)) {
+                    H5T_IS_COMPLEX(dt->shared->parent->shared->type)) {
```
derobins, line 6352: Weird indenting

derobins (approved):

qkoziol (commented):
What about the other places that use H5T_IS_COMPLEX?

jhendersonHDF:
> What about the other places that use H5T_IS_COMPLEX?

When I looked at the other occurrences previously, there are four occurrences in H5T.c, three of which are addressed here. The other one is in `H5T_detect_class`, which is just trying to call `H5T_detect_class` recursively on container datatypes and would return right away for `H5T_REFERENCE`.

There is one occurrence in H5Tvisit.c in `H5T_visit`, which only seems to be used for recursive calls of `H5T__upgrade_version_cb`, which returns right away for `H5T_REFERENCE`.

There are three occurrences in H5Tvlen.c, all in `H5T__vlen_reclaim`. Those ones are a bit trickier and might need an additional H5T_IS_REF check in case the datatype is something like a variable-length of array of references. I didn't see any memory leaks from this, but it's possible we aren't testing datatypes like this. I'll look closer.

jhendersonHDF:
Will address this in the future.

PR #4320, opened 2024-04-03
Update support branch version number to 1.14.5-1

glennsong09:


c++/src/cpp_doc_config
```diff
@@ -38,7 +38,7 @@ PROJECT_NAME           =
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = "1.14.4-1, currently under development"
+PROJECT_NUMBER         = "1.14.5-1, currently under development"
```
bmribler, line 41: Is 1.14.5 correct?

glennsong09: According to the release process document, it is correct.

byrnHDF: Now that the release branch is created

byrnHDF (approved):

c++/src/cpp_doc_config
lrknox: 1.14.5 is the correct version for the hdf5_1_14 branch - for the next 1.14 release.  The version in the release branch hdf5_1_14_4 is 1.14.4-2 until the -2 is removed for the actual release.

jhendersonHDF (approved):

PR #4321, opened 2024-04-04
Fix h5py testing failure due to invalid datatype IDs

jhendersonHDF:
Fixes an issue where invalid datatype IDs are passed to application conversion functions in the case where the top-level conversion function is a library-internal function that operates on a container-like datatype, but one or more of the base datatype members are converted with an application conversion function.

### DIFF
--- a/src/H5T.c
+++ b/src/H5T.c
@@ -3394,6 +3394,7 @@ H5T__create(H5T_class_t type, size_t size)
             /* Copy the default string datatype */
             if (NULL == (dt = H5T_copy(origin_dt, H5T_COPY_TRANSIENT)))
                 HGOTO_ERROR(H5E_DATATYPE, H5E_CANTINIT, NULL, "unable to copy");
+            dt->shared->type = type;
 
             /* Modify the datatype */
             if (H5T__set_size(dt, size) < 0)
@@ -4578,6 +4579,8 @@ H5T_cmp(const H5T_t *dt1, const H5T_t *dt2, bool superset)
             /*
              * Compound data types...
              */
+            if (dt1->shared->u.compnd.nmembs == 0 && dt2->shared->u.compnd.nmembs == 0)
+                HGOTO_DONE(0);
             if (dt1->shared->u.compnd.nmembs < dt2->shared->u.compnd.nmembs)
                 HGOTO_DONE(-1);
             if (dt1->shared->u.compnd.nmembs > dt2->shared->u.compnd.nmembs)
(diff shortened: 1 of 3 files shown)


src/H5T.c
```diff
@@ -3394,6 +3394,7 @@ H5T__create(H5T_class_t type, size_t size)
             /* Copy the default string datatype */
             if (NULL == (dt = H5T_copy(origin_dt, H5T_COPY_TRANSIENT)))
                 HGOTO_ERROR(H5E_DATATYPE, H5E_CANTINIT, NULL, "unable to copy");
+            dt->shared->type = type;
```
jhendersonHDF, line 3397: The logic for `H5Tcreate` with `H5T_INTEGER`, `H5T_FLOAT`, `H5T_TIME` and `H5T_STRING` weren't setting the new datatype's type to the given type, like the other cases below. This meant that newly-created types in any combination of these four categories could compare equal to each other if their sizes are the same.

src/H5T.c
```diff
@@ -4578,6 +4579,8 @@ H5T_cmp(const H5T_t *dt1, const H5T_t *dt2, bool superset)
             /*
              * Compound data types...
              */
+            if (dt1->shared->u.compnd.nmembs == 0 && dt2->shared->u.compnd.nmembs == 0)
```
jhendersonHDF, line 4582: Shortcut for comparing two compounds with 0 members. Also avoids a segfault in the debugging code below, but I added a check there to be safe.

src/H5T.c
```diff
@@ -4660,6 +4665,8 @@ H5T_cmp(const H5T_t *dt1, const H5T_t *dt2, bool superset)
             /*
              * Enumeration data types...
              */
+            if (dt1->shared->u.enumer.nmembs == 0 && dt2->shared->u.enumer.nmembs == 0)
```
jhendersonHDF, line 4668: Shortcut for comparing two enums with 0 members. Also avoids a segfault in the debugging code below, but I added a check there to be safe.

src/H5Tconv.c
```diff
@@ -2134,18 +2143,18 @@ H5T__conv_struct_init(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, co
             HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL,
                         "couldn't allocate destination compound member datatype array");
 
-        priv->need_ids = (cdata->command == H5T_CONV_INIT && conv_ctx->u.init.cb_struct.func) ||
```
jhendersonHDF, line 2137: For the container datatype conversions (compound, vlen, array), the code previously didn't take into account the case where the top-level datatype uses a library-internal conversion function, but one or more of the member types uses an application conversion function. Refactored so that we delay checking for and creating IDs until after we determine the conversion path to be used.

src/H5Tconv.c
```diff
@@ -2701,6 +2722,8 @@ H5T__conv_struct_opt(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, con
                 HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a datatype");
             if (NULL == conv_ctx)
                 HGOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "invalid datatype conversion context pointer");
+            if (!bkg)
```
jhendersonHDF, line 2725: Can hit an assertion below if `H5Tconvert` is called from an application with NULL for the background buffer. Converted this to an error check.

src/H5Tconv.c
```diff
@@ -2952,6 +2968,10 @@ H5T__conv_enum_init(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, cons
         if (NULL == (priv->dst_copy = H5T_copy(dst, H5T_COPY_ALL)))
             HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCOPY, FAIL, "unable to copy destination datatype");
 
+        /* Nothing more to do if enum has no members */
+        if (0 == src->shared->u.enumer.nmembs)
```
jhendersonHDF, line 2972: Moved the check for 0 enum members down so that the cached private data still gets a copy of the source and destination datatypes in case `H5T_cmp` is called later.

jhendersonHDF:
This is a fix after #4113

byrnHDF (approved):

lrknox (approved):

PR #4323, opened 2024-04-04
Revise _Float16 configure checks

jhendersonHDF:
Run configure checks with and without CFLAGS/CMAKE_C_FLAGS since some compilers work in one case while not working in the other case

Sync CMake configure checks with Autotools


byrnHDF (approved):

lrknox (approved):

config/cmake/ConfigureChecks.cmake
```diff
@@ -923,46 +923,56 @@ if (HDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16)
...
           FALSE
-          "Checking if compiler can convert _Float16 type with casts"
+          "Checking if compiler can convert _Float16 type with casts (without CMAKE_C_FLAGS)"
       )
 
       set (CMAKE_C_FLAGS "${cmake_c_flags_backup}")
 
-      if (${${HDF_PREFIX}_FLOAT16_CONVERSION_FUNCS_LINK})
+      if (${HDF_PREFIX}_FLOAT16_CONVERSION_FUNCS_LINK AND ${HDF_PREFIX}_FLOAT16_CONVERSION_FUNCS_LINK_NO_FLAGS)
         # Finally, MacOS 13 appears to have a bug specifically when converting
```
brtnfld, line 950: Is the issue related to the OS or the C clang version? Though I've not tried it, I would assume that gcc installed on 13 OS systems would work.

brtnfld (approved):

config/cmake/ConfigureChecks.cmake
jhendersonHDF: I'm just referring to the default AppleClang here for a particular MacOS version. I'd assume GCC should work as well, but haven't tested that.

PR #4330, opened 2024-04-04
Set available CMake build types for non-multi-config generators

jhendersonHDF:


jhendersonHDF:
Not needed for the release. @byrnHDF I don't believe this should affect anything in the CMake code, but it makes it convenient to set the build type with ccmake/cmake-gui since it becomes a toggle/drop-down rather than needing to type in the build type.

derobins (approved):

derobins:
> Not needed for the release. @byrnHDF I don't believe this should affect anything in the CMake code, but it makes it convenient to set the build type with ccmake/cmake-gui since it becomes a toggle/drop-down rather than needing to type in the build type.

Didn't we already have this? Did it regress?

jhendersonHDF:
> > Not needed for the release. @byrnHDF I don't believe this should affect anything in the CMake code, but it makes it convenient to set the build type with ccmake/cmake-gui since it becomes a toggle/drop-down rather than needing to type in the build type.
> 
> Didn't we already have this? Did it regress?

We used to have this, but it was changed long ago. Not sure the reason or if it was just unintentional.

byrnHDF (approved):

glennsong09 (approved):

PR #4338, opened 2024-04-07
Clean up random number generator code

derobins:
Depending on the platform, we use a mix of random, rand, and rand_r to generate pseudo-random numbers, along with a messy set of ifdefs in H5private.h. We are not a cryptographic library, only use random numbers in our test code, and have no need for anything more than the C standard's (s)rand(). There's no point dithering about using rand() vs random() when we're also doing bad things like using mod to restrict the range, which introduces bias.

Also removes CMake/configure checks for rand_r and random


derobins:
If we really want to dabble in true randomness, we'll need to do a lot of code cleanup and use things like /dev/urandom, arc4random(), and BCryptGenRandom() to generate random numbers. I feel like this is overkill for our meagre testing purposes, though.

byrnHDF (approved):

qkoziol (approved):
Good cleanup!

qkoziol (approved):

PR #4339, opened 2024-04-07
Remove the getting started guide

derobins:
This was moved to the wiki


derobins:
https://github.com/HDFGroup/hdf5/wiki/Getting-Started-with-HDF5-Development

byrnHDF (approved):

jhendersonHDF:
I feel this is something that should always live with the source code and not be tied to GitHub. Being on the wiki is a convenience and you run into issues with keeping the two in sync, but removing information like this from the source code seems like a mistake.

lrknox (approved):

derobins:
> I feel this is something that should always live with the source code and not be tied to GitHub. Being on the wiki is a convenience and you run into issues with keeping the two in sync, but removing information like this from the source code seems like a mistake.

Let's talk about it at the standup / HDF5WG. I'll revert if we decide to keep it here.

PR #4341, opened 2024-04-07
Remove TRACE macros

derobins:
Removes the TRACE macros from src/H5*.c and updates bin/trace to only handle H5ARG_TRACE macros.


.clang-format
```diff
@@ -68,12 +68,14 @@ StatementMacros:
...
   - END_MEMBERS
   - FUNC_ENTER_API
+  - FUNC_ENTER_API_NAMECHECK_ONLY
   - FUNC_ENTER_NOAPI
   - FUNC_ENTER_NOAPI_NOERR
   - FUNC_ENTER_NOAPI_NOINIT
   - FUNC_ENTER_NOAPI_NOINIT_NOERR
   - FUNC_ENTER_PACKAGE
   - FUNC_LEAVE_API
+  - FUNC_LEAVE_API_NAMECHECK_ONLY
```
derobins, line 78: These were causing clang-format to mess up H5TS.c

derobins:
The handing of H5ARG_TRACE isn't great. I've left it unchanged from the current behavior, but it's going to need some further attention. For example, right now it can't insert an H5ARG_TRACE macro, only update an existing one.

I'll remove H5_trace() in a future PR

bin/trace
derobins, line 1: Probably easier to view this file by itself than as a diff. Also probably want to turn whitespace off since I made the indents uniform.

src/H5I.c
```diff
@@ -90,14 +90,13 @@ static int H5I__iterate_pub_cb(void *obj, hid_t id, void *udata);
  *-------------------------------------------------------------------------
  */
 H5I_type_t
-H5Iregister_type(size_t H5_ATTR_DEBUG_API_USED hash_size, unsigned reserved, H5I_free_t free_func)
+H5Iregister_type(size_t H5_ATTR_UNUSED hash_size, unsigned reserved, H5I_free_t free_func)
```
jhendersonHDF, line 93: Interesting. Very weird to have an unused parameter in an API function.

jhendersonHDF:
Good cleanup. Two things though:

- This still seems like something that would be good to announce prior to removal, in line with https://github.com/HDFGroup/hdf5/wiki/Change-Policy, as I've seen recent examples of people still using this feature.
- There are probably additional things to update alongside this change, such as the entire section on API tracing in https://github.com/HDFGroup/hdf5/blob/develop/doxygen/examples/DebuggingHDF5Applications.html.

src/H5I.c
derobins: It's because the original implementation of ID storage was a hash table that took a size parameter. It's an internal implementation detail that never should have been exposed. We should remove it and version the API call for 2.0.

derobins:
> Good cleanup. Two things though:
> 
> * This still seems like something that would be good to announce prior to removal, in line with https://github.com/HDFGroup/hdf5/wiki/Change-Policy, as I've seen recent examples of people still using this feature.
> * There are probably additional things to update alongside this change, such as the entire section on API tracing in https://github.com/HDFGroup/hdf5/blob/develop/doxygen/examples/DebuggingHDF5Applications.html.

Yes. That guide should be reworked, but I was going to save that for v 2.0 of these changes. I'd like to have the bin/trace script be smarter about pasting the macros in, for example. I only want to update the guide once.

We can announce the change.

qkoziol (approved):

qkoziol:
> > Good cleanup. Two things though:
> > 
> > * This still seems like something that would be good to announce prior to removal, in line with https://github.com/HDFGroup/hdf5/wiki/Change-Policy, as I've seen recent examples of people still using this feature.
> > * There are probably additional things to update alongside this change, such as the entire section on API tracing in https://github.com/HDFGroup/hdf5/blob/develop/doxygen/examples/DebuggingHDF5Applications.html.
> 
> Yes. That guide should be reworked, but I was going to save that for v 2.0 of these changes. I'd like to have the bin/trace script be smarter about pasting the macros in, for example. I only want to update the guide once.
> 
> We can announce the change.

Agree that we should announce this change.

qkoziol (commented):
''All code shall yield to the mighty deleter-of-lines!''   ;-)

jhendersonHDF (approved):

PR #4353, opened 2024-04-08
Add Solaris into tested platforms

hyoklee:
This is a special commit for celebrating the Total Solar Eclipse 2024.


mattjala:
We have no plans to support Solaris beyond testing with it at the current time.

PR #4359, opened 2024-04-09
Add CXX flags to sanitize

byrnHDF:


derobins (approved):

config/toolchain/clang.cmake
```diff
@@ -9,6 +9,7 @@ if(WIN32)
 else()
   set (CMAKE_C_COMPILER clang)
   set (CMAKE_CXX_COMPILER clang++)
+  #set (CMAKE_Fortran_COMPILER flang-new)
```
hyoklee, line 12: Probably remove?

byrnHDF: Maybe - maybe not. Thought it was worthwhile to have a file with a one char change to enable it.
Can be persuaded either way.

jhendersonHDF (approved):

PR #4373, opened 2024-04-10
Add Daily rpm/deb Package Installation Test

hyoklee:
Run 1-hour later than daily snapshot build GH Action that takes 30 minutes.

### DIFF
--- a/.github/workflows/alien.yml
+++ b/.github/workflows/alien.yml
@@ -22,5 +22,5 @@ jobs:
           tar zxvf snapshot-1.14-ubuntu-2204_gcc.rpm.tar.gz
           cd hdf5
           sudo alien HDF5-1.14.4.2-1.x86_64.rpm
-          sudo dpkg -i hdf5-1.14.4.2-2_amd64.deb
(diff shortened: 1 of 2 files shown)


.github/workflows/alien.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: install
+        run:  |
+          sudo apt-get -q update
+          sudo apt install alien
+      - name: test
+        run:  |
+          curl -s -L -O https://github.com/HDFGroup/hdf5/releases/download/snapshot-1.14/snapshot-1.14-ubuntu-2204_gcc.rpm.tar.gz
+          tar zxvf snapshot-1.14-ubuntu-2204_gcc.rpm.tar.gz
+          cd hdf5
+          sudo alien HDF5-1.14.4.2-1.x86_64.rpm
```
byrnHDF, line 24: This should not use hard-coded values - it needs to parse the version from the files.
Why not check develop too?
Also use multiple steps where possible as it improves readability of the reporting.

byrnHDF (changes requested):
Needs work

PR #4377, opened 2024-04-10
Fix all doxygen versions

byrnHDF:


.github/workflows/cmake-ctest.yml
```diff
@@ -129,7 +129,7 @@ jobs:
       - name: Install Dependencies
         uses: ssciwr/doxygen-install@v1
         with:
-          version: "1.10.0"
```
jhendersonHDF, line 132: Are there significant difference between the "build_and_test" steps for the different platforms? Otherwise it'd be easy to have this be a single step instead of duplicated.

lrknox (approved):

jhendersonHDF (approved):

PR #4393, opened 2024-04-11
Split H5Tconv.c into modules by type

jhendersonHDF:


jhendersonHDF:
Happy to discuss other potential ways of organizing this, but this split makes it easier to find a particular conversion function, prevents H5Tconv.c from getting out of hand due to the combinations of conversions between types and also speeds up parallelized building of the library a little bit since it doesn't get hung up on H5Tconv.c. It also moves several functions, typedefs, etc. related to datatype conversions into the new H5Tconv.h header.

src/H5T.c
```diff
@@ -43,6 +43,16 @@
...
 
+/* Datatype conversion functions */
+#include "H5Tconv_char.h"
+#include "H5Tconv_short.h"
+#include "H5Tconv_int.h"
+#include "H5Tconv_long.h"
+#include "H5Tconv_llong.h"
+#include "H5Tconv_float.h"
+#include "H5Tconv_double.h"
+#include "H5Tconv_ldouble.h"
```
jhendersonHDF: Since all the conversion functions are currently only needed during H5T_init, I moved the prototypes into separate headers rather than H5Tconv.h so we only include them where we use them.

src/H5Tconv.c
qkoziol, line 1: I'd suggest moving the struct, string, bitfield, etc. conversions into their own files also.

qkoziol (approved):
Generally good, with a couple of suggestions.

src/H5Tconv.c
jhendersonHDF: Any potential name ideas? H5Tconv_soft, H5Tconv_general, ...?

qkoziol: No strong feelings, but I would suggest putting each remaining datatype class into it's own H5Tconv_<foo>.c file and just leave the utility routines and order "conversion" in H5Tconv.c

jhendersonHDF: I mostly left these in H5Tconv.c because I figured adding a new source/header combination for a single conversion function would probably be overkill, though I suppose some of them have helper functions for managing their cached private data and some have optimized versions of the functions. Really, I think it's an argument toward finally separating source files into directories by package, but that's for another day and another PR.

jhendersonHDF: > No strong feelings, but I would suggest putting each remaining datatype class into it's own H5Tconv_.c file and just leave the utility routines and order "conversion" in H5Tconv.c

Done. Reorganized so that this is strictly organized by datatype class. It moves the integer and float conversions back into single, larger files, but it's still easier to find what you're looking for compared to before. It also doesn't appear to have much of a negative effect on parallelized compile time either since the source files for the other datatype classes can be processed at the same time. Decided to do this since it then makes sense where to place the software conversion functions like "i_i" and "i_f".

qkoziol (approved):
Nice!

src/H5Tconv.c
qkoziol: Makes good sense to me, and agree that we're definitely still ahead even if the integer and float files are a little large.

src/CMakeLists.txt
```diff
@@ -610,6 +610,15 @@ set (H5T_SOURCES
...
     ${HDF5_SRC_DIR}/H5Tconv.c
+    ${HDF5_SRC_DIR}/H5Tconv_integer.c
+    ${HDF5_SRC_DIR}/H5Tconv_float.c
+    ${HDF5_SRC_DIR}/H5Tconv_string.c
+    ${HDF5_SRC_DIR}/H5Tconv_bitfield.c
+    ${HDF5_SRC_DIR}/H5Tconv_compound.c
+    ${HDF5_SRC_DIR}/H5Tconv_reference.c
+    ${HDF5_SRC_DIR}/H5Tconv_enum.c
+    ${HDF5_SRC_DIR}/H5Tconv_vlen.c
+    ${HDF5_SRC_DIR}/H5Tconv_array.c
```
derobins, line 621: The header files should be added to the list of private header files as well.

derobins (approved):

derobins:
Approved, but should add the private headers to the CMake lists so Visual Studio, etc. will show them in the IDE.

PR #4395, opened 2024-04-12
Correct doxygen GH pages folder for release

byrnHDF:


lrknox:
We realized that the links in the code to doxygen documentation all point to doxygen files in a directory that will be updated as soon as there is a daily build for the hdf5_1_14 branch, and those doxygen pages will show HDF5 version 1.14.5-1.  Also documentation will be updated for any hdf5 1.14.5 code changes as soon as they are committed.  This PR will create a dedicated and static v1_14_4 directory in github pages for the 1.14.4 documentation.  

The changes to be made are exclusively documentation.  Since the 1.14.4 release files were public since yesterday and there are documentation changes in the code (links), we believe will need to create a 1.14.4.1 release from hdf5_1.14.4.1 tag.  The hdf5_1.14.4 tag remains, and will not be affected by these changes.  The changes for the 1.14.4.1 release will remain a choice to be made until the release workflow is run.  

The Check Markdown links test will fail until the release workflow creates the new directory.

bmribler (approved):

lrknox (approved):

PR #4418, opened 2024-04-18
Update oneAPI version to 2024.1 for Daily Build

hyoklee:
2023.2 -> 2024.1


byrnHDF:
Please update all yml files that use Intel. Looks like the other files are updated, but line 508 in this file should be updated.

byrnHDF (changes requested):
Change line 508 in the build_and_test_linux_intel job too.

byrnHDF (approved):

lrknox (approved):

PR #4425, opened 2024-04-20
Add support for builtin_expect compiler hint

qkoziol:


derobins:
Do you have profiling that shows that this added complexity is worthwhile? I'm not opposed to adding it, but I don't want to add a bunch of /very/ compiler-specific complexity if this is just based on "thought experiment optimization".

qkoziol:
> Do you have profiling that shows that this added complexity is worthwhile? I'm not opposed to adding it, but I don't want to add a bunch of /very/ compiler-specific complexity if this is just based on "thought experiment optimization".

Here's a good stackoverflow that talks about the general improvements:  https://stackoverflow.com/questions/109710/how-do-the-likely-unlikely-macros-in-the-linux-kernel-work-and-what-is-their-ben

And here's a more in-depth article that measures the effect, which looks to be a 15-20% performance improvement:  http://blog.man7.org/2012/10/how-much-do-builtinexpect-likely-and.html

I think that 15-20% is worth it for the FUNC_ENTER/LEAVE statements, which are almost never going to throw errors, and the concurrency routines, which are also very unlikely to throw errors and are also latency sensitive.

src/H5private.h
```diff
@@ -1517,12 +1518,12 @@ H5_DLL herr_t H5CX_pop(bool update_dxpl_props);
 #define FUNC_LEAVE_API(ret_value)                                                                            \
     ;                                                                                                        \
     } /*end scope from end of FUNC_ENTER*/                                                                   \
-    if (api_ctx_pushed) {                                                                                    \
+    if (H5_LIKELY(api_ctx_pushed)) {                                                                         \
```
fortnern, line 1521: This will slow down FUNC_ENTER_API_NOINIT. Do we have a sense for how much? This applies to all passthrough VOL calls. Any other use cases we know of that use these NOINIT API functions heavily?
Functions (besides H5VL): H5dont_atexit, H5check_version, H5close, H5allocate_memory, H5resize_memory, H5free_memory, H5is_library_threadsafe, H5is_library_terminating, H5FDperform_init

fortnern: I guess your link suggests the simple benchmark runs slightly faster than half the speed of the default case. This is almost certainly (very slightly) good for everyone but passthrough VOLs, for whom it is (slightly more) bad, though I'm inclined to take it out.

qkoziol: Good points, I'll revert switching to H5_LIKELY(api_ctx_pushed) in FUNC_LEAVE_API

qkoziol: > Good points, I'll revert switching to H5_LIKELY(api_ctx_pushed) in FUNC_LEAVE_API

Actually, I looked at this again and don't understand how this change affects FUNC_ENTER_API_NOINIT, which uses FUNC_LEAVE_API_NOINIT, not this macro.

derobins (approved):

fortnern (approved):

src/H5private.h
fortnern: I see that now, thanks

PR #4426, opened 2024-04-20
Add new FUNC_*_REENTER macros for routines that user plugins can call

qkoziol:


src/H5private.h
```diff
@@ -1295,19 +1295,34 @@ H5_DLL herr_t H5CX_pop(bool update_dxpl_props);
...
+/*
+ * Use this macro for "developer" API functions that re-enter the library from
+ *      a plugin / connector / filter / etc, and shouldn't (need to) perform
+ *      initialization of the library or an interface, just perform tracing, etc.
+ *      Examples are: H5allocate_memory, public VOL callback
+ *      wrappers (e.g. H5VLfile_create, H5VLdataset_read, etc.), public VFL
+ *      callback wrappers (e.g. H5FDopen, H5FDwrite, etc.), etc.
+ *
+ */
+#define FUNC_ENTER_API_REENTER                                                                               \
```
mattjala, line 1309: The functions that use `_REENTER` don't have any functional changes from when they used `_NOINIT`, so why are the `_test(` wrappers now necessary to provide an API context for the FD functions?

src/H5private.h
```diff
@@ -1295,19 +1295,34 @@ H5_DLL herr_t H5CX_pop(bool update_dxpl_props);
...
- *      wrappers (e.g. H5VLfile_create, H5VLdataset_read, etc.), etc.
+ *      are: H5is_library_threadsafe, H5VLretrieve_lib_state, etc.
  *
  */
 #define FUNC_ENTER_API_NOINIT                                                                                \
     {                                                                                                        \
         {                                                                                                    \
             {                                                                                                \
                 FUNC_ENTER_API_COMMON                                                                        \
-                H5_PUSH_FUNC                                                                                 \
```
mattjala, line 1312: What's the motivation for removing utility functions that still use `_NOINIT` from the function stack?

qkoziol:
Those two changes are indeed related.  :-)   It was a bug to include pushing the function stack in the *_NOINIT macros, because they aren't supposed to perform actions that initialize the library.  However, there are functions that are called re-entrantly within the library (so any library initialization has already been performed) but should push themselves onto the function stack.  Hence, the need for the fix to the existing *_NOINIT macros and the need for the new *_REENTER macros.

qkoziol:
> Those two changes are indeed related. :-) It was a bug to include pushing the function stack in the *_NOINIT macros, because they aren't supposed to perform actions that initialize the library. However, there are functions that are called re-entrantly within the library (so any library initialization has already been performed) but should push themselves onto the function stack. Hence, the need for the fix to the existing *_NOINIT macros and the need for the new *_REENTER macros.

Perhaps unspoken in that comment: the *_NOINIT macros are used for several purposes, but the *_REENTER macros are *only* used for the re-entrant routines.

mattjala (approved):

fortnern:
I'm still unsure about making the H5FD API calls reentrant-only. Is this change needed to fix something or is it only to remove the overhead associated with the init check (just an if statement)?

qkoziol:
> I'm still unsure about making the H5FD API calls reentrant-only. Is this change needed to fix something or is it only to remove the overhead associated with the init check (just an if statement)?

Yes, this is necessary to fix pushing reentrant routines on the function stack, which has been broken for a while.

fortnern:
> > I'm still unsure about making the H5FD API calls reentrant-only. Is this change needed to fix something or is it only to remove the overhead associated with the init check (just an if statement)?
> 
> Yes, this is necessary to fix pushing reentrant routines on the function stack, which has been broken for a while.

I mean can we just leave the H5FD* functions as FUNC_ENTER_API. I can see now that this change may be necessary for passing things that are on the context but not the DXPL like (maybe) the hidden MPI types, though I would think this would be better fixed by exporting everything relevant to the DXPL when calling an external VFD. In the specific case of MPI types, that can be avoided in most cases (and is by default) using selection/vector I/O. Exceptions in parallel are I believe limited to when using the chunk cache with the file opened read only, and if there is a mixture of filtered and unfiltered datasets in a multi dataset I/O call. Are there any other cases where context info that is not on the DXPL needs to be passed through the VFD layer?

qkoziol:
> > > I'm still unsure about making the H5FD API calls reentrant-only. Is this change needed to fix something or is it only to remove the overhead associated with the init check (just an if statement)?
> > 
> > 
> > Yes, this is necessary to fix pushing reentrant routines on the function stack, which has been broken for a while.
> 
> I mean can we just leave the H5FD* functions as FUNC_ENTER_API. I can see now that this change may be necessary for passing things that are on the context but not the DXPL like (maybe) the hidden MPI types, though I would think this would be better fixed by exporting everything relevant to the DXPL when calling an external VFD. In the specific case of MPI types, that can be avoided in most cases (and is by default) using selection/vector I/O. Exceptions in parallel are I believe limited to when using the chunk cache with the file opened read only, and if there is a mixture of filtered and unfiltered datasets in a multi dataset I/O call. Are there any other cases where context info that is not on the DXPL needs to be passed through the VFD layer?

Yes, the API context is the main reason for not using FUNC_ENTER_API.  The error stack and other "global" state that was formerly in its own thread-local variable is moved to the API context in my threading_updates branch, since there's a limited # of thread-local variables, especially on Windows.  And also the locking protocol that's on the way needs to know the difference between a "real" (top-level?) API call and a reentrant one.   So, multiple reasons for moving the H5FD developer routines to use this macro pair.

qkoziol:
Following up on our discussion at the community meeting earlier today, I reviewed the difference between the FUNC_*_API and FUNC_*_API_REENTER macro pairs.  It's not just the API context that is getting pushed/popped, it's also the error stack getting cleared/reset, and the function stack getting pushed/popped.  So, multiple problems with using the FUNC_*_API macros for these routines.

Wrapping all of those push/pops in an 'if' (based on the API context already being pushed, say) would be possible, but seems open to confusion for users and would impose requirements on the existing code that weren't planned for (e.g. the MPI datatype in the API context).   If we want that for these kind of reentrant routines, we should do the full design process in that direction, and maybe add new API routines that are explicitly not for reentrant use.

fortnern:
> Following up on our discussion at the community meeting earlier today, I reviewed the difference between the FUNC___API and FUNC___API_REENTER macro pairs. It's not just the API context that is getting pushed/popped, it's also the error stack getting cleared/reset, and the function stack getting pushed/popped. So, multiple problems with using the FUNC_*_API macros for these routines.
> 
> Wrapping all of those push/pops in an 'if' (based on the API context already being pushed, say) would be possible, but seems open to confusion for users and would impose requirements on the existing code that weren't planned for (e.g. the MPI datatype in the API context). If we want that for these kind of reentrant routines, we should do the full design process in that direction, and maybe add new API routines that are explicitly not for reentrant use.

What are the new requirements for existing code? Wouldn't MPI datatypes be passed through correctly since the library will detect it's being called in a reentrant manner and will avoid pushing a new context? It seems to me we might want to apply this everywhere, since almost any API can be called in a reentrant manner.

For now, would this PR work with the changes to the H5FD functions taken out (temporarily)?

qkoziol:
I've been working on the locking protocol, API context, and error reporting more over the weekend and the approach in this PR may need to change.   I'll give it some more thought today and post another update.

qkoziol:
It looks like the locking protocol for reentrancy may require a new API context and possibly error stack for all types of reentrancy into the library (with normal API routines as well as these developer routines), along with some other kind of global (thread-local) state to manage the locking correctly.   So, I'm going to close this PR while I work that out.

qkoziol:
Closing to develop the concurrency design further.

PR #4439, opened 2024-04-23
Move 1.14.4-2 notes to release history

glennsong09:


hyoklee (approved):

lrknox (changes requested):
The file is correct, except that RELEASE.txt contents from the 1.14.4-2 release should be added to the hdf5_1_14 support branch, not to the release branch.  I'm sorry that wasn't clear.  It has been added to the RELEASE_PROCESS.md file. I think this PR will need to be closed and a new one created to hdf5_1_14.

The Check Markdown links failures for this file are not a concern because 1. the dead links are not in this file, and 2. what goes in HISTORY files is what was released.  fortunately there are no dead links in RELEASE.txt, because adding dead links to the HISTORY file in hdf5_1_14 would raise an issue that we would need to address otherwise.

PR #4446, opened 2024-04-29
H5R Fortran wrappers and misc. H5R API/DOC updates

brtnfld:
Added missing newest reference 3 H5R APIs, fixed C H5R issue and added test and doc updates related to H5R.
Please take a look at Release.txt updates for details.


src/H5Rpublic.h
```diff
@@ -524,7 +532,8 @@ H5_DLL herr_t H5Rget_obj_type3(H5R_ref_t *ref_ptr, hid_t rapl_id, H5O_type_t *ob
  *          application. If the length of the name, which determines
  *          the required value of size, is unknown, a preliminary
  *          H5Rget_file_name() call can be made. The return value of this
- *          call will be the size of the file name. That value can then be
+ *          call will be the size of the file name (the size does not include
+ *          the \p NULL terminator). That value can then be
  *          passed in for size in the second call to H5Rget_file_name(),
```
jhendersonHDF: Maybe clarify here for each of these that the return value can't be directly passed in since it doesn't include the NUL terminator. I think some of the other docs say something similar to "that size + 1"

brtnfld: Ok, I just copied the text from H5Rget_name. We also seem to mix "NULL" with \0 string in the docs. Some other functions don't call out if size includes this at all. Is there a particular description from another API that you prefer? It should probably mention when NULL is passed in as name.

brtnfld: nvm, I think I found an example from H5P. Since this is a common case, I will create a doxygen ALIAS.

brtnfld: added alias for doxygen.

byrnHDF (approved):

epourmal (approved):

PR #4451, opened 2024-05-01
Convert ERR test to use grep

byrnHDF:
This is to fix 2 tests failing on Lassen because h5stat_notexist and h5dump-nonexisting tests that return expected error messages also trigger irrelevant system warning text that use of Infiniband is deprecated in favor of UCX.
Removed all use of *.err files (except for format_convert) by grepping for expected phrase in error output. This eliminates the need to filter out system error messages.


lrknox (approved):

lrknox (approved):

derobins:
Would the Autotools tests fail since they'd still use the .err files?

byrnHDF:
> Would the Autotools tests fail since they'd still use the .err files?

No, this just eliminates the copy of the files and the compare of the error output against them.
Once autotools is fixed then the files can be deleted

lrknox (approved):

jhendersonHDF:
I think the CUDA warnings are coming about because the MPI wrappers being used in the NVHPC action are the ones under the `openmpi4` directory rather than the more commonly-used `openmpi` directory. The ones under the `openmpi4` seem to end up with a link to CUDA, while the ones under the `openmpi` directory often don't.

derobins (approved):

PR #4454, opened 2024-05-01
Removed "function/code stack" debugging configure option

qkoziol:
Easily replaced w/third-party tools, e.g. libbacktrace (https://github.com/ianlancetaylor/libbacktrace)


derobins:
This should probably not be merged to 1.14, even though it's removing an obscure, broken feature. We can discuss in the working group, though.

qkoziol:
> This should probably not be merged to 1.14, even though it's removing an obscure, broken feature. We can discuss in the working group, though.

Agree - no sense disrupting the branch, even if it's not working.

qkoziol:
The "free list tracking" code was also removed, since it was the reason that the function stack was added originally.  (And using libbacktrace can avoid all this framework)

derobins (approved):

derobins:
We decided this feature could be removed from 1.14 since it's currently broken (and didn't even compile until recently). Anyone who cared would have complained by now.

jhendersonHDF (approved):

PR #4456, opened 2024-05-02
Fix bug in MPI-IO VFD

qkoziol:


src/H5FDmpio.c
```diff
@@ -1951,7 +1951,7 @@ H5FD__mpio_vector_build_types(uint32_t count, H5FD_mem_t types[], haddr_t addrs[
 
 done:
     /* free sorted vectors if they exist */
-    if (!vector_was_sorted)
```
jhendersonHDF, line 1954: Not necessarily related to this PR, but it looks like the assertion `assert(vector_was_sorted);` on line 1807 above was probably meant to be `assert(*vector_was_sorted);`

jhendersonHDF (approved):

src/H5FDmpio.c
qkoziol: Yes, I agree, updated.

derobins (approved):

PR #4458, opened 2024-05-02
Fixed failures with xl compilers.

brtnfld:
Fixed issues with XL compilers when building with promoted reals and integers.


lrknox (approved):
Change fixed compile failure on echidna - passed all tests.

byrnHDF (approved):

PR #4459, opened 2024-05-02
Fix leak of internal IDs registered during compound datatype conversion

jhendersonHDF:


jhendersonHDF:
Will add some more fixes to this PR after testing, based on feedback in https://github.com/h5py/h5py/issues/2419

jhendersonHDF:
Before the type conversion optimization changes in the 1.14.4 release, the previous behavior of the library's compound datatype conversion function was to always register IDs for the members of the source and destination compound datatypes once during conversion path initialization and cache those IDs. After those changes, the registration of IDs was deferred until after the conversion path for each of the source compound members being converted was determined. That way, the library could determine if it needs to register IDs based on whether any of the conversion functions used is an application conversion function or whether a conversion exception function was provided.

However, this created a leak of these IDs when the compound conversion function recalculates cached data after the library's table of conversion functions is modified. Since this resulted in multiple IDs pointing to the same `H5T_t` structure, an application could encounter crashes, assertion failures or use-after-free issues, depending on the build type and whether or not free lists are enabled.

src/H5T.c
jhendersonHDF, line 1: The changes in this file are for making the library handle unlocking/closing of partially initialized datatypes more gracefully

src/H5T.c
```diff
@@ -1874,7 +1873,6 @@ H5T__close_cb(H5T_t *dt, void **request)
 
     /* Sanity check */
     assert(dt);
-    assert(dt->shared);
```
jhendersonHDF, line 1877: Handled further down when `dt` is passed to `H5T_close`

byrnHDF (approved):

src/H5Tconv_compound.c
```diff
@@ -264,21 +264,32 @@ H5T__conv_struct_init(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, co
...
-                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTREGISTER, FAIL,
-                                "can't register ID for source compound member datatype");
+                /* Only register new IDs for the source and destination member datatypes
+                 * if IDs weren't already registered for them. If the cached conversion
+                 * information has to be recalculated (in the case where the library's
+                 * table of conversion functions is modified), the same IDs can be reused
+                 * since the only information that needs to be recalculated is the conversion
+                 * paths used.
+                 */
+                if (priv->src_memb_id[i] == H5I_INVALID_HID) {
```
bmribler, line 274: Is it a guarantee that the memb_id's are initialized to H5I_INVALID_HID?

jhendersonHDF: Yes, they are initialized a bit above around line 174 and 180

lrknox (approved):

PR #4461, opened 2024-05-03
Remove use of err files in autotools test scripts

byrnHDF:
Format Convert in CMake and Autotools remains to be solved


hyoklee (approved):

qkoziol:
What will replace this kind of testing?

lrknox:
> What will replace this kind of testing?

The .err files contained 1 line error messages and the tests were comparing the output files with the expected .err file.  However, the HPC machines sometimes put system messages in the output in addition to the error message, in which case the output doesn't match the expected error message, even though the expected error message is in the output.  

The replacement for the error message in the .err file is to grep the output for the significant text in the error message.  This can be seen in the lines with "GREPTEST ERRTXT" added to the .sh.in files.  The same change has previously been added to CTestTest.cmake files for CMake.

lrknox (approved):

PR #4464, opened 2024-05-06
Use ADD_H5_ERR_TEST to not compare output

byrnHDF:


lrknox (approved):

lrknox:
Fixes H5STAT-h5stat_noexist failure on Lassen

bmribler (approved):

PR #4468, opened 2024-05-09
Update for blosc2 in plugins and prefix hdf5 cmake varnames

byrnHDF:
Java junit jar was out of date, updated from 4.11 (CVE issue) to 4.13.2


make.log
derobins, line 1: Does this belong here?

byrnHDF: fixed

lrknox (approved):

jhendersonHDF (approved):

PR #4475, opened 2024-05-10
Pause recording errors instead of clearing the error stack

qkoziol:
An internal capability that's similar to the H5E_BEGIN_TRY / H5E_END_TRY macros in H5Epublic.h, but more efficient since we can avoid pushing errors on the stack entirely (and those macros use public API routines).

This capability (and other techniques) can be used to remove use of H5E_clear_stack() and H5E_BEGIN_TRY / H5E_END_TRY within library routines.

We want to remove H5E_clear_stack() because it can trigger calls to the H5I interface from within the H5E code, which creates a great deal of complexity for threadsafe code.  And we want to remove H5E_BEGIN_TRY / H5E_END_TRY's because they make public API calls from within the library code.

Also some other minor tidying in routines related to removing the use of H5E_clear_stack() and H5E_BEGIN_TRY / H5E_END_TRY from H5Fint.c


qkoziol:
I have a change to make for this, please hold off on reviewing for the moment.   I should have something up later today.

qkoziol:
> I have a change to make for this, please hold off on reviewing for the moment. I should have something up later today.

OK, I've updated things to a cleaner solution.  Please go ahead with reviews now.

qkoziol:
Interesting - removing the H5E_clear_stack() from the virtual dataset storage code has revealed a failure in the VDS SWMR test.   I'm looking into it, but it's another motivator to get the clear stack calls out of the library.

jhendersonHDF:
How does this interact with the old way of "pausing" the error stack by just removing the callback function temporarily? I'm thinking that this PR might need to add a way of publicly retrieving whether an error stack has been paused. The error macros I used in different VFDs/VOLs currently decide whether to suppress that driver's/connector's own error reporting based on whether the default error stack currently has a callback function set, so with these changes it seems I'd likely also need to check if the error stack has been paused to be correct.

qkoziol:
> How does this interact with the old way of "pausing" the error stack by just removing the callback function temporarily? I'm thinking that this PR might need to add a way of publicly retrieving whether an error stack has been paused. The error macros I used in different VFDs/VOLs currently decide whether to suppress that driver's/connector's own error reporting based on whether the default error stack currently has a callback function set, so with these changes it seems I'd likely also need to check if the error stack has been paused to be correct.

Setting the reporting callback to NULL just suppresses error reports from being printed when leaving an API routine, it doesn't pause actually pushing errors on the stack.  This change allows the library to internally pause pushing errors on the stack in circumstances when it knows the operation could fail and would just call H5E_clear_stack() on a failure.

qkoziol:
Currently, I don't see a need for an API for this, but I'm open to ideas around one.

qkoziol:
> Interesting - removing the H5E_clear_stack() from the virtual dataset storage code has revealed a failure in the VDS SWMR test. I'm looking into it, but it's another motivator to get the clear stack calls out of the library.

I've put up a PR (https://github.com/HDFGroup/hdf5/pull/4489) that addresses the SWMR bug, and cherry-picked that commit into this branch, so the tests should pass now.

src/H5Fint.c
```diff
@@ -1850,56 +1860,54 @@ H5F_open(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id)
...
-     *
-     * However, if stripping away the H5F_ACC_CREAT flag and others left us
-     * with the same file access flags as before, then we will skip this
-     * tentative file open and only make a single attempt at opening the file.
-     * In this case, we don't want to suppress error output since the underlying
-     * file driver might provide more details on why the file open failed.
-     */
-    if (tent_flags != flags) {
-        /* Make tentative attempt to open file */
-        H5E_BEGIN_TRY
```
jhendersonHDF, line 1881: As an example, I believe switching from `H5E_BEGIN_TRY` -> `H5E_PAUSE_ERRORS` (inside `H5FD_try_open`) here will cause some VOL connectors/VFDs to start printing out error stacks from failures during speculative file opens because, at the time, the easiest way to determine whether or not to print an error stack was to check if the library's default error stack has error reporting disabled. Since I setup these connectors/drivers with their own error stack separate from the library, it generally only makes sense for that plugin to print out the contents of its own error stack on the way out from a "top-level" function (e.g., one of the callback functions the plugin implements). The error macros I setup check whether the library's default error stack has a callback function set before deciding whether to print out the error stack on the way out. With these changes, it seems I'd instead need to know whether error reporting on the error stack has been paused.

qkoziol: OK, that makes sense.

I can add an API routine (H5Eis_paused ?) that will allow plugin developers to query if the library has paused error reporting on the default stack.  I don't think it's a good idea to let those developers unpause the default stack, but I could also add a H5Epause & H5Eresume for them to have similar behavior for their own stacks.

I'll try to get these additional routines added tomorrow.

qkoziol: Added new routines

qkoziol:
@jhendersonHDF  - How do these new API routines look?

src/H5E.c
```diff
@@ -889,3 +889,114 @@ H5Eappend_stack(hid_t dst_stack_id, hid_t src_stack_id, hbool_t close_source_sta
...
+herr_t
+H5Epause_stack(hid_t stack_id)
+{
+    H5E_stack_t *stack;               /* Error stack */
+    herr_t       ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_API(FAIL)
+
+    /* Applications can't pause the default stack */
+    if (stack_id == H5E_DEFAULT)
```
jhendersonHDF: Hmm, I think this sort of comes down to deciding if we want to have multiple methods to essentially disable error reporting, but it seems to me that `H5Epause/resume_stack` on the default error stack is a reasonable use case. Otherwise, it seems like there's a bit of a disconnect where internal code is intended to use the internal `H5E_pause_stack` to temporarily disable error reporting, while "external" code has to use `H5Eget/set_auto`, which developers have to remember. Is there harm in allowing an application to call `H5Epause/resume_stack` on the default error stack and promoting that as a more natural way of temporarily disabling error reporting, as opposed to just temporarily NULLing out the printing callback?

qkoziol: I remember thinking "I don't think this will work from user space" when I was writing it, so I disabled it.   But, after your prompt, I went back and think it should be OK.  I think my main concern is that a plugin could manipulate the default error stack's pause status (unpause it, or make it paused when it returns), but that's true for the get_auto/set_auto also.

I'll update the code to allow it, and also switch the H5E_BEGIN_TRY / H5E_END_TRY macros to pause & resume pushing errors on the stack, since that's more logical than just disabling the reporting (as you say).

qkoziol: Now that I think about it a little more - I can add some code to save & restore library state around making a call to a plugin's function pointer (starting with the error stack reporting routine and pause status).   We're going to need that concept when we start converting routines that make plugin callbacks to be concurrent, and it's not bad to start now.

qkoziol: > I remember thinking "I don't think this will work from user space" when I was writing it, so I disabled it. But, after your prompt, I went back and think it should be OK. I think my main concern is that a plugin could manipulate the default error stack's pause status (unpause it, or make it paused when it returns), but that's true for the get_auto/set_auto also.
> 
> I'll update the code to allow it, and also switch the H5E_BEGIN_TRY / H5E_END_TRY macros to pause & resume pushing errors on the stack, since that's more logical than just disabling the reporting (as you say).

OK, I've pushed an update to allow pausing and resuming the default error stack.   I decided not to change the H5E_BEGIN_TRY / H5E_END_TRY macros, since that would modify application visible behavior.

qkoziol: > Now that I think about it a little more - I can add some code to save & restore library state around making a call to a plugin's function pointer (starting with the error stack reporting routine and pause status). We're going to need that concept when we start converting routines that make plugin callbacks to be concurrent, and it's not bad to start now.

I haven't made this change yet, since I want to think about it a little more.  Generally speaking, I'm planning to add the state of the default error stack to the information retrieved and restored with H5CX_retrieve_state / H5CX_restore_state, but I'm not 100% certain yet.  There's no urgency to this defensive state saving, so I've put it on my to-do list to tackle in a little while.

qkoziol:
@jhendersonHDF - I believe I've covered the changes necessary to allow plugin authors to use the error pause / resume capability correctly, so this should be ready for another review.

qkoziol:
@derobins - Anything holding this back from being merged?   There's other H5E_clear_stack() & H5E_BEGEN_TRY / H5E_END_TRY usage that I'd like to get started on, but it'll be based on the capabilities here.

jhendersonHDF:
> @derobins - Anything holding this back from being merged? There's other H5E_clear_stack() & H5E_BEGEN_TRY / H5E_END_TRY usage that I'd like to get started on, but it'll be based on the capabilities here.

I'll re-review a bit later this afternoon and see if we can get this merged. I don't think there were any outstanding issues. I'll just have to remember to go revisit the VOLs and VFDs that use H5E mechanisms.

qkoziol:
> > @derobins - Anything holding this back from being merged? There's other H5E_clear_stack() & H5E_BEGEN_TRY / H5E_END_TRY usage that I'd like to get started on, but it'll be based on the capabilities here.
> 
> I'll re-review a bit later this afternoon and see if we can get this merged. I don't think there were any outstanding issues. I'll just have to remember to go revisit the VOLs and VFDs that use H5E mechanisms.

Super, thanks

src/H5Epublic.h
```diff
@@ -322,6 +323,75 @@ H5_DLL hid_t H5Eget_current_stack(void);
...
+ *          The library may pause pushing errors on the default error stack
+ *          when performing "speculative" operations, such as testing for the
+ *          existence of something that could be located at one of many
+ *          locations. \p stack_id is the error stack to query, and the value
+ *          pointed to by \p is_paused is set to TRUE/FALSE.
+ *
+ *          If an error occurs while attempting to query the status of \p stack_id,
+ *          the value pointed to by \p is_paused is unchanged.
+ *
+ * \since 1.14.4
```
lrknox: I think these will first appear in version 1.14.5, and so should be "\since 1.14.5"?

qkoziol: OK, I'll update

qkoziol: Done

src/H5Eint.c
```diff
@@ -1791,3 +1790,79 @@ H5E_dump_api_stack(void)
...
+H5E_resume_stack(void)
+{
+    H5E_stack_t *estack = H5E__get_my_stack();
+
+    FUNC_ENTER_NOAPI_NOERR
+
+    assert(estack);
+
+    /* Decrement pause counter */
+    estack->paused--;
```
jhendersonHDF, line 1866: I'm thinking it would probably be a good idea to make sure this isn't 0 before decrementing it so that a `resume_stack` call doesn't accidentally pause the stack.

src/H5Epublic.h
```diff
@@ -322,6 +323,75 @@ H5_DLL hid_t H5Eget_current_stack(void);
...
+ *          existence of something that could be located at one of many
+ *          locations. \p stack_id is the error stack to query, and the value
+ *          pointed to by \p is_paused is set to TRUE/FALSE.
+ *
+ *          If an error occurs while attempting to query the status of \p stack_id,
+ *          the value pointed to by \p is_paused is unchanged.
+ *
+ * \since 1.14.5
+ */
+H5_DLL herr_t H5Eis_paused(hid_t stack_id, hbool_t *is_paused);
```
jhendersonHDF, line 352: I think going forward we just want to use plain `bool` instead of `hbool_t`

src/H5Epublic.h
```diff
@@ -322,6 +323,75 @@ H5_DLL hid_t H5Eget_current_stack(void);
...
+ * \estack_id{stack_id}
+ * \return \herr_t
+ *
+ * \details H5Epause_stack() pauses pushing errors on an error stack.  Pushing
+ *          an error on a paused error stack will be ignored (not fail).
+ *
+ *          H5Eresume_stack() is used to allow errors to be pushed on a stack.
+ *          Calls to H5Epause_stack() and H5Eresume_stack() must be matched.
+ *
+ *          Calls to H5Epause_stack()/H5Erusume_stack() may be nested.
```
jhendersonHDF: `H5Erusume_stack` -> `H5Eresume_stack`

src/H5Epublic.h
```diff
@@ -322,6 +323,75 @@ H5_DLL hid_t H5Eget_current_stack(void);
...
+ *
+ *          It is erroneous to attempt to pause the default error stack.
```
jhendersonHDF: This is allowed now

src/H5Epublic.h
```diff
@@ -322,6 +323,75 @@ H5_DLL hid_t H5Eget_current_stack(void);
...
+ * \estack_id{stack_id}
+ * \return \herr_t
+ *
+ * \details H5Eresume_stack() resumes pushing errors on an error stack.
+ *
+ *          Calls to H5Epause_stack() and H5Eresume_stack() must be matched.
+ *
+ *          Calls to H5Epause_stack()/H5Erusume_stack() may be nested.
+ *
+ *          It is erroneous to attempt to resume the default error stack.
```
jhendersonHDF: This is allowed now

src/H5Epublic.h
```diff
@@ -322,6 +323,75 @@ H5_DLL hid_t H5Eget_current_stack(void);
...
+ * \brief * Resume pushing errors on an error stack
+ *
+ * \estack_id{stack_id}
+ * \return \herr_t
+ *
+ * \details H5Eresume_stack() resumes pushing errors on an error stack.
+ *
+ *          Calls to H5Epause_stack() and H5Eresume_stack() must be matched.
+ *
+ *          Calls to H5Epause_stack()/H5Erusume_stack() may be nested.
```
jhendersonHDF: `H5Erusume_stack` -> `H5Eresume_stack`

src/H5E.c
```diff
@@ -889,3 +889,118 @@ H5Eappend_stack(hid_t dst_stack_id, hid_t src_stack_id, hbool_t close_source_sta
...
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Eis_paused(hid_t stack_id, hbool_t *is_paused)
+{
+    H5E_stack_t *stack;               /* Error stack */
+    herr_t       ret_value = SUCCEED; /* Return value */
+
+    /* Don't clear the error stack! :-) */
```
bmribler, line 908: Unnecessary comment?

src/H5FD.c
```diff
@@ -701,66 +701,77 @@ H5FDopen(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxaddr)
...
-    /* Dispatch to file driver */
     if (HADDR_UNDEF == maxaddr)
         maxaddr = driver->maxaddr;
-    if (NULL == (file = (driver->open)(name, flags, fapl_id, maxaddr)))
-        HGOTO_ERROR(H5E_VFL, H5E_CANTINIT, NULL, "open failed");
+
+    /* Try dispatching to file driver */
+    H5E_PAUSE_ERRORS
+    {
+        file = (driver->open)(name, flags, fapl_id, maxaddr);
```
jhendersonHDF: Shouldn't this function have some method of telling it whether or not to pause errors when dispatching to the file driver? Otherwise, since several places like `H5F_open` are being switched to using this new function, it's just always going to hide errors when file opens fail.

src/H5Fint.c
```diff
@@ -1850,56 +1860,54 @@ H5F_open(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id)
...
      * the error stack and reset the file access flags, then make another
      * attempt at opening the file.
      */
-    if ((tent_flags == flags) || (lf == NULL)) {
-        if (tent_flags != flags) {
-            H5E_clear_stack();
-            tent_flags = flags;
-        }
+    if (NULL == lf) {
+        if (H5FD_try_open(&lf, name, tent_flags, fapl_id, HADDR_UNDEF) < 0)
```
jhendersonHDF: This is a case where I'd think rather than switching to a try_open we'd want to keep the same behavior of just a simple open since this is the final open attempt and we're interested in knowing if something fails.

test/error_test.c
```diff
@@ -685,6 +682,166 @@ test_append(void)
...
+    is_paused = TRUE;
+    H5E_BEGIN_TRY
+    {
+        ret = H5Eis_paused(H5I_INVALID_HID, &is_paused);
+    }
+    H5E_END_TRY
+    if (ret >= 0)
+        TEST_ERROR;
+
+    /* Check if default stack is paused */
```
bmribler: Check if default stack is not paused?

test/error_test.c
```diff
@@ -685,6 +682,166 @@ test_append(void)
...
+    is_paused = TRUE;
+    if (H5Eis_paused(H5E_DEFAULT, &is_paused) < 0)
+        TEST_ERROR;
+    if (FALSE != is_paused)
+        TEST_ERROR;
+
+    /* Check if application stack is paused */
```
bmribler: Same as above, and for below.

bmribler (approved):
Just a few minor comments

src/H5Eint.c
qkoziol: Sure, I'll add an assert.

src/H5Epublic.h
qkoziol: I'm fine either way, but I thought we were trying to be consistent for now and switch them all at once.

@derobins @brtnfld - Any opinions about this?

src/H5FD.c
qkoziol: Yes, that's true - it always hides errors now, because of the changes to the H5F code.  The previous H5F code was already accommodating those failures by clearing the stack when it failed.   This change, and the changes to the H5F code, moves that down here and accommodates for the change here with more straightforward logic in the H5F code.

src/H5Fint.c
qkoziol: No, I believe that this is correct now.  This is in H5F_try_open and the code that calls this routine is built around trying the open, so that errors are only generated when there's no more opportunities for trying to open the file.   It's not an error to try to open the file and not succeed, the file pointer is just set to NULL, and the calling routine can determine what action it wants to take, which could involve trying again.

For example,  H5F_try_open is called from H5F__efc_try_open_file, which is called by H5F__efc_try_open, which is called in H5F_prefix_try_open() several times, to attempt opening the file with various paths, etc.

The logic for these new "try open" and "open" calls all follow the same pattern, to avoid clearing the error stack unnecessarily.

src/H5E.c
qkoziol: It's a reminder that the special "NOCLEAR" macro was intentional.  (All the other uses have the same reminder)

bmribler: I know it's silly, but I meant the ":-)"  It doesn't matter though.

src/H5Fint.c
jhendersonHDF: The previous code for `H5F_open`, for example, was clearing the error stack after the speculative file open fails, which is perfectly reasonable, but it wasn't clearing the error stack after the final file open attempt fails so that any errors from the VFD level can be captured. Since `H5F_open` is changed here to just call `H5F_try_open` once, which does a `H5FD_try_open` for both file open attempts, it seems like the error stack will be paused for both file open attempts and code that calls `H5F_open` won't be able to capture errors from the VFD used.

src/H5FD.c
jhendersonHDF: I get that one would want to hide errors for speculative file open attempts, but it seems like code (like `H5F_open`) that was previously performing both a speculative and non-speculative file open has been changed to only make use of these try_open calls, in essence making two speculative file open attempts.

jhendersonHDF: I should probably clarify that I'm talking about errors from the VOL/VFD level being hidden, not necessarily errors from the library's default error stack. For example, running a simple program that tries to open an HDF5 file which doesn't exist before and after these changes gives:

```C
#include "hdf5.h"

int
main(int argc, char **argv)
{
    H5Fopen("nonexist.h5", H5F_ACC_RDWR, H5P_DEFAULT);
    return 0;
}
```

Before
```
HDF5-DIAG: Error detected in HDF5 (1.15.0):
  #000: /home/jhenderson/git/hdf5/src/H5F.c line 827 in H5Fopen(): unable to synchronously open file
    major: File accessibility
    minor: Unable to open file
  #001: /home/jhenderson/git/hdf5/src/H5F.c line 788 in H5F__open_api_common(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #002: /home/jhenderson/git/hdf5/src/H5VLcallback.c line 3689 in H5VL_file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #003: /home/jhenderson/git/hdf5/src/H5VLcallback.c line 3501 in H5VL__file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #004: /home/jhenderson/git/hdf5/src/H5VLnative_file.c line 128 in H5VL__native_file_open(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #005: /home/jhenderson/git/hdf5/src/H5Fint.c line 1901 in H5F_open(): unable to open file: name = 'nonexist.h5', tent_flags = 1
    major: File accessibility
    minor: Unable to open file
  #006: /home/jhenderson/git/hdf5/src/H5FD.c line 763 in H5FD_open(): open failed
    major: Virtual File Layer
    minor: Unable to initialize object
  #007: /home/jhenderson/git/hdf5/src/H5FDsec2.c line 323 in H5FD__sec2_open(): unable to open file: name = 'nonexist.h5', errno = 2, error message = 'No such file or directory', flags = 1, o_flags = 2
    major: File accessibility
    minor: Unable to open file
```

After
```
HDF5-DIAG: Error detected in HDF5 (1.15.0):
  #000: /home/jhenderson/HDF5_Workspace/Worktrees/test_error/src/H5F.c line 827 in H5Fopen(): unable to synchronously open file
    major: File accessibility
    minor: Unable to open file
  #001: /home/jhenderson/HDF5_Workspace/Worktrees/test_error/src/H5F.c line 788 in H5F__open_api_common(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #002: /home/jhenderson/HDF5_Workspace/Worktrees/test_error/src/H5VLcallback.c line 3689 in H5VL_file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #003: /home/jhenderson/HDF5_Workspace/Worktrees/test_error/src/H5VLcallback.c line 3501 in H5VL__file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #004: /home/jhenderson/HDF5_Workspace/Worktrees/test_error/src/H5VLnative_file.c line 128 in H5VL__native_file_open(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #005: /home/jhenderson/HDF5_Workspace/Worktrees/test_error/src/H5Fint.c line 2247 in H5F_open(): can't open file
    major: File accessibility
    minor: Unable to open file
```

src/H5Fefc.c
```diff
@@ -231,54 +272,46 @@ H5F__efc_open(H5F_efc_t *efc, const char *name, unsigned flags, hid_t fcpl_id, h
...
-
-                HGOTO_DONE(ret_value);
+                HGOTO_DONE(SUCCEED);
             } /* end else */
         }     /* end if */
         else
             /* Allocate new entry */
             if (NULL == (ent = H5FL_MALLOC(H5F_efc_ent_t)))
-                HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, NULL, "memory allocation failed");
+                HGOTO_ERROR(H5E_FILE, H5E_CANTALLOC, FAIL, "memory allocation failed");
```
bmribler, line 319: Question: why is H5E_FILE instead of H5E_RESOURCE in this case?

qkoziol: Because the error occurred in a file module, not in a memory manager (i.e. "resource") module.

src/H5Fint.c
qkoziol: @jhendersonHDF - I updated to make the changes we talked about yesterday.  How's this look now?

derobins (changes requested):
Jordan's comments need to be addressed before this can be merged

jhendersonHDF:
Will do some more testing here to see if there's anything left to do for this PR.

jhendersonHDF (approved):

jhendersonHDF:
Minor new conflict from https://github.com/HDFGroup/hdf5/pull/4568, but this looks good to me now and local testing shows error stacks are retained at least from the H5F_open -> VFD path.

qkoziol:
> Minor new conflict from #4568, but this looks good to me now and local testing shows error stacks are retained at least from the H5F_open -> VFD path.

I've fixed the confilct; thanks for the review!  Back to @derobins now :-)

qkoziol:
@derobins - This should be ready to go, Jordan's approved.

src/H5Epublic.h
derobins: I think we use both in the API and the next major release will drop hbool_t from the API entirely (though the typdef will remain).

derobins (approved):

derobins:
Needs a RELEASE.txt entry, but otherwise this can be merged

src/H5Epublic.h
derobins: hbool_t is fine

Issue #4433, opened 2024-04-21
Segmentation fault in H5Oint.c


gabe-sherman:
A segmentation fault occurs in the below program when provided with a malformed input. This behavior occurs at line 1070 in H5Oint.c

```
#include <stdio.h>
#include <stdarg.h>
#include <string.h>
#include <stdlib.h>
#include <hdf5.h>

int main(int argc, char *argv[])
{
   hid_t v0 = H5Fopen(argv[1], H5F_ACC_RDONLY, H5P_DEFAULT);
   if (v0 == H5I_INVALID_HID) exit(1);
   H5G_info_t * v1;
   H5Gget_info(v0, v1);
   return 0;
}

```

### How to trigger

```./filename poc```

### POC File
https://github.com/FuturesLab/POC/blob/main/hdf5/poc-08

### Test Environment

Ubuntu 22.04, 64bit
### Version

Latest: 0394b03f66dc45fe96e2c772b7bce293e4316ad2

### Address Sanitizer Output
```
=================================================================
==1383990==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x555556fd85da bp 0x617000000508 sp 0x7fffffffd0a0 T0)
==1383990==The signal is caused by a READ memory access.
==1383990==Hint: address points to the zero page.
    #0 0x555556fd85da in H5O__assert /home/gabesherman/harness_test/AutoHarn-Evaluation/hdf5/lib_asan/src/H5Odbg.c:147:25
    #1 0x55555705391f in H5O_protect /home/gabesherman/harness_test/AutoHarn-Evaluation/hdf5/lib_asan/src/H5Oint.c:1071:5
    #2 0x5555570ad597 in H5O_msg_exists /home/gabesherman/harness_test/AutoHarn-Evaluation/hdf5/lib_asan/src/H5Omessage.c:787:23
    #3 0x555556ca3105 in H5G__open_oid /home/gabesherman/harness_test/AutoHarn-Evaluation/hdf5/lib_asan/src/H5Gint.c:526:62
    #4 0x555556ca3105 in H5G_open /home/gabesherman/harness_test/AutoHarn-Evaluation/hdf5/lib_asan/src/H5Gint.c:445:13
    #5 0x555556ced28a in H5G__obj_info /home/gabesherman/harness_test/AutoHarn-Evaluation/hdf5/lib_asan/src/H5Gobj.c:708:24
[...]

PR #4477, opened 2024-05-10
Properly clean up cache when failing to load an object header

fortnern:
Addresses #4433 

Since an object header is split into multiple cache objects but treated as a single object in the object header code, it is possible, if the cache load of the object header fails, for some of the object header objects to remain in cache pointing to an incomplete/inconsistent H5O_t struct. Modified H5O_protect to expunge all chunks and delete the main object header chunk if it fails for any reason. Also modified H5G__open_oid() to avoid the second call to H5O_msg_exists() if the first returns an error.

### DIFF
--- a/src/H5Gint.c
+++ b/src/H5Gint.c
@@ -506,6 +506,7 @@ static herr_t
 H5G__open_oid(H5G_t *grp)
 {
     bool   obj_opened = false;
+    htri_t msg_exists;
     herr_t ret_value  = SUCCEED;
 
     FUNC_ENTER_PACKAGE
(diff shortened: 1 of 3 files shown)


hyoklee (approved):

qkoziol:
H5HL_protect() is similar (w/two cache components), do you think it needs a similar change?

qkoziol (approved):

fortnern:
> H5HL_protect() is similar (w/two cache components), do you think it needs a similar change?

From a quick reading of the HL code, it looks like, while it doesn't remove the prefix from cache in case of datablock decoding failure, it should handle such a failure gracefully and leave the heap struct in a consistent state. It might still be a good idea to remove the prefix from cache.

jhendersonHDF (approved):

PR #4487, opened 2024-05-14
Add option to use zlib-ng as zlib library

byrnHDF:
Rework compression libs CMake handling
Update to latest libaec version


CMakeFilters.cmake
```diff
@@ -9,14 +9,23 @@
 # If you do not have access to either file, you may request a copy from
 # help@hdfgroup.org.
 #
-option (USE_LIBAEC_STATIC "Use static AEC library " OFF)
+option (USE_ZLIB_NG "Use zlib-ng library as zlib library" OFF)
```
jhendersonHDF: Seems like we should probably be prefixing these options as in `HDF5_USE_ZLIB_NG`

byrnHDF: Yes - good point

lrknox (changes requested):
Libaec 1.0.6 should still work?

byrnHDF:
> Libaec 1.0.6 should still work?

Need to test but yes it should

lrknox (approved):

release_docs/INSTALL_CMake.txt
```diff
@@ -531,12 +531,18 @@ These five steps are described in detail below.
             set (ZLIB_USE_LOCALCONTENT ON CACHE BOOL "Use local file for ZLIB FetchContent" FORCE)
```
tbeu, line 531: Why is zlib not updated to v1.3.1 on the fly?

byrnHDF: We want to use the latest release after testing.

tbeu: OK. Why was libaec updated in this PR? Seems unrelated to zlib-ng.

byrnHDF: Because I have a bad habit of adding something I think is related.

derobins (approved):

PR #4489, opened 2024-05-14
Disable EOF checks for SWMR readers in more cases.

qkoziol:
Fixes a race condition where the reader opens the file and sets its EOF from the file's size (from the stat() call in the driver open callback).  Then, before the reader can read the file's superblock, a SWMR writer races in, extends the file, and closes the file, writing an updated superblock with the 'writer' and 'SWMR writer' flags in the superblock off (appropriately).  Then the reader proceeds to read the superblock, and flags the EOF as wrong.  Taking out the check for the 'writer' and 'SWMR writer' flags will cause SWMR readers to avoid flagging the file as incorrect.

### DIFF
--- a/src/H5Fsuper.c
+++ b/src/H5Fsuper.c
@@ -583,13 +583,10 @@ H5F__super_read(H5F_t *f, H5P_genplist_t *fa_plist, bool initial_read)
 
     if (H5F_INTENT(f) & H5F_ACC_SWMR_READ) {
         /*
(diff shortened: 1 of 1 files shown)


qkoziol:
This was exposed by the code changes to error handling in https://github.com/HDFGroup/hdf5/pull/4475

lrknox (approved):

byrnHDF (approved):

PR #4491, opened 2024-05-15
Improve spec. reading superblock into cache (a little) by using v2 size

qkoziol:
Instead of reading the absolute minimal possible, use the likely value of a v2+ superblock w/8-byte addresses & lengths.


qkoziol:
For files with a v2 or v3 superblock, this reduces the read ops on the superblock from 2 to 1.

byrnHDF (approved):

lrknox (approved):

PR #4494, opened 2024-05-15
Release asset tarballs with no version filenames

brtnfld:
This should allow other CI and documentation to point to 

https://github.com/HDFGroup/hdf5/releases/latest/download/hdf5.tar.gz

to get the latest source release and to avoid complicated "one-liners" when trying to find the name of the versioned tarballs.


brtnfld:
It does not create assets when I release from my fork, so I have not checked if it works. Do you have any ideas on how I can get that to work?

byrnHDF:
Do we want this everytime daily-build runs or only under release-build?

brtnfld:
Just for a release build.

.github/workflows/tarball.yml
```diff
@@ -142,6 +142,8 @@ jobs:
         run: |
           mv hdf5-${{ steps.version.outputs.SOURCE_TAG }}.tar.gz ${{ inputs.use_tag }}.tar.gz
           mv hdf5-${{ steps.version.outputs.SOURCE_TAG }}.zip ${{ inputs.use_tag }}.zip
+          cp ${{ inputs.use_tag }}.tar.gz hdf5.tar.gz
```
byrnHDF, line 151: These should be moved to the two new steps below (line 166/169)

.github/workflows/tarball.yml
```diff
@@ -164,6 +166,20 @@ jobs:
               path: ${{ steps.set-file-base.outputs.FILE_BASE }}.zip
               if-no-files-found: error # 'warn' or 'ignore' are also available, defaults to `warn`
 
+      - name: Save tgz-tarball_nover
```
byrnHDF: These two new steps should be conditioned with:
        if: ${{ (inputs.use_environ == 'release') }}

.github/workflows/release-files.yml
```diff
@@ -153,6 +153,8 @@ jobs:
           sha256sum ${{ steps.get-file-base.outputs.FILE_BASE }}.doxygen.zip > ${{ steps.get-file-base.outputs.FILE_BASE }}.sha256sums.txt
           sha256sum ${{ steps.get-file-base.outputs.FILE_BASE }}.tar.gz >> ${{ steps.get-file-base.outputs.FILE_BASE }}.sha256sums.txt
           sha256sum ${{ steps.get-file-base.outputs.FILE_BASE }}.zip >> ${{ steps.get-file-base.outputs.FILE_BASE }}.sha256sums.txt
+          sha256sum hdf5.tar.gz >> hdf5.sha256sums.txt
```
byrnHDF: The files need a step to download the artifacts
then a separate step to add the sha.
All conditioned on:
        if: ${{ (inputs.use_environ == 'release') }}

.github/workflows/release-files.yml
```diff
@@ -190,6 +192,8 @@ jobs:
               ${{ steps.get-file-base.outputs.FILE_BASE }}.doxygen.zip
               ${{ steps.get-file-base.outputs.FILE_BASE }}.tar.gz
               ${{ steps.get-file-base.outputs.FILE_BASE }}.zip
+              hdf5.tar.gz
```
byrnHDF: These should not be here - not release environ

byrnHDF (changes requested):
Requires adding conditional steps.

.github/workflows/release-files.yml
```diff
@@ -148,6 +148,19 @@ jobs:
               name: abi-reports
               path: ${{ github.workspace }}
 
+#     - name: Get published non-versioned source
+#        uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+#        if: ${{ (inputs.use_environ == 'release') }}
+#        with:
+#              name: cp-to-non-versioned
```
byrnHDF: The artifact names here must match the artifact names uploaded.  Then the files will be available for the sha256sum.

.github/workflows/tarball.yml
byrnHDF: With the next step (line 149) these can be removed from this step (line 139)

.github/workflows/tarball.yml
```diff
@@ -164,6 +174,22 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        if: ${{ (inputs.use_environ == 'release') }}
+        with:
+              name: tgz-tarball_nover
```
byrnHDF: This is the artifact name to be used in the release-files download steps.

byrnHDF (approved):

PR #4500, opened 2024-05-17
Allow HDF5_LIB_INFIX to work with DLL

byrnHDF:
Allow library names to be customized without affecting CMake target names - in particular the hdf5*_EXPORTS defines used by windows for import/export created by target names.


derobins (changes requested):
The naming scheme here would change the names from the previous behavior.

Also, the description field of any non-trivial PR should be filled in so people can more easily interpret the changes.

lrknox:
> The naming scheme here would change the names from the previous behavior.
> 
> Also, the description field of any non-trivial PR should be filled in so people can more easily interpret the changes.

These have been addressed

lrknox (approved):

derobins (approved):

derobins (approved):

PR #4453, opened 2024-05-01
Refactor error handling code to eliminate internal ID calls

qkoziol:
All calls to the H5I routines are now made in API routines (sometimes in FUNC_ENTER/LEAVE_* macros), except for some calls to H5E_clear_stack() within the library, but I'm planning to remove those over time.

Also, made all the library internal error messages into static const variables, instead of malloc'ing them, which means that they can just be referenced and not copied.

Several new and updated auto-generated header files were necessary to enable this.

PR #4503, opened 2024-05-20
Sync develop changes May 6 - May 20 to hdf5_1_14

lrknox:


byrnHDF (approved):

CMakePresets.json
byrnHDF, line 1: The plugin tgzname needs changed from master to 1_14. (line 42)

byrnHDF: Also then the plugins repo needs the PR merged

CMakeFilters.cmake
```diff
@@ -71,20 +71,20 @@ if (HDF5_ENABLE_Z_LIB_SUPPORT)
   if (NOT H5_ZLIB_HEADER)
     if (NOT ZLIB_USE_EXTERNAL)
       find_package (ZLIB NAMES ${ZLIB_PACKAGE_NAME}${HDF_PACKAGE_EXT} COMPONENTS static shared)
-      if (NOT ZLIB_FOUND)
+      if (NOT H5_ZLIB_FOUND)
         find_package (ZLIB) # Legacy find
       endif ()
-      if (ZLIB_FOUND)
+      if (H5_ZLIB_FOUND)
```
jhendersonHDF, line 77: Looks like we'll either need to merge the zlib/szip fixes immediately after this PR or add them here.

src/H5public.h
```diff
@@ -88,6 +88,10 @@
  * For pre-releases like \c snap0. Empty string for official releases.
  */
 #define H5_VERS_SUBRELEASE "1"
+/**
+ * Short version string
+ */
+#define H5_VERS_STR "1.15.0"
```
jhendersonHDF: I think `h5vers` will maintain this in the future, but I'm guessing for this PR that this should be set to 1.14.5 or similar?

lrknox:
```suggestion
#define H5_VERS_STR "1.14.5-1"
```

lrknox: I updated the version in line94 to 1.14.5-1.  I'm not sure h5vers will maintain it as it is, because H5VERS_STR in line 94 is new, apparently duplicating line 98 but without the "HDF5 library version:" text.  I think it's unlikely that h5vers was updated to maintain H5_VERS_STR in either develop or hdf5_1_14.  It should be done, though.

jhendersonHDF: h5vers was updated in https://github.com/HDFGroup/hdf5/pull/4453 where H5_VERS_STR was introduced.

lrknox: ok, that's good.  #4453 is in the sync, so it's adding H5_VERS_STR, but h5vers didn't run yet in the hdf5_1_14 branch to update the version.  It should be fine.

PR #4495, opened 2024-05-16
add support for aarch64 _Float16 16-bit floating point type

wswsmao:
_Float16 was supported in commit 330b80a266.

However there is build error report in aarch64, like this: ../../../src/H5private.h:605:23: error: '_Float16' does not name a type; did you mean '_Float64'?
  605 | __extension__ typedef _Float16 H5__Float16;
      |                       ^~~~~~~~
      |                       _Float64
make[2]: *** [Makefile:972: dsets.o] Error 1
make[1]: *** [Makefile:910: all-recursive] Error 1
make: *** [Makefile:739: all-recursive] Error 1

In aarch64, __fp16 type is half-precision (16-bit) floating point

PR #4507, opened 2024-05-21
Add h5test_wrapper.h to isolate gcc and g++ to compile _Float16 files on aarch64

wswsmao:
_Float16 was supported in commit 330b80a266.

However there is build error report in aarch64, like this: ../../../src/H5private.h:605:23: error: '_Float16' does not name a type; did you mean '_Float64'?
  605 | __extension__ typedef _Float16 H5__Float16;
      |                       ^~~~~~~~
      |                       _Float64
make[2]: *** [Makefile:972: dsets.o] Error 1
make[1]: *** [Makefile:910: all-recursive] Error 1
make: *** [Makefile:739: all-recursive] Error 1

This commit adds a wrapper to isolate this difference

### DIFF
--- a/c++/src/CMakeLists.txt
+++ b/c++/src/CMakeLists.txt
@@ -75,6 +75,7 @@ set (CPP_HDRS
     ${HDF5_CPP_SRC_SOURCE_DIR}/H5PropList.h
     ${HDF5_CPP_SRC_SOURCE_DIR}/H5StrType.h
     ${HDF5_CPP_SRC_SOURCE_DIR}/H5VarLenType.h
+    ${HDF5_CPP_SRC_SOURCE_DIR}/H5Wrapper.h
 )
 
 if (BUILD_STATIC_LIBS)
--- a/c++/src/H5Wrapper.h
+++ b/c++/src/H5Wrapper.h
@@ -1,0 +1,21 @@
+// C++ informative line for the emacs editor: -*- C++ -*-
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * Copyright by The HDF Group.                                               *
+ * All rights reserved.                                                      *
+ *                                                                           *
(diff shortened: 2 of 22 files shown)


wswsmao:
fix #4495

jhendersonHDF:
Hi @wswsmao, we aren't currently planning on supporting the `__fp16` type at this time and when/if we do, we'd want to add configure time checks to make sure the type is available and useful before compiling it in to the library. `__fp16` is not the same type as `_Float16`, so using a `#define` or typedef in this way is a bit misleading. For the time being, a proper fix for aarch64 systems would be refactoring the C++ tests to not depend on private headers from the C library or at least reducing what's included so that bringing along H5private.h isn't necessary.

wswsmao:
> Hi @wswsmao, we aren't currently planning on supporting the `__fp16` type at this time and when/if we do, we'd want to add configure time checks to make sure the type is available and useful before compiling it in to the library. `__fp16` is not the same type as `_Float16`, so using a `#define` or typedef in this way is a bit misleading. For the time being, a proper fix for aarch64 systems would be refactoring the C++ tests to not depend on private headers from the C library or at least reducing what's included so that bringing along H5private.h isn't necessary.

OK，I get it. I am looking forward to see my idea wound be helpful.

bmribler:
> Hi @wswsmao, we aren't currently planning on supporting the `__fp16` type at this time and when/if we do, we'd want to add configure time checks to make sure the type is available and useful before compiling it in to the library. `__fp16` is not the same type as `_Float16`, so using a `#define` or typedef in this way is a bit misleading. For the time being, a proper fix for aarch64 systems would be refactoring the C++ tests to not depend on private headers from the C library or at least reducing what's included so that bringing along H5private.h isn't necessary.

FYI, we're working on eliminating H5private.h from the C++ tests.

jhendersonHDF:
Tracking this in #4517

PR #4512, opened 2024-05-21
Set H5 specific vars immediately if legacy find

byrnHDF:


CMakeFilters.cmake
```diff
@@ -75,6 +75,8 @@ if (HDF5_ENABLE_Z_LIB_SUPPORT)
       find_package (ZLIB NAMES ${ZLIB_PACKAGE_NAME}${HDF_PACKAGE_EXT} COMPONENTS static shared)
       if (NOT ZLIB_FOUND)
         find_package (ZLIB) # Legacy find
+        set (H5_ZLIB_INCLUDE_DIR ${ZLIB_INCLUDE_DIRS})
```
jhendersonHDF: Don't these also need to be set if the first `find_package` call succeeds?

byrnHDF: Maybe.

byrnHDF: I am confusing two different processes! No wonder it doesn't work

jhendersonHDF: Based on this refactoring, it even seems like the `H5_ZLIB_FOUND` and `H5_SZIP_FOUND` aren't really needed at a first glance, though there could be other places they're used that I'm not seeing in the limited diff context.

byrnHDF: Maybe - but let's make sure everything works first

CMakeFilters.cmake
```diff
@@ -10,8 +10,8 @@
 # help@hdfgroup.org.
 #
 option (USE_LIBAEC_STATIC "Use static AEC library" OFF)
-option (ZLIB_USE_EXTERNAL "Use External Library Building for ZLIB" OFF)
-option (SZIP_USE_EXTERNAL "Use External Library Building for SZIP" OFF)
+option (ZLIB_USE_EXTERNAL "Use External Library Building for ZLIB else search" OFF)
```
jhendersonHDF, line 13: I feel like "else search" doesn't give much information and would just make this even more confusing for users, but I don't have any particular evidence to back that up.

byrnHDF: User sstill complained it was unclear and that is what happens search using find_package

jhendersonHDF: Right, I'm saying maybe that needs to be even more explicit, such as "else search system with find_package", even though I still think that isn't necessarily helpful for most people that just want to build their HDF5 with zlib support and don't know what find_package is.

jhendersonHDF (approved):

CMakeFilters.cmake
lrknox: I'm confused -  is that evidence?  If it's OFF, what does that mean about "else search"?  If it's ON, does it mean "Build and use designated External Library or use find_package to search system"?

lrknox (approved):

PR #4519, opened 2024-05-23
Fix function name in USAGE for H5Pencode2()

hyoklee:
USAGE has `H5Pencode()`.

### DIFF
--- a/src/H5P.c
+++ b/src/H5P.c
@@ -788,7 +788,7 @@ done:
     The encoding of property values will be done according to the file format
     setting in fapl_id.
  USAGE
-    herr_t H5Pencode(plist_id, buf, nalloc, fapl_id)
+    herr_t H5Pencode2(plist_id, buf, nalloc, fapl_id)
         hid_t plist_id;         IN: Identifier to property list to encode
         void *buf:              OUT: buffer to gold the encoded plist
         size_t *nalloc;         IN/OUT: size of buffer needed to encode plist


src/H5P.c
```diff
@@ -788,7 +788,7 @@ H5Pget_size(hid_t id, const char *name, size_t *size /*out*/)
     The encoding of property values will be done according to the file format
     setting in fapl_id.
  USAGE
-    herr_t H5Pencode(plist_id, buf, nalloc, fapl_id)
+    herr_t H5Pencode2(plist_id, buf, nalloc, fapl_id)
```
byrnHDF, line 791: Actually the original is not functionally incorrect - the version macros will choose the correct procedure based on the arguments. Adding the "2" is necessary under certain conditions as we encounter from time to time in the tests.

derobins (approved):

src/H5P.c
lrknox: > Actually the original is not functionally incorrect - the version macros will choose the correct procedure based on the arguments. Adding the "2" is necessary under certain conditions as we encounter from time to time in the tests.

Yes, but in other "USAGE" comment blocks for specific functions the '2' is generally included.  As it is "USAGE" for the function it is probably appropriate.

Looking at "USAGE" comments for functions in just a few source files I found 8 instances where the function names, parameters or descriptions don't match the actual function, most often due to copy and paste from another function.

lrknox (approved):

PR #4533, opened 2024-05-29
Fix a few function names in USAGE comments that don't match the actual

lrknox:
function names.
Sometimes parameters and parameter documentation are missing.
Often copy/paste operations with incomplete modification.

I think it is likely there are more, and I'll add a couple of specific questions in the code.  I came across these while reviewing another PR, and thought I would create this one to see if these are something we want to fix.

### DIFF
--- a/src/H5FO.c
+++ b/src/H5FO.c
@@ -136,7 +136,7 @@ H5FO_opened(const H5F_t *f, haddr_t addr)
  PURPOSE
     Insert a newly opened object/pointer pair into the opened object info set
  USAGE
-    herr_t H5FO_insert(f,addr,obj)
+    herr_t H5FO_insert(f,addr,obj,delete_flag)
         H5F_t *f;               IN/OUT: File's opened object info set
         haddr_t addr;           IN: Address of object to insert
         void *obj;              IN: Pointer to object to insert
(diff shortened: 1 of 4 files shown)


src/H5Sselect.c
```diff
@@ -413,7 +413,7 @@ H5S_get_select_npoints(const H5S_t *space)
     Check whether the selection fits within the extent, with the current
     offset defined.
  USAGE
-    htri_t H5Sselect_void(dsid)
+    htri_t H5Sselect_valid(dsid)
         hid_t dsid;             IN: Dataspace ID to query
```
lrknox, line 417: I assume dsid here and spaceid in line 430 are the same.  Would it be better if they match?

src/H5FO.c
```diff
@@ -239,9 +239,10 @@ H5FO_delete(H5F_t *f, haddr_t addr)
  PURPOSE
     Mark an object to be deleted when it is closed
  USAGE
-    herr_t H5FO_mark(f,addr)
+    herr_t H5FO_mark(f,addr,deleted)
         const H5F_t *f;         IN: File opened object is in
         haddr_t addr;           IN: Address of object to delete
+        bool deleted;           IN: true for to be deleted
```
lrknox, line 245: Should this be added, or is it too obvious?

byrnHDF (approved):

derobins (approved):

PR #4536, opened 2024-05-30
correct plugin URL

byrnHDF:
The URL for the plugin repo releases was pointing to the develop version instead of the 1.14 tag


lrknox:
codespell issues?

lrknox (approved):

PR #4543, opened 2024-06-03
Remove an unnecessary check for parallel and thread-safety from examples

jhendersonHDF:


HDF5Examples/CMakeLists.txt
```diff
@@ -98,11 +98,11 @@ endif ()
...
-if (HDF_ENABLE_THREADSAFE)
-  # check for unsupported options
-  if (HDF_ENABLE_PARALLEL)
-    message (FATAL " **** Parallel and Threadsafe options are mutually exclusive **** ")
-  endif ()
+# Note that HDF_ENABLE_THREADSAFE is the CMake option for determining
+# whether to enable thread-safety in the examples. HDF5_ENABLE_THREADSAFE
+# is the CMake option determining whether HDF5 was configured with
+# thread-safety enabled.
+if (HDF_ENABLE_THREADSAFE AND HDF5_ENABLE_THREADSAFE)
```
jhendersonHDF, line 105: Just assume here that checking for unsupported combinations was done at library configure time. Since the message command was using `FATAL` instead of `FATAL_ERROR`, it was just printing out the word "FATAL" rather than stopping configuration anyway.

brtnfld (approved):

lrknox (approved):

PR #4552, opened 2024-06-06
Move C++ and Fortran and examples to HDF5Examples folder

byrnHDF:
Move the examples folder into the HDF5Examples folder. 
Since the examples are referenced from the LeanBasics.dox file as a tutorial, we name the subdirs as TUTR.
Also rename the referenced Java files from H5J to TUTR.


derobins (approved):
This is fine, but I think the name HDF5Examples should be changed to 'examples' or similar. HDF5EXamples looks jarring.

lrknox (approved):

PR #4561, opened 2024-06-11
Add compression includes to tests needing zlib support

byrnHDF:


jhendersonHDF (approved):

mattjala (approved):

emmenlau:
I think the same change is also required for `zip_perf` in tools, is that possible?

Issue #4508, opened 2024-05-21
Add option for page buffer cache size to command-line tools


ajelenak:
**Is your feature request related to a problem? Please describe.**
Performance when accessing cloud optimized HDF5 files greatly benefits when using page buffer cache. Currently, CLI tools, like h5dump or h5ls, do not have an option to set page bugger cache size.

**Describe the solution you'd like**
Add an option for all CLI tools to set page buffer cache.

**Describe alternatives you've considered**
There are no alternatives I am aware of.

byrnHDF:
PR #4562 adds this option to tools

byrnHDF:
merged to develop

PR #4562, opened 2024-06-12
Add page buffer cache command line option to tools

byrnHDF:


tools/lib/h5diff.c
```diff
@@ -639,6 +639,20 @@ h5diff(const char *fname1, const char *fname2, const char *objname1, const char
         }
     }
 
+    if (opts->page_cache > 0) {
+        if ((fapl1_id = h5tools_get_fapl(fapl1_id, NULL, NULL)) < 0) {
```
jhendersonHDF: This should probably be assigned to a temporary variable first so that the old `fapl1_id` can be closed before reassigning `fapl1_id` to the temporary variable.

tools/lib/h5diff.c
```diff
@@ -662,6 +681,15 @@ h5diff(const char *fname1, const char *fname2, const char *objname1, const char
...
+    if (opts->page_cache > 0) {
+        if (H5Pset_page_buffer_size(fapl2_id, opts->page_cache, 0, 0) < 0) {
+            error_msg("unable to set page buffer cache size for file access\n");
+            h5tools_setstatus(EXIT_FAILURE);
+            ret_value = -1;
+            goto done;
+        }
+    }
+
     if ((file2_id = h5tools_fopen(fname2, H5F_ACC_RDONLY, fapl2_id, (fapl2_id != H5P_DEFAULT), NULL,
```
jhendersonHDF: Not sure how much this matters in practice, but since this call to `h5tools_fopen` checks for `fapl2_id != H5P_DEFAULT` and the changes here now always create a fapl for `fapl2_id`, this fopen call will always fail right away when it fails to open the file with the default VOL/VFD, where previously it would have tried others. Unclear if that's still a good idea at this point or not.

tools/src/h5dump/h5dump.c
```diff
@@ -1368,10 +1375,15 @@ main(int argc, char *argv[])
     /* Initialize indexing options */
     h5trav_set_index(sort_by, sort_order);
 
-    if (use_custom_vol_g || use_custom_vfd_g) {
-        if ((fapl_id = h5tools_get_fapl(H5P_DEFAULT, use_custom_vol_g ? &vol_info_g : NULL,
-                                        use_custom_vfd_g ? &vfd_info_g : NULL)) < 0) {
-            error_msg("unable to create FAPL for file access\n");
+    if ((fapl_id = h5tools_get_fapl(H5P_DEFAULT, use_custom_vol_g ? &vol_info_g : NULL,
```
jhendersonHDF: Same comment here about this now always creating a fapl, which affects the logic used for `h5tools_fopen`. With there being other options now that may need a fapl, here and some other places we can probably switch to using `use_custom_vol_g || use_custom_vfd_g` in the `h5tools_fopen` call rather than `fapl_id != H5P_DEFAULT`.

tools/src/h5ls/h5ls.c
```diff
@@ -2995,23 +3002,27 @@ main(int argc, char *argv[])
...
+    if (custom_vfd_fapl && (0 == strcmp(vfd_info.u.name, drivernames[HDFS_VFD_IDX]))) {
+        if (!vfd_info.info)
+            vfd_info.info = &hdfs_fa;
+    }
 #endif
 
-        if ((fapl_id = h5tools_get_fapl(H5P_DEFAULT, custom_vol_fapl ? &vol_info : NULL,
-                                        custom_vfd_fapl ? &vfd_info : NULL)) < 0) {
-            error_msg("failed to setup file access property list (fapl) for file\n");
+    if ((fapl_id = h5tools_get_fapl(H5P_DEFAULT, custom_vol_fapl ? &vol_info : NULL,
```
jhendersonHDF: Same comment here about this now always creating a fapl, which affects the logic used for `h5tools_fopen`. Can probably become `custom_vol_fapl || custom_vfd_fapl` rather than `(fapl_id != H5P_DEFAULT)`.

tools/src/h5repack/h5repack_main.c
```diff
@@ -924,6 +933,32 @@ parse_command_line(int argc, const char *const *argv, pack_opt_t *options)
...
+        /* Close old FAPL */
+        if (options->fin_fapl != H5P_DEFAULT)
+            if (H5Pclose(options->fin_fapl) < 0) {
+                error_msg("failed to close FAPL\n");
+                h5tools_setstatus(EXIT_FAILURE);
+                ret_value = -1;
+                goto done;
+            }
+
+        options->fin_fapl = tmp_fapl;
```
jhendersonHDF: The logic in h5repack_copy.c is also checking `options->fin_fapl` against H5P_DEFAULT since a VOL/VFD used to be the only option that would create a non-default one. That logic may need to be changed as well to something like `custom_in_vol || custom_in_vfd` from above, but these booleans would probably need to be added to the `options` struct.

tools/src/h5stat/h5stat.c
```diff
@@ -1639,6 +1647,19 @@ main(int argc, char *argv[])
             goto done;
         }
     }
+    else {
+        if ((fapl_id = h5tools_get_fapl(H5P_DEFAULT, NULL, NULL)) < 0) {
```
jhendersonHDF: Same comment here about this now always creating a fapl, which affects the logic used for `h5tools_fopen`. May need to set a boolean to signify that a custom driver was set so that can be used in the `h5tools_fopen` logic below rather than `(fapl_id != H5P_DEFAULT)`

tools/lib/h5diff.c
byrnHDF: Shouldn't that be done inside the h5tools_get_fapl function? Or at least add an option BOOL close_old_fapl.

byrnHDF: Every place I looked where h5tools_get_fapl() is being used the old is getting closed.

tools/src/h5ls/h5ls.c
byrnHDF: Yes, I found those concerning, too.

tools/src/h5stat/h5stat.c
byrnHDF: Agree that something needs changed.

tools/lib/h5diff.c
jhendersonHDF: This appears to be the one place in the PR where the old one isn't getting closed. An option to the function could possibly make sense, but usually having a function that needs a bunch of `false, true, false` is bad design.

byrnHDF: Is there a reason not to close the previous fapl as I didn't find any current instances that continue to use it.

byrnHDF: I suggest just closing it inside and the future can add an option if necessary.

tools/lib/h5diff.c
byrnHDF: So I need to create a var here for custom_vol/vfd?

tools/lib/h5diff.c
jhendersonHDF: I'd say the main reason is that it's probably surprising behavior to find that an ID you passed in to the function got closed from underneath you. It adds extra work too since you need to make an extra copy of your ID if you do plan to use it afterward.

tools/lib/h5tools.c
```diff
@@ -1025,8 +1034,11 @@ h5tools_fopen(const char *fname, unsigned flags, hid_t fapl_id, bool use_specifi
                 vfd_info.info   = NULL;
                 vfd_info.u.name = drivernames[drivernum];
 
+                /* create a copy of the original fapl */
+                if ((cpy_fapl_id = H5Pcopy(fapl_id)) < 0)
```
jhendersonHDF: This seems like a good example of where it becomes awkward to pass a fapl ID in to a function where it's going to close it and you don't necessarily want it to. It also makes it seem like there's an ID leak until you realize `h5tools_get_fapl` will close it.

byrnHDF: The other choice is to pass in a DoNotClose bool. This was the only place that really needed it because of the do-loop

jhendersonHDF: It seems the simplest just to have the calling code close the ID itself, while just considering `h5tools_get_fapl` as giving out a new ID that's completely separate from the old one. In some cases that looks messier when the old ID needs to be closed, but adding a "DoNotClose" also invites future copy-paste errors.

byrnHDF: It's messy either way.

tools/lib/h5diff.c
jhendersonHDF: Looks like you can use something like `opts->custom_vol[0] || opts->custom_vol[1] || opts->custom_vfd[0] || opts->custom_vfd[1]`. A bit ugly, so maybe a temp var just to get the line length down?

byrnHDF:
Closes #4508

tools/src/misc/h5mkgrp.c
```diff
@@ -252,18 +252,24 @@ parse_command_line(int argc, const char *const *argv, mkgrp_opt_t *options)
 
     /* Setup a custom fapl for file accesses */
     if (custom_vol || custom_vfd) {
-        if ((tmp_fapl_id = h5tools_get_fapl(options->fapl_id, custom_vol ? &vol_info : NULL,
```
jhendersonHDF, line 255: I can't comment on the `h5tools_fopen` call below since it's outside the diff context, but another case where `(params_g.fapl_id != H5P_DEFAULT) ` should probably be switched to `custom_vol || custom_vfd`

tools/src/h5ls/h5ls.c
```diff
@@ -2816,6 +2820,9 @@ main(int argc, char *argv[])
                 leave(EXIT_FAILURE);
             }
         }
+        else if (!strncmp(argv[argno], "--page-buffer-size=", (size_t)19)) {
+            page_cache = strtoul(argv[argno] + 8, &rest, 0);
```
jhendersonHDF: Need to set this to 19 as well

jhendersonHDF:
Just a couple things left, but looking good. After thinking on it for a bit, I think making set_fapl_vol and set_fapl_vfd separate from get_fapl is a better design anyway based on how the functions were being used in the tools.

jhendersonHDF (approved):

lrknox (approved):

PR #4568, opened 2024-06-13
Allow usage of page buffering for serial file access from parallel HDF5 builds

jhendersonHDF:
Allows the page buffering feature to be used for serial file access with parallel builds of HDF5, where the library would previously throw an error.

### DIFF
--- a/src/H5Fint.c
+++ b/src/H5Fint.c
@@ -2008,15 +2008,6 @@ H5F_open(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id)
     if (H5P_get(a_plist, H5F_ACS_PAGE_BUFFER_SIZE_NAME, &page_buf_size) < 0)
         HGOTO_ERROR(H5E_FILE, H5E_CANTGET, NULL, "can't get page buffer size");
     if (page_buf_size) {
-#ifdef H5_HAVE_PARALLEL
-        /* Collective metadata writes are not supported with page buffering */
-        if (file->shared->coll_md_write)
-            HGOTO_ERROR(H5E_FILE, H5E_CANTOPENFILE, NULL,
-                        "collective metadata writes are not supported with page buffering");
-
-        /* Temporary: fail file create when page buffering feature is enabled for parallel */
-        HGOTO_ERROR(H5E_FILE, H5E_CANTOPENFILE, NULL, "page buffering is disabled for parallel");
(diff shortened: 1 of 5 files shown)


jhendersonHDF:
These changes are needed for #4562.

src/H5Fint.c
```diff
@@ -2008,15 +2008,6 @@ H5F_open(const char *name, unsigned flags, hid_t fcpl_id, hid_t fapl_id)
     if (H5P_get(a_plist, H5F_ACS_PAGE_BUFFER_SIZE_NAME, &page_buf_size) < 0)
         HGOTO_ERROR(H5E_FILE, H5E_CANTGET, NULL, "can't get page buffer size");
     if (page_buf_size) {
-#ifdef H5_HAVE_PARALLEL
-        /* Collective metadata writes are not supported with page buffering */
-        if (file->shared->coll_md_write)
```
jhendersonHDF, line 2013: Moved these checks down to a single `H5_HAVE_PARALLEL` block below.

test/page_buffer.c
```diff
@@ -2043,121 +2034,6 @@ test_stats_collection(hid_t orig_fapl, const char *driver_name)
...
- *
- * Return:      0 if test is successful
- *              1 if test fails
- *
- *-------------------------------------------------------------------------
- */
-
-#ifdef H5_HAVE_PARALLEL
-static unsigned
-verify_page_buffering_disabled(hid_t orig_fapl, const char *driver_name)
```
jhendersonHDF, line 2065: Not needed in this serial test now.

testpar/testphdf5.c
```diff
@@ -402,9 +402,7 @@ main(int argc, char **argv)
     AddTest("split", test_split_comm_access, NULL, "dataset using split communicators", PARATESTFILE);
     AddTest("h5oflusherror", test_oflush, NULL, "H5Oflush failure", PARATESTFILE);
 
-#ifdef PB_OUT /* temporary: disable page buffering when parallel */
     AddTest("page_buffer", test_page_buffer_access, NULL, "page buffer usage in parallel", PARATESTFILE);
```
jhendersonHDF, line 405: Make sure that we're always testing page buffer access, even if that means that in parallel we're just making sure it's disabled.

byrnHDF (approved):

lrknox (approved):

PR #4570, opened 2024-06-14
Document H5Punregister modifying default properties

mattjala:
`H5Punregister()` says that existing property lists are unaffected. However, property lists get default property values from their parent class(es). If list has a property with a default value, then calling `H5Punregister()` to remove that property from the class will also remove it from the list. 

This is probably something we'll want to change eventually, but for now we can at least document the actual behavior.

### DIFF
--- a/src/H5P.c
+++ b/src/H5P.c
@@ -1397,8 +1397,10 @@ done:
     Returns non-negative on success, negative on failure.
  DESCRIPTION
         Removes a property from a property list class.  Future property lists
-    created of that class will not contain this property.  Existing property
-    lists containing this property are not affected.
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

qkoziol (approved):

qkoziol:
> `H5Punregister()` says that existing property lists are unaffected. However, property lists get default property values from their parent class(es). If list has a property with a default value, then calling `H5Punregister()` to remove that property from the class will also remove it from the list.
> 
> This is probably something we'll want to change eventually, but for now we can at least document the actual behavior.

Yes, this looks like a bug.  Existing property lists with the property should retain the value.

PR #4573, opened 2024-06-17
Turn on parallel CI tests in Autotools & CMake

derobins:
Does not turn parallel CI on in NVHPC, etc. - just the main yml files.


.github/workflows/main-auto-par.yml
```diff
@@ -75,3 +75,8 @@ jobs:
       - name: Autotools Build
         run: make -j3
         working-directory: ${{ runner.workspace }}/build
+
+      - name: Autotools Run Tests
+        run: make check -j2
```
jhendersonHDF: For these actions we don't want to run the tests in parallel since that may cause slowdowns when the actual parallel tests run. Really what we want to do is run the serial tests in parallel separately and then run the parallel tests serially and set `RUNPARALLEL="mpiexec -np 2"` for autotools and `MPIEXEC_MAX_NUMPROCS=X` (probably X=2) for CMake.

derobins: I've thought about that, but that's more complicated since the serial tests are in multiple directories

byrnHDF:
For CMake you can easily use the REGEX form to run serial then parallel - Larry does it for HPC

derobins:
> For CMake you can easily use the REGEX form to run serial then parallel - Larry does it for HPC

Maybe we can start by running parallel tests just in CMake

.github/workflows/main-auto-par.yml
jhendersonHDF: > I've thought about that, but that's more complicated since the serial tests are in multiple directories

I'd imagine you can just do `cd test; make check -j2; cd ../tools/test; make check -j2` to only run all the serial tests that are interesting. It's not future-proof, but we can always start with just CMake like mentioned.

derobins: I'll update this when I get in on Tuesday

hyoklee (approved):

byrnHDF (approved):

PR #4575, opened 2024-06-18
Rename INSTALL_Auto.txt to INSTALL_Autotools.txt

derobins:


derobins:
Weird that the link checker isn't smart enough to handle updated links in the PR

byrnHDF (approved):

brtnfld (commented):
I think we should transition to using markdown for text docs, as many people now read this information from GitHub.

derobins:
> I think we should transition to using markdown for text docs, as many people now read this information from GitHub.

Agree, but that's a task for a different PR

brtnfld:
Why is this PR necessary if a future change to the markdown file will require renaming again?

derobins:
> Why is this PR necessary if a future change to the markdown file will require renaming again?

The odds of getting INSTALL_Auto.md seem high and I wasn't thinking about converting this to markdown when I created the PR. The name was just less than optimal.

brtnfld (approved):

PR #4595, opened 2024-06-21
Bring subfiling VFD code closer to typical library code

qkoziol:
Remove API calls, use FUNC_ENTER/LEAVE macros, use the library's error macros, rename functions to have more standardized names, etc.


qkoziol:
This PR addresses the issues in the subfiling code from the "pause_errors" PR, https://github.com/HDFGroup/hdf5/pull/4475

hyoklee (approved):

src/H5FDsubfiling/H5FDioc_threads.c
```diff
@@ -362,50 +334,44 @@ ioc_main(ioc_data_t *ioc_data)
...
-            int    tag    = status.MPI_TAG;
+            int count;
+            int source = status.MPI_SOURCE;
+            int tag    = status.MPI_TAG;
 
-            if ((tag != READ_INDEP) && (tag != WRITE_INDEP) && (tag != TRUNC_OP) && (tag != GET_EOF_OP))
-                H5_SUBFILING_GOTO_ERROR(H5E_ARGS, H5E_BADVALUE, -1, "invalid work request operation (%d)",
-                                        tag);
+            if (tag != READ_INDEP && tag != WRITE_INDEP && tag != TRUNC_OP && tag != GET_EOF_OP)
+                HGOTO_DONE((hg_thread_ret_t)FAIL);
```
jhendersonHDF, line 345: As a bit of irony, this seems like a very good example where separate error stacks would be nice so that when an error occurs in this thread, it can return an ID for its error stack and the process can merge that stack with its own with `H5Eappend_stack` so we keep errors like this rather than getting an unhelpful FAIL.

src/H5FDsubfiling/H5FDioc_threads.c
```diff
@@ -1098,11 +1040,11 @@ ioc_file_read_data(int fd, int64_t file_offset, void *data_buffer, int64_t data_
         else {
             if (retries == 0) {
 #ifdef H5FD_IOC_DEBUG
-                printf("[ioc(%d) %s]: TIMEOUT: file_offset=%" PRId64 ", data_size=%ld\n", ioc_idx, __func__,
+                printf("[ioc(%d) %s]: TIMEOUT: file_offset=%" PRId64 ", data_size=%lld\n", ioc_idx, __func__,
```
jhendersonHDF: Nitpick: use `%" PRId64 "` for `data_size` rather than `%lld`

src/H5FDsubfiling/H5FDioc_threads.c
qkoziol: I was thinking about that too while I was working on these background thread errors.  :-)

I'll add an item in my TODO list to work on how to make that happen.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -618,45 +599,59 @@ H5FD__subfiling_get_default_config(hid_t fapl_id, H5FD_subfiling_config_t *confi
...
+        H5FD_ioc_config_t ioc_config;
 
-        if (H5Pset_fapl_ioc(config_out->ioc_fapl_id, NULL) < 0)
-            H5_SUBFILING_GOTO_ERROR(H5E_PLIST, H5E_CANTSET, FAIL, "can't set IOC VFD on IOC FAPL");
+        if (H5P_set(plist, H5F_ACS_MPI_PARAMS_COMM_NAME, &comm) < 0)
+            HGOTO_ERROR(H5E_VFL, H5E_CANTSET, FAIL, "can't set MPI communicator");
+        if (H5P_set(plist, H5F_ACS_MPI_PARAMS_INFO_NAME, &info) < 0)
+            HGOTO_ERROR(H5E_VFL, H5E_CANTSET, FAIL, "can't set MPI info object");
+
+        if (H5FD__subfiling_get_default_ioc_config(&ioc_config) < 0)
```
jhendersonHDF, line 632: I think this highlights an architectural issue in that Subfiling needs to be able to set the underlying I/O concentrator VFD using a default config without being able to make a public API call here. Subfiling was meant to be a modular architecture, with the IOC VFD being a reference implementation of an I/O concentrator, but this tends to tie it right to the reference implementation. Fine for now, but I think a better solution is needed in general for the case of VFDs stacked on top of other VFDs.

src/H5FDsubfiling/H5FDioc.c
```diff
@@ -807,158 +749,131 @@ H5FD__ioc_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxaddr)
...
-    file_ptr->file_dir = NULL;
+    H5MM_free(file->file_path);
+    H5MM_free(file->file_dir);
 
     /* Release the file info */
-    file_ptr = H5FL_FREE(H5FD_ioc_t, file_ptr);
+    H5FL_FREE(H5FD_ioc_t, file);
 
-    H5_SUBFILING_FUNC_LEAVE;
+done:
```
jhendersonHDF: Why switch to skipping past the freeing of memory on error here? Just seems to leak memory for no reason. I'm also not a big fan of removing the

```
file_ptr->context_id = -1;
file_ptr->file_path = NULL;
file_ptr->file_dir = NULL;
```

statements since I like to be sure I leave memory in a sane state when free lists are involved, but it's not critical.

src/H5FDsubfiling/H5FDsubfiling.c
qkoziol: Agree - we should update the stackable VFDs RFC at some point, with the learnings from the VOL connectors.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -842,21 +837,12 @@ H5FD__subfiling_sb_encode(H5FD_t *_file, char *name, unsigned char *buf)
...
-        char ioc_name[9];
-
-        memset(ioc_name, 0, sizeof(ioc_name));
-
-        if (H5FD_sb_encode(file->sf_file, ioc_name, p + 9) < 0)
-            H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTENCODE, FAIL,
-                                    "unable to encode IOC VFD's superblock information");
-
-        /* Copy the IOC VFD's name into our buffer */
-        H5MM_memcpy(p, ioc_name, 9);
```
jhendersonHDF, line 855: Any reason why this was removed? The name of the underlying VFD is part of the Subfiling VFD's encoded information, mostly for future verification.

EDIT: Disregard, I see now that it's an out parameter from `H5FD_sb_encode`. The odd ordering of parameters makes the line read a little weird and misleading.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -1121,248 +1092,208 @@ H5FD__subfiling_fapl_free(void *_fa)
...
+        if (H5_mpi_comm_free(&file->comm) < 0)
+            HGOTO_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "unable to free MPI Communicator");
+        if (H5_mpi_info_free(&file->info) < 0)
+            HGOTO_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "unable to free MPI Info object");
+        if (H5_mpi_comm_free(&file->ext_comm) < 0)
+            HGOTO_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "can't free MPI communicator");
     }
 
-    file_ptr->fail_to_encode = false;
+    H5MM_free(file->file_path);
```
jhendersonHDF, line 1286: Another case where we should free memory even on errors.

src/H5FDsubfiling/H5FDsubfiling.c
```diff
@@ -1657,71 +1596,44 @@ H5FD__subfiling_write(H5FD_t *_file, H5FD_mem_t type, hid_t H5_ATTR_UNUSED dxpl_
...
     /* Set DXPL for operation */
     H5CX_set_dxpl(dxpl_id);
 
-    if (file_ptr->fa.require_ioc) {
-        if (H5FD__subfiling_io_helper(file_ptr, (size_t)count, types, addrs, sizes,
-                                      (H5_flexible_const_ptr_t *)bufs, IO_TYPE_READ) < 0)
-            H5_SUBFILING_GOTO_ERROR(H5E_IO, H5E_READERROR, FAIL, "couldn't read data");
-    }
-    else {
-        if (H5FD_read_vector(_file, count, types, addrs, sizes, bufs) < 0)
```
jhendersonHDF, line 1719: This called was removed here, but left for `H5FD__subfiling_read_vector`. While I'm not really in favor of removing things functionality-wise unless there is a demonstrated need for it, the two should be consistent either way.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -801,44 +730,39 @@ init_subfiling(const char *base_filename, uint64_t file_id, H5FD_subfiling_param
...
     /* Create a new subfiling context object with the created context ID */
-    if (NULL == (new_context = H5_get_subfiling_object(context_id)))
-        H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL, "couldn't create new subfiling object");
-    new_context->sf_context_id = -1;
-    new_context->topology      = NULL;
-    new_context->sf_msg_comm   = MPI_COMM_NULL;
-    new_context->sf_data_comm  = MPI_COMM_NULL;
-    new_context->sf_eof_comm   = MPI_COMM_NULL;
-    new_context->sf_node_comm  = MPI_COMM_NULL;
-    new_context->sf_group_comm = MPI_COMM_NULL;
```
jhendersonHDF, line 829: I'd generally prefer all the initialization and de-initialization code removed in this PR to stay intact. While removing these shouldn't generally pose a problem at the moment due to the current ordering of functions that are called, I can't really predict the future and this could easily turn into an `MPI_Comm_free` on an uninitialized communicator.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -940,46 +856,140 @@ init_subfiling(const char *base_filename, uint64_t file_id, H5FD_subfiling_param
      * Setup the application topology information, including the computed
      * number and distribution map of the set of I/O concentrators
      */
-    if (init_app_topology(context_id, subfiling_config, comm, node_comm, &app_topology) < 0)
-        H5_SUBFILING_GOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL, "couldn't initialize application topology");
+    if (H5FD__subfiling_init_app_topology(context_id, subfiling_config, comm, node_comm, &app_topology) < 0)
+        HGOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL, "couldn't initialize application topology");
+
+    /* Set non-zero fields */
```
jhendersonHDF: This function was called `init_subfiling` because it initializes both the context and topology objects. Moving the code out of the `init_subfiling_context` function to here just makes this harder to reason about IMO.

src/H5FDsubfiling/H5subfiling_common.c
qkoziol: These initializations are still intact, they were duplicated by later code (around lines 860-870 now), so I removed this copy.

src/H5FDsubfiling/H5subfiling_common.c
qkoziol: OK, the comment in the header seemed to imply that init_subfiling was mainly about initializing the context, so I brought up the context initialization into it.   I'll rename this routine to something more general (H5FD__subfiling_setup, possibly) and move all the context initialization back into one routine.

src/H5FDsubfiling/H5subfiling_common.c
jhendersonHDF: The main issue I was trying to get to is that if these fields aren't initialized right after the context object is allocated, there are several errors that could occur in between and the logic could hit the cleanup code with a partially-initialized context object.

qkoziol: Agree - now that I understand the intent of this routine better, I'll factor the context initialization back into a subroutine again.  That can bring the initialization back to following the allocation, while avoiding the duplication.  

Probably have it done later tonight.

qkoziol: Done

qkoziol:
@jhendersonHDF @derobins - Could you please review?  I'd like to get this merged, so I can return to debugging the performance regression in the subfiling code.

src/H5FDsubfiling/H5subfiling_common.c
```diff
@@ -901,85 +815,84 @@ init_subfiling(const char *base_filename, uint64_t file_id, H5FD_subfiling_param
...
-                H5_SUBFILING_DONE_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "couldn't free subfiling topology");
-        }
+        if (app_topology)
+            if (H5FD__subfiling_free_topology(app_topology) < 0)
+                HDONE_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "couldn't free subfiling topology");
 
-        if (H5_mpi_comm_free(&node_comm) < 0)
-            H5_SUBFILING_DONE_ERROR(H5E_VFL, H5E_CANTFREE, FAIL, "couldn't free MPI communicator");
+        if (node_comm != MPI_COMM_NULL)
+            if (H5_mpi_comm_free(&node_comm) < 0)
```
jhendersonHDF: Not a big deal at all, but `H5_mpi_comm_free` already checks for `MPI_COMM_NULL` and `MPI_COMM_WORLD`, which is why the code previously just called the function for simplicity.

jhendersonHDF (approved):
Generally looks good now, though the significant amount of cleanup and other mostly-unrelated changes definitely made this a decent bit harder to review

src/H5FDsubfiling/H5subfiling_common.c
qkoziol: Ah, missed that, thanks.  Fixed now.

derobins (approved):

PR #4604, opened 2024-06-25
Feature/awesome

gheber:
Use [Doxygen Awesome ](https://jothepro.github.io/doxygen-awesome-css/) for a more modern look.


byrnHDF:
Doxygen CI error: Correcting memberror: Style sheet '/home/runner/work/hdf5/hdf5/doxygen/doxygen-awesome-css/doxygen-awesome.css' specified by HTML_EXTRA_STYLESHEET does not exist!

byrnHDF (approved):

gheber:
`doxygen-awesome.css` comes from the Doxygen Awesome Git submodule specified in `.gitmodules`. Doesn't that get automatically cloned? Maybe there is a GH action missing that takes care of submodules?

gheber:
I'm reluctant to include the Doxygen Awesome code in the repo. The GH actions for cloning, building releases, etc., should be modified.

byrnHDF:
> I'm reluctant to include the Doxygen Awesome code in the repo. The GH actions for cloning, building releases, etc., should be modified.

Why? We could use DA (Doxygen Awesome) like how we do other external code - use a compressed file and extract it. This would also allow us to decide what versions we are compatible with and how to obtain the files.

lkurz:
> > I'm reluctant to include the Doxygen Awesome code in the repo. The GH actions for cloning, building releases, etc., should be modified.
> 
> Why? We could use DA like how we do other external code - use a compressed file and extract it. This would also allow us to decide what versions we are compatible with and how to obtain the files.

What is "DA" of which you speak?

submodules is the suggested method by Doxygen Awesome

and yes "there's an action for that" 

submodules can be pinned to a tag - much easier to change that in a config and workflows than to manually maintain local zips of external sources of deps

submodules are not without there difficulties - I remember OpenMPI went to submodules awhile back and it caused some confusion with cloning and builds.  
[GitSubmodules · open-mpi/ompi Wiki](https://github.com/open-mpi/ompi/wiki/GitSubmodules)

jhendersonHDF:
> `doxygen-awesome.css` comes from the Doxygen Awesome Git submodule specified in `.gitmodules`. Doesn't that get automatically cloned? Maybe there is a GH action missing that takes care of submodules?

Submodules in .gitmodules don't generally get automatically cloned. The file just tracks the information for the cloning process.

derobins:
Submodules can be difficult to wrap your head around. We used them briefly in VFD SWMR (I don't remember why), but they caused so many problems that we went back to not using them. In this case, we probably wouldn't have to modify the DA code, just occasionally update the submodule, so it should be less troublesome than with VFD SWMR. People would just need to remember to add `--recurse-submodules` or whatnot when they clone.

I'm not opposed and many other projects use them, but we'll have to educate staff and update the docs.

byrnHDF:
How many files are we talking about here, a half-dozen? Is it worth the extra headaches for user-support, release checks, development. Just copy them into a folder!

ajelenak:
I agree with @byrnHDF. What's the benefit of having this submodule?

lrknox (approved):

PR #4610, opened 2024-06-27
Intended URL link changes

byrnHDF:


mattjala (changes requested):
The new links in CITATION, README, and RELEASE are currently 404s - this shouldn't go in until those go to real pages.

PR #4611, opened 2024-06-27
Document VOL object wrapping context

mattjala:
The VOL object wrapping context is documented very sparsely at the moment. This PR adds an overview of what it's for and when it's used, and details what the `wrap_cls` callbacks should do in the VOL connector author guide.

### DIFF
--- a/doxygen/dox/VOLConnGuide.dox
+++ b/doxygen/dox/VOLConnGuide.dox
@@ -55,8 +55,7 @@ and, although it is nearing its final form, may change further before the stable
 for 2022.
 
 \subsection subsecVOLRelated VOL-Related HDF5 Header Files
-Use of the VOL, including topics such as registration and loading VOL plugins, is described in the \ref
-H5VL_UG.
+Use of the VOL, including topics such as registration and loading VOL plugins, is described in the \ref H5VL_UG.
 
 Public header Files you will need to be familiar with include:
 <table>
@@ -69,8 +68,7 @@ Public header Files you will need to be familiar with include:
 <tr>
 <td>H5VLconnector.h
 </td>
-<td>Main header for connector authors. Contains definitions for the main VOL struct and callbacks, enum
-values, etc.
+<td>Main header for connector authors. Contains definitions for the main VOL struct and callbacks, enum values, etc.
 </td>
 </tr>
 <tr>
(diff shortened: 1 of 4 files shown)


byrnHDF (approved):

doxygen/dox/VOLConnGuide.dox
```diff
@@ -696,14 +706,17 @@ Retrieves an underlying object.
...
+  obj  (IN): The object to be unwrapped.
 \endcode
 </td>
 </tr>
 </table>
 
 \subsubsection subsubsecVOLRefWrapctx wrap: get_wrap_ctx
 Get a VOL connector's object wrapping context.
+
+The context should be returned in dynamically allocated memory under <em>*wrap_ctx</em>.
```
jhendersonHDF, line 718: @qkoziol When trying to help with documenting the VOL wrap class, I couldn't quite figure out what the wrap context object was necessary for. Do you happen to remember why a passthrough VOL needs to return a wrapping context object for the library to keep around?

mattjala: The wrapping context object is saved on the API Context, which is where `H5VL__wrap_obj` retrieves it from to provide it to the wrap object callback

qkoziol: Wrapping objects is necessary for the library to return objects from user callbacks and other scenarios where the library has an object from a terminal VOL connector and needs to eventually pass back an ID to an application.  The library has to wrap the terminal object using pass-throughs 'wrap' callbacks, so that the resulting ID refers to an object that can later be unwrapped by the connector stack and reach the terminal connector again.

qkoziol (approved):

doxygen/dox/VOLConnGuide.dox
jhendersonHDF: Right, I was asking more about the wrapping context object in particular and why a connector needs to allocate one and pass it back to the library.

qkoziol: The wrapping context is what the library gives back to a VOL connector to say "wrap this object for me".  Without the wrapping context, the pass-through connector wouldn't know which connector stack it was supposed to be wrapping an object for.

derobins (approved):

PR #4619, opened 2024-07-01
Add H5F_LIBVER_V116 to internal library version structures

jhendersonHDF:
Also add H5F_LIBVER_V116 flag to Fortran wrappers


src/H5Aint.c
```diff
@@ -95,6 +95,7 @@ const unsigned H5O_attr_ver_bounds[] = {
     H5O_ATTR_VERSION_3,     /* H5F_LIBVER_V110 */
     H5O_ATTR_VERSION_3,     /* H5F_LIBVER_V112 */
     H5O_ATTR_VERSION_3,     /* H5F_LIBVER_V114 */
+    H5O_ATTR_VERSION_3,     /* H5F_LIBVER_V116 */
```
jhendersonHDF, line 98: Since these structures are indexed according to the library bound version value, "latest" previously represented the value for the next major release, such as 1.16. However, this makes maintenance a bit more awkward than necessary since the specific library version entries then don't get added until the following major release is in development. It also makes it difficult to write code that checks specifically against just the latest format version since checking against latest is a moving target. These changes mean that the "latest" entry should generally never be hit unless there's a point where the latest libver bound doesn't map to the latest-numbered version, but the current indexing will still work out correctly since "latest" currently always maps to the latest-numbered version.

byrnHDF (approved):

vchoi-hdfgroup (approved):

brtnfld (approved):
Fortran looks good

PR #4624, opened 2024-07-02
Remove some internal use of API calls and H5E_BEGIN_TRY/H5E_END_TRY

qkoziol:


src/H5VLcallback.c
```diff
@@ -3572,30 +3570,10 @@ H5VL__file_open_find_connector_cb(H5PL_type_t plugin_type, const void *plugin_in
     vol_cb_args.args.is_accessible.fapl_id    = fapl_id;
     vol_cb_args.args.is_accessible.accessible = &is_accessible;
 
-    /* Store current error stack size */
-    if ((num_errors = H5Eget_num(H5E_DEFAULT)) < 0)
-        HGOTO_ERROR(H5E_ERROR, H5E_CANTGET, H5_ITER_ERROR, "can't get current error stack size");
+    if (H5VL_file_specific(NULL, &vol_cb_args, H5P_DATASET_XFER_DEFAULT, H5_REQUEST_NULL) < 0)
```
jhendersonHDF, line 3573: This seems like another case where we need this to be a 'try'. The previous code was treating it like a 'try' by popping any errors that occurred from the error stack if the is_accessible call failed, but trying to keep the error messages (if any) around in the case where the file isn't accessible. Now if the is_accessible call fails, it just fails the iteration, which is not quite the behavior wanted here.

qkoziol: It is effectively a 'try' already - it's returning the flag in a parameter.  So, any failure means that the flag couldn't be determined, which is a real error.   I think that a file not being accessible isn't an error, it's just not accessible.

jhendersonHDF: The original intention (for better or worse) was to ignore errors that occur during the `H5VL_file_specific` call, treat them as the file not being accessible with the current VOL connector and then move on to the next one so that each available connector is tried. IIRC, the native VOL used to return `FAIL` when a file wasn't accessible, but I believe that has been fixed so it will return `false` now. I'm not sure whether or not the expected behavior of `H5VL_file_specific` in terms of `FAIL` vs `false` is really documented though.

qkoziol: I believe that it would be more consistent to have this routine act like the rest of the library's 'try' routines, where it is an actual error to not be able to answer the question "is this file accessible?".  That was the original intention of the 'is_accessible' callback.  As you said, the native VOL connector was already discriminating between actual errors and 'false', so it's behaving correctly.

jhendersonHDF: I'm not particularly stuck on the current implementation since this is functionality that is mostly unused outside of niche cases, but it would definitely be good to make sure that the distinction between `FAIL` and `false` for the is_accessible callback is documented appropriately.

qkoziol: Yes, definitely, I'll add some blurbs for it in the morning.

src/H5Fint.c
```diff
@@ -1029,19 +1029,22 @@ H5F_prefix_open_file(bool try, H5F_t **_file, H5F_t *primary_file, H5F_prefix_op
...
     H5FD_t       *lf        = NULL;        /* Low-level file struct            */
     H5F_shared_t *shared    = NULL;        /* Shared part of file              */
     haddr_t       sig_addr  = HADDR_UNDEF; /* Address of hdf5 file signature    */
-    htri_t        ret_value = FAIL;        /* Return value                     */
+    herr_t        ret_value = SUCCEED;     /* Return value                     */
 
     FUNC_ENTER_PACKAGE
 
+    /* Reset output parameter */
+    *is_hdf5 = false;
```
bmribler: This would potentially modify the argument even in the case of failure.

qkoziol: Yes, I considered that, but the previous behavior was aggregating FAIL with 'false' and it seemed best to continue that.

src/H5VLcallback.c
qkoziol: Done

qkoziol: @jhendersonHDF - How's this look?

jhendersonHDF (approved):

bmribler (approved):

src/H5Fint.c
fortnern: It looks like H5VL__native_file_specific() previously would not touch the user's buffer if H5F__is_hdf5() failed, whereas with this change it would, since the user buffer is passed through directly to H5F__is_hdf5(). Also, since it wouldn't be hard to do I think we should just follow the normal pattern where output parameters aren't changed in case of failure. We would of course need another variable here to handle the error check in the close section.

qkoziol: OK, I'll update the code.

fortnern (approved):

PR #4626, opened 2024-07-03
Reworked cleaning up API test files

brtnfld:
Only removing test files if exist to account for skipped tests

### DIFF
--- a/test/API/H5_api_object_test.c
+++ b/test/API/H5_api_object_test.c
@@ -4196,7 +4196,6 @@ test_object_copy_between_files(void)
     hid_t       attr_space_id   = H5I_INVALID_HID;
     hid_t       space_id        = H5I_INVALID_HID;
     hid_t       ocpypl_id       = H5I_INVALID_HID;
-    char        filename[H5_API_TEST_FILENAME_MAX_LENGTH];
 
     TESTING_MULTIPART("object copying between files");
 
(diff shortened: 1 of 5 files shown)


test/API/H5_api_test.c
```diff
@@ -263,7 +263,9 @@ main(int argc, char **argv)
     H5_api_test_run();
 
     printf("Cleaning up testing files\n");
-    H5Fdelete(H5_api_test_filename, fapl_id);
+    if (HDaccess(H5_api_test_filename, F_OK) != -1) {
```
jhendersonHDF: Probably need to use H5Fis_accessible everywhere instead. `access` won't work where there's not a physical file.

brtnfld: fixed

test/API/H5_api_object_test.c
```diff
@@ -7375,6 +7365,23 @@ object_visit_noop_callback(hid_t o_id, const char *name, const H5O_info2_t *obje
...
+ * Cleanup temporary test files
+ */
+static void
+cleanup_files(void)
+{
+    char filename[H5_API_TEST_FILENAME_MAX_LENGTH];
+
+    snprintf(filename, H5_API_TEST_FILENAME_MAX_LENGTH, "%s%s", test_path_prefix,
+             OBJECT_COPY_BETWEEN_FILES_TEST_FILE_NAME);
+    remove_test_file(NULL, filename);
```
jhendersonHDF: Could also just pass `test_path_prefix` to `remove_test_file` here to be even simpler.

jhendersonHDF (approved):

test/API/H5_api_object_test.c
brtnfld: updated

vchoi-hdfgroup (approved):

derobins (approved):

PR #4630, opened 2024-07-08
Implement support for complex number datatypes

jhendersonHDF:
Adds the new datatype class H5T_COMPLEX

Adds the new API function H5Tcomplex_create which creates a complex number datatype from an ID of a base floating-point datatype

Adds the new feature check macros H5_HAVE_COMPLEX_NUMBERS and H5_HAVE_C99_COMPLEX_NUMBERS

Adds the new datatype size macros H5_SIZEOF_FLOAT_COMPLEX, H5_SIZEOF_DOUBLE_COMPLEX and H5_SIZEOF_LONG_DOUBLE_COMPLEX

Adds the new datatype ID macros H5T_NATIVE_FLOAT_COMPLEX, H5T_NATIVE_DOUBLE_COMPLEX, H5T_NATIVE_LDOUBLE_COMPLEX, H5T_CPLX_IEEE_F16LE, H5T_CPLX_IEEE_F16BE,
H5T_CPLX_IEEE_F32LE, H5T_CPLX_IEEE_F32BE,
H5T_CPLX_IEEE_F64LE and H5T_CPLX_IEEE_F64BE

Adds hard and soft datatype conversion paths between complex number datatypes and all the integer and floating-point datatypes, as well as between other complex number datatypes

Adds a special conversion path between complex number datatypes and array or compound datatypes where the in-memory layout of data is the same between the datatypes and data can be converted directly

Adds support for complex number datatypes to the h5dump, h5ls and h5diff/ph5diff tools. Allows h5dump '-m' option to change floating-point printing format for float complex and double complex datatypes, as well as long double complex if it has the same size as double complex

Adds minimal support to the h5watch and h5import tools

Adds support for the predefined complex number datatypes and H5Tcomplex_create function to the Java wrappers. Also adds initial, untested support to the JNI for future use with HDFView

Adds support for just the H5T_COMPLEX datatype class to the Fortran wrappers

Adds support for the predefined complex number datatypes and H5Tcomplex_create function to the high level library H5LT interface for use with the H5LTtext_to_dtype and H5LTdtype_to_text functions

Changes some usages of "complex" in the library since it conflicts with the "complex" keyword from the complex.h header. Also changes various usages of the word "complex" throughout the library to distinguish compound datatypes from complex datatypes.


jhendersonHDF:
Additional complex number features that haven't been implemented yet are mentioned in https://github.com/HDFGroup/hdf5/discussions/4631.

fortran/src/H5Tf.c
```diff
@@ -239,17 +240,7 @@ h5tget_class_c(hid_t_f *type_id, int_f *classtype)
...
-    /*
-      if (c_classtype == H5T_INTEGER)   *classtype = H5T_INTEGER_F;
-      if (c_classtype == H5T_FLOAT)     *classtype = H5T_FLOAT_F;
-      if (c_classtype == H5T_TIME)      *classtype = H5T_TIME_F;
-      if (c_classtype == H5T_STRING)    *classtype = H5T_STRING_F;
-      if (c_classtype == H5T_BITFIELD)  *classtype = H5T_BITFIELD_F;
-      if (c_classtype == H5T_OPAQUE)    *classtype = H5T_OPAQUE_F;
-      if (c_classtype == H5T_COMPOUND)  *classtype = H5T_COMPOUND_F;
-      if (c_classtype == H5T_REFERENCE) *classtype = H5T_REFERENCE_F;
-      if (c_classtype == H5T_ENUM)      *classtype = H5T_ENUM_F;
```
jhendersonHDF, line 251: Didn't look like this commented out section was useful, but can be put back.

release_docs/RELEASE.txt
```diff
@@ -431,6 +431,198 @@ New Features
...
+
+        - H5T_NATIVE_LDOUBLE_COMPLEX
+
+        This macro maps to the ID of an HDF5 datatype representing the native C
+        long double complex datatype (either "long double _Complex" or "_Lcomplex")
+        for the platform. If support for a native long double complex datatype is
+        not available (H5_HAVE_COMPLEX_NUMBERS is not defined), the macro will map
+        to H5I_INVALID_HID and should not be used.
+
+        - H5T_CPLX_IEEE_F16LE / H5T_CPLX_IEEE_F16BE
```
jhendersonHDF: The current datatype docs say that datatype names usually follow a "H5T_arch_base" convention, but that made these names a bit long so I chose this convention for simplicity. Names can be changed if there is a better alternative.

test/dsets.c
```diff
@@ -3915,7 +3915,7 @@ test_nbit_compound_2(hid_t file)
         unsigned int v;
         char         b[2][2];
         atomic       d[2][2];
-    } complex;
```
jhendersonHDF, line 3921: This structure had to be renamed since "complex" is a keyword from complex.h and conflicted

tools/src/h5dump/h5dump_xml.c
jhendersonHDF, line 1: h5dump XML support was added and I verified it with a couple files, but no XML tests were added because the XML output always prints data (either dataset data or fill values) and `long double`/`long double _Complex` data displays differently across platforms due to `sizeof(long double) == sizeof(double)` with MSVC. h5dump would need some work for these tests to be portable.

tools/src/h5repack/h5repack_opttable.c
```diff
@@ -44,47 +44,47 @@ init_packobject(pack_info_t *obj)
...
  *
- * Purpose: auxiliary function, inserts the filter in object OBJS[ I ]
+ * Purpose: auxiliary function, inserts the filter in object OBJS[ idx ]
  *
  * Return: void
  *-------------------------------------------------------------------------
  */
 
 static void
-aux_tblinsert_filter(pack_opttbl_t *table, unsigned int I, filter_info_t filt)
```
jhendersonHDF: `I` had to be changed since it is a keyword from the complex.h header and conflicted.

tools/src/h5dump/h5dump.c
```diff
@@ -1051,11 +1053,41 @@ parse_command_line(int argc, const char *const *argv)
                 h5tools_nCols = 0;
                 break;
 
-            case 'm':
+            case 'm': {
+                char *format_specifier, *plus_flag;
```
jhendersonHDF: h5dump's "-m" option allows specifying the printf specifier for `float` and `double` data, as well as `long double` data if `sizeof(long double) == sizeof(double)`. Rather than having a separate flag for `float/double _Complex` data and a possible additional one for `long double _Complex` data, I just re-used this flag. A better solution is likely needed if XML output will continue to be supported, since the "-m" flag applying to `long double` (`_Complex`) data or not, depending on the platform, makes portable h5dump tests difficult.

tools/lib/h5diff_array.c
```diff
@@ -2391,6 +2530,872 @@ diff_float16_element(unsigned char *mem1, unsigned char *mem2, hsize_t elem_idx,
...
+
+/*-------------------------------------------------------------------------
+ * Function: diff_float_complex
+ *
+ * Purpose:  helper function to diff a single float complex type
+ *
+ * Return:   number of differences found
+ *
+ *-------------------------------------------------------------------------
+ */
```
jhendersonHDF: The `diff_float_complex`/`diff_double_complex`/`diff_ldouble_complex` functions below have the common diffing code that is shared between `diff_float_complex_element`/`diff_double_complex_element`/`diff_ldouble_complex_element` and `diff_complex_element` (the former are used when complex number support is available; the latter otherwise). The code is generally copied directly from `diff_float_element`/`diff_double_element`/etc., but takes into account both the real and imaginary parts.

test/dtypes.c
```diff
@@ -205,7 +211,7 @@ user_compound_convert(hid_t H5_ATTR_UNUSED src_id, hid_t H5_ATTR_UNUSED dst_id,
 static int
 test_classes(void)
 {
-    struct complex { /* Struct with complex fields */
```
jhendersonHDF: Another struct that had to be renamed so it doesn't conflict with the "complex" keyword.

src/H5Tpkg.h
```diff
@@ -41,9 +41,10 @@
 /* Other public headers needed by this file */
 #include "H5Spublic.h" /* Dataspace functions			*/
 
-/* Macro to ease detecting "complex" datatypes (i.e. those with base types or fields) */
-#define H5T_IS_COMPLEX(t)                                                                                    \
-    ((t) == H5T_COMPOUND || (t) == H5T_ENUM || (t) == H5T_VLEN || (t) == H5T_ARRAY || (t) == H5T_REFERENCE)
+/* Macro to ease detecting composite datatypes (i.e. those with base types or fields) */
+#define H5T_IS_COMPOSITE(t)                                                                                  \
```
jhendersonHDF, line 45: Renamed this macro from "H5T_IS_COMPLEX" -> "H5T_IS_COMPOSITE" to distinguish from complex number datatypes.

src/H5Tpkg.h
```diff
@@ -136,8 +144,6 @@
  * floating exception. */
 #if (H5_WANT_DATA_ACCURACY && defined(H5_LDOUBLE_TO_LLONG_ACCURATE)) || (!H5_WANT_DATA_ACCURACY)
 #define H5T_CONV_INTERNAL_LDOUBLE_ULLONG 1
-#else
-#define H5T_CONV_INTERNAL_LDOUBLE_ULLONG 0
```
jhendersonHDF, line 140: The related code that was checking this macro was using `#ifdef` in some places and `#if` in others, so those were updated to just use `#ifdef` and this was changed so that it just doesn't define the macro instead of defining it to 0.

src/H5Tmodule.h
jhendersonHDF, line 1: I haven't yet updated any of the graphics included in this documentation as it would have been difficult to match the original text and styles when updating them. It would be nice to have the graphics in some format that's easier to update.

src/H5Tdbg.c
```diff
@@ -182,6 +182,31 @@ H5T_debug(const H5T_t *dt, FILE *stream)
             s1 = "struct";
             break;
 
+        case H5T_REFERENCE:
```
jhendersonHDF, line 185: Added in missing reference datatype debugging code while adding the new code for complex number datatypes below.

src/H5Tconv_bitfield.c
```diff
@@ -105,7 +105,9 @@ H5T__conv_b_b(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, const H5T_
             }
 
             /* Allocate space for order-reversed source buffer */
-            src_rev = (uint8_t *)H5MM_calloc(src->shared->size);
+            if (conv_ctx->u.conv.cb_struct.func)
```
jhendersonHDF, line 109: This check was added to a few different conversion functions. The `src_rev` buffer will only be used (and therefore only needs to be allocated/freed) if an application datatype conversion exception function is registered and in use.

src/H5Tconv_bitfield.c
```diff
@@ -161,8 +163,7 @@ H5T__conv_b_b(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, const H5T_
                 if (src->shared->u.atomic.prec > dst->shared->u.atomic.prec) {
                     /*overflow*/
                     if (conv_ctx->u.conv.cb_struct.func) { /*If user's exception handler is present, use it*/
-                        H5T__reverse_order(src_rev, s, src->shared->size,
-                                           src->shared->u.atomic.order); /*reverse order first*/
+                        H5T__reverse_order(src_rev, s, src); /*reverse order first*/
```
jhendersonHDF, line 167: Since byte-swapping a complex number value involves swapping the real and imaginary parts separately, `H5T__reverse_order` was updated to just take a pointer to the datatype involved so it can do the right thing.

src/H5Tconv.h
```diff
@@ -25,6 +25,15 @@
 /* Length of debugging name buffer */
 #define H5T_NAMELEN 32
 
+/* Swap two elements (I & J) of an array using a temporary variable */
+#define H5_SWAP_BYTES(ARRAY, I, J)                                                                           \
```
jhendersonHDF, line 29: Moved this macro out of H5Tconv.c to a more general place so it can be reused to make some of the conversion functions more compact.

src/H5T.c
```diff
@@ -746,6 +770,47 @@ H5T__init_inf(void)
         } /* end for */
     }     /* end if */
 
+    /* Get the long double datatype */
```
jhendersonHDF, line 804: +/- Inf for `long double` was previously never set, presumably due to IBM `long double` issues and also since there's currently no larger standard type than `long double`, so the values weren't needed for overflow/underflow checks during conversion. However, complex numbers have some special logic that needed +/- Inf for `long double _Complex` conversions, so I set +/- Inf for `long double` and use it to construct +/- Inf for `long double _Complex`. The value will likely be wrong on machines that use the IBM `long double` format until fixes are made to the library, but this shouldn't be any more problematic than the current situation.

src/H5Odtype.c
```diff
@@ -830,6 +830,44 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
                 dt->shared->force_conv = true;
             break;
 
+        case H5T_COMPLEX:
+            /*
+             * Complex number datatypes...
+             */
+
+            /* Set whether the complex number datatype is homogeneous */
+            dt->shared->u.cplx.homogeneous = (bool)(flags & 0x01);
```
jhendersonHDF: For possible future expansion, use the first bit flag for the datatype to store whether the real and imaginary parts are the same floating-point datatype.

src/H5Odtype.c
```diff
@@ -830,6 +830,44 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
...
+
+            if (!dt->shared->u.cplx.homogeneous)
+                HGOTO_ERROR(H5E_DATATYPE, H5E_UNSUPPORTED, FAIL,
+                            "heterogeneous complex number datatypes are currently unsupported");
+
+            /* Set the form of the complex number datatype */
+            dt->shared->u.cplx.form = (H5T_complex_form_t)((flags >> 1) & 0x03);
```
jhendersonHDF, line 848: From previous discussions, use the next two bits to store the type of the complex number. Currently, only rectangular form is supported.

src/H5Tconv_macros.h
```diff
@@ -804,6 +853,564 @@ typedef struct H5T_conv_hw_t {
...
+
+/* Map all complex number casts to direct casts */
+#define H5T_CONV_CAST_Z(SYMBOLS, STYPE, DTYPE, S, S_REAL, S_IMAG, D, ST, DT)                                 \
+    do {                                                                                                     \
+        *(D) = (DT)(*(S));                                                                                   \
+    } while (0)
+
+#endif
+
+#define H5T_CONV_zZ_CORE(STYPE, DTYPE, S, D, ST, DT, D_MIN, D_MAX)                                           \
```
jhendersonHDF, line 952: All of the macros below are generally identical to their floating-point counterparts (replace "Z" with "F" and "z" with "f"), but some additional logic was needed to extract and handle the real and imaginary parts as necessary. Since MSVC uses structure types for complex numbers, these macros also had to take into account that direct casts to/from complex number types can't be made during conversions when compiled with MSVC.

test/dt_arith.c
```diff
@@ -1630,393 +1746,688 @@ test_derived_integer(void)
...
- * Return:    Success:    0
- *            Failure:    number of errors
+ * Return:      Success:    0
+ *              Failure:    number of errors
  *
  *-------------------------------------------------------------------------
  */
 static int
-test_conv_int_1(const char *name, hid_t src, hid_t dst)
+test_derived_complex(void)
```
jhendersonHDF, line 1760: The git diff isn't great here; `test_derived_complex` was added and `test_conv_int_1` is just now below this function.

test/dt_arith.c
```diff
@@ -2904,6 +3420,39 @@ test_conv_flt_1_hw_conv_from_flt16(void *hw_dst, unsigned char *src_buf, size_t
             *((long double *)hw_dst) = (long double)aligned;
             break;
 #endif
+
+#ifdef H5_HAVE_COMPLEX_NUMBERS
+#ifdef H5_HAVE_C99_COMPLEX_NUMBERS
+        case FLT_COMPLEX:
+            *((H5_float_complex *)hw_dst) = (H5_float_complex)aligned;
```
jhendersonHDF, line 3427: Several places in this file deal with casting values differently depending on whether C99 or MSVC complex number types are supported. While the `H5_CMPLXF` macro below could cover both cases, I wanted to exactly match how each platform would handle casting values to avoid running into possible odd behavior and testing failures. The `H5_CMPLXF` macro has issues with specific special floating-point values that can't be fixed without C11 support.

test/dt_arith.c
```diff
@@ -3167,864 +3838,629 @@ test_conv_flt_1_hw_conv_from_ldouble(void *hw_dst, unsigned char *src_buf, size_
...
- * Return:    Success:    0
- *            Failure:    number of errors
+ * Return:      enum conv_func_ret_t value
  *
  *-------------------------------------------------------------------------
  */
-static int
-test_conv_flt_1(const char *name, int run_test, hid_t src, hid_t dst)
+static conv_func_ret_t
+test_conv_flt_1_hw_conv_from_fcomplex(void *hw_dst, unsigned char *src_buf, size_t idx, dtype_t dst_type)
```
jhendersonHDF, line 3855: Another part of the diff that's not super useful. Similar to changes in the `_Float16` PR, I extracted out the logic that deals with casting from a specific complex number type to another type into helper functions starting here. `test_conv_flt_1` moved down below all these helper functions.

test/dt_arith.c
```diff
@@ -4043,829 +4479,3588 @@ test_conv_int_fp(const char *name, int run_test, hid_t src, hid_t dst)
...
+ * Purpose:     Helper function for test_conv_int_fp to perform conversion
+ *              from a datatype to signed char by casting.
+ *
+ * Return:      -1 on failure
+ *              0 on success
+ *
+ *-------------------------------------------------------------------------
+ */
+static int
+test_conv_int_fp_conv_to_schar(void *hw_p, unsigned char *src_buf, size_t idx, dtype_t src_type)
```
jhendersonHDF, line 5436: Also similar to the changes in the `_Float16` PR, the logic in `test_conv_int_fp` was getting a bit lengthy due to all the types supported, including the new complex number types. I moved each type into a helper function and `test_conv_int_fp` moved below these helper functions. It make this particular diff difficult to follow, but makes the test much easier to read and follow.

src/H5Tconv_integer.c
```diff
@@ -475,425 +477,537 @@ H5T__conv_i_i(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata, const H5T_
 } /* end H5T__conv_i_i() */
 
 /*-------------------------------------------------------------------------
- * Function:    H5T__conv_i_f
+ * Function:    H5T__conv_i_f_loop
```
jhendersonHDF, line 481: The entirety of the conversion logic in `H5T__conv_i_f` was moved into this new helper function so that it could be reused when converting integers to complex numbers as well. `H5T__conv_i_f` moved down below and was updated to just call `H5T__conv_i_f_loop`. The only addition to the conversion logic was updating the byte-swapping portion to do the right thing for complex number values.

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+ *
+ * Purpose:     Implements the body of the conversion loop when converting
+ *              complex number values to another complex number type.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5T__conv_complex_loop(const H5T_t *src_p, const H5T_t *dst_p, const H5T_conv_ctx_t *conv_ctx, size_t nelmts,
```
jhendersonHDF, line 130: Converting between complex number types is fairly similar to converting between floating-point types, but some additional logic is needed to deal with the special rules around when a complex number type is +/- Infinity, NaN, 0, etc.

- A complex or imaginary number is infinite if one of its parts is infinite, even if the other part is NaN.

- A complex or imaginary number is finite if both parts are neither infinities nor NaNs.

- A complex or imaginary number is a zero if both parts are positive or negative zeroes.

Once those rules have been dealt with, each part of the complex number value is converted using the `H5T__conv_complex_part` helper routine, which has nearly identical logic to `H5T__conv_f_f_loop`.

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+ *              and H5T__conv_f_complex functions. Values can be directly
+ *              converted between the types after checking for conversion
+ *              exceptions.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5T__conv_complex_f_matched(const H5T_t *src_p, const H5T_t *dst_p, const H5T_conv_ctx_t *conv_ctx,
```
jhendersonHDF, line 912: When converting between a complex number type and a floating-point type where the complex number type's base type is that same floating-point type, data values can be directly converted with a `memcpy` after checking for conversion exceptions. This function implements that specialized logic.

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+ *                substring of "real", such as "r", or "re" and
+ *                "i(m)(a)(g)(i)(n)(a)(r)(y)" means the field may be named
+ *                any substring of "imaginary", such as "im" or "imag".
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5T__conv_complex_compat(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata,
```
jhendersonHDF, line 1254: This function is just a no-op conversion when converting between complex number datatypes and other datatypes that are in-memory compatible. This provides efficient conversions from h5py-written data in compound datatype format to in-memory complex number format.

src/H5Tconv_float.c
```diff
@@ -42,54 +44,34 @@ H5T__conv_f_f(const H5T_t *src_p, const H5T_t *dst_p, H5T_cdata_t *cdata, const
               size_t nelmts, size_t buf_stride, size_t H5_ATTR_UNUSED bkg_stride, void *buf,
               void H5_ATTR_UNUSED *bkg)
 {
-    /* Traversal-related variables */
```
jhendersonHDF, line 46: The entirety of the conversion logic here was moved into a new `H5T__conv_f_f_loop` helper function so it can be reused when converting between complex number types. No changes to the existing actual conversion logic were made, but some additions were made and refactoring done for readability and reusability. Most notably, the logic that checks for special floating-point values (+/- Inf, NaN, 0) was refactored into a helper function `H5T__conv_float_find_special` that returns an enum value indicating the type of special value found, if any. This compacted the logic a bit and made it reusable since both parts of a complex number have to be checked and the logic would have to have been duplicated otherwise. A few small additions were made just to do the right thing for complex numbers vs. floating-point numbers.

src/H5Tconv_float.c
```diff
@@ -688,602 +755,628 @@ H5T__conv_f_i(const H5T_t *src_p, const H5T_t *dst_p, H5T_cdata_t *cdata, const
             if (NULL == conv_ctx)
                 HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "invalid datatype conversion context pointer");
 
-            src = src_p->shared->u.atomic;
```
jhendersonHDF, line 692: All the conversion logic was extracted to `H5T__conv_f_i_loop` helper function for reuse with complex numbers, with some refactoring to use new `H5T__conv_float_find_special` function and an addition to do the byte-swapping correctly for complex numbers.

HDF5Examples/C/H5T/116/h5ex_t_complex.c
jhendersonHDF, line 1: I added these examples since they're useful, but they currently aren't hooked into the examples project build system since AFAICT the cmake code isn't really setup for distinguishing between MSVC and other compilers. I didn't want to complicate the examples by having more code in them than is necessary, so I added separate examples for C99 vs MSVC complex number types. One of the two should be built depending on the compiler, but not both.

doxygen/dox/DDLBNF116.dox
jhendersonHDF, line 1: Copied from DDLBNF114 and complex number section added

tools/src/h5dump/h5dump_xml.c
```diff
@@ -3446,12 +3548,20 @@ xml_dump_fill_value(hid_t dcpl, hid_t type)
     ctx.indent_level++;
     dump_indent += COL;
 
-    space = H5Tget_size(type);
+    H5E_BEGIN_TRY
+    {
+        n_type = H5Tget_native_type(type, H5T_DIR_DEFAULT);
```
jhendersonHDF: These changes were copied from the DDL output code and fix an issue with XML output where the native type wasn't being used when retrieving the fill value of a dataset. The values displayed would be wrong when dumping a dataset with a different endianness than the machine, for example.

HDF5Examples/C/H5T/116/h5ex_t_complex.c
byrnHDF: There should be CMake variables already from the Compiler checks. Then just use if else in the source_files files for choosing. If we need another check we can add it to both configs.

byrnHDF: There is a config/cmake folder in HDF5Examples that should have the same checks as the library.

release_docs/RELEASE.txt
Dave-Allured: Request.  Use the full word "COMPLEX", not abbreviation, in type names, to improve readability etc.  An extra three characters is not important.  **H5T_COMPLEX_IEEE_F16LE**, etc.

Dave-Allured:
> Adds hard and soft datatype conversion paths between complex number datatypes and all the integer and floating-point datatypes, as well as between other complex number datatypes

Request.  Do not include **any** datatype conversion paths between complex and simple numeric data types.  This will significantly reduce library size and avoid controversies around conventions and other nuances.  Conversions between complex and simple should be left up to the application domain, not baked into the HDF5 library.

Dave-Allured:
> Request. Do not include **any** datatype conversion paths between complex and simple numeric data types. This will significantly reduce library size and avoid controversies around conventions and other nuances. Conversions between complex and simple should be left up to the application domain, not baked into the HDF5 library.

P.S.  My apologies for not catching this sooner and preventing extra work.

src/H5Odtype.c
Dave-Allured: @jhendersonHDF  Thank you for thinking ahead about this.  Polar form is an important basic data type in many high bandwidth radial scanning real world processes.  Conversions between rectangular and polar are problematical.  I hope to see polar supported some time later.

jhendersonHDF:
> > Adds hard and soft datatype conversion paths between complex number datatypes and all the integer and floating-point datatypes, as well as between other complex number datatypes
> 
> Request. Do not include **any** datatype conversion paths between complex and simple numeric data types. This will significantly reduce library size and avoid controversies around conventions and other nuances. Conversions between complex and simple should be left up to the application domain, not baked into the HDF5 library.

Could you describe some of these conventions that applications may use? The current conversion paths should be considered as defaults and they just follow the C99 conversion rules from the standard. An application could always use [H5Tregister](https://portal.hdfgroup.org/hdf5/develop/group___c_o_n_v.html#ga0a23a3cb9f24bd79fae7d2d8c412a25a) to register its own conversion function that takes precedence if there is a specific way of converting data that is desired.

release_docs/RELEASE.txt
jhendersonHDF: Sure, no issue with renaming these if "H5T_COMPLEX_IEEE_F16LE" is better for readability.

ajelenak:
Just to clarify... are these conversions like $1.0$ as a float to $1.0 + 0.0 \cdot i$ as a complex number or something else?

Dave-Allured:
> Could you describe some of these conventions that applications may use? The current conversion paths should be considered as defaults and they just follow the C99 conversion rules from the standard ...

If your conversion paths are simply breaking out the real and imaginary parts into separate parts, and the reverse, then I withdraw my request.  I was momentarily confusing this with complex polar types and their issues.

jhendersonHDF:
> Just to clarify... are these conversions like 1.0 as a float to 1.0+0.0⋅i as a complex number or something else?

Correct; here are the relevant conversion rules from the standard:

"
**6.3.1.6 Complex types**
1 When a value of complex type is converted to another complex type, both the real and
imaginary parts follow the conversion rules for the corresponding real types.
**6.3.1.7 Real and complex**
1 When a value of real type is converted to a complex type, the real part of the complex
result value is determined by the rules of conversion to the corresponding real type and
the imaginary part of the complex result value is a positive zero or an unsigned zero.
2 When a value of complex type is converted to a real type, the imaginary part of the
complex value is discarded and the value of the real part is converted according to the
conversion rules for the corresponding real type.
"

Dave-Allured:
> " **6.3.1.7 Real and complex** 1 When a value of real type is converted to a complex type, the real part of the complex result value is determined by the rules of conversion to the corresponding real type and the imaginary part of the complex result value is a positive zero or an unsigned zero. 2 When a value of complex type is converted to a real type, the imaginary part of the complex value is discarded and the value of the real part is converted according to the conversion rules for the corresponding real type. "

So these are not simple mechanical conversions to extract or combine parts.  They are genuine mathematical transforms.  In my opinion these do not belong in the HDF5 library and should be left out.  The main reason is that they are simply extraneous to the mission of high performance I/O, and would be rarely used.  You are not obliged to support every conversion listed in the standard C library.

Also I think such conversions could lead to misunderstandings and data loss in some scenarios.  For example, someone forgets the in-memory type is simple rather than complex, and code compiles and runs quietly.  Then they find out months later that the imaginary part in their new data set was written all zeros.

I do not think that even extract or combine parts are appropriate.  IMO most applications dealing with type complex will want to read and write full complex slabs as opaque blocks, then deal with extractions and transforms within their own code and their own language.

epourmal:
I completely agree with Dave.

jhendersonHDF:
> So these are not simple mechanical conversions to extract or combine parts. They are genuine mathematical transforms. In my opinion these do not belong in the HDF5 library and should be left out. The main reason is that they are simply extraneous to the mission of high performance I/O, and would be rarely used. You are not obliged to support every conversion listed in the standard C library.
> 
> Also I think such conversions could lead to misunderstandings and data loss in some scenarios. For example, someone forgets the in-memory type is simple rather than complex, and code compiles and runs quietly. Then they find out months later that the imaginary part in their new data set was written all zeros.

I agree that these conversion paths would likely not be used often and could be problematic, but I'd also note that there are other conversion paths currently supported where an application could make a mistake during writing data. It's also worth mentioning that there's currently a bit of overhead associated with user-defined conversion functions. If these paths were removed, an application that does want the standard behavior would need to register its own conversion function and may encounter some of that overhead.

Depending on whether data is being written or read, some of the conversion paths such as `float _Complex` -> `float` seem fairly natural to me. During reads that would essentially be equivalent to extracting the real part of the complex values inside the application code after reading data. `float` -> `float _Complex` would probably be uncommon for both the read and write case though. Not that any of this means the conversion paths need to stay, but I think it's something worth considering.

> I do not think that even extract or combine parts is appropriate. IMO most applications dealing with type complex will want to read and write full complex slabs as opaque blocks, then deal with extractions and transforms within their own code and their own language.

Agreed here; my intention was that applications would just use the standard C functions to extract the parts as necessary after reading data, skipping the datatype conversion pipeline for efficiency. I'm not sure how useful a conversion path to combine parts would be either. I'd generally assume that an application would have data in a final, appropriate in-memory format before I/O. Though, maybe a path to combine parts might be useful in the case of complex types where there isn't a standard C type, such as a complex number of float16 values?

release_docs/RELEASE.txt
qkoziol: Agree - "COMPLEX" is more clear.

byrnHDF (approved):
CMake, Java, tools look good

qkoziol:
> > > Adds hard and soft datatype conversion paths between complex number datatypes and all the integer and floating-point datatypes, as well as between other complex number datatypes
> > 
> > 
> > Request. Do not include **any** datatype conversion paths between complex and simple numeric data types. This will significantly reduce library size and avoid controversies around conventions and other nuances. Conversions between complex and simple should be left up to the application domain, not baked into the HDF5 library.
> 
> Could you describe some of these conventions that applications may use? The current conversion paths should be considered as defaults and they just follow the C99 conversion rules from the standard. An application could always use [H5Tregister](https://portal.hdfgroup.org/hdf5/develop/group___c_o_n_v.html#ga0a23a3cb9f24bd79fae7d2d8c412a25a) to register its own conversion function that takes precedence if there is a specific way of converting data that is desired.

I think that following the C Standard is appropriate, especially since it's possible to override.

qkoziol:
> > > > Adds hard and soft datatype conversion paths between complex number datatypes and all the integer and floating-point datatypes, as well as between other complex number datatypes
> > > 
> > > 
> > > Request. Do not include **any** datatype conversion paths between complex and simple numeric data types. This will significantly reduce library size and avoid controversies around conventions and other nuances. Conversions between complex and simple should be left up to the application domain, not baked into the HDF5 library.
> > 
> > 
> > Could you describe some of these conventions that applications may use? The current conversion paths should be considered as defaults and they just follow the C99 conversion rules from the standard. An application could always use [H5Tregister](https://portal.hdfgroup.org/hdf5/develop/group___c_o_n_v.html#ga0a23a3cb9f24bd79fae7d2d8c412a25a) to register its own conversion function that takes precedence if there is a specific way of converting data that is desired.
> 
> I think that following the C Standard is appropriate, especially since it's possible to override.

However - have you confirmed that overriding works correctly with a regression test?   (Which will also be useful for users who wish to override)

jhendersonHDF:
> However - have you confirmed that overriding works correctly with a regression test? (Which will also be useful for users who wish to override)

Currently, there's only a test to make sure that the registered function gets called instead and that the datatype IDs passed are valid, but I could add a more comprehensive test.

config/cmake/ConfigureChecks.cmake
```diff
@@ -816,6 +817,71 @@ macro (H5ConversionTests TEST def msg)
...
+      set (${HDF_PREFIX}_SIZEOF_LONG_DOUBLE_COMPLEX ${${HDF_PREFIX}_SIZEOF__LCOMPLEX}
+           CACHE INTERNAL "SizeOf for long double _Complex" FORCE)
+
+      unset (H5_HAVE_C99_COMPLEX_NUMBERS)
+    endif ()
+  endif ()
+
+  if (${HDF_PREFIX}_SIZEOF_FLOAT_COMPLEX AND ${HDF_PREFIX}_SIZEOF_DOUBLE_COMPLEX AND
+      ${HDF_PREFIX}_SIZEOF_LONG_DOUBLE_COMPLEX)
+    # Check if __STDC_NO_COMPLEX__ macro is defined, in which case complex number
```
qkoziol, line 870: Why depend on all three sizes?  Shouldn't this be an 'OR'?

qkoziol:
> > However - have you confirmed that overriding works correctly with a regression test? (Which will also be useful for users who wish to override)
> 
> Currently, there's only a test to make sure that the registered function gets called instead and that the datatype IDs passed are valid, but I could add a more comprehensive test.

Seems sufficient to me.

configure.ac
```diff
@@ -574,6 +574,58 @@ AC_CHECK_SIZEOF([float])
...
+HAVE_COMPLEX_NUMBERS="no"
+AC_CHECK_SIZEOF([float complex], [], [#include <complex.h>])
+AC_CHECK_SIZEOF([double complex], [], [#include <complex.h>])
+AC_CHECK_SIZEOF([long double complex], [], [#include <complex.h>])
+
+if test "$ac_cv_sizeof_float_complex" != 0 &&
+   test "$ac_cv_sizeof_double_complex" != 0 &&
+   test "$ac_cv_sizeof_long_double_complex" != 0; then
+  # Check if __STDC_NO_COMPLEX__ macro is defined, in which case complex number
+  # support is not available
```
qkoziol, line 589: Same here ('or' instead of 'and')

config/cmake/ConfigureChecks.cmake
jhendersonHDF: Since all three types are required by the standard if complex numbers are supported, it didn't seem worth the effort and extra special-case code to deal with each type individually. No doubt that there probably is or was some compiler that only supports 1 or 2 of the types, but all modern ones appear to support complex numbers according to the standard (except for the oddness with MSVC).

doxygen/examples/H5.format.html
```diff
@@ -11523,6 +11543,119 @@ <h4><a name="DatatypeMessage">IV.A.2.d. The Datatype Message</a></h4>
...
+
+      <tr>
+        <th width="25%">Byte</th>
+        <th width="25%">Byte</th>
+        <th width="25%">Byte</th>
+        <th width="25%">Byte</th>
+      </tr>
+
+      <tr>
+        <td colspan="4"><br />Parent Type Message<br /><br /></td>
```
qkoziol, line 11635: This may be awkward if/when heterogenous complex types are supported.

doxygen/examples/H5.format.html
```diff
@@ -11523,6 +11543,119 @@ <h4><a name="DatatypeMessage">IV.A.2.d. The Datatype Message</a></h4>
                         </table>
                       </div>
 
+<br />
+<br />
+<a name="ClassComplex"></a>
+  <p>Class specific information for the Complex class (Class 11):</p>
```
qkoziol, line 11551: I suggest adding a version to this info, which will make supporting heterogenous complex types easier if they are added in the future.

release_docs/RELEASE.txt
```diff
@@ -431,6 +431,198 @@ New Features
...
+        H5T_NATIVE_FLOAT16 <-> H5T_NATIVE_DOUBLE_COMPLEX | H5T_NATIVE_FLOAT   <-> H5T_NATIVE_DOUBLE_COMPLEX 
+        H5T_NATIVE_DOUBLE  <-> H5T_NATIVE_DOUBLE_COMPLEX | H5T_NATIVE_LDOUBLE <-> H5T_NATIVE_DOUBLE_COMPLEX
+
+        H5T_NATIVE_SCHAR   <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_UCHAR   <-> H5T_NATIVE_LDOUBLE_COMPLEX
+        H5T_NATIVE_SHORT   <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_USHORT  <-> H5T_NATIVE_LDOUBLE_COMPLEX
+        H5T_NATIVE_INT     <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_UINT    <-> H5T_NATIVE_LDOUBLE_COMPLEX
+        H5T_NATIVE_LONG    <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_ULONG   <-> H5T_NATIVE_LDOUBLE_COMPLEX
+        H5T_NATIVE_LLONG   <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_ULLONG  <-> H5T_NATIVE_LDOUBLE_COMPLEX
+        H5T_NATIVE_FLOAT16 <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_FLOAT   <-> H5T_NATIVE_LDOUBLE_COMPLEX 
+        H5T_NATIVE_DOUBLE  <-> H5T_NATIVE_LDOUBLE_COMPLEX | H5T_NATIVE_LDOUBLE <-> H5T_NATIVE_LDOUBLE_COMPLEX
```
qkoziol, line 286: I suggest mentioning that these follow the C Standard for their behavior.

release_docs/RELEASE.txt
```diff
@@ -431,6 +431,198 @@ New Features
...
+
+        H5T_NATIVE_FLOAT_COMPLEX  <-> H5T_NATIVE_DOUBLE_COMPLEX
+        H5T_NATIVE_FLOAT_COMPLEX  <-> H5T_NATIVE_LDOUBLE_COMPLEX
+        H5T_NATIVE_DOUBLE_COMPLEX <-> H5T_NATIVE_LDOUBLE_COMPLEX
+
+      Alternative software implementation conversion paths have been added for all of
+      the above for use when native complex number support is not available.
+
+      Additionally, a special datatype conversion path has been added between complex number
```
qkoziol, line 297: It would be useful to add these conversions as options to h5repack as well.   I'll add a comment on https://github.com/HDFGroup/hdf5/discussions/4631 .

src/H5Odtype.c
```diff
@@ -830,6 +830,44 @@ H5O__dtype_decode_helper(unsigned *ioflags /*in,out*/, const uint8_t **pp, H5T_t
...
+
```
qkoziol, line 849: Can you also protect against fuzzers by throwing an error for any flags (and anything else) that are currently undefined?

src/H5Odtype.c
```diff
@@ -1644,6 +1710,13 @@ H5O__dtype_size(const H5F_t *f, const void *_mesg)
             ret_value += H5O__dtype_size(f, dt->shared->parent);
             break;
 
+        case H5T_COMPLEX:
+            if (dt->shared->u.cplx.homogeneous)
+                ret_value += H5O__dtype_size(f, dt->shared->parent);
+            else
+                ret_value = 0;
```
qkoziol: '0' seems unhelpful.  Throw an error instead?

jhendersonHDF: The function currently returns a `size_t` (0 for failure), though it could be updated.

src/H5Odtype.c
jhendersonHDF: I'm not sure which things to check for here. Just throw an error if any of the bits above bits 0,1,2 are set?

src/H5T.c
```diff
@@ -789,6 +854,22 @@ H5T__init_inf(void)
...
+     * positive/negative infinity value for the base floating-point type.
+     * The imaginary part should be zeroed-out bits.
+     */
+    memcpy(&H5T_NATIVE_FLOAT_COMPLEX_POS_INF_g, &H5T_NATIVE_FLOAT_POS_INF_g, sizeof(float));
+    memcpy(&H5T_NATIVE_FLOAT_COMPLEX_NEG_INF_g, &H5T_NATIVE_FLOAT_NEG_INF_g, sizeof(float));
+    memcpy(&H5T_NATIVE_DOUBLE_COMPLEX_POS_INF_g, &H5T_NATIVE_DOUBLE_POS_INF_g, sizeof(double));
+    memcpy(&H5T_NATIVE_DOUBLE_COMPLEX_NEG_INF_g, &H5T_NATIVE_DOUBLE_NEG_INF_g, sizeof(double));
+    memcpy(&H5T_NATIVE_LDOUBLE_COMPLEX_POS_INF_g, &H5T_NATIVE_LDOUBLE_POS_INF_g, sizeof(long double));
+    memcpy(&H5T_NATIVE_LDOUBLE_COMPLEX_NEG_INF_g, &H5T_NATIVE_LDOUBLE_NEG_INF_g, sizeof(long double));
+#endif
```
qkoziol, line 902: This seems odd.   Why are the complex infinities not of _Complex type?   If it's correct, then why not just use the regular float, double, etc, types and not define complex infinities?

src/H5T.c
```diff
@@ -4294,12 +4730,28 @@ H5T__set_size(H5T_t *dt, size_t size)
...
+        if (dt->shared->type == H5T_COMPLEX) {
+            if (!dt->shared->u.cplx.homogeneous)
+                HGOTO_ERROR(H5E_DATATYPE, H5E_BADTYPE, FAIL,
+                            "heterogeneous complex number datatypes are currently unsupported");
+
+            if ((new_size % 2) != 0)
+                HGOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL,
+                            "new datatype size is not evenly divisible by 2");
+
+            new_size /= 2;
```
qkoziol: The 'size' parameter is for the parent type (e.g. the way the array type works), so I don't think we want to divide by 2.

src/H5Odtype.c
qkoziol: Ah, OK, didn't expand upward.  :-)

src/H5Odtype.c
qkoziol: That's good for the flags.  I guess that the parent type is the only other thing, so that should be all.

config/cmake/ConfigureChecks.cmake
qkoziol: OK, seems sensible.

src/H5Tconv.c
```diff
@@ -185,27 +177,47 @@ H5T_get_force_conv(const H5T_t *dt)
...
+    size = dtype->shared->size;
+
+    if (H5T_IS_ATOMIC(dtype->shared))
+        order = dtype->shared->u.atomic.order;
+    else
+        order = dtype->shared->parent->shared->u.atomic.order;
 
     if (H5T_ORDER_VAX == order) {
-        for (i = 0; i < size; i += 2) {
+        for (size_t i = 0; i < size; i += 2) {
```
qkoziol, line 199: Throw an error for complex here?

src/H5Tconv.c
```diff
@@ -281,12 +296,38 @@ H5T__conv_order(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata,
...
+            }
+            else {
+                dst_offset = dst->shared->u.atomic.offset;
+                dst_order  = dst->shared->u.atomic.order;
+            }
+
+            if (0 != src_offset || 0 != dst_offset)
+                HGOTO_ERROR(H5E_DATATYPE, H5E_UNSUPPORTED, FAIL, "conversion not supported");
+
+            if (!((H5T_ORDER_BE == src_order && H5T_ORDER_LE == dst_order) ||
```
qkoziol, line 329: Are you disallowing VAX conversions for atomic types?

src/H5T.c
jhendersonHDF: They are of `_Complex` type, but only their real part gets set to +/- Infinity. See the comment above these lines. Since the H5Tconv_macros.h macros take in a partial type name like `DOUBLE_COMPLEX` and then do token pasting to create the name of the matching +/- Infinity variable, I would have to have another layer of macro indirection to redirect the macro from `DOUBLE_COMPLEX` to `DOUBLE` to get the variable `H5T_NATIVE_DOUBLE_NEG_INF_g`, for example.

```C
#define H5T_CONV_Zf_NOEX_CORE_IMP(STYPE, DTYPE, S, D, ST, DT, SBT, D_MIN, D_MAX)                             \
        H5T_CONV_##STYPE##_REALVAL(S); /* Extract "real" part of complex number */                           \
        if ((sr_val) > (SBT)(D_MAX))                                                                         \
            *(D) = H5_GLUE3(H5T_NATIVE_, DTYPE, _POS_INF_g);                                                 \
        else if ((sr_val) < (SBT)(D_MIN))                                                                    \
            *(D) = H5_GLUE3(H5T_NATIVE_, DTYPE, _NEG_INF_g);                                                 \
        else                                                                                                 \
            *(D) = (DT)((sr_val));                                                                           \
    }
```

src/H5T.c
jhendersonHDF: When it comes to 'container' types, it's inconsistent whether setting the size is for the whole type or a base type, even though the docs for `H5Tset_size` say "H5Tset_size() sets the total size, size, in bytes, for a datatype". To follow the docs for complex numbers, I chose to have `H5Tset_size` set the size for the whole datatype, in which case the size of the parent type is half the datatype size.

src/H5Tconv.c
jhendersonHDF: This is the same logic from before, just compacted a bit with temporary variables for readability. For complex numbers, I decided to throw an error for VAX during conversion based off this comment: https://github.com/HDFGroup/hdf5/blob/96acf6c73910b7271bdb2c0248d3b588d9f45e2b/src/H5Tconv_float.c#L302-L306. Looks like that comment got removed when I added the helper function to retrieve the type of floating-point value involved; I'll put that back.

src/H5T.c
qkoziol: OK, I see that they are _Complex, and I understand the standard's behavior.  It's the way that only half of the value is being initialized that was concerning me.  I suppose that it's in the C89 standard to have the other half zero-initialized, so should be fine.

src/H5T.c
qkoziol: I could go either way on this, I suppose, since Complex straddles the line between array / compound types and atomic types.  I'd lean toward the way that arrays work, but don't feel strongly enough about it to insist.   @fortnern @derobins - where do you land on this?

Definitely needs to be documented, either way - could you add a blurb to the H5Tset_size()  documentation?

src/H5Tconv.c
qkoziol: Ah, the original code had scrolled off the screen and I missed that this was the existing behavior.

src/H5T.c
jhendersonHDF: See the changes to H5Tpublic.h (~line 2519)

qkoziol: > See the changes to H5Tpublic.h

OK, found that, thanks.   (I had been looking for them in H5Tmodule.h)

Still makes me feel like H5Tset_size() should be applied to the base type...

src/H5Tconv.c
```diff
@@ -185,27 +177,47 @@ H5T_get_force_conv(const H5T_t *dt)
  *-------------------------------------------------------------------------
  */
 herr_t
-H5T__reverse_order(uint8_t *rev, uint8_t *s, size_t size, H5T_order_t order)
+H5T__reverse_order(uint8_t *rev, uint8_t *s, const H5T_t *dtype)
```
qkoziol, line 180: Good catch!

src/H5Tconv.c
```diff
@@ -307,6 +348,23 @@ H5T__conv_order(const H5T_t *src, const H5T_t *dst, H5T_cdata_t *cdata,
                     } /* end if */
                     break;
 
+                case H5T_COMPLEX: {
+                    const H5T_shared_t *src_base_sh = src->shared->parent->shared;
```
qkoziol, line 352: Maybe use H5T_cmp() here instead?

src/H5T.c
jhendersonHDF: My personal take is that it may be surprising to some that certain datatypes adjust their size after calling `H5Tset_size` on them, despite the fact that it's mentioned in the documentation. Especially so since Compound datatypes are a container type where `H5Tset_size` actually does set the size of the whole datatype. For some datatypes like variable-length this is of course awkward, but I would've had that type be the sole exception. I likely would have implemented `H5Tset_size` so that it sets the whole size for an array datatype. It makes one have to think a bit more about the details, but avoids having "extra" behavior in `H5Tset_size` and allows for catching some possible programming errors like an array datatype where the size isn't evenly divisible by the number of array elements.

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+        }
+        else if (real_except || imag_except) {
+            except_ret = H5T_CONV_UNHANDLED;
+
+            /* If user's exception handler is present, use it */
+            if (conv_ctx->u.conv.cb_struct.func) {
+                H5T_conv_except_t except_type; /* type of conversion exception that occurred */
+
+                /* reverse source buffer order first */
+                H5T__reverse_order(src_rev, s, src_p);
```
qkoziol, line 284: Why do you always reverse the order?

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+                                                               conv_ctx->u.conv.dst_type_id, src_rev, d,
+                                                               conv_ctx->u.conv.cb_struct.user_data);
+            }
+
+            if (except_ret == H5T_CONV_UNHANDLED) {
+                if (realval_type == H5T_CONV_FLOAT_SPECVAL_POSINF ||
+                    realval_type == H5T_CONV_FLOAT_SPECVAL_NEGINF ||
+                    imagval_type == H5T_CONV_FLOAT_SPECVAL_POSINF ||
+                    imagval_type == H5T_CONV_FLOAT_SPECVAL_NEGINF) {
+                    H5T__bit_copy(d, dst_atomic.u.f.sign, s, src_atomic.u.f.sign, (size_t)1);
```
qkoziol, line 321: Could you copy the global infinity values here?

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+                         * the 1st bit of mantissa to 1. The Intel-Linux "long double" is
+                         * this case. */
+                        if (H5T_NORM_NONE == dst_atomic.u.f.norm)
+                            H5T__bit_set(d + dst_part_size, dst_atomic.u.f.mpos + dst_atomic.u.f.msize - 1,
+                                         (size_t)1, true);
+                    }
+                }
+                else {
+                    /* There are many NaN values, so we just set all bits of the significand. */
+                    if (realval_type == H5T_CONV_FLOAT_SPECVAL_NAN) {
```
qkoziol, line 349: Would probably be worthwhile setting up globals for NANs also (in additional to infinities), so that memcpy() could be used here to speed things up.

src/H5Tconv.c
jhendersonHDF: It would be more compact, though `H5T_cmp` will compare the datatypes as different if they have different byte orders and then there's no information about whether or not that was the only thing different about the datatypes.

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+                goto next;
+            }
+            else if (except_ret == H5T_CONV_ABORT)
+                HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCONVERT, FAIL, "can't handle conversion exception");
+
+            goto padding;
+        }
+
+        if (real_zero) {
+            H5T__bit_copy(d, dst_atomic.u.f.sign, s, src_atomic.u.f.sign, (size_t)1);
```
qkoziol, line 374: Globals for memcpy() would be good here too.

src/H5Tconv_complex.c
jhendersonHDF: This just follows the logic in all the other conversion functions. They all convert the values to little-endian first to make the loops less complicated and then at the point that an exception is thrown the buffer is reversed if necessary before handing it to the exception callback function. If the data is already little-endian this function just copies `s` into `src_rev` (though that's not really necessary and could be optimized).

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+        dp += dst_delta;
+    } /* end conversion loop */
+
+done:
+    H5MM_free(src_rev);
+
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5T__conv_complex_loop() */
+
+/*-------------------------------------------------------------------------
```
qkoziol, line 470: Maybe move this code to a 'static inline' routine in H5Tconv.h and invoke it from H5T__conv_f_f_loop and H5T__conv_complex_loop?   I bet that the compiler will inline it correctly and eliminate any performance gap.   Getting rid of the duplicated code would be very worthwhile from a maintenance perspective.

src/H5Tconv_complex.c
jhendersonHDF: Sure, as long as the type being converted matches one of the native types. Would probably make it a bit uglier here with special-case code and realistically this function is only ever going to be used for non-standard complex types like a complex of float16 anyway. I'm sure several of the library's software conversion functions could be heavily optimized like this, though I don't know if it's worth doing for this PR.

src/H5Tconv_complex.c
```diff
@@ -0,0 +1,2323 @@
...
+             * type consisting of the same floating-point type?
+             */
+            equal_cplx_conv = (0 == H5T_cmp(src_p->shared->parent, dst_p, false));
+            if (!equal_cplx_conv) {
+                /* If floating-point types differ, use generic f_f loop */
+                if (H5T__conv_f_f_loop(src_p, dst_p, conv_ctx, nelmts, buf_stride, buf) < 0)
+                    HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCONVERT, FAIL, "unable to convert data values");
+            }
+            else {
+                /* If floating-point types are the same, use specialized loop */
```
qkoziol, line 881: Isn't this the same for floating-point type conversion?  Could the code be shared with them?

src/H5Tconv_complex.c
jhendersonHDF: Sure, I can probably do this. I must admit that I was a bit concerned about potential function call overhead in such a hot loop, but this function also isn't likely to be used very often unless applications make use of non-standard complex number types.

src/H5Tconv_float.c
```diff
@@ -100,524 +82,625 @@ H5T__conv_f_f(const H5T_t *src_p, const H5T_t *dst_p, H5T_cdata_t *cdata, const
...
+         * Set external padding areas
+         */
+        if (dst_atomic.offset > 0) {
+            assert(H5T_PAD_ZERO == dst_atomic.lsb_pad || H5T_PAD_ONE == dst_atomic.lsb_pad);
+            H5T__bit_set(d, (size_t)0, dst_atomic.offset, (bool)(H5T_PAD_ONE == dst_atomic.lsb_pad));
+        }
+        {
+            size_t type_size;
+
+            if (dst_p->shared->type == H5T_FLOAT)
```
qkoziol, line 569: This could be hoisted out of the loop

src/H5Tconv_float.c
```diff
@@ -688,602 +755,628 @@ H5T__conv_f_i(const H5T_t *src_p, const H5T_t *dst_p, H5T_cdata_t *cdata, const
...
-                                goto next;
-                            }
-                            else if (except_ret == H5T_CONV_ABORT)
-                                HGOTO_ERROR(H5E_DATATYPE, H5E_CANTCONVERT, FAIL,
-                                            "can't handle conversion exception");
-                        }
-                        goto padding;
+
+                    if (except_ret == H5T_CONV_UNHANDLED)
+                        H5T__bit_set(d, (dst_atomic.offset + dst_atomic.prec - 1), (size_t)1, true);
```
qkoziol, line 1200: We're using the H5T__bit_set, etc. routines so heavily now, I suggest moving them into one of the H5T headers and making them 'static inline', to speed them up.

src/H5Tconv_complex.c
qkoziol: OK, I see that, but why are we always assuming that?   It doesn't necessarily match the datatype IDs.   Seems like an existing bug that would show up with exception handing callbacks on big-endian machines.

src/H5Tconv.c
qkoziol, line 180: After looking at this again, the function name is very misleading.  It should be "H5T__to_le_order" or similar.  If you are on a little-endian machine, it definitely doesn't reverse the order.

src/H5Tconv_complex.c
jhendersonHDF: This would just be putting the value buffer back as it was at the start of the loop before initial byte-swapping, so I'm not sure where the bug would be at.

src/H5Tconv.c
jhendersonHDF: Agreed and ideally we would be able to refactor the conversion functions to not have to byte-swap data into little-endian order in the first place, removing the need for reversing the buffer before calling the exception callback function.

src/H5Tpkg.h
```diff
@@ -291,6 +297,19 @@ typedef struct H5T_array_t {
     size_t   dim[H5S_MAX_RANK]; /* size in each dimension       */
 } H5T_array_t;
 
+/* Form (Rectangular, Polar, Exponential) of complex number */
+typedef enum H5T_complex_form_t {
+    H5T_COMPLEX_RECTANGULAR,
+    H5T_COMPLEX_POLAR,
+    H5T_COMPLEX_EXPONENTIAL,
```
qkoziol, line 304: Some compilers are grumpy about the last item in an enum having a ','

src/H5Tpkg.h
```diff
@@ -291,6 +297,19 @@ typedef struct H5T_array_t {
...
+} H5T_complex_form_t;
+
+/* A complex number datatype */
+typedef struct H5T_complex_t {
+    H5T_complex_form_t form;        /* Form (Rectangular, Polar, Exponential) of complex number */
+    bool               homogeneous; /* Whether the parts of the complex number have the same datatype */
```
qkoziol: The 'homogeneous' flag is just extra overhead, since we don't support heterogeneous complex types (and probably never will, I'm guessing).   I suggest removing it, to simplify and clarify the code.

src/H5Tpkg.h
jhendersonHDF: It's allowed in C99 though! Sounds like some compilers need to be fixed.

src/H5T.c
qkoziol: I see that H5Tset_size is working on the entire type, but H5Tset_order and H5Tset_precision are working on the base type (which I agree with).  This just adds to the confusion here and I now think we should align all three and have H5Tset_size work on the base type also.

jhendersonHDF: The current situation makes perfect sense to me though because container types don't generally have an order or precision since they aren't atomic types.

src/H5Tconv_complex.c
qkoziol: OK, yes, I agree with you.  The asymmetry is annoying: on non-LE types, it byte swaps, but is memcpy for LE types.  Very weird.

src/H5Tpkg.h
qkoziol: 🤷‍♂️

tools/lib/h5tools.h
```diff
@@ -227,67 +229,82 @@ typedef struct h5tool_format_t {
...
-     *               printed as numeric values.  The default is zero.
+     *   fmt_float:            The printf() format to use when rendering data which is
+     *                         typed `float'. The default is `%g'.
      *
-     *   str_locale: Determines how strings are printed. If zero then strings
-     *               are printed like in C except. If set to ESCAPE_HTML then
-     *               strings are printed using HTML encoding where each
-     *               character not in the class [a-zA-Z0-9] is substituted
-     *               with `%XX' where `X' is a hexadecimal digit.
+     *   fmt_float_complex:    The printf() format to use when rendering data which is
```
qkoziol: Should probably follow the same (big->little) ordering as for floating-point types.

qkoziol: Or switch the floating-point types to be little->big also.

tools/test/h5ls/expected/tattr2.ls
```diff
@@ -12,21 +12,30 @@ dset                     Dataset {2/2}
...
+    Attribute: complex3D {4, 3, 2}
+        Type:      complex number of
+                   IEEE 32-bit little-endian float
     Attribute: compound {2}
         Type:      struct {
                    "a"                +0    8-bit integer
-                   "b"                +4    IEEE 64-bit little-endian float
-               } 12 bytes
+                   "b"                +8    IEEE 64-bit little-endian float
+               } 16 bytes
```
qkoziol: Why did this change?

qkoziol (changes requested):
Amazing amount of great work here!

Four changes seem important to me:
- Add the fuzzing guards on the complex datatype's encoding.
- Add the version field to the complex datatype encoding, so that if we ever add heterogeneous complex types, we don't have to write as complex of a parser.
- Switch H5Tset_size's behavior
- Remove the 'homogeneous' flag from the complex datatype struct.

tools/test/h5ls/expected/tattr2.ls
jhendersonHDF: My guess is that the file must have previously been generated on some system where the compiler used different structure padding.

src/H5Tpkg.h
jhendersonHDF: Sure, this was mostly thinking ahead since the related forum post already mentioned some use of heterogeneous complex types, but it's not really necessary to have around right now. The same could be said about the form of the complex number, but it seems appropriate to have at least something for the datatype if for no other purpose than error checking.

jhendersonHDF:
>     * Add the fuzzing guards on the complex datatype's encoding.

The previous comment was made in the decode helper routine. Do we want checks both during encode and decode?

>     * Add the version field to the complex datatype encoding, so that if we ever add heterogeneous complex types, we don't have to write as complex of a parser.

What's the purpose of having a complex number-specific encoding version field here? Is it just to check that no heterogeneous complex types are allowed in, e.g. version 0?

qkoziol:
> > ```
> > * Add the fuzzing guards on the complex datatype's encoding.
> > ```
> 
> The previous comment was made in the decode helper routine. Do we want checks both during encode and decode?

Sorry, I meant decoding.

> 
> > ```
> > * Add the version field to the complex datatype encoding, so that if we ever add heterogeneous complex types, we don't have to write as complex of a parser.
> > ```
> 
> What's the purpose of having a complex number-specific encoding version field here? Is it just to check that no heterogeneous complex types are allowed in, e.g. version 0?

🤔   I worked through all the permutations I can think of again, and I think it'll be OK.  I was concerned about the possible addition of another parent datatype for the heterogeneous case, but I think the existing flag will be OK for the forward/backward compatibility  So, I think we can skip adding the version field.

jhendersonHDF:
Most of the major previous comments should be addressed now. I haven't made the change to `H5Tset_size` yet as I still feel fairly strongly about that and I don't think we necessarily came to any real conclusion on it. There are various other comments on other possible optimizations or refactoring that could be made, but many of them seem like things that could fall outside this PR.

qkoziol:
> Most of the major previous comments should be addressed now. I haven't made the change to `H5Tset_size` yet as I still feel fairly strongly about that and I don't think we necessarily came to any real conclusion on it. There are various other comments on other possible optimizations or refactoring that could be made, but many of them seem like things that could fall outside this PR.

Perhaps we should have a short, small audience meeting to discuss the H5Tset_size() issue?

qkoziol (approved):

derobins (approved):

Issue #4634, opened 2024-07-09
H5VL_t not freed properly if closed with no references


mattjala:
**Describe the bug**
`H5VL_new_connector()` constructs an `H5VL_t` instance from a connector class. The `nrefs` field of the new `H5VL_t` is initialized to zero by `H5FL_CALLOC`. If the connector object is closed before the reference count is incremented (for example, within `H5ESinsert_request()` - if `H5ES__insert_request()` fails to insert a request) then the cleanup routine `H5VL_conn_dec_rc()` will decrement the reference count to a negative value, and then not free the memory because it expects the reference count to be exactly zero if the connector needs to be freed.

**Expected behavior**
Memory allocated for `H5VL_t` should be freed properly regardless of later failures.

mattjala:
Resolved in #4856

PR #4635, opened 2024-07-09
Avoid attempting to close connector with zero references

mattjala:
Fix for #4634 (H5VL_t not freed properly if closed with no references).

I don't think an assertion would make sense here, since it's possible for regular internal library failures to elicit this issue.

### DIFF
--- a/src/H5VLint.c
+++ b/src/H5VLint.c
@@ -265,8 +265,8 @@ H5VL_term_package(void)
                 /* Destroy the VOL connector ID group */
                 n += (H5I_dec_type_ref(H5I_VOL) > 0);
             } /* end else */
-        }     /* end else */
-    }         /* end else */
+        } /* end else */
+    } /* end else */
 
     FUNC_LEAVE_NOAPI(n)
 } /* end H5VL_term_package() */
@@ -330,14 +330,14 @@ H5VL__get_connector_cb(void *obj, hid_t id, void *_op_data)
             op_data->found_id = id;
             ret_value         = H5_ITER_STOP;
         } /* end if */
-    }     /* end if */
+    } /* end if */
     else {
         assert(H5VL_GET_CONNECTOR_BY_VALUE == op_data->key.kind);
         if (cls->value == op_data->key.u.value) {
             op_data->found_id = id;
             ret_value         = H5_ITER_STOP;
         } /* end if */
-    }     /* end else */
+    } /* end else */
 
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5VL__get_connector_cb() */
@@ -425,7 +425,7 @@ H5VL__set_def_conn(void)
                     0)
                     HGOTO_ERROR(H5E_VOL, H5E_CANTREGISTER, FAIL, "can't register connector");
             } /* end else */
-        }     /* end else */
+        } /* end else */
 
         /* Was there any connector info specified in the environment variable? */
         if (NULL != (tok = HDstrtok_r(NULL, "\n\r", &lasts)))
@@ -627,8 +627,8 @@ H5VL_conn_copy(H5VL_connector_prop_t *connector_prop)
                 /* Set the connector info to the copy */
                 connector_prop->connector_info = new_connector_info;
             } /* end if */
-        }     /* end if */
-    }         /* end if */
+        } /* end if */
+    } /* end if */
 
 done:
     FUNC_LEAVE_NOAPI(ret_value)
@@ -827,9 +827,14 @@ H5VL_register_using_vol_id(H5I_type_t type, void *obj, hid_t connector_id, bool 
     if (NULL == (connector = H5VL_new_connector(connector_id)))
         HGOTO_ERROR(H5E_VOL, H5E_CANTCREATE, H5I_INVALID_HID, "can't create VOL connector object");
 
+    /* Increment reference count above 0 so that connector can be closed if VOL object creation fails */
+    connector->nrefs++;
+
     /* Get an ID for the VOL object */
     if ((ret_value = H5VL_register(type, obj, connector, app_ref)) < 0)
         HGOTO_ERROR(H5E_VOL, H5E_CANTREGISTER, H5I_INVALID_HID, "unable to register object handle");
+
+    connector->nrefs--;
 
 done:
     /* Clean up on error */
@@ -982,12 +987,13 @@ H5VL_conn_dec_rc(H5VL_t *connector)
 
     /* Check arguments */
     assert(connector);
+    assert(connector->nrefs >= 0);
 
     /* Decrement refcount for connector */
     connector->nrefs--;
 
     /* Check for last reference */
-    if (0 >= connector->nrefs) {
+    if (0 == connector->nrefs) {
         if (H5I_dec_ref(connector->id) < 0)
             HGOTO_ERROR(H5E_VOL, H5E_CANTDEC, FAIL, "unable to decrement ref count on VOL connector");
         H5FL_FREE(H5VL_t, connector);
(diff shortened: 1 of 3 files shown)


src/H5VLint.c
```diff
@@ -987,7 +987,7 @@ H5VL_conn_dec_rc(H5VL_t *connector)
     connector->nrefs--;
 
     /* Check for last reference */
-    if (0 == connector->nrefs) {
+    if (0 >= connector->nrefs) {
```
qkoziol: This doesn't look correct to me.  We should only decrement the refcount on the ID when the connector struct is no longer using it.

mattjala: Attempting to decrement the reference count below zero can happen within the library if a failure occurs at certain times - for example, in `H5ESinsert_request()`. As far as I can tell, most things that attempt to create a new object for the connector will have the potential to fail, so this is a case we should handle somehow.

qkoziol: Agree, but that may need to be handled at the source (in the code which increments and doesn't decrement).

qkoziol (changes requested):
Also probably not the best way to go.  How about extracting these two lines from H5VL_conn_dec_rc() into a new "destroy connector" routine:

        if (H5I_dec_ref(connector->id) < 0)
            HGOTO_ERROR(H5E_VOL, H5E_CANTDEC, FAIL, "unable to decrement ref c
        H5FL_FREE(H5VL_t, connector);

Then, call that routine from both H5VL_conn_dec_rc() and when there's an error and a connector from H5VL_new_connector() needs to be released.

src/H5VLint.c
```diff
@@ -1003,6 +1003,33 @@ H5VL_conn_dec_rc(H5VL_t *connector)
...
+{
+    herr_t ret_value = SUCCEED; /* Return value */
+
+    FUNC_ENTER_NOAPI(FAIL)
+
+    if (H5I_dec_ref(connector->id) < 0)
+        HGOTO_ERROR(H5E_VOL, H5E_CANTDEC, FAIL, "unable to decrement ref count on VOL connector");
+    H5FL_FREE(H5VL_t, connector);
+
+    connector = NULL;
```
mattjala: Setting the connector to NULL on free is new behavior, but it's necessary to avoid potential attempts to double free the connector.

Specifically within `H5VL_register_using_vol_id()` -> `H5VL_register()` -> `H5VL__new_vol_obj()`.  `H5VL__new_vol_obj()` is provided with a connector with zero references and attempts to create a new object. If it fails after incrementing the nrefs of the connector, it uses `H5VL_conn_dec_rc`, which frees the connector upon bringing it (back) to `nrefs = 0`. If the connector isn't set to NULL on free, then `H5VL_register_using_vol_id()` would attempt to double free it.

qkoziol: Setting connector to NULL has only a local affect and therefore is unnecessary here.

qkoziol (approved):
Yes, good solution.  Thanks for iterating on it.

src/H5VLint.c
mattjala: Good catch. This still needs to be resolved somehow, but it looks like it will be a little involved and can be done in a separate PR.

lrknox (approved):

src/H5VLint.c
lrknox: I'm not sure the change addressed this.

mattjala:
The previous solution left open the possibility for a double-free of a connector, since the routines that tried to create the first VOL object could fail, free the VOL connector, and be unable to propagate the fact that the connector was freed to their caller.

The current solution is to initialize `H5VL_t` with a 'dummy' reference count so it starts with nrefs = 1. After the first object is successfully created, the nrefs count will be manually decremented. This is not a perfect solution, but it seems cleaner than having subfunctions return pointers to the connector.

In every case, this should be safe:

- Failures prior to the creation of the first object will properly free it with a single decrement operation
- Failures between first object creation and the end of the function that created the connector will properly free it due to releasing the object (which decrements the reference count of the connector from 2 to 1) and then the function that created the connector will free it (1 -> 0)
- The dummy reference is removed at the end of the creation process, so failures that should result in the destruction of the connector can't occur after this point.

src/H5VLint.c
```diff
@@ -705,6 +705,10 @@ H5VL_register(H5I_type_t type, void *object, H5VL_t *vol_connector, bool app_ref
         HGOTO_ERROR(H5E_VOL, H5E_CANTREGISTER, H5I_INVALID_HID, "unable to register handle");
 
 done:
+    if (ret_value < 0)
+        if (vol_obj && H5VL_free_object(vol_obj) < 0)
+            HDONE_ERROR(H5E_VOL, H5E_CANTRELEASE, H5I_INVALID_HID, "unable to release VOL object");
```
mattjala, line 710: This is required to properly clean up the VOL object if a failure occurs during `H5I_register()` even without the changes to `H5VL_t`'s nrefs field.

qkoziol (changes requested):
This is a bad coding pattern.  A correctly created object should not require a decrement on the refcount.

qkoziol:
I can iterate on this with you to get to a better pattern.   What is the failure you are attempting to address?

mattjala:
> I can iterate on this with you to get to a better pattern. What is the failure you are attempting to address?

If a failure occurs after the creation of a VOL object but before the conclusion of the routine that invoked `H5VL_new_connector()`, and H5VL_t starts off with a reference count of zero, then the release of the VOL object during cleanup will indirectly free the connector before the connector is freed directly (in `H5ESinsert_request()` or `H5VL_register_using_vol_id()`), resulting in a double free. This wasn't addressed by the previous solution of extracting the connector's actual release into its own routine.

The callers need to directly free the connector on failure in at least some cases - since failure can occur before or after the VOL object is created. Since the routines that eventually try to create the VOL object can fail before or after creating it, just checking the failure status is not enough to determine how the connector needs to be treated.

One potential solution to this is to update the routines which might result in the provided connector being freed due to the first VOL object creation failing (`H5VL__new_vol_obj()`, and `H5VL_register()`) to take an `H5VL_t **` parameter which may be set to NULL if the connector is freed. This could then be checked to avoid double frees.

Another solution - the one currently implemented in this PR - is to use a 'dummy' ref count to avoid the possibility of a new connector being freed 'early' in subroutines.

hyoklee (approved):

qkoziol:
> I can iterate on this with you to get to a better pattern. What is the failure you are attempting to address?

@mattjala - I traced this around a little more today and have some ideas.   Would you like to set up a call to talk about it?

mattjala:
> @mattjala - I traced this around a little more today and have some ideas. Would you like to set up a call to talk about it?

Yes. We could pencil this in for discussion during tomorrow's working group, or set up a one on one discussion.

qkoziol:
> > @mattjala - I traced this around a little more today and have some ideas. Would you like to set up a call to talk about it?
> 
> Yes. We could pencil this in for discussion during tomorrow's working group, or set up a one on one discussion.

It's likely to be a fairly coding-intense discussion, so probably not a great fit for the community meeting.   I'll send you an email for setting up a 1:1 meeting and we can try to work it forward.

mattjala:
Quincey and I had a discussion about the best way to resolve this. We came to the conclusion that the root cause of this error is bad usage of VOL structures between different layers of the library, and that the solution is to:

1. Replace usage of the VOL class H5I id with the class itself (`H5VL_class_t`) in internal routines. 
2. Instead of registering the VOL class structure with H5I, register each VOL connector instance (`H5VL_t`) with H5I.
3. Connector creation should be pushed down to the same layer as object creation when possible, to remove the potential for free calls from two different layers during a failure, and in turn prevent the VOL connector from being passed around with a reference count of zero.

Quincey has started work on this set of changes, so I'm closing this PR for now.

Issue #4633, opened 2024-07-09
H5VL not freed properly during failure in H5O_refresh_metadata


mattjala:
**Describe the bug**
`H5O_refresh_metadata` increases the reference count of the VOL connector to prevent it from potentially being closed during the refresh operations  `H5O__refresh_metadata_close()` and `H5O_refresh_metadata_reopen()`. The reference count is decremented afterwards. However, if either of those two routines fail, then the reference count will be left erroneously high, likely preventing it from being freed properly later on.

**Expected behavior**
Connector should be freed properly regardless of failures.

Relevant section of code:

```
    /* Bump the number of references on the VOL connector.
     * If you don't do this, VDS refreshes can accidentally close the connector.
     */
    connector->nrefs++;

    /* Close object & evict its metadata */
    if (H5O__refresh_metadata_close(oloc, &obj_loc, oid) < 0)
        HGOTO_ERROR(H5E_OHDR, H5E_CANTLOAD, FAIL, "unable to refresh object");

    /* Re-open the object, re-fetching its metadata */
    if (H5O_refresh_metadata_reopen(oid, H5P_DEFAULT, &obj_loc, connector, false) < 0)
        HGOTO_ERROR(H5E_OHDR, H5E_CANTLOAD, FAIL, "unable to refresh object");

    /* Restore the number of references on the VOL connector */
    connector->nrefs--;

...

done:
    if (objs_incr && file)
        H5F_decr_nopen_objs(file);

    FUNC_LEAVE_NOAPI(ret_value)
```

PR #4636, opened 2024-07-09
Correct H5VL_t ref count on H5O_refresh_metadata failure

mattjala:
Fix for #4633 (H5VL not freed properly during failure in H5O_refresh_metadata)

This is the same pattern used to handle a similar failure case in `H5VL_new_connector()`.

### DIFF
--- a/src/H5Oflush.c
+++ b/src/H5Oflush.c
@@ -185,11 +185,10 @@ done:
 herr_t
 H5O_refresh_metadata(H5O_loc_t *oloc, hid_t oid)
 {
-    H5VL_object_t *vol_obj      = NULL;  /* VOL object associated with the ID */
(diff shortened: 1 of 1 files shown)


src/H5Oflush.c
```diff
@@ -256,6 +259,10 @@ H5O_refresh_metadata(H5O_loc_t *oloc, hid_t oid)
     if (objs_incr && file)
         H5F_decr_nopen_objs(file);
 
+    if (vol_obj && vol_obj->connector)
```
qkoziol: This seems overly complex.  Just decrement the connector's refcount before throwing an error, for the two places where it matters.

qkoziol (approved):

derobins (approved):

PR #4638, opened 2024-07-10
Fix bad ref counting in async object close with empty event set

mattjala:
If the event set is not none, the async object close functions increment the reference count of the VOL connector to prevent it from being closed if the object closure results in the entire file being closed. 

The end of the close routines always decrement the connector ref count, presumably in order to undo this earlier addition. (This decrement shouldn't correspond to the closing of the target object, since that is not guaranteed to occur if the target object has a reference count above 1.)

If these routines are invoked with an empty event set, then the earlier ref count increment never occurs and the decrement at the end of the function might result in the connector object being unexpectedly freed. 

This PR changes the decrement at the end to be controlled by the same condition that triggers the earlier increment (`H5ES_NONE != es_id`). `H5ES_insert()` shouldn't be able to take the event set from from a non-none value to a none value, so this should decrement the reference count in exactly the cases where it is incremented.

### DIFF
--- a/src/H5A.c
+++ b/src/H5A.c
@@ -2244,8 +2244,9 @@ H5Aclose_async(const char *app_file, const char *app_func, unsigned app_line, hi
             HGOTO_ERROR(H5E_ATTR, H5E_CANTINSERT, FAIL, "can't insert token into event set");
 
 done:
-    if (connector && H5VL_conn_dec_rc(connector) < 0)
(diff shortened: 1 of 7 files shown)


qkoziol (changes requested):
The previous code is already correct, since the 'connector' variable is only set to non-NULL when es_id != H5ES_NONE.

PR #4640, opened 2024-07-10
Update doxygen Learn Basics / example refs. Add Reference sections

byrnHDF:


mattjala (approved):

src/H5Dmodule.h
```diff
@@ -67,17 +67,21 @@
  * The dataset does not have to open to be linked or unlinked.
  *
  * \subsubsection subsubsec_dataset_intro_obj Object Reference
- * A dataset may be the target of an object reference. The object reference is created by
- * #H5Rcreate with the name of an object which may be a dataset and the reference type
+ * A file, group, dataset, named datatype, or attribute may be the target of an object reference.
+ * The object reference is created by
+ * #H5Rcreate_object with the name of an object which may be a dataset and the reference type
```
lrknox, line 72: May the object be any of the objects listed, or just a dataset?
I suppose the sentence beginning with "The object reference" is specific to a reference to a dataset, since this is in H5Dmodule.h.  Maybe it would be better if it began with "The dataset object reference"?
Is there (or will there be) documentation elsewhere for creating object references to file, group, named datatype, and attribute targets?

src/H5Tmodule.h
```diff
@@ -2943,8 +2943,63 @@ filled according to the value of this property. The padding can be:
  * \endcode
  *
  * \subsubsection subsubsec_datatype_other_refs Reference
+ * In HDF5, objects (groups, datasets, attributes, and committed datatypes) are usually accessed by name.
+ * There is another way to access stored objects - by reference. Before HDF5 1.12.0, were only two
```
lrknox: Before HDF5 1.12.0, **there** were only two

lrknox (approved):

src/H5Dmodule.h
byrnHDF: I believe it is dataset only

src/H5Tmodule.h
byrnHDF: fixed

src/H5Dmodule.h
lrknox: Then how about "The dataset object reference is created by" for line 71?

byrnHDF: Sorry, misunderstood the question - region refs are dataset only -  object refs are to objects.
And yes, there should be other examples for other types.

PR #4648, opened 2024-07-12
Eliminate more H5E_BEGIN/END_TRY macros and H5E_clear_stack() calls

qkoziol:


qkoziol:
Also cleaned up some of the error codes and a couple of warnings

mattjala (approved):

src/H5SM.c
```diff
@@ -288,12 +288,12 @@ H5SM__type_to_flag(unsigned type_id, unsigned *type_flag)
...
  */
-ssize_t
-H5SM__get_index(const H5SM_master_table_t *table, unsigned type_id)
+herr_t
+H5SM__get_index(const H5SM_master_table_t *table, unsigned type_id, ssize_t *idx)
 {
-    size_t   x;
     unsigned type_flag;
-    ssize_t  ret_value = FAIL;
+    ssize_t  indx      = -1;
```
derobins: I'd probably use found_idx instead of indx. One letter differences are easy to confuse.

qkoziol: Fine w/me  :-)

derobins (approved):

src/H5FDfamily.c
```diff
@@ -1082,7 +1082,7 @@ H5FD__family_get_eof(const H5FD_t *_file, H5FD_mem_t type)
      * loop with i==0.
      */
     assert(file->nmembs > 0);
-    for (i = (int)file->nmembs - 1; i >= 0; --i) {
+    for (i = (int)(file->nmembs - 1); i >= 0; --i) {
```
fortnern, line 1085: What's the reason for this change? It looks like it could create a warning about subtracting a signed number from an unsigned (I assume the (int) was added in the first place to avoid that).

src/H5Fsuper.c
```diff
@@ -361,19 +361,13 @@ H5F__super_read(H5F_t *f, H5P_genplist_t *fa_plist, bool initial_read)
...
         if (0 == mpi_rank) {
-            herr_t status;
-
             /* Try detecting file's signature */
             /* (Don't leave before Bcast, to avoid hang on error) */
-            H5E_BEGIN_TRY
+            H5E_PAUSE_ERRORS
             {
-                status = H5FD_locate_signature(file, &super_addr);
+                H5FD_locate_signature(file, &super_addr);
```
fortnern, line 368: This should be fine, but interesting note: I looked at H5FD_locate_signature to make sure it won't update sig_addr on failure, and it looks like it's actually meant to set it to HADDR_UNDEF on failure, but it does so *after* the call to HGOTO_ERROR...

src/H5FDfamily.c
qkoziol: It actually removed a warning.  :-)   I can't remember exactly, but one of the ones like "assuming that the range doesn't wrap around".

(It can't actually go negative, since the assert guarantees that nmembs is > 0)

src/H5Fsuper.c
qkoziol: 🙃

PR #4649, opened 2024-07-14
Clean up Fortran __float128 configure-time checks

derobins:
Cleans up the __float128 checks in both Autotools and CMake, which are only needed for the Fortran wrappers.

Fixes the CI failures that started about a week ago.


config/cmake/ConfigureChecks.cmake
```diff
@@ -646,22 +650,38 @@ endif()
...
+  #
+  # If __float128 exists and we can determine its precision, we will use
+  # it in the Fortran interface. The checks for this require that the
+  # precision be specified via a symbol named FLT128_DIG, which might be
+  # found in quadmath.h.
+  #
+  # The checks here are based on the GNU __float128 extension type from
+  # libquadmath, which is now part of gcc. Other compilers (clang, Intel)
+  # also expose __float128 and/or __float128 may be an alias for some
+  # other 128-bit floating point type.
```
lrknox, line 664: The sentence above and also in configure.ac might be worth clarifying;  I think I understand it's related to the first paragraph in the comment, but only after having read through all of the PR.

lrknox (approved):

config/cmake/ConfigureChecks.cmake
derobins: I'm going to go over this code with Scot when he gets back. We can revise then.

byrnHDF (approved):

PR #4654, opened 2024-07-16
Make VOL CI run against current branch

mattjala:
This PR also changes the VOL CI to be run against the current HDF branch instead of HDFGroup/develop so that other repos can detect if their changes break CI. This shouldn't change anything about the daily VOL connector CI, since that is run on develop anyway.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1202,12 +1202,8 @@ endif ()
 # Assuming they don't
 foreach (libs ${LINK_COMP_LIBS})
 #  set (_PKG_CONFIG_REQUIRES_PRIVATE "${_PKG_CONFIG_REQUIRES_PRIVATE} -l${libs}")
-if(TARGET ${libs})
   get_target_property (libname ${libs} OUTPUT_NAME)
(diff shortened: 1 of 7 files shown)


src/CMakeLists.txt
```diff
@@ -1202,8 +1202,12 @@ endif ()
 # Assuming they don't
 foreach (libs ${LINK_COMP_LIBS})
 #  set (_PKG_CONFIG_REQUIRES_PRIVATE "${_PKG_CONFIG_REQUIRES_PRIVATE} -l${libs}")
+if(TARGET ${libs})
```
byrnHDF: While this is a workaround - it is not the solution for the root problem.
The actual problem is that the CMake FindZLIB.cmake module does not set an OUTPUT_NAME on the target it creates, which is the second problem. The first problem is that the variable returned in ZLIB_LIBRARIES is not a target.

There will be a proper PR soon.

byrnHDF (changes requested):
remove the changes to src/CMakeLists.txt

byrnHDF (approved):

lrknox (approved):

PR #4655, opened 2024-07-16
Fixed usage issue with FindZLIB.cmake module

byrnHDF:


mattjala:
This doesn't fix the VOL CI. From a [run of VOL connectors CI against this change](https://github.com/mattjala/hdf5/actions/runs/9960706075/job/27520564302):

```
 /usr/bin/ld: /usr/lib/x86_64-linux-gnu/libz.a(deflate.o): warning: relocation against `z_errmsg' in read-only section `.text'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libz.a(zutil.o): relocation R_X86_64_PC32 against symbol `z_errmsg' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
gmake[2]: *** [src/CMakeFiles/hdf5-shared.dir/build.make:5668: bin/libhdf5.so.1000.0.0] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2705: src/CMakeFiles/hdf5-shared.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2
```

mattjala (changes requested):
Fix VOL CI

byrnHDF:
> Fix VOL CI

But that is a different problem as it complains about -fPic. Which means that the zlib it is finding was not built with -fPIC .
The other possibility is that it is using a newer CMake and needs to override the search for static compression libraries.

mattjala:
> But that is a different problem as it complains about -fPic. Which means that the zlib it is finding was not built with -fPIC . The other possibility is that it is using a newer CMake and needs to override the search for static compression libraries.

I'm not sure I understand exactly what has changed and why such that the VOL CI should change. I don't believe any of the failing VOL connectors retrieve Zlib in any special way. It looks like this set of changes removes the search for a dynamic zlib that would have been built with fPIC - is this somehow necessary in order to create the desired targets and define their output name property?

mattjala (approved):
The release note should probably mention the new HDF5_USE_ZLIB_STATIC and HDF5_USE_LIBAEC_STATIC options, but overall looks good

byrnHDF:
> The release note should probably mention the new HDF5_USE_ZLIB_STATIC and HDF5_USE_LIBAEC_STATIC options, but overall looks good

Actually only the ZLIB one is new but that reminds me that it needs to be added to the INSTALL_CMake.txt file.

lrknox (approved):

PR #4670, opened 2024-07-24
Update URL documentation links

byrnHDF:
Updated all links that need to originate from common URL.
Doxygen files use aliases, *.md files use footnotes


tools/test/h5dump/expected/pbits/tnofilename-with-packed-bits.ddl
```diff
@@ -105,7 +105,7 @@ usage: h5dump [OPTIONS] files
 
 --------------- Option Argument Conventions ---------------
   D - is the file driver to use in opening the file. Acceptable values are available from
-      https://portal.hdfgroup.org/documentation/hdf5-docs/registered_virtual_file_drivers_vfds.html.
+      https://support.hdfgroup.org/documentation/HDF5/registered_virtual_file_drivers_vfds.html.
```
derobins, line 108: We should consider managing VFD plugins like we do the filters (though this is a lower-order concern than filters).

derobins (approved):

PR #4675, opened 2024-07-29
Remove another H5E_BEGIN/END_TRY within the library

qkoziol:


src/H5Gdeprec.c
```diff
@@ -1133,7 +1127,7 @@ H5Gget_objname_by_idx(hid_t loc_id, hsize_t idx, char *name /*out*/, size_t size
     size_t               name_len = 0; /* Length of object name */
     ssize_t              ret_value;    /* Return value */
 
-    FUNC_ENTER_API(FAIL)
+    FUNC_ENTER_API(-1)
```
mattjala, line 1154: Why replace the use of `FAIL` here?

mattjala (approved):

src/H5Gdeprec.c
derobins: Because FAIL is semantically an herr_t value. -1 is better for an ssize_t, even though they are the same.

src/H5Gdeprec.c
```diff
@@ -1075,23 +1075,17 @@ H5G__get_objinfo(const H5G_loc_t *loc, const char *name, bool follow_link, H5G_s
         HGOTO_ERROR(H5E_SYM, H5E_EXISTS, FAIL, "name doesn't exist");
 
     /* If we're pointing at a soft or UD link, get the real link length and type */
-    if (statbuf && follow_link == 0) {
```
fortnern, line 1078: What happens if the user, for example, passes "/group/."?

qkoziol: That will be handled by the H5G__iterate() routine internally and won't fail, since there's an actual object that is found during path iteration.   This case is only for the literal "." string.

qkoziol: Hah!  I was wrong, good call @fortnern.  :-)

I've updated the code and added a test for this case.

lrknox (approved):

PR #4679, opened 2024-07-30
Use local variable in btree2 and print value

byrnHDF:


test/btree2.c
```diff
@@ -9917,19 +9917,21 @@ main(void)
...
     /* Reset library */
     h5_test_init();
-    fapl = h5_fileaccess();
+    fapl             = h5_fileaccess();
+    localTestExpress = TestExpress;
 
     /* For the Direct I/O driver, skip intensive tests due to poor performance */
-    if (!strcmp(driver_name, "direct"))
-        SetTestExpress(2);
+    if (localTestExpress < 2 && strcmp(driver_name, "direct"))
```
fortnern: As Larry mentioned in the meeting I think this should be !strcmp(...

test/btree2.c
```diff
@@ -9917,19 +9917,21 @@ main(void)
...
+        localTestExpress = 2;
 
-    if (TestExpress > 1)
-        printf("***Express test mode on.  Some tests may be skipped\n");
+    if (localTestExpress > 0)
```
fortnern, line 9933: Why the change from 1 to 0 here? All the other places in this PR use 1.

test/earray.c
```diff
@@ -2303,7 +2303,7 @@ main(void)
     h5_test_init();
     fapl = h5_fileaccess();
     if (TestExpress > 1)
-        printf("***Express test mode on.  Some tests may be skipped\n");
+        printf("***Express test mode %d.  Some tests may be skipped\n", TestExpress);
```
fortnern, line 2306: I prefer to explicitly specify the type of printf value parameters (int here) because there is absolutely no type checking, and a change in the global variable type could cause hard to track down errors. The LocalTestExpress ones aren't as bad since that variable is declared in the function but it wouldn't hurt to add it there too.

test/btree2.c
byrnHDF: Fixed

test/btree2.c
byrnHDF: Technically, level one could skip tests. The other uses actually skip tests if level 2 or 3.

test/earray.c
byrnHDF: Not sure what you want here, since the variable is just an int?

test/btree2.c
byrnHDF: > Technically, level one could skip tests. The other uses actually skip tests if level 2 or 3.

Fixed other files.

test/earray.c
qkoziol: Cast LocalTestExpress to an int:  "LocalTestExpress" => "(int)LocalTestExpress"

jhendersonHDF: Seems like a bad idea to cast something that doesn't really need casting since it'll just hide future possible errors.

fortnern: If the type of the variable ever changes it can produce memory errors. Modern compilers might be able to detect and throw a warning in this case (if the type doesn't match the format string), I'm not sure.

byrnHDF: I think you want the compiler to report the issue and not cover it up?

fortnern: I'm not certain the compiler will report an issue if there's no cast. And it is much more unsafe to pass the wrong type to printf than to cast something to the right type - the only potential problem there is wrap around. printf looks at the format string and reads that many bytes from the input parameters, it does  not check if enough data was passed in the parameters, so if the types don't match exactly it's possible for printf to read past the end of the input parameters which is a memory error.

fortnern: It looks like gcc does not warn by default but does if -Wformat is on

jhendersonHDF: I believe we have Wformat on by default for GCC and most of the other compilers. My main concern is that casting to int can hide other problems if the variable ever does change to something else, but it seems unlikely that it will at this point. Absolutely not an important issue though; either way is fine.

byrnHDF:
This is a test not the library - requiring this scrutiny is beyond the usual for a test! Other tests have more issues then if a simple value (0,1,2,3) would ever be more then an int?!

jhendersonHDF (approved):

fortnern (approved):

PR #4681, opened 2024-07-30
Rework Dynamic Analysis and sanitize testing

byrnHDF:
Sanitizers do not exhibit the same issues as valgrind did with the CMake test macros. Therefore allow sanitizers to use the same testing processes as regular testing. Also consideration should be given to removing the workarounds used for valgrind testing as sanitizers seem to do much better reporting of issues.


tools/test/h5ls/CMakeTests.cmake
```diff
@@ -320,22 +320,28 @@
     # rather than "IEEE 16-bit little-endian float".
     if (H5_WORDS_BIGENDIAN)
       ADD_H5_TEST (tfloat16_be 0 -w80 -v tfloat16_be.h5)
-      ADD_H5_TEST (tfloat16_be_nosupport 0 -w80 -v tfloat16_be.h5)
-      set_tests_properties (H5LS-tfloat16_be_nosupport PROPERTIES WILL_FAIL "true")
+      if (NOT HDF5_USING_ANALYSIS_TOOL)
```
jhendersonHDF: Why do we need to be checking whether an analysis tool is used when tests that are expected to fail are being tested? Surely there's a better way to do this.

byrnHDF: The trouble with these tests are that the sanitzer will avoid the runtest.cmake macro and therefore the test will not fail. I could execute the test and skip the property setting.

jhendersonHDF: What is it about the sanitizers that causes them to not run through the normal testing macros/framework? Do they do more than just add in the `-fsanitize=address` flag for example?

byrnHDF: I am investigating that now - it might just be valgrind that has the issue.

byrnHDF: Looks like the sanitizers do not have a problem with the macros - so will convert the checks to MEMCHECKER only.

jhendersonHDF: Checking for a specific analysis tool doesn't seem much better either. I don't necessarily want to suggest adding another test macro for this purpose, but it's not great that anybody adding new tests needs to know that tests which are expected to fail have to be wrapped in `if (NOT HDF5_USING_ANALYSIS_TOOL)` or similar. It should be simpler than that.

byrnHDF: Reworked this whole analysis as sanitisers do not have the problems that valgrind has.
In fact, I think the sanitizers negate the need to support valgrind limitations and even the MEMCHECKER checks should just be eliminated.

byrnHDF:
The sanitizers do not need any special handling for the CMake macros like valgrind. The variety of sanitizers make the issues with valgrind moot, as the sanitizers provide more information during a test run on failures.
I will create another PR after this one to remove the special checks for memchecker and CMake macros.

lrknox (approved):

jhendersonHDF (approved):

PR #4683, opened 2024-07-31
Fix segfault when closing datatype during failure in H5Topen2

jhendersonHDF:
When opening a committed datatype fails (usually due to an issue during datatype decoding), `H5VL_datatype_close` would be called on an incorrect VOL object pointer and result in a segfault or heap buffer overflow. This has been fixed by creating a temporary VOL object that wraps the datatype object pointer handed back by the stack of VOL connectors and using that for the close operation.

### DIFF
--- a/src/H5Tcommit.c
+++ b/src/H5Tcommit.c
@@ -59,6 +59,7 @@ static herr_t H5T__commit_api_common(hid_t loc_id, const char *name, hid_t type_
 static hid_t  H5T__open_api_common(hid_t loc_id, const char *name, hid_t tapl_id, void **token_ptr,
                                    H5VL_object_t **_vol_obj_ptr);
 static H5T_t *H5T__open_oid(const H5G_loc_t *loc);
+static herr_t H5T_destruct_datatype(void *datatype, H5VL_t *vol_connector);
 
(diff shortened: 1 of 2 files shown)


src/H5Tcommit.c
```diff
@@ -662,7 +663,7 @@ H5T__open_api_common(hid_t loc_id, const char *name, hid_t tapl_id, void **token
 done:
     /* Cleanup on error */
     if (H5I_INVALID_HID == ret_value)
-        if (dt && H5VL_datatype_close(*vol_obj_ptr, H5P_DATASET_XFER_DEFAULT, H5_REQUEST_NULL) < 0)
```
jhendersonHDF, line 665: This VOL object pointer actually points to the VOL object that was passed in to `H5Topen2` for `loc_id` rather than pointing to the VOL object for the committed datatype. When `H5VL_register` above fails, `dt` is left as a typeless pointer not attached to a VOL object, so we have to wrap it in one before being able to call `H5VL_datatype_close` on it.

qkoziol (approved):

mattjala (approved):

lrknox (approved):

PR #4685, opened 2024-08-01
CMake link line needs to use new HDF5_ENABLE_THREADS

byrnHDF:


jhendersonHDF (approved):

lrknox:
Fixes link error, but ping pong test in test/ttsafe_semaphore.c sometimes passes, sometimes fails:

 UNEXPECTED VALUE from ping pong should be 1000001, but is 1000000 at line  111 in /home/hdftest/snapshots-hdf5/current/test/ttsafe_semaphore.c

Also failed similarly with MPICH 3.3 on jelly and maybe some other machines before this PR, after yesterday's commit.

lrknox:
@qkoziol, this PR eliminates link errors for thread functions when building with CMake on various systems today.  We also see the ping pong error above, not related to the link error.  It occurred for one of the github test configurations when this PR was created, but the configuration passed when it was re-run.  Is the lightweight system semaphores (semaphore) subtest always expected to have value 1000001, or will it sometimes correctly return value 1000000?

qkoziol:
Thank you for fixing this.

I'll check into the intermittent ping pong failures for the semaphores ASAP tomorrow.  (It should always be 1000001)

qkoziol (approved):

PR #4701, opened 2024-08-07
Add Windows SHLWAPI lib to public interface

### LINKED ISSUE #3663: The cmake configuration should provide transitive dependencies (like `shlwapi` on MSVC)
cmake is very good in collecting the list of library dependencies transitively for project trees. This typically works via the list of linker libraries. In order to make use of the automatic dependency tracking feature, dependencies that need to be linked downstream must be listed `PUBLIC` in the project. Every such `PUBLIC` dependency will then be linked into the final executable, and in turn, its dependencies are also (transitively) resolved and linked.

Since https://github.com/HDFGroup/hdf5/pull/2407, hdf5 has a dependency on `shlwapi` on MSVC. However this seems not to be exposed to cmake. At least, when I use the following code in cmake, the build lacks a linker dependency on `shlwapi`. I could not find out how to instruct cmake with hdf5 to get this dependency resolved. I can resolve it by linking `shlwapi` myself. But this is not ideal, because I need to maintain the dependency list of hdf5 in my own (user) code. This is unfortunate, because it creates an extra burden on users to keep dependency lists updated, which can be quite tedious for large software projects.

**Expected behavior**
The following used to work until about a year ago. But it is no longer sufficient for me since the current release. This, or something similar, should just work:
```
  find_package(HDF5 MODULE REQUIRED COMPONENTS C)

  target_include_directories(mylibrary SYSTEM INTERFACE ${HDF5_INCLUDE_DIRS})
  target_link_libraries(mylibrary INTERFACE ${HDF5_LIBRARIES}) 
  target_compile_definitions(mylibrary INTERFACE ${HDF5_DEFINITIONS})
```

 - HDF5 version: 1.14.2
 - OS and version: Windows 11
 - Compiler and version: Visual Studio 17.7.5 with LLVM clang-cl 16.0.6
 - Build system (e.g. CMake, Autotools) and version: cmake 3.27.7
 - Any configure options you specified:
```
    "-DBUILD_SHARED_LIBS=OFF"
    "-DBUILD_STATIC_LIBS=ON"
    "-DONLY_SHARED_LIBS=OFF"
```
 - MPI library and version (parallel HDF5): No MPI

**Additional context**
None.

byrnHDF:
This addresses the missing dependency for Windows of the shlwapi library, this will close #3663


jhendersonHDF:
I'd definitely still like to know if we're doing something wrong since I don't understand the reason why this needs to be in the public link interface. I'm guessing that it has to do with functions which are meant to be library-private not having restricted visibility so they're in the link interface for anything which links to HDF5.

lrknox:
> This addresses the missing dependency for Windows of the shlwapi library, this will close #3663

I believe it will close #3663 automatically if the description includes "Fixes #3663".  Otherwise it may need to be closed manually.

byrnHDF:
> I'd definitely still like to know if we're doing something wrong since I don't understand the reason why this needs to be in the public link interface. I'm guessing that it has to do with functions which are meant to be library-private not having restricted visibility so they're in the link interface for anything which links to HDF5.

see https://stackoverflow.com/questions/26037954/cmake-target-link-libraries-interface-dependencies

lrknox (approved):

jhendersonHDF:
> > I'd definitely still like to know if we're doing something wrong since I don't understand the reason why this needs to be in the public link interface. I'm guessing that it has to do with functions which are meant to be library-private not having restricted visibility so they're in the link interface for anything which links to HDF5.
> 
> see https://stackoverflow.com/questions/26037954/cmake-target-link-libraries-interface-dependencies

That just seems to reinforce the idea that this shouldn't be a public link because there's no reason the functions that use the SHLWAPI lib should be in the public link interface. The library was linked against due to use of StrStrIA, but this is used in a function that isn't meant to be used publicly. It's in H5private.h through H5system.c and is meant for internal library use. But since we don't restrict the visibility of library-wide private functions, I'm guessing it ends up in the link interface unintentionally.

brtnfld:
If we want a definitive answer, we should ask on CMake's discourse https://discourse.cmake.org/

byrnHDF:
I could make it an INTERFACE instead of PUBLIC.

jhendersonHDF:
> I could make it an INTERFACE instead of PUBLIC.

We need to link against the shlwapi lib anyway, so I don't think INTERFACE is correct since "Libraries following INTERFACE are appended to the link interface and are not used for linking \<target\>."

jhendersonHDF:
This will likely have to remain a PUBLIC link for now since it would take some time to figure out why shlwapi is a transitive dependency currently and some work to resolve. However, I'm not convinced that PUBLIC is ultimately correct and can't see how shlwapi is anything but a private dependency.

jhendersonHDF:
> > This addresses the missing dependency for Windows of the shlwapi library, this will close #3663
> 
> I believe it will close #3663 automatically if the description includes "Fixes #3663". Otherwise it may need to be closed manually.

You can see the GitHub issue linked on the right side under the "Development" section. Here's the list of GitHub keywords in a description that will automatically link an issue: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

byrnHDF:
INTERFACE didn't work - reverting back to PUBLIC

src/CMakeLists.txt
```diff
@@ -1099,7 +1099,7 @@ if (BUILD_STATIC_LIBS)
   TARGET_C_PROPERTIES (${HDF5_LIB_TARGET} STATIC)
   target_link_libraries (${HDF5_LIB_TARGET}
       PRIVATE ${LINK_LIBS} ${LINK_COMP_LIBS}
-      PUBLIC "$<$<NOT:$<PLATFORM_ID:Windows>>:${CMAKE_DL_LIBS}>" "$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>"
+      PUBLIC "$<$<PLATFORM_ID:Windows>:${SHLWAPI_LIB}>" "$<$<NOT:$<PLATFORM_ID:Windows>>:${CMAKE_DL_LIBS}>" "$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:MPI::MPI_C>"
```
jhendersonHDF: Can we have a `LINK_LIBS`-style variable for public libs, just so we don't have to introduce unnecessary extra variables like `SHLWAPI_LIB` above?

byrnHDF: Sounds like a good idea

jhendersonHDF (approved):
Approving, though I hope two things come after at some point in the future:

- We should figure out why this dependency had to be made public since the library is really only used by library-internal functions and not exposed publicly in any way (at least not intentionally)
- Would be nice to have a public `LINK_LIBS` style variable for public libraries, but that should probably come in a separate PR

byrnHDF:
> * Would be nice to have a public `LINK_LIBS` style variable for public libraries, but that should probably come in a separate PR
I'll do it in this PR.
I'll leave it dependent on Windows since it seems to be the one platform that is affected.

scivision:
HDF5 v1.14.5, 1.14.6, 2.0.0 on Windows with Intel oneAPI 2024.2 or 2025.3 still have the error in #3663 

>  error LNK2019: unresolved external symbol __imp_StrStrIA referenced in function Wstrcasestr_wrap

etc. are still preset.

scivision:
still an issue in HDF5 2.1

Issue #1498, opened 2022-03-13
Segmentation fault when opening unseekable file


kalvdans:
I tried to use `/dev/null` in a unit test (since I'm not interested in the result) but the library crashes hard. This is a minimal test case:
```
#include <H5Fpublic.h>
#include <stdio.h>

int main(int argc, char *argv[])
{
    hid_t fileobject = H5Fcreate("/dev/null", H5F_ACC_TRUNC, 0, 0);
    return H5Fclose(fileobject);
}
```
The output when running is:
```
$ gcc -Wall h5py-bug-1775.c -I /usr/include/hdf5/serial -l hdf5_serial && ./a.out 
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 139699578592128:
  #000: ../../../src/H5F.c line 668 in H5Fclose(): closing file ID failed
    major: File accessibilty
    minor: Unable to close file
  #001: ../../../src/H5Fint.c line 2047 in H5F__close(): decrementing file ID failed
    major: File accessibilty
    minor: Unable to decrement reference count
  #002: ../../../src/H5I.c line 1309 in H5I_dec_app_ref(): can't decrement ID ref count
    major: Object atom
    minor: Unable to decrement reference count
  #003: ../../../src/H5Fint.c line 2105 in H5F__close_cb(): can't close file
    major: File accessibilty
    minor: Unable to close file
  #004: ../../../src/H5Fint.c line 2276 in H5F_try_close(): problems closing file
    major: File accessibilty
    minor: Unable to close file
  #005: ../../../src/H5Fint.c line 1130 in H5F__dest(): unable to flush cached data (phase 2)
    major: File accessibilty
    minor: Unable to flush data from cache
  #006: ../../../src/H5Fint.c line 1896 in H5F__flush_phase2(): low level truncate failed
    major: File accessibilty
    minor: Write failed
  #007: ../../../src/H5FD.c line 1581 in H5FD_truncate(): driver truncate request failed
    major: Virtual File Layer
    minor: Can't update object
  #008: ../../../src/H5FDsec2.c line 892 in H5FD_sec2_truncate(): unable to extend file properly, errno = 22, error message = 'Invalid argument'
    major: Low-level I/O
    minor: Seek failed
Segmentation fault
```

I understand that hdf5 library wants to seek in the file, but it should fail gracefully. My version:
```
$ dpkg -l | grep libhdf5-dev
[...]

PR #4720, opened 2024-08-16
Test opening unseekable file

mattjala:
Test for the issue described in #1498. 

This works in 1.12/1.14 and segfaults in 1.10, so this is a regression test.

### DIFF
--- a/test/tfile.c
+++ b/test/tfile.c
@@ -8102,6 +8102,49 @@ test_min_dset_ohdr(void)
 
 /****************************************************************
 **
+**  test_unseekable_file():
+**    Test that attempting to open an unseekable file fails gracefully
+**    without a segfault (see hdf5#1498)
+****************************************************************/
+static void
+test_unseekable_file(void)
+{
+    hid_t file_id = H5I_INVALID_HID; /* File ID */
+
+    /* Output message about test being performed */
+    MESSAGE(5, ("Testing creating/opening an unseekable file\n"));
+
+    /* Creation */
+#ifdef H5_HAVE_WIN32_API
+    file_id = H5Fcreate("NUL", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+#else
+    file_id = H5Fcreate("/dev/null", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+#endif
+
+    H5Fclose(file_id);
+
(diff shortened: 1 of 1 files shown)


lrknox:
Rerunning the check where ttsafe failed - it's a know issue and often passes when rerun.  There's also this error in some other checks:

/home/runner/work/hdf5/hdf5/test/tfile.c: In function ‘test_unseekable_file’:
/home/runner/work/hdf5/hdf5/test/tfile.c:8146:1: error: label ‘error’ defined but not used [-Werror=unused-label]
 8146 | error:

Ypu'll probably need to fix that one.

lrknox (approved):

qkoziol (approved):

test/tfile.c
```diff
@@ -8419,10 +8462,11 @@ test_file(void)
...
     test_libver_bounds_low_high(driver_name);
-    test_libver_macros();  /* Test the macros for library version comparison */
-    test_libver_macros2(); /* Test the macros for library version comparison */
-    test_incr_filesize();  /* Test H5Fincrement_filesize() and H5Fget_eoa() */
-    test_min_dset_ohdr();  /* Test dataset object header minimization */
+    test_libver_macros();   /* Test the macros for library version comparison */
+    test_libver_macros2();  /* Test the macros for library version comparison */
+    test_incr_filesize();   /* Test H5Fincrement_filesize() and H5Fget_eoa() */
+    test_min_dset_ohdr();   /* Test dataset object header minimization */
+    test_unseekable_file(); /* Test attempting to open/create an unseekable file */
```
kalvdans, line 8469: No need to keep the comments aligned, better keep diff size down.

test/tfile.c
```diff
@@ -8100,6 +8100,49 @@ test_min_dset_ohdr(void)
...
+**    Test that attempting to open an unseekable file fails gracefully
+**    without a segfault (see hdf5#1498)
+****************************************************************/
+static void
+test_unseekable_file(void)
+{
+    hid_t file_id = H5I_INVALID_HID; /* File ID */
+
+    /* Output message about test being performed */
+    MESSAGE(5, ("Testing creating/opening an unseekable file\n"));
```
kalvdans, line 8115: Maybe fflush(stdout) after this, to know which test that causes a Segmentation Fault.

kalvdans (approved):
Looks good besides two minor remarks.

PR #4725, opened 2024-08-17
Fix semaphore test

qkoziol:
Don't access counter variable outside of semaphore to avoid race


qkoziol:
@derobins @jhendersonHDF @fortnern - Could you review this at your earliest convenience?  I believe it will fix the intermittent failures in the semaphore regression test on develop.

byrnHDF (approved):

lrknox (approved):

PR #4736, opened 2024-08-20
Fix segfault in ROS3 credential parsing

mattjala:
The whitespace trimming could sometimes run off the end of the buffer and segfault. 

Fixes an issue where, if two profiles were provided in the same file, the information provided in the second would overwrite the first in unpredictable ways. Now, parsing stops after the first profile. (This is the result of resetting the `found_setting` flag.)

### DIFF
--- a/src/H5FDs3comms.c
+++ b/src/H5FDs3comms.c
@@ -1671,6 +1671,9 @@ H5FD_s3comms_HMAC_SHA256(const unsigned char *key, size_t key_len, const char *m
 
     FUNC_ENTER_NOAPI_NOINIT
 
+    if (!key)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "signing key not provided");
+
     if (dest == NULL)
         HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "destination cannot be null.");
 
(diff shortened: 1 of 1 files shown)


hyoklee (approved):

qkoziol (approved):

src/H5FDs3comms.c
```diff
@@ -1812,10 +1815,11 @@ H5FD__s3comms_load_aws_creds_from_file(FILE *file, const char *profile_name, cha
                 strncpy(setting_pointers[setting_i], (const char *)line_buffer, strlen(line_buffer));
 
                 /* "trim" tailing whitespace by replacing with null terminator*/
-                buffer_i = 0;
-                while (!isspace(setting_pointers[setting_i][buffer_i]))
-                    buffer_i++;
-                setting_pointers[setting_i][buffer_i] = '\0';
+                end = strlen(line_buffer) - 1;
```
jhendersonHDF, line 1818: Be careful that the length of `line_buffer` isn't 0 here, which would cause `end` to wrap around.

src/H5FDs3comms.c
```diff
@@ -1812,10 +1815,11 @@ H5FD__s3comms_load_aws_creds_from_file(FILE *file, const char *profile_name, cha
...
+                while (end > 0 && isspace((unsigned char)setting_pointers[setting_i][end])) {
```
jhendersonHDF: Why the cast to `unsigned char` here when `isspace()` expects an int?

src/H5FDs3comms.c
mattjala: This should be covered by the earlier check on line 1810 that throws an errors if ` *(line_buffer + 1) == 0`.

PR #4579, opened 2024-06-18
Clean up ros3 VFD stats code

derobins:
* Removes printf debugging
* Simplifies and centralizes stats code
* Other misc tidying

PR #4738, opened 2024-08-20
Return basic HTTP range GET logging to ROS3

ajelenak:
These logging statements disappeared after the recent ROS3 code cleanup.

### DIFF
--- a/src/H5FDs3comms.c
+++ b/src/H5FDs3comms.c
@@ -50,7 +50,7 @@
 /* manipulate verbosity of CURL output
  *
  * 0 -> no explicit curl output
- * 1 -> on error, print failure info to stderr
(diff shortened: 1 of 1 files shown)


mattjala:
This was something specifically removed in [this commit](https://github.com/HDFGroup/hdf5/pull/4579/commits/3eb9bc1c1dc6d59fd34b8883782c6f9963de191c) - perhaps we need to have a discussion about the best way to handle logging here?

ajelenak:
I did have a conversation with that commit's author and we concluded my changes were worth putting back in.

I'm fine with a discussion on ROS3 logging only if it will also not be a build configuration, for example, something that can be triggered by an env. variable. Otherwise, I am fine with as-is.

byrnHDF (approved):

derobins (approved):

PR #4747, opened 2024-08-22
Check if DECIMAL_DIG is defined during CMake configuration

hyoklee:
This PR will fix the following error on Echidna:

```
 CMake Error at config/cmake/ConfigureChecks.cmake:729 (message):
  Compilation of C maximum decimal precision for C - Failed
  config/cmake/ConfigureChecks.cmake:754 (C_RUN)
  CMakeLists.txt:496 (include)
```

### DIFF
--- a/config/cmake/ConfigureChecks.cmake
+++ b/config/cmake/ConfigureChecks.cmake
@@ -745,7 +745,11 @@ if (HDF5_BUILD_FORTRAN)
 #else\n\
 #  define C_FLT128_DIG 0\n\
 #endif\n\
-#define C_LDBL_DIG DECIMAL_DIG\n\
+#ifdef DECIMAL_DIG\n\
+#  define C_LDBL_DIG DECIMAL_DIG\n\
+#else\n\
+#  define C_LDBL_DIG 0\n\
+#endif\n\
 \n\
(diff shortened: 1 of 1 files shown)


config/cmake/ConfigureChecks.cmake
```diff
@@ -745,7 +745,11 @@ if (HDF5_BUILD_FORTRAN)
 #else\n\
 #  define C_FLT128_DIG 0\n\
 #endif\n\
-#define C_LDBL_DIG DECIMAL_DIG\n\
+#ifdef DECIMAL_DIG\n\
```
jhendersonHDF, line 748: What compiler is this with? `DECIMAL_DIG` is C99, so I think we should be failing configure if C99 isn't supported rather than defining `C_LDBL_DIG` to 0 and hoping it will work.

hyoklee: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)

brtnfld: If you use -std=c99, it will be defined. HDF5 requires a minimum of C99. This option should be added for these older GNU compilers.

brtnfld (changes requested):
There are better ways to fix this than setting it to zero.

PR #4749, opened 2024-08-23
Restore rand_r in a few parallel tests

derobins:
The t_pmulti_dset and t_select_io_dset tests rely on the behavior of the previous private rand_r-like implementation to get the correct sequence of random numbers to pass. This has been restored using a fully private rand_r-like implementation that doesn't rely on rand_r and will work on Windows and other platforms where rand_r doesn't exist.

### DIFF
--- a/test/h5test.c
+++ b/test/h5test.c
@@ -2563,3 +2563,24 @@ h5_driver_uses_multiple_files(const char *drv_name, unsigned flags)
 
     return ret_val;
 }
+
+/* Deterministic random number functions that don't modify the underlying
+ * C/POSIX library rand/random state, as this can cause spurious test failures.
+ *
+ * Adapted from the example code in the POSIX.1-2001 standard.
+ */
+
+static unsigned int next_g = 1;
+
+int
+h5_local_rand(void)
+{
+    next_g = next_g * 1103515245 + 12345;
+    return next_g & RAND_MAX;
+}
+
+void
+h5_local_srand(unsigned int seed)
+{
+    next_g = seed;
+}
--- a/testpar/t_select_io_dset.c
+++ b/testpar/t_select_io_dset.c
@@ -1588,7 +1588,7 @@ test_multi_dsets_no_bkg(hid_t fid, unsigned chunked, unsigned dtrans, unsigned s
                  mwbuf ? "mwbuf" : "nomwbuf");
 
         /* Flip a coin to see if we're doing type conversion */
-        tconv = rand() % 2;
+        tconv = h5_local_rand() % 2;
         if (tconv)
             any_tconv = true;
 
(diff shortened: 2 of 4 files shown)


test/h5test.c
```diff
@@ -2563,3 +2563,24 @@ h5_driver_uses_multiple_files(const char *drv_name, unsigned flags)
...
+ */
+
+static unsigned int next_g = 1;
+
+int
+h5_local_rand(void)
+{
+    next_g = next_g * 1103515245 + 12345;
+    return next_g & RAND_MAX;
+}
```
derobins, line 2580: This seems to be the simplest consensus algorithm that uses RAND_MAX (the POSIX example assumes RAND_MAX is 32767).

derobins:
The two tests that are "failing":

* An h5diff test is barfing on mkdir (a not uncommon error, unrelated to this change)
* The EOS code is hidden behind a login now. We'll have to modify that action.

testpar/t_select_io_dset.c
```diff
@@ -1588,7 +1588,7 @@ test_multi_dsets_no_bkg(hid_t fid, unsigned chunked, unsigned dtrans, unsigned s
                  mwbuf ? "mwbuf" : "nomwbuf");
 
         /* Flip a coin to see if we're doing type conversion */
-        tconv = rand() % 2;
```
fortnern, line 1591: It's surprising to me that this test doesn't call srand (neither does the serial version). We should probably change that, though it doesn't necessarily need to be in this PR.

fortnern (approved):

testpar/t_select_io_dset.c
derobins: I also thought that was weird, but the docs say not calling srand is like using 1 as the seed, IIRC, and since we just used 42 as the seed in the original parallel test I wasn't going to worry too much about it.

qkoziol (approved):
A good fix, but still dependent on the parallel testing algorithms always calling h5_local_rand() in a "globally collective" way (i.e. always with all the nodes and always in the same order).   Probably fine for now, but could be a source of confusion in the future.

lrknox:
t_pmulti_dsets and t_sect_io_dset tessts pass now, at least on perlmutter (Cray at NERSC) and skybridge (Linux 3.10 at Sandia).

jhendersonHDF:
> A good fix, but still dependent on the parallel testing algorithms always calling h5_local_rand() in a "globally collective" way (i.e. always with all the nodes and always in the same order). Probably fine for now, but could be a source of confusion in the future.

Or broadcasting the value from a single rank, however even this didn't seem to work on the HPC machines that were failing.

jhendersonHDF (approved):

PR #4752, opened 2024-08-23
Don't run AOCC parallel tests with -j2

derobins:
Don't run parallel tests in both Autotools and CMake with multiple processes. ph5diff still runs with -j2 w/ Autotools since the test script is in the tools/test/h5diff directory.

### DIFF
--- a/.github/workflows/aocc-auto.yml
+++ b/.github/workflows/aocc-auto.yml
@@ -86,12 +86,24 @@ jobs:
           make -j3
         working-directory: ${{ runner.workspace }}/build
 
+      # ph5diff tests are in the tools/tests directory so they will get run
+      # here
       - name: Autotools Run Tests
         env:
           NPROCS: 2
         run: |
           export PATH=/home/runner/work/hdf5/hdf5/openmpi-4.1.6-install/bin:/usr/local/bin:$PATH
-          make check -j
+          cd test && make check -j2 && cd ..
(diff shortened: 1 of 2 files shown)


.github/workflows/aocc-auto.yml
```diff
@@ -86,12 +86,24 @@ jobs:
           make -j3
         working-directory: ${{ runner.workspace }}/build
 
+      # ph5diff tests are in the tools/tests directory so they will get run
+      # here
       - name: Autotools Run Tests
         env:
           NPROCS: 2
```
jhendersonHDF, line 91: Should NPROCS be unset here?

derobins: It could be. It's not necessary.

derobins: Done. We only use NPROCS in the parallel tests and configure (just in case).

derobins: Wait, no - I'm putting it back in the serial tests since we run the ph5diff tests there.

jhendersonHDF: That seems worth reworking since they aren't serial tests

derobins: Agree, but I'll save that for a different PR.

jhendersonHDF (approved):

lrknox (approved):

PR #4757, opened 2024-08-26
Remove early test exit

mattjala:
The unseekable file test `exit`s on success and skips any tests that would have been executed afterwards. This removes the `exit`, which shouldn't have been there in the first place.

### DIFF
--- a/test/tfile.c
+++ b/test/tfile.c
@@ -8115,6 +8115,9 @@ test_unseekable_file(void)
     /* Output message about test being performed */
     MESSAGE(5, ("Testing creating/opening an unseekable file\n"));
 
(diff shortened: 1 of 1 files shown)


jhendersonHDF (approved):

mattjala:
I've removed the checks that the file id returned is H5I_INVALID HID, because it seems that on some systems the calls succeed and return a valid ID. ( [CI results](https://github.com/HDFGroup/hdf5/actions/runs/10562340838)) This is still bad, but the main point of this test is just to make sure a segfault doesn't occur; the lack of proper failure can be investigated separately.

byrnHDF (approved):

Issue #4539, opened 2024-06-02
Fix broken Julia CI


derobins:
The Julia GitHub CI actions have been broken for the past week or two, both in Autotools and CMake. There were no obvious changes that could have caused these failures. They will often pass when re-run.

We will need to investigate why they are failing. Since it's random, it may be a memory issue, either in the Julia wrappers or the HDF5 library.

derobins:
Sample test failure output:

```
Test Summary:                      | Pass  Fail  Broken  Total
HDF5.jl                            | 1497     2       3   1502
  plain                            |  151             1    152
  complex                          |   13                   13
  undefined and null               |    4                    4
  abstract arrays                  |    2                    2
  empty and 0-size arrays          |   39                   39
  generic read of native types     |   17                   17
  show                             |   44                   44
  split1                           |   13                   13
  haskey                           |   18                   18
  AbstractString                   |   51                   51
  opaque data                      |    7                    7
  FixedStrings and FixedArrays     |   18                   18
  Object Exists                    |    8                    8
  HDF5 existance                   |    4                    4
  bounds                           |    2                    2
  create_dataset                   |  264                  264
  Strings                          |    8                    8
  h5a_iterate                      |    7     1              8
  h5l_iterate                      |    7     1              8
  h5dchunk_iter                    |    3                    3
  compound                         |   10                   10
  create_dataset (compound)        |    4                    4
  write_compound                   |   27                   27
  custom                           |    6                    6
  reference                        |    6                    6
  null dataspace                   |   13                   13
  scalar dataspace                 |   15                   15
  simple dataspaces                |   98                   98
  BlockRange                       |   42                   42
  hyperslab                        |    6                    6
  Datatypes                        |   15                   15
  hyperslab                        |    5                    5
  read 0-length arrays: issue #859 |                     No tests
  attrs interface                  |   92                   92
  variable length strings          |    1                    1
  readremote                       |   23                   23
  extend                           |   29                   29
  gc                               |  101                  101
  external                         |    6                    6
  swmr                             |    4                    4
  mmap                             |    9                    9
  properties                       |   46             1     47
  filter                           |   80                   80
  Raw Chunk I/O                    |   80                   80
  fileio                           |    6                    6
  track order                      |   18                   18
  h5f_get_dset_no_attrs_hint       |    6                    6
  non-allocating methods           |   11             1     12
  Compression Filter Unit Tests    |    6                    6
  Object API                       |   38                   38
[...]

PR #4763, opened 2024-08-26
Restore Julia CI

derobins:


.github/workflows/julia-auto.yml
```diff
@@ -0,0 +1,79 @@
+name: hdf5 1.14 autotools julia
```
byrnHDF: 1.14 to dev

.github/workflows/julia-cmake.yml
```diff
@@ -0,0 +1,82 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Sources
+        uses: actions/checkout@v4.1.1
+
+      - name: Install Dependencies
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install ninja-build doxygen graphviz
```
byrnHDF, line 26: Might want to use the doxygen action to specify version 1.9.6

byrnHDF: - name: Install Dependencies
        uses: ssciwr/doxygen-install@v1
        with:
          version: "1.9.7"

.github/workflows/julia-auto.yml
derobins: Fixed

hyoklee (approved):

byrnHDF (approved):

derobins:
Fixes #4539

PR #4774, opened 2024-08-27
Sync develop changes August 21 - 27 to hdf5_1_14

lrknox:


.github/workflows/julia-auto.yml
```diff
@@ -0,0 +1,79 @@
+name: hdf5 dev autotools julia
```
byrnHDF: dev to 1.14

.github/workflows/julia-cmake.yml
```diff
@@ -0,0 +1,82 @@
+name: hdf5 dev CMake julia
```
byrnHDF: dev to 1.14

byrnHDF (approved):

.github/workflows/julia-auto.yml
lrknox:
```suggestion
name: hdf5 1.14 autotools julia
```

derobins (approved):

PR #4781, opened 2024-08-28
Update URL documentation links to support site

byrnHDF:
Updated all links that need to originate from common support URL.
Doxygen files use aliases, *.md files use footnotes


derobins (approved):

HDF5Examples/README.md
```diff
@@ -48,17 +48,17 @@ HDF5 SNAPSHOTS, PREVIOUS RELEASES AND SOURCE CODE
...
 Full Documentation and Programming Resources for this HDF5 can be found at
 
-   https://portal.hdfgroup.org/documentation/index.html
+   https://support.hdfgroup.org/documentation/HDF5/index.html
 
 Periodically development code snapshots are provided at the following URL:
-    
-   https://gamma.hdfgroup.org/ftp/pub/outgoing/hdf5/snapshots/
+
+   https://github.com/HDFGroup/hdf5/releases
```
hyoklee, line 55: https://github.com/HDFGroup/hdf5/releases/tag/snapshot

byrnHDF: releases is the more general location - snapshot is just develop, snapshot-1.14 is for the releasable branch

lrknox: The snapshot URL is probably better as it is more closely related to the preceding sentence "Periodically development code snapshots are provided at the following URL:", and there are no develop code "releases".  However, the last snapshot listed for develop is dated May 25.  Should there be monthly/bi-weekly/weekly develop and 1.14 snapshots deployed, maybe with a reduced number of assets?

byrnHDF: As examples can be used with any HDF5 binary - I think it it is better to point to more general location.
Merges always trigger builds - snapshot named binaries need to be forced. Not sure what the policy should be on forcing a snapshot build, maybe a week after a merge build?

lrknox (approved):

Issue #4586, opened 2024-06-19
A memory leak was found  while testing h5_extended_fuzzer with libfuzzer


MageWeiG:
While testing h5_extended_fuzzer with libfuzzer, I found a memory leak 

The reason for this vulnerability is:
There is a direct leak that occurs when object creation is handled in the HDF5 library. The H5O_create_ohdr function attempted to allocate memory via H5FL_reg_calloc (it is based on H5FL_reg_malloc), but the memory could not be freed before the end of the program.
In addition, H5O__chunk_deserialize, H5O__chunk_deserialize, H5O__alloc_msgs three functions also have the same leakage problem.

the following crash information:
```
=================================================================
==14==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 416 byte(s) in 1 object(s) allocated from:
    #0 0x55ab3dea1c5e in malloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:69:3
    #1 0x55ab3e0a3120 in H5FL__malloc /src/hdf5/src/H5FL.c:203:30
    #2 0x55ab3e0a3120 in H5FL_reg_malloc /src/hdf5/src/H5FL.c:355:34
    #3 0x55ab3e0a3408 in H5FL_reg_calloc /src/hdf5/src/H5FL.c:387:30
    #4 0x55ab3e1cc07d in H5O_create_ohdr /src/hdf5/src/H5Oint.c:321:10
    #5 0x55ab3e1cbd6f in H5O_create /src/hdf5/src/H5Oint.c:278:10
    #6 0x55ab3e0db26e in H5G__obj_create_real /src/hdf5/src/H5Gobj.c:248:9
    #7 0x55ab3e0dab38 in H5G__obj_create /src/hdf5/src/H5Gobj.c:137:9
    #8 0x55ab3e0e179b in H5G_mkroot /src/hdf5/src/H5Groot.c:173:13
    #9 0x55ab3e04b57f in H5F_open /src/hdf5/src/H5Fint.c:2070:13
    #10 0x55ab3e66a103 in H5VL__native_file_open /src/hdf5/src/H5VLnative_file.c:127:9
    #11 0x55ab3e63dc02 in H5VL__file_open /src/hdf5/src/H5VLcallback.c:3500:30
    #12 0x55ab3e63dc02 in H5VL_file_open /src/hdf5/src/H5VLcallback.c:3649:30
[...]

PR #4784, opened 2024-08-29
Add oss-fuzz CI Action

hyoklee:
See also "A memory leak was found while testing h5_extended_fuzzer with libfuzzer" #4586

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,37 @@
+name: fuzz
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


byrnHDF:
This workflow is triggered on a push to develop only; is this the correct trigger type? Maybe it should be scheduled and add a dispatch trigger to allow a force run.

Also, should the workflow be allowed to be run against branches?

Finally how and where does this workflow report?

derobins:
Closing since CI is not where we should be running oss-fuzz

PR #4786, opened 2024-08-29
Add bounds checking to avoid Out-of-bounds Write for gif2h5

wswsmao:
fix CVE-2022-25972
https://nvd.nist.gov/vuln/detail/CVE-2022-25972

### DIFF
--- a/hl/tools/gif2h5/decompress.c
+++ b/hl/tools/gif2h5/decompress.c
@@ -296,6 +296,10 @@ Decompress(GIFIMAGEDESC *GifImageDesc, GIFHEAD *GifHead)
              * Build the hash table on-the-fly. No table is stored in the
              * file.
              */
+            if (FreeCode >= 4096) {
+                printf("Error: FreeCode out of bounds\n");
+                exit(EXIT_FAILURE);
+            }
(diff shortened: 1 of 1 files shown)


hyoklee (approved):

hl/tools/gif2h5/decompress.c
```diff
@@ -296,6 +296,10 @@ Decompress(GIFIMAGEDESC *GifImageDesc, GIFHEAD *GifHead)
              * Build the hash table on-the-fly. No table is stored in the
              * file.
              */
+            if (FreeCode >= 4096) {
```
mattjala, line 299: Prefix/Suffix buffer size is `4096 * sizeof(int)`, so this should probably be `4096 * sizeof(int)` too.

jhendersonHDF: In this case the check would be correct since `FreeCode` is an index value into the `int` arrays

mattjala (approved):

derobins:
Do you have a test file? There's no test to ensure this is really fixed.

wswsmao:
> Do you have a test file? There's no test to ensure this is really fixed.

Ok，I can add test files as appendix

PR #4792, opened 2024-08-30
Resolve race around single counter by using two counters

qkoziol:
Gave this more thought and using a single counter is going to be racey, unless we use thread barriers, which takes away the whole point of the ping+pong semaphores.  So, use two different counters to verify correct operation.

### DIFF
--- a/test/ttsafe_semaphore.c
+++ b/test/ttsafe_semaphore.c
@@ -20,14 +20,15 @@
 
 #if defined(H5_HAVE_THREADS)
 
-#define NUM_PINGPONG     (500 * 1000)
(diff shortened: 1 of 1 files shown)


qkoziol:
@lrknox @hyoklee

byrnHDF (approved):

jhendersonHDF:
Where does the race on the counter value come into play here?

qkoziol:
> Where does the race on the counter value come into play here?

When there was a single counter, sometimes the other thread would race ahead and increment the counter again before the first thread would check the (shared) value and then the first thread would leave its loop one iteration too early.

jhendersonHDF (approved):

PR #4794, opened 2024-08-30
Fix a few issues with error reporting during sec2 reads/writes

jhendersonHDF:


src/H5FDsec2.c
```diff
@@ -700,15 +700,16 @@ H5FD__sec2_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_UNU
             int    myerrno = errno;
             time_t mytime  = time(NULL);
 
+#ifndef H5_HAVE_PREADWRITE
             offset = HDlseek(file->fd, 0, SEEK_CUR);
```
jhendersonHDF, line 704: Since `pread` doesn't change the file offset, this `HDlseek` call ends up resetting the file offset that gets printed out on errors.

src/H5FDsec2.c
```diff
@@ -700,15 +700,16 @@ H5FD__sec2_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_UNU
...
             HGOTO_ERROR(H5E_IO, H5E_READERROR, FAIL,
                         "file read failed: time = %s, filename = '%s', file descriptor = %d, errno = %d, "
-                        "error message = '%s', buf = %p, total read size = %llu, bytes this sub-read = %llu, "
-                        "bytes actually read = %llu, offset = %llu",
-                        ctime(&mytime), file->filename, file->fd, myerrno, strerror(myerrno), buf,
-                        (unsigned long long)size, (unsigned long long)bytes_in,
-                        (unsigned long long)bytes_read, (unsigned long long)offset);
+                        "error message = '%s', buf = %p, total read size = %z, bytes this sub-read = %llu, "
+                        "offset = %llu",
+                        ctime(&mytime), file->filename, file->fd, myerrno, strerror(myerrno), buf, size,
```
jhendersonHDF, line 711: Switched to `%zu` for `size` rather than casting to `unsigned long long`.

src/H5FDsec2.c
```diff
@@ -700,15 +700,16 @@ H5FD__sec2_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_UNU
...
             offset = HDlseek(file->fd, 0, SEEK_CUR);
+#endif
 
             HGOTO_ERROR(H5E_IO, H5E_READERROR, FAIL,
                         "file read failed: time = %s, filename = '%s', file descriptor = %d, errno = %d, "
-                        "error message = '%s', buf = %p, total read size = %llu, bytes this sub-read = %llu, "
-                        "bytes actually read = %llu, offset = %llu",
-                        ctime(&mytime), file->filename, file->fd, myerrno, strerror(myerrno), buf,
-                        (unsigned long long)size, (unsigned long long)bytes_in,
-                        (unsigned long long)bytes_read, (unsigned long long)offset);
```
jhendersonHDF, line 711: Removed the printing out of `bytes_read` since it will always be -1 in this case, so casting it to `unsigned long long` and printing it out just gives misleading output.

qkoziol (approved):

lrknox (approved):

PR #4797, opened 2024-08-31
Fix un-seekable test error on Cygwin

hyoklee:
This fixes ```Subprocess aborted***Exception:``` error.

### DIFF
--- a/test/tfile.c
+++ b/test/tfile.c
@@ -8101,6 +8101,7 @@ test_min_dset_ohdr(void)
     CHECK(ret, FAIL, "H5Fclose");
 } /* end test_min_dset_ohdr() */
 
+#ifndef __CYGWIN__
 /****************************************************************
 **
 **  test_unseekable_file():
@@ -8137,6 +8138,8 @@ test_unseekable_file(void)
     /* TODO - Does not properly fail on all systems */
     /* VERIFY(file_id, H5I_INVALID_HID, "H5Fopen"); */
 }
+#endif
+
 /****************************************************************
 **
 **  test_deprec():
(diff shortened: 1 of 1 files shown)


qkoziol (changes requested):
We try hard to avoid adding OS/compiler/system-specific #ifdef's in the source code as they quickly become out of date and are mysterious after a few years.   Can you find a behavior that is triggering this and help (others) solve it?   Otherwise, some way of querying the environment to determine if the test should be run is a better way of skipping tests.

mattjala:
This should have been detected in CI. Is this an intermittent failure? If unseekable files really do cause segfaults on Cygwin, then that's an issue that needs to be fixed outside of this test.

hyoklee:
@qkoziol, I hope this error message helps:

`assertion "H5AC_cache_is_clean(f, H5AC_RING_MDFSM)" failed: file "/cygdrive/d/a/hdf5/hdf5/src/H5Fint.c", line 1423, function: H5F__dest`

https://my.cdash.org/tests/197770754

@mattjala, it's a consistent failure since un-seekable test was introduced:

https://github.com/hdfeos/hdf5/actions/workflows/cyg-ninja.yml

mattjala:
> @mattjala, it's a consistent failure since un-seekable test was introduced:
> 
> [hdfeos/hdf5/actions/workflows/cyg-ninja.yml](https://github.com/hdfeos/hdf5/actions/workflows/cyg-ninja.yml)

That action should probably be added to the main repo's CI, then. 

I suspect the root issue is that Cygwin doesn't treat `NUL` as a keyword and expects `/dev/null` - I'm investigating this currently.

mattjala:
> > @mattjala, it's a consistent failure since un-seekable test was introduced:
> > [hdfeos/hdf5/actions/workflows/cyg-ninja.yml](https://github.com/hdfeos/hdf5/actions/workflows/cyg-ninja.yml)
> 
> That action should probably be added to the main repo's CI, then.
> 
> I suspect the root issue is that Cygwin doesn't treat `NUL` as a keyword and expects `/dev/null` - I'm investigating this currently.

When built on Cygwin, `H5_HAVE_WIN32_API` is false, so Cygwin is correctly trying to access /dev/null - the root issue is probably the one indicated by the MDC assertion failure.

qkoziol:
> > > @mattjala, it's a consistent failure since un-seekable test was introduced:
> > > [hdfeos/hdf5/actions/workflows/cyg-ninja.yml](https://github.com/hdfeos/hdf5/actions/workflows/cyg-ninja.yml)
> > 
> > 
> > That action should probably be added to the main repo's CI, then.
> > I suspect the root issue is that Cygwin doesn't treat `NUL` as a keyword and expects `/dev/null` - I'm investigating this currently.
> 
> When built on Cygwin, `H5_HAVE_WIN32_API` is false, so Cygwin is correctly trying to access /dev/null - the root issue is probably the one indicated by the MDC assertion failure.

Then I suggest changing that assertion into an error check, instead of the #ifdef's

jhendersonHDF:
> > > > @mattjala, it's a consistent failure since un-seekable test was introduced:
> > > > [hdfeos/hdf5/actions/workflows/cyg-ninja.yml](https://github.com/hdfeos/hdf5/actions/workflows/cyg-ninja.yml)
> > > 
> > > 
> > > That action should probably be added to the main repo's CI, then.
> > > I suspect the root issue is that Cygwin doesn't treat `NUL` as a keyword and expects `/dev/null` - I'm investigating this currently.
> > 
> > 
> > When built on Cygwin, `H5_HAVE_WIN32_API` is false, so Cygwin is correctly trying to access /dev/null - the root issue is probably the one indicated by the MDC assertion failure.
> 
> Then I suggest changing that assertion into an error check, instead of the #ifdef's

Seems we should probably determine why the metadata free space manager still has something left in it

PR #4803, opened 2024-09-04
Skip unseekable file test on Cygwin

mattjala:
Different implementation of hdf5#4797. Ideally the root issue will be fixed eventually, but for now we should avoid a known segfault during testing.

The metadata cache assertion that triggers an error in debug builds hasn't been converted to a regular error because it uses a debug-build-exclusive routine.

### DIFF
--- a/test/h5test.c
+++ b/test/h5test.c
@@ -2564,6 +2564,31 @@ h5_driver_uses_multiple_files(const char *drv_name, unsigned flags)
     return ret_val;
 }
 
+/*-------------------------------------------------------------------------
+ * Function:    h5_on_cygwin
+
+ *
+ * Purpose:     Determine whether tests are running on Cygwin
+ *
+ * Return:      True if running on Cygwin, false otherwise.
+ *
+ *-------------------------------------------------------------------------
+ */
+bool
+h5_on_cygwin(void)
+{
+    bool  ret_value = false;
+    FILE *fp        = NULL;
+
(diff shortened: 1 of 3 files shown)


hyoklee (approved):

qkoziol (changes requested):
Please don't put OS/system-specific #ifdefs in.

test/h5test.c
```diff
@@ -2564,6 +2564,31 @@ h5_driver_uses_multiple_files(const char *drv_name, unsigned flags)
...
+ *-------------------------------------------------------------------------
+ */
+bool
+h5_on_cygwin()
+{
+    bool  ret_value = false;
+    FILE *fp        = NULL;
+
+    /* If /cygdrive/ exists and is readable, assume HDF5 is running on Cygwin */
+    if ((fp = fopen("/cygdrive", "r")) != NULL) {
```
mattjala, line 2584: A perverse user could define this drive on a non-Cygwin system to make this function incorrect, but I'm not aware of any other way to check that Cygwin is in use besides the `__CYGWIN__` directive.

qkoziol (changes requested):
This is fundamentally no different than an #ifdef on the OS/system name.  Like #ifdefs, this will quickly become mysterious and forgotten, disabled forever.

derobins:
I'm not a fan of papering over failing tests. We should try to understand and fix the underlying issue.

mattjala:
Closing this, since the root issue should be fixed instead. Details in #4804

Issue #4672, opened 2024-07-24
links test has unreported failures


qkoziol:
**Describe the bug**
When running the 'links' test, it reports failures in the output, but doesn't cause the test to fail.

**Expected behavior**
Not fail

**Platform (please complete the following information)**
Current develop branch on MacOS

**Additional context**

Sample output:

```
============================
Test log for links
VOL connector: pass_through under_vol=1;under_info={under_vol=0;under_info={}}
============================

-Testing with unminimzed dataset object headers-

--Testing with 'old format'--
Testing link creation                                                  PASSED
Testing link queries                                                   PASSED
Testing link queries using deprecated routines                         PASSED
Testing H5Lcreate functions                                            PASSED
Testing new link queries                                               PASSED
Testing long names for objects & links                                 PASSED
Testing too many links                                                 PASSED
Testing H5Lcreate path handling                                        PASSED
Testing link creation property lists                                   PASSED
Testing link creation property lists using deprecated routines         PASSED
Testing H5Lmove                                                        PASSED
Testing H5Lcopy                                                        PASSED
Testing moving and copying links preserves their properties            PASSED
Testing moving and copying links using deprecated routines preserves their properties PASSED
Testing backwards compatibility                                        PASSED
Testing external file cache with external links                        PASSED
Testing external links across multiple files                           PASSED
Testing H5Pset/get_elink_acc_flags()                                   PASSED

---Testing without external file cache---
Testing external link to root                                          PASSED
Testing external link to root using deprecated routines                PASSED
Testing external link to object on path                                PASSED
Testing external link to self                                          PASSED
Testing external links back and forth                                  PASSED
Testing too many external links                                        PASSED
Testing dangling external links                                        PASSED
Testing recursive external links                                       PASSED
Testing query aspects of external link                                 PASSED
Testing query aspects of external link using deprecated routines       PASSED
Testing unlinking external link in compact group                       PASSED
Testing unlinking external link in dense group                         PASSED
Testing move external link                                             PASSED
Testing external link along for the ride                               PASSED
Testing that external files are closed during traversal                PASSED
Testing that external files are closed during traversal                PASSED
Testing endianness of external links                                   PASSED
Testing that external files work with strong file close degree         PASSED
Testing external links via H5Pset_elink_prefix()                       PASSED
Testing external links via main file's absolute path                   PASSED
Testing external links via main file's CWD + relative path             PASSED
Testing external links via current working directory                   PASSED
[...]

PR #4809, opened 2024-09-05
Fix use of public API calls

qkoziol:
Switch public API calls to private ones.

Also minor code cleanups


qkoziol:
Root cause of https://github.com/HDFGroup/hdf5/issues/4672, which it fixes.

mattjala (approved):

derobins (approved):

Issue #4684, opened 2024-08-01
Inconsistent naming of hdf5_hl_fortran library between autotools and CMake build


jsberg-bnl:
**Describe the bug**
A CMake build will create libraries named `libhdf5_hl_fortran.*`. The autotools build will instead create libraries named `libhdf5hl_fortran.*`, then create soft links `libhdf5_hl_fortran.so` and `libhdf5_hl_fortran.a` pointing to the corresponding `libhdf5hl_fortran.*` files. On MacOS, the dynamic library is named `libhdf5hl_fortran.dylib`, and no corresponding link is made there. 

**Expected behavior**
Both the autotools build and the CMake build should create libraries with the same names, and for hl_fortran, those names should be `libhdf5_hl_fortran.*`

**Platform (please complete the following information)**
 - HDF5 version: 1.14.4-3
 - OS and version: Arch fully updated; MacOS 14.5 w/ HomeBrew fully updated
 - Compiler and version: gcc 14.1
 - Build system (e.g. CMake, Autotools) and version: CMake 3.30.1, latest autotools
 - Any configure options you specified: configure: --enable-fortran --enable-cxx --without-szlib --without-zlib; CMake: -G "Unix Makefiles" -DHDF5_BUILD_JAVA=OFF
 - MPI library and version (parallel HDF5): n/a

**Additional context**
Also see issue #4526, addressing more inconsistencies between the two build methods.

lrknox:
Names corrected to be libhdf5_hl_fortran.* in develop, will be in next release (2.0.0).

PR #4811, opened 2024-09-06
Change name of libhdf5hl_fortran installed by autotools 

lrknox:
to libhdf5_hl_fortran to be consistent with CMake install and with other hl lib files.  Switched corresponding symlink to libhdf5hl_fortran.

Fixes issue #4684.


byrnHDF (approved):

mattjala:
The issue was flagged for 1.16 - is this the kind of thing we should  delay until the next major update to avoid surprising users?

derobins:
This can go in now, but should be develop only. Renaming libraries should only be done in a major release.

derobins:
Also, this needs a RELEASE.txt entry

derobins:
Decision is to:

1) Remove the symlink
2) Rename the library to be consistent with CMake and other hl libraries
3) This will go to develop only
4) Will announce in a RELEASE.txt note and on the forum

derobins (approved):

PR #4815, opened 2024-09-07
Update NVHPC optimization settings

derobins:
* Use -gopt in Autotools/CMake instead of -g
* Autotools uses -O3 for release, -O1 for debug
* CMake uses whatever the defaults are for the configuration
* Remove CMake optimization flag removal hack

### DIFF
--- a/config/cmake/HDFCompilerFlags.cmake
+++ b/config/cmake/HDFCompilerFlags.cmake
@@ -54,46 +54,8 @@ if (CMAKE_C_COMPILER_ID STREQUAL "NVHPC" )
       set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -s")
     endif ()
   else ()
-    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Mbounds -g")
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Mbounds -gopt")
   endif ()
-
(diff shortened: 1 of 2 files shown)


bmribler (approved):

lrknox (approved):

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -54,46 +54,8 @@ if (CMAKE_C_COMPILER_ID STREQUAL "NVHPC" )
...
-
-  # With at least NVHPC 23.5 - 23.9, compiling with -O2 or higher and -DNDEBUG
-  # appears to have issues that manifest in the tests as incorrect metadata
-  # checksums being read or memory being corrupted. Compiling without -DNDEBUG
-  # does not appear to have these issues, but is not ideal due to compiling in
-  # asserts and other library debug code. Compiling with -O1 also does not appear
-  # to have these issues, so set maximum optimization level to -O1 for now until
-  # it can be determined whether these issues are compiler-specific or issues
-  # in the library.
-  set (cmake_c_flags_minsizerel_edited "${CMAKE_C_FLAGS_MINSIZEREL}")
```
jhendersonHDF, line 68: Shouldn't we be keeping this in and checking the version of NVHPC that's used, or are we only going to commit to supporting the latest version of NVHPC, even though older versions are used on HPC machines?

PR #4816, opened 2024-09-07
Add CMake build mode flags to hdf5lib.settings

derobins:
Flags from the CMake build mode (e.g., optimization) are not a part of
CMAKE_<language>_FLAGS and were not exported to the libhdf5.settings file. This
has been fixed and the C, Fortran, and C++ build mode flags are now exported to
the file.

This also affects the text output of H5check_version() and the libhdf5.settings
string stored in the library (for those who use strings(1), etc. to get
build info from the binary).


config/cmake/HDF5DeveloperBuild.cmake
```diff
@@ -17,11 +17,8 @@ set (CMAKE_CXX_FLAGS_DEVELOPER ${CMAKE_CXX_FLAGS_DEBUG} CACHE STRING
   "Flags used by the C++ compiler during developer builds." FORCE
 )
 
-# Set CMake C flags based off of Debug build flags. Add in -Og
-# option to disable some GCC optimizations that might affect
-# debugging negatively and also include some GCC compiler passes
-# that collect debugging information
-set (CMAKE_C_FLAGS_DEVELOPER "${CMAKE_C_FLAGS_DEBUG} -Og" CACHE STRING
+# Set CMake C flags based off of Debug build flags
+set (CMAKE_C_FLAGS_DEVELOPER "${CMAKE_C_FLAGS_DEBUG}" CACHE STRING
```
derobins, line 21: This is already set in the debug flags

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -465,3 +465,33 @@ if (HDF5_ENABLE_OPTIMIZATION)
...
 endif ()
 MARK_AS_ADVANCED (HDF5_ENABLE_OPTIMIZATION)
+
+#-----------------------------------------------------------------------------
+# The build mode flags are not added to CMAKE_C_FLAGS, so create a separate
+# variable for them so they can be written out to libhdf5.settings and
+# H5build_settings.c
+#-----------------------------------------------------------------------------
+set (HDF5_BUILD_MODE_C_FLAGS "")
+set (HDF5_BUILD_MODE_Fortran_FLAGS "")
```
byrnHDF, line 475: Shouldn't the Fortran and CXX flags be set in the HDFFortranCompilerFlags.cmake and HDFCXXCompilerFlags.cmake files?

derobins: I can do that

config/cmake/HDF5DeveloperBuild.cmake
jhendersonHDF: This is `CMAKE_C_FLAGS_DEBUG`, which will only capture the debug flags that are added by default by CMake, or explicitly added at configure time by the user. Not the flags that we add separately in the library's configuration.

config/cmake/HDFCompilerFlags.cmake
```diff
@@ -465,3 +465,33 @@ if (HDF5_ENABLE_OPTIMIZATION)
...
+set (HDF5_BUILD_MODE_CXX_FLAGS "")
+if ("${HDF_CFG_NAME}" STREQUAL "Debug")
```
jhendersonHDF, line 477: This is going to be problematic for multi-config generators

config/cmake/HDF5DeveloperBuild.cmake
jhendersonHDF: That said, some compilers don't have `-Og`, so this does need to be refactored at some point, but I still find it useful.

mattjala (changes requested):
Move CXX/Fortran flag setting to correct cmake files

PR #4820, opened 2024-09-10
Remove extensions from the package binaries - msi, dmg, deb,rpm

### LINKED ISSUE #4716: Change release/snapshot binaries to plain deb, rpm, dmg, msi packages
Currently binaries include license file and readme file in addition to the installer.
Those binaries should just be the installers.

byrnHDF:
Closes #4716

### DIFF
--- a/.github/workflows/cmake-ctest.yml
+++ b/.github/workflows/cmake-ctest.yml
@@ -273,7 +273,7 @@ jobs:
         id: publish-ctest-deb-binary
         run: |
           mkdir "${{ runner.workspace }}/builddeb"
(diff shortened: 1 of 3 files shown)


jhendersonHDF (approved):
Consider that it's not usually a bad idea to have both versions of all these, as long as the compression is actually doing something worthwhile in terms of resulting size

byrnHDF:
> Consider that it's not usually a bad idea to have both versions of all these, as long as the compression is actually doing something worthwhile in terms of resulting size

The compressed file was including a COPYING and README that was decided to not be needed. So therefore the removal of the extensions.

hyoklee (approved):

PR #4821, opened 2024-09-10
Fix a bug in Subfiling VFD vector I/O setup

jhendersonHDF:
Fixes a bug where the vector I/O sizes weren't being extended when one of the entries in the array is 0. This caused an overread of the I/O sizes buffer and on some machines would cause a memory allocation failure due to the calculated I/O vector size being too large.

### DIFF
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -823,6 +823,24 @@ Bug Fixes since HDF5-1.14.0 release
 ===================================
     Library
     -------
+    - Fixed a bug in the Subfiling VFD that could cause a buffer over-read
+      and memory allocation failures
+
+      When performing vector I/O with the Subfiling VFD, making use of the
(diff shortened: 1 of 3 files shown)


jhendersonHDF:
Will add a release note after checking testing on an HPC machine. The current testing in t_vfd.c wasn't generally catching this as the I/O sizes array was allocated on the stack and passed in. The buffer over-read was then usually reading valid memory for the process. Adding a test with an array allocated on the heap has a much better chance of triggering the issue.

lrknox:
MPI_TEST_t_vfd and 2930 other tests all passed on frontier, testing the subfiling_vector_io_bugfix branch with gcc-native/13.2 compiler:  https://my.cdash.org/builds/2656168.  The test has been failing consistently in this configuration.

qkoziol (approved):

lrknox (approved):

PR #4839, opened 2024-09-16
Fix the CMake code to use zlibng option properly

byrnHDF:
Until a better method is proposed, used #ifdef of new define to use the correct name of function/variable between zlibng and zlib.


derobins (approved):

byrnHDF:
fixes #4840

ajelenak:
Is this going to be included in 1.14.5?

PR #4842, opened 2024-09-17
Fix memory leaks in ttsafe tests

mattjala:
Valgrind still reports possibly lost memory in `tts_error()/tts_dcreate()`:

```
==3035271== 352 bytes in 1 blocks are possibly lost in loss record 2 of 3
==3035271==    at 0x484DA83: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==3035271==    by 0x40147D9: calloc (rtld-malloc.h:44)
==3035271==    by 0x40147D9: allocate_dtv (dl-tls.c:375)
==3035271==    by 0x40147D9: _dl_allocate_tls (dl-tls.c:634)
==3035271==    by 0x52A07B4: allocate_stack (allocatestack.c:430)
==3035271==    by 0x52A07B4: pthread_create@@GLIBC_2.34 (pthread_create.c:647)
==3035271==    by 0x4FC9C33: H5TS_thread_create (H5TSthread.c:270)
==3035271==    by 0x1134E0: tts_error (ttsafe_error.c:122)
==3035271==    by 0x4867384: PerformTests (testframe.c:319)
==3035271==    by 0x10C3BF: main (ttsafe.c:178)
```

But I think it's spurious due to something about the order of valgrind's operations and can be ignored. I've verified that all the test threads are joined in each case, so I don't think the thread stack memory is actually leaked.

### DIFF
--- a/test/ttsafe_acreate.c
+++ b/test/ttsafe_acreate.c
@@ -63,7 +63,11 @@ tts_acreate(void)
     int    buffer, i;
     herr_t status;
 
-    ttsafe_name_data_t *attrib_data;
+    ttsafe_name_data_t *attrib_data[NUM_THREADS];
+
+    char *attribute_name = NULL;
+
+    memset(attrib_data, 0, sizeof(attrib_data));
 
     /*
      * Create an HDF5 file using H5F_ACC_TRUNC access, default file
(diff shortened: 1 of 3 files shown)


qkoziol (approved):

test/ttsafe_acreate.c
```diff
@@ -63,7 +63,11 @@ tts_acreate(void)
     int    buffer, i;
     herr_t status;
 
-    ttsafe_name_data_t *attrib_data;
+    ttsafe_name_data_t *attrib_data[NUM_THREADS];
+
+    char *attribute_name = NULL;
+
+    memset(attrib_data, 0, sizeof(attrib_data));
```
derobins, line 70: Does this generate size warnings?

mattjala: Nothing cropped up in the -Werror builds. NUM_THREADS is 16, so this is 16 * 8 bytes per pointer = 128, which I think should be fine.

fortnern (approved):

test/ttsafe_acreate.c
derobins: The -Werror builds don't check --enable-threadsafe

PR #4852, opened 2024-09-18
Clean up thread-local error stacks in all threads

mattjala:
When threadsafety is enabled, threadlocal error stacks in threads that don't call `H5_term_library` do not have any error records stored on them cleaned up properly. This leads to leaked memory and, before the recent changes to H5E,  an infinite loop during library close. 

Set of changes:
-    Cleanup threadlocal error stack at individual thread exit time
-    Rename H5E__clear_stack() to H5E__destroy_stack() to contrast its intention with H5E_clear_stack()
-    Add a ttsafe test

This PR was originally much more complex, and the below description is out-of-date.

=====
Fixing this is complicated by the fact that H5TS's cleanup is currently performed after every individual module is terminated. In order for the records to be released properly, H5TS must terminate before H5E and after H5CX, necessitating rearranging the termination order. As a side effect, most H5TS resources that were previously set up and released on a per-process basis (`H5TS_tinfo_mtx_s`, `H5TS_api_info_p`, `H5TS_thrd_info_key_g`) are now handled per-thread.

This is currently a draft PR due to an issue with thread once macros on MacOS builds. The flag the library uses to perform only one initialization of H5TS across all threads, `H5TS_first_thread_init_s`, is not currently cleaned up at library close. If the library is closed and re-opened, the library will see the old value and not properly initialize H5TS.

This can be solved on most systems by re-initializing `H5TS_first_init_s = H5TS_ONCE_INITIALIZER` during H5TS termination. However, it appears that on MacOS, the `PTHREAD_ONCE_INIT` macro is defined in a manner that causes problems with this approach:

```
 In file included from /Users/runner/work/hdf5/hdf5/src/H5private.h:65,
                 from /Users/runner/work/hdf5/hdf5/src/H5TSint.c:35:
/Users/runner/work/hdf5/hdf5/src/H5TSint.c: In function 'H5TS_term_package':
/Users/runner/work/hdf5/hdf5/src/H5TSprivate.h:71:36: error: expected expression before '{' token
   71 | #define H5TS_ONCE_INITIALIZER      PTHREAD_ONCE_INIT
      |                                    ^~~~~~~~~~~~~~~~~
/Users/runner/work/hdf5/hdf5/src/H5TSint.c:151:25: note: in expansion of macro 'H5TS_ONCE_INITIALIZER'
  151 |     H5TS_first_init_s = H5TS_ONCE_INITIALIZER;
      |                         ^~~~~~~~~~~~~~~~~~~~~
```

My best guess is  that MacOS enforces the requirement that `PTHREAD_ONCE_INIT` only be used for static initializations, and this initialization during termination is dynamic. Windows threads have both a static and dynamic once flag initializer (`INIT_ONCE_STATIC_INIT` and `InitOnceInitialize`) but I haven't been able to find a dynamic counterpart for pthreads.

### DIFF
--- a/src/H5Eint.c
+++ b/src/H5Eint.c
@@ -646,7 +646,7 @@ H5E__get_current_stack(void)
     estack_copy->auto_data = current_stack->auto_data;
 
     /* Empty current error stack */
-    H5E__clear_stack(current_stack);
+    H5E__destroy_stack(current_stack);
 
     /* Set the return value */
     ret_value = estack_copy;
@@ -685,7 +685,7 @@ H5E__set_current_stack(H5E_stack_t *estack)
         HGOTO_ERROR(H5E_ERROR, H5E_CANTGET, FAIL, "can't get current error stack");
 
     /* Empty current error stack */
-    H5E__clear_stack(current_stack);
+    H5E__destroy_stack(current_stack);
 
     /* Copy new stack to current error stack */
     current_stack->nused = estack->nused;
@@ -715,7 +715,7 @@ H5E__close_stack(H5E_stack_t *estack, void H5_ATTR_UNUSED **request)
     assert(estack);
 
     /* Release the stack's error information */
-    H5E__clear_stack(estack);
+    H5E__destroy_stack(estack);
 
     /* Free the stack structure */
     estack = H5FL_FREE(H5E_stack_t, estack);
@@ -1699,16 +1699,20 @@ done:
 } /* end H5E_clear_stack() */
 
 /*-------------------------------------------------------------------------
- * Function:    H5E__clear_stack
+ * Function:    H5E__destroy_stack
  *
- * Purpose:     Clear the specified error stack
+ * Purpose:     Clear all internal state within an error stack, as a precursor to freeing it.
+ *
+ *              At present, this is nearly identical to H5E_clear_stack(),
+ *              but if additional resources are added to the error stack in the future,
+ *              they will only be released by this routine and not by H5E_clear_stack().
  *
  * Return:      SUCCEED/FAIL
  *
  *-------------------------------------------------------------------------
  */
 herr_t
-H5E__clear_stack(H5E_stack_t *estack)
+H5E__destroy_stack(H5E_stack_t *estack)
 {
     herr_t ret_value = SUCCEED; /* Return value */
 
@@ -1726,7 +1730,7 @@ H5E__clear_stack(H5E_stack_t *estack)
 
 done:
     FUNC_LEAVE_NOAPI(ret_value)
-} /* end H5E__clear_stack() */
+} /* end H5E__destroy_stack() */
 
 /*-------------------------------------------------------------------------
  * Function:    H5E__pop
--- a/src/H5TSint.c
+++ b/src/H5TSint.c
@@ -525,10 +525,12 @@ H5TS__tinfo_destroy(void *_tinfo_node)
     FUNC_ENTER_PACKAGE_NAMECHECK_ONLY
 
     if (tinfo_node) {
-        /* Add thread info node to the free list */
         H5TS_mutex_lock(&H5TS_tinfo_mtx_s);
+        /* Add thread info node to the free list */
         tinfo_node->next       = H5TS_tinfo_next_free_s;
         H5TS_tinfo_next_free_s = tinfo_node;
+        /* Release resources held by error records in thread-local error stack */
+        H5E__destroy_stack(&tinfo_node->info.err_stack);
         H5TS_mutex_unlock(&H5TS_tinfo_mtx_s);
     }
 
--- a/test/ttsafe_error_stacks.c
+++ b/test/ttsafe_error_stacks.c
@@ -1,0 +1,114 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * Copyright by The HDF Group.                                               *
+ * All rights reserved.                                                      *
+ *                                                                           *
+ * This file is part of HDF5.  The full HDF5 copyright notice, including     *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the COPYING file, which can be found at the root of the source code       *
+ * distribution tree, or in https://www.hdfgroup.org/licenses.               *
+ * If you do not have access to either file, you may request a copy from     *
+ * help@hdfgroup.org.                                                        *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+#include "ttsafe.h"
+
+#ifdef H5_HAVE_THREADS
+
+#define ERR_CLS_NAME        "Custom error class"
+#define ERR_CLS_LIB_NAME    "example_lib"
+#define ERR_CLS_LIB_VERSION "0.1"
+
+#define ERR_MAJOR_MSG "Okay, Houston, we've had a problem here"
+#define ERR_MINOR_MSG "Oops!"
+
+H5TS_THREAD_RETURN_TYPE generate_hdf5_error(void *arg);
+H5TS_THREAD_RETURN_TYPE generate_user_error(void *arg);
+
+/* Helper routine to generate an HDF5 library error */
+H5TS_THREAD_RETURN_TYPE
+generate_hdf5_error(void H5_ATTR_UNUSED *arg)
+{
+    H5TS_thread_ret_t ret_value = 0;
+    ssize_t           nobjs     = 0;
+
+    H5E_BEGIN_TRY
+    {
+        nobjs = H5Fget_obj_count(H5I_INVALID_HID, H5F_OBJ_ALL);
+    }
+    H5E_END_TRY
+
+    /* Expect call to fail */
+    VERIFY(nobjs, FAIL, "H5Fget_obj_count");
+
+    return ret_value;
+}
+
+/* Helper routine to generate a user-defined error */
+H5TS_THREAD_RETURN_TYPE
+generate_user_error(void H5_ATTR_UNUSED *arg)
+{
+    H5TS_thread_ret_t ret_value = 0;
+    hid_t             cls       = H5I_INVALID_HID;
+    hid_t             major     = H5I_INVALID_HID;
+    hid_t             minor     = H5I_INVALID_HID;
+    herr_t            status    = FAIL;
+
+    cls = H5Eregister_class(ERR_CLS_NAME, ERR_CLS_LIB_NAME, ERR_CLS_LIB_VERSION);
+    CHECK(cls, H5I_INVALID_HID, "H5Eregister_class");
+
+    major = H5Ecreate_msg(cls, H5E_MAJOR, ERR_MAJOR_MSG);
+    CHECK(major, H5I_INVALID_HID, "H5Ecreate_msg");
+
+    minor = H5Ecreate_msg(cls, H5E_MINOR, ERR_MINOR_MSG);
+    CHECK(minor, H5I_INVALID_HID, "H5Ecreate_msg");
+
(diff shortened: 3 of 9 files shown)


src/H5.c
```diff
@@ -208,15 +208,6 @@ H5_init_library(void)
      */
     if (!H5_dont_atexit_g) {
```
qkoziol: Why is this change necessary?

src/H5Eint.c
```diff
@@ -283,20 +283,11 @@ H5E_term_package(void)
...
-         * clear that stack manually.
-         *
-         * Error message IDs will typically still be
-         * in use by the default error stack when the
-         * application does H5E_BEGIN/END_TRY cleanup
-         * at the very end.
-         */
+        /* Clear the global error stack if it exists.
+         * Thread-local error stack cleanup is handled in H5TS. */
+#ifndef H5_HAVE_THREADSAFE
```
qkoziol: I don't think this should be skipped for threadsafe builds.

src/H5TSint.c
```diff
@@ -135,26 +135,32 @@ H5TS__init(void)
...
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
+    /* Clean up per-thread library info */
+    H5TS__tinfo_term();
+
+    /* Tell subsequent library opens that thread information is not initialized */
+    H5TS_first_init_s = H5TS_ONCE_INITIALIZER;
+
+    /* Release critical section / mutex for modifying the thread info globals */
+    H5_UNLIKELY(H5TS_mutex_destroy(&H5TS_tinfo_mtx_s));
```
qkoziol: Why did you move these destroy+delete out of the H5TS_tinfo_term call?

src/H5TSint.c
```diff
@@ -556,18 +568,30 @@ H5TS__tinfo_term(void)
         HGOTO_DONE(FAIL);
     while (H5TS_tinfo_next_free_s) {
         H5TS_tinfo_node_t *next = H5TS_tinfo_next_free_s->next;
+
+        /* Free any entries within the error stack */
+        H5E__clear_stack(&H5TS_tinfo_next_free_s->info.err_stack);
```
qkoziol: Two problems: there are no routines called here that can push errors on the stack, and I am removing all the calls to H5E_clear_stack from the library (except for within the error code and the single call in the FUNC_LEAVE_* macros.

src/H5TSint.c
```diff
@@ -556,18 +568,30 @@ H5TS__tinfo_term(void)
...
-    /* Release critical section / mutex for modifying the thread info globals */
-    if (H5_UNLIKELY(H5TS_mutex_destroy(&H5TS_tinfo_mtx_s) < 0))
+    if (tinfo_node) {
+        /* Free any entries within the error stack */
+        H5E__clear_stack(&tinfo_node->info.err_stack);
+        H5MM_free(tinfo_node);
+    }
+
+    /* Clear the thread info key to prevent bad reuse if the library is re-opened later in this process */
+    if (H5_UNLIKELY(H5TS_key_set_value(H5TS_thrd_info_key_g, NULL) < 0))
```
qkoziol: Have you seen evidence of this reuse?

src/H5TSint.c
```diff
@@ -135,26 +135,32 @@ H5TS__init(void)
...
+int
 H5TS_term_package(void)
 {
     FUNC_ENTER_NOAPI_NOINIT_NOERR
 
+    /* Clean up per-thread library info */
+    H5TS__tinfo_term();
+
+    /* Tell subsequent library opens that thread information is not initialized */
+    H5TS_first_init_s = H5TS_ONCE_INITIALIZER;
```
qkoziol: This looks like a good idea, but I don't know if the standard(s) supports it, so I don't think it'll work.   That's why I put the H5TS_term_package as an atexit() callback, instead of when the library is closed.

src/H5TSint.c
```diff
@@ -556,18 +568,30 @@ H5TS__tinfo_term(void)
...
+    /* Release the thread info node for this thread since it's not on the free list */
+    if (H5_UNLIKELY(H5TS_key_get_value(H5TS_thrd_info_key_g, (void **)&tinfo_node) < 0))
         HGOTO_DONE(FAIL);
 
-    /* Release critical section / mutex for modifying the thread info globals */
-    if (H5_UNLIKELY(H5TS_mutex_destroy(&H5TS_tinfo_mtx_s) < 0))
+    if (tinfo_node) {
+        /* Free any entries within the error stack */
+        H5E__clear_stack(&tinfo_node->info.err_stack);
+        H5MM_free(tinfo_node);
```
qkoziol: Freeing the memory is good, but not the H5E_clear_stack call.

qkoziol (changes requested):
I don't think that most of the library changes should go in.  I've added comments, but I'm also happy to talk in-person or on a call.

src/H5.c
mattjala: In order to be able to clean up error records properly, H5TS needs to perform the cleanup before `H5E_term_package()`. It would be possible to split `H5TS_term_package()` into a part that runs at process exit here and a part that runs during library shutdown so it can precede H5E, but I figured it was better practice to keep it all in one place.

src/H5Eint.c
mattjala: At the time `H5TS_term_package()` is performed during a threadsafe build, H5TS has to clean up the thread info node containing the stack, so it must also clean up the stack. That's done before H5E_term_package(), so its stack is already cleaned, so H5E can't clean it up again. I agree this is a bad pattern. It could be resolved by splitting H5TS termination into two routines.

src/H5TSint.c
mattjala: Are you saying it's a problem because we can't push any errors in the event that this fails, or that this routine is unnecessary because no errors can be pushed here? If it's the latter, this is meant to clean up errors that were pushed before thread termination (see `ttsafe_error_stacks.c`). If the former, I would assume that returning `FAIL` if this fails should be sufficient, same as the rest of the termination routine. (I'll add a check for that.)

To your second point - I agree that's a problem! We may need to discuss that at greater length.

src/H5TSint.c
mattjala: I think this is a holdover from an initial design that split H5TS_term into two routines - I've moved them back into tinfo_term.

src/H5TSint.c
mattjala: I saw it during testing, but I think this is another holdover from an initial design that had two H5TS term routines. Removed in newer commit.

src/H5.c
qkoziol: What part of the H5TS code needs cleaned up before the H5E shuts down?

src/H5TSint.c
qkoziol: I meant the latter.   But, if there's any errors on the default error stack, it's the H5E package's responsibility to clear them, not the H5TS package.

src/H5Eint.c
qkoziol: I can't see how the H5TS shutdown could possibly occur before the H5E package shuts down.

src/H5TSint.c
```diff
@@ -353,8 +353,13 @@ H5TS__tinfo_create(void)
         HGOTO_DONE(NULL);
 
     /* Reuse an info struct that's on the free list if possible */
-    if (NULL != (tinfo_node = H5TS_tinfo_next_free_s))
+    if (NULL != (tinfo_node = H5TS_tinfo_next_free_s)) {
+        /* Clear the error stack associated with that thread before reuse */
+        if (H5E__clear_stack(&tinfo_node->info.err_stack) < 0)
```
qkoziol: It's not about checking the return value, this routine shouldn't be called here.

src/H5.c
mattjala: It's not the H5TS code that needs to be cleaned up before H5E per se; it's part of H5E that needs to be cleaned up before  H5E shuts down, and only H5TS has the information necessary to perform the clean up.

As the library is presently, multiple pieces of memory don't get cleaned up properly during the execution of `ttsafe_error_stacks`:

- the error descriptions allocated during `H5E__set_stack_entry()`
- the temporary function and files strings allocated during `H5Epush2()`

If you just add `H5E__clear_stack()` calls to the thread info cleanup in `H5TS_term_package()`, then the H5I_dec_ref() calls fails due to having already cleaned up the error classes/messages. (This then cascades into an assertion failure, since it tries to push an error after having already cleaned up H5E.)

src/H5TSint.c
mattjala: These errors are in the thread-local error stacks from other threads. H5E could clear them itself, but that would require exposing more H5TS functionality to H5E.

src/H5.c
qkoziol: Thanks, that helps.   I see what you are thinking about now.

This is a more fundamental problem, which we're going to have to partially solve with "don't do that" instructions to application developers.

For instance, it may not be supported for one thread to shut down the library while other threads exist, since they may be using or about to use the library.  And since we can't detect when an application creates or joins threads, it's likely that we will have to take a solution like MPI: you must initialize the library before your application creates threads and can only shut it down (possibly via an atexit() callback) after all the threads are joined.

Otherwise, I don't think there's a way to correctly get the 'once' initialization to work correctly.

qkoziol: All that said, I think that the H5E__clear_stack() call _does_ need to be made within the H5TS__tinfo_destroy() routine, so that the error stack can be cleaned up when a thread is joined.

qkoziol: I don't think it belongs in the other places, and I don't think the split shutdown that you are describing for releasing the main thread's error stack is necessary, since the main thread will call H5E_term_package() to release it.

Hmm - might have to refcount (with an atomic int) the number of threads active in the library, in order to know when the last one (the main one) is shutting down and the error stack destroy in H5TS__tinfo_destroy() should be skipped.

qkoziol: Probably worthwhile to set up a call to talk, if you'd like.

mattjala: I'm not sure if "don't do that" instructions will suffice here - this problem comes up whenever the last API routine used in a thread raises an HDF5 error.  Even if the applications is nice and closes all of its threads before terminating the library, the memory will still be lost, because all the per-thread cleanup does is add its thread info node to the global free list, which is still subject to the cleanup ordering problems.

I'll reach out to set up a time to talk through this.

qkoziol: Agree about the resource leak for threads that are joined, which is why I suggested adding the H5E__clear_stack() to H5TS_tinfo_destroy().

Cool, happy to talk.

mattjala:
Vastly simplified this after a discussion with Quincey.
- Cleanup threadlocal error stack at individual thread exit time
- Rename `H5E__clear_stack()` to `H5E__destroy_stack()` to contrast its intention with `H5E_clear_stack()`
- Add a ttsafe test

test/ttsafe_error_stacks.c
```diff
@@ -0,0 +1,112 @@
...
+ * This file is part of HDF5.  The full HDF5 copyright notice, including     *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the COPYING file, which can be found at the root of the source code       *
+ * distribution tree, or in https://www.hdfgroup.org/licenses.               *
+ * If you do not have access to either file, you may request a copy from     *
+ * help@hdfgroup.org.                                                        *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+#include "ttsafe.h"
+
+#ifdef H5_HAVE_THREADS
```
qkoziol: This ifdef should check "H5_HAVE_THREADSAFE" instead, since per-thread error stacks (etc) are only set up when the API is threadsafe.  (As opposed to the library being thread-enabled for internal use)

qkoziol (approved):
Small change, otherwise good  :-)

derobins (approved):

PR #4853, opened 2024-09-18
Reduce test iterations, etc. to keep time down on slower systems

qkoziol:


test/ttsafe_rwlock.c
```diff
@@ -23,8 +23,8 @@
 #define NUM_THREADS 16
 #define NUM_WRITERS 4
 
-#define NUM_ITERS 32
-#define COUNT_MAX 1024
+#define NUM_ITERS 12
+#define COUNT_MAX 512
```
lrknox, line 27: 16 and 512 reduced runtime on echidna to 23 minutes;  8 and 512 reduced it to 8 minutes.  12 seems like a happy medium, I'll test it.  This isn't a problem on most other tested machines.  If it works on echidna I expect it has a good chance to eliminate any occasional failures elsewhere.

lrknox: Some, but maybe not all of the echidna test configurations will pass.  It may need one last revision of the numbers, but I think it's worthwhile to try these to see where it stands,

lrknox (approved):

derobins (approved):

test/ttsafe_rwlock.c
qkoziol: Feel free to reduce the # of iterations down to 10 (or 8, even) if necessary.

PR #4863, opened 2024-09-20
Replace deprecated vlen release in ttsafe test

mattjala:


qkoziol:
As you mentioned, this header should already be getting included, so I have a feeling this is not the solution.  I'll look at it today and see if I can learn more.

qkoziol:
Found the problem: H5Dvlen_reclaim is deprecated in favor of H5Treclaim.   So, the code in ttsafe_attr_vlen.c should use the new routine.  (And doesn't need any additional headers)

qkoziol (approved):

hyoklee (approved):

PR #4871, opened 2024-09-23
Remove H5E_clear_stack() from library

qkoziol:
Use existing H5G_loc_exists() routine

### DIFF
--- a/src/H5Ocopy.c
+++ b/src/H5Ocopy.c
@@ -1464,12 +1464,13 @@ H5O__copy_search_comm_dt(H5F_t *file_src, H5O_t *oh_src, H5O_loc_t *oloc_dst /*i
 
             /* Walk through the list of datatype suggestions */
             while (suggestion) {
+                bool exists = false;
+
                 /* Find the object */
-                if (H5G_loc_find(&dst_root_loc, suggestion->path, &obj_loc /*out*/) < 0)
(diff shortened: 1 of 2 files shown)


mattjala (approved):

src/H5Ocopy.c
```diff
@@ -1464,22 +1464,28 @@ H5O__copy_search_comm_dt(H5F_t *file_src, H5O_t *oh_src, H5O_loc_t *oloc_dst /*i
 
             /* Walk through the list of datatype suggestions */
             while (suggestion) {
+                bool exists = false;
+
                 /* Find the object */
-                if (H5G_loc_find(&dst_root_loc, suggestion->path, &obj_loc /*out*/) < 0)
-                    /* Ignore errors - i.e. suggestions not present in
-                     * destination file */
-                    H5E_clear_stack();
```
jhendersonHDF, line 1471: If you haven't already, it's worth documenting that developers should avoid calling `H5E_clear_stack()` in the library since I can easily see this cropping up again.

jhendersonHDF (approved):

src/H5Ocopy.c
qkoziol: Good point, I'll add that in the next PR.

derobins (approved):

PR #4875, opened 2024-09-24
Remove a call to H5E_clear_stack() in H5O code

qkoziol:


src/H5Oint.c
```diff
@@ -1594,37 +1594,39 @@ H5O_obj_type(const H5O_loc_t *loc, H5O_type_t *obj_type)
...
+        HGOTO_ERROR(H5E_OHDR, H5E_CANTGET, FAIL, "unable to determine object class");
 
-        /* Set type to "unknown" */
-        *obj_type = H5O_TYPE_UNKNOWN;
-    }
-    else
+    if (obj_class)
         /* Set object type */
         *obj_type = obj_class->type;
+    else
```
mattjala: It doesn't seem like it's actually possible for this case to occur -  if `obj_class` is NULL, then that means `H5O__obj_class_real()` encountered an error and the `HGOTO_ERROR` up above was triggered. There's no way for `obj_class` to be NULL and for execution of `H5O__obj_type_real()` to continue.

qkoziol: Agree; I'll update

qkoziol: Updated

mattjala (approved):

derobins (approved):

PR #4881, opened 2024-09-26
Split H5TS termination into threadlocal/thread-global parts

mattjala:


src/H5TSint.c
```diff
@@ -151,8 +153,8 @@ H5TS_term_package(void)
     H5TS_mutex_destroy(&H5TS_api_info_p.api_mutex);
     H5TS_atomic_destroy_uint(&H5TS_api_info_p.attempt_lock_count);
 
-    /* Clean up per-thread library info */
-    H5TS__tinfo_term();
+    /* Release critical section / mutex for modifying the thread info globals */
+    H5TS_mutex_destroy(&H5TS_tinfo_mtx_s);
```
qkoziol: Why did you move this up?  Seems like it would be better where it was.

src/H5.c
```diff
@@ -433,11 +433,14 @@ H5_term_library(void)
             /* Shut down the following packages in strictly the order given
              * by the table.
              */
+        ,   TERMINATOR(CX, true)
```
qkoziol: Why did you need to move the H5CX termination earlier?

src/H5TSint.c
mattjala: The mutex is a thread-global resource guarding access to thread-global resources (the free list of thread-info nodes, the next thread id), so it makes sense to clean it up with the other global resources.

src/H5.c
mattjala: I was under the impression that the H5TS cleanup would remove a context node that H5CX expected to exist, but this doesn't actually seem to be true. I'll undo this.

src/H5TSint.c
qkoziol: Understood, but the global resources for the thread-local info are created in H5TS__tinfo_init and it's more logical to destroy all of them in H5TS__tinfo_term.

mattjala: Fair point. If more threadlocal setup is ever added it might make sense to breakup tinfo_init() into a thread-local/thread-global counterparts to match this idea, but right now it makes sense to keep the mutex destroy where it is.

qkoziol (approved):

jhendersonHDF (approved):

PR #4884, opened 2024-09-26
Add threadsafety warning and release notice

mattjala:
Separate 1.14 merge of two recent PRs, since I initially flagged them as develop only.

### DIFF
--- a/doc/threadsafety-warning.md
+++ b/doc/threadsafety-warning.md
@@ -1,0 +1,16 @@
+## A Warning
+
+Any application that creates threads that use the HDF5 library must join those threads before either process exit or library close through H5close(). If all HDF5-using threads aren't joined, the threads may exhibit undefined behavior.
+
+## Discussion for Developers on Potential Improvements
+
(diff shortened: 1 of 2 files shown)


lrknox:
Release notes were cleaned from the hdf5_1_14 branch when the hdf5_1_14_5 release branch was created.  Any release notes for fixes and changes in the HDF5 1.14.5 release should only go to the hdf5_1_14_5 branch as they will be added to the HISTORY-1_14 .txt file when the release is announced.  The only release notes in the RELEASE.txt file on the hdf5_1_14 branch in the meantime should be those for HDF5 1.14.6 that will not be released in 1.14.5. 

Probably it will be easiest to just cherry-pick the two PRs to the hdf5_1_14_5 branch, and sync develop changes to hdf5_1_14 after release, taking care not to bring in release notes for already released changes.

jhendersonHDF:
We also usually want to keep the "1.14" out of the PR name and commit message

lrknox (approved):

PR #4891, opened 2024-09-27
Refactor h5test.c, testframe.c and testpar.h testing frameworks

jhendersonHDF:
Added new testframe.h header to document testing framework functions and split them away from h5test.h and from test programs that don't integrate with the testframe.c testing framework

Added new test setup callback to testframe.c testing framework

Added parameters to AddTest() to specify size of test parameters so they can be copied for later use

Enabled HDF5 error stacks in testframe.c framework by default and added some error stack suppressions to some testhdf5 tests

Added new maxthreads option to testframe.c framework to allow specifying the maximum number of threads a multi-threaded test can use

Moved TestExpress functionality out of testframe.c and into more general h5test.c for wider use by tests through getter and setter

Updated some tests to not mix and match functionality between h5test.c/h and testframe.c/h

Moved some functionality from testphdf5.h into testpar.h for parallel tests that aren't part of testphdf5

Added new parallel test library that contains common shared functionality for parallel tests (similar to h5test library)


c++/test/dsets.cpp
```diff
@@ -1492,9 +1492,6 @@ test_dset()
     catch (Exception &E) {
         test_report(nerrors, H5std_string(" Dataset"));
     }
-
-    // Clean up data file
-    cleanup_dsets();
```
jhendersonHDF, line 1497: No need to explicitly call this here since the testing framework will do this now

c++/test/dsets.cpp
```diff
@@ -1506,8 +1503,10 @@ test_dset()
  *-------------------------------------------------------------------------
  */
 extern "C" void
-cleanup_dsets()
+cleanup_dsets(void *params)
 {
-    HDremove(FILE1.c_str());
-    HDremove(FILE_ACCPLIST.c_str());
+    if (GetTestCleanup()) {
```
jhendersonHDF, line 1508: This is a similar pattern replicated everywhere. The testing framework always calls the cleanup callback for a test now as the callback can do more than just cleanup temp. testing files. `GetTestCleanup()` just surrounds the part where the testing files are cleaned up.

hl/test/test_packet_vlen.c
```diff
@@ -861,9 +861,11 @@ verify_attribute(hid_t fid, const char *table_name, const char *attr_name)
 
     /* Verify values read in */
     for (ii = 0; ii < ATTR_DIM; ii++)
-        if (attr_data[ii] != read_data[ii])
-            TestErrPrintf("%d: attribute data different: attr_data[%d]=%d, read_data[%d]=%d\n", __LINE__, ii,
```
jhendersonHDF, line 865: `TestErrPrintf` was converted to `fprintf(stderr, ...)` for a few test files that weren't really integrated with the testframe.c framework.

test/dsets.c
```diff
@@ -990,14 +990,16 @@ test_compact_io(hid_t fapl)
 
             /* Verify the dataset's layout and fill message versions */
             if (fp->shared->low_bound == H5F_LIBVER_EARLIEST) {
-                VERIFY(dsetp->shared->layout.version, H5O_LAYOUT_VERSION_DEFAULT, "layout_ver_bounds");
```
jhendersonHDF, line 993: Replaced usages of the `VERIFY` macro with equivalent logic for some test files that weren't really integrated with the testframe.c framework

test/dsets.c
```diff
@@ -16165,8 +16168,7 @@ main(void)
         goto error;
     printf("All dataset tests passed.\n");
 #ifdef H5_HAVE_FILTER_SZIP
-    if (GetTestCleanup())
```
jhendersonHDF, line 16168: This test never changes the test cleanup status and wasn't really integrated with the testframe.c framework, so the call to GetTestCleanup() wasn't useful.

test/earray.c
```diff
@@ -2302,8 +2302,6 @@ main(void)
     /* Reset library */
     h5_test_init();
     fapl = h5_fileaccess();
-    if (TestExpress > 0)
```
jhendersonHDF, line 2305: This test doesn't do anything with the value of TestExpress

test/enc_dec_plist.c
```diff
@@ -191,9 +191,6 @@ main(void)
                      "Testing ENCODE/DECODE with file version bounds: (%s, %s):", low_string, high_string);
             puts(msg);
 
-            if (VERBOSE_MED)
```
jhendersonHDF, line 194: This test never used the testframe.c framework and never changed the test verbosity, so checking for `VERBOSE_MED` had no effect.

test/enc_dec_plist_cross_platform.c
```diff
@@ -26,9 +26,6 @@ static int test_plists(const char *filename_prefix);
 int
 main(void)
 {
-    if (VERBOSE_MED)
```
jhendersonHDF, line 29: Same comment here as above

test/farray.c
```diff
@@ -1633,8 +1633,6 @@ main(void)
     /* Reset library */
     h5_test_init();
     fapl = h5_fileaccess();
-    if (TestExpress > 0)
```
jhendersonHDF, line 1636: This test doesn't do anything with the value of TestExpress

test/h5test.c
```diff
@@ -114,11 +114,15 @@ const char *LIBVER_NAMES[] = {"earliest", /* H5F_LIBVER_EARLIEST = 0  */
 static H5E_auto2_t err_func = NULL;
 
 /* Global variables for testing */
-size_t   n_tests_run_g     = 0;
-size_t   n_tests_passed_g  = 0;
-size_t   n_tests_failed_g  = 0;
-size_t   n_tests_skipped_g = 0;
-uint64_t vol_cap_flags_g   = H5VL_CAP_FLAG_NONE;
+static int TestExpress_g     = -1; /* Whether to expedite testing. -1 means not set yet. */
```
jhendersonHDF, line 117: TestExpress functionality was moved back out of the testframe.c framework for more widespread usage in the future, but the global variable is once again hidden behind get/set functions so that tests don't have direct access to it.

test/h5test.c
```diff
@@ -114,11 +114,15 @@ const char *LIBVER_NAMES[] = {"earliest", /* H5F_LIBVER_EARLIEST = 0  */
...
+size_t     n_tests_run_g     = 0;
+size_t     n_tests_passed_g  = 0;
+size_t     n_tests_failed_g  = 0;
+size_t     n_tests_skipped_g = 0;
+uint64_t   vol_cap_flags_g   = H5VL_CAP_FLAG_NONE;
+
+/* Whether h5_cleanup should clean up temporary testing files */
+static bool do_test_file_cleanup_g = true;
```
jhendersonHDF, line 125: While this duplicates the `GetTestCleanup()` functionality from testframe.c into h5test.c, it prevents conflating the two functionalities. h5test.c's usage was also less structured (the application had to call `h5_fixname` to influence the status), so it seemed best to leave them separate for now until a potential unified testing framework can come together in the future.

test/h5test.c
```diff
@@ -1403,24 +1482,6 @@ h5_get_file_size(const char *filename, hid_t fapl)
...
 H5_GCC_CLANG_DIAG_ON("format-nonliteral")
 
-/*
- * This routine is designed to provide equivalent functionality to 'printf'
- * and allow easy replacement for environments which don't have stdin/stdout
- * available. (i.e. Windows & the Mac)
- */
-H5_ATTR_FORMAT(printf, 1, 2)
-int
-print_func(const char *format, ...)
```
jhendersonHDF, line 1413: No longer needed at this point

test/h5test.h
```diff
@@ -24,77 +24,6 @@
 #include "H5private.h"
 #include "H5Eprivate.h"
 
-/*
```
jhendersonHDF, line 27: Nearly all of the below code was moved into testframe.h

test/stab.c
```diff
@@ -1466,10 +1466,9 @@ main(void)
     puts("All symbol table tests passed.");
 
     /* Cleanup */
-    if (GetTestCleanup()) {
```
jhendersonHDF, line 1469: This test wasn't integrated with testframe.c and never changed the cleanup status; `GetTestCleanup()` calls weren't useful here

test/tattr.c
```diff
@@ -4009,7 +4009,16 @@ test_attr_big(hid_t fcpl, hid_t fapl)
     /* Create attribute */
     u = 2;
     snprintf(attrname, sizeof(attrname), "attr %02u", u);
-    attr = H5Acreate2(dataset, attrname, H5T_NATIVE_UINT, big_sid, H5P_DEFAULT, H5P_DEFAULT);
+
+    if (vol_is_native && low != H5F_LIBVER_LATEST) {
```
jhendersonHDF, line 4013: This replicates the logic below and suppresses an error stack output from an expected failure

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
+static int TestMaxNumThreads_g = -1; /* Max number of threads that can be spawned */
+
+static bool TestDoSummary_g = false; /* Show test summary. Default is no. */
+static bool TestDoCleanUp_g = true;  /* Do cleanup or not. Default is yes. */
+
+int TestFrameworkProcessID_g = 0;         /* MPI process rank value for parallel tests */
+int TestVerbosity_g          = VERBO_DEF; /* Default Verbosity is Low */
 
 /*
- * Setup a test function and add it to the list of tests.
```
jhendersonHDF, line 55: Documentation for all the functions here went into testframe.h, so header comments became brief.

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
- *    Since only the pointer is copied, the contents should not change.
- * Return: Void
- *    exit EXIT_FAILURE if error is encountered.
+ * Add a new test to the list of tests to be executed
  */
-void
-AddTest(const char *TheName, void (*TheCall)(void), void (*Cleanup)(void), const char *TheDescr,
-        const void *Parameters)
+herr_t
+AddTest(const char *TestName, void (*TestFunc)(const void *), void (*TestSetupFunc)(void *),
```
jhendersonHDF, line 62: Added a test setup callback and a parameter for the size of test data so that the framework can copy the data and the same structure can be changed between calls to `AddTest()`

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
- *      provided.
- * private_parser: Optional routine provided by test program to parse the
- *      private options.  Default to NULL which means none is provided.
- *
+ * Initialize the testing framework
  */
-void
-TestInit(const char *ProgName, void (*private_usage)(void), int (*private_parser)(int ac, char *av[]))
+herr_t
+TestInit(const char *ProgName, void (*TestPrivateUsage)(FILE *stream),
```
jhendersonHDF, line 138: Added `FILE *` parameters here and to other functions that print output. Also added an extra `TestProcessID` parameter that is used to control output for parallel tests, while not requiring any MPI datatypes or calls in this function.

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
         }
         else if (((strcmp(*argv, "-only") == 0) || (strcmp(*argv, "-o") == 0))) {
             if (argc > 0) {
-                unsigned Loop;
-
                 --argc;
                 ++argv;
 
-                /* Skip all tests, then activate only one. */
-                if (!skipped_all) {
```
jhendersonHDF, line 262: Moved this logic into `SetTest()`

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
             else {
-                TestUsage();
-                exit(EXIT_FAILURE);
+                ret_value = FAIL;
+                goto done;
             }
         }
         else if ((strcmp(*argv, "-summary") == 0) || (strcmp(*argv, "-s") == 0))
-            Summary = 1;
-        else if (strcmp(*argv, "-enable-error-stack") == 0)
```
jhendersonHDF, line 276: To prevent hiding of problems, enabled error stacks by default for testframe.c tests. This includes testhdf5, so some new error stack suppressions had to be added to some tests.

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
+            TestEnableErrorStack = false;
+        }
         else if ((strcmp(*argv, "-help") == 0) || (strcmp(*argv, "-h") == 0)) {
-            TestUsage();
+            TestUsage(stdout);
             exit(EXIT_SUCCESS);
         }
         else if ((strcmp(*argv, "-cleanoff") == 0) || (strcmp(*argv, "-c") == 0))
             SetTestNoCleanup();
+        else if ((strcmp(*argv, "-maxthreads") == 0) || (strcmp(*argv, "-t") == 0)) {
```
jhendersonHDF, line 330: Added a new `-maxthreads` command-line option for testframe.c tests for use with multi-threaded tests in the future.

test/testframe.c
```diff
@@ -391,140 +530,77 @@ TestShutdown(void)
...
-                express_val = 2;
-            else
-                express_val = 3;
-        }
-
-        SetTestExpress(express_val);
-    }
-
-    return (TestExpress);
+    return h5_get_testexpress();
```
jhendersonHDF, line 561: Uses the functionality from h5test.c. The goal is to make it easy for testframe.c tests to use functionality from both frameworks if need be, while making it more difficult for h5test.c-style tests to use functionality from testframe.c.

test/testframe.c
```diff
@@ -552,16 +646,7 @@ GetTestNumErrs(void)
...
 IncTestNumErrs(void)
 {
-    num_errs++;
-}
-
-/*
- * Retrieve the current Test Parameters pointer.
- */
-H5_ATTR_PURE const void *
-GetTestParameters(void)
```
jhendersonHDF, line 562: Previously, the testing framework had a global pointer that it set to the uncopied test data stored for each test when the test was executing. Then `GetTestParameters()` just returned that pointer to the test. Since the testing framework now copies test data to make it easy to re-use test data between calls to `AddTest()`, this was simplified by passing the stored pointer directly and removing the need to call `GetTestParameters()`.

test/testhdf5.h
```diff
@@ -161,94 +164,70 @@
         }                                                                                                    \
     } while (0)
 
-/* Used to document process through a test */
-#if defined(H5_HAVE_PARALLEL) && defined(H5_PARALLEL_TEST)
-#define MESSAGE(V, A)                                                                                        \
```
jhendersonHDF, line 166: Moved into testframe.h and simplified by using the new 'test process ID' parameter

test/testhdf5.h
```diff
@@ -161,94 +164,70 @@
...
-                       "invalid result\n",                                                                   \
-                       where, (int)__LINE__, __FILE__);                                                      \
+            printf("   Call to routine: %15s at line %4d in %s returned "                                    \
+                   "invalid result\n",                                                                       \
+                   where, (int)__LINE__, __FILE__);                                                          \
         TestErrPrintf("*** UNEXPECTED RESULT from %s at line %4d in %s\n", where, (int)__LINE__, __FILE__);  \
     } while (0)
 
-/* definitions for command strings */
-#define VERBOSITY_STR "Verbosity"
```
jhendersonHDF, line 193: These macros aren't used anywhere

testpar/CMakeLists.txt
jhendersonHDF, line 1: Added a new parallel testing library in testpar/testpar.c that contains shared functionality among parallel tests. Only 2 functions are in it currently, but it separates parallel tests with a `main()` function from testframe.c-style tests like testphdf5.

testpar/t_2Gio.c
jhendersonHDF, line 1: This file had a lot of functions copied into it from testphdf5 without having made any changes to those functions and without using them in this file, so they were removed. This file also had included testphdf5.h just for some of the macros and support functions, which were mostly moved into the parallel testing library.

testpar/t_2Gio.c
```diff
@@ -1348,6 +1266,22 @@ dataset_writeAll(void)
     /* Dataset5: point selection in File - Hyperslab selection in Memory*/
     /* create a file dataspace independently */
     point_set(start, count, stride, block, num_points, coords, OUT_OF_ORDER);
+    if (VERBOSE_MED) {
```
jhendersonHDF, line 1269: This previously existed in `point_set()`, but that function was moved into the parallel testing library where `VERBOSE_MED` isn't available (since some tests don't integrate with the testframe.c framework). So this output section was moved into individual tests that do have access to `VERBOSE_MED`.

testpar/t_prestart.c
```diff
@@ -115,8 +122,6 @@ main(int argc, char **argv)
     if (data_array)
         free(data_array);
 
-    nerrors += GetTestNumErrs();
```
jhendersonHDF, line 118: This test never used testframe.c functionality, so `GetTestNumErrs()` isn't useful here

testpar/t_shapesame.c
```diff
@@ -4319,8 +4282,6 @@ main(int argc, char **argv)
         return -1;
     }
 
-    mpi_rank_framework_g = mpi_rank;
```
jhendersonHDF, line 4322: Instead of relying on parallel tests to set this global variable for the testframe.c framework somewhere in main, it gets passed to `TestInit()` now instead

test/testframe.h
```diff
@@ -0,0 +1,803 @@
...
+/* Prototypes */
+/**************/
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * --------------------------------------------------------------------------
+ * \ingroup H5TEST
```
mattjala, line 128: `H5TEST` -> `TestInit`

test/testframe.h
```diff
@@ -0,0 +1,803 @@
...
+ *
+ *          <b>Note:</b> when a parallel test calls TestInfo(), the output is
+ *          only printed from the MPI process with rank value 0.
+ *
+ */
+H5TEST_DLL void TestInfo(FILE *stream);
+
+/**
+ * --------------------------------------------------------------------------
+ * \ingroup H5TEST
```
mattjala, line 256: Would it make more sense to have the doxygen group for these routines be called something like H5TESTFRAME?

test/testhdf5.c
```diff
@@ -47,46 +47,54 @@ main(int argc, char *argv[])
...
     /* Perform requested testing */
     PerformTests();
 
     /* Display test summary, if requested */
     if (GetTestSummary())
-        TestSummary();
+        TestSummary(stdout);
 
     /* Clean up test files, if allowed */
     if (GetTestCleanup() && !getenv(HDF5_NOCLEANUP))
```
mattjala, line 100: It seems like test-level cleanup functions are invoked twice here. Once by `PerformTests()` after each test, and then during `TestCleanup()`.

jhendersonHDF: Yes, this is rather an open question of whether the old behavior is useful in any way. testhdf5 previously waited until all tests have run before cleaning up, which has its own problems even when cleanup callbacks were only handling the removal of temporary testing files. I left that behavior in for now since it's currently harmless, but I'm inclined toward removing the TestCleanup() function entirely and just letting the testing framework do it. `GetTestCleanup()` would probably need to be modified to handle the `HDF5_NOCLEANUP` environment variable.

test/testframe.h
jhendersonHDF: No particular opinion on the group name, especially since we don't currently have any testing documentation and this documentation won't get parsed by doxygen yet anyway until we decide to start doing that. H5TEST was just a generic name to keep all testing documentation together, but having a separate group for this framework would work just as well.

test/tskiplist.c
```diff
@@ -1555,7 +1555,7 @@ test_skiplist_term(void)
 **
 ****************************************************************/
 void
-test_skiplist(void)
```
mattjala, line 1558: `test_skiplist()` has a initialization routine (`test_skiplist_init()`) and a termination routine (`test_skiplist_term()`) that handle some global buffers. Are the test-framework level setup/cleanup functions just for HDF5-library level setup and cleanup?

test/testframe.c
```diff
@@ -11,368 +11,504 @@
...
-                MESSAGE(5, ("There were %d errors detected.\n\n", (int)Test[Loop].NumErrors));
-            }
+
+        if (TestFrameworkProcessID_g == 0) {
+            MESSAGE(2, ("Testing  -- %s (%s) \n", TestArray[Loop].Description, TestArray[Loop].Name));
+            MESSAGE(5, ("===============================================\n"));
         }
 
-    Test_parameters = NULL; /* clear it. */
+        TestAlarmOn();
```
mattjala, line 407: It would be good to check for failure here

test/testhdf5.h
```diff
@@ -17,17 +17,20 @@
 #ifndef TESTHDF5_H
 #define TESTHDF5_H
 
-/* Include generic testing header also */
+/* Include generic testing header */
 #include "h5test.h"
 
+/* Include testing framework functionality */
+#include "testframe.h"
```
mattjala, line 24: Testframe currently includes h5test, so including them both here is redundant. I'm guessing the goal is to eventually separate the two?

test/tskiplist.c
jhendersonHDF: I hadn't looked into the content of any of the tests for possible changes. In this case, the test's setup and cleanup is performed just once before and after all the tests, where the test setup and cleanup callbacks are used on a per-test basis. Though it could make sense to allow specifying pre- and post-tests callbacks in `TestInit`, for example.

test/testhdf5.h
jhendersonHDF: Separate or ideally unify. In the meantime, both headers are included here without assuming what each header includes. `testframe.h` mostly just includes `h5test.h` as a convenience since library developers have generally assumed they can just include `h5test.h` and get everything needed. This is moving a bit more toward a test either including `testframe.h` or `h5test.h`, depending on the functionality needed, until tests are maybe made consistent at some point in the future. The general rule with these changes is the testhdf5-style tests can access functionality from both `testframe.h` and `h5test.h`, but not the other way around.

mattjala (approved):

qkoziol:
Looking at this now...

jhendersonHDF:
Rebased off of develop

test/testhdf5.c
qkoziol: One possibility would be to have separate cleanups for individual tests and for the entire framework.

testpar/t_2Gio.c
qkoziol: All of these replicated blocks would really benefit from a refactor into a subroutine.

testpar/t_2Gio.c
```diff
@@ -1892,324 +1922,96 @@ dataset_readAll(void)
...
     /* -------------------
      * START AN HDF5 FILE
      * -------------------*/
     /* setup file access template */
-    acc_tpl = create_faccess_plist(comm, info, facc_type);
-    VRFY((acc_tpl >= 0), "");
-
-    /* Reduce the number of metadata cache slots, so that there are cache
-     * collisions during the raw data I/O on the chunked dataset.  This stresses
-     * the metadata cache and tests for cache bugs. -QAK
```
qkoziol, line 1958: Why was this removed?

testpar/t_2Gio.c
```diff
@@ -1892,324 +1922,96 @@ dataset_readAll(void)
...
+        for (i = 0; i < (int)orig_size; i++)
+            printf("%s%d", i ? ", " : "", written[i]);
+        printf("\n");
     }
-
-    /* create a memory dataspace independently */
-    mem_dataspace = H5Screate_simple(MAX_RANK, block, NULL);
-    VRFY((mem_dataspace >= 0), "");
-
-    /* Extend its current dim sizes before writing */
```
qkoziol, line 2029: Also, why was this test removed?

testpar/t_2Gio.c
```diff
@@ -1892,324 +1922,96 @@ dataset_readAll(void)
...
-}
-
-/*
- * Example of using the parallel HDF5 library to create an extendable dataset
- * and perform I/O on it in a way that verifies that the chunk cache is
- * bypassed for parallel I/O.
- */
-
-void
-extend_writeInd2(void)
```
qkoziol, line 2129: Same here - why was this test removed?

testpar/t_2Gio.c
```diff
@@ -2274,631 +2076,22 @@ extend_writeInd2(void)
...
-    VRFY((ret > 0), "H5Sget_simple_extent_dims succeeded");
-    dims[0]++;
-    ret = H5Dset_extent(dataset1, dims);
-    VRFY((ret < 0), "H5Dset_extent failed as expected");
-
-    /* restore auto error reporting */
-    H5Eset_auto2(H5E_DEFAULT, old_func, old_client_data);
-    H5Sclose(file_dataspace);
-
-    /* Read dataset1 using BYROW pattern */
```
qkoziol, line 2367: Same here - why removed?

testpar/t_2Gio.c
```diff
@@ -2274,631 +2076,22 @@ extend_writeInd2(void)
...
-/*
- * Example of using the parallel HDF5 library to create two extendible
- * datasets in one HDF5 file with collective parallel MPIO access support.
- * The Datasets are of sizes (number-of-mpi-processes x dim0) x dim1.
- * Each process controls only a slab of size dim0 x dim1 within each
- * dataset.
- */
-
-void
-extend_writeAll(void)
```
qkoziol, line 2466: Same here - removed?

testpar/t_2Gio.c
```diff
@@ -2274,631 +2076,22 @@ extend_writeInd2(void)
...
-     * -------------------*/
-    /* setup file access template */
-    acc_tpl = create_faccess_plist(comm, info, facc_type);
-    VRFY((acc_tpl >= 0), "");
-
-    /* Reduce the number of metadata cache slots, so that there are cache
-     * collisions during the raw data I/O on the chunked dataset.  This stresses
-     * the metadata cache and tests for cache bugs. -QAK
-     */
-    {
```
qkoziol, line 2520: Same here - removed?

testpar/t_2Gio.c
```diff
@@ -2274,631 +2076,22 @@ extend_writeInd2(void)
...
-
-    /* Release file-access template */
-    ret = H5Pclose(acc_tpl);
-    VRFY((ret >= 0), "");
-
-    /* --------------------------------------------------------------
-     * Define the dimensions of the overall datasets and create them.
-     * ------------------------------------------------------------- */
-
-    /* set up dataset storage chunk sizes and creation property list */
```
qkoziol, line 2545: Same here - removed?

testpar/t_2Gio.c
```diff
@@ -2274,631 +2076,22 @@ extend_writeInd2(void)
...
-
-    /* put some trivial data in the data_array */
-    dataset_fill(start, block, data_array1);
-    MESG("data_array initialized");
-    if (VERBOSE_MED) {
-        MESG("data_array created");
-        dataset_print(start, block, data_array1);
-    }
-
-    /* create a memory dataspace independently */
```
qkoziol, line 2634: So many tests removed, I'll stop commenting :-)

qkoziol (commented):
Generally speaking, this looks like a solid improvement to the testing infrastructure, particularly the TLC invested in improving the parallel tests.  Just concerned about some of the tests that were removed.

testpar/t_2Gio.c
jhendersonHDF: My comment above may not have showed up inline, but every single test function that was removed from this file wasn't being tested in the first place. They were exact or near-exact duplicates of the same functions from testphdf5's t_dset.c and they weren't actually being run by the t_2Gio.c testing; they simply existed in this file for no reason. Since t_2Gio.c previously included testphdf5.h, these functions matched the prototypes declared there for t_dset.c's tests and caused confusion for some code editors.

test/testhdf5.c
jhendersonHDF: Yes I think at this point it makes sense to add in a callback or two that applies for the whole test program to cover the old way that testhdf5 performed cleanup after all the tests ran. It's also been pointed out that one or more tests have a run-once setup callback.

testpar/t_2Gio.c
jhendersonHDF: Will look into pulling that back out into a separate function that isn't tied to `VERBOSE_MED`

lrknox (approved):

PR #4899, opened 2024-09-28
merge missing changes from 1.14.5

byrnHDF:


hl/test/test_ds.c
jhendersonHDF, line 1: I think this file is missing changes, specifically:

```C
    /* test finding dataset dsname */
    if ((ds_existed = H5LTfind_dataset(fid, dsname)) < 0)
        goto out;
    if (ds_existed == 0) {
        printf("Unexpected result: Dataset \"%s\" does exist\n", dsname);
        goto out;
    }
```

The -Werror build failed previously because the `ds_existed` is unused currently in this PR.

lrknox: After this is merged I'll do a sync from develop, and will watch for this change in the sync.

jhendersonHDF (approved):

glennsong09 (approved):

PR #4922, opened 2024-10-03
Create HISTORY file and clear RELEASE

byrnHDF:


release_docs/RELEASE.txt
```diff
@@ -36,7 +36,7 @@ CONTENTS
 
 - New Features
 - Support for new platforms and languages
-- Bug Fixes since HDF5-1.14.0
+- Bug Fixes since HDF5-1.16.0
```
lrknox: Shouldn't this be since HDF5-1.14.5?  This will be the 1.16.0 release.
Same for line 131.

byrnHDF: Well 1.14 was since 1.12.0 which likely wasn't correct either - but yes I like 1.14.5

lrknox: I started wondering about what previous .0 versions used, and saw what you did for 1.14.0, and then for 1.12.0 there were these 2 lines:
- Bug Fixes since HDF5-1.12.0-alpha1
- Major Bug Fixes since HDF5-1.10.0

for this one I think since 1.14.5 might be appropriate, since 1.16 is shaping up as a continuation of 1.14 and not a something totally new with log-running parallel development.

lrknox (approved):

release_docs/RELEASE.txt
```diff
@@ -818,1066 +121,23 @@ New Features
...
-
+    -
 
 
 Support for new platforms, languages and compilers
 ==================================================
     -
 
-Bug Fixes since HDF5-1.14.0 release
+Bug Fixes since HDF5-1.16.0 release
```
```suggestion
Bug Fixes since HDF5-1.14.5 release
```

PR #4929, opened 2024-10-04
Use H5VLcmp_connector_cls to compare VOL connector IDs in API tests

jhendersonHDF:


jhendersonHDF:
This should fix the current issue with the VOL connector test CI

mattjala (approved):

derobins (approved):

PR #4931, opened 2024-10-04
Update clang-format to 17

derobins:
clang-format 13 is getting old. This updates the repository to 17.

Formatting changes are minimal. A few unnecessary `clang-format off/on` lines
were removed, but no new options have been added to .clang-format.


derobins:
fwiw, I would probably not cherry pick this directly to 1.14 and 1.16. I would just update the GitHub actions to use v17 and then simply reformat the source and commit that.

derobins:
I checked out all the options that were added in clang-format 14-17 and there wasn't anything I thought made a useful difference. I also tried pulling all the clang-format comments to see if we could ditch them. Only a few were unnecessary and they've been removed in this PR.

derobins:
One thing I've noticed is that my Ubuntu 23.10 VM uses 17.0.2 while the action uses 17.0.4 and there are minor differences in the output. Not sure if anyone cares about that.

qkoziol (approved):
Generally a little better than previously  👍🏼

hyoklee (approved):

PR #4936, opened 2024-10-07
Use H5VLcmp_connector_cls to compare VOL connector IDs in parallel API tests

jhendersonHDF:


jhendersonHDF:
Forgot to fix the parallel API tests as well. I believe they will still fail until the log VOL is fixed to not use H5VLpeek_connector_id_by_(name/value)

mattjala (approved):

qkoziol (approved):

Issue #4849, opened 2024-09-18
Setting strict libver bounds in FAPL causes unexpected H5Fopen failures that do not occur with file switch


markcmiller86:
In the code below, I compare setting setting libver bounds in FAPL used in H5Fopen to using *switch* to change libver bounds after the file is open.

Since the libver bounds in the fapl is intended to effect *only* the objects *created* thereafter, I am puzzled why the attempts to open a file created with libver set to LATEST then fail when FAPL sets libver to v18. I mean, shouldn't the open succeed and then behave as if `H5Fset_libver_bounds()` was called. Its particularly strange in the read-only case because no objects will be created under those conditions and the libver setting is kinda meaningless.

```
#include <hdf5.h>
  
int main(int argc, char **argv)
{
    hid_t fapl1 = H5Pcreate(H5P_FILE_ACCESS);
    hid_t fapl2 = H5Pcreate(H5P_FILE_ACCESS);
    hid_t fapl3 = H5Pcreate(H5P_FILE_ACCESS);

    H5Pset_libver_bounds(fapl1, H5F_LIBVER_LATEST, H5F_LIBVER_LATEST);
    hid_t fid = H5Fcreate("foo.hdf5", H5F_ACC_TRUNC, H5P_DEFAULT, fapl1);
    H5Pclose(fapl1);
    H5Fclose(fid);

    /* fails in read-write mode */
    H5Pset_libver_bounds(fapl2, H5F_LIBVER_V18, H5F_LIBVER_V18);
    fid = H5Fopen("foo.hdf5", H5F_ACC_RDWR, fapl2);
    H5Fclose(fid);

    /* fails in read-only mode */
    fid = H5Fopen("foo.hdf5", H5F_ACC_RDONLY, fapl2);
    H5Fclose(fid);
    H5Pclose(fapl2);

    /* But, this works fine */
    fid = H5Fopen("foo.hdf5", H5F_ACC_RDWR, fapl3);
    H5Fset_libver_bounds(fid, H5F_LIBVER_V18, H5F_LIBVER_V18);
    H5Pclose(fapl3);
    H5Fclose(fid);

    return 0;
}
```

markcmiller86:
I believe that if you reverse the ordering of the libver settings and instead create file using a fapl with libver v18 in both high and low and then attempt to open it with a fapl that sets libver to latest, that also fails.

markcmiller86:
[...]

PR #4939, opened 2024-10-08
Fix for issue #4849 that settings in fapl libver bounds causes unexpe…

vchoi-hdfgroup:
…cted H5Fopen failures.

File with non-SWMR-write access can now be opened without regard for superblock version. Due to the fix, H5Fstart_swmr_write() also needs to be modified as well as the tests for libver bounds. The "RFC: Setting Bounds for Object Creation in HDF5 1.10.0" is also updated to reflect the changes.

### DIFF
--- a/src/H5Fint.c
+++ b/src/H5Fint.c
@@ -3181,7 +3181,8 @@ H5F__set_libver_bounds(H5F_t *f, H5F_libver_t low, H5F_libver_t high)
     assert(f->shared);
 
     /* Set the bounds only if the existing setting is different from the inputs */
-    if (f->shared->low_bound != low || f->shared->high_bound != high) {
+    if ((f->shared->low_bound != low || f->shared->high_bound != high) &&
+        !(H5F_INTENT(f) & H5F_ACC_SWMR_WRITE)) {
         /* Call the flush routine, for this file */
         /* Note: This is done in case the binary format for representing a
          *      metadata entry class changes when the file format low / high
@@ -3704,6 +3705,7 @@ done:
  *                  --only allow datasets and groups without attributes
  *                  --disallow named datatype with/without attributes
  *                  --disallow opened attributes attached to objects
+ *                  --disallow opened objects below 1.10
  *
  * NOTE:        Currently, only opened groups and datasets are allowed
  *              when enabling SWMR via H5Fstart_swmr_write().
@@ -3746,7 +3748,7 @@ H5F__start_swmr_write(H5F_t *f)
     if (f->shared->sblock->super_vers < HDF5_SUPERBLOCK_VERSION_3)
         HGOTO_ERROR(H5E_FILE, H5E_BADVALUE, FAIL, "file superblock version - should be at least 3");
 
-    /* Check for correct file format version */
+    /* Check for correct file format version to start SWMR writing */
     if ((f->shared->low_bound < H5F_LIBVER_V110) || (f->shared->high_bound < H5F_LIBVER_V110))
         HGOTO_ERROR(H5E_FILE, H5E_BADVALUE, FAIL,
                     "file format version does not support SWMR - needs to be 1.10 or greater");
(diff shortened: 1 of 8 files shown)


test/swmr.c
```diff
@@ -1554,20 +1554,26 @@ test_metadata_read_retry_info(hid_t in_fapl)
...
- *    Close "dataset2"
- *      Create "dataset3"--should be using latest chunk indexing
- *      Close "dataset3"
- *      Close the group and file
+ *  (B) Open the file with write + non-latest-format (earliest, latest)
+ *      --file has v3 superblock
+ *      Create a group in the file
+ *      Create a chunked dataset "dataset2" in the group--keep it open
+ *      Verify "dataset2" is using v1 btree indexing
+ *      Should fail to enable SWMR writing
```
mattjala, line 1562: I thought the idea behind the libver bounds was to try and use earlier versions for file format compatibility, but to "fallback" to later versions if necessary to complete the requested operations. Is this a violation of that principle?

vchoi-hdfgroup: If I understand the question correctly:  this test is for testing H5Fstart_swmr_write() which requires V1.10 version 3 superblock to begin with.

mattjala: I was trying to understand the significance of the changes to libver bounds handling. I think I was confusing object creation versus later operations on those objects. I guess it makes sense for SWMR to fail if old objects exist that it can't handle.

mattjala (approved):

src/H5Fint.c
```diff
@@ -3783,6 +3785,31 @@ H5F__start_swmr_write(H5F_t *f)
...
+            if (NULL == (oloc = H5O_get_loc(obj_ids[u])))
+                HGOTO_ERROR(H5E_SYM, H5E_NOTFOUND, FAIL, "object not found");
+
+            if (H5O_get_native_info(oloc, &ninfo, H5O_NATIVE_INFO_HDR) < 0)
+                HGOTO_ERROR(H5E_FILE, H5E_CANTGET, FAIL, "H5O_get_native__info failed");
+
+            if (H5O_get_version_bound(f->shared->low_bound, &version) < 0)
+                HGOTO_ERROR(H5E_FILE, H5E_CANTGET, FAIL, "H5O_get_version_bound failed");
+
+            if (ninfo.hdr.version < version)
```
qkoziol, line 3809: The code and error message seem to be out-of-sync.  Which is correct?

vchoi-hdfgroup: Fixed the error message.

qkoziol (approved):

markcmiller86:
So, can I just confirm something about this fix. Does it mean that when running on 1.14, if a `fapl` that specifies a `LATEST` for both upper and lower bound is used in the `H5Fcreate` call, it will not wind up creating a file that cannot be opened by say `v18`?

I mean does the fix in this PR make 1.14's `H5FCreate` with `LATEST` in upper/lower has same effect as `H5FCreate` with `DEFAULT` fapl followed by `H5Pset_libver_bounds` with `LATEST` in upper/lower? That is kinda sorta the key question. In both cases, the fapl *should* (at least in my interpretation of this functionality) effect *access* to the objects in the file and not whether the file can actuall succesfully open or not.

fortnern:
> So, can I just confirm something about this fix. Does it mean that when running on 1.14, if a `fapl` that specifies a `LATEST` for both upper and lower bound is used in the `H5Fcreate` call, it will not wind up creating a file that cannot be opened by say `v18`?

The library may create a file that cannot be opened by 1.8 if the upper bound is anything above V18. If the lower bound is not above V18 and you are careful not to use any features that require newer file formats then the file will be readable by 1.8 even if the upper bound is higher.

> I mean does the fix in this PR make 1.14's `H5FCreate` with `LATEST` in upper/lower has same effect as `H5FCreate` with `DEFAULT` fapl followed by `H5Pset_libver_bounds` with `LATEST` in upper/lower? That is kinda sorta the key question. In both cases, the fapl _should_ (at least in my interpretation of this functionality) effect _access_ to the objects in the file and not whether the file can actuall succesfully open or not.

My understanding is that this fix eliminates a check for the superblock version that is not necessary in non-SWMR mode. In SWMR write mode, when opening the file, the library needs to write a message to the superblock, and that message is only supported in newer superblock versions, so the library needs to upgrade the superblock if it is on the older format. If it is prevented from upgrading the superblock by the libver bounds setting then it must issue an error (and it still does). The bug was it would check for this ability even if not in SWMR write mode. The libver bounds setting only affects newly created objects, it does not prevent reading previously created objects, no matter the format version, unless there's a bug like this one.

PR #4942, opened 2024-10-09
Refactor allocation of API context

qkoziol:
Since each API context is local to a thread, use the stack to store the context instead of allocating & releasing it each time.  This improves performance (slightly), reduces alloc/free calls, and eliminates the H5FL package from the push & pop operations, which helps simplify threadsafe operation.

One effect of this change is that the H5VLstart_lib_state / H5VLfinish_lib_state API routines for pass through connector authors now require a parameter that can be used to store the library's context.  It was probably a mistake to assume that these two routines would not do this previously, so this is essentially a bug fix for them.

Some other minor things:
- Added API context push+pop operations to cache tests (I'm not actually certain why this was working before) and a few other places
- Cleaned up a bunch of warnings in test code (calloc args, mainly)
- Made header file inclusions more standard in some source files


src/H5CXprivate.h
```diff
@@ -47,6 +50,236 @@ typedef struct H5CX_state_t {
...
+    bool  vol_connector_prop_valid;           /* Whether property for VOL connector ID & info is valid */
+    void *vol_wrap_ctx;                       /* VOL connector's "wrap context" for creating IDs */
+    bool  vol_wrap_ctx_valid; /* Whether VOL connector's "wrap context" for creating IDs is valid */
+} H5CX_t;
+
+/* Typedef for nodes on the API context stack */
+/* Each entry into the library through an API routine invokes H5CX_push()
+ * in a FUNC_ENTER_API* macro, which pushes an H5CX_node_t on the API
+ * context [thread-local] stack, after initializing it with default values
+ * in H5CX__push_common().
```
mattjala: Comment references deleted routine `H5CX__push_common()`

src/H5VL.c
```diff
@@ -771,26 +771,33 @@ H5VLretrieve_lib_state(void **state /*out*/)
 /*---------------------------------------------------------------------------
  * Function:    H5VLstart_lib_state
  *
- * Purpose:     Opens a new internal state for the HDF5 library.
+ * Purpose:     Opens a new internal context for the HDF5 library.
```
mattjala: It would be good to explicitly mention that the modified buffer here should eventually be passed to `finish_state`.

test/vol.c
```diff
@@ -1780,7 +1781,7 @@ exercise_reg_opt_oper(hid_t fake_vol_id, hid_t reg_opt_vol_id, H5VL_subclass_t s
     /* Push a new API context on the stack */
     /* (Necessary for the named datatype construction routines) */
     if (H5VL_SUBCLS_DATATYPE == subcls)
-        H5CX_push();
+        H5CX_push(&api_ctx);
```
mattjala, line 1785: It looks like this context might reuse values from the one pushed at `vol.c:1728`

src/H5CXprivate.h
qkoziol: Updated in next push

src/H5VL.c
qkoziol: Included in next push.

test/vol.c
qkoziol: No, the context pushed at vol.c:1728 is popped on vol.c:1736.

mattjala: True, but as I understand it the block of memory isn't modified by `H5CX_pop()` (beyond potentially setting a few context return values), it's just removed from the context stack. Any modified values would persist.

mattjala (approved):

hyoklee (approved):

src/H5CXprivate.h
```diff
@@ -47,6 +50,235 @@ typedef struct H5CX_state_t {
...
+ *
+ *      These properties have both a 'valid' and 'set' flag. <foo>_valid is true if the field has ever been
+ *      populated from its underlying property list. <foo>_set flag is true if this field has ever been set on
+ *      the context for application introspection. The naming of these fields is important for the
+ *      H5CX_RETRIEVE_PROP_VALID_SET macro to work properly.
+ *
+ *      If a field has been set on the context but never read internally, <foo>_valid will be false
+ *      despite the context containing a meaningful cached value.
+ */
+typedef struct H5CX_t {
```
jhendersonHDF, line 97: An interesting point I hadn't considered but may be a problem in the future - the size of this struct may soon grow large enough to cause stack warnings.

test/vol.c
qkoziol: Yes, I agree with you, I'll memset it back to zeros for this (unusual) case.

src/H5CXprivate.h
qkoziol: I thought about that too, and I checked the size: ~600 bytes currently.   So, we should have a lot of margin for growth in the future.  If it gets more than 10x bigger, we'll have to think about it again.  :-)

derobins (approved):

PR #4968, opened 2024-10-17
Remove another couple of H5E_clear_stack calls

qkoziol:
And clean up the links test, which had a ton of copy-and-pasted incorrect negative error checks.   (Calling TEST_ERROR within H5E_BEGIN_TRY / H5E_END_TRY pairs will `goto` out of the pair, leaving errors suppressed)


mattjala (approved):

qkoziol:
Note that this change corrects a failure in H5Oexists / H5Oexists_by_name, which should have returned 'false' and not failed.   (This is why the Julia test is failing, since it is checking for the failure instead of false)

derobins (approved):

PR #4969, opened 2024-10-18
Remove TRUE and FALSE from H5private.h

derobins:


derobins:
We purged hbool_t/TRUE/FALSE in lieu of bool/true/false a long time ago. There's no reason to keep TRUE/FALSE macros around (though note that Windows still defines them in Win32).

byrnHDF (approved):

mattjala (approved):

qkoziol (approved):

PR #4979, opened 2024-10-19
Remove some Solaris Studio work-arounds

derobins:
Solaris Studio hasn't been updated in almost a decade and the last version (12.4, circa 2015) doesn't seem to fully support C11.

This PR removes some work-arounds for things like __attribute__() support.

### DIFF
--- a/config/cmake/ConfigureChecks.cmake
+++ b/config/cmake/ConfigureChecks.cmake
@@ -29,11 +29,6 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set (${HDF_PREFIX}_HAVE_DARWIN 1)
 endif ()
 
-# Check for Solaris
-if (${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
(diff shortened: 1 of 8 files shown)


derobins:
H5LTparse.c and H5LTanalyze.c are generated files, but I decided to not regenerate them since the only thing that changed was the warning block we add from genparser.

config/cmake/ConfigureChecks.cmake
```diff
@@ -29,11 +29,6 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set (${HDF_PREFIX}_HAVE_DARWIN 1)
 endif ()
 
-# Check for Solaris
-if (${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
-    set (${HDF_PREFIX}_HAVE_SOLARIS 1)
-endif ()
```
derobins, line 35: H5_HAVE_SOLARIS is unused in the library

mattjala (approved):

bmribler (approved):

qkoziol (approved):

PR #4980, opened 2024-10-19
Always check for __attribute__() in CMake

derobins:
The CMake compiler checks skip checking for things like __attribute__() on Windows. Now that Visual Studio can use clang, we should be checking for this, even on non-MinGW Windows.


derobins:
This should dramatically reduce the number of warnings we see in the Windows builds in CDash since we use __attribute__() to suppress fall-through, unused parameters, etc.

mattjala (approved):

qkoziol (approved):

PR #4981, opened 2024-10-19
Check in generated files in src

derobins:
These files are infrequently updated and generating them adds an annoying dependency on Perl. We're checking them in and will probably add a GitHub action to check if anything is stale when creating a PR.

Adds:
    * H5Edefin.h
    * H5Einit.h
    * H5Emajdef.h
    * H5Emindef.h
    * H5Epubgen.h
    * H5Eterm.h
    * H5overflow.h
    * H5version.h

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,15 +32,7 @@ m4/ltoptions.m4
 m4/ltsugar.m4
 m4/ltversion.m4
 m4/lt~obsolete.m4
-src/H5Edefin.h
-src/H5Einit.h
-src/H5Emajdef.h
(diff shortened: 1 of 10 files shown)


derobins:
Adding these back in allows us to more easily test MinGW, which is a platform we care about.

derobins:
I'll add a GitHub issue to check for freshness at a later date

mattjala (approved):

bmribler (approved):
Thank you!

qkoziol (approved):
Still have a dependency on Perl, ultimately...

PR #4982, opened 2024-10-19
Prefix DEFAULT_API_VERSION with HDF5_ in CMake

derobins:
All other HDF5 library CMake options are prefixed with HDF5_

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -857,9 +857,16 @@ if (HDF5_ENABLE_SUBFILING_VFD)
   endif()
 endif()
 
+# Select the default API version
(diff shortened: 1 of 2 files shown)


derobins:
I did not bother to update the versions to 2.0, etc. since that should be done in a separate PR

qkoziol (approved):

bmribler (approved):

byrnHDF (changes requested):
DEFAULT_API_VERSION has at least 43 references in code that needs updated.

mattjala (approved):

byrnHDF:
Superceded by #4990

PR #4983, opened 2024-10-19
Update CI to use gcc-14

derobins:


derobins:
I'll probably do the warnings in separate PRs before this will be ready to merge

bmribler (approved):

qkoziol (approved):

bmribler (approved):

hl/src/H5TB.c
```diff
@@ -3185,7 +3185,7 @@ H5TB_create_type(hid_t loc_id, const char *dset_name, size_t type_size, const si
     if (H5TBget_table_info(loc_id, dset_name, &nfields, NULL) < 0)
         goto out;
 
-    if (NULL == (fnames = (char **)calloc(sizeof(char *), (size_t)nfields)))
+    if (NULL == (fnames = (char **)calloc((size_t)nfields), sizeof(char *)))
```
lrknox:
```suggestion
    if (NULL == (fnames = (char **)calloc((size_t)nfields, sizeof(char *))))
```

lrknox: Misplaced ')' causes build error.

hl/src/H5TB.c
```diff
@@ -3185,7 +3185,7 @@ H5TB_create_type(hid_t loc_id, const char *dset_name, size_t type_size, const si
     if (H5TBget_table_info(loc_id, dset_name, &nfields, NULL) < 0)
         goto out;
 
-    if (NULL == (fnames = (char **)calloc(sizeof(char *), (size_t)nfields)))
+    if (NULL == (fnames = (char **)calloc((size_t)nfields, sizeof(char *)))
```
lrknox:
```suggestion
    if (NULL == (fnames = (char **)calloc((size_t)nfields, sizeof(char *))))
```

.github/workflows/abi-report.yml
```diff
@@ -28,7 +28,7 @@ permissions:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
byrnHDF, line 31: Should we not specify latest and always use a specific version on all CI?

bmribler (approved):

.github/workflows/abi-report.yml
derobins: I definitely prefer that, but ubuntu-latest is 22.04, which does not have gcc-14

derobins:
I'll recreate when we've fixed the warning issues

PR #4989, opened 2024-10-21
Replace COPYING with LICENSE in workflows

byrnHDF:
Finish replacing COPYING* with LICENSE* - mostly in the workflows


derobins (changes requested):
The HISTORY files shouldn't be updated

byrnHDF:
> The HISTORY files shouldn't be updated

Completed

derobins (approved):

PR #4991, opened 2024-10-21
Add msys2 workflow for CMake

byrnHDF:


.github/workflows/msys2-cmake.yml
```diff
@@ -0,0 +1,85 @@
...
+            $GITHUB_WORKSPACE
+
+      - name: CMake Build
+        run: |
+          cmake --build . --parallel 3 --config ${{ inputs.build_mode }}
+        working-directory: ${{ runner.workspace }}/build
+
+      - name: CMake Run Tests
+        run: |
+          ctest . --parallel 2 -C ${{ inputs.build_mode }} -V -E "tfloatsattrs|testhdf5-base|cache_api|dt_arith|dtypes|err_compat"
```
jhendersonHDF: Does this action report to cdash? Otherwise, excluding tests just to get this to pass is 100% going to be forgotten about for a long time.

byrnHDF: NO but this is not the only one that does this.

byrnHDF: Maybe we should have an ENV option that tests all the test to verify they are still failing?

byrnHDF: Only daily builds report to CDash

jhendersonHDF: It may be more appropriate for msys2 to be a daily build anyway, but I wonder if we can do something with splitting out the failing tests into a separate step with `continue-on-error: true` and report the results somewhere.

byrnHDF: I could add a next step that just runs those and add continue on error

jhendersonHDF: Right, that's the idea; I'm just wondering if we can have those test results reporting somewhere so that we can keep track of when they start passing/if they're still failing.

byrnHDF: The other one is the intel cmake workflow
Those failing tests will show up in the workflow individual - maybe there is a action for email?

jhendersonHDF: > Those failing tests will show up in the workflow individual

Sure, I'm just thinking that if the workflow is always passing, no one will think to ever go and check the workflow status again.

.github/workflows/nvhpc-cmake.yml
```diff
@@ -76,6 +76,13 @@ jobs:
           ctest . -E "MPI_TEST|H5TEST-dt_arith|H5TEST-dtransform" --parallel 2 -C ${{ inputs.build_mode }} -V
         working-directory: ${{ runner.workspace }}/build
 
+      - name: CMake Run Expected To Fail Tests
+        shell: bash
+        run: |
+          ctest . -R "H5TEST-dt_arith|H5TEST-dtransform" --parallel 2 -C ${{ inputs.build_mode }} -V
+        working-directory: ${{ runner.workspace }}/build
+        continue-on-error: true
+
```
derobins: Please put this in a separate PR

derobins (changes requested):
The nvhpc change needs to go in a separate PR

.github/workflows/nvhpc-cmake.yml
byrnHDF: done

derobins (approved):

jhendersonHDF (approved):

PR #4993, opened 2024-10-22
Update nvhpc version to 24.9

hyoklee:
NVHPC 24.9 is released in September.


derobins:
I would change this PR to just update CMake. We're going to drop the Autotools after SC24, so I'm not concerned with the Fortran failures.

derobins:
I removed the install check since we're dropping the Autotools. Figuring out why that fails would be a waste of time.

derobins (approved):

byrnHDF (approved):

PR #4994, opened 2024-10-22
Add full-text search to Doxygen

gheber:
Restore fulltext search based on Google PSE.

### DIFF
--- a/doxygen/dox/FTS.dox
+++ b/doxygen/dox/FTS.dox
@@ -1,7 +1,7 @@
 /** \page FTS Full-Text Search
 
 \htmlonly
-<script async src="https://cse.google.com/cse.js?cx=75c754173f9b90804"></script>
+<script async src="https://cse.google.com/cse.js?cx=f671574fa55cc44e6"></script>
 <div class="gcse-search"></div>
 \endhtmlonly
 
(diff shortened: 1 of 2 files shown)


loricooperhdf:
I would rather replace the corner doxygen-generated search box with the fully functional google search box.

brtnfld:
If I'm searching in the RM, I don't want search results outside the RM. Maybe that is just me. Is that what this accomplishes? Or the default is to search the RM but have the option to do an expanded search.

loricooperhdf:
> If I'm searching in the RM, I don't want search results outside the RM. Maybe that is just me. Is that what this accomplishes? Or the default is to search the RM but have the option to do an expanded search.

No--Gerd's PR here adds a link in the sidebar for Full Text Search. I set up that search to look at the following "directories" and everything under them. For the record, it's here: https://cse.google.com/cse?cx=f671574fa55cc44e6 
- https://support.hdfgroup.org/documentation/
- https://support.hdfgroup.org/downloads/
- https://support.hdfgroup.org/documentation/hdf5/latest/

I would like to change the box in the upper right to function the same way. I think people believe that a search box looks at all of the content on a website. No PR for that yet.

We COULD provide search boxes in certain places, like somewhere in the RM that searches RM only and specifically states that usage. Let's discuss at an upcoming Monday meeting.

hyoklee (approved):

byrnHDF (approved):

PR #4996, opened 2024-10-22
Add 32-bit CI test for linux

### LINKED ISSUE #4921: Add 32-bit CI test for linux
Add to GitHub actions/workflow testing on a 32-bit OS, such as alpine Linux,  https://github.com/marketplace/actions/setup-alpine-linux-environment

hyoklee:
close #4921

### DIFF
--- a/.github/workflows/i386-alp.yml
+++ b/.github/workflows/i386-alp.yml
@@ -1,0 +1,38 @@
+name: i386 alpine
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "6 0 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: setup alpine
+        uses: jirutka/setup-alpine@v1
+        with:
+          arch: x86
+          packages: >
+            build-base
+            libaec-dev
+            libgit2-dev
+            cmake
+      - name: run script
+        run: |
+          mkdir build
+          cd build
+          cmake \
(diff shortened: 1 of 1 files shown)


.github/workflows/i386-alp.yml
```diff
@@ -0,0 +1,38 @@
+name: i386 alpine
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "6 0 * * *"
```
derobins, line 6: Why cron and not PR? Does it take a long time to run? Also, anything that isn't a PR check should probably get a badge so it's easy to see when it fails.

derobins (approved):
Approved. This is mainly so we can check out warnings on CDash.

.github/workflows/i386-alp.yml
byrnHDF: Can we add this to the daily-build workflow instead?

.github/workflows/i386-alp.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: jirutka/setup-alpine@v1
+        with:
+          arch: x86
+          packages: >
+            build-base
+            libaec-dev
+            libgit2-dev
+            cmake
+      - name: run script
+        run: |
```
byrnHDF, line 27: It would be better to follow the cmake-script.yml file as a template instead of reinventing the workflows every time! Makes it so much harder to maintain.

byrnHDF (changes requested):
Please use the existing workflows as templates (suggest improvements)

hyoklee:
#5081

PR #5007, opened 2024-10-25
Fix h5py CI failure

hyoklee:
Update minimal hdf5 version to 10.6 for Python 3.12,

Ref: https://github.com/spack/spack/pull/47132/files#diff-e14a9be8ba65e358b688fb91ac00a931d029c34c0db4f9384ea5f50cffee8444

### DIFF
--- a/.github/workflows/h5py.yml
+++ b/.github/workflows/h5py.yml
@@ -21,7 +21,7 @@ jobs:
 
     - name: Run a multi-line script
       run: |
-        sed -i 's/hdf5@1.10.4:1.14/hdf5@1.10.4:/g' \
+        sed -i 's/hdf5@1.10.6:1.14/hdf5@1.10.6:/g' \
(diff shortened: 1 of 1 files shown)


derobins (approved):

.github/workflows/h5py.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     - name: Run a multi-line script
       run: |
-        sed -i 's/hdf5@1.10.4:1.14/hdf5@1.10.4:/g' \
+        sed -i 's/hdf5@1.10.6:1.14/hdf5@1.10.6:/g' \
```
byrnHDF, line 24: What exactly does this step do?

jhendersonHDF (approved):

.github/workflows/h5py.yml
lrknox: I think this extends the required range for hdf5 from 1.10.6 - 1.10.14 to any version 1.10.6 or greater.

PR #5008, opened 2024-10-25
Remove Perl from CMake

hyoklee:
See also #5005.


byrnHDF:
Why remove this? The GENERATE_HEADERS option is still available.
AND there are tests in test folder that use perl

byrnHDF (changes requested):
Cannot be committed - there are still requirements for perl

hyoklee:
v2.0.0 doesn't remove Perl dependency.

PR #5009, opened 2024-10-25
Get the Autotools version number from H5public.h

derobins:
The version number in the Autotools is scraped from README.md in configure.ac, which makes README.md count as 'code', which is a problem since changes to markdown files are ignored in CI.

This change scrapes the version number from H5public.h instead of README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-HDF5 version 1.17.0 currently under development
+HDF5 version 2.0.0 currently under development
 
 ![HDF5 Logo](doxygen/img/HDF5.png)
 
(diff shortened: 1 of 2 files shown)


derobins:
We still haven't updated to v2.0.0 throughout the library yet, so we won't change the version in H5public.h in this PR. I just changed it in README.md since that should be harmless now.

README.md
```diff
@@ -1,4 +1,4 @@
-HDF5 version 1.17.0 currently under development
+HDF5 version 2.0.0 currently under development
```
lrknox, line 1: Generally h5vers changes this along with the rest of the version settings, but as you say, probably harmless now.

lrknox (approved):

byrnHDF (approved):

hyoklee (approved):

Issue #4949, opened 2024-10-10
Build breaks with libszip-2.1.1: undefined reference due to --no-allow-shlib-undefined: SZ_encoder_enabled


yurivict:
**Describe the bug**
```
ld: error: undefined reference due to --no-allow-shlib-undefined: SZ_encoder_enabled
>>> referenced by ../bin/libhdf5.so.310.5.0

ld: error: undefined reference due to --no-allow-shlib-undefined: SZ_BufftoBuffCompress
>>> referenced by ../bin/libhdf5.so.310.5.0

ld: error: undefined reference due to --no-allow-shlib-undefined: SZ_BufftoBuffDecompress
>>> referenced by ../bin/libhdf5.so.310.5.0
cc: error: linker command failed with exit code 1 (use -v to see invocation)
--- java/src/hdf/hdf5lib/CMakeFiles/jarhdf5.dir/all ---
--- java/src/hdf/hdf5lib/jarhdf5-1.14.5.jar ---
--- test/CMakeFiles/testhdf5.dir/all ---
[ 73%] Building C object test/CMakeFiles/testhdf5.dir/testhdf5.c.o
[...]

PR #5010, opened 2024-10-25
Fix CMake build detection of `libaec`

cho-m:
Partial revert of https://github.com/HDFGroup/hdf5/pull/4567 as libaec is the correct name of the package and should be searched first.

Also fixes pkg-config generation handling. This does mean the output CMake configuration files need to `find_package(libaec)` prior to `hdf5-static`, but it also avoids using absolute path to shared library so that users can do `set(libaec_USE_STATIC_LIBS ON)`.

---

`libaec` package installs `libaec-config.cmake`.

CMake does not provide a `FindSZIP.cmake` module, internal copy was removed (https://github.com/HDFGroup/hdf5/commit/2c3797d7bdb397c25f82f25be6bbabe921e42255), and older SZIP didn't include CMake config files. So, could consider dropping older `find_package(SZIP ...)` as the only valid one is now `find_package(libaec ...)`.

Resulting `hdf5.pc` has:
```pc
Libs.private:   -lz -lsz
```

Resulting `hdf5-targets.cmake` has:
```cmake
set_target_properties(hdf5-static PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:m>;\$<LINK_ONLY:dl>;\$<LINK_ONLY:ZLIB::ZLIB>;\$<LINK_ONLY:libaec::sz>;\$<\$<PLATFORM_ID:Windows>:>;\$<\$<NOT:\$<PLATFORM_ID:Windows>>:>;\$<\$<BOOL:OFF>:MPI::MPI_C>;\$<LINK_ONLY:\$<\$<OR:\$<BOOL:OFF>,\$<BOOL:OFF>>:Threads::Threads>>"
)
```

---

Related to #4949

Note that Debian does not include `libaec` CMake files so they would need a different fix (though maybe better for them to support CMake files).


CMakeFilters.cmake
```diff
@@ -160,15 +160,27 @@ if (HDF5_ENABLE_SZIP_SUPPORT)
     endif ()
     set(libaec_USE_STATIC_LIBS ${HDF5_USE_LIBAEC_STATIC})
     set(SZIP_FOUND FALSE)
-    find_package (SZIP NAMES ${LIBAEC_PACKAGE_NAME}${HDF_PACKAGE_EXT} COMPONENTS ${LIBAEC_SEACH_TYPE})
+    find_package (libaec CONFIG)
```
byrnHDF: original szip could still be used with previous but probably need to make sure that LIBAEC_PACKAGE_NAME matches libaec otherwise might be a mismatch if LIBAEC_PACKAGE_NAME is used later and doesn't match

CMakeFilters.cmake
```diff
@@ -160,15 +160,27 @@ if (HDF5_ENABLE_SZIP_SUPPORT)
...
+      if (NOT SZIP_FOUND)
+        find_package (SZIP) # Legacy find
+      endif ()
     endif ()
     set(H5_SZIP_FOUND ${SZIP_FOUND})
     if (H5_SZIP_FOUND)
       set (H5_SZIP_INCLUDE_DIR_GEN ${SZIP_INCLUDE_DIR})
       set (H5_SZIP_INCLUDE_DIRS ${H5_SZIP_INCLUDE_DIRS} ${SZIP_INCLUDE_DIR})
-      set (LINK_COMP_LIBS ${LINK_COMP_LIBS} ${SZIP_LIBRARIES})
+      if (TARGET libaec::sz)
```
byrnHDF, line 175: not always sz (like on windows)

byrnHDF: Should avoid using hardcodded names - always if possible

byrnHDF (changes requested):
Hardcoded names should not be used

CMakeFilters.cmake
cho-m: What CMake file is outputting something different on Windows? Can you provide reference for this?

The target names are created via `libaec-config.cmake` https://gitlab.dkrz.de/k202009/libaec/-/blob/master/cmake/libaec-config.cmake.in?ref_type=heads#L25.

If target doesn't exist, it should fall back on prior logic for `else()` case. 

This is only way to get both static and shared support. If you use `SZIP_LIBRARIES`, it will hardcode the specific library found `.a`, `.lib`, `.dll`, `.dylib`, or `.so`

---

With prior logic, the resulting HDF5 cmake config file (if built with shared `libaec`) will cause:
```cmake
set(libaec_USE_STATIC_LIBS ON)
find_package(ZLIB REQURIED)
find_package(libaec CONFIG REQURIED)
find_package(hdf5 CONFIG REQURIED)
add_executable(test test.cpp)
target_link_libraries(test hdf5-static)
```

to run `(c|clang|g)++ test.cpp ... /path/to/libhdf5.a /path/to/libsz.{so,dylib}`.

Whereas most users would expect: `(c|clang|g)++ test.cpp ... /path/to/libhdf5.a /path/to/libsz.a`.

byrnHDF: hdf5 only needs static compression libs

cho-m: > hdf5 only needs static compression libs

But most external distribution require shared libraries and do not allow bundling libraries, e.g.

* Alpine - only shared - https://pkgs.alpinelinux.org/contents?name=libsz&repo=community&branch=edge&arch=x86_64
* Arch Linux - only shared - https://archlinux.org/packages/extra/x86_64/libaec/
* Debian - only shared - https://packages.debian.org/sid/amd64/libsz2/filelist
* ...

---

The default externally will always be shared. If static libraries are provided, then ideally should work as users expect.

Otherwise, most repositories/distros will just drop support for SZIP in HDF5. If that is the recommendation, can just go this route and close PR.

CMakeFilters.cmake
cho-m: Mismatching names are not recommended, which is why CMake will output warning https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPackageHandleStandardArgs.cmake#L441-446

This also breaks `libaec` detection due to https://gitlab.dkrz.de/k202009/libaec/-/blob/master/cmake/libaec-config.cmake.in?ref_type=heads#L49 (also COMPONENTS are not defined in `libaec`).

---

Original szip should still be supported by fallbacks. As long as you don't add `REQUIRED` (and there is no conflicting `libaec` somewhere) it should always work.

If you don't want output warnings, then the recommendation is usually running with `QUIET` and only outputting warning for last attempt.

If there is equal preference, then a dedicated CMake option would be better to control priority

CMakeFilters.cmake
byrnHDF: hdf5 doesn't expose the API for compression, the reason for shared - static lib is only provided for linking the static hdf5. So if there is no static hdf5, then there is not a need for providing a static compression lib.
Also, the compression libraries do not require inline build with hdf5 - they be can be built externally and just use the proper include and lib paths to pull them into hdf5

cho-m: Anyway, without modification, existing logic is broken with `libaec` and will result in 
```
CMake Error at src/CMakeLists.txt:1204 (get_target_property):
  get_target_property() called with non-existent target
```

cho-m: > So if there is no static hdf5, then there is not a need for providing a static compression lib.

In Homebrew's case, we have both. We distribute a shared and static HDF5 from single build. We always default to shared as main library.

User expectation here is that each would behave as expected, depending on if they use `hdf5-shared` or `hdf5-static` target (or [FindHDF5.cmake](https://cmake.org/cmake/help/latest/module/FindHDF5.html) - `HDF5_USE_STATIC_LIBRARIES=[ON|OFF]`)


---

> Also, the compression libraries do not require inline build with hdf5 - they be can be built externally and just use the proper include and lib paths to pull them into hdf5

That is what we've done in Homebrew until 1.14.5 introduced a build regression. We have not updated to HDF5 1.14.5 yet due to these issues.

Issue was also reported by FreeBSD. It will impact pretty much all repositories/distros that use CMake build. Linux distros are pretty much all using autotools so won't be impacted until v2.

cho-m:
> Hardcoded names should not be used

Root causing further, the change in https://github.com/HDFGroup/hdf5/pull/4567 and https://github.com/HDFGroup/hdf5/commit/4fa004e73a820b46304c1194f7aa1444584d7597 did something worse than "hardcoding". It relied on unofficial / build-system-specific (vcpkg) patch.

It introduced regression on official builds of `libaec` and `szip`. Even builds with original SZIP are broken without vcpkg patch (https://github.com/microsoft/vcpkg/blob/master/ports/hdf5/szip.patch) as linker flags are missing

EDIT 1: Further regressions in `libaec` are from expecting the same `find_package` behavior. There are no COMPONENTS defined in `libaec`.

EDIT 2: Given various breakages, would recommend adding a CMake build test with system `libaec` on Linux CI, but the limitation here is lack of `libaec` CMake files on Ubuntu. You would need to set up another CI runner on Fedora, Arch Linux, Alpine, or something else.

---

So, yes, relying on Vcpkg's "hardcoded" workaround should be avoided. Overriding `SZIP_LIBRARIES` (and thus causing confusion on intended output) should be avoided.

Also using `SZIP_LIBRARIES` as a target when it is not guaranteed to be so is also a bad idea.

Instead, the correct targets should be used maybe like
```cmake
if (TARGET libaec::sz)
  set(SZIP_TARGET libaec::sz)
elseif (TARGET szip-shared)
  set(SZIP_TARGET szip-shared)
elseif (TARGET szip-static)
  set(SZIP_TARGET szip-static)
endif()
```
Or
```cmake
set (SZIP_TARGET "szip-${LIBAEC_SEACH_TYPE}")
if (TARGET libaec::sz)
  set (SZIP_TARGET libaec::sz)
endif ()
```

Since the logic has to plug into the pkgconfig handler which needs CMake targets (`SZIP_LIBRARIES` is not one):
https://github.com/HDFGroup/hdf5/blob/ee61dd51ac3b3eda721dc91eab03685f4b4d029c/src/CMakeLists.txt#L1225-L1229

Otherwise you get:
```
Libs.private:   -lz -llibname-NOTFOUND
```

hyoklee (approved):

byrnHDF:
I will be investigating all the issues with the zlib and szip functionality in the HDF5 CMake code. There are a variety of ways to build hdf5 with zlib/szip support and it's time to re-evaluate that code before the end of the year.

byrnHDF:
Please check #5182 - I think it should solve the problem.

byrnHDF:
1.14.6 and develop was reworked and this change would be in conflict with the changes.
If there is still a problem please create an issue.

PR #5011, opened 2024-10-25
Remove version number from h5repack test plugin

derobins:
The h5repack test uses an older copy of a plugin from test that appended the version number to group names. The plugin in test was updated to use a fixed string instead, but the h5repack copy wasn't updated.

We'll remove this copy and just use the plugin from test after the Autotools are removed.

### DIFF
--- a/tools/test/h5repack/dynlib_vrpk.c
+++ b/tools/test/h5repack/dynlib_vrpk.c
@@ -9,31 +9,34 @@
  * If you do not have access to either file, you may request a copy from     *
  * help@hdfgroup.org.                                                        *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
 /*
- * Purpose:    Tests the plugin module (H5PL)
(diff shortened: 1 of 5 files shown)


byrnHDF (approved):

tools/test/h5repack/dynlib_vrpk.c
```diff
@@ -9,31 +9,34 @@
  * If you do not have access to either file, you may request a copy from     *
  * help@hdfgroup.org.                                                        *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
 /*
- * Purpose:    Tests the plugin module (H5PL)
+ * Purpose:	Test group filter plugin for the filter_pluging.c test.
```
lrknox, line 14:
```suggestion
 * Purpose:	Test group filter plugin for the filter_plugin.c test.
```

lrknox (approved):
Probably one non-critical typo.

PR #5012, opened 2024-10-25
Yank duplicate sec2 vfd variable

derobins:


derobins:
Stupid wrong branch...

PR #5014, opened 2024-10-25
Set C11 standard in CMake examples

hyoklee:
See also #4975.


derobins (approved):

HDF5Examples/config/cmake/HDFExampleMacros.cmake
```diff
@@ -31,13 +31,13 @@ macro (BASIC_SETTINGS varname)
     set (CMAKE_MFC_FLAG 0)
   endif ()
 
-  set (CMAKE_C_STANDARD 99)
+  set (CMAKE_C_STANDARD 11)
```
byrnHDF, line 34: The examples need to work for all versions of the library - should this be conditional on the lib version?

config/cmake/HDFCompilerCXXFlags.cmake
```diff
@@ -32,7 +32,7 @@ endif ()
 if (CMAKE_CXX_COMPILER_ID STREQUAL SunPro AND CMAKE_CXX_COMPILER_LOADED)
   if (NOT DEFINED CMAKE_CXX${CMAKE_CXX_STANDARD}_STANDARD_COMPILE_OPTION)
     if (NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.13)
-      if (NOT CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD EQUAL 98)
+      if (NOT CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD EQUAL 11)
```
byrnHDF, line 35: This is fine for the library.

byrnHDF (changes requested):
Need to determine if this is a problem for older library versions.

HDF5Examples/config/cmake/HDFExampleMacros.cmake
derobins: We can leave it at 99. I don't think the current examples do anything special for C11.

fortnern:
Closed due to consensus at standup. Examples should not require C11

PR #5016, opened 2024-10-27
Fix remaining calloc param reversals

derobins:
gcc 14 identifies likely places where the sizeof() call is used for the first parameter to calloc(). This PR fixes the remaining places in the library where this occurs and reworks a few other allocations to be more uniform.


derobins:
Autogen removes files!

PR #5020, opened 2024-10-28
Fix bad private FUNC_ENTER macros

derobins:
The name check in the FUNC_ENTER checks for private FUNC_ENTER macros (which have NOAPI in the name) only checked for not having the public form (H5Xfoo), so they could be inappropriately applied to package-level functions (H5X__foo).

This PR fixes the bug and updates the inappropriate macros.


src/H5Fprivate.h
```diff
@@ -124,7 +124,7 @@ typedef struct H5F_t H5F_t;
 #define H5F_HAS_FEATURE(F, FL)           (H5F_has_feature(F, FL))
 #define H5F_BASE_ADDR(F)                 (H5F_get_base_addr(F))
 #define H5F_SYM_LEAF_K(F)                (H5F_sym_leaf_k(F))
-#define H5F_KVALUE(F, T)                 (H5F_Kvalue(F, T))
+#define H5F_KVALUE(F, T)                 (H5F_kvalue(F, T))
```
derobins, line 127: The capital K causes problems with the name check macros

src/H5Oshared.h
```diff
@@ -42,7 +42,8 @@ H5O_SHARED_DECODE(H5F_t *f, H5O_t *open_oh, unsigned mesg_flags, unsigned *iofla
 {
     void *ret_value = NULL; /* Return value */
 
-    FUNC_ENTER_NOAPI_NOINIT
+    /* H5O_SHARED_ENCODE will be defined as a package function */
```
qkoziol: decode

qkoziol (approved):
Nice!

mattjala (approved):

src/H5Oshared.h
derobins: Fixed!

PR #5025, opened 2024-10-28
Add long double format option to h5dump

byrnHDF:
In trying to fix tfloatattrs, discovered that the floating point format option did not apply to long double datatypes.
This adds the missing option.


derobins:
Needs a release note!

lrknox (approved):

derobins (approved):

mattjala (approved):

tools/test/h5dump/expected/tfloatsattrs.ddl
jhendersonHDF, line 1: @byrnHDF Was this file updated correctly? It looks like the `--lformat` wasn't applied; the data still has varying numbers of digits after the decimal

byrnHDF:
Yes, but I used %.4Lg and not %.04Lg

jhendersonHDF:
> Yes, but I used %.4Lg and not %.04Lg

`%.4Lg` is the correct specifier; it just looks like it wasn't applied

PR #5029, opened 2024-10-29
Remove comment about multi VFD duplication

derobins:
We used to duplicate the warning macros in H5FDmulti.c when they were in
H5private.h. Now that they are in a separate header, we can use them
directly.

### DIFF
--- a/src/H5warnings.h
+++ b/src/H5warnings.h
@@ -19,10 +19,6 @@
 
 /* Macros for enabling/disabling particular GCC / clang warnings
  *
- * These are duplicated in H5FDmulti.c (we don't want to put them in the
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

qkoziol (commented):
Unless we are installing this header, we shouldn't use any non-public header files in the multi (or stdio) VFDs.  They should not depend on anything internal to the library.

derobins:
> Unless we are installing this header, we shouldn't use any non-public header files in the multi (or stdio) VFDs. They should not depend on anything internal to the library.

They are just the warning suppression macros. There's no reason for them to be considered private. I wouldn't make them public, but anyone could use that file.

hyoklee (approved):

PR #5032, opened 2024-10-29
Add parallel cdash script to daily-build workflow

byrnHDF:


.github/workflows/cmake-par-script.yml
```diff
@@ -0,0 +1,230 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install CMake Dependencies (OpenMPI)
+        run: |
+          sudo apt-get update
+          sudo apt-get install ninja-build graphviz curl
+          sudo apt install libssl3 libssl-dev libcurl4 libcurl4-openssl-dev
+          sudo apt install gcc-12 g++-12 gfortran-12
+          sudo apt install libaec0 libaec-dev
+          sudo apt-get install libopenmpi-dev
```
jhendersonHDF: I'm sure it's just from copy pasting, but we really don't need pretty much any of these dependencies other than libopenmpi-dev.

And if we just install the MPI package, then this workflow isn't much different from the other ones that run parallel tests. Ideally we should be building OpenMPI and MPICH from source and then caching the results so we can sometimes skip the long build process if their head commit hasn't changed.

jhendersonHDF (approved):
I will have a follow on PR that switches the source builds over to building OpenMPI and MPICH from source rather than installing from a package manager

lrknox (approved):

PR #5034, opened 2024-10-29
Detect C11 threads in CMake

hyoklee:
v2.0.0 uses C11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,6 @@
 cmake_minimum_required (VERSION 3.18)
+# set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "" FORCE)
+set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "" FORCE)
 project (HDF5 C)
 
 if (POLICY CMP0074)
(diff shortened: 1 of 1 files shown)


jhendersonHDF:
May need input from @qkoziol

derobins:
The reason this is failing is that we're turning Pthreads on/off via the presence of the pthreads.h header, which is checked elsewhere in CMake. If we pick C11 threads here and then also have H5_HAVE_PTHREAD_H set, the ifdefs in H5TS are likely to be confused.

The Autotools are probably fine since it looks like configure.ac is smarter about checking for the Pthreads header, but ideally there should be an H5_USE_POSIX_THREADS symbol that is set independently of H5_HAVE_PTHREADS_H and used in the H5TS code (and the other thread options should be renamed to USE vs HAVE, IMHO). If you do that here, be sure to alert @qkoziol since he'll have to update his branches.

qkoziol:
> The reason this is failing is that we're turning Pthreads on/off via the presence of the pthreads.h header, which is checked elsewhere in CMake. If we pick C11 threads here and then also have H5_HAVE_PTHREAD_H set, the ifdefs in H5TS are likely to be confused.
> 
> The Autotools are probably fine since it looks like configure.ac is smarter about checking for the Pthreads header, but ideally there should be an H5_USE_POSIX_THREADS symbol that is set independently of H5_HAVE_PTHREADS_H and used in the H5TS code (and the other thread options should be renamed to USE vs HAVE, IMHO). If you do that here, be sure to alert @qkoziol since he'll have to update his branches.

I like this idea.   I'm tied up with something else for a few days, but could tackle it later this week or early next, f someone else doesn't get to it first.

seanm:
If I'm not mistaken, even the newest versions of macOS / Xcode do not support C11 threads.

qkoziol:
> If I'm not mistaken, even the newest versions of macOS / Xcode do not support C11 threads.

Agree

qkoziol:
I gave a try at making this work on MacOS last week and ultimately failed.  It may be possible, but it seems difficult.

PR #5058, opened 2024-11-01
Update daily build action to build MPICH and OpenMPI from source

jhendersonHDF:


.github/workflows/build_mpich_source.yml
```diff
@@ -53,8 +56,8 @@ jobs:
 
       # Enable threads=multiple for testing with Subfiling and
       # VOL connectors that require MPI_THREAD_MULTIPLE
-      - name: Install MPICH (GCC) (Production)
-        if: ${{ steps.cache-mpich-ubuntu-gcc.cache-hit != 'true' && (inputs.build_mode != 'debug') }}
```
jhendersonHDF, line 57: Fixed the caching here in both the MPICH and OpenMPI workflows; forgot to add the `.outputs`

.github/workflows/build_mpich_source.yml
```diff
@@ -80,3 +83,13 @@ jobs:
             --enable-threads=multiple
           make -j2
           make install
+
+      - name: Tar MPICH installation to preserve permissions for artifact
+        run: tar -cvf mpich.tar ${{ runner.workspace }}/mpich
```
jhendersonHDF, line 88: Needed as a workaround. The default artifact zipping method doesn't preserve permissions, so GitHub mentions tar'ing first as a workaround.

byrnHDF (approved):

lrknox (approved):

PR #5062, opened 2024-11-02
Ensure C language when requiring MPI

dg0yt:
With PARALLEL enabled, the exported targets depend on `MPI::MPI_C`. However, this target isn't created by `find_package(MPI)` when the consuming project has only language `CXX` enabled. There is also no good diagnostics because `MPI` is found successfully.

(Alternative: Use `find_package(MPI ... COMPONENTS C)`. This would enable proper diagnostics. But simply enabling language `C` is more effective. This might be reconsidered when fully supporting cmake config search by switching to `find_dependency()` instead of `find_package(... REQUIRED)`.)

Noticed while testing vcpkg port gdal with hdf5 being a transitive dependency - the minimal test port had just `CXX` enabled.

This PR: Build successful.

Before this PR:
~~~
CMake Error at /Users/vcpkg/Data/installed/arm64-osx/share/hdf5/hdf5-targets.cmake:61 (set_target_properties):
  The link interface of target "hdf5::hdf5-static" contains:

    MPI::MPI_C

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
~~~

Diagnostics-only alternative:
~~~
CMake Error at /<redacted>/vcpkg/vcpkg/downloads/tools/cmake-3.30.1-linux/cmake-3.30.1-linux-x86_64/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find MPI (missing: MPI_C_FOUND C)

      Reason given by package: MPI component 'C' was requested, but language C is not enabled.  
~~~


hyoklee (approved):

byrnHDF:
While I think this is just fine - the only worry is if there will be any unintended issues. I Can't think of any since the library is designed to be built with "C".

dg0yt:
I acknowledge that there are some restrictions on where `enable_language` can be used:
https://cmake.org/cmake/help/latest/command/enable_language.html
I believe that these cases should be rare for normal use of `find_package`. But OTOH I have seen enough weird cmake code...

derobins (approved):

PR #5063, opened 2024-11-03
Remove top-level structure from MPI source build workflow tar artifacts

jhendersonHDF:


jhendersonHDF:
Missed these changes when creating the branch for https://github.com/HDFGroup/hdf5/pull/5058. This changes to the parent directory before tar'ing the openmpi or mpich directories in order to remove all the top-level hierarchy bits.

derobins (approved):

byrnHDF (approved):

Issue #3091, opened 2023-06-09
file_offset_t < size_t on mingw


mulimoen:
**Describe the bug**
Using the mingw toolchain we encounter a problem in the assertation on the following https://github.com/HDFGroup/hdf5/blob/ac07382ca947de2b936204dc9d60ce9ac245a21f/src/H5FDstdio.c#L354 which crashes the program

**Expected behavior**
The assertation to not trigger a crash.

**Platform**
 - Default windows github runner
 - HDF5 version 1.14.1-2 from source
 - Microsoft Windows Server 2022
 - Mingw-w64 11.2.0
 - CMake 3.26.4

**Additional context**
- A build log can be found here https://github.com/aldanor/hdf5-rust/actions/runs/5225628766/jobs/9435350340?pr=243
- Found when trying to use 1.14.1 in [`hdf5-rust`](https://github.com/aldanor/hdf5-rust/pull/243)

derobins:
We don't have great testing for MinGW, but I would like to get that set up as a GitHub action ASAP so we can fix these issues.

aldanor:
@derobins mingw builds are very easy to set up on github actions (you can check hdf5-rust CI as an example).

Would really appreciate if this was looked into since it's currently a blocker for us bundling 1.14 for static builds.

derobins:
@aldanor and/or @mulimoen - shoot me an email at derobins at hdfgroup dot org and we'll get this fixed.

derobins:
Is this still a problem? We build and test MinGW in several environments now and we don't see this assert being tripped.

mulimoen:
We still see the problem: https://github.com/mulimoen/hdf5-rust/actions/runs/11579804898/job/32236915194

derobins:
Let me know if this PR fixes the problem. It should be merged in the next day or so.

https://github.com/HDFGroup/hdf5/pull/5064

Update: It's in. Let me know if this fixes your problem!

bmribler:
Hi @mulimoen @aldanor 
I just want to draw your attention to derobins's message on Nov 3, 2024.  Please let us know if this issue can be closed.  Thanks!

magnusuMET:
Was this fix included in 1.14.6, as the problem is still present there: https://github.com/metno/hdf5-rust/actions/runs/16470318613/job/46557649259?pr=88#step:5:220

derobins:
This was not fixed in 1.14.6, but will be fixed in 2.0.

PR #5064, opened 2024-11-04
Update stdio API calls

derobins:
The stdio VFD set up ftell, etc. equivalents differently than the rest of the library. This PR aligns them with H5private.h and H5win32defs.h and should fix some issues with MinGW. The stdio VFD now also uses HDoff_t, which will be a public symbol in the next major release of HDF5.

In the main library, HDftell was set to be ftell and not ftello. This probably isn't a big deal since the systems where it's used all have 64-bit longs, but we should match types correctly.

### DIFF
--- a/src/H5FDstdio.c
+++ b/src/H5FDstdio.c
@@ -119,47 +119,45 @@ typedef struct H5FD_stdio_t {
 #endif /* H5_HAVE_WIN32_API */
 } H5FD_stdio_t;
 
-/* Use similar structure as in H5private.h by defining Windows stuff first. */
-#ifdef H5_HAVE_WIN32_API
-#ifndef H5_HAVE_MINGW
-#define file_fseek     _fseeki64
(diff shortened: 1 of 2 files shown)


derobins:
NOTE: If this is merged to a future 1.14.6 release, HDoff_t will have to be replaced with a restored file_offset_t.

derobins:
NOTE NOTE: This PR will need to be updated after #5065 goes in.

byrnHDF (approved):

qkoziol (approved):

bmribler:
> NOTE NOTE: This PR will need to be updated after #5065 goes in.

@derobins Hi Dana, #5065 is closed un-merged, so does it affect #5064 in any way, merged or unmerged?  A user said #5064 still didn't fix the issue [#3091](https://github.com/HDFGroup/hdf5/issues/3091) in 1.14.6.

PR #5065, opened 2024-11-04
Rename HDoff_t to hoff_t

derobins:
hoff_t is in line with the other typedefs in H5public.h (hsize_t, etc.).


byrnHDF (approved):

lrknox (approved):

hyoklee (approved):

qkoziol (changes requested):
I disagree with the revised naming.  The HDoff_t indicates that it's for portability betweens OS platforms' offset types.   The 'hsize_t', haddr_t', etc. are HDF5-specific types and don't have anything to do with OS platforms.  I believe that HDoff_t (and other platform portability types & function) should stay distinct from the HDF5 types.

derobins:
> I disagree with the revised naming. The HDoff_t indicates that it's for portability betweens OS platforms' offset types. The 'hsize_t', haddr_t', etc. are HDF5-specific types and don't have anything to do with OS platforms. I believe that HDoff_t (and other platform portability types & function) should stay distinct from the HDF5 types.

hsize_t is basically a platform-independent size. hbool_t was a platform-independent Boolean before bool was a standard.

derobins:
I'm going to turn that H5Pset/get_external() parameter into a haddr_t so we don't have to expose a platform-independent off_t. HDoff_t will go back into H5private.h, where it belongs.

PR #5074, opened 2024-11-05
Rename HDF5_ENABLE_Z_LIB_SUPPORT to HDF5_ENABLE_ZLIB_SUPPORT

byrnHDF:


jhendersonHDF:
As a bikeshed argument, this makes more sense to me as `Z_LIB` since it's the 'z' library (libz.so, for example). I think the main argument though is that this will probably cause some headache downstream somewhere for no real good reason.

brtnfld:
I think it is more in line with FindZLIB for cmake, which always references it as zlib with no underscore.

byrnHDF:
> As a bikeshed argument, this makes more sense to me as `Z_LIB` since it's the 'z' library (libz.so, for example). I think the main argument though is that this will probably cause some headache downstream somewhere for no real good reason.

Yes, it is a good argument, except everyone seems to refer to it as the zlib library. This would be the best time to make the change.

jhendersonHDF (approved):

lrknox (approved):

PR #5077, opened 2024-11-06
Fix failures in dtypes test on cygwin

jhendersonHDF:


test/dtypes.c
```diff
@@ -6436,16 +6436,17 @@ test__Float16(void)
     }
 
     if (is_little_endian) {
-        if (path->conv.u.lib_func != H5T__conv_noop) {
```
jhendersonHDF, line 6439: In at least cygwin, `H5T__conv_noop` (and some of the other functions used below) has a different address from `path->conv.u.lib_func`, so the two can't be compared directly. Instead, compare based on the conversion path name.

src/H5T.c
```diff
@@ -5660,8 +5660,8 @@ H5T__path_find_real(const H5T_t *src, const H5T_t *dst, const char *name, H5T_co
     old_npaths = H5T_g.npaths;
 
     /* Set a few convenience variables */
-    new_api_func = (matched_path && conv->is_app && conv->u.app_func);
```
jhendersonHDF, line 5663: Add an additional check to prevent the no-op path from being considered a new conversion path, which had some odd side effects from replacing the path with a slightly different, new no-op path.

qkoziol (approved):

mattjala (approved):

derobins (approved):

PR #5082, opened 2024-11-07
Switch HDoff_t to uint64_t in the public API

derobins:
Moves HDoff_t out of the public API and replaces the HDoff_t
parameter type in H5Pset/get_external() with a uint64_t.

Also removes OFF_T from the Fortran public API.


byrnHDF (dismissed):

mattjala (dismissed):

src/H5private.h
```diff
@@ -591,6 +591,16 @@ typedef double _Complex H5_double_complex;
...
+ */
+#ifdef H5_HAVE_WIN32_API
+typedef int64_t HDoff_t;
+#else
+typedef off_t HDoff_t;
+#endif
+
 /* __int64 is the correct type for the st_size field of the _stati64
  * struct on Windows (MSDN isn't very clear about this). POSIX systems use
  * off_t. Both of these are typedef'd to HDoff_t in H5public.h.
```
lrknox: This comment is out of date - the typedefs are just above in this file now.

brtnfld (dismissed):

src/H5private.h
derobins: Fixed

mattjala (approved):

bmribler (approved):

byrnHDF (approved):

epourmal:
I am puzzled with this the switch from off_t to haddr_t. The parameter is used for the offsets in **external** files that ARE not HDF5 files. haddr_t corresponds to an offset in a **logical** HDF5 file. Are you sure, this is a correct approach?

derobins:
> I am puzzled with this the switch from off_t to haddr_t. The parameter is used for the offsets in **external** files that ARE not HDF5 files. haddr_t corresponds to an offset in a **logical** HDF5 file. Are you sure, this is a correct approach?

off_t isn't portable, so it really should be replaced with something that is (HDoff_t is what we use internally). I originally had it as HDoff_t, which I moved to the public API, then wanted to switch the name to something more in line with the other HDF5 primitive types (hoff_t, like hsize_t) and got pushback from Quincey about the semantics of HD* vs h*. We compromised on haddr_t, which is logically reasonable, a 64-bit type, and portable.

We're not using the offset parameter to seek to negative offsets, so the semantics of off_t aren't really necessary and haddr_t makes more semantic sense than an arbitrary integer.

I see your point, though. A haddr_t isn't really a physical offset if there's a user block, so the semantics are a bit different. I just really want to get rid of off_t. Non-standard-C types are annoying to support.

epourmal:
> I just really want to get rid of off_t. Non-standard-C types are annoying to support.

I don't think we can have a good solution here.

epourmal:
Let's use uint64_t

fortnern (approved):
We'll need to make sure not to merge this to 1.14

derobins:
> We'll need to make sure not to merge this to 1.14

That won't be a problem. We don't maintain the 1.14 branch anymore and a 1.14.6 release (if that happens) will basically be a patch release with a few bugfixes cherry-picked into a copy of the 1.14.5 release branch.

release_docs/RELEASE.txt
brtnfld, line 1: Add to the Fortran Section:

The _**offset**_ parameter was changed for H5P(set|get)_external_f() from an INTEGER of KIND OFF_T to HADDR_T, and the OFF_T parameter is no longer a defined INTEGER KIND.

brtnfld (changes requested):
Add to the Release.txt the Fortran changes.

brtnfld:
@derobins, can we close this PR until you have the chance to take up the work again?

derobins:
Closed in lieu of #5809

PR #5083, opened 2024-11-08
Make H5F_LIBVER_LATEST a real enum value

derobins:
It was formerly a macro, but is now an enum value with the same value as the latest numbered API.

### DIFF
--- a/java/src/hdf/hdf5lib/HDF5Constants.java
+++ b/java/src/hdf/hdf5lib/HDF5Constants.java
@@ -575,9 +575,9 @@ public class HDF5Constants {
     /** */
     public static final int H5F_LIBVER_V200 = H5F_LIBVER_V200();
     /** */
-    public static final int H5F_LIBVER_NBOUNDS = H5F_LIBVER_NBOUNDS();
-    /** */
     public static final int H5F_LIBVER_LATEST = H5F_LIBVER_LATEST();
+    /** */
+    public static final int H5F_LIBVER_NBOUNDS = H5F_LIBVER_NBOUNDS();
     /** */
     public static final int H5F_OBJ_ALL = H5F_OBJ_ALL();
     /** */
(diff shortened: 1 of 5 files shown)


java/src/hdf/hdf5lib/HDF5Constants.java
derobins, line 1: The Java changes are just a re-ordering, putting LATEST in front of NBOUNDS

qkoziol (approved):

src/H5Fpublic.h
```diff
@@ -163,17 +163,16 @@ typedef struct H5F_sect_info_t {
...
-    H5F_LIBVER_V114     = 4, /**< Use the latest v114 format for storing objects */
-    H5F_LIBVER_V200     = 5, /**< Use the latest v200 format for storing objects */
+    H5F_LIBVER_EARLIEST = 0, /**< Use the earliest possible file format for storing objects */
+    H5F_LIBVER_V18      = 1, /**< Use the 1.8 file format for storing objects */
+    H5F_LIBVER_V110     = 2, /**< Use the 1.10 file format for storing objects */
+    H5F_LIBVER_V112     = 3, /**< Use the 1.12 file format for storing objects */
+    H5F_LIBVER_V114     = 4, /**< Use the 1.14 file format for storing objects */
+    H5F_LIBVER_V200     = 5, /**< Use the 2.0 file format for storing objects */
+    H5F_LIBVER_LATEST   = 5, /**< Use the latest file format for storing objects */
     H5F_LIBVER_NBOUNDS       /**< Sentinel */
```
jhendersonHDF, line 173: We probably want to check what consequences this will have by making `H5F_LIBVER_NBOUNDS` inaccurate as a counting sentinel value. `H5F_LIBVER_NBOUNDS` is used in several places in the library.

derobins: I'll take a look

brtnfld (approved):
Fortran should not be affected.

byrnHDF (approved):

PR #5091, opened 2024-11-12
Remove CMake-3.31.0 warnings

hyoklee:
This PR will remove warnings like below:

```
CMake Warning (dev) at fortran/src/CMakeLists.txt:132 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): DEPENDS.
```

Ref: https://github.com/lukka/get-cmake/releases/tag/v3.31.0

### DIFF
--- a/fortran/src/CMakeLists.txt
+++ b/fortran/src/CMakeLists.txt
@@ -133,7 +133,6 @@ add_custom_command (TARGET H5match_types POST_BUILD
     BYPRODUCTS ${HDF5_F90_BINARY_DIR}/H5f90i_gen.h ${HDF5_F90_BINARY_DIR}/H5fortran_types.F90
     COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:H5match_types>
     WORKING_DIRECTORY ${HDF5_F90_BINARY_DIR}
-    DEPENDS H5match_types
 )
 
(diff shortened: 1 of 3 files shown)


byrnHDF (approved):
Seems okay as this is using the TARGET form of add_custom_command.
Wonder why we didn't use the other form, OUTPUT?

lrknox (approved):

brtnfld:
Can you explain this further so I understand? The DEPENDS option specifies the files on which the command depends. Are you saying this DEPENDS is redundant because TARGET already handles the interdependency between build files?

Is one of the Github actions to check this with the minimum version of CMake to ensure it still works for the older versions?

byrnHDF:
> Can you explain this further so I understand? The DEPENDS option specifies the files on which the command depends. Are you saying this DEPENDS is redundant because TARGET already handles the interdependency between build files?

There are two forms of the add_custom_command. 
add_custom_command(OUTPUT ...)  and add_custom_command(TARGET ...)
The OUTPUT command supports DEPENDS the TARGET version does not and never did since at least 3.18.

brtnfld:
Right, so how is CMake getting informed of the dependency if it ignored the DEPENDCY command?

byrnHDF:
Most likely from the actual target properties.

brtnfld (approved):

PR #5092, opened 2024-11-12
fixed true/false definition and added missing h5_interm_group example

brtnfld:


hyoklee (approved):

byrnHDF (approved):

derobins (approved):

HDF5Examples/C/TUTR/h5_interm_group.c
```diff
@@ -64,13 +64,13 @@ main(void)
     /* Next commented call causes error stack to be printed out; the next one
      * works fine; is it a bug or a feature? EIP 04-25-07
```
derobins, line 65: We should address the comment!

brtnfld: This behavior is documented, so we should remove the comment:
 
> If name includes either a relative path or an absolute path to the target link, intermediate steps along the path must be verified before the existence of the target link can be safely checked. If the path is not verified, and an intermediate element of the path does not exist, the API will fail.

That said, as this behavior seems to demand attention, as evidenced by workarounds (https://gist.github.com/jzrake/3025642), we should look into handling this use case.

PR #5095, opened 2024-11-13
Add option to use HighFive C++

byrnHDF:
Until the old code is removed this will allow use of HighFive.  Only one example has been converted and working.


byrnHDF:
Converted to draft as examples need updated to show how to use HighFive in place of the existing C++ library.

PR #5124, opened 2024-11-15
Disable complex dtypes in API tests for now

mattjala:
The REST VOL CI is failing because the REST VOL/HSDS don't currently support complex datatypes. Since none of the other VOLs in CI do anything particular with complex datatypes, it should be fine to disable this until the REST VOL/HSDS are updated.


jhendersonHDF (changes requested):
The REST VOL CI is currently failing in its own internal `test_rest_vol.c` file, not the API tests, presumably because the changes in develop for creating a complex datatype haven't been moved over to that file yet. The REST VOL isn't being tested against the API tests at the moment, though when that is changed the tests will likely fail in a different way since the connector likely won't know how to handle complex datatypes. If `test_rest_vol.c` is updated to just reroll the datatype when it rolls for complex types, that should be enough to fix the CI for now. I'm also generally against temporarily disabling test functionality in this manner, just because it will be forgotten for a long time when there's no obvious indication that something is being skipped.

mattjala:
Right, I forgot that the REST VOL is using its own tests. I'll replace this with a similar change to the REST VOL modifying `test_rest_vol.c`.

PR #5128, opened 2024-11-18
Fix the expected DDL output value

hyoklee:
The test source code uses `-99.0`:

`    Particle1   fill_data[1]     = {{"no data", -1, -1, -99.0F, -99.0}};`

I386 Alpine CI test needs this patch to pass all tests.

This change may fail some other 64-bit systems. 
If so, I suspect a bug in the HDF5 library.

### DIFF
--- a/HDF5Examples/C/HL/tfiles/h5ex_table_11.ddl
+++ b/HDF5Examples/C/HL/tfiles/h5ex_table_11.ddl
@@ -178,7 +178,7 @@ GROUP "/" {
          DATATYPE  H5T_IEEE_F64LE
          DATASPACE  SCALAR
          DATA {
-         (0): 0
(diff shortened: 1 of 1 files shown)


byrnHDF (changes requested):
This is not the correct approach - why did this not fail until testing i386. First investigate the why of the failure.
And if i386 testing is correct and the 64-bit tests are correct - then add a special case.

byrnHDF:
Looking at the example code the fill date is defined:
    Particle1   fill_data[1]     = {{"no data", -1, -1, -99.0F, -99.0}};

So why is the test not always failing?

byrnHDF:
Closing this in favor of: HL TB examples are incorrect #5332

Issue #4394, opened 2024-04-12
Skip H5SHELL-test_flush_refresh Test for CMake


hyoklee:
If Perl is not found, CTest should skip it instead of reporting FAIL:

```
/usr/local/bin/bash: /home/runner/work/actions/actions/hdf5-1.14.4/build/test/H5TEST/test_flush_refresh.sh: No such file or directory
```

byrnHDF:
Closed by #5130

PR #5130, opened 2024-11-18
Add check for PERL_FOUND around actual test

byrnHDF:
Fixes Skip H5SHELL-test_flush_refresh Test for CMake #4394

### DIFF
--- a/test/ShellTests.cmake
+++ b/test/ShellTests.cmake
@@ -47,9 +47,6 @@ elseif (UNIX)
     ##############################################################################
     #  configure scripts to test dir
     ##############################################################################
-    if (H5_PERL_FOUND)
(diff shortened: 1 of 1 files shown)


derobins (approved):

derobins:
I wish there were a portable way of doing this without the Perl dependency

derobins:
You don't need Perl for this test, though. The test will run fine if the system's date command takes %s (true on Linux and Mac, but not some oddballs like Solaris).

Also, this doesn't have a Powershell equivalent and doesn't get tested on Windows. We should add GitHub issues for anything that is tested under Linux but not Windows due to no Powershell equivalent.

byrnHDF:
However  -the .sh file is not created earlier so cannot execute the test.
We would need to allow the configure file to happen.

lrknox (approved):

hyoklee:
Test passed: https://github.com/hyoklee/actions/actions/runs/11926479719

PR #5138, opened 2024-11-21
Add check for broken links in Doxygen generated html docs

### LINKED ISSUE #4928: Add check for broken links in Doxygen generated html docs.
Having this would have caught the issues in #4927

hyoklee:
This PR will report 36 broken links.  
CI will fail until they are all fixed.

close #4928


.github/workflows/main-cmake.yml
byrnHDF, line 1: I think this would be better served by a dedicated workflow that just needs to configure and build on one platform then having it check on all the systems.  After all the links don't change between platforms.

byrnHDF (changes requested):
Add link check workflow

.github/workflows/main-cmake.yml
byrnHDF: Also nothing in CMake tests finished because the Macs claim to not know about the command.

byrnHDF:
You could use the first 65 lines of the netcdf.yml file for the configure and build and then add the link check.

byrnHDF (dismissed):

byrnHDF:
Could you add one more thing - put the actual linkcheck into a separate step
```
      - name: check
        run: |
          mkdir build
          cd build
          cmake -DHDF5_BUILD_DOC:BOOL=ON ..
          make doxygen
          cd hdf5lib_docs/html
          linkchecker --check-extern ./index.html

```
Change to:
```
      - name: create
        run: |
          mkdir build
          cd build
          cmake -DHDF5_BUILD_DOC:BOOL=ON ..
          make doxygen

     - name: check
        run: |
          cd build/hdf5lib_docs/html
          linkchecker --check-extern ./index.html

```

It would be easier to focus on the actual issues rather then scrolling through the build stuff.

byrnHDF:
We should get this in now, to keep track of issues. The remaining errors are in files that will be converted to doxygen and will fix the URLs.
After the merge we will temporarily add a continue after error to allow PR merges to pass. This will be reverted once the remaining files are converted.

byrnHDF (approved):

lrknox (approved):

PR #5139, opened 2024-11-21
Convert Debugging HDF5 Applications to doxygen

byrnHDF:
Is this info even correct anymore?


doxygen/dox/TechnicalNotes.dox
```diff
@@ -50,7 +50,249 @@
...
+</tr>
+<tr>
+<td>z</td><td>Yes</td><td>Raw data filters</td>
+</tr>
+</table>
+
+In addition to including the code at compile time the application must enable each package at
+runtime. This is done by listing the package names in the HDF5_DEBUG environment variable. That
+variable may also contain file descriptor numbers (the default is 2') which control the output
+for all following packages up to the next file number. The word all` refers to all packages. Any
```
lrknox: Are these stray ' and ` in the 2 lines above?

Also should the ` in in line 287 be changed to '?

lrknox (dismissed):

doxygen/dox/TechnicalNotes.dox
```diff
@@ -1063,7 +1063,249 @@ been used at the lowest level to process the request.
...
+</td>
+<td>This disables all debugging even if the default was to debug something, then output
+from the meta data cache is send to the standard error stream and output from data types
+and spaces is sent to file descriptor 255 which should be redirected by the shell.
+</td>
+</tr>
+</table>
+The components of the HDF5_DEBUG value may be separated by any non-lowercase letter.
+
+\section sec_adddbg_trace API Tracing
```
jhendersonHDF: Essentially everything from this point down can just be removed since the API tracing functionality was removed a bit ago.

doxygen/dox/TechnicalNotes.dox
```diff
@@ -1063,7 +1063,249 @@ been used at the lowest level to process the request.
 
 /** \page APPDBG Debugging HDF5 Applications
 
-\htmlinclude DebuggingHDF5Applications.html
+\section sec_adddbg_intro Introduction
+The HDF5 library contains a number of debugging features to make programmers' lives
+easier including the ability to print detailed error messages, check invariant
+conditions, display timings and other statistics, and trace API function calls
```
jhendersonHDF: Which would also warrant revision here

doxygen/dox/TechnicalNotes.dox
byrnHDF: That is what I thought! Good so this goes away.

doxygen/dox/TechnicalNotes.dox
byrnHDF: But wait - we still have H5ARG_TRACE and bin/trace?

byrnHDF: |trace|Updates H5ARG_TRACE macros in H5ES_insert() calls (run by autogen.sh)|

jhendersonHDF: They trace function calls just for internal use with the Async APIs but in general the H5TRACE macros were removed from all API routines.

byrnHDF: So safe to remove in this file - good!

jhendersonHDF (approved):
I believe there are a couple other things here that could be modernized, but getting it into doxygen is a first step

lrknox (approved):

PR #5144, opened 2024-11-24
Fix a broken link in release process

hyoklee:
`snapshot-2.0.0` tag doesn't exist.


derobins (changes requested):
Why are we referring to a tag for a release that will never exist?

hyoklee:
> Why are we referring to a tag for a release that will never exist?

I just want to make this CI pass: https://github.com/HDFGroup/hdf5/actions/runs/11986265289/job/33419127227

byrnHDF (changes requested):
The release process doc is for future releases - therefore by definition this will fail.
The text needs to be changed that isn't parsed as a link but indicates what it should be.

PR #5145, opened 2024-11-25
Fix broken links in parallel compression document.

hyoklee:


byrnHDF (changes requested):
This is wrong  - the links need to point to doxygen

PR #5151, opened 2024-11-26
Skip some parallel tests for OpenMPI version 5.0.5

jhendersonHDF:
An issue in the 5.0.5 release of OpenMPI causes some parallel HDF5 tests to fail. The issue was resolved in the OpenMPI 5.0.6 release.

### DIFF
--- a/testpar/testphdf5.c
+++ b/testpar/testphdf5.c
@@ -346,7 +346,19 @@ main(int argc, char **argv)
     test_params.count = 0;
 
     /* Tests are generally arranged from least to most complexity... */
+
+    /* Skip this test for OpenMPI version 5.0.5 since an issue was
+     * introduced in that release which causes this test to fail.
+     * The issue was fixed for the 5.0.6 release.
+     */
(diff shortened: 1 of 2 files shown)


testpar/testphdf5.c
```diff
@@ -346,7 +346,19 @@ main(int argc, char **argv)
     test_params.count = 0;
 
     /* Tests are generally arranged from least to most complexity... */
+
+    /* Skip this test for OpenMPI version 5.0.5 since an issue was
+     * introduced in that release which causes this test to fail.
+     * The issue was fixed for the 5.0.6 release.
+     */
+#if defined(OPEN_MPI) && defined(OMPI_MAJOR_VERSION) && (OMPI_MAJOR_VERSION == 5) &&                         \
+    defined(OMPI_MINOR_VERSION) && (OMPI_MINOR_VERSION == 0) && defined(OMPI_RELEASE_VERSION) &&             \
```
jhendersonHDF, line 355: It would probably make sense to have a macro for MPI library version checking in case we need to do the same for an MPICH release in the future, but it didn't seem too worth it yet.

byrnHDF (approved):

lrknox (approved):

PR #5162, opened 2024-12-02
Convert Library Version html to doxygen

byrnHDF:


doxygen/dox/TechnicalNotes.dox
```diff
@@ -26,7 +26,264 @@
 
 /** \page RELVERSION HDF5 Library Release Version Numbers
 
-\htmlinclude LibraryReleaseVersionNumbers.html
```
bmribler, line 29: Shouldn't this be in a separate dox file?

byrnHDF: It was short enough to be here.

doxygen/dox/RelVersion.dox
```diff
@@ -0,0 +1,263 @@
...
+</td>
+</tr>
+</table>
+
+The <b>trunk</b> is the center of the system. New features are
+implemented in <b>feature branches</b> and aggregated in the trunk.
+<b>Release branches</b> are then created from the trunk.
+
+Before the 2.0 version, the minor version number of the trunk is always an odd number. From
+the time of Release 1.8.0 to the first 1.10 release, the trunk will be
```
mattjala: `the trunk will be` -> `the trunk was`

mattjala (dismissed):

doxygen/dox/RelVersion.dox
byrnHDF: Changed

lrknox (approved):
The text will need more thoughtful review before release.
1. I don't think the major/minor/release difference between the  1.x releases and the 2.x releases is sufficiently clear.
2. I think we changed H5check behavior at some point.  We should verify that the sections around line 170 accurately describe current behavior.

mattjala (approved):

PR #5169, opened 2024-12-07
H5Pf: Use off_t instead HDoff_t

jschueller:
Fixes this error:

H5Pf.c:1630:72: error: passing argument 5 of 'H5Pget_external' from incompatible pointer type [-Wincompatible-pointer-types]

cc @derobins since your PR #5082 maybe be related


derobins:
Thanks! I'm going to change the type to uint64_t, so off_t will go away shortly. I'm going to go ahead and close this.

PR #5170, opened 2024-12-08
Remove the hash_size param from H5Iregister_type()

### LINKED ISSUE #4344: Remove the `hash_size` parameter from `H5Iregister_type()`
This parameter is currently unused since the implementation has changed since the API call was designed. This internal detail should never have been exposed to users.

We should remove this parameter in the next major version of HDF5 and version the API call.

derobins:
The hash_size parameter of H5Iregister_type() hasn't been used since 1.8.
It's been removed and the API call has been versioned.

This PR also updates the make_vers script to handle v2.0.0.

Fixes #4344

### DIFF
--- a/bin/make_vers
+++ b/bin/make_vers
@@ -28,6 +28,21 @@ $min_sup_idx = 3;
 # Number of spaces to indent preprocessor commands inside ifdefs
 $indent = 2;
 
+# Global hash of functions ==> # symbol versions parsed from H5vers.txt
+$functions = {};
+
+# Global hash of functions ==> versioned function params parsed from H5vers.txt
+$func_params = {};
+
+# Global hash of typedefs ==> # symbol versions parsed from H5vers.txt
+$typedefs = {};
+
+# Array of {func name ==> symbol version} hashes. One hash per API version.
+@func_vers =  map { {} } 1..($max_idx + 1);
+
+# Array of {type name ==> symbol version} hashes. One hash per API version.
+@type_vers =  map { {} } 1..($max_idx + 1);
+
 #
 # Copyright by The HDF Group.
 # All rights reserved.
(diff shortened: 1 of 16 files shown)


bin/make_vers
```diff
@@ -301,149 +386,67 @@ sub parse_line ($) {
...
+        foreach my $vers (@vers_list) {
+            foreach my $hash_vers (@versions) {
+                if ($vers > $hash_vers) {
+                    next;
+                } else {
+                    if ($line_type == 1) {
+                        $api_vers_to_function_vers{$hash_vers}{$name} = $curr_sym_number;
                     } else {
-                        die "unknown line type: $line";
+                        $api_vers_to_type_vers{$hash_vers}{$name} = $curr_sym_number;
```
derobins, line 428: All that index manipulation nonsense reduced to about a dozen lines of code that even handles the v111 complication without additional logic.

bin/make_vers
derobins, line 1: I reworked the whole script to just use an explicit list of supported versions. All that array index complexity was giving me a headache.

mattjala (approved):

src/H5Iint.c
```diff
@@ -156,6 +156,76 @@ H5I_term_package(void)
...
+ * Function:    H5I__register_type_common
+ *
+ * Purpose:     Common functionality for H5Iregister_type(1|2)
+ *
+ * Return:      Success:    Type ID of the new type
+ *              Failure:    H5I_BADID
+ *-------------------------------------------------------------------------
+ */
+H5I_type_t
+H5I__register_type_common(unsigned reserved, H5I_free_t free_func)
```
jhendersonHDF, line 169: Seems like this function should return a pointer to the registered class rather than `H5I_type_t` so that if an error occurs later the code can free the registered class if need be.

qkoziol (approved):

derobins:
I'm going to go ahead and merge this to get the make_vers changes in. I'll address Jordan's comment in a future PR.

PR #5172, opened 2024-12-09
Prevent H5Pset* routines from modifying default plists

mattjala:
This prevents the connector in `H5VL_def_conn_s` and the one used by `H5P_DEFAULT/H5P_FILE_ACCESS_DEFAULT` from diverging.


qkoziol (changes requested):
In principal, yes, but rejecting the default property list of each type should apply to *all* the 'set' routines, for all types of property lists.

And, you don't need to reject H5P_DEFAULT, since H5P_object_verify() will already fail for that value.

qkoziol:
I suggest that you put the appropriate checking inside H5P_object_verify()

jhendersonHDF:
> In principal, yes, but rejecting the default property list of each type should apply to _all_ the 'set' routines, for all types of property lists.

In this case I'd call it a "not letting perfect be the enemy of good enough" situation. I asked Matt to create this PR due to the issue I encountered specifically with changing the VOL on the default FAPL. Fixing all the cases (and testing them) would be a decently larger effort.

> And, you don't need to reject H5P_DEFAULT, since H5P_object_verify() will already fail for that value.

I'd actually prefer to explicitly reject `H5P_DEFAULT` and `H5P_FILE_ACCESS_DEFAULT` first as an invalid argument check and not rely on `H5P_object_verify()` doing that, in case someone updates it in the future to accept `H5P_DEFAULT` and another parameter to allow it to translate to the correct property list.

qkoziol:
I understand your perspective, but I think the problem should be easy to address in a few hours of work:  add a new 'is_lib_default' flag to the property list structure, which is only set to true for the actual default property lists, and is set to false when one of those property lists is copied.

Then add a parameter to H5P_object_verify(), or someplace else that's central to the 'set' calls, that indicates that H5P_object_verify() is being called from a 'set' routine.   Then fail in H5P_object_verify() when both are true.

Probably only ~50 lines of actual code.  (and 600 lines of trivial changes to calls to H5P_object_verify :-)

src/H5Pfapl.c
```diff
@@ -5829,8 +5829,10 @@ H5Pset_vol(hid_t plist_id, hid_t new_vol_id, const void *new_vol_info)
     FUNC_ENTER_API(FAIL)
 
     /* Check arguments */
-    if (NULL == (plist = (H5P_genplist_t *)H5I_object_verify(plist_id, H5I_GENPROP_LST)))
+    if (NULL == (plist = (H5P_genplist_t *)H5P_object_verify(plist_id, H5P_FILE_ACCESS, false)))
         HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a property list");
+    if (H5P_is_default_plist(plist_id))
```
qkoziol: This shouldn't be necessary now.

test/plists.c
qkoziol, line 1: This test looks fine, but there's already a property list test (src/gen_plist.c).

qkoziol (commented):
Please merge the new and old tests together, so there's only one property list test.

Otherwise, looks good - thanks for taking the time to implement a holistic approach that addresses the problem at its root.

test/plists.c
qkoziol: (It doesn't matter to me which name you choose)

mattjala: `gen_plists` is associated with test files and is only built when a specific argument is provided. I think it makes sense to isolate that from "regular" property list tests.

qkoziol: Apologies, I had the wrong name, that should have been: test/tgenprop.c

qkoziol (approved):

lrknox (approved):

PR #5178, opened 2024-12-15
Update CMakeLists.txt to allow building tools before "install" step

poelmanc:
The INTERFACE include directory `$<INSTALL_PREFIX>/include` only gets populated after the install step has been performed. Adding `${CMAKE_CURRENT_LIST_DIR}` enables developers on a fresh build to build HDF5 and the libraries and tools in one go.


byrnHDF:
This really should not be necessary. 
What is your use case and build commands that cause the issue?

byrnHDF:
No activity - we can reopen discussion if necessary

PR #5085, opened 2024-11-08
Rework the compression library search

byrnHDF:
Added default names for search use.

PR #5181, opened 2024-12-16
Fix LOG VOL CI not including Zlib

mattjala:
#5085 changed the default for zlip support to 'off', so the CI should now specifically turn this one. The LOG VOL CI is currently failing two compression tests due to this issue.

### DIFF
--- a/.github/workflows/vol_log.yml
+++ b/.github/workflows/vol_log.yml
@@ -41,7 +41,7 @@ jobs:
             -DHDF5_ENABLE_PARALLEL:BOOL=ON \
             -DHDF5_ENABLE_THREADSAFE:BOOL=ON \
             -DHDF5_ALLOW_UNSUPPORTED:BOOL=ON \
-            -DHDF5_ENABLE_ZLIB_SUPPORT:BOOL=OFF \
+            -DHDF5_ENABLE_ZLIB_SUPPORT:BOOL=ON \
             -DHDF5_ENABLE_SZIP_SUPPORT:BOOL=OFF \
             ${{ github.workspace }}/hdf5
           cat src/libhdf5.settings


byrnHDF (approved):

jhendersonHDF (changes requested):
Since the workflow was already explicitly turning off zlib support before, we need to determine why the default being changed to OFF is now causing the LOG VOL CI to fail and fix that issue, rather than enabling zlib support to paper over the issue.

mattjala:
> Since the workflow was already explicitly turning off zlib support before, we need to determine why the default being changed to OFF is now causing the LOG VOL CI to fail and fix that issue, rather than enabling zlib support to paper over the issue.

The  workflow wasn't explicitly turning off zlib before. It used to include zlib by default like everything else. In the PR that changed the zlib default to off, the "ZLIB=OFF" line was redundantly added to the workflow. That should have been ZLIB=ON.

hyoklee (approved):

jhendersonHDF:
> > Since the workflow was already explicitly turning off zlib support before, we need to determine why the default being changed to OFF is now causing the LOG VOL CI to fail and fix that issue, rather than enabling zlib support to paper over the issue.
> 
> The workflow wasn't explicitly turning off zlib before. It used to include zlib by default like everything else. In the PR that changed the zlib default to off, the "ZLIB=OFF" line was redundantly added to the workflow. That should have been ZLIB=ON.

I don't believe the line was redundant; more that it was either a copy-paste or wasn't functioning as intended before. Either way, this still seems to point to an issue that needs to be fixed on the LOG VOL's side before re-enabling ZLIB. Otherwise, the issue will go unfixed. @brtnfld do you happen to know anything about this?

brtnfld:
I'm unaware of how zlib is being tested in the log-vol; I've created an issue https://github.com/HDFGroup/vol-log-based/issues/75 to determine what's happening.

jhendersonHDF (approved):

PR #5185, opened 2024-12-18
Update file-locking.md

donners-atos:
Changed the "writer process" to the "reader process" that can crash when reading.


byrnHDF (approved):

lrknox (approved):

lrknox:
Thank you @donners-atos!

PR #5195, opened 2024-12-30
Make H5FL package threadsafe

qkoziol:
Library:

- Added mutex that obeys disable-locking-for-this-thread (DLFTT) semantics, to enable threadsafe data structures within the library that "suspend" locking when a thread has exclusive ownership of the library. (in src/H5TSdlftt_mutex.[ch])

- Initialize the threadsafe aspects of the H5FL package at library startup, when concurrency is enabled. (in src/H5.c)

- Added new H5_global_t type, with H5_GLOBAL_INIT macro, to hold and initialize threadsafe globals with complex initialization that needs to be performed dynamically at runtime. (in src/H5private.h)

- Added wrapper for atomic size_t variables. (in src/H5TSprivate.h and src/H5TSatomic.c)

- Added new H5TS_bootstrap_g DLFTT mutex in src/H5TSint.c to support new threadsafe globals. (in src/H5TSint.c)

- H5FL package changes: (in src/H5FL.c and src/H5FLprivate.h)
    - Made free list limits atomic variables when concurrency is enabled
    - Access & modify free list limit globals through wrapper macros that work in both concurrent and non-concurrent builds of the library
    - Added DLFTT mutex and info for H5_global_t to each free list
    - Use new H5_GLOBAL_INIT macro to safely initialize each free list dynamically at runtime
    - Acquire & release free list's mutex when accessing information in the list to allocate & free objects (ifdef'd out in non-concurrent builds)
    - Added DLFTT mutex to each global list-of-lists
    - Acquire & release global list-of-list's mutex when accessing information in it, mainly to add new free lists to it, as they are initialized by use within the library (ifdef'd out in non-concurrent builds)
    - Small reworkings of the garbage collection routines to separate locking from accessing free list limits better
    - Switched free list initialization macros in private header  to use named fields, to make them portable between concurrency and non-concurrency builds
    - Removed unnecessary 'init' field from a few structs
    - Minor variable and struct field renamings, to help clarify their purposes

Testing:

- Added new h5_setup_local_rand() support routine to initialize h5_local_rand() with gettimeofday() or a predefined seed value.

- Expand test/tmisc.c, test_misc35(), to cover additional cases.

- Added tts_h5fl() to ttsafe test, skipped by default for non-concurrency builds.

- The tts_h5fl() testing generates multiple random test vectors of operations for each type of free list (regular, block, factory, and array) and then spawns threads that randomly chooses a set of test vectors to execute concurrently.


qkoziol:
Can I please get some reviews on this?  @fortnern @mattjala @jhendersonHDF @derobins

lrknox:
@qkoziol the conflicts are Makefile.ams that are gone in develop since autotools support has been removed there.  Do we need to fix the conflicts, or is this PR to be superseeded by another one that will be coming?

qkoziol:
I have been waiting for the CMake builds to stabilize, which I believe we're at.   I'll update this to resolve the conflicts this week.

jhendersonHDF:
Hi @qkoziol,

where are you at with regards to this PR? As there hasn't been much discussion internally or externally on this PR for a while, we'd like to close it, even if only temporarily, to keep our open PRs focused and progressing where possible.

qkoziol:
I have mixed feelings about closing it - I do understand that it's been cluttering up the pipeline for months.  On the other hand, I've made tons of progress on this stack of branches (this is the first in the stack), and anticipate getting back to finishing this work in the next couple of weeks.

On the whole though, I'm OK with closing it for now.  I can always re-open it in a little while, when I've finished the stack and we can discuss and work through them.

Issue #5037, opened 2024-10-30
Creating files with UTF-8 filenames broken on Windows


takluyver:
**Describe the bug**

Up to 1.14.3, HDF5 functions taking filenames expected UTF-8 filenames on Windows, and internally converted this to UTF-16 to pass to Windows Unicode (wide character) APIs. This allows you to create or open any filename Windows can represent, so long as you pass it as UTF-8, and h5py is relying on this behaviour to handle non-ASCII filenames consistently.

In #4172, this was changed to try the code-page dependent API first and fall back to the unicode API, although this is subtly different to what the user on the forum had requested:

> try MultiByteToWideChar (with CP_UTF8) first, and then rather than failing completely, fallback to simply calling _open rather than _wopen.

The problem is that in most of the common Windows code pages, almost any  UTF-8 string is valid, but representing the wrong characters. E.g.:

```
>>> "spent €3 in the café".encode("utf-8").decode("cp1252")
'spent â‚¬3 in the cafÃ©'
```

That's using [codepage 1252](https://en.wikipedia.org/wiki/Windows-1252), which is the default in Western Europe & many English-speaking areas.

And this is precisely what we see in h5py's tests. A file is created through HDF5 with the character U+201A "Single Low-9 Quotation Mark" (which confusingly looks like a regular comma, but isn't). This is encoded as 3 bytes in UTF-8: 

```
>>> chr(0x201a).encode('utf-8')
b'\xe2\x80\x9a'
```

And the filename that gets created contains `â€š`, which is those same 3 bytes interpreted as codepage 1252.

**Expected behavior**

The previous behaviour was, from our perspective, fine: `H5Fopen` & `H5Fcreate` simply took UTF-8 strings on Windows. So long as you know that and know how to convert encodings, you can do everything with that, with no ambiguity.

However, this may be confusing for Windows programmers who assume that a `char *` filename parameter will be interpreted as their current code page, since that's how Windows APIs work. You could document that this is different from Windows APIs and leave it alone.

If you want to allow a `char *` path to mean either UTF-8 or the current code page, there is unavoidable ambiguity: you cannot know if the bytes `'\xc3\x9f'` are UTF-8 (ß) or code page 1252 (ÃŸ). Favouring the code page in this scenario, as the current version does, makes UTF-8 essentially unusable. Favouring UTF-8 will mostly work: it's rare to hit on a UTF-8 sequence (except the ASCII part) by chance, but not impossible.

[...]

PR #5197, opened 2024-12-31
Revert "Address code page issues w/ Windows file paths (#4172)"

hmaarrfk:
This reverts commit eb0351efffe987a9c1882ccc9b03b5b0aec7f2dd.

This isn't a full solution for the problems that arose due to https://github.com/HDFGroup/hdf5/pull/4172 but as expressed by many https://github.com/HDFGroup/hdf5/issues/5037 the concerns for allowing users to "upgrade" past 1.14.3 are real

This ~proves~ provides you all a way to click merge as a way to stop the bleed.

edit: proves -> provides

### DIFF
--- a/src/H5system.c
+++ b/src/H5system.c
@@ -480,21 +480,27 @@ error:
 } /* end H5_get_utf16_str() */
 
 /*-------------------------------------------------------------------------
- * Function:     Wopen
+ * Function:     Wopen_utf8
  *
- * Purpose:      Equivalent of open(2) for use on Windows. Necessary to
- *               handle code pages and Unicode on that platform.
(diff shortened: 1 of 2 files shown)


derobins:
We still have plenty of time before the 2.0 release. Let's actually fix the problem instead of reverting to different broken behavior. We can always revert the PR at the last minute if we run out of time.

h-vetinari:
> We still have plenty of time before the 2.0 release.

Targeting 2.0 is entirely the wrong approach, as I've [outlined](https://github.com/HDFGroup/hdf5/issues/5037#issuecomment-2540072467) in the issue. Conda-forge will be stuck on 1.14.3 indefinitely due to this, and until a 2.0-migration can happen (much less conclude), this means months of gridlock.

PR #5198, opened 2025-01-02
Fix flipped calloc args

derobins:


derobins:
gcc-14 complains when the first arg to calloc() is a sizeof()

byrnHDF (approved):

mattjala (approved):

PR #5199, opened 2025-01-02
Fix 'unused parameter' warnings in C++ tests

derobins:


derobins:
I'm not super excited about maintaining the C++ wrappers, but while they exist they should at least compile quietly.

byrnHDF (approved):

qkoziol (commented):
Why not use the "unused" attribute?

hyoklee (approved):

jhendersonHDF:
This was something I missed when updating the testing frameworks since I don't usually build with the wrappers on, but like @qkoziol mentioned, using the unused attribute macro like everywhere else that got update seems better

derobins:
The unused macros aren't public and we avoid H5private.h in the C++ wrappers

PR #5213, opened 2025-01-07
Add H5Tdecode2, rename and deprecate H5Tdecode

### LINKED ISSUE #4661: H5Tdecode should require buffer size parameter
`H5Tdecode()` receives a buffer from the user. In the event that this buffer is malformed, the library may read values that cause it to run off the end of the buffer and read invalid memory. The bounds checking that the library performs in other, similar situations is not possible since the library does not know the size of the provided buffer. See #4436.

`H5Tdecode()` should be deprecated in favor of a new `H5Tdecode2()` which takes the buffer size as a parameter. 

**Additional context**
Some other places in the library face a similar problem - H5S also internally uses the macro which skips bounds checking if buffer size is unknown (`H5_IS_KNOWN_BUFFER_OVERFLOW`), so it may be productive to update parts of the H5S API as well.

mattjala:
When provided malformed or too-small buffers, H5Tdecode() could crash due to walking off the end of the buffer. The new buffer size parameter allows this to be reliably avoided.

Resolves #4661


release_docs/RELEASE.txt
```diff
@@ -178,6 +178,24 @@ New Features
...
+        hid_t H5Tdecode1(const void *buf);
+
+      The new signature is H5Tdecode2(). New code should use this version:
+
+      hid_t H5Tdecode2(const void *buf, size_t buf_size);
+
+      H5Tdecode() will map to the new signature unless the library is
+      explicitly configured to use an older version of the API.
+
+      The old signature
```
qkoziol: Typo?

test/dtypes.c
qkoziol, line 1: There should be a new test that exercises the deprecated H5Tdecode1 routine.

qkoziol (changes requested):
Looks very good, just needs a test for H5Tdecode1

c++/src/H5DataType.cpp
```diff
@@ -250,7 +250,7 @@ DataType::p_decode() const
     }
 
     // Call C function to decode the binary object description
-    hid_t encoded_dtype_id = H5Tdecode(encoded_buf);
+    hid_t encoded_dtype_id = H5Tdecode2(encoded_buf, buf_size);
```
bmribler, line 254: buf_size was initialized to 0 at creation.  Should there be a call to determine the size first?

c++/src/H5DataType.cpp
```diff
@@ -249,8 +249,19 @@ DataType::p_decode() const
...
+        throw DataTypeIException("DataType::p_decode", "Unable to retrieve encoded buffer size");
+    }
+
+    if (tmp_buf_size == 0) {
+        throw DataTypeIException("DataType::p_decode", "Invalid encoded buffer size");
+    }
+
     // Call C function to decode the binary object description
-    hid_t encoded_dtype_id = H5Tdecode(encoded_buf);
+    hid_t encoded_dtype_id = H5Tdecode2(encoded_buf, tmp_buf_size);
```
bmribler, line 254: Thank you for addressing the issue, but I'm sorry, I made a mistake in the original comment.  If encoded_buf is not NULL then buf_size will already be the size of the encoded buffer.  You can safely use buf_size in H5Tdecode2 as you originally did.  I'm sorry again!

java/src/hdf/hdf5lib/H5.java
```diff
@@ -13883,7 +13886,7 @@ public static long H5Tdecode(byte[] buf) throws HDF5LibraryException, NullPointe
         return id;
     }
 
-    private synchronized static native long _H5Tdecode(byte[] buf)
```
jhendersonHDF, line 13886: At first I was going to mention that there should be the corresponding JNI changes, but it seems there is no `_H5Tdecode` implementation currently. @byrnHDF We should either implement that or remove this function from being available.

src/H5T.c
```diff
@@ -3690,8 +3690,8 @@ H5Tencode(hid_t obj_id, void *buf, size_t *nalloc)
     FUNC_LEAVE_API(ret_value)
 } /* end H5Tencode() */
 
-/*-------------------------------------------------------------------------
```
jhendersonHDF, line 3693: Unintentionally removed?

src/H5T.c
```diff
@@ -3812,6 +3807,9 @@ H5T_decode(size_t buf_size, const unsigned char *buf)
     if (NULL == (f = H5F_fake_alloc((uint8_t)0)))
         HGOTO_ERROR(H5E_DATATYPE, H5E_CANTALLOC, NULL, "can't allocate fake file struct");
 
+    if (buf_size < 2)
```
jhendersonHDF: Not a big fan of an explicit comparison against the value 2 here. Using 2 occurrences of the `H5_IS_BUFFER_OVERFLOW` with a size of 1, for example, might be more verbose, but IMO is a lot clearer about what's going on.

src/H5T.c
```diff
@@ -3714,13 +3714,8 @@ H5Tdecode(const void *buf)
...
 
-    /* Create datatype by decoding buffer
-     * There is no way to get the size of the buffer, so we pass in
-     * SIZE_MAX and assume the caller knows what they are doing.
-     * Really fixing this will require an H5Tdecode2() call that
-     * takes a size parameter.
-     */
-    if (NULL == (dt = H5T_decode(SIZE_MAX, (const unsigned char *)buf)))
+    /* Create datatype by decoding buffer */
+    if (NULL == (dt = H5T_decode(buf_size, (const unsigned char *)buf)))
```
jhendersonHDF: Just a nitpick and not at all important for this PR, but I'd prefer to remove this cast. It's completely useless here (as with many other casts in the library).

src/H5Tpublic.h
```diff
@@ -1452,10 +1452,11 @@ H5_DLL herr_t H5Tencode(hid_t obj_id, void *buf, size_t *nalloc);
  *        object handle
  *
  * \param[in] buf Buffer for the datatype object to be decoded
+ * \param[in] buf_size Size of the buffer
  *
  * \return \hid_t{datatype}
  *
- * \details H5Tdecode() Given an object description of datatype in binary in a
+ * \details H5Tdecode2() Given an object description of datatype in binary in a
  *          buffer, H5Tdecode() reconstructs the HDF5 datatype object and
```
jhendersonHDF: `H5Tdecode()` -> `H5Tdecode2()`, just so the doxygen is right

c++/src/H5DataType.cpp
bmribler: Without having to call H5Tencode to get the size, that is.  Checking buf_size is a good idea though.  However, if you want to put a fix in DataType::close() to set encoded_buf to NULL then checking for NULL encoded_buf alone will be good enough.  Question me if you see anything I said questionable. :D

qkoziol (dismissed):

bmribler (dismissed):

src/H5T.c
```diff
@@ -3812,10 +3807,16 @@ H5T_decode(size_t buf_size, const unsigned char *buf)
...
         HGOTO_ERROR(H5E_DATATYPE, H5E_CANTALLOC, NULL, "can't allocate fake file struct");
 
+    if (buf_size != SIZE_MAX && H5_IS_BUFFER_OVERFLOW(buf, 1, buf + buf_size))
+        HGOTO_ERROR(H5E_DATATYPE, H5E_BADMESG, NULL, "buffer too small to be datatype message");
+
     /* Decode the type of the information */
     if (*buf++ != H5O_DTYPE_ID)
         HGOTO_ERROR(H5E_DATATYPE, H5E_BADMESG, NULL, "not an encoded datatype");
 
+    if (buf_size != SIZE_MAX && H5_IS_BUFFER_OVERFLOW(buf, 1, buf + buf_size))
```
jhendersonHDF: I believe both of these should be `buf + buf_size - 1`

test/dtypes.c
```diff
@@ -8954,8 +8954,14 @@ test_encode(void)
     unsigned char *vlstr_buf      = NULL;
     hid_t          ret_id;
     herr_t         ret;
+    char           test_msg[128];
 
-    TESTING("functions of encoding and decoding datatypes");
+    /* Silence unused parameter warning if built with no deprecated symbols */
+    (void)use_old_decode_api;
```
jhendersonHDF: Instead of silencing this way, using the `H5_ATTR_DEPRECATED_USED` macro would be a little cleaner

jhendersonHDF (dismissed):

qkoziol (dismissed):

fortran/src/H5Tff.F90
```diff
@@ -1857,28 +1857,31 @@ END SUBROUTINE h5tcommitted_f
...
+         IMPORT :: HID_T, SIZE_T
          IMPLICIT NONE
          CHARACTER(KIND=C_CHAR), DIMENSION(*), INTENT(IN) :: buf
+         INTEGER(SIZE_T), INTENT(IN) :: buf_size
          INTEGER(HID_T), INTENT(OUT) :: obj_id
        END FUNCTION h5tdecode_c
     END INTERFACE
 
-    hdferr = h5tdecode_c(buf, obj_id)
+    hdferr = h5tdecode_c(buf, buf_size, obj_id)
```
brtnfld: I would get rid of calling the c wrapper and call h5tdecode2 directly.
The buf_size should be an optional argument. Since we don't version the Fortran APIs, you must overload the function when adding new arguments. If buf_size is absent, then you can use the size of buf (LEN(buf)) passed in as the default. Please look at some of the H5A async APIs (or other APIs that call the C APIs) to get an idea of how this is done.

fortran/test/tH5T.F90
```diff
@@ -555,14 +555,14 @@ SUBROUTINE compoundtest(cleanup, total_error)
 
      !  Try decoding bogus buffer
 
-     CALL H5Tdecode_f(cmpd_buf, decoded_tid1, error)
+     CALL H5Tdecode_f(cmpd_buf, cmpd_buf_size, decoded_tid1, error)
```
brtnfld: Could you modify the test to have and have not the optional parameter?

brtnfld (changes requested):
Address Fortran issues.

fortran/src/H5Tff.F90
mattjala: Overloading makes sense, but why remove the C wrapper? It's been changed to use `H5Tdecode2` internally, so it seems like it would break the usage pattern with no clear benefit.

brtnfld: Because it is redundant code, instead of updating just code in the Fortran file, you had to update additionally another file's code (and the signature file) when it could have been removed instead.

src/H5vers.txt
```diff
@@ -86,6 +86,7 @@ FUNCTION: H5Tarray_create; ; v14, v18
 FUNCTION: H5Tcommit; ; v10, v18
 FUNCTION: H5Tget_array_dims; ; v14, v18
 FUNCTION: H5Topen; ; v10, v18
+FUNCTION: H5Tdecode; ; v12, v200
```
lrknox: I find H5Tdecode in the code tag hdf5-1_8_0, but not in tag hdf5-1_6_10, so I think this line should end with ; v18, v200.

I believe that change would also not add lines 151 - 154 in H5version.h.

lrknox (changes requested):

src/H5vers.txt
mattjala: That's strange. I can't find it in 1.6 either, but the [reference manual](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_t.html#ga186d7ed91fc0f9c55e19dcd96c7b229b) lists it as `since 1.2`. I'll update both of these to reflect a 1.8 addition.

fortran/src/H5Tff.F90
mattjala: If this logic applies to removing `h5tdecode_c`, I'd think it would apply to reworking the way that most fortran wrappers work and should probably be considered separately from this change.

brtnfld: That is the plan to remove the use of wrappers, but we don't have any resources to do it, so we've been doing it as part of other changes, such as this one, since it is usually less work.

lrknox (dismissed):

brtnfld (approved):
I made the Fortran API use an optional parameter instead since it is not all that needed from a Fortran perspective.

lrknox (approved):

byrnHDF (approved):

java/src/hdf/hdf5lib/H5.java
byrnHDF: Definitely looks like an oversight! I wonder if it got dropped along the way?

byrnHDF: Most likely the issue is the byte buffer and translating C memory to java memory

PR #5219, opened 2025-01-09
Remove gmnow() helper fxn from ros3 VFD

derobins:
Replace with gmtime(time(NULL))

The unused pointer arg in time() is obsolescent and should be NULL in new code. When the pointer is NULL, the time call cannot fail.


derobins:
Source: https://man7.org/linux/man-pages/man2/time.2.html

PR #5231, opened 2025-01-14
More H5FDs3comms.c/h cleanup

derobins:
* Make more error lines like the rest of the library (i.e., check for negative herr_t instead of FAIL)
* Rename functions that create AWS strings to include `_make_aws_` in the name
* Separate out the test-only load AWS credentials function
* Other minor cleanup

### DIFF
--- a/src/H5FDs3comms.h
+++ b/src/H5FDs3comms.h
@@ -452,52 +452,37 @@ typedef struct {
 extern "C" {
 #endif
 
-/****************************
- * HTTP FIELD LIST ROUTINES *
- ****************************/
-
+/* HTTP request buffer routines */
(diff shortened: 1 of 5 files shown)


src/H5FDs3comms.h
```diff
@@ -452,52 +452,37 @@ typedef struct {
 extern "C" {
 #endif
 
-/****************************
```
derobins, line 455: This looks messy, but if you look at the raw files you'll see it's just some reorg.

byrnHDF (approved):

mattjala (approved):

bmribler (approved):

PR #5233, opened 2025-01-14
Update install notes

byrnHDF:


glennsong09 (dismissed):

bmribler (dismissed):

release_docs/INSTALL_CMake.txt
```diff
@@ -814,20 +706,53 @@ These five steps are described in detail below.
       WiX--the Windows Installer XML toolset--lets developers create installers for
       Windows Installer, the Windows installation engine. See http://wixtoolset.org.
 
+   10. Backward compatibility
```
lrknox, line 709: Is "The 2.0.0 version of the HDF5 library can be configured to operate identically to the v1.14 library (or other versions)... overstating the effect and potentially misleading?  The 2.0.0 library with flags for earlier versions will still run as the 2.0.0 library, but will compile and run with earlier versions of the versioned functions.  
 
If any 1.x.x functions are removed in 2.0.0 and the application code used the old functions, it won't compile.  If a 2.0.0 function has the same signature as a 1.x.x function but has a different implementation or different output, it could behave differently, which I don't think would be operating identically to the earlier version.

lrknox (dismissed):

release_docs/INSTALL_CMake.txt
```diff
@@ -503,39 +508,37 @@ These five steps are described in detail below.
             * MinGW Makefiles
             * NMake Makefiles
             * Unix Makefiles
-            * Visual Studio 14 2015
-            * Visual Studio 14 2015 Win64
```
mattjala, line 507: Is dropping support for Visual Studio 2014 a 2.0.0 thing?

byrnHDF: Actually, I think we dropped support for everything before VS 2019!
CMake though supports 2015 - I think

mattjala (approved):

hyoklee (approved):

PR #5235, opened 2025-01-14
Fix AddressSanitizer issue in cache test

jhendersonHDF:
Fixes issues with usages of stack variables after function return related to API context pushing/popping

### DIFF
--- a/test/cache.c
+++ b/test/cache.c
@@ -222,7 +222,7 @@ static unsigned check_stats(unsigned paged);
 static void check_stats__smoke_check_1(H5F_t *file_ptr);
 #endif /* H5C_COLLECT_CACHE_STATS */
 
-static H5F_t *setup_cache(size_t max_cache_size, size_t min_clean_size, unsigned paged);
+static H5F_t *setup_cache(size_t max_cache_size, size_t min_clean_size, unsigned paged, H5CX_node_t *api_ctx);
 
 static void takedown_cache(H5F_t *file_ptr, bool dump_stats, bool dump_detailed_stats);
 
@@ -298,13 +298,10 @@ smoke_check_1(int express_test, unsigned paged)
     if (show_progress) /* 2 */
         fprintf(stdout, "%s() - %0d -- pass = %d\n", __func__, mile_stone++, (int)pass);
 
-    file_ptr = setup_cache((size_t)(4 * 1024 * 1024), (size_t)(2 * 1024 * 1024), paged);
+    file_ptr = setup_cache((size_t)(4 * 1024 * 1024), (size_t)(2 * 1024 * 1024), paged, &api_ctx);
 
     if (show_progress) /* 3 */
         fprintf(stdout, "%s() - %0d -- pass = %d\n", __func__, mile_stone++, (int)pass);
-
-    /* Push API context */
-    H5CX_push(&api_ctx);
 
     row_major_scan_forward(/* file_ptr               */ file_ptr,
                            /* max_index              */ max_index,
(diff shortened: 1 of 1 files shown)


test/cache.c
jhendersonHDF, line 1: These tests previously followed the general order of

1. call `setup_cache()` (which pushes its own API context)
2. call `H5CX_push()` to push a new API context
3. run the entirety of the test logic
4. call `takedown_cache()` (which pops the API context pushed in 1.)
5. call `H5CX_pop()` to pop the API context pushed in 2.

This out-of-order arrangement of pushing and popping was causing the tests
to try and make use of stack variables that were already gone at the point that
the `H5CX_pop()` calls were being made. Putting this in the correct order of
`setup_cache()` -> `H5CX_push()` -> `H5CX_pop()` -> `takedown_cache()` fixes
the issue, though this does seem to kind of point to a design issue in general.

jhendersonHDF:
AddressSanitizer is happy with this arrangement, but valgrind isn't for other reasons

jhendersonHDF:
Reworked this a little bit to handle the other stack issues as well

test/cache.c
```diff
@@ -298,14 +298,11 @@ smoke_check_1(int express_test, unsigned paged)
...
         fprintf(stdout, "%s() - %0d -- pass = %d\n", __func__, mile_stone++, (int)pass);
 
-    file_ptr = setup_cache((size_t)(4 * 1024 * 1024), (size_t)(2 * 1024 * 1024), paged);
+    file_ptr = setup_cache((size_t)(4 * 1024 * 1024), (size_t)(2 * 1024 * 1024), paged, &api_ctx);
 
     if (show_progress) /* 3 */
         fprintf(stdout, "%s() - %0d -- pass = %d\n", __func__, mile_stone++, (int)pass);
 
-    /* Push API context */
-    H5CX_push(&api_ctx);
```
jhendersonHDF, line 307: All of the push/pop calls in these tests appear to have been unnecessary, as API contexts are all pushed right after `setup_cache()` (which pushes its own API context) and popped right after `takedown_cache()` (which pops its own API context), at least before the initial version of this fix. Just removed all the extra push/pop calls and kept the single context pushed/popped by the setup/takedown functions.

test/cache.c
```diff
@@ -32367,20 +32054,19 @@ check_write_permitted(const H5F_t H5_ATTR_UNUSED *f, bool *write_permitted_ptr)
...
-    bool        verbose       = true;
-    int         mile_stone    = 1;
-    hid_t       fid           = H5I_INVALID_HID;
-    H5F_t      *file_ptr      = NULL;
-    H5C_t      *cache_ptr     = NULL;
-    H5F_t      *ret_val       = NULL;
-    haddr_t     actual_base_addr;
-    hid_t       fapl_id = H5P_DEFAULT;
-    hid_t       fcpl_id = H5P_DEFAULT;
-    H5CX_node_t api_ctx = {{0}, NULL}; /* API context node to push */
```
jhendersonHDF, line 32383: As `H5CX_push()` has since been refactored to use the passed in context object directly, rather then allocating a new node, this variable needs to be hoisted out of this function. Otherwise, the library will be holding onto a stale address once this function returns. The simplest way seemed to be just to pass in a pointer to a `H5CX_node_t`, but, again, this seems to point to design issues.

test/cache.c
```diff
@@ -26856,20 +26577,16 @@ check_flush_deps_err(unsigned paged)
         } /* end switch */
 
         takedown_cache(file_ptr, false, false);
-        if (!pass)
+        if (!pass) {
+            file_ptr = NULL;
```
jhendersonHDF, line 26581: Prevent accidentally calling `takedown_cache()` twice in case the first time fails

byrnHDF (approved):

qkoziol (approved):
Good change!  👍

PR #5246, opened 2025-01-17
Integrate shared chunk cache (H5SC) package with upper layers of library

fortnern:
Also some updates to shared chunk cache interface, and other general fixes. Non functional currently since the shared chunk cache hasn't been implemented.

### DIFF
--- a/src/H5CX.c
+++ b/src/H5CX.c
@@ -2901,6 +2901,40 @@ H5CX_set_mpio_actual_chunk_opt(H5D_mpio_actual_chunk_opt_mode_t mpio_actual_chun
 } /* end H5CX_set_mpio_actual_chunk_opt() */
 
 /*-------------------------------------------------------------------------
+ * Function:    H5CX_or_mpio_actual_chunk_opt
+ *
+ * Purpose:     Performs a bitwise "or" operation on the the actual chunk
+ *              optimization used for parallel I/O for the current API
+ *              call context.
+ *
+ * Return:      <none>
+ *
+ *-------------------------------------------------------------------------
+ */
+void
+H5CX_or_mpio_actual_chunk_opt(H5D_mpio_actual_chunk_opt_mode_t mpio_actual_chunk_opt)
+{
+    H5CX_node_t **head = NULL; /* Pointer to head of API context list */
+
+    FUNC_ENTER_NOAPI_NOINIT_NOERR
+
+    /* Sanity checks */
+    head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
+    assert(head && *head);
+    assert(!((*head)->ctx.dxpl_id == H5P_DEFAULT || (*head)->ctx.dxpl_id == H5P_DATASET_XFER_DEFAULT));
+
+    /* If the value is already set, "or" it with the requested value, othewise simply set it */
+    if ((*head)->ctx.mpio_actual_chunk_opt_set)
+        (*head)->ctx.mpio_actual_chunk_opt |= mpio_actual_chunk_opt;
+    else {
+        (*head)->ctx.mpio_actual_chunk_opt     = mpio_actual_chunk_opt;
+        (*head)->ctx.mpio_actual_chunk_opt_set = true;
+    }
+
+    FUNC_LEAVE_NOAPI_VOID
+} /* end H5CX_or_mpio_actual_chunk_opt() */
+
+/*-------------------------------------------------------------------------
  * Function:    H5CX_set_mpio_actual_io_mode
  *
  * Purpose:     Sets the actual I/O mode used for parallel I/O for the current API call context.
@@ -2926,7 +2960,40 @@ H5CX_set_mpio_actual_io_mode(H5D_mpio_actual_io_mode_t mpio_actual_io_mode)
     (*head)->ctx.mpio_actual_io_mode_set = true;
 
     FUNC_LEAVE_NOAPI_VOID
-} /* end H5CX_set_mpio_actual_chunk_opt() */
+} /* end H5CX_set_mpio_actual_io_mode() */
+
+/*-------------------------------------------------------------------------
+ * Function:    H5CX_or_mpio_actual_io_mode
+ *
+ * Purpose:     Performs a bitwise "or" operation on the actual I/O mode
+ *              used for parallel I/O for the current API call context.
+ *
+ * Return:      <none>
+ *
+ *-------------------------------------------------------------------------
+ */
+void
+H5CX_or_mpio_actual_io_mode(H5D_mpio_actual_io_mode_t mpio_actual_io_mode)
+{
+    H5CX_node_t **head = NULL; /* Pointer to head of API context list */
+
+    FUNC_ENTER_NOAPI_NOINIT_NOERR
+
+    /* Sanity checks */
+    head = H5CX_get_my_context(); /* Get the pointer to the head of the API context, for this thread */
+    assert(head && *head);
+    assert(!((*head)->ctx.dxpl_id == H5P_DEFAULT || (*head)->ctx.dxpl_id == H5P_DATASET_XFER_DEFAULT));
+
+    /* If the value is already set, "or" it with the requested value, othewise simply set it */
+    if ((*head)->ctx.mpio_actual_io_mode_set)
+        (*head)->ctx.mpio_actual_io_mode |= mpio_actual_io_mode;
+    else {
+        (*head)->ctx.mpio_actual_io_mode     = mpio_actual_io_mode;
+        (*head)->ctx.mpio_actual_io_mode_set = true;
+    }
+
+    FUNC_LEAVE_NOAPI_VOID
+} /* end H5CX_or_mpio_actual_io_mode() */
 
 /*-------------------------------------------------------------------------
  * Function:    H5CX_set_mpio_local_no_coll_cause
--- a/src/H5CX.c
+++ b/src/H5CX.c
@@ -2923,7 +2923,7 @@ H5CX_or_mpio_actual_chunk_opt(H5D_mpio_actual_chunk_opt_mode_t mpio_actual_chunk
     assert(head && *head);
     assert(!((*head)->ctx.dxpl_id == H5P_DEFAULT || (*head)->ctx.dxpl_id == H5P_DATASET_XFER_DEFAULT));
 
-    /* If the value is already set, "or" it with the requested value, othewise simply set it */
+    /* If the value is already set, "or" it with the requested value, otherwise simply set it */
     if ((*head)->ctx.mpio_actual_chunk_opt_set)
         (*head)->ctx.mpio_actual_chunk_opt |= mpio_actual_chunk_opt;
     else {
@@ -2984,7 +2984,7 @@ H5CX_or_mpio_actual_io_mode(H5D_mpio_actual_io_mode_t mpio_actual_io_mode)
     assert(head && *head);
     assert(!((*head)->ctx.dxpl_id == H5P_DEFAULT || (*head)->ctx.dxpl_id == H5P_DATASET_XFER_DEFAULT));
 
-    /* If the value is already set, "or" it with the requested value, othewise simply set it */
+    /* If the value is already set, "or" it with the requested value, otherwise simply set it */
     if ((*head)->ctx.mpio_actual_io_mode_set)
         (*head)->ctx.mpio_actual_io_mode |= mpio_actual_io_mode;
     else {
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -379,7 +379,8 @@ H5FL_EXTERN(H5S_sel_iter_t);
  *-------------------------------------------------------------------------
  */
 herr_t
-H5D__chunk_direct_write(H5D_t *dset, uint32_t filters, const hsize_t *offset, uint32_t data_size, const void *buf)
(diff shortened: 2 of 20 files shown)


src/H5Dio.c
```diff
@@ -436,21 +454,28 @@ H5D__read(size_t count, H5D_dset_io_info_t *dset_info)
...
-                    H5CX_set_mpio_actual_chunk_opt(H5D_MPIO_LINK_CHUNK);
+                    H5CX_or_mpio_actual_chunk_opt(H5D_MPIO_LINK_CHUNK);
             }
         }
 #endif /* H5_HAVE_PARALLEL */
     }
 
+    /* Make shared chunk cache read call if appropriate */
+    if (any_scc) {
+        assert(H5F_get_shared_cache(dset_info[0].dset->oloc.file));
```
mattjala: Does it make sense to check the first dataset's file for a shared cache, when the first dataset may not be using the shared chunk cache?

src/H5Dio.c
```diff
@@ -1003,30 +1055,33 @@ H5D__dset_ioinfo_init(H5D_t *dset, H5D_dset_io_info_t *dset_info, H5D_storage_t
...
-    } /* end if */
-    else {
-        /*
-         * This is the general case (type conversion, usually).
-         */
-        dset_info->io_ops.single_read  = H5D__scatgath_read;
-        dset_info->io_ops.single_write = H5D__scatgath_write;
-    } /* end else */
+    /* Only set layout ops if weŕe not using the shared chunk cache */
+    if (!dset->shared->layout.sc_ops) {
```
mattjala, line 1059: If the behavior from `H5D__read/writer()` of skipping IO info initialization for shared chunk cache datasets is going to be a consistent pattern, it might be better to just assert that the dataset isn't using the shared chunk cache. It would remove a layer of nesting, which is always nice.

src/H5VLnative_dataset.c
```diff
@@ -755,53 +756,75 @@ H5VL__native_dataset_optional(void *obj, H5VL_optional_args_t *args, hid_t dxpl_
...
-            /* Copy the user's offset array so we can be sure it's terminated properly.
-             * (we don't want to mess with the user's buffer).
-             */
-            if (H5D__chunk_get_offset_copy(dset, chunk_read_args->offset, offset_copy) < 0)
-                HGOTO_ERROR(H5E_DATASET, H5E_CANTCOPY, FAIL, "failure to copy offset array");
+            /* Check for shared chunk cache support */
+            if (dset->shared->layout.sc_ops) {
+                H5D_chunk_scc_udata_t udata;
+                size_t                buf_size = SIZE_MAX;
+                assert(0 && "set up the buf size from API when available. Either point directly into args "
```
mattjala, line 768: Was this hardcoded assertion failure supposed to be left in for now?

src/H5Dio.c
fortnern: The shared chunk cache is global to the file so this code should find it even if that dataset doesn't use it. All datasets must share the same file.

src/H5VLnative_dataset.c
fortnern: Yes. We should be updating the API soon and will be able to get rid of this assertion.

src/H5Dio.c
fortnern: This function is called even for datasets using the SCC. Do you mean we should avoid calling it for those datasets (and move up the setting of the dset and store parameters)?

mattjala: Disregard my comment - I had IO info init and IO init confused.

src/H5CX.c
```diff
@@ -2900,6 +2900,40 @@ H5CX_set_mpio_actual_chunk_opt(H5D_mpio_actual_chunk_opt_mode_t mpio_actual_chun
     FUNC_LEAVE_NOAPI_VOID
 } /* end H5CX_set_mpio_actual_chunk_opt() */
 
+/*-------------------------------------------------------------------------
```
fortnern, line 2903: Added these since this should be more convenient than get/or/set. I think we could possibly use these in a few more places instead of tracking the property in a separate variable.

src/H5Dchunk.c
```diff
@@ -8222,20 +8234,16 @@ H5D__chunk_iter(H5D_t *dset, H5D_chunk_iter_op_t op, void *op_data)
 } /* end H5D__chunk_iter() */
 
 /*-------------------------------------------------------------------------
- * Function:    H5D__chunk_get_offset_copy
+ * Function:    H5D__chunk_verify_offset
```
fortnern, line 8238: Turns out nothing in the direct chunk code modifes the offset array so there's no need to copy it and we can just pass it directly from the user qualified as const. Added this function to perform verification.

src/H5Dio.c
```diff
@@ -302,22 +315,25 @@ H5D__read(size_t count, H5D_dset_io_info_t *dset_info)
     if (io_skipped == count)
         HGOTO_DONE(SUCCEED);
 
-    /* Perform second phase of type info initialization */
-    if (H5D__typeinfo_init_phase2(&io_info) < 0)
-        HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "unable to set up type info (second phase)");
+    /* Initialize type info differently depending on if weŕe using the shared chunk cache or not */
```
fortnern, line 318: The shared chunk cache will need to perform its own type conversion initialization.

src/H5Dpkg.h
```diff
@@ -115,6 +115,13 @@
 /* Package Private Typedefs */
 /****************************/
 
+/* Typedef for cached dataset creation property list information */
+typedef struct H5D_dcpl_cache_t {
```
fortnern, line 119: This is not used outside of the H5D package so I demoted it to this header to minimize issues with circular dependencies

src/H5VLnative_dataset.c
```diff
@@ -755,53 +756,75 @@ H5VL__native_dataset_optional(void *obj, H5VL_optional_args_t *args, hid_t dxpl_
...
+                /* Make sure the dataset is chunked */
+                if (H5D_CHUNKED != dset->shared->layout.type)
+                    HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, FAIL, "not a chunked dataset");
 
-            /* Read the raw chunk */
-            if (H5D__chunk_direct_read(dset, offset_copy, &chunk_read_args->filters, chunk_read_args->buf) <
-                0)
-                HGOTO_ERROR(H5E_DATASET, H5E_READERROR, FAIL, "can't read unprocessed chunk data");
+                /* Read the raw chunk */
+                if (H5D__chunk_direct_read(dset, chunk_read_args->offset, &chunk_read_args->filters,
```
fortnern, line 789: Checking the validity of the offset is layout specific and has been moved to H5Dchunk.c for the chunked non-SCC case.

test/struct_chunk_storage.c
```diff
@@ -235,6 +235,9 @@ test_sparse_direct_chunk(hid_t fapl)
 
     TESTING("APIs for direct chunk I/O on structured chunks");
 
+    SKIPPED();
```
fortnern, line 245: Skipped since struct chunk direct I/O requires a shared chunk cache client implementation, which we don't have yet.

src/H5SCprivate.h
```diff
@@ -58,71 +61,79 @@ typedef herr_t (*H5SC_chunk_lookup_t)(H5D_t *dset, size_t count, hsize_t *scaled
...
-                                                     size_t *alloc_size /*in,out*/, void **chunk /*in,out*/);
+                                                     size_t *alloc_size /*in,out*/, bool partial_bound,
+                                                     void **chunk /*in,out*/, void *udata);
 
 /* Creates a new empty chunk. Does not insert into on disk chunk index. If fill is true, writes the fill value
  * to the chunk (unless this is a sparse chunk). The number of bytes used is returned in *nbytes and the size
  * of the chunk buffer is returned in *buf_size. */
 typedef herr_t (*H5SC_new_chunk_t)(H5D_t *dset, bool fill, size_t *nbytes /*out*/, size_t *buf_size /*out*/,
-                                   void **chunk /*chunk*/);
+                                   void **chunk /*chunk*/, void **udata[] /*out*/);
```
vchoi-hdfgroup: Is the udata[] correct here?

src/H5SCprivate.h
```diff
@@ -58,71 +61,79 @@ typedef herr_t (*H5SC_chunk_lookup_t)(H5D_t *dset, size_t count, hsize_t *scaled
...
 /* Inserts (or reinserts) count chunks into the chunk index if necessary. Old address and size (if any) of the
  * chunks on disk are passed as addr and old_disk_size, the new size is passed in as new_disk_size. This
  * function resizes and reallocates on disk if necessary, returning the address of the chunks on disk in
- * *addr. If an element in chunk is passed as NULL then this function shall insert a chunk large enough and
- * with properties set to (initially) hold only fill values. */
+ * *addr. */
 typedef herr_t (*H5SC_chunk_insert_t)(H5D_t *dset, size_t count, hsize_t *scaled[] /*in*/,
                                       haddr_t *addr[] /*in,out*/, hsize_t old_disk_size[],
-                                      hsize_t new_disk_size[], void *chunk[] /*in*/);
+                                      hsize_t new_disk_size[], void *chunk[] /*in*/, void *udata);
```
vchoi-hdfgroup: Should the udata be an array?

src/H5SCprivate.h
fortnern: Yes, because it's an array of udata pointers, one for each chunk

src/H5SCprivate.h
fortnern: Looks like I vectorized the wrong udata, thanks!

qkoziol:
@fortnern - Is this ready (enough :-) to review?

fortnern:
> @fortnern - Is this ready (enough :-) to review?

You're welcome to review it if you like. It's nowhere near production ready yet though - just part of the machinery for the shared chunk cache (designed to support sparse chunks), which hasn't been implemented yet. Note the PR is to a feature branch not develop.

src/H5Dio.c
```diff
@@ -286,10 +289,20 @@ H5D__read(size_t count, H5D_dset_io_info_t *dset_info)
             dset_info[i].skip_io = false;
         }
 
-        /* Call storage method's I/O initialization routine */
-        if (dset_info[i].layout_ops.io_init &&
-            (dset_info[i].layout_ops.io_init)(&io_info, &(dset_info[i])) < 0)
-            HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "can't initialize I/O info");
+        /* Check for shared chunk cache support */
+        if (dset_info[i].dset->shared->layout.sc_ops)
```
jhendersonHDF, line 293: What are the current cases where a dataset will not support the shared chunk cache? It would be fantastic if we could eliminate large amounts of the code that falls in the other path here.

src/H5SC.c
```diff
@@ -155,21 +157,24 @@ H5SC_flush_dset(H5SC_t *cache, H5D_t *dset, bool evict)
 /*-------------------------------------------------------------------------
  * Function: H5SC_read
  *
- * Purpose:  Reads raw data through a shared chunk cache.
+ * Purpose:  Reads raw data through a shared chunk cache. There may be
+ * datasets in the dset_info array that do not support the shared chunk
+ * cache. These datasets must be ignored by the shared chunk cache. There
+ * may also be datasets that have skip_io set. These datasets must also be
+ * skipped.
```
jhendersonHDF, line 164: While it may look a bit awkward at the higher levels, I propose that we partition up the `H5D_dset_io_info_t` structs into separate arrays (or do a sort) based on those that support H5SC and those which don't (also the same for the `skip_io` case) to simplify the logic here and not push the concerns of skipping over datasets down to these functions.

src/H5Dchunk.c
```diff
@@ -8260,70 +8262,11 @@ H5D__chunk_get_offset_copy(const H5D_t *dset, const hsize_t *offset, hsize_t *of
...
- * Function:    H5D__write_struct_chunk_direct
- *
- * Purpose:    Internal routine to write a structured chunk directly into the file.
- *
- * Return:    Non-negative on success/Negative on failure
- *
- *-------------------------------------------------------------------------
- */
-herr_t
-H5D__write_struct_chunk_direct(H5D_t H5_ATTR_UNUSED *dset, hsize_t H5_ATTR_UNUSED *offset,
```
jhendersonHDF, line 8281: Unused at this point or were these moved elsewhere?

src/H5Dio.c
fortnern: Indeed, the idea is to eventually have every dataset type support the shared chunk cache. Currently none of them do, though structured chunk will support it from the beginning

src/H5SC.c
fortnern: We could certainly do that. My initial idea is that it won't be difficult or costly to just skip over the unused datasets (as is the case in H5Dio.c, where it skips over datasets that DO support SCC), and is not worth the overhead, but if it ends up being difficult we can create a new list.

src/H5Dchunk.c
fortnern: Unused, should not be in H5Dchunk.c, and should be implemented in terms of H5SC layout callbacks. The H5SC code will handle the actual direct I/O, the only thing the layout callbacks need to do is insert the chunk into the index.

PR #5247, opened 2025-01-17
Enable S3 VFD Test for Windows CI

hyoklee:


byrnHDF:
I think I would prefer that instead of modifying this workflow to create a new vcpkg workflow - vcpkg-cmake.yml.
And then we can specify vcpg (https://github.com/microsoft/vcpkg ) specifid tests.

byrnHDF (changes requested):
Create vcpkg workflow.

lrknox:
VCPkg says it runs on Windows, macOS and Linux.  Should we add a VCPkg workflow for enabling S3 on multiple platforms?

PR #5254, opened 2025-01-21
Enable julia tests in develop.

lrknox:


byrnHDF (dismissed):

mattjala:
I don't think these should be re-enabled until we expect them to pass. From the logs: `1348 passed, 5 failed, 1 errored, 2 broken.`

What was the original reason these were disabled?

jhendersonHDF:
> I don't think these should be re-enabled until we expect them to pass. From the logs: `1348 passed, 5 failed, 1 errored, 2 broken.`
> 
> What was the original reason these were disabled?

This PR was mostly to check the current status of the Julia tests in develop and see if they can be re-enabled. The original issue that they were disabled for (https://github.com/JuliaIO/HDF5.jl/pull/1170) has since been fixed in develop, but it looks like these are new issues we haven't seen before and will need to be fixed.

PR #5257, opened 2025-01-21
Update 1.14 workflows for 1.14.6

byrnHDF:


.github/workflows/cmake.yml
```diff
@@ -93,18 +105,24 @@ jobs:
...
         uses: ./.github/workflows/testxpr-cmake.yml
 
-    call-release-cmake-julia:
-        name: "CMake Julia Workflows"
-        uses: ./.github/workflows/julia-cmake.yml
+#    call-release-cmake-julia:
+#        name: "CMake Julia Workflows"
+#        uses: ./.github/workflows/julia-cmake.yml
+#        with:
+#         build_mode: "Release"
```
lrknox, line 116: Are we not fixing julia test failures?

byrnHDF: Separate PR

.github/workflows/release.yml
```diff
@@ -55,7 +55,7 @@ jobs:
     needs: [log-the-inputs, call-workflow-tarball, call-workflow-ctest]
     uses: ./.github/workflows/abi-report.yml
     with:
-      file_ref: '1.14.4.3'
+      file_ref: '1.14.5'
```
jhendersonHDF, line 58: Should this be 1.14.6?

byrnHDF: NO. abi report needs the previous release to compare against

lrknox: I think this is the version for the abi test to compare 1.14.6 to, which should be 1.14.5.

lrknox (approved):

jhendersonHDF (approved):

PR #5258, opened 2025-01-21
Update compression / plugins CMake code

byrnHDF:


.github/workflows/cmake-script.yml
```diff
@@ -95,7 +95,9 @@ jobs:
             contents: |
               set (CTEST_DROP_SITE_INIT "my.cdash.org")
               # Change following line to submit to your CDash dashboard to a different CDash project
-              set (CTEST_DROP_LOCATION_INIT "/submit.php?project=HDF5")
+              #set (CTEST_DROP_LOCATION_INIT "/submit.php?project=HDF5")
```
lrknox, line 98: Not submitting results to CDash?  Or does it default to cdash.hdfgroup.org if not set?
Or maybe the commented line is just an example for submitting to another project.

lrknox (approved):

lrknox (changes requested):
Several tests failed because of the unused variable buffer_i that was removed yesterday from H5FDs3comms.c in PR #5257.  That change is apparently missing in this one;  is an update from hdf5_1_14_6 branch needed?  Otherwise I approve.

PR #5261, opened 2025-01-22
Upgrade doxygen to develop standard

byrnHDF:


doxygen/dox/VOLConnGuide.dox
jhendersonHDF, line 1: I believe pretty much all of this file needs reverted except for the first line change

doxygen/dox/file-locking.dox
```diff
@@ -55,7 +52,7 @@ it wrong could result in corrupt files or crashed readers, we decided to add
 a file locking scheme to help users get it right. Since this would also help
 prevent harmful accesses when SWMR is not in use, we decided to switch the
 file locking scheme on by default. This scheme has been carried forward into
-HDF5 1.12 and 1.13 (soon to be 1.14).
+HDF5 1.12 and 1.14 (soon to be 2.0).
```
jhendersonHDF: Minor nitpick to the point that I don't know if it's worth worrying about, but "1.13 (soon to be 1.14)" made sense, "1.14 (soon to be 2.0)" doesn't really.

doxygen/dox/VOLConnGuide.dox
byrnHDF: OK

doxygen/dox/file-locking.dox
byrnHDF: I'll just remove the parens

lrknox (approved):

PR #5265, opened 2025-01-23
Fix Julia wrappers

jhendersonHDF:
This PR brings the fix for `H5Oexists_by_name` for the Julia wrappers back from the develop branch.


jhendersonHDF:
Have to bring over the changes to the testing as well

jhendersonHDF:
The original issue is fixed with these changes and the current failure is the same as develop (with a small change in the workflows in develop). I'm going to disable the workflows again as the failure doesn't appear to be related to any changes and may need help from upstream.

byrnHDF (approved):

lrknox (approved):

PR #5268, opened 2025-01-24
Simplify the .gitattributes file

derobins:
* Removes all per-file attributes, many of which were incorrectly removing the text attribute from text/source files
* Removes all svneol attributes (not sure where those came from, but they don't appear in the git attributes docs)
* Marks *.h5 and certain test files as binary
* Removes the !eol attribute, which is unnecessary given text=auto and default git text handling


derobins:
text=auto will handle common binary formats like jpg and jar. Even marking HDF5 files as binary is probably overkill.

Helpful git attribute docs:

https://git-scm.com/docs/gitattributes

Also, FWIW, the crazy attributes are from the svn --> git migration we did back in 2016.

byrnHDF (approved):

lrknox (approved):

PR #5277, opened 2025-01-24
Revert 'Address code page issues w/ Windows file paths (#4172)'

jhendersonHDF:


jhendersonHDF:
While the intention was to address issues with this change in a major release, the change has caused a lot of downstream issues. Given that there is no definite timeline yet on a future major release, the decision was made to simply revert this change for the 1.14.6 release to help downstream software and then address this properly in the future.

lrknox (approved):

brtnfld (approved):
lgtm

PR #5279, opened 2025-01-24
Fix a few illegal pointer usages

### LINKED ISSUE #5278: Erroneous assignment of `false` to `partial` in `src/H5Shyper.c`
`src/H5Shyper.c` assigns a boolean value to a `_Bool*` and `test/cache.c` sends in `false` to `H5C_get_evictions_enabled` which takes a `_Bool*`

**Expected behavior**
I'd expect the code to have defined behavior.

 - HDF5 version 1f62484a938c76cc2c5b2c547dc2d741fd23071a (latest on `develop` right now)
 - OS and version: Ubuntu 24.04
 - Compiler and version: g++ (Ubuntu 14.2.0-4ubuntu2~24.04) 14.2.0
 - Build system (e.g. CMake, Autotools) and version: cmake version 3.28.3
 - Any configure options you specified: N/A
 - MPI library and version (parallel HDF5): N/A

**Additional context**
N/A

TedLyngmo:
Fixes #5278

### DIFF
--- a/src/H5Shyper.c
+++ b/src/H5Shyper.c
@@ -12334,7 +12334,7 @@ H5S_hyper_get_first_inc_block(const H5S_t *space, hsize_t clip_size, bool *parti
     if (diminfo->start >= clip_size) {
         ret_value = 0;
         if (partial)
-            partial = false;
+            *partial = false;
     } /* end if */
     else {
         /* Calculate index of first incomplete block */
--- a/test/cache.c
+++ b/test/cache.c
@@ -13800,12 +13800,12 @@ check_check_evictions_enabled_err(unsigned paged)
 
         (cache_ptr->resize_ctl).incr_mode = H5C_incr__threshold;
 
-        result = H5C_get_evictions_enabled(cache_ptr, false);
+        result = H5C_set_evictions_enabled(cache_ptr, false);
 
         if (result == SUCCEED) {
 
             pass         = false;
-            failure_mssg = "H5C_get_evictions_enabled succeeded() 1.\n";
+            failure_mssg = "H5C_set_evictions_enabled succeeded() 1.\n";
         }
         else if (cache_ptr->evictions_enabled == true) {
         }
@@ -13817,12 +13817,12 @@ check_check_evictions_enabled_err(unsigned paged)
 
         (cache_ptr->resize_ctl).decr_mode = H5C_decr__threshold;
 
-        result = H5C_get_evictions_enabled(cache_ptr, false);
+        result = H5C_set_evictions_enabled(cache_ptr, false);
 
         if (result == SUCCEED) {
 
             pass         = false;
-            failure_mssg = "H5C_get_evictions_enabled succeeded() 2.\n";
+            failure_mssg = "H5C_set_evictions_enabled succeeded() 2.\n";
         }
 
         (cache_ptr->resize_ctl).decr_mode = H5C_decr__off;


test/cache.c
```diff
@@ -13800,7 +13800,7 @@ check_check_evictions_enabled_err(unsigned paged)
 
         (cache_ptr->resize_ctl).incr_mode = H5C_incr__threshold;
 
-        result = H5C_get_evictions_enabled(cache_ptr, false);
+        result = H5C_get_evictions_enabled(cache_ptr, NULL);
```
jhendersonHDF, line 13803: It looks like `H5C_get_evictions_enabled` expects that non-`NULL` will be passed in here, so you may have to simply pass a temporary variable set to false.

src/H5Shyper.c
```diff
@@ -12334,7 +12334,7 @@ H5S_hyper_get_first_inc_block(const H5S_t *space, hsize_t clip_size, bool *parti
     if (diminfo->start >= clip_size) {
         ret_value = 0;
         if (partial)
-            partial = false;
+            *partial = false;
```
jhendersonHDF, line 12337: Great catch! Did you happen to see this issue with a tool or just from reading the code?

test/cache.c
TedLyngmo: I looked at the other code using `H5C_get_evictions_enabled` and it sends in `NULL` so I assumed it was ok. `H5C_get_evictions_enabled` also checks if it's `NULL` or not and seems to assert.
```
 HGOTO_ERROR(H5E_CACHE, H5E_SYSTEM, FAIL, "Bad evictions_enabled_ptr on entry.");
 ```
... but, ok, that might not be what's wanted in the test case there :-) I'll update

**Review by qkoziol — DISMISSED** [review]
Nice!

File: src/H5Shyper.c (hunk 2 above)
Comment (TedLyngmo) [reply src/H5Shyper.c:12337]: I actually didn't see it before I put `set (C_STANDARD 23)` everywhere. :-) Then the compiler suddenly started barking.

File: test/cache.c (hunk 1 above)
Comment (TedLyngmo) [reply test/cache.c:13803]: @jhendersonHDF I looked at the test case again. It's expected to fail if I read it correctly.
```
        result = H5C_get_evictions_enabled(cache_ptr, NULL);

        if (result == SUCCEED) {

            pass         = false;
            failure_mssg = "H5C_get_evictions_enabled succeeded() 2.\n";
        }
```

Comment (jhendersonHDF) [reply test/cache.c:13803]: Yep that particular test case is just checking to see that a failure happens when NULL is passed in. I'm assuming the other two cases fixed in this PR were probably a copy-paste issue from the nearby `H5C_set_evictions_enabled` call.

Comment (jhendersonHDF) [reply test/cache.c:13803]: The CI will pass in the meantime since it looks like the other two cases that were fixed were intending for the test to fail anyway, but definitely want to make sure it's failing for the intended reason rather than because `NULL` is passed.

Comment (TedLyngmo) [reply test/cache.c:13803]: Ah, ok, I see. I'll update - I'll just let my local test-run finish before uploading.
Hmm, `H5TEST-accum .......................***Failed    0.14 sec` - Is that a known failure?

Full list:
```
The following tests FAILED:
         21 - H5TEST-accum (Failed)
         76 - H5TEST-cache (Failed)
         79 - H5TEST-external_env (Failed)
         81 - H5TEST-vds_env (Failed)
         85 - H5TEST-flush2 (Failed)
        350 - H5LS-tfloat16_nosupport (Failed)
        352 - H5LS-tcomplex_nosupport (Failed)
```

Comment (jhendersonHDF) [reply test/cache.c:13803]: Not as far as I'm currently aware, but adding `-VV` to ctest should give you some output to see what's going on. `ctest -VV -R H5TEST-accum .` to make life easier.

Comment (TedLyngmo) [reply test/cache.c:13803]: Ok, there seems to be some problem loading the shared libraries in some tests:
```
    Start 21: H5TEST-accum

21: Test command: /home/ted/git/hdf5/build/bin/accum
21: Working Directory: /home/ted/git/hdf5/build/test/H5TEST
21: Environment variables:
21:  srcdir=/home/ted/git/hdf5/build/test/H5TEST
21: Test timeout computed to be: 1200
21: Testing the metadata accumulator
21: Testing simple write/read to/from metadata accumulator                 PASSED
21: Testing simple write/read to/from before metadata accumulator          PASSED
21: Testing simple write/read to/from after metadata accumulator           PASSED
21: Testing overlapping write to metadata accumulator                      PASSED
21: Testing overlapping write to partially clean metadata accumulator      PASSED
21: Testing overlapping write to accumulator larger then accum_size        PASSED
21: Testing non-overlapping write to accumulator larger then accum_size    PASSED
21: Testing accumulator adjustments after append/prepend of data           PASSED
21: Testing reading data from both accumulator and disk                    PASSED
21: Testing simple freeing metadata accumulator                            PASSED
21: Testing large metadata I/O operations                                  PASSED
21: Testing random writes to accumulator                                   PASSED
21: Testing SWMR write of large metadata: with latest format              accum_swmr_reader: error while loading shared libraries: libhdf5_test.so.1000: cannot open shared object file: No such file or directory
21: *FAILED*
21:    at /home/ted/git/hdf5/test/accum.c:2248 in test_swmr_write_big()...
21: child process exited abnormally
21: Testing SWMR write of large metadata: with non-latest-format          accum_swmr_reader: error while loading shared libraries: libhdf5_test.so.1000: cannot open shared object file: No such file or directory
21: *FAILED*
21:    at /home/ted/git/hdf5/test/accum.c:2248 in test_swmr_write_big()...
21: child process exited abnormally
21: *** TESTS FAILED ***
2/3 Test #21: H5TEST-accum .....................***Failed    0.15 sec
```
So, I assume it's not my patch that's doing that and I've uploaded a new version.

Comment (jhendersonHDF) [reply test/cache.c:13803]: Interesting. Yes your changes almost certainly aren't the cause of those failures, though if `H5TEST-cache` is still failing it would be good to know why.

Comment (TedLyngmo) [reply test/cache.c:13803]: I'm actually getting these two from `H5TEST-cache`:
```
76: check_check_evictions_enabled_err: failure_mssg = "H5C_get_evictions_enabled succeeded() 1.
76: check_check_evictions_enabled_err: failure_mssg = "H5C_get_evictions_enabled succeeded() 2.
```
With the dummy `bool` - so... change back to `NULL` I guess? With that it passes.

jhendersonHDF: Looks like you'll have to initialize at least this variable to `false` since the logic is checking to make sure that the function both failed and didn't set the variable to `true`, which it must be accidentally getting initialized as here.

jhendersonHDF: Hmm, based on the logic here, I'm not sure whether the test was fully finished in the first place. It looks like it was intending to check that the function fails in both cases and doesn't set the variable to `true` in this first one, but if both cases aren't actually failing, I'm not sure what the real intention was.

(Deleted old comment after re-reading the logic a bit)

TedLyngmo: Yeah, me neither :) Set it back to `NULL` or remove the tests?

jhendersonHDF: As long as `NULL` isn't passed for either of the parameters, `H5C_get_evictions_enabled` should never fail, so it looks like this may be another copy-paste issue. Going back through the git history, I believe both of those calls are actually supposed to be calls to `H5C_set_evictions_enabled` instead, where passing `false` for both would actually make sense again.

TedLyngmo: Well, that works too and the test passes since `H5C_set_evictions_enabled(cache_ptr, false);` fails in both cases.

I uploaded a new version with that in.

jhendersonHDF: Thanks! Based off the git history and the comment at the start of the test, it definitely looks like the intention was for these to be calls to `H5C_set_evictions_enabled` and something got missed in translation many years ago.

TedLyngmo: Funny what adventures trying to help people debug issues at Stackoverflow can lead to :-)
I came to this repo via https://stackoverflow.com/questions/79385144/cannot-write-a-variable-length-string-attribute-to-hdf5-file

jhendersonHDF: Funnily enough, we've had questions very recently about a similar situation. In that particular case I believe the program needs to pass `&value.c_str()` rather than `value.c_str()`. Variable-length strings are a bit of a special case in HDF5. It's potentially a bit misleading due to the way the array is declared, but https://github.com/HDFGroup/hdf5/blob/develop/HDF5Examples/C/H5T/h5ex_t_vlstring.c is an example of variable-length strings from the perspective of the C API. The short of it is essentially that the library expects `char *` for strings and `char **` for variable-length strings.

TedLyngmo: Ahhhh... wow! Yes, I noticed that the inner function did `H5MM_memcpy(&s, _vl, sizeof(char *));` so that makes sense. I'll try it out. Thanks! Edit: Works like a charm!

jhendersonHDF (approved):

qkoziol (approved):

Issue #5155, opened 2024-11-27
On macOS, using CMake, the OS-provided zlib is not found. Worse, it seems not possible to point to it.


seanm:
**Describe the bug**
On macOS, using CMake, the OS-provided zlib is not found. Worse, it seems not possible to point to it.

HDF5 provides a `ZLIB_DIR` setting, but it's described as "The directory containing a CMake configuration file for ZLIB." macOS does not provide any CMake configuration file for ZLIB, and in fact CMake is not even zlib's usual build system, so it's odd to look for that at all.

Other libraries I use, such as [VTK](https://vtk.org/) and [ITK](https://itk.org/), provide CMake variables like:

- `ZLIB_INCLUDE_DIR`
- `ZLIB_LIBRARY_DEBUG`
- `ZLIB_LIBRARY_RELEASE`

Which you can just point directly to the zlib include directory, and library files. This is most flexible, and then you don't care how zlib was built.

These tickets seem related, but maybe not identical: https://github.com/HDFGroup/hdf5/issues/4614 https://github.com/HDFGroup/hdf5/issues/4904

**Expected behavior**
macOS provides zlib, therefore I'd expect it to be detected by default, or at least be able to point to it.

**Platform (please complete the following information)**
 - HDF5 version: git master
 - OS and version: macOS 14
 - Compiler and version: Xcode 16
 - Build system (e.g. CMake, Autotools) and version: CMake

hyoklee:
Hi, @seanm !

Thank you for your report but I cannot duplicate your concern. 

```
-- Found ZLIB: /Applications/Xcode_16.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk/usr/lib/libz.tbd (found version "1.2.12")
-- H5_ZLIB_HEADER=zlib.h
-- H5_ZLIB_HEADER=zlib.h
-- H5_ZLIB_HEADER=zlib.h
-- H5_ZLIB_HEADER=zlib.h
-- H5_ZLIB_HEADER=zlib.h
```

Please see the complete action [log](https://github.com/hyoklee/actions/actions/runs/12093423460/job/33724388102) and [workflow](https://github.com/hyoklee/actions/actions/runs/12093423460/workflow).

All (incl. deflate) tests [passed](https://my.cdash.org/viewTest.php?onlypassed&buildid=2740030).

seanm:
[...]

Issue #4904, opened 2024-10-01
HDF5 finds the wrong zlib


haampie:
**Describe the bug**

HDF5 uses `find_package(ZLIB ... COMPONENTS ...)` before it falls back to `find_package(ZLIB)` in [CMakeFilters.cmake](https://github.com/HDFGroup/hdf5/blob/develop/CMakeFilters.cmake). The former call scans the system for a CMake config file (due to `COMPONENTS`). The latter call uses the builtin FindZLIB.cmake module.

Calling `find_package` twice is problematic because it messes up the search order: if a user configures HDF5 with `CMAKE_PREFIX_PATH=/path/to/my/zlib`, and this zlib does *not* have an associated CMake config file, HDF5 will ignore it and pick up a system zlib if it has a config file.

The problem happens for example on Fedora 40 after `dnf install zlib-devel`, which installs the file

```
/usr/lib64/cmake/ZLIB/ZLIB.cmake
```

which is picked up by the first `find_package` call. The second `find_package` call which would have found the intended zlib in the higher priority search directory is then ignored.

**Expected behavior**

As a user, if I pass `-DCMAKE_PREFIX_PATH=/path/to/zlib` (without cmake config file), HDF5 should use this zlib, not system zlib (with cmake config file).

Further, I would not expect HDF5 to prefer CMake config files at all, given that the promoted build system for zlib is their configure script, which obviously does not create a CMake config file.

I would strongly suggest to use one and only one `find_package` call, and use `ZLIB_USE_STATIC_LIBS` introduced in CMake 3.24 instead of rolling your own.

**Platform (please complete the following information)**
 - HDF5 version: any version in the last decade, specifically any after 4d0187a4832a60d56bc077a43855b91b8518271c
 - OS and version: any OS that has a system ZLIB.cmake, for example Fedora 40 with `zlib-devel` installed
 - Build system and version: CMake at any version.

**Additional context**

byrnHDF:
I would prefer a different solution then this as this is not a guarantee for all the systems we support. CMake has a number of options to find_package:
`             [NO_DEFAULT_PATH]
             [NO_CMAKE_ENVIRONMENT_PATH]
             [NO_CMAKE_PATH]
             [NO_SYSTEM_ENVIRONMENT_PATH]
             [NO_CMAKE_PACKAGE_REGISTRY]
             [NO_CMAKE_BUILDS_PATH]
             [NO_CMAKE_SYSTEM_PATH]
[...]

Issue #4614, opened 2024-06-28
Unable to build HDF5 with locally compiled szip or Zlib


jml06:
Dear developers,

**Platform (please complete the following information)**
 - HDF5 version : hdf5-1.14.4-3
 - OS and version: Centos-7.9.2009
 - Compiler and version: OneApi Intel 2024 or Gnu-11.2
 - Build system : cmake 3.29
 - MPI library and version : Intel/latest or OpenMPI-5.0.3 (compiled with gcc 11.2)

Context:
```shell
-bash-4.2$ more ./tools/BuildGnuLibs 
#!/usr/bin/env bash

ProjectDir=$1
echo "ProjectDir: "$ProjectDir

# --- OpenMPI-5.0.3 ---
if [ -d $ProjectDir/3rdParty/openmpi-5.0.3 ]; then
    rm -rf $ProjectDir/3rdParty/openmpi-5.0.3
fi

tar zxf $ProjectDir/3rdParty/openmpi-5.0.3.tar.gz -C $ProjectDir/3rdParty
cd $ProjectDir/3rdParty/openmpi-5.0.3

./configure --prefix=$ProjectDir/lib/Gnu/OpenMPI
make -j 36
make -j 36 install

printf "\n----------------------------------------------------------------------\n"
printf "   OpenMPI-5.0.3 Built and Installed"
printf "\n----------------------------------------------------------------------\n"

# --- ZLIB ---
if [ -d $ProjectDir/3rdParty/zlib-1.3.1 ]; then
    rm -rf $ProjectDir/3rdParty/zlib-1.3.1
fi

tar zxf $ProjectDir/3rdParty/zlib-1.3.1.tar.gz -C $ProjectDir/3rdParty
cd $ProjectDir/3rdParty/zlib-1.3.1

mkdir build
cd build
cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=$ProjectDir/lib/Gnu/zlib  -DCMAKE_C_COMPILER="$MPI_C_Compiler"
ninja -j 36
ninja install

printf "\n----------------------------------------------------------------------\n"
printf "   ZLIB libraries Built and Installed"
printf "\n----------------------------------------------------------------------\n"

# --- SZIP---
if [ -d $ProjectDir/3rdParty/szip-2.1.1 ]; then
    rm -rf $ProjectDir/3rdParty/szip-2.1.1
fi

[...]

PR #5280, opened 2025-01-27
Use only one find_package call for zlib.

byrnHDF:
Addresses #5155, and related #4614 #4904

zlib system installs tend to not have CMake config files and therefore require that a plain vanilla find_package call be used. This uses the find module for zlib. For more control over what gets found, an option to use config mode of find package is added. Which allows the use of static libs and better target properties. Also native zlib-ng, which usually has config files, would be better served with the config mode.


bmribler (dismissed):

brtnfld (dismissed):

CMakeFilters.cmake
```diff
@@ -117,11 +120,13 @@ if (HDF5_ENABLE_ZLIB_SUPPORT)
...
-        get_filename_component (libname ${ZLIB_LIBRARIES} NAME_WLE)
-        string (REGEX REPLACE "^lib" "" libname ${libname})
-        set_target_properties (ZLIB::ZLIB PROPERTIES OUTPUT_NAME ${libname})
+        if (NOT WIN32) #windows has a list of names
+          # The FindZLIB.cmake module does not set an OUTPUT_NAME
+          # on the target. The target returned is: ZLIB::ZLIB
+          get_filename_component (libname ${ZLIB_LIBRARIES} NAME_WLE)
+          string (REGEX REPLACE "^lib" "" libname ${libname})
+          set_target_properties (ZLIB::ZLIB PROPERTIES OUTPUT_NAME ${libname})
+      endif ()
```
lrknox:
```suggestion
        endif ()
```

release_docs/RELEASE.txt
```diff
@@ -565,6 +565,13 @@ Bug Fixes since HDF5-2.0.0 release
 
     Configuration
     -------------
+    - When using a system installed zlib library, the shared library is expected to
+      found in the system library path.
```
```suggestion
      be found in the system library path.
```

lrknox (dismissed):
2 code suggestions - indentation and missing "be".  Currently testing cmake configure commands;  ok so far.

CMakeFilters.cmake
```diff
@@ -85,30 +85,33 @@ endif ()
 if(NOT DEFINED ZLIBNG_PACKAGE_NAME)
   set(ZLIBNG_PACKAGE_NAME "zlib-ng")
 endif ()
-option (HDF5_ENABLE_ZLIB_SUPPORT "Enable Zlib Filters" OFF)
 if (HDF5_ENABLE_ZLIB_SUPPORT)
   if (NOT H5_ZLIB_HEADER)
     if (NOT ZLIB_USE_EXTERNAL)
+      option (HDF5_DEFAULT_ZLIB_SYSTEM "Use the default system installed ZLIB" ON)
```
jhendersonHDF: I'm somewhat iffy on this one. This seems more like something that should be handled by setting CMake path search variables rather than explicitly having to specify whether to use the system default. Using the system default should pretty much always be the default.

byrnHDF: In a perfect world yes, but I tried and there were unintended consequences. This is the best solution until zlib and installs are improved.

jhendersonHDF: Not to mention, this would obviously seem to imply an option being needed for every single other plugin we support, which is pretty much a non-starter IMO.

release_docs/RELEASE.txt
byrnHDF: fixed

CMakeFilters.cmake
byrnHDF: fixed

CMakeFilters.cmake
brtnfld: I agree that the system installation should be the default. Also, some of our CI tests should install zlib from a package manager, which is probably how most users would do it instead of compiling it from source, as it is done now.

byrnHDF: > Not to mention, this would obviously seem to imply an option being needed for every single other plugin we support, which is pretty much a non-starter IMO.

No - most all others including libaec have a proper install that can be found normally.

brtnfld: What makes a proper install? Is Zlib aware of the issue, or is there an issue I can refer to? Or is this an issue with CMake?

byrnHDF: ZLIB did no CMake packaging, which if done would allow finding the libs/include so much easier and accurate. On my latest fedora they tried, but messed it up so bad that the proper CMake config errors out. On top of that, the static build is useless for use in a shared library because they did not enable relocation. Every distro/installer looks like they have their own ideas of what to provide and most probably just use the autotools builds for zlib. Thereby the need to use the CMake FIND_ZLIB.cmake module. Some now use zlib-ng but make it look like zlib. The lack of full CMake support defeats the advantage of CMake targets. 
On top of all that, even if ZLIB fixes the problems, how long before it is commanplace to use? Therefore I added the option for future use.

jhendersonHDF: Can we flip this around a bit? Rather than the approach of "turn this option off" to go looking for a non-system default zlib, what about an option "turn this on" to go looking for one? This would hopefully be cleaner in the future when the option (hopefully) gets removed and it seems to me that the default "use the system zlib" should eventually work in the future, if distros aren't already patching this to work correctly. It seems the option on our end is for niche, currently broken cases and we should be explicit about that and tell users exactly when and why they might need to turn this option on.

lrknox:
Tested minimal cmake configure commands, e.g. cmake -G "Unix Makefiles" -DHDF5_ENABLE_ZLIB_SUPPORT=ON ../hdf5, on jelly, ec2-ubuntu24.04, and macmini-m1.  In a spack environment and outside the spack environment on jelly and macmini-m1.  In the spack environment the previous zlib build in spack is added;  outside it picks up the one beneath /usr.

On Windows 11 I tested with a VCPkg installed zlib, szip(libaec), and curl and ssl for ros3.  I added CMake ROOT variables for these  since they were not system installs.  I ran builds and tests with no errors/failures on all the machines.

lrknox (dismissed):

CMakeFilters.cmake
byrnHDF: I understand what your saying, however to me this option in the future would be normally OFF. It's ON because currently the non-cmake way forces the module use. Maybe a better name? HDF5_FIND_SYSTEM_ZLIB_MODULE

jhendersonHDF: Why would this option be OFF in the future? Isn't using the default system ZLIB what people are going to want in 99% of cases, assuming that ZLIB does cmake packaging or otherwise fixes this issue at some point in the future?

byrnHDF: That's why I suggested the name change - it is telling the build to use the module mode first (limited target info) instead of using config mode. The future should use config mode if zlib uses proper CMake practice.

jhendersonHDF: I see. In that case, yes I think the current name is misleading. I'm still undecided on which way should be the default though. I have to imagine some distros are patching this to make it work. Surely `find_package(ZLIB)` (without `MODULE`) is common enough that some other projects must have dealt with this. In which case they'd have to know to turn this option OFF to get the normal, expected behavior. I'm hesitant to make the "broken" behavior the default, even if it works in several places due to broken cmake configure files.

byrnHDF: I think the typical is the module for system installed zlib and will be that way for a while. Even if in the future, config mode becames the norm we would still need to update this to take full advantage of targets.

byrnHDF: Of course we are still using JAM and it's ancient compiler!

byrnHDF: How about HDF5_MODULE_SYSTEM_ZLIB

jhendersonHDF: It seems we might want to check how `CMAKE_FIND_PACKAGE_PREFER_CONFIG` comes into play here.

jhendersonHDF: With that variable, it seems like we can let MODULE be the default and if someone wants config mode they can specify that until we determine the best way to handle this in the future (just to avoid having yet another option people have to know the "when" and "why" of).

byrnHDF: Unfortunately those CMake variables are global in nature.

jhendersonHDF: That seems like a fairly acceptable trade-off though, all things considered, and is exactly what was commented on in #5155. I would keep the idea of adding a future-use option separate from the changes really needed here, since there doesn't necessarily seem to even be a use case for config searches yet and just complicates the logic.

byrnHDF: No that will force everything including plugins and libaec to use module mode.

jhendersonHDF: As I understand it, we should end up with module mode as the default anyway, since that's what the basic `find_package` does before falling back to config mode.

jhendersonHDF: I think these changes have outgrown the original purpose here though; we should just be eliminating the two `find_package` calls in favor of a single call and bringing that back into the 1.14.6 branch (along with the fix for `get_filename_component`, I suppose) and then adding this option could go into develop, but I wouldn't have it in this same PR.

byrnHDF: No CMake uses config mode first (the problem here) and then module mode. The preference in CMake is config mode and do away with modules.

byrnHDF: You may ask why - because the modules get created and then nobody maintains them. Config files are created and maintained by the library.

jhendersonHDF: Config mode may be the preferred way, but the docs very specifically say that, if a specific mode isn't given, module mode is used first with fallback to config mode, which makes sense as the ordering for finding packages that don't provide a config file by first looking for a module file specific to that package.

byrnHDF: This change by default does a module find for zlib. It can also use config mode by changing this option. Both cases handled!
Why force just one to be available?

jhendersonHDF: For two reasons, one is that the CMake docs say "Where possible, projects should find packages using this form. This reduces complexity and maximizes the ways in which the package can be found or provided." referring to providing the basic `find_package` form that does a module search followed by a config search, which should generally handle both cases. Secondly, because every option we add complicates the CMake logic and introduces the potential for bugs. Of course, the `CMAKE_FIND_PACKAGE_PREFER_CONFIG` isn't an ideal solution since it's global, but from this discussion it also seems like the current status is to always use module files anyway, so why not just use what works until config files work and there's a use case?

byrnHDF: I think we read the docs two different ways. There are two forms of the find_package and thus different search processes.

I disagree relying on module mode is wrong way of thinking - reading all the docs and from searches - config mode is really preferred.

byrnHDF: This PR satisfies the problem, allows a different use case and if users don't care - they are not forced to use it.

jhendersonHDF: Since the docs say that the preferred version of `find_package` to use is the basic signature, which by default does a module search followed by a config search, I think that's the signature that we should be using, if possible. When we switch to the full signature by using arguments outside the basic signature, that's when CMake switches to config-only search, and is presumably why we have the problem being fixed with this PR. If we really need the full signature, then so be it, but apparently we don't necessarily need it since using the basic signature by default works.

I'm not against giving users more options, but I don't think it's needed alongside the fix here and if there are other broken plugins in the future, I don't want to keep adding options like this for them. And I completely agree that config mode seems to be the preferred method for CMake, but they still have to support legacy and other packages that don't provide config files. So, starting by looking for a module file makes sense. It's not relying solely on module mode since there's a fallback to config mode, but the module file is a good place to start and the docs seem to say this right up front.

byrnHDF: Frankly, zlib is the only problem right now and if one installs a proper zlib build that has config files then the option can be changed by the user. Otherwise users don't need to care, as it does what is needed right now.
That's the trouble with flexibility - it is messy - but I would rather error on flexible if it is easy (as here, an option that does one or the other).

jhendersonHDF: Personally, I think this gives users the wrong incentive to subvert CMake's intended `find_package` mechanisms rather than taking the issue to upstream zlib where it can properly be fixed. Plus, users should still be able to use `ZLIB_ROOT` to point to an install, regardless of if it has CMake config files or not. As a workaround in develop this isn't particularly problematic since we have plenty of time to think about a better way, but I wouldn't want to see this option go into the release since it's likely to be confusing and I don't know that there's a real use case yet.

jhendersonHDF (dismissed):
Approved for develop, but we should discuss whether this is needed in 1.14.6

CMakeFilters.cmake
jhendersonHDF: It may also seem needlessly redundant, but a change to this approach might be to have the option being ON mean that we surround the call to `find_package` as in:

```
if (NOT HDF5_MODULE_MODE_ZLIB)
    set (CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
endif ()
find_package(ZLIB)
if (NOT HDF5_MODULE_MODE_ZLIB)
    set (CMAKE_FIND_PACKAGE_PREFER_CONFIG OFF)
endif ()
```

That at least removes the branching on the two different `find_package` calls and is explicit about what we're doing.

brtnfld:
What does this get us to put into develop? Develop is not a production branch, so why is there an urgency? We have time before the 2.0 release, so why not resolve this with other interested parties if needed? We should not add a new option for 1.14 if we are not convinced it is the correct solution. We also have the option of not fixing this in 1.14.6, as this needs a targeted long-term fix plan that can involve Kitware and other parties.

lrknox (dismissed):

brtnfld:
Also, nearly zero comments in the code explain the logic of this fix.

byrnHDF:
> Also, nearly zero comments in the code explain the logic of this fix.

Because it is in the issues?
But I can update it.

lrknox (dismissed):

CMakeFilters.cmake
```diff
@@ -85,30 +89,35 @@ endif ()
...
       else ()
         set (PACKAGE_NAME ${ZLIB_PACKAGE_NAME}${HDF_PACKAGE_EXT})
       endif ()
       set(ZLIB_FOUND FALSE)
-      if (HDF5_USE_ZLIB_STATIC)
-        set(ZLIB_SEARCH_TYPE static)
+      message (VERBOSE "Filter HDF5_ZLIB package name:${PACKAGE_NAME}")
+      if (HDF5_MODULE_MODE_ZLIB)
+        # Expect that the default shared library is expected with FindZLIB.cmake
+        find_package (ZLIB MODULE)
```
haampie, line 107:
```suggestion
        find_package (ZLIB MODULE REQUIRED)
```

CMakeFilters.cmake
```diff
@@ -85,30 +89,35 @@ endif ()
...
+        # Expect that a correctly built library with CMake config files is available
+        if (HDF5_USE_ZLIB_STATIC)
+          set(ZLIB_SEARCH_TYPE static)
+          if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
+            set(ZLIB_USE_STATIC_LIBS  ${HDF5_USE_ZLIB_STATIC})
+          endif()
+        else ()
+          set(ZLIB_SEARCH_TYPE shared)
+        endif ()
+        find_package (ZLIB NAMES ${PACKAGE_NAME} CONFIG OPTIONAL_COMPONENTS ${ZLIB_SEARCH_TYPE})
```
haampie, line 118:
```suggestion
        find_package (ZLIB NAMES ${PACKAGE_NAME} CONFIG REQUIRED OPTIONAL_COMPONENTS ${ZLIB_SEARCH_TYPE})
```

haampie:
Can you use `find_package(... REQUIRED)`? That improves error messages on failure. Also removes the need for branching on whether zlib is found.

CMakeFilters.cmake
```diff
@@ -85,30 +89,35 @@ endif ()
...
       endif ()
-      set(H5_ZLIB_FOUND ${ZLIB_FOUND})
+      set(H5_ZLIB_FOUND ZLIB_FOUND})
       if (H5_ZLIB_FOUND)
         if (HDF5_USE_ZLIB_NG)
```
haampie, line 122: not related to the change directly, but isn't this wrong? zlib-ng can be built and found as libz.so and zlib.h too.

byrnHDF:
> Can you use `find_package(... REQUIRED)`? That improves error messages on failure. Also removes the need for branching on whether zlib is found.

I understand the reasoning, however, there is still a group of devs, which are not convinced that a FATAL_ERROR should happen for not finding a compression library - also, if zlib is built in-line then that process also would need to be changed. 
Therefore at this moment, maybe it is best to make a simpler change and then work on a followup that implements REQUIRED processing.

haampie:
> which are not convinced that a FATAL_ERROR

Right now it's already a fatal error:

```cmake
if (HDF5_ENABLE_ZLIB_SUPPORT)
  ... find_package calls
  if (H5_ZLIB_FOUND)
    ...
  else ()
    set (HDF5_ENABLE_ZLIB_SUPPORT OFF CACHE BOOL "" FORCE)
    message (FATAL_ERROR " ZLib support in HDF5 was enabled but not found")
  endif ()
endif ()
```

`message (FATAL_ERROR` stops cmake.

byrnHDF:
> > which are not convinced that a FATAL_ERROR
> 
> Right now it's already a fatal error:
Yes, for develop, and it should stay that way. Just trying to get this implemented in smaller steps. Let's create an issue to fix the inline build code to behave the same as find_package REQUIRED and eliminate extra code?

byrnHDF:
Issue #5306 created to deal with the REQUIRED argument of find_package.

lrknox (dismissed):

release_docs/INSTALL_CMake.txt
```diff
@@ -877,6 +877,7 @@ if (BUILD_TESTING)
 HDF5_ALLOW_EXTERNAL_SUPPORT "Allow External Library Building (NO GIT TGZ)"       "NO"
 HDF5_ENABLE_PLUGIN_SUPPORT  "Enable PLUGIN Filters"                              OFF
 HDF5_ENABLE_SZIP_SUPPORT    "Use SZip Filter"                                    OFF
+HDF5_MODULE_MODE_ZLIB       ""Prefer module mode to find ZLIB"                   ON
```
bmribler: Two "s?

release_docs/RELEASE.txt
```diff
@@ -572,6 +572,15 @@ Bug Fixes since HDF5-2.0.0 release
 
     Configuration
     -------------
+    - When using a system installed zlib library, the shared library is expected to
+      be found in the system library path.
+
+      Setting the HDF5_MODULE_MODE_ZLIB option to OFF will force find_package to
+      use config mode first, an installed zlib, or an alternate installed zlib
```
bmribler: Perhaps, "use config mode first, an installed zlib" should be "use config mode first.  An installed zlib"?

byrnHDF: fixed.

release_docs/INSTALL_CMake.txt
byrnHDF: Fixed

CMakeFilters.cmake
byrnHDF: Issue #5306 created

CMakeFilters.cmake
byrnHDF: Issue #5306 created

lrknox (approved):
homebrew-installed libaec and szip were found with command "cmake -DHDF5_ENABLE_SZIP_SUPPORT -DHDF5_ENABLE_ZLIB_SUPPORT ../byrn-hdf5", which contained changes for PRs #5280 and #5298.  ctest ran deflate and szip tests;  all tests passed.  This was on an older mac 10.13, the only one where I have privileges for system installs.

jhendersonHDF (approved):

PR #5284, opened 2025-01-30
Only call find_package once, by default module mode for zlib

byrnHDF:
Merge from #5280


lrknox (approved):

lrknox:
tested again on macmini-m1 (macOS 15.3) and delta (Linux 4.18.0-477.81.1.el8_8.x86_64) after last change.  System zlib still found on both, system libaec also found on delta (not installed on macmini-m1).

PR #5286, opened 2025-01-30
Add new -testexpress option to testframe testing framework

jhendersonHDF:
Adds new -testexpress command-line option to the testframe testing framework to allow setting or overriding of the TestExpress level at runtime

Adds macros for the different currently defined TestExpress levels

### DIFF
--- a/test/btree2.c
+++ b/test/btree2.c
@@ -9928,10 +9928,10 @@ main(void)
     localTestExpress = h5_get_testexpress();
 
     /* For the Direct I/O driver, skip intensive tests due to poor performance */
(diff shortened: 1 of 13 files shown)


jhendersonHDF:
This option will be particularly useful for tests that we may want to run for a very long time, such as multi-threaded tests, while being restricted by the library's current default of level 3 for the purposes of CI.

byrnHDF (approved):

mattjala (approved):

qkoziol (approved):
Solid!

PR #5294, opened 2025-02-05
Update files based on issues discovered during release

byrnHDF:


jhendersonHDF (approved):

.github/workflows/tarball.yml
```diff
@@ -145,16 +145,16 @@ jobs:
         id: ren-basename
```
lrknox, line 145: The names of the steps in line 128 and 136 would reflect the action more clearly if "with" were inserted, as in 
- name: Create snapshot file with base name
and 
- name: Create release file with base name

lrknox (approved):

PR #5296, opened 2025-02-06
Fix segfault in H5Scombine_select

jhendersonHDF:
Checks for a hyperslab selection in the resulting dataspace before attempting to set a hyperslab-specific field


qkoziol (approved):
Nice catch

mattjala (approved):

bmribler (approved):

PR #5299, opened 2025-02-07
Fix error in H5Ddebug()

fortnern:
H5Ddebug() currently does not tag the dataset object header for the metadata cache, causing attempts to iterate over the index (which may require protecting a piece of metadata) to throw an error. This was causing h5ls -va to silently fail to print any chunk addresses. Added tagging and test cases for h5ls to prevent regression.


mattjala (dismissed):

fortnern:
Hmm I guess we changed the settings so we need to re-review after changes? Sorry Matt. Only thing I did was add a note to RELEASE.txt.

mattjala (approved):

qkoziol (approved):
Wow!  That's obscure!  Good find though.  :-)

jhendersonHDF:
Looks like the mirror VFD tests are failing due to an old issue that I thought was fixed at this point.

lrknox:
> Looks like the mirror VFD tests are failing due to an old issue that I thought was fixed at this point.

Test passed on retry.  We should keep an eye out for it, but I'll go ahead and merge the PR

PR #5307, opened 2025-02-10
Move TestAlarmOn()/TestAlarmOff() back into PerformTests()

jhendersonHDF:
Moves call to TestAlarmOn() in testframe framework back into the PerformTests() function due to the timing of when the framework processes command-line arguments

### DIFF
--- a/test/testframe.c
+++ b/test/testframe.c
@@ -158,12 +158,6 @@ TestInit(const char *ProgName, void (*TestPrivateUsage)(FILE *stream),
     /* Initialize value for TestExpress functionality */
     h5_get_testexpress();
 
-    /* Enable alarm timer for test program once TestExpress setting
-     * has been determined
-     */
-    if (TestAlarmOn() < 0)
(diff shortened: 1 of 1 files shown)


test/testframe.c
```diff
@@ -456,6 +450,10 @@ TestParseCmdLine(int argc, char *argv[])
 herr_t
 PerformTests(void)
 {
+    /* Enable alarm timer for tests */
+    if (TestAlarmOn() < 0)
```
jhendersonHDF, line 454: While I'd like for this timer to cover essentially the whole test program, the timing of when testframe tests process command-line arguments means that this call has to come late in test execution to react to the `-testexpress` command-line option. The timer is still for all of the tests as intended in #5292, but not for the init/shutdown parts of the program. To cover that as well, `TestParseCmdLine()` would need to be called as part of `TestInit()`, which would be a good idea but would require re-writing how that function deals with skipping tests, so this is the easier solution for now.

byrnHDF (approved):

lrknox (approved):

PR #5331, opened 2025-02-21
Fix objects not being closed on objcopy test failure

mattjala:
If the call to `compare_datasets` returns a test failure, then `oid` and `oid2` won't be cleaned up properly, which can propagate failures in some cases due to delaying file closes.


qkoziol (dismissed):

hyoklee (changes requested):
Please fix clang-formatting.

lrknox (approved):

hyoklee (approved):

PR #5333, opened 2025-02-24
Fix the HL H5TB fill-info #5332

### LINKED ISSUE #5332: HL TB examples are incorrect
Fix the expected DDL output value #5128 exposed a failure of the HL package to correctly insert fill data. In addition, because of badly created coding and testing, the issue was masked until the 32-bit testing exposed the problem.

byrnHDF:
Changed the test values to be unique to verify correct values are used.
Added a h5dump test to test_table so it could verify that fill values are correct.

Fixes #5332


hyoklee (changes requested):
Please update release documentation.

hyoklee (dismissed):
https://my.cdash.org/viewTest.php?onlypassed&buildid=2833655

It would be nice if you include GH Actions for i386 as well in this PR.

byrnHDF:
> https://my.cdash.org/viewTest.php?onlypassed&buildid=2833655
> 
> It would be nice if you include GH Actions for i386 as well in this PR.

i386 action already exists - just needed to remove special check for this failure

bmribler (dismissed):

hl/src/H5TB.c
```diff
@@ -2676,6 +2686,13 @@ H5TBdelete_field(hid_t loc_id, const char *dset_name, const char *field_name)
     if (H5TB_attach_attributes(table_title, loc_id, dset_name, nfields, tid_3) < 0)
         goto out;
 
+    if (NULL == (src_offset = (size_t *)malloc((size_t)nfields * sizeof(size_t))))
```
jhendersonHDF, line 2689: It doesn't look like `src_offset` is used below here? The lines below are still assigning `member_offset = H5Tget_member_offset(tid_3, (unsigned)i);`

byrnHDF: Looks like I didn't save the change.

jhendersonHDF: It looks similar to the other cases where the `H5TBget_field_info()` call is needed, but it's hard to say. Would probably need to have a specific test for it to check.

jhendersonHDF (approved):

hyoklee (approved):

lrknox (approved):

PR #5337, opened 2025-02-26
Fix bad VOL env var handling in CMake tests

mattjala:
Some CMake tests which deal with dynamic loading modify the HDF5_PLUGIN_PATH temporarily, but don't modify HDF5_VOL_CONNECTOR. If the user's environment has HDF5_PLUGIN_PATH and HDF5_VOL_CONNECTOR set in order to load a VOL connector on library initialization, then these tests fail due to searching for the connector named in HDF5_VOL_CONNECTOR but being unable to find it.

This PR changes CMake to unset HDF5_VOL_CONNECTOR as well to consistently ignore the user's VOL configuration during these tests


test/CMakeTests.cmake
```diff
@@ -906,7 +906,7 @@ if (BUILD_SHARED_LIBS)
 
   add_test (NAME H5PLUGIN-filter_plugin COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:filter_plugin>)
   set_tests_properties (H5PLUGIN-filter_plugin PROPERTIES
-      ENVIRONMENT "HDF5_PLUGIN_PATH=${CMAKE_BINARY_DIR}/filter_plugin_dir1${CMAKE_SEP}${CMAKE_BINARY_DIR}/filter_plugin_dir2;srcdir=${HDF5_TEST_BINARY_DIR}"
+      ENVIRONMENT "HDF5_PLUGIN_PATH=${CMAKE_BINARY_DIR}/filter_plugin_dir1${CMAKE_SEP}${CMAKE_BINARY_DIR}/filter_plugin_dir2;srcdir=${HDF5_TEST_BINARY_DIR};HDF5_VOL_CONNECTOR="
```
jhendersonHDF: Hmm I think this should rather go the other way around. If a user sets `HDF5_VOL_CONNECTOR`, we should assume they know what they are doing and may have set `HDF5_PLUGIN_PATH` as well, in which case the tests that modify `HDF5_PLUGIN_PATH` should be appending on to anything that exists rather than overriding it and not override `HDF5_VOL_CONNECTOR`.

test/CMakeTests.cmake
```diff
@@ -1005,7 +1005,7 @@ if (BUILD_SHARED_LIBS)
 
   add_test (NAME H5PLUGIN-vol_plugin COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:vol_plugin>)
   set_tests_properties (H5PLUGIN-vol_plugin PROPERTIES
-      ENVIRONMENT "HDF5_PLUGIN_PATH=${CMAKE_BINARY_DIR}/null_vol_plugin_dir;srcdir=${HDF5_TEST_BINARY_DIR}"
+      ENVIRONMENT "HDF5_PLUGIN_PATH=${CMAKE_BINARY_DIR}/null_vol_plugin_dir;srcdir=${HDF5_TEST_BINARY_DIR};HDF5_VOL_CONNECTOR="
```
jhendersonHDF, line 1008: This particular case is a little trickier though, since we want to test the NULL VOL connector in particular. Overriding `HDF5_VOL_CONNECTOR` in this case probably makes sense.

test/CMakeTests.cmake
byrnHDF: Remember that these tests are only temporarily modifying the ENV for the test duration. From cmake docs:
`ENVIRONMENT
Specify environment variables that should be defined for running a test.

Set to a [semicolon-separated list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) list of environment variables and values of the form MYVAR=value. Those environment variables will be defined while running the test. The environment changes from this property do not affect other tests.` 

Note this improvement:
`ENVIRONMENT_MODIFICATION
(https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT_MODIFICATION.html#environment-modification)
Added in version 3.22.`

bmribler (dismissed):

qkoziol (dismissed):

test/CMakeTests.cmake
```diff
@@ -906,7 +906,7 @@ if (BUILD_SHARED_LIBS)
 
   add_test (NAME H5PLUGIN-filter_plugin COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:filter_plugin>)
   set_tests_properties (H5PLUGIN-filter_plugin PROPERTIES
-      ENVIRONMENT "HDF5_PLUGIN_PATH=${CMAKE_BINARY_DIR}/filter_plugin_dir1${CMAKE_SEP}${CMAKE_BINARY_DIR}/filter_plugin_dir2;srcdir=${HDF5_TEST_BINARY_DIR}"
+      ENVIRONMENT "HDF5_PLUGIN_PATH=${CMAKE_BINARY_DIR}/filter_plugin_dir1${CMAKE_SEP}${CMAKE_BINARY_DIR}/filter_plugin_dir2${CMAKE_SEP}$ENV{HDF5_PLUGIN_PATH};srcdir=${HDF5_TEST_BINARY_DIR}"
```
byrnHDF: I'm not sure this is what we want - the purpose of the test is to find the two plugin files not a general test hdf5 library.
If vols are having test problems then maybe this test should be skipped.

byrnHDF (changes requested):
Need to check what is really needed.

test/CMakeTests.cmake
jhendersonHDF: @mattjala Did you run into this issue from `ctest` with a VOL connector, or while running the test manually?

mattjala: I ran into it while my environment was configured to load a VOL connector. I'll change the PR to consistently env VOL information (both plugin path and the `HDF5_VOL_CONNECTOR` value).

qkoziol (approved):

mattjala:
> Need to check what is really needed.

Both tests now make sure to ignore the VOL connector specified by the environment, since it doesn't make sense to use a VOL Connector for these tests.

byrnHDF (approved):

lrknox (approved):

Issue #2510, opened 2023-03-04
t_bigio test failure with 1.14.0 and mpich on Fedora rawhide


opoplawski:
**Describe the bug**
Test segfault.

**Platform (please complete the following information)**
 - HDF5 version
1.14.0 plus pull #2480 
 - OS and version
Fedora Rawhide i686
 - Compiler and version
gcc 13.0.1
 - Build system (e.g. CMake, Autotools) and version
autotools
 - Any configure options you specified
```
configure --build=i686-redhat-linux-gnu --host=i686-redhat-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib --libexecdir=/usr/libexec --localstatedir=/var --runstatedir=/run --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --disable-silent-rules --enable-fortran --enable-hl --enable-shared --with-szlib CC=mpicc CXX=mpicxx F9X=mpif90 'FCFLAGS=-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m32 -march=i686 -mtune=generic -msse2 -mfpmath=sse -mstackrealign -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -I/usr/lib/gfortran/modules  -I/usr/lib/gfortran/modules/mpich' --enable-parallel --exec-prefix=/usr/lib/mpich --libdir=/usr/lib/mpich/lib --bindir=/usr/lib/mpich/bin --sbindir=/usr/lib/mpich/sbin --includedir=/usr/include/mpich-i386 --datarootdir=/usr/lib/mpich/share --mandir=/usr/lib/mpich/share/man --with-default-plugindir=/usr/lib/mpich/hdf5/plugin
```
 - MPI library and version (parallel HDF5)
mpich 4.0.3

**Additional context**
[...]

PR #5344, opened 2025-03-02
Fix an oversize allocation in t_bigio on 32-bit systems

derobins:
A test in t_bigio.c attempts to allocate more than 4 GiB of memory, which will overflow (and wrap) the size_t type on 32-bit systems, creating a very small allocation instead of a very large allocation. The test then segfaults when it accesses memory outside of  the small buffer.

The test has been fixed by limiting the buffer to 2 GiB on 32-bit systems.

Fixes GitHub #2510

### DIFF
--- a/testpar/t_bigio.c
+++ b/testpar/t_bigio.c
@@ -1092,11 +1092,24 @@ single_rank_independent_io(void)
         file_id = H5Fcreate(FILENAME[1], H5F_ACC_TRUNC, H5P_DEFAULT, fapl_id);
         VRFY_G((file_id >= 0), "H5Dcreate2 succeeded");
 
-        /*
-         * Calculate the number of elements needed to exceed
-         * MPI's INT_MAX limitation
-         */
-        dims[0] = (INT_MAX / sizeof(int)) + 10;
+        if (sizeof(size_t) > 4) {
+            /*
+             * Calculate the number of elements needed to exceed
+             * MPI's INT_MAX limitation
+             */
+            dims[0] = (INT_MAX / sizeof(int)) + 10;
+        }
+        else {
+            /* 32-bit systems can't allocate a buffer > MPI's INT_MAX
+             * limitation, so that aspect of this test can't really
+             * be tested. Instead, just get enough elements to fill
+             * a 2 GiB buffer, which should be safe on most platforms
+             * (even though 4 GiB technically fits in the data type,
+             * the operating system's allocator probably can't fulfil
+             * a request that large).
+             */
+            dims[0] = ((INT_MAX / 2) / sizeof(int));
+        }
 
         fspace_id = H5Screate_simple(1, dims, NULL);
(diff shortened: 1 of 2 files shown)


qkoziol (approved):
Nice!

hyoklee (approved):
Fix Confirmed: https://my.cdash.org/tests/257263007

FYI, MPI_TEST_t_2Gio still fails though: https://my.cdash.org/tests/257263020

derobins:
Yes. I'll work on fixing all the 32-bit and big-endian problems before the release.

byrnHDF (approved):

testpar/t_bigio.c
```diff
@@ -1092,11 +1092,24 @@ single_rank_independent_io(void)
...
+        else {
+            /* 32-bit systems can't allocate a buffer > MPI's INT_MAX
+             * limitation, so that aspect of this test can't really
+             * be tested. Instead, just get enough elements to fill
+             * a 2 GiB buffer, which should be safe on most platforms
+             * (even though 4 GiB technically fits in the data type,
+             * the operating system's allocator probably can't fulfil
+             * a request that large).
+             */
+            dims[0] = ((INT_MAX / 2) / sizeof(int));
```
jhendersonHDF, line 1111: Good fix, though this entire test is largely useless on 32-bit systems since it won't really test anything that other tests don't.

jhendersonHDF (approved):

testpar/t_bigio.c
jhendersonHDF: I should also note that's basically the same case for t_2gio, so we should really decide if running it on 32-bit systems even makes sense, as it would basically be defeating the purpose of the test.

derobins: Agree. I thought about just skipping it, but testing the 2 GiB barrier is possibly useful.

jhendersonHDF: I'm assuming INT_MAX will be 2147483647 in most 32-bit places, in which case wouldn't this be grabbing ~1GiB of elements instead?

Ah no, this is the element count so ~4GiB.

derobins: I'll look at t_2gio tonight and think about it

derobins: t_2Gio passes on my 32-bit Debian 12 VM w/ the develop branch

jhendersonHDF: I think the t_2Gio test really needs rethought anyway. It first sets the "big I/O size" to 1MiB then writes only ~128MiB datasets, in which case it's not actually testing anything about 2GiB of I/O, or the MPI int limitation, but rather just whether the derived datatype creation functionality in the library works as expected. Did you happen to try running the test with a testexpress level of 0? Either set it to 0 at configure time or run the test with the `-testexpress 0` command-line argument. That at least runs a test that's supposed to test 4GiB I/O.

derobins: I'll give it a try tonight

PR #5347, opened 2025-03-03
Fix for external test for cray compiler

brtnfld:
Failing on Frontier and Perlmutter with Cray compilers. The original code works for -O0, so the compiler does something problematic with the optimizations.

### DIFF
--- a/test/external.c
+++ b/test/external.c
@@ -831,6 +831,10 @@ test_write_file_set(hid_t fapl)
     if (reset_raw_data_files(false) < 0)
         TEST_ERROR;
 
+    /* Generate the data */
+    for (i = 0; i < cur_size; i++)
+        whole[i] = (int)i;
+
     /* Create the dataset */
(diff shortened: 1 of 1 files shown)


test/external.c
```diff
@@ -831,6 +831,10 @@ test_write_file_set(hid_t fapl)
     if (reset_raw_data_files(false) < 0)
         TEST_ERROR;
 
+    /* Generate the data */
+    for (i = 0; i < cur_size; i++)
+        whole[i] = (int)i;
+
     /* Create the dataset */
     if ((mem_space = H5Screate_simple(1, &cur_size, &max_size)) < 0)
```
jhendersonHDF, line 839: If looking to create a reproducer at some point, I get the feeling it may have something to do with the fact that the loop invariant variable `cur_size` is passed into `H5Screate_simple()` when creating the dataspace.

jhendersonHDF (approved):

hyoklee (approved):

glennsong09 (approved):

PR #5352, opened 2025-03-05
Fix assertion failure in H5S__copy_pnt_list()

jhendersonHDF:
> [!IMPORTANT]
> Fix assertion failure in `H5S__copy_pnt_list()` by adding a zero rank check and update tests.
> 
>   - **Bug Fix**:
>     - Fix assertion failure in `H5S__copy_pnt_list()` in `H5Spoint.c` by adding a check for zero rank and converting the assertion to a normal error check.
>   - **Testing**:
>     - Add `test_h5s_bug4()` in `th5s.c` to test copying a dataspace with a point selection after extent reset with `H5Sset_extent_none()`.
>   - **Documentation**:
>     - Update `RELEASE.txt` to document the fix for the assertion failure in `H5S__copy_pnt_list()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for d9e20323715e42861baeb8ebf2f23b4de64c6912. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5Spoint.c
+++ b/src/H5Spoint.c
@@ -800,11 +800,14 @@ H5S__copy_pnt_list(const H5S_pnt_list_t *src, unsigned rank)
 
     /* Sanity checks */
     assert(src);
-    assert(rank > 0);
 
     /* Allocate room for the head of the point list */
     if (NULL == (dst = H5FL_CALLOC(H5S_pnt_list_t)))
         HGOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, NULL, "can't allocate point list node");
+
+    /* If source dataspace has no extent, just return new point list node */
+    if (rank == 0)
+        HGOTO_DONE(dst);
 
     curr     = src->head;
     new_tail = NULL;
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -514,6 +514,14 @@ Bug Fixes since HDF5-2.0.0 release
 ===================================
     Library
     -------
+    - Fixed an assertion failure in H5S__copy_pnt_list()
+
+      If a dataspace with a point selection in it had its extent reset with
+      the H5Sset_extent_none() function, any copy operations on that dataspace
+      would result in an assertion failure in debug builds of the library
+      due to the dataspace having an extent with a rank value of 0. This has
+      been fixed by converting the assertion failure into a normal error check.
+
     - Fixed an error in H5Ddebug
 
       H5Ddebug would fail for any chunked dataset with a chunk index, due to its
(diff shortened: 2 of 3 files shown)


jhendersonHDF:
Foiled again by the linkchecker

qkoziol (dismissed):

jhendersonHDF:
@qkoziol apologies on the timing; realized I should add a release note even though the issue is a minor one.

qkoziol (approved):

bmribler (approved):

src/H5Spoint.c
```diff
@@ -800,12 +800,15 @@ H5S__copy_pnt_list(const H5S_pnt_list_t *src, unsigned rank)
...
     assert(src);
-    assert(rank > 0);
 
     /* Allocate room for the head of the point list */
     if (NULL == (dst = H5FL_CALLOC(H5S_pnt_list_t)))
         HGOTO_ERROR(H5E_DATASPACE, H5E_CANTALLOC, NULL, "can't allocate point list node");
 
+    /* If source dataspace has no extent, just return new point list node */
+    if (rank == 0)
+        HGOTO_DONE(dst);
```
jhendersonHDF, line 810: @qkoziol I'm considering revising this. While it makes sense to me to return an empty point list node when the dataspace's extent has 0 dimensions (as we can't even tell how to copy the points over in that case), this causes problems elsewhere where the library expects `space->select.sel_info.pnt_lst->head` to be non-NULL. I have a fix for that as well, but I'm thinking of just having `H5Scopy()` / `H5Sselect_copy()` return failure for point and hyperslab selection dataspaces with a 0 dimension extent, since they both don't support `H5S_SCALAR` or `H5S_NULL` dataspaces and can only end up here when `H5Sset_extent_none()` is called on them. Thoughts?

qkoziol: That would be fine.  Another solution might be to have H5Sset_extent_none() clear the selection also (since you can't select anything within a "none" extent").   Generically, I would suggest reseting the selection any time that the extent's rank changes (also).

jhendersonHDF: That seems like a better solution to me. I suppose it makes sense to reset the selection to 'all' since that's what dataspaces start with?

qkoziol: Agree, back to 'all' for all changes in rank would be consistent with dataspace creation.

brtnfld:
@ellipsis-dev, review PR.

jhendersonHDF:
Will address with a different solution later

PR #5355, opened 2025-03-06
Fix two assertion failures in hyperslab code

jhendersonHDF:


src/H5Sselect.c
```diff
@@ -218,26 +218,39 @@ H5Sselect_copy(hid_t dst_id, hid_t src_id)
...
+    herr_t ret_value    = FAIL; /* Return value */
 
     FUNC_ENTER_NOAPI(FAIL)
 
     /* Check args */
     assert(dst);
     assert(src);
 
-    /* Release the current selection */
-    if (H5S_SELECT_RELEASE(dst) < 0)
```
jhendersonHDF, line 230: Part of this PR was fixing `H5Scopy()` to return failure for datasets with a NULL extent (rank value of 0). However, the previous code didn't really account for failure and would release the selection in the destination space before determining if the selection copy portion actually succeeded or not. This could result in segfaults later on when the application tries to close the destination dataspace. This reworks things so the copy is made into a temporary dataspace and only releases the selection in the destination dataspace once things have succeeded.

src/H5Shyper.c
```diff
@@ -3215,11 +3217,11 @@ H5S__hyper_copy(H5S_t *dst, const H5S_t *src, bool share_selection)
     assert(dst);
 
     /* Allocate space for the hyperslab selection information */
-    if (NULL == (dst->select.sel_info.hslab = H5FL_MALLOC(H5S_hyper_sel_t)))
```
jhendersonHDF, line 3218: The code here and below was directly manipulating the destination dataspace, which could leave things in a bad state if failure occurs. This function had to be reworked to use the local `dst_hslab` and only manipulate the destination dataspace after things have succeeded.

src/H5Shyper.c
```diff
@@ -3229,25 +3231,38 @@ H5S__hyper_copy(H5S_t *dst, const H5S_t *src, bool share_selection)
...
-            dst->select.sel_info.hslab->span_lst = src->select.sel_info.hslab->span_lst;
-            dst->select.sel_info.hslab->span_lst->count++;
+            dst_hslab->span_lst = src_hslab->span_lst;
+            dst_hslab->span_lst->count++;
         } /* end if */
-        else
+        else {
             /* Copy the hyperslab span information */
-            dst->select.sel_info.hslab->span_lst =
-                H5S__hyper_copy_span(src->select.sel_info.hslab->span_lst, src->extent.rank);
```
jhendersonHDF, line 3241: Code wasn't previously checking for errors here, which would hide problems when trying to copy a dataspace with an extent of 0 dims.

qkoziol (approved):

mattjala (approved):

PR #5356, opened 2025-03-06
Define __STDC_WANT_IEC_60559_TYPES_EXT__ before including other headers

mmuetzel:
Some standard headers might implicitly include `float.h`. If that happens before `__STDC_WANT_IEC_60559_TYPES_EXT__` is defined, the macros that are needed for the `_Float16` data type (like `FLT16_MAX`) might not be defined.

That is happening with mingw-w64 headers since the following change: https://github.com/mingw-w64/mingw-w64/commit/b40b6a021dcb6c2396e248e22f682b7bd2599d7a

Define `__STDC_WANT_IEC_60559_TYPES_EXT__` before including other headers in `H5private.h` to make that scenario less likely.

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -20,6 +20,11 @@
 #ifndef H5private_H
 #define H5private_H
 
+/* Define __STDC_WANT_IEC_60559_TYPES_EXT__ for _FloatN support, if available.
+ * Do that before including any other headers in case they include <float.h>
+ * implicitly. */
+#define __STDC_WANT_IEC_60559_TYPES_EXT__
+
 #include "H5public.h" /* Include Public Definitions    */
 
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


jhendersonHDF (dismissed):

hyoklee (dismissed):

qkoziol (changes requested):
Please add a comment about this, so it won't get broken in the future.

Here's something from the GCC docs that contains useful info:

"__FP_FAST_FMAF16
__FP_FAST_FMAF32
__FP_FAST_FMAF64
__FP_FAST_FMAF128
__FP_FAST_FMAF32X
__FP_FAST_FMAF64X
__FP_FAST_FMAF128X
These macros are defined with the value 1 if the backend supports the fma functions using the additional _Floatn and _Floatnx types that are defined in ISO/IEC TS 18661-3:2015. The include file math.h can define the FP_FAST_FMAFn and FP_FAST_FMAFnx macros if the user defined __STDC_WANT_IEC_60559_TYPES_EXT__ before including math.h."

glennsong09 (dismissed):

mmuetzel:
> Please add a comment about this, so it won't get broken in the future.

I added an inline comment with a note why the definition is early in that header file. Is that ok?

qkoziol (approved):

hyoklee (approved):

PR #5387, opened 2025-03-17
Suppress GCC warnings about invalid float16 literal suffix before C23

jhendersonHDF:


config/gnu-warnings/9.3
```diff
@@ -1 +1 @@
--Wc11-c2x-compat
```
jhendersonHDF: `-Wc11-c2x-compat` is deprecated in favor of `-Wc11-c23-compat` (apparently not recently enough, however).

src/H5Tconv_complex.c
```diff
@@ -1626,9 +1626,14 @@ H5T__conv_fcomplex__Float16(const H5T_t *st, const H5T_t *dt, H5T_cdata_t *cdata
...
                             size_t H5_ATTR_UNUSED bkg_stride, void *buf, void H5_ATTR_UNUSED *bkg)
 {
-    /* Suppress warning about non-standard floating-point literal suffix */
+    /* Suppress warning about non-standard floating-point literal suffix (F16) */
     H5_GCC_CLANG_DIAG_OFF("pedantic")
+    /* Suppress GCC warning about non-standard floating-point literal suffix (F16)
+     * before C23. This should be fine when _Float16 support is available.
+     */
+    H5_GCC_DIAG_OFF("c11-c23-compat")
     H5T_CONV_Zf(FLOAT_COMPLEX, FLOAT16, H5_float_complex, H5__Float16, -FLT16_MAX, FLT16_MAX);
```
jhendersonHDF, line 1635: These warnings are caused by usage of float16 macros when we have the `-Wc11-c2x-compat` flag enabled. The values behind the macros use a F16 suffix which is being warned about. Since keeping the `-Wc11-c2x-compat` flag seems valuable, just suppress the warnings here since the F16 suffix comes from GCC anyway. An alternative would be to create macros which wrap the usage of float16 macros with `__extension`, but there are several macros used and this also becomes awkward when '-' is involved, such as in `-FLT16_MAX`.

byrnHDF (approved):

mattjala (approved):

qkoziol (approved):

Issue #5377, opened 2025-03-14
Remove checking of external URLs in linkchecker workflow


byrnHDF:
linkchecker randomly fails on checking external URLs not in the repo/doxygen files. Due to this inconsistency, we should not check the couple of external URLs
Possible URLs:

https://www.youtube.com
https://www.gnu.org
https://www.oreilly.com
https://youtu.be
https://en.wikipedia.org
https://preshing.com
https://my.cdash.org
https://semver.org

PR #5391, opened 2025-03-18
Ignore gnu.org for #5377

hyoklee:
Linkchecker almost always fails for gnu.org.

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -42,6 +42,6 @@ jobs:
       - name: check
         run: |
           cd build/hdf5lib_docs/html
-          linkchecker --no-warnings --ignore-url=/doxygen.css --ignore-url=gnu.org --check-extern ./index.html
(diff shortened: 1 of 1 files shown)


byrnHDF (changes requested):
We should check if others should be added now - instead of multiple PRs - definitely add youtube

hyoklee:
@byrnHDF , linkchecker passed.

byrnHDF:
Just looking at the past half dozen failures shows that besides the high failure for gnu.org, that also failing is libpng.org and sourceforge (also help.hdfgroup.org)

.github/workflows/linkchecker.yml
```diff
@@ -42,6 +42,6 @@ jobs:
       - name: check
         run: |
           cd build/hdf5lib_docs/html
-          linkchecker --no-warnings --ignore-url=/doxygen.css --check-extern ./index.html
+          linkchecker --no-warnings --ignore-url=/doxygen.css --ignore-url=gnu.org --check-extern ./index.html
```
lrknox:
```suggestion
          linkchecker --no-warnings --ignore-url=/doxygen.css --ignore-url=gnu.org --ignore-url=en.wikipedia.org --ignore-url=help.hdfgroup.org --ignore-url=libpng.org --ignore-url=my.cdash.org --ignore-url=www.oreilly.com --ignore-url=preshing.com --ignore-url=semver.org --ignore-url=sourceforge.net --ignore-url=www.youtube.com --ignore-url=youtu.be --check-extern ./index.html
```

lrknox: I think this suggestion, if committed, will add all of the problematic links that Allen has listed.

byrnHDF (approved):

vchoi-hdfgroup (approved):

PR #5396, opened 2025-03-20
Add Developer build mode workflow

### LINKED ISSUE #2883: Add a GitHub Action that tests `--enable-internal-debug=all`.
**Describe the bug**

```
  CC       H5Ocache.lo
/home/hdftest/snapshots-hdf5/current/src/H5Ocache.c: In function ‘H5O__cache_notify’:
/home/hdftest/snapshots-hdf5/current/src/H5Ocache.c:539:15: error: ‘H5O_t’ has no member named ‘ndecode_dirtied’
             s oh->ndecode_dirtied = 0;
               ^
gmake[2]: *** [H5Ocache.lo] Error 1

test command:  bin/snapshot srcdir test clean setenv HDF5TestExpress 1 setenv HDF5_CXXFLAGS -std=c++11 deploydir /mnt/scr1/pre-release/hdf5/vdev setenv HDF5_USE_SELECTION_IO yes --enable-build-mode=production --enable-internal-debug=all zlib /usr/include,/usr/lib srcdirname moohan
```

**Expected behavior**

No compilation error

 - HDF5 version: develop
 - OS and version: CentOS7
 - Compiler and version: gcc
 - Build system (e.g. CMake, Autotools) and version: autotools
 - Any configure options you specified: --enable-internal-debug=all
 - MPI library and version (parallel HDF5): n/a 

**Additional context**

`grep` returns nothing:

```
nene-[9:11am]~/src/hdf5.HDFGroup/.github/workflows>grep -r enable-internal-debug *
nene-[9:11am]~/src/hdf5.HDFGroup/.github/workflows>
```

Credit: @lrknox

byrnHDF:
Only for linux - need to verify other platforms/configuration options.

Closes #2883, which reported a compile error. The goal of this PR is to verify that it compiles and executes minimal tests.


jhendersonHDF:
Note that the developer mode doesn't turn any of the extra debug options on by default. I'm assuming you may want to turn all/some of those on in the spirit of #2883.

.github/workflows/dev-cmake.yml
```diff
@@ -0,0 +1,89 @@
...
+           FILE_NAME_BASE=$(echo "${{ inputs.file_base }}")
+           echo "FILE_BASE=$FILE_NAME_BASE" >> $GITHUB_OUTPUT
+
+      - name: Install Linux Dependencies
+        run: |
+           sudo apt-get update
+           sudo apt-get install ninja-build doxygen graphviz
+           sudo apt install libssl3 libssl-dev libcurl4 libcurl4-openssl-dev
+           sudo apt install libaec0 libaec-dev
+           sudo apt install gcc-12 g++-12 gfortran-12
```
jhendersonHDF, line 35: I'm guessing these and the lines below are just copy-pasted, but we should probably be using the default software on the runners where possible, especially in the case of compilers. Sticking to gcc 12 will hide new warnings and/or errors that crop up in gcc 13 and now 14 (and there are some new warnings at least).

.github/workflows/dev-cmake.yml
```diff
@@ -0,0 +1,89 @@
...
+                 ${{ github.workspace }}/hdfsrc
+
+      - name: CMake Build
+        run: cmake --build . --parallel 3 --config Developer
+        working-directory: ${{ runner.workspace }}/build
+
+      - name: CMake Run Tests
+        env:
+          HDF5TestExpress: 0
+        run: ctest . --parallel 2 -C Developer -V -R H5TESTXPR
```
jhendersonHDF: Why not run all the tests? The testexpress tests aren't the only ones that would be affected.

byrnHDF: The issue was a compile issue - do all the tests need to be run? TestExpress 0 does test a lot.

jhendersonHDF: If there's a compile issue we should probably fix that. But if I remember right, the H5TESTXPR category doesn't cover many tests at all. Most of the library's tests don't even pay attention to TestExpress.

hyoklee (dismissed):

byrnHDF:
> extra debug options on by default

Which options?

.github/workflows/dev-cmake.yml
byrnHDF: I can fix that to use the latest gcc

jhendersonHDF:
> > extra debug options on by default
> 
> Which options?

I would say at least `HDF5_ENABLE_DEBUG_H5B2=ON` and `HDF5_ENABLE_DEBUG_H5FS_ASSERT=ON`. There are several others but unfortunately they pretty much all add extra debugging while also including printfs that will mess up testing. It would be nice to separate out the real debug code from printing debug code.

I believe historically it's been found that that first one slows things down a lot though, so might be good to check the before/after runtime.

byrnHDF:
Added the options and windows job.
Windows failed with:
[1/3070] Building C object src\CMakeFiles\hdf5-static.dir\H5.c.obj
FAILED: src/CMakeFiles/hdf5-static.dir/H5.c.obj 
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1443~1.348\bin\Hostx64\x64\cl.exe  /nologo -DH5AC_DEBUG -DH5B2_DEBUG -DH5CX_DEBUG -DH5D_CHUNK_DEBUG -DH5D_DEBUG -DH5FS_DEBUG_ASSERT -DH5F_DEBUG -DH5MM_DEBUG -DH5O_DEBUG -DH5T_DEBUG -DH5Z_DEBUG -DH5_NO_FREE_LISTS -DH5_STRICT_FORMAT_CHECKS -D_BIND_TO_CURRENT_VCLIBS_VERSION=1 -D_CONSOLE -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -ID:\a\hdf5\hdf5\hdfsrc\src -ID:\a\hdf5\hdf5\hdfsrc\src\H5FDsubfiling -ID:\a\hdf5\build\src -std:c11 /DWIN32 /D_WINDOWS -wd5105 /Zi /Ob0 /Od /RTC1 -Og -std:c11 -MD /Wall /wd4668 /showIncludes /Fosrc\CMakeFiles\hdf5-static.dir\H5.c.obj /Fdbin\libhdf5_D.pdb /FS -c D:\a\hdf5\hdf5\hdfsrc\src\H5.c
cl : Command line warning D9035 : option 'Og' has been deprecated and will be removed in a future release
cl : Command line error D8016 : '/RTC1' and '/Og' command-line options are incompatible

byrnHDF:
Fixed the windows issue but now have one test timeout:
8/9 Test #12: H5TESTXPR-fheap ..................***Timeout 3600.02 sec

jhendersonHDF:
> Fixed the windows issue but now have one test timeout: 8/9 Test #12: H5TESTXPR-fheap ..................***Timeout 3600.02 sec

Just removing the flag there removes it for all the other compilers than GCC which is not exactly what I'd intended, but the flags for the other compilers also do need to be reviewed. I'm not too surprised by the timeout, but we'll need to make sure the test is actually making progress and then deal with the timeout somehow.

byrnHDF:
> Just removing the flag there removes it for all the other compilers than GCC

Right, but it was interfering more then helping. And probably should be in the compiler specific files.

lrknox (approved):

.github/workflows/dev-cmake.yml
```diff
@@ -0,0 +1,346 @@
...
+    #
+    name: "Ubuntu Clang CMake"
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install CMake Dependencies (Linux_clang)
+        run: |
+          sudo apt-get update
+          sudo apt-get install ninja-build doxygen graphviz curl libtinfo5
+
+      - name: add clang to env
```
jhendersonHDF, line 268: This is fine for now I suppose, but we really should stop installing things that are already there and just use the defaults. Ubuntu has clang and installing a specific version here just adds maintenance for no reason. Same with forcing a particular version of GCC, installing and using Ninja (for some reason), etc. We're just moving our builds away from what most people are going to be using.

jhendersonHDF (approved):

.github/workflows/dev-cmake.yml
byrnHDF: Sure - just didn't think about it and just copy-pasted from another workflow.

PR #5397, opened 2025-03-20
S3 credentials order/testing and add S3 object URIs support

### LINKED ISSUE #4493: allow AWS_* environment variables to replace `--s3-cred` for ROS3 driver
Short term problem: Sending S3 credentials to the ROS3 driver for h5ls and h5dump command line tools is cumbersome and not fully documented.
Long term problem: Our software should eventually rely on hdf5 using the ROS3 driver when called via netcdf-4. We think this is more likely to happen if the ROS3 driver picks up S3 credentials from environment variables or configuration files [standardized by the AWS-CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html).

The following currently (using hdf5 built from [develop](https://github.com/HDFGroup/hdf5/tree/6203a4455e27d9497f1e89eefd36885692910c20)) returns `https://ob-cumulus-prod-public.s3.us-west-2.amazonaws.com/PACE_OCI.20240411T164652.L1B.nc: unable to open file`
```
export AWS_REGION=us-west-2
export AWS_ACCESS_KEY_ID=<key>
export AWS_SECRET_ACCESS_KEY=<secret>
export AWS_SESSION_TOKEN=<token>
~/.local/bin/h5ls --vfd=ros3 https://ob-cumulus-prod-public.s3.us-west-2.amazonaws.com/PACE_OCI.20240411T164652.L1B.nc
```
Anyone can create an Earthdata account at https://urs.earthdata.nasa.gov/, and get the temporary credentials for this bucket at https://obdaac-tea.earthdatacloud.nasa.gov/s3credentials.

For the above `h5ls` call to work, I find I have to include the credentials directly.
```
~/.local/bin/h5ls --vfd=ros3 --s3-cred="($AWS_REGION,$AWS_ACCESS_KEY_ID,$AWS_SECRET_ACCESS_KEY,$AWS_SESSION_TOKEN)" https://ob-cumulus-prod-public.s3.us-west-2.amazonaws.com/PACE_OCI.20240411T164652.L1B.nc
```
Note that the inclusion of 4 comma-separated strings is overlooked in the `h5ls -h` and `h5dump -h` documentation. Configuration files described at the link above should be similarly supported.

### LINKED ISSUE #4600: Support S3 object URIs in ROS3

S3 objects are more commonly referenced as `s3://bucket_name/object_key`. ROS3 virtual file driver uses S3 object URLs instead in the form: `https://s3.<aws_region>.amazonaws.com/<bucket_name>/<object_key>`. Constructing an object's URL requires extracting bucket name and object key from its URI. AWS region comes from the usual sources (config file or env. variable) unless specified by the user application. Bucket names are up to 63 bytes long, while object keys can be up to 1024 bytes long.

Be able to use S3 object URIs when working with HDF5 files in S3 cloud stores. Example:

    h5ls -r --vfd=ros3 s3://my-bucket/a/b/c/file.h5.

Write code to parse S3 object URI and assemble object's URL prior to invoking libhdf5 or its tools. This is what h5py currently does.

**Additional context**

Below is how ChatGPT would extract bucket and key names.

```c
#include <stdio.h>
#include <string.h>

// Function to parse S3 URI
void parseS3URI(const char *uri, char *bucket_name, char *object_key) {
    // Check if the URI starts with "s3://"
    if (strncmp(uri, "s3://", 5) != 0) {
        printf("Invalid S3 URI format. It should start with 's3://'\n");
        return;
    }

    // Skip "s3://"
    const char *path_start = uri + 5;

    // Find the '/' delimiter to separate bucket name and object key
    const char *slash_pos = strchr(path_start, '/');
    if (slash_pos == NULL) {
        printf("Invalid S3 URI format. Missing '/' after bucket name\n");
        return;
    }

    // Calculate lengths
    size_t bucket_len = slash_pos - path_start;
    size_t key_len = strlen(slash_pos + 1);

    // Copy bucket name
    strncpy(bucket_name, path_start, bucket_len);
    bucket_name[bucket_len] = '\0';

    // Copy object key
    strncpy(object_key, slash_pos + 1, key_len);
    object_key[key_len] = '\0';
}

int main() {
    // Example S3 URI
    const char *s3_uri = "s3://my-test-bucket/my-folder/my-file.h5";

    // Buffer to store parsed components
    char bucket_name[64];
    char object_key[1025];

    // Parse the S3 URI
    parseS3URI(s3_uri, bucket_name, object_key);

    // Print the parsed components
    printf("Bucket Name: %s\n", bucket_name);
    printf("Object Key: %s\n", object_key);

    return 0;
}
```

byrnHDF:
Replaces #5346 
Fixes #4600
Fixes #4493


src/H5FDros3.c
```diff
@@ -680,6 +688,215 @@ H5Pset_fapl_ros3_token(hid_t fapl_id, const char *token)
...
+ *
+ * Purpose:     Modify the file access property list to use the H5FD_ROS3
+ *              driver defined in this source file by adding or
+ *              modifying the endpoint url property.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Pset_fapl_ros3_endpoint(hid_t fapl_id, const char *endpoint_url)
```
jhendersonHDF, line 809: Why add in a new API function for setting an endpoint when that's generally passed in via `H5Fcreate`/`H5Fopen` anyway?

src/H5FDros3.h
```diff
@@ -97,13 +102,17 @@
...
+ * \var char H5FD_ros3_fapl_t::session_token[H5FD_ROS3_MAX_SECRET_TOK_LEN + 1]
+ *      A string which specifies the security token.
  */
 typedef struct H5FD_ros3_fapl_t {
     int32_t version;
     hbool_t authenticate;
     char    aws_region[H5FD_ROS3_MAX_REGION_LEN + 1];
     char    secret_id[H5FD_ROS3_MAX_SECRET_ID_LEN + 1];
     char    secret_key[H5FD_ROS3_MAX_SECRET_KEY_LEN + 1];
+    char    session_token[H5FD_ROS3_MAX_SECRET_TOK_LEN + 1];
```
jhendersonHDF, line 115: There's already the `H5Pset_fapl_ros3_token` function/property which was intended to cover the case of passing in session tokens. I'm also generally against this type of configuration structure because it gets to the point you can't instantiate one on the stack without compiler warnings.

src/H5FDros3.c
```diff
@@ -949,6 +1179,16 @@ H5FD__ros3_cmp(const H5FD_t *_f1, const H5FD_t *_f2)
     else if (f2->fa.secret_key[0] != '\0')
         HGOTO_DONE(-1);
 
+    /* FAPL: SESSION_TOKEN */
+    if (f1->fa.session_token[0] != '\0' && f2->fa.session_token[0] != '\0') {
```
jhendersonHDF, line 1183: I don't think it necessarily makes sense to compare two files based on a session token and would generally be in favor of removing the `session_token` field from the structure and just passing around the value retrieved from the property.

test/hdfs.c
```diff
@@ -28,6 +28,9 @@
 #define HDFS_TEST_DEBUG        0
 #define HDFS_TEST_MAX_BUF_SIZE 256
 
+#define S3_TEST_RESOURCE_TEXT_RESTRICTED_SIZE 5582655
+#define S3_TEST_RESOURCE_TEXT_PUBLIC_SIZE     6589
```
jhendersonHDF, line 32: Seems like the changes to this file may have been unintentional. Are these files in S3 considering that this is the HDFS test? The values for the macros also don't seem to match up with the values they're replacing below.

src/H5FDs3comms.c
```diff
@@ -753,13 +757,15 @@ H5FD__s3comms_s3r_getsize(s3r_t *handle)
...
  *              fapl_token can be NULL
+ *              fapl_endpoint can be NULL
  *
  * Return:      SUCCESS:    Pointer to new request handle.
  *              FAILURE:    NULL
  *----------------------------------------------------------------------------
  */
 s3r_t *
-H5FD__s3comms_s3r_open(const char *url, const H5FD_ros3_fapl_t *fa, const char *fapl_token)
+H5FD__s3comms_s3r_open(const char *url, const H5FD_ros3_fapl_t *fa, const char *fapl_token,
```
jhendersonHDF, line 767: This seems a bit odd to add an extra parameter for an endpoint that is generally just going to be `NULL` when it should just be what's passed in `url` in the first place.

tools/src/h5ls/h5ls.h
```diff
@@ -70,6 +70,9 @@
  * \li <strong>--page-buffer-size=N</strong> Set the page buffer cache size, N=non-negative integers
  * \li <strong>--vfd=DRIVER</strong>   Use the specified virtual file driver
  * \li <strong>--hexdump</strong>      Show raw data in hexadecimal format
+ * \li <strong>--endpoint-url=P</strong> Supply S3 endpoint url information to "ros3" vfd.
+ *                   P is the AWS service endpoint.
+ *                   Has no effect if vfd flag not set to "ros3".
```
jhendersonHDF, line 78: I'm very much against continuing to add options to the tools that are only relevant to one particular VFD. If there's a good use case for it, I think the whole "endpoint URL" feature should be split out into a separate PR and discussed there. The idea of updating the VFD interface to account for VFD-specific configuration was brought up somewhat recently and I think this is a good fit for that.

PR #5401, opened 2025-03-21
Fortran configury: Ensure F08 real_kinds are actually valid kinds

### LINKED ISSUE #5090: Compilation failure on aarch64 with Arm Compiler for Linux's flang
Using Spack to set up config and compile, the following error occurs: 5 errors found in build log.

It makes no difference if I disable the non-standard float16 - using an arg to cmake of: `-DHDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16:BOOL=OFF`

```
5 errors found in build log:
     3047    F90-S-0081-Illegal selector - KIND parameter has unknown value for data type  (/tmp/user/spack-stage/s
             pack-stage-hdf5-1.14.5-rzhstpbozkrcypkkfn5ntevpg2gwfof7/spack-build-rzhstpb/fortran/static/H5_gen.F90: 604
             9)
     3048      0 inform,   0 warnings,   1 severes, 0 fatal for h5pget_kind_16
     3049    F90-S-0081-Illegal selector - KIND parameter has unknown value for data type  (/tmp/user/spack-stage/s
             pack-stage-hdf5-1.14.5-rzhstpbozkrcypkkfn5ntevpg2gwfof7/spack-build-rzhstpb/fortran/static/H5_gen.F90: 609
             7)
     3050      0 inform,   0 warnings,   1 severes, 0 fatal for h5pregister_kind_16
     3051    F90-S-0081-Illegal selector - KIND parameter has unknown value for data type  (/tmp/user/spack-stage/s
             pack-stage-hdf5-1.14.5-rzhstpbozkrcypkkfn5ntevpg2gwfof7/spack-build-rzhstpb/fortran/static/H5_gen.F90: 614
             5)
     3052      0 inform,   0 warnings,   1 severes, 0 fatal for h5pinsert_kind_16
  >> 3053    make[2]: *** [fortran/src/CMakeFiles/hdf5_fortran-static.dir/build.make:325: fortran/src/CMakeFiles/hdf5_f
             ortran-static.dir/__/static/H5_gen.F90.o] Error 1
     3054    make[2]: Leaving directory '/tmp/user/spack-stage/spack-stage-hdf5-1.14.5-rzhstpbozkrcypkkfn5ntevpg2gw
             fof7/spack-build-rzhstpb'
```
This is using Arm Compiler for Linux - which is LLVM derivative but the NVIDIA/PGI fortran Flang.

As the float16 is disabled, it ought not to be trying to generate it?

lrbison:
Closes https://github.com/HDFGroup/hdf5/issues/5090

### DIFF
--- a/m4/aclocal_fc.f90
+++ b/m4/aclocal_fc.f90
@@ -183,8 +183,11 @@ END PROGRAM FC_AVAIL_KINDS
 PROGRAM FC08_AVAIL_KINDS
       USE, INTRINSIC :: ISO_FORTRAN_ENV, ONLY : stdout=>OUTPUT_UNIT, integer_kinds, real_kinds, logical_kinds
       IMPLICIT NONE
-      INTEGER :: ik, jk, k, max_decimal_prec
+      INTEGER :: ik, jk, k, kk, max_decimal_prec
       INTEGER :: num_rkinds, num_ikinds, num_lkinds
+      LOGICAL :: found_rkinds( 1:SIZE(real_kinds) )
+      CHARACTER(LEN=1) :: sep
+
 
       ! Find integer KINDs
 
@@ -203,28 +206,33 @@ PROGRAM FC08_AVAIL_KINDS
 
       num_rkinds = SIZE(real_kinds)
 
+      ! some compilers (ACfL 24) reported REAL=16 kind, but refused to
+      ! compile with it.  Verify the kind can be selected in SELECTED_REAL_KIND.
+      found_rkinds(:) = .FALSE.
       max_decimal_prec = 1
 
       prec: DO ik = 2, 36
          exp: DO jk = 1, 700
             k = SELECTED_REAL_KIND(ik,jk)
             IF(k.LT.0) EXIT exp
+            DO kk = 1,num_rkinds
+               IF (real_kinds(kk) == k) found_rkinds(kk) = .TRUE.
+            ENDDO
             max_decimal_prec = ik
          ENDDO exp
       ENDDO prec
 
+      sep = ""
       DO k = 1, num_rkinds
-         WRITE(stdout,'(I0)', ADVANCE='NO') real_kinds(k)
-         IF(k.NE.num_rkinds)THEN
-            WRITE(stdout,'(A)',ADVANCE='NO') ','
-         ELSE
-            WRITE(stdout,'()')
-         ENDIF
+         IF(.NOT. found_rkinds(k)) CYCLE
+         WRITE(stdout,'(A,I0)', ADVANCE='NO') TRIM(sep), real_kinds(k)
+         sep = ","
       ENDDO
+      WRITE(stdout,'()')
 
      WRITE(stdout,'(I0)') max_decimal_prec
      WRITE(stdout,'(I0)') num_ikinds
-     WRITE(stdout,'(I0)') num_rkinds
+     WRITE(stdout,'(I0)') COUNT(found_rkinds)
 
      ! Find logical KINDs
 


m4/aclocal_fc.f90
```diff
@@ -203,24 +206,29 @@ PROGRAM FC08_AVAIL_KINDS
...
       max_decimal_prec = 1
 
       prec: DO ik = 2, 36
          exp: DO jk = 1, 700
             k = SELECTED_REAL_KIND(ik,jk)
             IF(k.LT.0) EXIT exp
+            do kk = 1,num_rkinds
+               if (real_kinds(kk) == k) found_rkinds(kk) = .true.
+            end do
             max_decimal_prec = ik
```
brtnfld, line 221: If the KIND  is not supported, then the maximum decimal precision, max_decimal_prec, should not be set for that KIND.

brtnfld (changes requested):

brtnfld:
For consistency, the same changes should be made to FC_AVAIL_KINDS.

m4/aclocal_fc.f90
```diff
@@ -183,8 +183,11 @@ END PROGRAM FC_AVAIL_KINDS
 PROGRAM FC08_AVAIL_KINDS
       USE, INTRINSIC :: ISO_FORTRAN_ENV, ONLY : stdout=>OUTPUT_UNIT, integer_kinds, real_kinds, logical_kinds
       IMPLICIT NONE
-      INTEGER :: ik, jk, k, max_decimal_prec
+      INTEGER :: ik, jk, k, kk, max_decimal_prec
       INTEGER :: num_rkinds, num_ikinds, num_lkinds
+      logical :: found_rkinds( 1:SIZE(real_kinds) )
```
brtnfld: please use uppercase keyword intrinsics

brtnfld:
I have to ensure this does not suppress detected kinds of other compilers since that scenario would not cause tests to fail.

m4/aclocal_fc.f90
lrbison: Addressed.

lrbison:
> please use uppercase keyword intrinsics

Done!

> I have to ensure this does not suppress detected kinds of other compilers since that scenario would not cause tests to fail.

Sure.  In my testing I saw that GNU Fortran was unaffected.  You can see [output here](https://github.com/HDFGroup/hdf5/issues/5090#issuecomment-2744303359).

For these next two points let me clarify that SELECTED_REAL_KIND is correct in ACfL, it is only the F08 ISO_FORTRAN_ENV array REAL_KINDS that was wrong.  That array was _only_ used to print the real kinds the compiler supported (not max precision), so I don't think the following comments items require a change:

> If the KIND is not supported, then the maximum decimal precision, max_decimal_prec, should not be set for that KIND.

The REAL_KINDS array was not used in the loop to determine max decimal precision, so it is already correct.

> For consistency, the same changes should be made to FC_AVAIL_KINDS.

The non-F08 version of this loop builds the list of kinds based on the return values of SELECTED_REAL_KIND, so it never had the problem.

... however it makes me wonder if the preferred fix is just to ignore REAL_KINDS altogether and use FC_AVAIL_KINDS's logic to find real kinds.

brtnfld:
> The non-F08 version of this loop builds the list of kinds based on the return values of SELECTED_REAL_KIND, so it never had the problem.

I'm not sure the compiler does not also have an issue with SELECTED_REAL_KIND, as from our discussion, #5090, 

```
k = SELECTED_REAL_KIND(19,2)
print*,k
end
```

Printed -1, so it does not support KIND=16.

lrbison:
>Printed -1, so it does not support KIND=16.

Correct.  ACfL does not support `KIND=16`, yet it includes 16 in `REAL_KINDS` (a compiler bug).  This caused HDF5 configuration to make code with `REAL(KIND=16)` which then fails to compile.  I am suggesting this patch to work around the compiler bug.

lrbison:
@brtnfld so are there other changes you think are required?

brtnfld:
`num_rkinds` gets its value from `size(real_kinds)`, which is going to include KIND=16.
`max_decimal_prec` is going to be for KIND=16, and not KIND=8, or whatever the highest kind supported.

lrbison:
@brtnfld Yes, `num_rkinds` will be `size(real_kinds)`, but in the loop before we print the kinds we check the new logical array with `IF(.NOT. found_rkinds(k)) CYCLE`.  I've also updated the logic to ensure we don't print trailing commas, double commas, or anything like that.

>  max_decimal_prec is going to be for KIND=16, and not KIND=8, or whatever the highest kind supported.

No.  On line 216 `k = SELECTED_REAL_KIND(ik,jk)` never selects kind=16 in ACfL, therefore the max_precision remains correct.  When `ik` goes above 15 selected_real_kind  returns -1 and the existing logic exits the loop and max_decimal_prec is never updated further.  This is also seen from the issue where I show the test output has `-- ....MAX DECIMAL PRECISION 15` rather than `33` as it does in GFortran.

brtnfld:
` num_rkinds ` is printed later in the program, so it should reflect the number of real KINDS actually printed.

lrbison:
> `num_rkinds` is printed later

I see, I had missed that, thank you for pointing it out.  Testing with `cmake -S ../hdf5 -DCMAKE_Fortran_COMPILER=gfortran/armflang -DHDF5_BUILD_FORTRAN=ON`: Gfortran cmake prints:

```
-- ....NUMBER OF INTEGER KINDS FOUND 5
-- ....REAL KINDS FOUND {4,8,16}
-- ....INTEGER KINDS FOUND {1,2,4,8,16}
-- ....MAX DECIMAL PRECISION 33
-- ....LOGICAL KINDS FOUND {1,2,4,8,16}
-- ....FOUND SIZEOF for REAL KINDs {4,8,16}
```

and ACfL prints
```
-- ....NUMBER OF INTEGER KINDS FOUND 4
-- ....REAL KINDS FOUND {4,8}
-- ....INTEGER KINDS FOUND {1,2,4,8}
-- ....MAX DECIMAL PRECISION 15
-- ....LOGICAL KINDS FOUND {1,2,4,8}
-- ....FOUND SIZEOF for REAL KINDs {4,8}
```

The test program itself prints:

```
gfortran ./test.f90 && ./a.out
1,2,4,8,16
4,8,16
33
5
3
5
1,2,4,8,16
```

```
armflang ./test.f90 && ./a.out
1,2,4,8
4,8
15
4
2
4
1,2,4,8
```

lrbison:
@brtnfld fixed and rebased

lrbison:
@brtnfld can you take another look?

brtnfld (approved):
LGTM, thanks for the fix.\!

mattjala (approved):

lrknox (approved):

PR #5404, opened 2025-03-26
Enable building of test generator programs against shared libraries

jhendersonHDF:
Build test generator programs against shared libraries, if available, by default. Otherwise, build them against static libraries. If HDF5_BUILD_STATIC_TOOLS is set to ON, the tools test generator programs will be built against static libraries even if shared libraries are available.


jhendersonHDF:
This allows the test generator programs to be built when only building shared libraries, which is a little more convenient when working on tools test files.

tools/test/h5copy/CMakeLists.txt
```diff
@@ -4,11 +4,16 @@ project (HDF5_TOOLS_TEST_H5COPY C)
 # --------------------------------------------------------------------
 # Add the h5copy test executables
 # --------------------------------------------------------------------
-if (HDF5_BUILD_GENERATORS AND BUILD_STATIC_LIBS)
+if (HDF5_BUILD_GENERATORS)
   add_executable (h5copygentest ${HDF5_TOOLS_TEST_H5COPY_SOURCE_DIR}/h5copygentest.c)
   target_include_directories (h5copygentest PRIVATE "${HDF5_SRC_INCLUDE_DIRS};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>")
-  TARGET_C_PROPERTIES (h5copygentest STATIC)
-  target_link_libraries (h5copygentest PRIVATE ${HDF5_LIB_TARGET})
+  if (HDF5_BUILD_STATIC_TOOLS)
```
jhendersonHDF, line 10: One of the existing tools test generator programs, h5watchgentest,  used `HDF5_BUILD_STATIC_TOOLS` to determine whether to build against static or shared. I used the same for the other tools, though it may make more sense just to use `BUILD_STATIC_LIBS`/`BUILD_SHARED_LIBS`.

hl/test/CMakeLists.txt
```diff
@@ -87,16 +87,25 @@ endif ()
...
   add_executable (hl_gen_test_ds gen_test_ds.c)
   target_compile_options(hl_gen_test_ds PRIVATE "${HDF5_CMAKE_C_FLAGS}")
   target_include_directories (hl_gen_test_ds PRIVATE "${HDF5_SRC_INCLUDE_DIRS};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>")
-  TARGET_C_PROPERTIES (hl_gen_test_ds STATIC)
-  target_link_libraries (hl_gen_test_ds PRIVATE
-      ${HDF5_HL_LIB_TARGET}
-      ${HDF5_TEST_LIB_TARGET}
-      ${HDF5_LIB_TARGET}
-  )
+  if (NOT BUILD_SHARED_LIBS)
```
jhendersonHDF, line 94: Slightly different logic for static vs. shared here by using NOT, but this just matches the existing logic in the file.

byrnHDF:
Makes sense - I think at the time autotools only generated with static libs.

byrnHDF (approved):

mattjala (approved):

PR #5406, opened 2025-03-27
Fix incorrect link size for VOL connectors in h5ls/h5diff

mattjala:
h5ls and h5diff operate on the value of external links retrieved through the top-level `H5Lget_val()`. In their calls to `H5Lunpack_elink_val()`, these tools provide the size of the external link value as `linfo->u.val_size`, a value passed in from the sub-VOL connector layers of the library.

When using the Native VOL or a VOL connector that doesn't modify external link values, this works fine since the external link size reported by the API agrees with the link size that the sub-Native VOL layers see. 

When using a passthrough connector that changes the external link values,  the link size from outside will reflect the 'internal' link size, and the link value will reflect the externally visible link value, leading to bad memory reads depending on how different the real size is.

This PR changes the tools to use the externally-visible link size  (retrieved via `H5tools_get_symlink_info() -> H5Lget_info2()`) for consistency with the externally-visible link value.

### DIFF
--- a/tools/lib/h5diff.c
+++ b/tools/lib/h5diff.c
@@ -521,7 +521,8 @@ trav_grp_symlinks(const char *path, const H5L_info2_t *linfo, void *udata)
                 H5TOOLS_GOTO_DONE(SUCCEED);
             }
 
-            if (H5Lunpack_elink_val(lnk_info.trg_path, linfo->u.val_size, NULL, &ext_fname, &ext_path) < 0)
+            if (H5Lunpack_elink_val(lnk_info.trg_path, lnk_info.linfo.u.val_size, NULL, &ext_fname,
+                                    &ext_path) < 0)
                 H5TOOLS_GOTO_DONE(SUCCEED);
 
             /* check if already visit the target object */
--- a/tools/src/h5ls/h5ls.c
+++ b/tools/src/h5ls/h5ls.c
@@ -2441,7 +2441,7 @@ list_lnk(const char *name, const H5L_info2_t *linfo, void *_iter)
             else if (no_dangling_link_g && ret == 0)
                 iter->symlink_list->dangle_link = true;
 
-            if (H5Lunpack_elink_val(buf, linfo->u.val_size, NULL, &filename, &path) < 0)
+            if (H5Lunpack_elink_val(buf, lnk_info.linfo.u.val_size, NULL, &filename, &path) < 0)
                 goto done;
 
             h5tools_str_append(&buffer, "External Link {");


qkoziol (approved):

jhendersonHDF:
When do the values in `linfo->u.val_size` and `lnk_info.linfo.u.val_size` start to differ from each other? Or rather, do you happen to have an example of the problem? `lnk_info.linfo.u.val_size` comes from `H5Lget_info2()` which goes through the 'link get' callback and `linfo->u.val_size` comes from the connector populating a `H5L_info2_t` structure during link iteration, which often ends up using the same logic as the 'link get' callback to populate the structure. I'm just wondering why the two values aren't the same in this case.

mattjala:
> When do the values in `linfo->u.val_size` and `lnk_info.linfo.u.val_size` start to differ from each other? Or rather, do you happen to have an example of the problem? `lnk_info.linfo.u.val_size` comes from `H5Lget_info2()` which goes through the 'link get' callback and `linfo->u.val_size` comes from the connector populating a `H5L_info2_t` structure during link iteration, which often ends up using the same logic as the 'link get' callback to populate the structure. I'm just wondering why the two values aren't the same in this case.

I ran into this issue when using a connector I made for testing purposes that appends a suffix to filenames under the hood. For consistency, the connector also has to manage appending/removing that suffix during external link operations. Using h5ls on a file with external links created through that VOL connector led to a read from bad memory:

```
==4169618== Invalid read of size 1
==4169618==    at 0x4CEF99C: H5Lunpack_elink_val (H5L.c:1936)
==4169618==    by 0x1140A6: list_lnk (h5ls.c:2444)
==4169618==    by 0x48B233A: traverse_cb (h5trav.c:220)
==4169618==    by 0x4C35C2A: H5G__visit_cb (H5Gint.c:1027)
==4169618==    by 0x4C3A543: H5G__link_iterate_table (H5Glink.c:520)
==4169618==    by 0x4C22EBB: H5G__dense_iterate (H5Gdense.c:966)
==4169618==    by 0x4C4F60A: H5G__obj_iterate (H5Gobj.c:647)
==4169618==    by 0x4C36A92: H5G_visit (H5Gint.c:1249)
==4169618==    by 0x519B556: H5VL__native_link_specific (H5VLnative_link.c:373)
==4169618==    by 0x51670E4: H5VL__link_specific (H5VLcallback.c:5558)
==4169618==    by 0x51677AD: H5VLlink_specific (H5VLcallback.c:5633)
==4169618==    by 0x538DCB2: debug_pt_link_specific (debug_pt_vol_connector.c:2219)
==4169618==  Address 0x57e94c6 is 8 bytes after a block of size 30 alloc'd
==4169618==    at 0x484DA83: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==4169618==    by 0x48B1938: H5tools_get_symlink_info (h5tools_utils.c:870)
==4169618==    by 0x11402F: list_lnk (h5ls.c:2433)
==4169618==    by 0x48B233A: traverse_cb (h5trav.c:220)
==4169618==    by 0x4C35C2A: H5G__visit_cb (H5Gint.c:1027)
==4169618==    by 0x4C3A543: H5G__link_iterate_table (H5Glink.c:520)
==4169618==    by 0x4C22EBB: H5G__dense_iterate (H5Gdense.c:966)
==4169618==    by 0x4C4F60A: H5G__obj_iterate (H5Gobj.c:647)
==4169618==    by 0x4C36A92: H5G_visit (H5Gint.c:1249)
==4169618==    by 0x519B556: H5VL__native_link_specific (H5VLnative_link.c:373)
==4169618==    by 0x51670E4: H5VL__link_specific (H5VLcallback.c:5558)
==4169618==    by 0x51677AD: H5VLlink_specific (H5VLcallback.c:5633)
```

The link size for the allocation was provided by low-level H5G routines, which saw the 'real' filename that had the suffix, so the provided link size was longer than the actual value of the buffer retrieved from `H5Lget_val()`. This led to a bad read when it tried to verify that the buffer ended in a null byte.

jhendersonHDF:
Hmm if I'm understanding correctly this seems like it might be bad behavior from the VOL connector. Based on the stack trace, the size provided for the allocation directly comes from the result of `H5Lget_info2()` which goes through the connector's 'link get' callback; the size shouldn't be provided by any H5G routines here.

jhendersonHDF:
The changes make sense to me and seem correct; I can see where the tools aren't quite correctly matching up the size they should be using when calling `H5Lunpack_elink_val` and should be using what was stored from `H5tools_get_symlink_info`. I'm just trying to work through whether this is something that should be "allowed" as it seems a simple case that other similarly-written applications could run into.

hyoklee (approved):

mattjala:
> Hmm if I'm understanding correctly this seems like it might be bad behavior from the VOL connector. Based on the stack trace, the size provided for the allocation directly comes from the result of `H5Lget_info2()` which goes through the connector's 'link get' callback; the size shouldn't be provided by any H5G routines here.

I don't think this is bad behavior on the connector's part. The allocation does correctly go through the VOL layer, but the size that the tool expects the allocation to be comes from `debug_pt_link_specific (Iterate) -> ... -> H5G__obj_iterate() -> H5G__iterate_cb() -> H5G_link_to_info() -> H5L__extern_query()`. These act on the underlying value with the suffix, and pass that suffix-aware length to be used with the suffix-unaware allocation.

jhendersonHDF:
The tools clearly seem to have been making the assumption that the size for a link's value that is received for a link during iteration on the links in a group will be the same as the size retrieved from calling `H5Lget_info2()` on that link. That seems reasonable to me and the fact that that may not hold due to changes made by a passthrough connector above is what concerns me.

jhendersonHDF (approved):
Approved because the changes make sense regardless of any potential underlying issues. It would be good to create a minimal example of this problem and maybe dig deeper into it though; I still feel like there's an issue here around whether it makes sense that the two link value sizes can differ between iteration and `H5Lget_info2()` calls.

PR #5407, opened 2025-03-27
Improve the default plugin path #5399

### LINKED ISSUE #5399: When building plugins inline set the default plugin path
When building plugins with the hdf5 library build, we know where the plugins will be installed. Override the default PLUGIN PATh by setting H5_PLUGIN_PATH to the install location.

byrnHDF:
Add CMAKE_INSTALL_PREFIX to the default plugin path
Closes #5399

### DIFF
--- a/config/cmake/ConfigureChecks.cmake
+++ b/config/cmake/ConfigureChecks.cmake
@@ -547,9 +547,9 @@ set (${HDF_PREFIX}_DEFAULT_VFD H5FD_SEC2)
 
 if (NOT DEFINED "${HDF_PREFIX}_DEFAULT_PLUGINDIR")
   if (WINDOWS)
-    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "%ALLUSERSPROFILE%\\\\hdf5\\\\lib\\\\plugin")
+    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "${CMAKE_INSTALL_PREFIX}\\\\lib\\\\plugin;%ALLUSERSPROFILE%\\\\hdf5\\\\lib\\\\plugin")
   else ()
-    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "/usr/local/hdf5/lib/plugin")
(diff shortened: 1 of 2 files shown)


lrknox (approved):

vchoi-hdfgroup (approved):

config/cmake/ConfigureChecks.cmake
```diff
@@ -547,9 +547,9 @@ set (${HDF_PREFIX}_DEFAULT_VFD H5FD_SEC2)
 
 if (NOT DEFINED "${HDF_PREFIX}_DEFAULT_PLUGINDIR")
   if (WINDOWS)
-    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "%ALLUSERSPROFILE%\\\\hdf5\\\\lib\\\\plugin")
+    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "${CMAKE_INSTALL_PREFIX}\\\\lib\\\\plugin;%ALLUSERSPROFILE%\\\\hdf5\\\\lib\\\\plugin")
   else ()
-    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "/usr/local/hdf5/lib/plugin")
+    set (${HDF_PREFIX}_DEFAULT_PLUGINDIR "${CMAKE_INSTALL_PREFIX}/lib/plugin:/usr/local/hdf5/lib/plugin")
```
jhendersonHDF, line 552: Won't this have issues with lib64?

byrnHDF: Likely not, because the plugins and libs are stored under the default install location. The actual setting can be changed by variables. Of greater concern, is the established path manipulations that happen in configure phase. Those assumptions and settings probably need reviewed and updated.

jhendersonHDF: > Likely not, because the plugins and libs are stored under the default install location.

But the default install location could just as likely be under lib64 couldn't it?

> The actual setting can be changed by variables. Of greater concern, is the established path manipulations that happen in configure phase. Those assumptions and settings probably need reviewed and updated.

Agreed.

PR #5413, opened 2025-04-02
ph5diff: remove message that breaks MPI_TEST_H5DIFF-h5diff_601

andreas-schwab:
Start 225: MPI_TEST_H5DIFF-h5diff_601
1/1 Test #225: MPI_TEST_H5DIFF-h5diff_601 .......***Failed    7.82 sec
-- COMMAND:  /usr/lib64/mpi/gcc/openmpi5/bin/mpiexec -n;1;;/home/abuild/rpmbuild/BUILD/hdf5-openmpi5-1.14.6-build/hdf5-1.14.6/build/bin/ph5diff;;h5diff_basic1.h5;h5diff_basic1.h5;nono_obj
-- COMMAND Result: 0
-- COMMAND Error:
Only 1 task available...doing serial diff

CMake Error at /home/abuild/rpmbuild/BUILD/hdf5-openmpi5-1.14.6-build/hdf5-1.14.6/config/cmake/grepTest.cmake:119 (message):
  Failed: The error output of /usr/lib64/mpi/gcc/openmpi5/bin/mpiexec did not
  contain 'Object could not be found'.  Error output was: 'Only 1 task
  available...doing serial diff

  '

### DIFF
--- a/tools/src/h5diff/ph5diff_main.c
+++ b/tools/src/h5diff/ph5diff_main.c
@@ -63,8 +63,6 @@ main(int argc, char *argv[])
     g_Parallel    = 1;
 
     if (g_nTasks == 1) {
-        fprintf(stderr, "Only 1 task available...doing serial diff\n");
-
         g_Parallel = 0;
 
         parse_command_line(argc, (const char *const *)argv, &fname1, &fname2, &objname1, &objname2, &opts);


byrnHDF:
That should not be the problem, because that test is doing a grep for a phrase. However, there was an issue with the text it was searching for and has been fixed in development branch.

andreas-schwab:
Why do you think this changes anything?

byrnHDF:
No this is not the answer - the correct solution is that 601 should be using a grep test and not a compare test.

byrnHDF:
Aha - the problem is not that text - the problem is that the phrase to be grepped is in stdout - not sterr.

byrnHDF:
Closing this in favor of #5416

PR #5416, opened 2025-04-04
Correct fix for h5diff-601 tools test

byrnHDF:
The h5diff tests that grep for an error result will look in the stderr file first and fail because the text is actually in stdout.


tools/test/h5diff/CMakeTests.cmake
```diff
@@ -533,11 +533,9 @@
               #-D "TEST_EXPECT=${resultcode}"
               -D "TEST_EXPECT=0" # ph5diff currently always exits with a zero status code due to
```
hyoklee, line 534: Why does comment say `# ph5diff` for `tools/test/h5diff`?

mattjala:
Why are the error results from h5diff pushed to stdout instead of stderr?

byrnHDF:
> Why are the error results from h5diff pushed to stdout instead of stderr?

Need to ask the author of that decision. ?? Actually, there is a difference between reporting a difference and reporting an error. This, "object not found", is the former.

tools/test/h5diff/CMakeTests.cmake
byrnHDF: ph5diff runs the same tests as h5diff except in parallel mode. And this is the parallel test macro.

jhendersonHDF:
I think the situation is being interpreted backward here. The message _is_ in `stderr`. It was in `stdout` but I changed that because `stderr` is better suited for informational messages that aren't part of the main program output so that they can be filtered out with redirects. It's not an error (and `stderr` is somewhat misnamed IMO), but it really shouldn't be in `stdout` either.

tools/test/h5diff/CMakeTests.cmake
```diff
@@ -533,11 +533,9 @@
...
                                  # output from some MPI implementations from a non-zero exit code
-              -D "TEST_REFERENCE=${resultfile}.txt"
-              -D "TEST_ERRREF=${result_errcheck}"
+              -D "TEST_REFERENCE=${result_errcheck}"
```
mattjala, line 536: I don't think this change makes sense. grepTest expects `TEST_REFERENCE` to be the filename of the reference file to compare to `${TEST_OUTPUT}`, and `result_errcheck` in this case is the string `could not be found`.

mattjala:
The `parallel_print()` routine used by h5diff writes the "not found" message to stdout in serial, but to stderr in parallel. `grepTest.cmake` always accepts an empty error file, but if there is any error output, then it requires that it contain the error reference - in this case, the string `could not be found`. 

The serial h5diff tests normally work because all output goes to stdout, and stderr is empty and unchecked. The parallel h5diff tests normally work because stderr contained the error string 'could not be found'. The run in #5413 failed because, when h5diff detects that there's only one task available, it reverts to serial mode. This causes later output from `parallel_print()` to go to stdout. The "one task found" message goes to stderr, so stderr is both non-empty and does not contains the error reference string.

Since moving the 'one task found' message to stdout would still break the grepTest comparison in the one-task case, I think removing that message might be the best solution.

mattjala (changes requested):
The `parallel_print()` routine used by h5diff writes the "not found" message to stdout in serial, but to stderr in parallel. `grepTest.cmake` always accepts an empty error file, but if there is any error output, then it requires that it contain the error reference - in this case, the string `could not be found`. 

The serial h5diff tests normally work because all output goes to stdout, and stderr is empty and unchecked. The parallel h5diff tests normally work because stderr contained the error string 'could not be found'. The run in #5413 failed because, when h5diff detects that there's only one task available, it reverts to serial mode. This causes later output from `parallel_print()` to go to stdout. The "one task found" message goes to stderr, so stderr is both non-empty and does not contains the error reference string.

Since moving the 'one task found' message to stdout would still break the grepTest comparison in the one-task case, I think removing that message might be the best solution.

jhendersonHDF:
> The `parallel_print()` routine used by h5diff writes the "not found" message to stdout in serial, but to stderr in parallel. `grepTest.cmake` always accepts an empty error file, but if there is any error output, then it requires that it contain the error reference - in this case, the string `could not be found`.

To start with, this seems like something that should fixed. I'm split on whether the 'could not be found' message should go to `stdout` or `stderr` but at the very least it should be consistent between serial and parallel output.

> Since moving the 'one task found' message to stdout would still break the grepTest comparison in the one-task case, I think removing that message might be the best solution.

I'm of the opinion that removing output just to get a test to pass is not a particularly good way to go about things. Normal program output should go to `stdout`, errors and other informational type messages should go to `stderr` and `grepTest.cmake` should assume that and adjust as needed. Just because `stderr` is non-empty and doesn't contain the string being looked for doesn't mean that an error occurred or that the message being looked for isn't in `stdout`. It rather seems that `grepTest.cmake` should handle this better.

tools/test/h5diff/CMakeTests.cmake
byrnHDF: Actually, grepTest.cmake reuses arguments and in this case if it is not grepping the error file (TEST_REFERENCE) for text (TEST_ERRREF) then it simply greps the stdout for text (TEST_REFERENCE).

mattjala: So it does. That's very unintuitive behavior on the part of grepTest that I think we should change eventually.

I think the ideal fix for the original issue would be to make `parallel_print()` consistent about where it writes to in serial/parallel. If that has substantial knock-on effects, then this could be acceptable as a temporary solution.

jhendersonHDF: To do that, we should probably first determine why the two are inconsistent. In serial, `vprintf` is always used in `parallel_print()` so that one's obvious. In parallel, the `outBuff` buffer is used until the number of bytes to be printed exceeds `OUTBUFF_SIZE`, at which point a tempfile is used. We'd need to figure which one the output is going into and at what point the contents are dumped to `stderr` instead of `stdout`.

lrknox (approved):

jhendersonHDF:
Created related issue https://github.com/HDFGroup/hdf5/issues/5460 for dealing with the output differences.

jhendersonHDF (approved):

mattjala (approved):

PR #5422, opened 2025-04-07
Add file exists checks and fix policy 53 warnings

byrnHDF:
Fix CMake policy 0053, Simplify variable reference and escape sequence evaluation.
Fixes issue #5443


HDF5Examples/config/cmake/grepTest.cmake
```diff
@@ -23,7 +24,7 @@ if (NOT TEST_OUTPUT)
   message (FATAL_ERROR "Require TEST_OUTPUT to be defined")
 endif ()
 if (NOT TEST_FILTER)
-  message (STATUS "Optional TEST_FILTER to be defined")
+  message (VERBOSE "Optional TEST_FILTER to be defined")
```
lrknox, line 27: The meaning of this message seems unclear.  Following the if condition and with no action to define TEST_FILTER afterwards, would "Optional TEST_FILTER is not defined" be more accurate?

HDF5Examples/config/cmake/jrunTest.cmake
```diff
@@ -21,7 +21,7 @@ if (NOT TEST_PROGRAM)
   message (FATAL_ERROR "Require TEST_PROGRAM to be defined")
 endif ()
 if (NOT TEST_LIBRARY_DIRECTORY)
-  message (STATUS "Require TEST_LIBRARY_DIRECTORY to be defined")
+  message (VERBOSE "Require TEST_LIBRARY_DIRECTORY to be defined")
```
lrknox, line 24: Does "Require TEST_LIBRARY_DIRECTORY to be defined" mean defining it is required and it will be defined later, or it must be defined and it isn't?

HDF5Examples/config/cmake/grepTest.cmake
byrnHDF: Probably.

HDF5Examples/config/cmake/jrunTest.cmake
byrnHDF: Should be optional

HDF5Examples/config/cmake/jrunTest.cmake
```diff
@@ -102,10 +102,14 @@ endif ()
 if (TEST_MASK_ERROR)
   if (NOT TEST_ERRREF)
     # the error stack has been appended to the output file
-    file (READ ${TEST_FOLDER}/${TEST_OUTPUT} TEST_STREAM)
+    if (EXISTS "${TEST_FOLDER}/${TEST_OUTPUT}")
```
jhendersonHDF, line 105: Thinking about this more, would it be better to just exit/return an error/etc. when the test file doesn't exist? Otherwise, it seems like `TEST_STREAM` could be unset which might cause another problem further on.

byrnHDF: Not having an output file is not necessarily an error - just need to make sure there are no logic errors. And make sure of what is an error, i.e. no output file but reference file as an arg means there should be something.

byrnHDF: Also TEST_STREAM is a local variable for the block - need to be sure it is not used outside the block.

jhendersonHDF: In most of these files it looks like `TEST_STREAM` isn't particularly local since it's just set inside an if block and used outside those blocks later on. For example, it's printed out just a bit below this line which won't particularly do anything when it's empty, but also wouldn't be very helpful. It seems like a good argument for use of functions if we want to restrict scope of certain variables (which I'm all for).

byrnHDF: I think functions would work in these macro files - "cmake -P" - haven't investigated.

byrnHDF: > In most of these files it looks like `TEST_STREAM` isn't particularly local

Reviewing the files again.

lrknox (approved):

jhendersonHDF (approved):

PR #5423, opened 2025-04-07
Fix incorrect link size for VOL connectors in h5dump

mattjala:
Similar to the change in #5406, this is updating a spot where a tool conflated an API-level external link value and the internally (within the Native VOL) reported external link length. It now consistently uses the reported API-level values.

### DIFF
--- a/tools/src/h5dump/h5dump_ddl.c
+++ b/tools/src/h5dump/h5dump_ddl.c
@@ -168,8 +168,10 @@ dump_all_cb(hid_t group, const char *name, const H5L_info2_t *linfo, void H5_ATT
     h5tool_format_t  *outputformat = &h5tools_dataformat;
     h5tool_format_t   string_dataformat;
     hsize_t           curr_pos = 0; /* total data element position   */
+    H5L_info2_t       local_linfo;
 
     /* setup */
+    memset(&local_linfo, 0, sizeof(H5L_info2_t));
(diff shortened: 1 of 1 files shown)


tools/src/h5dump/h5dump_ddl.c
```diff
@@ -455,7 +457,13 @@ dump_all_cb(hid_t group, const char *name, const H5L_info2_t *linfo, void H5_ATT
                 break;
 
             case H5L_TYPE_EXTERNAL:
-                if ((targbuf = (char *)malloc(linfo->u.val_size)) == NULL) {
+                if (H5Lget_info2(group, name, &local_linfo, H5P_DEFAULT) < 0) {
```
jhendersonHDF, line 460: No particular problem with these changes; I'll just bring up again that I think this points to a fundamental design problem that must be fixed and these changes are hiding that. I believe it's reasonable that if you write some tool/code to iterate over links in a group, you should be able to assume that using the info in the `H5L_info2_t` that is passed to your callback will always be safe. By having to call `H5Lget_info2` here, you're essentially doing the same work twice and this may contribute to more performance problems in link iteration. I know that there are applications out there using this exact design pattern, so we will continue to hit this issue.

hyoklee (approved):

lrknox (approved):

PR #5429, opened 2025-04-09
Fix bad connector ID management during file open fallback

mattjala:
When #5015 reworked the VOL connector objects, one instance of the old connector ID management was missed in `H5VL__file_open_find_connector_cb()`. The ID is left invalid, and the H5I_dec_app_ref() call doesn't do anything.

The connector is now freed properly if the file open callback does not succeed in opening a file (i.e. if `ret_value != H5_ITER_STOP`).


qkoziol (approved):
Good catch!

bmribler (approved):

PR #5432, opened 2025-04-09
Refactor warning macros

### LINKED ISSUE #5424: Rework the warning suppression macros
The macros in H5warnings.h have several flaws:

- The compiler name(s) appear in the macros, which means that we sometimes either have multiple macro lines around problematic code or we wind up with "combo" macros like the gcc-clang macros that will be awkward to extend to new compilers.
- The macros are simple replacements for individual pragma lines, so any ifdef complications have to be handled in the library code, not H5warnings.h. These ifdefs and the comments that explain them will have to be duplicated throughout the library code.
- The name of the suppressed warning may not be very informative. For example, passing MPI_STATUSES_IGNORE as a parameter to MPI calls can raise spurious stringop-overflow warnings w/ gcc. It'd be nice to have a warning suppression that clearly indicated the nature of the (non-)problem.

We should replace the existing scheme with one that uses a wider set of macros that hide ifdef complexity and can be more easily extended to other compilers (e.g., MSVC).

Example:

```
/* Disable warnings concerning non-standard extensions, like F16 */
/* clang */
#if defined(__clang__)
#define H5_WARN_NONSTD_SUFFIX_OFF H5_WARN_OFF("pedantic")
#define H5_WARN_NONSTD_SUFFIX_ON  H5_WARN_ON("pedantic")
/* gcc 14+ */
#elif defined(__GNUC__) && __GNUC__ >= 14
#define H5_WARN_NONSTD_SUFFIX_OFF H5_WARN_OFF("c11-c23-compat")
#define H5_WARN_NONSTD_SUFFIX_ON  H5_WARN_ON("c11-c23-compat")
/* gcc 9-13 */
#elif defined(__GNUC__) && __GNUC__ >= 9
#define H5_WARN_NONSTD_SUFFIX_OFF H5_WARN_OFF("c11-c2x-compat")
#define H5_WARN_NONSTD_SUFFIX_ON  H5_WARN_ON("c11-c2x-compat")
#else
/* Everything else */
#define H5_WARN_NONSTD_SUFFIX_OFF
#define H5_WARN_NONSTD_SUFFIX_ON
#endif

```

derobins:
Fixes #5424


derobins:
This would be ready to review if Java worked on Ubuntu, which it does not. Without a working Java setup, I can't debug the Java JNI header issue, where it appears to have a problem seeing H5warnings.h via H5private.h.

I'd also like to extend this to MSVC warnings, but I can't get HDF5 to compile with Visual Studio due to complex number errors. It's unclear why HDF5 builds in CI but not locally with Visual Studio.

jhendersonHDF:
> This would be ready to review if Java worked on Ubuntu, which it does not. Without a working Java setup, I can't debug the Java JNI header issue, where it appears to have a problem seeing H5warnings.h via H5private.h.

@byrnHDF or I might be able to take a look at java. What issue are you having with Java on Ubuntu?

> I'd also like to extend this to MSVC warnings, but I can't get HDF5 to compile with Visual Studio due to complex number errors. It's unclear why HDF5 builds in CI but not locally with Visual Studio.

Could you create an issue with the problems you're seeing? I might be able to take a look if I can get Visual Studio working somewhere.

tools/src/h5repack/h5repack_filters.c
```diff
@@ -219,7 +217,6 @@ aux_assign_obj(const char  *name,        /* object name from traverse list */
         }
     }
 
-    *obj = tmp;
```
derobins, line 222: Not a good idea to return a stack object through a pointer...

derobins:
This is ready for review now.

FWIW, if we can fix a few stragglers in C++, Fortran, clang, and MSVC, we can start building with -Werror always and get rid of the special jobs.

testpar/t_subfiling_vfd.c
qkoziol, line 1: Using snprintf eliminates the warning?

tools/src/h5repack/h5repack_filters.c
jhendersonHDF: This isn't returning a stack object, this is using struct assignment from `tmp` to the struct pointed to by `obj`, so it's perfectly fine. Plus it's a lot cleaner than the new solution if errors were to occur because all the fields were previously assigned all at once at the end here only on success, whereas the new solution could half-initialize the pointed to structure.

src/H5warnings.h
```diff
@@ -12,62 +12,315 @@
...
+#endif
+
+/* Suppress warnings about bad cast alignment. These should be corrected,
+ * not suppressed in the main library, but might appear in test code.
  */
-#ifndef H5_SHOW_ALL_WARNINGS
-#define H5_DIAG_OFF(x) H5_DIAG_PRAGMA(push) H5_DIAG_PRAGMA(ignored H5_DIAG_JOINSTR("-W", x))
-#define H5_DIAG_ON(x)  H5_DIAG_PRAGMA(pop)
+#if defined(__clang__) || defined(__GNUC__)
+#define H5_WARN_CAST_ALIGNMENT_OFF H5_WARN_OFF("cast-align")
```
jhendersonHDF, line 126: Out of all the warning types, `cast-align` is one that I think we should never allow suppressing warnings for. While it's fine on some platforms it's certainly not on others and will likely lead to subtle issues.

tools/src/h5repack/h5repack_filters.c
qkoziol: Yeah, I tend to agree

qkoziol (approved):
Looks good to me, modulo the possible revert on the pack option (but I don't have a strong opinion either way on that)

testpar/t_subfiling_vfd.c
jhendersonHDF: Seconding this; is the warning just no longer issued?

c++/test/tcompound.cpp
```diff
@@ -98,7 +98,9 @@ test_compound_2()
         bkg  = static_cast<unsigned char *>(malloc(nelmts * sizeof(dst_typ_t)));
         orig = static_cast<unsigned char *>(malloc(nelmts * sizeof(src_typ_t)));
         for (i = 0; i < nelmts; i++) {
-            s_ptr       = (reinterpret_cast<src_typ_t *>(orig)) + i;
+            H5_WARN_CAST_ALIGNMENT_OFF
+            s_ptr = (reinterpret_cast<src_typ_t *>(orig)) + i;
```
jhendersonHDF, line 102: If these few places are the only locations where `cast-align` is currently suppressed, I'd much rather these be fixed (maybe in a separate PR) than continuing to be suppressed. The solution is very simple so no real need to keep suppressing `cast-align`.

jhendersonHDF:
Great changes that I'd like to see merged to fix the new annoying warnings from C23-related issues; just a couple minor things.

testpar/t_subfiling_vfd.c
derobins: No longer complains (if it did, the -Werror checks would fail)

jhendersonHDF: I'm not sure that we're currently building parallel with -Werror or at least that it works correctly. Some of the parallel builds use the HDF5_ENABLE_WARNINGS_AS_ERRORS option, but I can see this warning in https://github.com/HDFGroup/hdf5/actions/runs/14818490824/job/41602216637, for example.

jhendersonHDF: I think the option may simply not be covering all the cases we're interested in turning into an error from a warning. Looks like this particular warning may need to be suppressed again. We'll check into the issues with the option outside of this PR.

jhendersonHDF (approved):
I'll look into the remaining issue with the new warning later on; I'd much rather have the warning suppression reworking in.

PR #5441, opened 2025-04-11
Include 400/500 HTTP status codes in ROS3 VFD error stacks

### LINKED ISSUE #5434: Update ROS3 VFD to include HTTP status code in error stack on failure
Unless `S3COMMS_CURL_VERBOSITY` in `H5FDs3comms.c` is defined to a value > 0 (the default is 0), the ROS3 VFD does not currently show the HTTP status code from a request failure, which can be useful in identifying why problems occurred. Further, when `S3COMMS_CURL_VERBOSITY > 0` the code is printed to `stdout`, which could be a bit disruptive.

The ROS3 VFD should be updated to include the HTTP status code in HDF5's error stack when failures occur.

jhendersonHDF:
Fixes #5434.

### DIFF
--- a/src/H5FDs3comms.c
+++ b/src/H5FDs3comms.c
@@ -66,6 +66,19 @@
 /* Size of buffers that hold hex representations of SHA256 digests */
 #define S3COMMS_SHA256_HEXSTR_LENGTH (SHA256_DIGEST_LENGTH * 2 + 1)
 
+/* Defines for the use of HTTP status codes */
+#define HTTP_CLIENT_ERROR_MIN 400 /* Minimum and maximum values for the 400 class of */
+#define HTTP_CLIENT_ERROR_MAX 499 /* HTTP client error responses */
+
+#define HTTP_SERVER_ERROR_MIN 500 /* Minimum and maximum values for the 500 class of */
+#define HTTP_SERVER_ERROR_MAX 599 /* HTTP server error responses */
+
+/* Macros to check for classes of HTTP response */
+#define HTTP_CLIENT_ERROR(status_code)                                                                       \
+    (status_code >= HTTP_CLIENT_ERROR_MIN && status_code <= HTTP_CLIENT_ERROR_MAX)
+#define HTTP_SERVER_ERROR(status_code)                                                                       \
+    (status_code >= HTTP_SERVER_ERROR_MIN && status_code <= HTTP_SERVER_ERROR_MAX)
+
 /******************/
 /* Local Typedefs */
 /******************/
@@ -80,7 +93,8 @@
  */
 struct s3r_datastruct {
     char  *data;
-    size_t size;
+    size_t cur_size;
+    size_t max_size;
 };
 
 /********************/
@@ -105,6 +119,8 @@ static herr_t H5FD__s3comms_make_iso_8661_string(time_t time, char iso8601[ISO86
 static parsed_url_t *H5FD__s3comms_parse_url(const char *url);
 
 static herr_t H5FD__s3comms_free_purl(parsed_url_t *purl);
+
+static const char *H5FD__s3comms_httpcode_to_str(long httpcode);
 
 /*********************/
 /* Package Variables */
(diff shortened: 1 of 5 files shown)


src/H5FDs3comms.c
```diff
@@ -776,7 +802,7 @@ H5FD__s3comms_s3r_open(const char *url, const H5FD_ros3_fapl_t *fa, const char *
         HGOTO_ERROR(H5E_VFL, H5E_BADVALUE, NULL, "error while setting CURL option (CURLOPT_HTTPGET)");
     if (CURLE_OK != curl_easy_setopt(curlh, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1))
         HGOTO_ERROR(H5E_VFL, H5E_BADVALUE, NULL, "error while setting CURL option (CURLOPT_HTTP_VERSION)");
-    if (CURLE_OK != curl_easy_setopt(curlh, CURLOPT_FAILONERROR, 1L))
+    if (CURLE_OK != curl_easy_setopt(curlh, CURLOPT_FAILONERROR, 0L)) /* Defaults off anyway, but make sure */
```
jhendersonHDF, line 805: This option makes error handling slightly simpler, but also throws away the HTTP response body that might contain more details on what a client should do for a particular error. To handle these situations in the future, it's better to make sure this option is turned off and then directly handle HTTP status codes ourself. This also means that some buffer overflow checking logic had to be added to this file since the ROS3 VFD wasn't initially designed with this use case in mind.

src/H5FDs3comms.c
```diff
@@ -1887,4 +1925,63 @@ H5FD__s3comms_make_aws_stringtosign(char *dest, const char *req, const char *now
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5FD__s3comms_make_aws_stringtosign() */
 
+/*
+ * Maps HTTP status codes (currently, just 400 and 500 codes)
+ * to generic strings
+ */
+static const char *
+H5FD__s3comms_httpcode_to_str(long httpcode)
```
jhendersonHDF, line 1933: The particular status codes mapped to generic strings are mostly chosen arbitrarily here and would be better handled by a third party library in the future. This is just an intermediate solution.

src/H5FDs3comms.c
```diff
@@ -921,21 +947,23 @@ H5FD__s3comms_s3r_configure_aws(s3r_t *handle, const H5FD_ros3_fapl_t *fa, const
  *----------------------------------------------------------------------------
  */
 herr_t
-H5FD__s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest)
+H5FD__s3comms_s3r_read(s3r_t *handle, haddr_t offset, size_t len, void *dest, size_t dest_size)
```
jhendersonHDF, line 950: With `CURLOPT_FAILONERROR` turned off, the data that comes back from a request on failure might be an additional response body rather than the bytes being looked for. In this case, it's possible that the previous logic could have tried to put more bytes into the destination buffer than what's available, so a buffer size parameter was added here. The ROS3 VFD needs to be redesigned a bit to correctly handle these situations, but this is still a useful check anyway.

src/H5FDs3comms.c
```diff
@@ -136,12 +152,21 @@ H5FD__s3comms_curl_write_callback(char *ptr, size_t size, size_t nmemb, void *us
     struct s3r_datastruct *sds    = (struct s3r_datastruct *)userdata;
     size_t                 nbytes = size * nmemb;
 
-    /* Write bytes and size to userdata/sds struct */
-    if (size > 0) {
```
jhendersonHDF, line 140: `size` will always be 1 here, so changed this to check the total data size instead.

jhendersonHDF:
Example:

Before:
```bash
h5dump -f ros3 -pH --enable-error-stack=2 <URL>/<Bucket>/bogus.h5
...
#010: H5FDs3comms.c line 1213 in H5FD__s3comms_s3r_read(): curl cannot perform request
    major: Virtual File Layer
    minor: Unable to open file
```

After:
```bash
h5dump -f ros3 -pH --enable-error-stack=2 <URL>/<Bucket>/bogus.h5
...
#010: H5FDs3comms.c line 1246 in H5FD__s3comms_s3r_read(): 404 - Resource not found
    major: Virtual File Layer
    minor: Read failed
```

mattjala (approved):

lrknox (approved):

PR #5454, opened 2025-04-18
Add subfolder argument to gentest scripts

mattjala:
Part of a project to re-integrate the gentest scripts with VOL tools testing. This makes it simple to have the gentest scripts match the current expected behavior where test files end up in the `testfiles` directory.

`h5perf_gentest.c` is excluded from these changes since it already takes in filenames which can have a subfolder prepended to them.

### DIFF
--- a/tools/test/h5copy/CMakeLists.txt
+++ b/tools/test/h5copy/CMakeLists.txt
@@ -23,7 +23,7 @@ if (HDF5_BUILD_GENERATORS)
     clang_format (HDF5_TOOLS_TEST_H5COPY_FORMAT h5copygentest)
   endif ()
 
-  #add_test (NAME h5copygentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5copygentest>)
+  #add_test (NAME h5copygentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5copygentest> "testfiles")
 endif ()
 
 #-----------------------------------------------------------------------------
--- a/tools/test/h5diff/CMakeLists.txt
+++ b/tools/test/h5diff/CMakeLists.txt
@@ -24,7 +24,11 @@ if (HDF5_BUILD_GENERATORS)
     clang_format (HDF5_TOOLS_TEST_H5DIFF_FORMAT h5diffgentest)
   endif ()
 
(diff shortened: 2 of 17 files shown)


tools/test/h5diff/CMakeLists.txt
```diff
@@ -24,7 +24,11 @@ if (HDF5_BUILD_GENERATORS)
     clang_format (HDF5_TOOLS_TEST_H5DIFF_FORMAT h5diffgentest)
   endif ()
 
-  #add_test (NAME h5diffgentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diffgentest>)
+  #add_test (NAME h5diffgentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diffgentest> "testfiles")
+
+  # if (H5_HAVE_PARALLEL)
+  #   add_test (NAME h5diffgentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diffgentest> "PAR/testfiles")
```
byrnHDF, line 30: Doesn't this generate the same files? just copy the files?

tools/test/h5copy/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if (HDF5_BUILD_GENERATORS)
     clang_format (HDF5_TOOLS_TEST_H5COPY_FORMAT h5copygentest)
   endif ()
 
-  #add_test (NAME h5copygentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5copygentest>)
+  #add_test (NAME h5copygentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5copygentest> "testfiles")
```
byrnHDF, line 26: Another way would just to set the WORKING_DIRECTORY to testfiles. No need to update the code.

tools/test/h5diff/CMakeLists.txt
mattjala: For the native connector, yes. But VOL connectors aren't guaranteed to use a single file per HDF5 file, or to use files at all. This for compatibility with connector that do something different on the backend

mattjala:
Closing this, since the same change can be accomplished with the `WORKING_DIRECTORY` option at the same time that the gentest tests are enabled

PR #5456, opened 2025-04-18
Bring h5diffgentest in line with expected test files

mattjala:
- Add files to gentest script: `trefer_attr.h5`, `tvlstr.h5`, `tvlstr2.h5`

The routines to build these are copied from h5dumpgentest. It might make sense to eventually have a separate collection of shared build file generator routines that can be shared between gentest scripts.

- Fix cases where the expected datataype does not match the generated one

- Fix some object names not matching their expected values 

- Update test files and expected files for `test_comps_array_vlen` and  `test_comps_vlen_arry`

- Fixed incorrect datasets being compared for compound dataset test

- Update the (disabled for now) h5diffgentest CMake test to create files in the correct directories

We can't just copy the testfiles to the parallel directory because the active VOL connector may not use files.


byrnHDF:
One thing that is missing and needs to be handled - the tools will need reference files generated if any of the testfiles change.

mattjala:
> One thing that is missing and needs to be handled - the tools will need reference files generated if any of the testfiles change.

None of the testfiles are changed here. The gentest script is out of date and currently creates testfiles that are different than the one we have committed. This is changing gentest to create testfiles identical to the ones we already have.

byrnHDF (dismissed):

tools/test/h5diff/CMakeLists.txt
```diff
@@ -24,7 +24,11 @@ if (HDF5_BUILD_GENERATORS)
     clang_format (HDF5_TOOLS_TEST_H5DIFF_FORMAT h5diffgentest)
   endif ()
 
-  #add_test (NAME h5diffgentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diffgentest>)
+  #add_test (NAME h5diffgentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diffgentest> WORKING_DIRECTORY "testfiles")
+
+  # if (H5_HAVE_PARALLEL)
+  #   add_test (NAME h5diffgentest COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diffgentest> WORKING_DIRECTORY "PAR/testfiles")
```
jhendersonHDF: I know this is still disabled, but what's the reason for adding another command here for parallel? I don't generally expect there to be any differences for parallel in the near future, if ever. Plus, this has the same name as the serial test above, so it would need to be changed.

mattjala: This is to match the current behavior, where h5diff testfiles are copied to the PAR directory when the library is built with parallel (`h5diff/CMakeTests.cmake:330-370`). 

I'll fix the test name conflict.

tools/test/h5diff/h5diffgentest.c
```diff
@@ -1461,8 +1471,8 @@ test_datatypes(const char *fname)
     short   buf2b[3][2] = {{1, 1}, {3, 4}, {5, 6}};
     int     buf3a[3][2] = {{1, 1}, {1, 1}, {1, 1}};
     int     buf3b[3][2] = {{1, 1}, {3, 4}, {5, 6}};
-    long    buf4a[3][2] = {{1, 1}, {1, 1}, {1, 1}};
```
jhendersonHDF: The point here seems to have been to test the various C datatypes (`char` -> `short` -> `int` -> `long` -> `float` -> `double`), in which case I would leave this as a `long`. For most of the changes made in this file, in the spirit of "more testing" I would think it's better to just update the testing files to match what's currently being generated by h5diffgentest rather than trying to "revert" h5diffgentest to match the testing files that are currently checked in. So, I'd also not remove the attributes in the other test files and instead check in newly generated files. Most likely what happened is someone updated h5diffgentest in the past and forgot to check in newly generated test files. I'd keep the new `gent_trefer_attr` and `gent_vlstr` changes though.

tools/test/h5diff/CMakeLists.txt
jhendersonHDF: Based on reading of the CMake docs, what's specified for `WORKING_DIRECTORY` may also need modifying as well. If not specified, CMake uses `[CMAKE_CURRENT_BINARY_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_BINARY_DIR.html#variable:CMAKE_CURRENT_BINARY_DIR)`, which to me implies that if you do specify it, you'll need to use a full path. Or rather it's not very clear what the starting point is if you do specify it. Fine for now, I just generally prefer not to add code until it's actually needed since it tends to stick around commented out for a long time.

tools/test/h5diff/h5diffgentest.c
byrnHDF: Yes, I'd agree that someone failed to check in new files.

tools/test/h5diff/CMakeLists.txt
byrnHDF: Yes working dir should be full path - see the root CMakeLists.txt file for predefined path variables. Also need to understand the CMake default *_BINARY_DIR/*_SOURCE_DIR temporary variables for CURRENT vs <project> types

tools/test/h5diff/h5diffgentest.c
mattjala: I've checked in new files for those two tests.

For the datatypes test, it looks like there was a typo in the CMake causing the wrong datasets to be compared.

For the attributes tests, the checked-in files were missing several attributes.

tools/test/h5diff/h5diffgentest.c
```diff
@@ -8929,3 +8940,135 @@ write_dset(hid_t loc_id, int rank, hsize_t *dims, const char *name, hid_t tid, v
 
     return FAIL;
 } /* end write_dset() */
+
+/* TBD: Duplicated from h5dumpgentest */
+static void
```
jhendersonHDF: We probably need to think of a better solution for files which are generated by one generator program and used by multiple tools. CMake can't necessarily copy those files as needed, but duplicating the code to generate the files means you have to know and remember to update multiple places when you change one of the testing files.

jhendersonHDF: I believe the VOL tools tests originally used h5repack to read the files in with the native VOL and repack them to the VOL connector set in `HDF5_VOL_CONNECTOR`.

lrknox (approved):

byrnHDF (approved):

PR #5458, opened 2025-04-21
Bring h5dumpgentest in line with expected test files

mattjala:
- Add routines to generate several test files that were missing from the gentest script

- Fix several datatypes having unexpected endianness on certain systems

- Fix several issues on systems with small long double types

- Fix created file for long links test not matching expected file

- Use double values during complex data setup to fix inaccuracies with the intel compiler at -O3

- Add several missing testfiles to h5dumpgentest

### DIFF
--- a/tools/test/h5dump/h5dumpgentest.c
+++ b/tools/test/h5dump/h5dumpgentest.c
@@ -5099,7 +5099,7 @@ gent_compound_complex(void)
 {
     /* Structure and array for compound types                             */
     typedef struct Array1Struct {
-        int32_t         a;
+        int32_t     a;
         const char *b[F41_DIMb];
         char        c[F41_ARRAY_DIMc];
         short       d[F41_ARRAY_DIMd1][F41_ARRAY_DIMd2];
@@ -10529,7 +10529,7 @@ gent_floatsattrs(void)
     aset128 = calloc(F89_XDIM * F89_YDIM128, sizeof(long double));
 
     fid = H5Fcreate(FILE89, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
-   
+
     if ((tid = H5Tcopy(H5T_NATIVE_FLOAT)) < 0)
         goto error;
 
--- a/tools/test/h5dump/h5dumpgentest.c
+++ b/tools/test/h5dump/h5dumpgentest.c
@@ -118,6 +118,17 @@
 #define FILE95 "tcomplex.h5"
 #define FILE96 "tcomplex_be.h5"
 #endif
+#define FILE97   "trefer_attr.h5"
+#define FILE98   "tattr4_be.h5"
+#define FILE99   "tno-subset.h5"
+#define F99_DSET "AHFINDERDIRECT::ah_centroid_t[0] it=0 tl=0"
+#define FILE100  "trefer_compat.h5"
+#define FILE101  "trefer_grp.h5"
+#define FILE102  "trefer_obj_del.h5"
+#define FILE103  "trefer_obj.h5"
+#define FILE104  "trefer_param.h5"
+#define FILE105  "trefer_reg.h5"
+#define FILE106  "trefer_reg_1d.h5"
 
 #define ONION_TEST_FIXNAME_SIZE 1024
 #define ONION_TEST_PAGE_SIZE    (uint32_t)32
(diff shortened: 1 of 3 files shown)


tools/test/h5dump/h5dumpgentest.c
```diff
@@ -24,109 +24,125 @@
 #include "h5test.h"
 #include "h5tools.h"
 
-#define FILE1      "tgroup.h5"
-#define FILE2      "tdset.h5"
-#define FILE3      "tattr.h5"
-#define FILE4      "tslink.h5"
+#define TESTFILE_DIR "testfiles/"
```
jhendersonHDF: "testfiles/" should not be prepended to these test file names. There shouldn't be any assumptions about where the test file generator programs are being run from; a "testfiles" directory may not exist (nor should we try to create it from wherever the generator programs are being run from).

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -1177,6 +1193,10 @@ gent_compound_dt(void)
     H5Gclose(group);
 
     /* unnamed data type */
+    /* TBD: h5dump generates a placeholder name for the datataype based on the VOL connector's
```
jhendersonHDF: In general, the generator programs should not try to concern themselves with adapting to VOL connectors. Files should be generated "as is" and it should be up to the build system or something external to determine whether it makes sense to try testing against a particular file.

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -11521,12 +11629,18 @@ gent_onion_1d_dset(void)
 
     /* Setup */
     onion_info.backing_fapl_id = H5Pcreate(H5P_FILE_ACCESS);
+    /* Force onion files to be generated via Native connector */
+    if (H5Pset_vol(onion_info.backing_fapl_id, H5VL_NATIVE, NULL) < 0)
```
jhendersonHDF: Forcing the native VOL is not the way to go about this, especially since this could be performed on various passthrough VOL connectors which should generally work fine. Skipping generation of certain test files for some connectors could potentially be acceptable, but I'd much rather that skipping be done at the build system level or higher rather than here. I'm thinking more in the direction of "all test files should always be generated, regardless of VOL connector". It's ok if some files can't be generated due to something a connector doesn't support, like going through the VFL interface in this case, but we shouldn't try to force these situations.

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -12764,7 +13466,9 @@ main(void)
     gent_hardlink();
     gent_extlink();
     gent_udlink();
-    gent_compound_dt();
+    /* TBD: Disabled for now , since the generated file will be assigned inconsistent datatype names by h5dump
+     */
+    // gent_compound_dt();
```
jhendersonHDF: Completely disabling these tests is definitely not the way to go. Skipping these for some VOL connectors would be a slightly better way to go, but again I'd still much prefer not to do that directly in this code. The generator programs should be "dumb".

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -11726,10 +11847,21 @@ gent_onion_create_delete_objects(void)
         for (int j = 0; j < 4; j++)
             wdata[i][j] = i + j;
 
+    /* Ensure that the onion file does not exist, to avoid tracking revisions across multiple test runs */
```
jhendersonHDF: To me this seems like it may be an issue with the Onion VFD that should be solved there rather than here (again, in the spirit of the generator programs being "dumb"). What does it mean to truncate an onion file? I would think that removing/clearing the revisions file would make sense, but that's something that would need to be thought about. But having a generator program directly remove files immediately sticks out to me as something not to do.

tools/test/h5dump/h5dumpgentest.c
mattjala: I agree, but in this case this is a problem even for the Native connector. The generated `tcompound.h5` matches the checked-in version in every externally visible way, but h5dump assigns a different datatype number to the generated version.

We could replace the checked-in file with this new version and update the expected output, but I was wary to do so given that there's no apparent reason for the change in behavior, so later library updates might force us to repeat the process again.

jhendersonHDF: The name comes from the object token, which in turn comes from the object's address in the file. So yes, it could change for any number of reasons. This is why the original approach for VOL tools testing masked out certain things in the output. See https://github.com/HDFGroup/vol-tests/blob/master/tools/testh5dump.sh#L516-L517 and https://github.com/HDFGroup/vol-tests/blob/master/tools/output_filter.sh. The idea behind VOL tools testing is that we should be taking that same approach of generating all the files with a VOL connector and then masking out what information can't reasonably be checked. The main difficulty is translating all that new logic back into the library's existing testing infrastructure since the external repository mostly just did this through ad-hoc scripts.

mattjala: That seems like a workable solution. I've re-enabled the tests, the masking will be included in a later PR.

byrnHDF (approved):

lrknox (approved):

PR #5465, opened 2025-04-22
h5cc shell portability fixes

mgorny:
Two related changes to the CMake `h5cc.in` file:

1. Replace the `${@:2}` construct that is specific to bash shell with a more portable approach based on `shift`, in order to restore compatibility with more strict POSIX shells.  This fixes a regression introduced in #5361.
2. Pass arguments to subprocesses via quoted `"$@"` rather than plain `$@`, in order to prevent the shell from applying word splitting, filename expansion, etc., and therefore ensure that they are passed through to the compiler process unchanged.

### DIFF
--- a/config/cmake/libh5cc.in
+++ b/config/cmake/libh5cc.in
@@ -66,7 +66,8 @@ for arg in $@ ; do
       exit $status
       ;;
     -show)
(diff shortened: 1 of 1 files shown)


byrnHDF (approved):

lrknox (approved):

mgorny:
Thank you!

PR #5467, opened 2025-04-22
h5cc: Allow overriding the compilers written into the file

### LINKED ISSUE #5481: Please add an option to override the compilers stored in `h5cc`
Currently, the HDF5 build system stores the exact compiler paths used to build HDF5 into `h5cc` and related executables. This can be problematic when using tools such as ccache or distcc, as these tools (used on the package build system) end up being written into the file and used afterwards, even though they may not be relevant to the user use of `h5cc`, for example:

    /usr/lib/ccache/bin/x86_64-pc-linux-gnu-gcc

I think the cleanest approach would be to add a CMake variables to override the compilers stored in `h5cc`, etc. The variables could default to the currently used compilers to preserve backwards compatibility, and therefore need to be used only by advanced users.

The alternative would be to force using the exact system compiler without ccache or distcc, as [openSUSE is already doing](https://build.opensuse.org/projects/science/packages/hdf5/files/hdf5.spec?expand=1). However, this is suboptimal and nontrivial to implement in Gentoo.

**Additional context**
PR #5467

mgorny:
Add a set of `HDF5_H5CC_C_COMPILER`, `HDF5_H5CC_CXX_COMPILER` and `HDF5_H5CC_Fortran_COMPILER` variables that can be used to override the compiler string written into `h5cc`, `h5c++`, etc.  This is particularly useful when e.g. using ccache during the build, as the ccache path otherwise ends up in `h5cc`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1058,7 +1058,7 @@ else ()
 endif ()
 
 #-----------------------------------------------------------------------------
-# Option to override the compiler for pkg-config & h5cc
+# Option to override the compiler for h5cc
 #-----------------------------------------------------------------------------
 if (HDF5_ENABLE_PARALLEL AND MPI_C_FOUND)
   set (_HDF5_H5CC_C_COMPILER ${MPI_C_COMPILER})
@@ -1230,13 +1230,6 @@ if (EXISTS "${HDF5_SOURCE_DIR}/fortran" AND IS_DIRECTORY "${HDF5_SOURCE_DIR}/for
       endif ()
     endif ()
 
-    if (HDF5_ENABLE_PARALLEL AND MPI_Fortran_FOUND)
-      set (_HDF5_H5CC_Fortran_COMPILER ${MPI_Fortran_COMPILER})
-    else ()
-      set (_HDF5_H5CC_Fortran_COMPILER ${CMAKE_Fortran_COMPILER})
-    endif ()
-    set (HDF5_H5CC_Fortran_COMPILER ${_HDF5_H5CC_Fortran_COMPILER} CACHE STRING "Fortran compiler to use in h5cc")
-
     add_subdirectory (fortran)
     if (HDF5_BUILD_HL_LIB)
       if (EXISTS "${HDF5_SOURCE_DIR}/hl/fortran" AND IS_DIRECTORY "${HDF5_SOURCE_DIR}/hl/fortran")
@@ -1263,13 +1256,6 @@ if (EXISTS "${HDF5_SOURCE_DIR}/c++" AND IS_DIRECTORY "${HDF5_SOURCE_DIR}/c++")
     endif ()
 
     include (${HDF_RESOURCES_DIR}/HDFCompilerCXXFlags.cmake)
-
-    if (HDF5_ENABLE_PARALLEL AND MPI_CXX_FOUND)
-      set (_HDF5_H5CC_CXX_COMPILER ${MPI_CXX_COMPILER})
-    else ()
-      set (_HDF5_H5CC_CXX_COMPILER ${CMAKE_CXX_COMPILER})
-    endif ()
-    set (HDF5_H5CC_CXX_COMPILER ${_HDF5_H5CC_CXX_COMPILER} CACHE STRING "C++ compiler to use in h5cc")
 
     add_subdirectory (c++)
     if (HDF5_BUILD_HL_LIB)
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -1057,6 +1057,16 @@ else ()
...
 
+#-----------------------------------------------------------------------------
+# Option to override the compiler for pkg-config & h5cc
+#-----------------------------------------------------------------------------
+if (HDF5_ENABLE_PARALLEL AND MPI_C_FOUND)
+  set (_PKG_CONFIG_C_COMPILER ${MPI_C_COMPILER})
+else ()
+  set (_PKG_CONFIG_C_COMPILER ${CMAKE_C_COMPILER})
+endif ()
+set (PKG_CONFIG_C_COMPILER ${_PKG_CONFIG_C_COMPILER} CACHE STRING "C compiler to use in h5cc")
```
jhendersonHDF: Bikeshedding a bit here, but as these are HDF5-specific options I'd generally prefer that they follow that naming. Maybe something like `HDF5_H5CC_C_COMPILER`, etc. `h5cc` using pkg-config is really just an implementation detail.

mgorny: Sure. I've just followed the original names used in the individual files but I definitely agree that better names are better :-).

mgorny: Done.

jhendersonHDF: Thanks! Also makes it a little clearer for me to read now compared to before:

`set (_PKG_CONFIG_C_COMPILER ...)` -> `set (PKG_CONFIG_C_COMPILER ${_PKG_CONFIG_C_COMPILER} ...)` -> `set (_PKG_CONFIG_COMPILER ${PKG_CONFIG_C_COMPILER})`

jhendersonHDF:
@mgorny It's a bit of a hassle, but could you create an issue just quickly detailing the issue you ran into / the motivation for this PR and link it to this PR? It's pretty obvious (at least to me) the "why", but we tend to use issues for tracking various things rather than looking for descriptions in PRs.

mgorny:
No problem. Issue #5481

jhendersonHDF (dismissed):
Thanks! LGTM. @byrnHDF Any issues you see with this?

CMakeLists.txt
```diff
@@ -1247,6 +1264,13 @@ if (EXISTS "${HDF5_SOURCE_DIR}/c++" AND IS_DIRECTORY "${HDF5_SOURCE_DIR}/c++")
 
     include (${HDF_RESOURCES_DIR}/HDFCompilerCXXFlags.cmake)
 
+    if (HDF5_ENABLE_PARALLEL AND MPI_CXX_FOUND)
```
byrnHDF: This probably should go into ${HDF_RESOURCES_DIR}/HDFCompilerCXXFlags.cmake file which protects from the configuration from force-enabling the CXX language if not requested.

Same for Fortran.

byrnHDF (changes requested):
Move the set commands for Fortran and CXX to appropriate files.  Otherwise, those languages will be enabled even if they were not selected to be used.

Needs to be verified, these commands may not cause the issue.

CMakeLists.txt
mgorny: I'm not sure if I understand that. The code is inside `HDF5_BUILD_CPP_LIB` condition, so I don't see how it would be enabled accidentally.

That said, I can move it. Do I understand correctly that I should move it inside the `CMAKE_CXX_COMPILER_LOADED` block?

byrnHDF:
So the basis for the comment is that in the past there was a problem in configuration that C++ was enabled and there was no C++ compiler, so config failed. Turned out to be a command that triggered it. The command seemed to be general purpose. So since then we try to protect the optional languages from being accidentally being enabled. One way is with the  CMAKE_CXX_COMPILER_LOADED check. Plus for Fortran too.
Also the reason for the language-specific files.

mgorny:
Done.

byrnHDF (changes requested):
Only one last item, a short description of the change in the release_docs/REALEASE.txt file. I think under New Features/Configuration right at the begining.

mgorny:
Added.

byrnHDF (approved):

jhendersonHDF (approved):

brtnfld (approved):

mgorny:
Thank you!

Issue #5460, opened 2025-04-21
Align output between serial and parallel h5diff


jhendersonHDF:
See related discussion in https://github.com/HDFGroup/hdf5/pull/5416. Output from (p)h5diff may end up in either stdout or stderr, depending on whether serial diff is used, parallel diff is used or parallel diff is used but with only 1 MPI process. Differences between the 3 cases should be aligned so that checking of testing output works as expected.

PR #5469, opened 2025-04-22
Use stream variables in tools

### LINKED ISSUE #5466: tools should use stream vars
All tools should use the initialized stream variables for stder/stdout - using stderr/stdout could result in incompatibilities with tests/results.

byrnHDF:
Fixes #5466


tools/lib/h5diff_util.c
```diff
@@ -333,7 +333,7 @@ print_found(hsize_t nfound)
     if (g_Parallel)
         parallel_print("%" PRIuHSIZE " differences found\n", nfound);
     else
-        fprintf(stdout, "%" PRIuHSIZE " differences found\n", nfound);
+        fprintf(rawoutstream, "%" PRIuHSIZE " differences found\n", nfound);
```
jhendersonHDF, line 336: May also want to update `parallel_print` to use `vfprintf` to `rawoutstream` rather than `vprintf` in the serial case at least. Parallel is a little more complicated due to the multiple buffers, but `print_manager_output()` in ph5diff_main.c is generally where most of the real output comes from. It uses regular `printf` and `putchar`, which could probably be updated to `fprintf` and `fputc` as well.

tools/src/h5jam/h5unjam.c
```diff
@@ -114,7 +115,7 @@ parse_command_line(int argc, const char *const *argv)
                 if (ub_file)
                     h5tools_set_output_file(ub_file, 1);
                 else
-                    rawoutstream = stdout;
+                    rawoutstream = rawoutstream;
```
jhendersonHDF: Looks like an accidental find-replace

tools/src/h5stat/h5stat.c
```diff
@@ -1086,7 +1086,7 @@ print_file_info(const iter_t *iter)
     printf("\t# of unique links: %lu\n", iter->uniq_links);
     printf("\t# of unique other: %lu\n", iter->uniq_others);
     printf("\tMax. # of links to object: %lu\n", iter->max_links);
-    fprintf(stdout, "\tMax. # of objects in group: %" PRIuHSIZE "\n", iter->max_fanout);
```
jhendersonHDF, line 1089: Seems like all the other prints above should be consistent here so that you don't have split output when `rawoutstream` is something other than `stdout` (like a file). I imagine there are probably many other printf, putc, etc. calls like this.

tools/src/h5jam/h5unjam.c
byrnHDF: so careful yet so sloppy!

tools/lib/h5diff_util.c
byrnHDF: A little worried about that one.

jhendersonHDF: Yes, ph5diff is a little unique and that should probably be covered related to #5460.

byrnHDF: parallel_print goes to overflow_file then what prints that? So i'm not sure where to edit

jhendersonHDF: It goes to `outBuff` first and then over a certain amount of output will go to `overflow_file`. Both are printed in `print_manager_output()` but could potentially get printed elsewhere as well.

tools/src/h5stat/h5stat.c
byrnHDF: That is one of the tools that is out of step and needs to fixed. I'll see if I can convert those printfs.

lrknox (approved):

qkoziol (approved):

PR #5471, opened 2025-04-22
Rename Cmake test script parameter

mattjala:
Renamed the `TEST_FILTER` grepTest.cmake parameter to `TEST_EXPECTED_STRING`. 

grepTest and runTest both have a TEST_FILTER parameter, but they're used for different things.

- runTest's TEST_FILTER is regex expression used to perform replacements in the output file before correctness checking

- grepTest's TEST_FILTER is a regex expression used to search for a particular expected string in the output file, and fail if it is not found

This should prevent confusion between the two scripts and make it more clear what grepTest is doing.


mattjala:
This was already taken care of in #5422; closing this.

PR #5473, opened 2025-04-22
Mask values in h5dump output files

mattjala:
Some values in h5dump output are potentially inconsistent even between files created using the exact same set of instructions with the Native connector. Specifically, these values are the offset of objects within the file and the index assigned to anonymous datatypes by h5dump.

- Replace the displayed values of these fields with 'XXX' in saved reference files

- Update the CMake testing code to perform this masking on output files before comparing them to the saved reference files 

This is done through the TEST_FILTER argument to runTest/grepTest. Because some tests already provide one filter to apply, adding the masking capability required changing runTest and grepTest to support applying multiple filters from a provided semicolon-separated list of filters.

- Add a command line script to convert specific output files to the masked form (h5dump_mask_output.sh)

This was used to convert existing test files, and will be useful if more test files are added, though it is not currently used during testing runs.


byrnHDF (changes requested):
Adding the capability to have  multiple filters  is a good idea. However masks should use the TEST_MASK argument.

The runTest.cmake file probably should pull the TEST_MASK_STORE and TEST_MASK_MOD options out of runTest.cmake to the calling tools and make it consistent.

TEST_MASK_ERROR is a different issue and probably should be dealt with in a different issue.

mattjala:
> Adding the capability to have multiple filters is a good idea. However masks should use the TEST_MASK argument.
> 
> The runTest.cmake file probably should pull the TEST_MASK_STORE and TEST_MASK_MOD options out of runTest.cmake to the calling tools and make it consistent.

Right now, `TEST_MASK` just removes the expression entirely. If we modified it to perform a replace operation in order to do the masking, then it'd be an exact copy of `TEST_FILTER/TEST_FILTER_APPEND`.

Masking should probably use a generic find/replace argument to runTest - there's no reason for runTest to care whether the specific operation is 'masking'. 

Regardless, changes to runTest seem out of scope for this. `TEST_FILTER` is already used for some masks (see `ADD_H5_COMP_TEST`), so I think reworking runTest should be considered separately.

byrnHDF:
> Regardless, changes to runTest seem out of scope for this. `TEST_FILTER` is already used for some masks (see `ADD_H5_COMP_TEST`), so I think reworking runTest should be considered separately.

Agree the larger changes is a separate issue - the whether to use in the CMake add_test or incorporate in the runTests should be decided on how much it is to be used - if every test in the use case needs to do it - put it in the runTest - if only a special couple tests need to use the mask, put it in the add_test command.

mattjala:
- Output files and reference files are now temporarily masked at test time, with the copies on disk still having their real values

- Factored out the masking work into a macro

- In order to avoid running into secondary behavior for TEST_FILTER, I moved the masking behavior to the (formerly unused) TEST_MASK and TEST_MASK_REPLACE arguments

config/cmake/HDF5Macros.cmake
```diff
@@ -118,3 +118,36 @@ macro (H5_CREATE_VFD_DIR)
     file (MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/${vfdtest}")
   endforeach ()
 endmacro ()
+
+# Apply mask replacements to FILENAME and write the results to FILENAME_masked
+# Uses TEST_MASKS and TEST_MASKS_REPLACE from its execution environment
+macro(H5_MASK_FILE FILENAME)
```
mattjala, line 124: This macro uses variables from its environment that aren't specified as arguments. This is not ideal, but trying to directly pass in the list of mask replace expressions as a parameter proved difficult due to 1. It collapsing from a list to a string, and 2. Loss of special character escaping in a way that created a string CMake can't directly evaluate at all, preventing a simple `\ -> \\` replacement to re-escape it.

byrnHDF (approved):

lrknox (approved):

PR #5475, opened 2025-04-22
Use ubuntu-22.04 for ubuntu clang

byrnHDF:
ubuntu clang needs libtinfo5 which can only be used on ubuntu-22.04 runners.


lkurz (approved):
LGTM

lrknox (approved):

Issue #5132, opened 2024-11-19
h5repack oddities


markcmiller86:
**Describe the bug**
I have been playing with `h5repack` in the context of user-defined filters (e.g. passing `-f <list of objects>:<name of filter>=<filter parameters>`) and have run across a number of issues which I briefly enumerate here.

#### Use of colon character
- When no specific objects are listed ahead of the colon character (`:`) in the argument to the `-f` option, is the colon character still needed? I think it is unclear in the documentation whether it is or not. I would recommend that it *work* in this circumstance either way. Then, you don't have to worry about making documentation any clearer. So, if no specific objects are listed ahead of the colon, both `-f :UD=XXXXX,...` and `-f UD=XXXXXX,...` should both work.

#### Handling of `cd_value_count` for `UD=` filters
- For a `UD` filter with no params (e.g. `cd_vals`), is a `0` still nonetheless needed for the `cd_value_count` parameter. This does seem to work either way. That is good.
- If you specify a `cd_value_count` of `K` and provide fewer than `K` values, this is caught as an error and `h5repack` does not proceed. This is also good. The reverse is not, however, true. If you specify a `cd_value_count` of zero but provide more values, no error is generated and, indeed, the resulting repacked datasets show those extra values as part of the filter parameters.
 
#### When `filter_flag` is `1` (mandatory)
- When a specific object is listed ahead of the colon character and the `filter_flag` is `1` indicating the filter is *mandatory*, then if the `can_apply` for the associated filter returns anything other than 1 for that object, that should be treated as an outright error by `h5repack`. An error message should be printed and the return value from `h5repack` should not be `0`.
- When the requested filter ID is not succesfully loaded (e.g. after all `dlopen()` calls there is nothing found matching the specified filter ID) and `filter_flag` is `1` indicating the filter is *mandatory, that should be treated as an outright error by `h5repack`. An error message should be printed and the return value from `h5repack` should not be `0`.

A stub filter I was using to test is [foo.c.gz](https://github.com/user-attachments/files/17822078/foo.c.gz)

markcmiller86:
Related to this, I am getting a segv on HDF5-1.14.3 with a slight modification of the filter attached above.

[foo.c.gz](https://github.com/user-attachments/files/17823894/foo.c.gz)
[Makefile.gz](https://github.com/user-attachments/files/17823905/Makefile.gz)

I used the attached makefile with the command...

```
[...]

PR #5480, opened 2025-04-23
Corrected colon behavior and UD with zero elements

byrnHDF:
Addresses issues stated in #5132, but the comment about a segfault might have been addressed in the code changes.


mattjala:
This should include a test to prevent regression. It looks like a duplicate of one of the existing tests in `h5repack/CMakeTests.cmake` with a slight change to the arguments would suffice.

byrnHDF:
> This should include a test to prevent regression. It looks like a duplicate of one of the existing tests in `h5repack/CMakeTests.cmake` with a slight change to the arguments would suffice.

Maybe for the zero case with added params. but it would need to fail if extra params are added.

bmribler (dismissed):

lrknox (approved):

lrknox:
> This should include a test to prevent regression. It looks like a duplicate of one of the existing tests in `h5repack/CMakeTests.cmake` with a slight change to the arguments would suffice.

@mattjala, Allen added a test after your comment.  Does it sufficiently address this?

mattjala (approved):

PR #5482, opened 2025-04-24
Add missing test files to h5copygentest

mattjala:
Generate `tudfilter.h5` and `tudfilter2.h5` through h5copygentest, to match the checking in h5copy testfiles

### DIFF
--- a/tools/test/h5copy/h5copygentest.c
+++ b/tools/test/h5copy/h5copygentest.c
@@ -22,6 +22,8 @@
 #define HDF_FILE2        "h5copy_ref.h5"
 #define HDF_EXT_SRC_FILE "h5copy_extlinks_src.h5"
 #define HDF_EXT_TRG_FILE "h5copy_extlinks_trg.h5"
+#define HDF_FILE3        "tudfilter.h5"
+#define HDF_FILE4        "tudfilter2.h5"
 
 /* objects in HDF_FILE1 */
 #define DATASET_SIMPLE     "simple"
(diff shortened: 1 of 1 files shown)


tools/test/h5copy/h5copygentest.c
```diff
@@ -960,6 +962,36 @@ Test_Extlink_Copy(void)
...
+        for (j = 0; j < 10; j++) {
+            data[i][j] = count++;
+        }
+    }
+
+    file_id      = H5Fcreate(filename, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+    dataspace_id = H5Screate_simple(2, dims, NULL);
+    datatype_id  = H5Tcopy(H5T_NATIVE_INT32);
+    dataset_id =
+        H5Dcreate2(file_id, "dynlibud", datatype_id, dataspace_id, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
```
jhendersonHDF, line 986: This won't create quite the same file. A user-defined filter with the ID 300 and comment "dynlibud" (which I think is just the name that h5dump is interpreting) needs to be added to the dataset. If you can track down the original code that created the files (if it exists), that would be better.

jhendersonHDF: (That also means the dataset will need to be chunked 10x5 according to h5dump)

jhendersonHDF: Looks like the filter is from https://github.com/HDFGroup/hdf5/blob/develop/tools/test/h5dump/dynlib_dump.c and the file is from https://github.com/HDFGroup/hdf5/blob/develop/tools/test/h5dump/h5dumpgentest.c#L113 / https://github.com/HDFGroup/hdf5/blob/develop/tools/test/h5dump/h5dumpgentest.c#L11196. Probably best again not to duplicate the code, but at least we have the original.

mattjala:
Superseded by #5515

PR #5497, opened 2025-04-28
Generate VOL tests for h5copy

mattjala:
- Repack h5copy testfiles with h5repack to put them into the form that the VOL connector loaded through fetchcontent expects (e.g. multiple files, objects in object store)

- If a VOL connector is provided at build time, CMake will create counterparts of the h5copy tests for that connector, with names of the form <original test name>-<vol name>

- Add h5copy tests to the VOL workflows

The h5copy tests aren't added to the adios2 and the LOG VOL CI, because the workflows don't use fetchcontent to set up those connectors.

The h5copy tests are mostly disabled for the REST VOL because it doesn't factor in current working directory to the HSDS paths it accesses, so the h5copy tests fail due to being unable to find the files.

- Factor some shared VOL parsing logic into a new macro

VOL CI results after this change are [here](https://github.com/mattjala/hdf5/actions/runs/14735468898/job/41360331314)

### DIFF
--- a/.github/workflows/vol_cache.yml
+++ b/.github/workflows/vol_cache.yml
@@ -162,8 +162,3 @@ jobs:
         if: ${{ ! matrix.async }}
         run: |
           ctest --build-config ${{ inputs.build_mode }} -VV -R "HDF5_VOL_vol-cache" .
-
-      - name: Test HDF5 cache VOL connector with HDF5 tools
-        working-directory: ${{ github.workspace }}/hdf5/build
-        run: |
-          ctest --build-config ${{ inputs.build_mode }} -VV -R "H5COPY.*async"
--- a/.github/workflows/vol_cache.yml
+++ b/.github/workflows/vol_cache.yml
@@ -162,3 +162,8 @@ jobs:
         if: ${{ ! matrix.async }}
         run: |
           ctest --build-config ${{ inputs.build_mode }} -VV -R "HDF5_VOL_vol-cache" .
+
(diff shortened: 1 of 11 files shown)


.github/workflows/vol_cache.yml
```diff
@@ -162,3 +162,8 @@ jobs:
         if: ${{ ! matrix.async }}
         run: |
           ctest --build-config ${{ inputs.build_mode }} -VV -R "HDF5_VOL_vol-cache" .
+
+      - name: Test HDF5 cache VOL connector with HDF5 tools
+        working-directory: ${{ github.workspace }}/hdf5/build
+        run: |
+          ctest --build-config ${{ inputs.build_mode }} -VV -R "H5COPY.*async"
```
jhendersonHDF: Still need to review the rest of this PR, but at a first glance the tools test target names should be consistent with the other tests and begin with "HDF5_VOL_<vol_name>-" (instead of appending to the name) to group them together. Otherwise you might accidentally exclude tests unintentionally with this naming scheme + regex. Consider if tools with support for async I/O and the associated tests are added in the future.

jhendersonHDF: By matching the naming you also don't need this separate workflow step just to run the tools tests. It just gets added in with `ctest --build-config ${{ inputs.build_mode }} -VV -R "HDF5_VOL_vol-cache" .`, for example.

mattjala:
Marking this as draft until I resolve issues with repack file generation

mattjala:
Superseded by #5551

PR #5506, opened 2025-05-02
Add H5Dread_chunk2()

### LINKED ISSUE #5224: Create memory safe version of H5Dread_chunk()
H5Dread_chunk() currently has the signature: `herr_t H5Dread_chunk(hid_t dset_id, hid_t dxpl_id, const hsize_t *offset, uint32_t *filters, void *buf);`

This is potentially not memory safe because hte caller does not specify the size of buf. This is especialy problematic because the fitler can potentially increase the size of the chunk, meaning the only way for a caller to definitely know how big buf needs to be is to call H5Dget_chunk_info() first. We should add a size parameter to H5Dread_chunk(), which the caller initializes with the size of buf, and on exit contains the size of the filtered chunk (the needed size of buf). If the buffer is not big enough to read the entire chunk, the contents of buf are unchanged. This would allow the caller to pass a buffer they think is big enough without creating memory (and potentially security) errors if they're wrong and without needed to call H5Dget_Chunk_info() every time, which adds overhead and thereby partially defeats the purpose of the direct chunk I/O interface.

fortnern:
Version H5Dread_chunk() according to the standard HDF5 API versioning scheme. H5Dread_chunk2() takes an additional parameter that points to a variable containing the size of the buffer on entry, and is set to the needed size on exit. Deprecated H5Dread_chunk1(). The FORTRAN interface still needs to be updated.

Resolves #5224


release_docs/RELEASE.txt
```diff
@@ -171,6 +171,31 @@ New Features
 
     Library:
     --------
+    - The H5Dread_chunk() signature has changed
+
+      A new parameter, nalloc, has been added to H5Dread_chunk(). This parameter
+      contains a pointer to a variable that holds the size of the buffer buf. If
+      *nalloc is not large enough to hol dhte entire chunk being read, no data
```
brtnfld: "hol dhte"

src/H5Dpublic.h
```diff
@@ -1250,10 +1250,11 @@ H5_DLL herr_t H5Dwrite_chunk(hid_t dset_id, hid_t dxpl_id, uint32_t filters, con
  *                      dataspace
  * \param[in,out]  filters  Mask for identifying the filters in use
  * \param[out]  buf     Buffer containing data to be read from the chunk
+ * \param[in,out]  nalloc   Size of buf in bytes
```
brtnfld: Is the naming convention `nalloc`, or `buf_size`? I see both being used. What should be used going forward? For Fortran, I think we mostly use buf_size.

src/H5Dpublic.h
```diff
@@ -1276,22 +1277,31 @@ H5_DLL herr_t H5Dwrite_chunk(hid_t dset_id, hid_t dxpl_id, uint32_t filters, con
  *          \p buf is the memory buffer containing the chunk read from
  *          the dataset in the file.
  *
- * \attention Exercise caution when using H5Dread_chunk() and
+ *          \p nalloc must be passed as a pointer to a variable holding the
+ *          allocated size, in bytes, of the memory buffer \p buf. On exit,
+ *          \p *nalloc is set to the buffer size needed to read the chunk, which
+ *          is the same as the size of the chunk on disk. If the value of
+ *          \p *nalloc passed in was insufficient to read the entire, chunk, no
+ *          data is read. \p buf may be passed as NULL as long as \p *nalloc
```
brtnfld: Why does `nalloc` need to be set to zero when `buf ` is NULL?

src/H5Dpublic.h
fortnern: I used nalloc because that's what the closest existing analogs for this pattern, H5Tencode etc., use. I do think buf_size is a better name though, so I wouldn't be opposed to changing it.

src/H5Dpublic.h
fortnern: I suppose we could allow a nonzero (uninitialized *nalloc) in this case for convenience. My reasoning was that by passing *nalloc > 0 the app is asserting there is space in the buffer, and therefore it's inconsistent to pass a NULL buffer.

fortnern: Now that I look at it, H5Tencode allows *nalloc > 0 with buf == NULL, so I'll change this.

release_docs/RELEASE.txt
brtnfld: fixed

src/H5Dpublic.h
brtnfld: I changed it to `buf_size`.

src/H5Dpublic.h
brtnfld: Ok, then I will allow this behaviour in the FORTRAN wrapper.

brtnfld:
@fortnern , I finished with the Fortran additions, you might want to bring the PR in sync with develop

src/H5Dpublic.h
brtnfld: done

brtnfld:
@fortnern, I don't think those failing test are related to my changes. All tests pass on my OpenSUSE

fortran/src/H5Dff.F90
```diff
@@ -1780,6 +1786,62 @@ SUBROUTINE h5dfill_f(fill_value, fill_type_id, buf, buf_type_id, space_id, hdfer
...
+!! See C API: @ref H5Dread_chunk2()
+!!
+ SUBROUTINE h5dread_chunk_f(dset_id, offset, filters, buf, buf_size, hdferr, dxpl_id)
+    IMPLICIT NONE
+
+    INTEGER(HID_T)    , INTENT(IN)  :: dset_id
+    INTEGER(HSIZE_T)  , INTENT(IN), DIMENSION(:) :: offset
+    INTEGER           , INTENT(INOUT) :: filters
+    TYPE(C_PTR)                       :: buf
+    INTEGER(HID_T)    , INTENT(INOUT) :: buf_size
```
epourmal: Should the type be INTEGER(SIZE_T)?

brtnfld: good catch, fixed.

hl/src/H5DO.c
```diff
@@ -60,7 +60,7 @@ herr_t
 H5DOread_chunk(hid_t dset_id, hid_t dxpl_id, const hsize_t *offset, uint32_t *filters, void *buf)
 {
     /* Call underlying H5D function */
-    if (H5Dread_chunk(dset_id, dxpl_id, offset, filters, buf) < 0)
+    if (H5Dread_chunk1(dset_id, dxpl_id, offset, filters, buf) < 0)
```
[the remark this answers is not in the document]
fortnern: Thanks copilot, but this code is within an #ifndef H5_NO_DEPRECATED_SYMBOLS block so there should be no issue. Also calling the macro in the library is a bad idea since the signature is different depending on which version gets called.

src/H5version.h
```diff
@@ -368,6 +368,10 @@
   #define H5Dopen_vers 2
 #endif /* !defined(H5Dopen_vers) */
 
+#if !defined(H5Dread_chunk_vers)
+  #define H5Dread_chunk_vers 1
+#endif /* !defined(H5Dread_chunk_vers) */
```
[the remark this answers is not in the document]
fortnern: Copilot doesn't understand our API versioning scheme

PR #5512, opened 2025-05-04
build: fix i386 openbsd-7.7 threads link error

### LINKED ISSUE #5508: `HDF5_ENABLE_THREADS:BOOL=OFF` causes build failure
Platform:  i386 OpenBSD 7.7 

```
ld: error: undefined reference due to --no-allow-shlib-undefined: pthread_mutexattr_init
```

Full report with 13 errors:

https://my.cdash.org/viewBuildError.php?buildid=2909570

hyoklee:
close #5508 

proof: https://my.cdash.org/viewTest.php?onlyfailed&buildid=2911051

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,4 @@
 cmake_minimum_required (VERSION 3.18)
-# set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "" FORCE)
-# set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "" FORCE)
 project (HDF5 C)
 
 if (POLICY CMP0074)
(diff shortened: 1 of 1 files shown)


byrnHDF (changes requested):
There are problems with the HDF5_ENABLE_THREADS option - it is used improperly

CMakeLists.txt
```diff
@@ -799,35 +799,37 @@ endif ()
 
 # Determine if a threading package is available on this system
 option (HDF5_ENABLE_THREADS "Enable thread support" ON)
```
byrnHDF, line 801: This option should be removed 

Then in src/CMakeLists.txt should remove the generator expressions involving the option.
Also other options should not use the state of options to forces defines.
Also there should probably be a line that adds the Threads::Threads library either to the link_libs variable or a new threads_lib variable

PR #5514, opened 2025-05-06
Add bitgroom filter support

byrnHDF:
Depends on https://github.com/HDFGroup/hdf5_plugins/pull/179
Update compression lib version settings and add bitgroom examples.
Add settings for next filter requested, bitround.


HDF5Examples/C/H5FLT/tfiles/h5ex_d_bitgroom.ddl
```diff
@@ -0,0 +1,287 @@
...
+      DATATYPE  H5T_IEEE_F32LE
+      DATASPACE  SIMPLE { ( 32, 64 ) / ( 32, 64 ) }
+      STORAGE_LAYOUT {
+         CHUNKED ( 4, 8 )
+         SIZE 8192 (1.000:1 COMPRESSION)
+      }
+      FILTERS {
+         USER_DEFINED_FILTER {
+            FILTER_ID 32022
+            COMMENT BitGroom filter (Zender, 2016 GMD: http://www.geosci-model-dev.net/9/3199/2016)
```
hyoklee, line 13: Use https.

byrnHDF: That usage is from the original code  and could cause issues with current netcdf files that use that filter.

lrknox (dismissed):

HDF5Examples/C/H5FLT/h5ex_d_bitgroom.c
```diff
@@ -0,0 +1,239 @@
...
+    dset_id = -1;
+    H5Pclose(dcpl_id);
+    dcpl_id = -1;
+    H5Sclose(space_id);
+    space_id = -1;
+    H5Fclose(file_id);
+    file_id = -1;
+    status  = H5close();
+    if (status < 0) {
+        printf("/nFAILED to close library/n");
```
jhendersonHDF, line 147: `/` -> `\` (May want to check if this is a copy-paste from other examples)

config/cmake/cacheinit.cmake
```diff
@@ -86,6 +86,16 @@ set (PLUGIN_PACKAGE_NAME "pl" CACHE STRING "Name of PLUGIN package" FORCE)
...
 
+###########
+# bitgroom
+###########
+set (BITGROOM_PACKAGE_NAME "bitgroom" CACHE STRING "Name of BITGROOM package" FORCE)
+
+###########
+# bitround
+###########
+set (BITGROUND_PACKAGE_NAME "bitround" CACHE STRING "Name of BITROUND package" FORCE)
```
jhendersonHDF: "BITGROUND" -> "BITROUND" ?

HDF5Examples/C/H5FLT/h5ex_d_bitgroom.c
byrnHDF: This is from the original source

config/cmake/cacheinit.cmake
byrnHDF: Fixed

HDF5Examples/C/H5FLT/h5ex_d_bitgroom.c
jhendersonHDF: Sure but it's not a bad idea to go ahead and fix it since it's wrong

byrnHDF: It then needs to fixed in the plugins code first

byrnHDF: But then all the examples in the plugins will need to fixed because they all do the same

jhendersonHDF: Which is why:

> May want to check if this is a copy-paste from other examples

Not exactly a big deal but definitely a why not

byrnHDF: Yes it is probably all cut-n-paste from the beginning

jhendersonHDF: Might as well fix them all at once separately I guess

jhendersonHDF (approved):

lrknox (approved):

PR #5515, opened 2025-05-06
Merge test file generators for tools

### LINKED ISSUE #5117: Move VOL tools testing to the library

mattjala:
This will allow gentest scripts to share routines between each other, so that cases where multiple tools tests require the same file can be handled without directly copying the files or duplicating file generation routines.

- Split the tool gentest files into source and header portions.

- Actual test file generation is now done through a single script with command line options


tools/test/h5stat/h5stat_gentest.c
```diff
@@ -623,9 +65,9 @@ main(void)
     if (gen_err_refcount(ERR_REFCOUNT_FILE) < 0)
         goto error;
 
-    return EXIT_SUCCESS;
```
mattjala, line 626: EXIT_SUCCESS/FAILURE aren't used consistently across all of the gentest scripts, and were the only remaining symbol that required including the HDF5 header directly here, so it seemed like a good idea to change these to direct values to remove the dependency.

tools/test/misc/h5clear_gentest.c
```diff
@@ -11,308 +11,21 @@
...
-        H5Sclose(sid);
-        H5Dclose(did);
-        H5Fclose(fid);
-        H5Pclose(fcpl);
-    }
-    H5E_END_TRY
-    return 1;
-} /* gen_enhance_files() */
-
+#define FILENAME_ENHANCE_6                                                                                   \
```
mattjala: These entries were removed from the FILENAME_ENHANCE array since they're used in a different manner, and because doing so allowed the core/script split to be done more cleanly. Entries 6 and 7 were only used in `main`, while 0-5 were only used in the routines.

tools/test/h5stat/h5stat_gentest.c
jhendersonHDF: Note that it's generally best practice to use `EXIT_SUCCESS` and `EXIT_FAILURE` rather than direct values and they aren't from HDF5, they're from `<stdlib.h>`. You could just include that somewhere in these files, but I don't see why you shouldn't just keep including `hdf5.h`.

h5repackgentest_core.c
jhendersonHDF, line 1: Looks like an accidentally checked in file.

tools/test/h5copy/h5copygentest.c
jhendersonHDF, line 1: What goes in a "core" vs regular .c file here? If `main()` is the only thing left in all the main files and the entirety of the test generator functions are in the "core" files, then at that point the main files are basically useless and we might as well go the last few steps of having a single generator program that just links with the various new libraries (I'd also probably remove "core" from the names if we do that).

mattjala: Having the core and regular .c file all in the same file led to linker conflicts over the 'main' routine when trying to share specific generator routines between different scripts. The idea was to separate the library of usable generator functions from the way that the particular test setup for e.g. h5copy actually uses them.

jhendersonHDF: Sure, but at this point do any of the `main()`s for these generators actually do anything beyond just calling the test file generator functions in sequence? If not, I don't see any reason to keep a bunch of separate executables around.

mattjala:
- Moved the 'core' definition files back to the original h5<tool>gentest.c files
- Merged the actual scripts into h5gentest.c, which takes an argument to decide which sets of files to create

h5clear_gentest was excluded from h5gentest because it exits in a specific way to leave IDs open.
Similarly, h5perf_gentest was excluded because it doesn't seem to be used for h5perf testing in the same way as the other gentest scripts.

tools/test/h5gentest.c
```diff
@@ -0,0 +1,607 @@
...
+    if (gent_ub(H5JAM_FILE7, 0, 0) < 0)
+        goto error;
+    if (gent_ub(H5JAM_FILE8, 512, PATTERN_LEN) < 0)
+        goto error;
+    if (gent_ub(H5JAM_FILE9, 1024, 513) < 0)
+        goto error;
+
+    return;
+error:
+    fprintf(stderr, "h5jam test generator FAILED\n");
```
jhendersonHDF: May want to make the main generator functions for each tool return some sort of value so that can at least be captured when failures happen.

tools/test/h5gentest.c
```diff
@@ -0,0 +1,607 @@
...
+#include "h5copygentest.h"
+#include "h5diffgentest.h"
+#include "h5dumpgentest.h"
+#include "h5format_convertgentest.h"
+#include "h5jamgentest.h"
+#include "h5repackgentest.h"
+#include "h5statgentest.h"
+#include "h5repartgentest.h"
+
+#define H5DIFF_FILE1  "h5diff_basic1.h5"
```
jhendersonHDF: Where possible these `#define`s, typedefs, etc. for a tool should be moved backed to the new header file for that tool to keep everything together in separate modules.

tools/test/h5gentest.c
```diff
@@ -0,0 +1,607 @@
...
+gen_h5repart_files(void)
+{
+    gent_repart_family();
+}
+
+/*
+ * Generate the binary hdf5 files used for tools tests
+ */
+/*
+ * Generate the binary hdf5 files used for tools tests
```
jhendersonHDF, line 411: Duplicate

tools/test/h5gentest.c
```diff
@@ -0,0 +1,607 @@
...
+    bool run_h5repack = false;
+    bool run_h5stat   = false;
+    bool run_h5repart = false;
+
+    /* Parse command line arguments */
+    if (argc > 1) {
+        run_all = false;
+
+        for (i = 1; i < argc; i++) {
+            if (strcmp(argv[i], "-h") == 0 || strcmp(argv[i], "--help") == 0) {
```
jhendersonHDF: This is fine, though probably better to use `H5_get_option()`.

jhendersonHDF:
Just a couple minor comments but this looks pretty good to me.

tools/test/h5gentest.c
mattjala: The defines remaining in `h5gentest.c` are all filenames passed as parameters to the file creation functions. Moving them into tool-specific headers would require changing the functions to use fixed test names, which will create issues if we want to create duplicates of certain test files later on.

jhendersonHDF: `h5gentest.c` includes all the header files for all the tools, so every generator function in this file will still have access to the macros. It's just nitpicking on location as the filenames are specific to the test files for a particular tool rather than specific to this file. If some tool really wants to create a file from another tool with a different name than the one defined in that tool's header file, it can just do that or can re-use the name from the tool's header file as a base name and modify it as needed.

jhendersonHDF: My initial phrasing might have been confusing - I just meant that we should move the macros into _the_ new header file for each tool (`h5dumpgentest.h`, `h5diffgentest.h`), not _a_ new header file for each tool.

tools/test/h5diff/h5diffgentest.h
```diff
@@ -0,0 +1,116 @@
+/*
```
jhendersonHDF: One last change (or two): should have the HDF5 copyright header here and I'm not sure what the below comment was related to but it would be good to find that.

mattjala: This was a duplicated comment regarding ` H5TOOLS_MALLOCSIZE` in h5diffgentest.c. Removed.

jhendersonHDF (approved):

fortnern (approved):

jhendersonHDF:
The commit title/description for this PR could likely use a bit of editing before merge as the purpose has changed a bit from the original idea.

PR #5415, opened 2025-04-02
Suppress unhelpful error messages from tentative file opens in h5tools_fopen()

jhendersonHDF:
When developing the DAOS VOL connector, it was requested that the library/tools try to fallback to opening a file with what's available in terms of VOLs/VFDs when the file can't be opened by the default setup, which is usually native+sec2. While I'm now thinking this feature was misguided and should be reworked, it's currently polluting the error stack from at least `h5dump` in a very unhelpful way. The new pause/resume feature for error stacks allows cleaning up of these messages to put the error stacks back to how they used to be years ago. Compare the current (snipped a bit due to length):

```bash
h5dump --enable-error-stack=2 bogus.h5
```

```bash
HDF5-DIAG: Error detected in HDF5 (2.0.0):
  #000: H5F.c line 821 in H5Fopen(): unable to synchronously open file
    major: File accessibility
    minor: Unable to open file
  #001: H5F.c line 782 in H5F__open_api_common(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #002: H5VLcallback.c line 3869 in H5VL_file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #003: H5VLcallback.c line 3718 in H5VL__file_open(): open failed
    major: Virtual Object Layer
    minor: Can't open object
  #004: H5VLnative_file.c line 128 in H5VL__native_file_open(): unable to open file
    major: File accessibility
    minor: Unable to open file
  #005: H5Fint.c line 1924 in H5F_open(): can't try opening file
    major: File accessibility
    minor: Unable to open file
  #006: H5FD.c line 966 in H5FD_open(): can't open file
    major: Virtual File Layer
    minor: Unable to open file
  #007: H5FDsec2.c line 285 in H5FD__sec2_open(): unable to open file: name = 'bogus.h5', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0
    major: File accessibility
    minor: Unable to open file
... (repeat of the above)
...
  #007: H5FDstdio.c line 348 in H5FD_stdio_open(): file doesn't exist and CREAT wasn't specified
    major: Low-level I/O
    minor: Unable to open file
...
  #007: H5FDcore.c line 759 in H5FD__core_open(): unable to open file
    major: File accessibility
    minor: Unable to open file
...
  #007: H5FDfamily.c line 745 in H5FD__family_open(): file names not unique
    major: Virtual File Layer
    minor: File already exists
...
...
[...]

PR #5518, opened 2025-05-06
Set subfiling vfd default enabled=ON 

### LINKED ISSUE #3985: Enable the subfiling VFD by default

lrknox:
and option available when parallel is enabled and not on win32.

Also turned on sorting for brief lists in doxygen files.

Fixes #3985.


CMakeLists.txt
```diff
@@ -835,32 +836,18 @@ set (HDF5_SRC_INCLUDE_DIRS
     ${HDF5_SRC_INCLUDE_DIRS}
     ${H5FD_SUBFILING_DIR}
 )
-option (HDF5_ENABLE_SUBFILING_VFD "Build Parallel HDF5 Subfiling VFD" OFF)
+cmake_dependent_option(HDF5_ENABLE_SUBFILING_VFD "Build Parallel HDF5 Subfiling VFD" ON "HDF5_ENABLE_PARALLEL AND NOT WIN32" OFF)
 if (HDF5_ENABLE_SUBFILING_VFD)
-  if (WIN32)
```
jhendersonHDF, line 840: I believe this check is still needed. `cmake_dependent_option` just hides the option, but someone could still set it on Win32.

jhendersonHDF:
The last I tried to do this, nvhpc had issues due to `MPI_Init_thread()` failing, so I guess we'll see if that's still the case.

CMakeLists.txt
byrnHDF: I'd like to know for sure - as I believe the documentation states:
"condition": A boolean expression involving other CMake options. It must evaluate to true or false.
Of course NOT WIN32 might not evaluate correctly. Then we would need to add if NOT WIN32 around the whole block.

jhendersonHDF: What I mean is that CMake hides the option, but I don't think that means it's actually gone. I think it's similar to just being marked with `mark_as_advanced()`.

byrnHDF: Yes it hides the option but the condition forces a value if it is hidden

CMakeLists.txt
```diff
@@ -835,32 +836,18 @@ set (HDF5_SRC_INCLUDE_DIRS
     ${HDF5_SRC_INCLUDE_DIRS}
     ${H5FD_SUBFILING_DIR}
 )
-option (HDF5_ENABLE_SUBFILING_VFD "Build Parallel HDF5 Subfiling VFD" OFF)
+cmake_dependent_option(HDF5_ENABLE_SUBFILING_VFD "Build Parallel HDF5 Subfiling VFD" ON "HDF5_ENABLE_PARALLEL AND NOT WIN32" OFF)
```
jhendersonHDF: Also, it looks like this syntax for the condition is only supported as of CMake 3.22 and otherwise has to be a semicolon-separated list.

lrknox:
> The last I tried to do this, nvhpc had issues due to `MPI_Init_thread()` failing, so I guess we'll see if that's still the case.

Yes,  *** An error occurred in MPI_Init_thread
2155: *** on a NULL communicator
Test #2155: H5WATCH_ARGS-h5watch-w-err-dset2 ...........................................***Failed 
1 of 110 failed tests.  Also with this error message:
[fv-az2211-563:37654] [[INVALID],INVALID] ORTE_ERROR_LOG: A system-required
2162:   executable either could not be found or was not executable by this user in
2162:   file ../../../../../orte/mca/ess/singleton/ess_singleton_module.c at line
2162:   172

CMakeLists.txt
byrnHDF: Since we are still on 3.18 then we need to use the semis

CMakeLists.txt
lrknox: I ran " ccmake -DMPIEXEC_PREFLAGS:STRING="-n;6" -DHDF5_ENABLE_SUBFILING_VFD=ON ../lrk-hdf5", then configured without any other changes.  The subfiling vfd option was missing and libhdf5.settings reported Subfiling VFD  OFF.

CMakeLists.txt
lrknox: I will fix this.

jhendersonHDF:
> Yes, *** An error occurred in MPI_Init_thread 2155: *** on a NULL communicator Test #2155: H5WATCH_ARGS-h5watch-w-err-dset2 ...........................................***Failed 1 of 110 failed tests. Also with this error message: [fv-az2211-563:37654] [[INVALID],INVALID] ORTE_ERROR_LOG: A system-required 2162: executable either could not be found or was not executable by this user in 2162: file ../../../../../orte/mca/ess/singleton/ess_singleton_module.c at line 2162: 172

Yep, this is related to the description I added for https://github.com/HDFGroup/hdf5/pull/5415. One issue is that the tools VFD fallback mechanism attempts to initialize and use the Subfiling VFD for tests where the file can't be opened with the sec2 VFD (the file doesn't exist, it's in a different VFD's format, etc.) which normally wouldn't be a problem, except that (I believe) NVHPC's MPI seems to have issues when NULL is passed for the arguments to `MPI_Init_thread`. Removing the VFD fallback mechanism could "fix" this, but it's really just avoiding the problem.

There's also the question of testing. Subfiling is tested in the daily VFD build workflow currently, but we only run the subfiling-specific tests and none of the tools tests, for example, so is that enough testing?

The last issue is that turning on Subfiling by default means threads becomes a default requirement for parallel builds. Probably not really a big issue, but something to consider.

lrknox:
Yes, threads becomes a default requirement for parallel when subfiling is on by default.  However, if threads are not available, the error message in configure identifies subfiling as problematic:  message (FATAL_ERROR "Subfiling requires thread operations support").  Do we need to add "Disable subfiling with -DHDF5_ENABLE_SUBFILING_VFD=OFF to build parallel HDF5."?

lrknox:
Disabled subfiling_vfd in cmake-nvhpc.yml to avoid MPI error with NULL communicator.  No failing checks now.

byrnHDF (approved):

PR #5537, opened 2025-05-15
Delay dataset layout copy to DCPL

mattjala:
- Optimize dataset opening by delaying the copy of dataset layout to its internal DCPL

Copying the layout immediately slows down workloads that open many small datasets.

The layout copy is delayed until either a user requests the dataset's DCPL, or until the start of an operation that may need to read the layout from the DCPL.

Whether the copying of the layout has been performed yet for a dataset is tracked by a new field on `H5D_shared_t`.

- Added a new test for this behavior in `dsets.c`

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -245,7 +245,7 @@ typedef struct H5D_chunk_coll_fill_info_t {
         haddr_t addr;       /* File address of the chunk */
         size_t  chunk_size; /* Size of the chunk in the file */
         bool    unfiltered_partial_chunk;
-    } * chunk_info;
+    } *chunk_info;
 } H5D_chunk_coll_fill_info_t;
 #endif /* H5_HAVE_PARALLEL */
 
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -245,7 +245,7 @@ typedef struct H5D_chunk_coll_fill_info_t {
         haddr_t addr;       /* File address of the chunk */
         size_t  chunk_size; /* Size of the chunk in the file */
         bool    unfiltered_partial_chunk;
-    } *chunk_info;
+    } * chunk_info;
 } H5D_chunk_coll_fill_info_t;
 #endif /* H5_HAVE_PARALLEL */
 
@@ -764,9 +764,6 @@ H5D__chunk_set_sizes(H5D_t *dset)
     /* Sanity checks */
     assert(dset);
 
-    /* Increment # of chunk dimensions, to account for datatype size as last element */
-    dset->shared->layout.u.chunk.ndims++;
-
     /* Set the last dimension of the chunk size to the size of the datatype */
     dset->shared->layout.u.chunk.dim[dset->shared->layout.u.chunk.ndims - 1] =
         (uint32_t)H5T_GET_SIZE(dset->shared->type);
@@ -829,6 +826,9 @@ H5D__chunk_construct(H5F_t H5_ATTR_UNUSED *f, H5D_t *dset)
         HGOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL, "no chunk information set?");
     if (dset->shared->layout.u.chunk.ndims != dset->shared->ndims)
         HGOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL, "dimensionality of chunks doesn't match the dataspace");
+
+    /* Increment # of chunk dimensions, to account for datatype size as last element */
+    dset->shared->layout.u.chunk.ndims++;
 
     /* Set chunk sizes */
     if (H5D__chunk_set_sizes(dset) < 0)
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -763,6 +763,7 @@ H5D__chunk_set_sizes(H5D_t *dset)
 
     /* Sanity checks */
     assert(dset);
+    assert(dset->shared->layout.u.chunk.ndims > 0);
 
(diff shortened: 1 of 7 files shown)


src/H5Dint.c
```diff
@@ -1262,6 +1270,10 @@ H5D__create(H5F_t *file, hid_t type_id, const H5S_t *space, hid_t dcpl_id, hid_t
             HGOTO_ERROR(H5E_ARGS, H5E_CANTINIT, NULL, "H5Z_has_optional_filter() failed");
 
         if (false == ignore_filters) {
+            /* Flush layout to DCPL before reading */
```
fortnern: This seems wrong to me. H5D_new() populates the DCPL field but not the layout field. The layout field is populated from the DCPL later in this function. So it seems this line would overwrite the layout in the DCPL with the layout from the default dataset?

src/H5Dint.c
```diff
@@ -3021,6 +3033,10 @@ H5D__check_filters(H5D_t *dataset)
         if (fill_status == H5D_FILL_VALUE_DEFAULT || fill_status == H5D_FILL_VALUE_USER_DEFINED) {
             if (fill->fill_time == H5D_FILL_TIME_ALLOC ||
                 (fill->fill_time == H5D_FILL_TIME_IFSET && fill_status == H5D_FILL_VALUE_USER_DEFINED)) {
+                /* Flush layout to DCPL before reading */
+                if (H5D_flush_layout_to_dcpl(dataset) < 0)
```
fortnern, line 3035: Does H5Z_can_apply require a valid layout? I wonder if we could pass the layout separately.

fortnern: Probably not now that I think about it, though maybe we could hack it to work for internal filters

src/H5Dlayout.c
```diff
@@ -630,25 +630,18 @@ H5D__layout_oh_read(H5D_t *dataset, hid_t dapl_id, H5P_genplist_t *plist)
...
-    if (H5D_CHUNKED == dataset->shared->layout.type)
-        dataset->shared->layout.u.chunk.ndims--;
-
-    /* Copy layout to the DCPL */
-    if (H5P_set(plist, H5D_CRT_LAYOUT_NAME, &dataset->shared->layout) < 0)
-        HGOTO_ERROR(H5E_DATASET, H5E_CANTSET, FAIL, "can't set layout");
-
     /* Set chunk sizes */
-    if (H5D_CHUNKED == dataset->shared->layout.type)
+    if (H5D_CHUNKED == dataset->shared->layout.type) {
```
fortnern: Why the braces?

src/H5Dint.c
fortnern: I suppose it doesn't cause an error because H5D__new always sets the layout_copied_to_dcpl flag to true, so this function is a no-op. Still we shouldn't call this here, we can instead add an assertion that that flag is true.

mattjala: You're correct - I'll replace this with an assert.

src/H5Dlayout.c
mattjala: No particular reason - removed for consistency.

src/H5Dint.c
mattjala: As I understand it, `H5Z_can_apply()` eventually invokes filter-defined callbacks to determine whether the filter can be applied with the current DCPL/datatype/dataspace setup. The callback can access arbitrary fields on the DCPL including the layout, so we want the layout available there to reflect the dataset accurately.

qkoziol:
I have significantly changed this piece of code (and related areas) in my incoming threadsafety patches.  Would you consider backburnering this until those are in?

fortnern:
I just finished checking for other uses of dset->shared->dcpl_id, and it looks like this PR handles everything. Would it be possible though, for you to add code to, when NDEBUG is not set, push an invalid layout to dset->shared->dcpl_id whenever it goes out of syync with dset->shared->layout? You could set the type field in the layout to H5D_LAYOUT_ERROR. Along with code to assert the type is not H5D_LAYOUT_ERROR in places that actually use that layout.

This should help minimze the chance that this change creates problems with future code changes.

mattjala:
> I just finished checking for other uses of dset->shared->dcpl_id, and it looks like this PR handles everything. Would it be possible though, for you to add code to, when NDEBUG is not set, push an invalid layout to dset->shared->dcpl_id whenever it goes out of syync with dset->shared->layout? You could set the type field in the layout to H5D_LAYOUT_ERROR. Along with code to assert the type is not H5D_LAYOUT_ERROR in places that actually use that layout.
> 
> This should help minimze the chance that this change creates problems with future code changes.

As far as I'm aware, the layout can't be modified after creation. So the only time the dataset layout and the DCPL layout are out of sync is at creation time. But I can set up an error value layout for that case.

fortnern (dismissed):

src/H5Dchunk.c
```diff
@@ -764,9 +764,6 @@ H5D__chunk_set_sizes(H5D_t *dset)
     /* Sanity checks */
     assert(dset);
 
-    /* Increment # of chunk dimensions, to account for datatype size as last element */
-    dset->shared->layout.u.chunk.ndims++;
```
jhendersonHDF, line 768: It may be useful to add an assertion or error check here that `ndims > 0`. Previously it was inherent to the code but now that this increment has been moved out of this function, a future re-org could mean that the below code ends up performing negative indexing.

jhendersonHDF (approved):

fortnern (approved):

PR #5538, opened 2025-05-16
refactor run scripts to a single macros with includes

### LINKED ISSUE #5484: Remove appending the error test and exit code to the output in tests
CMake provides better ways of getting those results in testing. It could simplify checking test results against reference files.

### LINKED ISSUE #5485: Refactor the cmake scripts (runTest, grepTest,jrunTest) into a single script with CMake includes to break out functionality
There is a lot of duplication in these scripts and is becoming a maintenance burden. The scripts could be broken out into single purpose macros that could be included into a single script used by the the CMake tests.
Also the main script could use the optional arguments to determine which sub-scripts get included.

byrnHDF:
Fixes #5485
Fixes #5484


HDF5Examples/config/cmake/runExecute.cmake
```diff
@@ -70,40 +53,137 @@ if (TEST_ENV_VAR)
...
+  endif ()
+  message (FATAL_ERROR "Failed: Test program ${TEST_PROGRAM} exited != ${TEST_EXPECT}.\n${TEST_ERROR}")
+endif ()
+endmacro ()
+
+#############################################
+# Begin of file filtering
+#############################################
+macro (FILTER_TEST)
+cmake_parse_arguments (TEST "MASK_FILEMASK_ERROR;MASK_STORE;ERRREF" "FOLDER;OUTPUT;REFERENCE;REGEX;MATCH;MASK;FILTER;REF_FILTER;FILTER_REPLACE;MASK_MOD;ARGS" "TEST_" ${ARGN})
```
mattjala: Missing semicolon between `MASK_FILE` and `MASK_ERROR`

HDF5Examples/config/cmake/runExecute.cmake
```diff
@@ -70,40 +53,137 @@ if (TEST_ENV_VAR)
...
+    endif ()
+  endif ()
+  message (FATAL_ERROR "Failed: Test program ${TEST_PROGRAM} exited != ${TEST_EXPECT}.\n${TEST_ERROR}")
+endif ()
+endmacro ()
+
+#############################################
+# Begin of file filtering
+#############################################
+macro (FILTER_TEST)
```
mattjala, line 122: Since `FILTER_TEST` is shared between `config/cmake/runTest.cmake` and `HDF5Examples/config/cmake/runTest.cmake`, it should probably be defined in HDF5Macros.cmake or some other shared file.

HDF5Examples/config/cmake/runExecute.cmake
```diff
@@ -9,15 +9,10 @@
...
-# against a reference file. Exit status of command can also be compared.
-cmake_policy(SET CMP0007 NEW)
-cmake_policy(SET CMP0053 NEW)
+# runExecute.cmake executes a command included by runTest.cmake.
 
-# arguments checking
-if (NOT TEST_TESTER)
-  message (FATAL_ERROR "Require TEST_TESTER to be defined")
-endif ()
+macro (EXECUTE_TEST)
```
mattjala, line 14: It looks like two identical versions of this macro are defined in `config/cmake/runExecute.cmake` and in `HDF5Examples/config/cmake`. Is there a reason to go with this separation instead of using a shared macro?

mattjala: For that matter, is there a good reason to duplicate the entire `runTest.cmake` script between HDF5Examples and the rest of the library? Perhaps this is something that will immediately be addressed by #5540

tools/test/h5diff/expected/h5diff_629.txt
```diff
@@ -1,2 +1,2 @@
 h5diff: <file1.h6>: unable to open file
-EXIT CODE: 2
+
```
mattjala: Extra newline added here

HDF5Examples/config/cmake/runExecute.cmake
byrnHDF: NO.
HDF5Examples is provided as a separate project when a binary is created. There is two processes involved with HDF5Examples 1. build and test during library build and 2. Build and test with binary library installs.

tools/test/h5diff/expected/h5diff_629.txt
byrnHDF: I believe this is necessary as the test passes.

HDF5Examples/config/cmake/runExecute.cmake
byrnHDF: Fixed

tools/test/h5diff/expected/h5diff_629.txt
byrnHDF: I was wrong the test is not even run!
```
# Disabling this test as it hangs - LRK 20090618
# 6.29  non valid files
#ADD_H5_TEST (h5diff_629 2 file1.h6 file2.h6)

```

byrnHDF: Therefore will remove extra nl

byrnHDF:
One thing to keep in mind that the runTest and runExecute(better name?) files execute in a private environment from all the other macro and test files. While possible to include other files it makes the code for these test macros more difficult.

config/cmake/runTest.cmake
```diff
@@ -46,6 +46,7 @@ EXECUTE_TEST (TEST_FOLDER ${TEST_FOLDER}
                TEST_CLASSPATH ${TEST_CLASSPATH}
                TEST_NOERRDISPLAY ${TEST_NOERRDISPLAY}
 )
+  message (STATUS "TEST_MASK_ERROR: ${TEST_MASK_ERROR}")
```
mattjala: Looks like a debug message that can be removed

mattjala (dismissed):

config/cmake/runTest.cmake
byrnHDF: yes.

byrnHDF: fixed

mattjala (approved):

lrknox (approved):

PR #5540, opened 2025-05-19
Refactor the config/cmake folder #5446

byrnHDF:
Fixes #5446

PR #5544, opened 2025-05-21
Fix incorrect test file masking

mattjala:
- Fix the reference file not being masked

- Fix the output comparison using the unmasked reference file


byrnHDF:
Can we wait on this till after #5538 and #5540 are in

PR #5551, opened 2025-05-28
Generate VOL tests for h5copy

mattjala:
- If a VOL connector is provided at build time, CMake will create counterparts of the h5copy tests for that connector, with names of the form `HDF5_VOL_<vol_name>-<test_name>`

- Add h5copy tests to the VOL workflows

The h5copy tests aren't added to the adios2 and the LOG VOL CI, because the workflows don't use fetchcontent to set up those connectors.

The h5copy tests are mostly disabled for the REST VOL because it doesn't factor in current working directory to the HSDS paths it accesses, so the h5copy tests fail due to being unable to find the files.

----

> [!IMPORTANT]
> Generate VOL-specific `h5copy` tests using CMake for each VOL connector, updating workflows and macros to support this functionality.
> 
>   - **Behavior**:
>     - Generate `h5copy` tests for each VOL connector at build time using `CMakeTests.cmake`.
>     - Tests are named `HDF5_VOL_<vol_name>-<test_name>`.
>     - Excludes `adios2` and `LOG VOL` due to setup method.
>     - REST VOL tests mostly disabled due to path issues.
>   - **CMake Macros**:
>     - `HDF5_GET_VOL_TGT_INFO` in `HDF5Macros.cmake` to extract VOL connector info.
>     - Modify `ADD_H5_TEST`, `ADD_H5_UD_TEST`, and `ADD_H5_UD_ERR_TEST` in `CMakeTests.cmake` to support VOL-specific tests.
>   - **Workflows**:
>     - Update `vol_cache.yml` and `vol_rest.yml` to include `h5copy` tests.
>     - Exclude `H5COPY` tests in REST VOL workflow due to path issues.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for cbbf08e9a379d8ee3ee43e1a748f4ddb3707bb12. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


bmribler (dismissed):

fortnern:
Ideally we should find a way to get rid of the repeated code in the h5copy cmake file, but this looks good otherwise (as far as I can tell, I am not a cmake expert) and should be fine for now

fortnern (dismissed):

tools/test/h5copy/CMakeTests.cmake
```diff
@@ -58,215 +100,328 @@
   #
   # Perform h5copy according to passing parameters
   #
-  macro (ADD_H5_F_TEST testname resultcode infile fparam vparam sparam srcname dparam dstname)
+  macro (ADD_H5_F_TEST testname vol env resultcode infile fparam vparam sparam srcname dparam dstname)
     # Remove any output file left over from previous test run
     add_test (
-        NAME H5COPY_F-${testname}-clear-objects
-        COMMAND ${CMAKE_COMMAND} -E remove ./testfiles/${testname}.out.h5
+        NAME HDF5_VOL_${vol}-H5COPY_F-${testname}-clear-objects
```
jhendersonHDF: In general these tests should keep their original name when using the native VOL and only include the VOL prefix when using an external VOL that's being tested against. I'd suggest setting a local variable for the test name based on the VOL name given.

tools/test/h5copy/CMakeTests.cmake
```diff
@@ -58,215 +100,328 @@
...
+        COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5delete> ./testfiles/${testname}.out.h5
     )
 
+    set_tests_properties (HDF5_VOL_${vol}-H5COPY_F-${testname}-clear-objects PROPERTIES
+      ENVIRONMENT "${env}"
```
jhendersonHDF: Similarly, we should probably avoid modifying any of these properties when using the native VOL. Currently, this seems like it would pass NONE for the environment, whereas we shouldn't be passing anything. It would also be good to avoid the generator expressions below that are trying to fix things up when using the native VOL.

tools/test/h5copy/CMakeTests.cmake
```diff
@@ -492,124 +728,146 @@
...
-  ADD_SIMPLE_TEST (h5copy_help2 0 --help)
+  list(LENGTH HDF5_EXTERNAL_VOL_TARGETS num_ext_vols)
 
-  # "Test copying various forms of datasets"
-  ADD_H5_TEST (simple 0 ${HDF_FILE1}.h5 -v -s simple -d simple)
-  ADD_H5_TEST (chunk 0 ${HDF_FILE1}.h5 -v -s chunk -d chunk)
-  ADD_H5_TEST (compact 0 ${HDF_FILE1}.h5 -v -s compact -d compact)
-  ADD_H5_TEST (compound 0 ${HDF_FILE1}.h5 -v -s compound -d compound)
+  # Add a test for the native connector and each external VOL connector
+  foreach (vol_idx RANGE 0 ${num_ext_vols})
```
jhendersonHDF: I think this is handy for preventing duplication of the `ADD_` macros below, but it complicates the logic a bit when the native VOL is being used and has other issues due to VOL testing really being separate from regular testing. For example, tests that use VOL plugins should currently only be run when shared tools are built (or, the flipside of that, avoided when static tools are built), whereas the regular tools tests using the native VOL can always be run. I think it would go a long way if we were able to collapse the vast majority of the different `ADD_` macros and ideally have one for regular testing and one for VOL testing, though it's not clear if that would be more complicated than what we currently have.

In the worst case, one alternative might be to move the `foreach (VOL ...)` logic into each of the macros so that we don't duplicate the calls here.

mattjala: If the primary motivation is to separate VOL testing from regular testing, does moving the per-VOL loop to inside the `ADD_` macros make that any simpler? It seems like any conditions we want to enforce as requisite for VOL testing (e.g. shared libraries) are simpler to set up when all the `ADD_` macros fall within a single loop.

The other case - where VOL tests are managed by an invocation of a counterpart `ADD_VOL_` call for each `ADD_` - seems cumbersome and likely to lead to newly added tests being accidentally left out of some test suites. The majority of tool tests can be run against most VOL connectors, so the default case being to run the test for native and non-native connectors makes sense.

I agree that it would be good to condense the number of distinct `ADD_` macros, especially because they are duplicated across the different tools

jhendersonHDF: I would say the primary motivation is trying to separate VOL testing from regular testing while also trying to prevent the need for duplicating calls to the `ADD_` macros, since, as you mention, it opens the possibility for someone to forget to add new calls for both the native VOL and external VOLs. Moving the foreach logic into each macro allows us to keep only one set of calls to the macros and allows us to easily separate the VOL-specific test setup logic from the logic for the native VOL; mixing logic between "regular" testing and external VOL testing is always going to be messy. However, it will likely make each of those macros substantially larger, which is why I think the ideal case is condensing the majority of the existing macros.

In many cases, the macro variants aren't substantially different from each other, so I think this is work that needs to happen regardless. The current situation of having many variants of tools test macros doesn't lend itself well to adding in VOL testing without either duplication or hacky workarounds for the native VOL.

> The majority of tool tests can be run against most VOL connectors, so the default case being to run the test for native and non-native connectors makes sense.

As I recall, several tools tests had to be skipped or left out for the majority of VOL connectors the last time this was tested, so I'm not sure this is necessarily the case. And to me this is all the more reason to keep VOL testing as separate from regular testing as possible.

mattjala:
- Merged several variants of the `ADD_H5_TEST` macro together
- Moved the VOL iteration logic inside the `ADD_` macros
- The native VOL tests no longer have the `HDF5_VOL_` prefix
- Native VOL tests no longer modify the environment/working directory

tools/test/h5copy/CMakeTests.cmake
```diff
@@ -38,16 +46,49 @@
       h5copy_help2.ddl
   )
 
+  # Testfile directory setup
   file (MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/testfiles")
 
-   
+  # Generate testfiles for VOL connector(s) through script, if enabled
+  if (HDF5_BUILD_GENERATORS)
```
jhendersonHDF: Since the `HDF5_BUILD_GENERATORS` option has been removed this clause can be removed

brtnfld:
@ellipsis-dev, review PR.

mattjala:
This is waiting on #5650

mattjala:
@ellipsis-dev, review PR.

byrnHDF (approved):

lrknox (approved):

PR #5553, opened 2025-05-28
Add static only workflows to CI

byrnHDF:
Depends on #5554


lrknox (approved):

hyoklee (approved):

lrknox:
Fortran change is one line in CMake code only - overriding requirement for Fortran code owner review.

PR #5554, opened 2025-05-28
Fix symbol definition conflict in tools

mattjala:
Linking h5gentest to the tools library creates a conflict over multiple definitions of `H5TOOLS_MALLOCSIZE` in certain builds.  This PR removes the link to the tools library and replaces use of a routine in h5dumpgentest that depended on it.

### DIFF
--- a/tools/test/CMakeLists.txt
+++ b/tools/test/CMakeLists.txt
@@ -50,14 +50,12 @@ if (HDF5_BUILD_GENERATORS)
         TARGET_C_PROPERTIES(${tool}gentest STATIC)
         target_link_libraries (${tool}gentest PRIVATE
           ${HDF5_LIB_TARGET}
-          ${HDF5_TOOLS_LIB_TARGET}
           ${HDF5_REQUIRED_LIBRARIES} # For math lib
         )
     else ()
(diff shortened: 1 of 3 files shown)


byrnHDF (dismissed):

jhendersonHDF:
I don't think this is quite the way to go about this. The `h5tools_can_encode()` function is good to use for maintenance reasons. The reason you're encountering the link error for `H5TOOLS_MALLOCSIZE` is because it's now defined in both `tools/lib/h5tools_utils.c` and `tools/test/h5diff/h5diffgentest.c` which is of course a conflict since it's `extern` in `tools/lib/h5tools_utils.h` so that all the tools can have access to it if needed. It should be removed from `tools/test/h5diff/h5diffgentest.c`, which should just be using the common definition in `tools/lib/h5tools_utils.c`.

jhendersonHDF:
On second thought and a bit of discussion, it may make more sense for the `h5tools_can_encode()` function to just be a macro in, say H5Zpublic.h. In either case, I think the `H5TOOLS_MALLOCSIZE` situation still needs fixed.

byrnHDF (approved):

lrknox (approved):

PR #5559, opened 2025-05-30
Add h5ls test files to gentest

mattjala:
h5ls currently has no script to generate the test files it uses on the fly. This adds support for h5ls test file generation to the gentest script.

### DIFF
--- a/tools/test/h5gentest.c
+++ b/tools/test/h5gentest.c
@@ -428,7 +428,7 @@ gen_h5ls_files(void)
     return nerrors;
 }
 
- /*-------------------------------------------------------------------------
+/*-------------------------------------------------------------------------
  * Function: usage
  *
  * Purpose: Prints a usage message on stdout stream and then returns.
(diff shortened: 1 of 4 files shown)


hyoklee (approved):

tools/test/h5ls/h5lsgentest.c
```diff
@@ -0,0 +1,104 @@
...
+
+    /* Initialize data for dataset 2 */
+    for (i = 0; i < 30; i++) {
+        for (j = 0; j < 10; j++) {
+            data2[i][j] = (double)j;
+        }
+    }
+
+    /* Create a new file using default properties */
+    file_id = H5Fcreate(TDSET_FILENAME, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
```
bmribler, line 51: Do we have a standard on error checking for test file generators?  I looked at the existing ones, it's a mix.  It's not in https://github.com/HDFGroup/hdf5/wiki/Getting-Started-with-HDF5-Development, BTW.

mattjala: No. Some test generators return an error code, some print an error message and try to continue, some use asserts, and some directly exit on failure. I've been keeping the error handling as-is and just taking advantage of it where it exists to track nerrors in h5gentest. It would be nice to make this consistent across generators, but that should probably be considered separately.

bmribler (approved):

tools/test/h5gentest.c
```diff
@@ -476,14 +521,16 @@ main(int argc, char *argv[])
                 case 'p':
                     run_h5repart = true;
                     break;
+                case 'l':
+                    run_h5ls = true;
```
jhendersonHDF, line 525: @mattjala This added case statement doesn't have a `break`, causing it to fall into the `continue` statement. Currently, that doesn't cause bad behavior but a `break` should be added.

PR #5548, opened 2025-05-23
Re-write ROS3 VFD S3 backend to use AWS' C S3 library

jhendersonHDF:
Replaces the ROS3 VFD's libcurl-based S3 backend with a new backend which uses the aws-c-s3 library

Adds support for sourcing credentials from standard AWS configuration files, environment variables, etc. with a fallback to anonymous credentials when credentials can't be found

Adds support for sourcing AWS region from standard AWS configuration files and environment variables

Adds support for new environment variables HDF5_ROS3_VFD_DEBUG, HDF5_ROS3_VFD_LOG_LEVEL and HDF5_ROS3_VFD_LOG_FILE for debugging purposes

Replaces function to load testing credentials in the ros3.c and s3comms.c tests with new function that delegates to the aws-c-s3 library for better compatibility with standard AWS configuration files and environment variables

Adds support for s3:// object URIs

Adds HTTP User-Agent header in form of
'libhdf5/<hdf5_version> (vfd:ros3) libaws-c-s3'

Fixes #4493
Fixes #4600
Fixes #5276

PR #5560, opened 2025-06-01
Add documentation and s3proxy testing for ROS3 VFD

### LINKED ISSUE #5109: Expand ros3 VFD testing
The ros3 VFD testing is very minimal (mainly just the ros3 and s3comms unit tests). We should expand the ros3 VFD testing to exercise more HDF5 API calls.

An easy way to do this might be to upload the tools test files to S3 and run h5dump and/or h5stat over them using the ros3 VFD.

jhendersonHDF:
Adds testing of ROS3 VFD with s3proxy and docker

Adds workflow to build ROS3 VFD and optionally build aws-c-s3 library from source or use package
managers where possible

Adds new H5Pset_fapl_ros3_endpoint()/H5Pget_fapl_ros3_endpoint() API functions to set/get an
alternative endpoint URL to use when opening files with the ROS3 VFD

Cleans up warnings in tools and tests related to ROS3 VFD structure size

Fixes #5109


jhendersonHDF:
This PR is the followup to #5548 and replaces and largely copies the remaining features needed from #5397, but with several changes and additions made to work with the aws-c-s3 library. The ROS3 is built on the 3 major platforms for PRs, but only tested on Ubuntu as testing currently relies on docker and the aws cli, which are preinstalled on the Ubuntu runners. It should be fairly straightforward to get testing running on Windows and Mac as well, but this is a good start. For PRs, aws-c-s3 is installed from vcpkg on Windows and homebrew on Mac, but Ubuntu doesn't have a package yet, so it gets built from source. The VFD is also built and tested on Ubuntu with a source build of aws-c-s3 in the VFD daily build action.

It's fairly large due to all the changes that had to be made to work well with #5397, but if need be I can try to break it down into more pieces. The large line change count is just due to the inclusion of the large text file for testing, which I hope we can remove in the future.

release_docs/INSTALL_S3.txt
jhendersonHDF, line 1: @ajelenak @lrknox @byrnHDF Can you review and let me know if there's anything else that would be good to know here?

src/H5FDros3_s3comms.c
```diff
@@ -669,6 +681,14 @@ H5FD__s3comms_s3r_open(const char *url, const H5FD_ros3_fapl_t *fa, const char *
                     aws_error_str(aws_last_error()));
     cond_var_init = true;
 
+    /* Check if path-style requests should be forced */
+    pathstyle_env = getenv(HDF5_ROS3_VFD_FORCE_PATH_STYLE);
```
jhendersonHDF, line 683: This is needed for s3proxy testing, as I couldn't get virtual-hosted-style requests to work correctly with it, but path-style requests work just fine.

src/H5Fmodule.h
```diff
@@ -1041,6 +1041,50 @@
  * additional variable settings associated with the Split driver, there is no H5Pget_fapl_split
  * function.
  *
+ * \subsubsection subsubsec_file_alternate_drivers_ros3 The ROS3 Driver
+ * The ROS3 (read-only S3) driver is used to enable read-only access to HDF5 files which are stored
```
jhendersonHDF, line 1045: I added a user's guide entry for the VFD, but as the entries for the other VFDs were brief I didn't add too many details here. Most information about the VFD is in its header file currently. But I think we should revise these entries into separate pages or something similar so they can serve as the main location for information about a particular VFD. That or maybe free-form doxygen like this should be put into each VFD's header file.

test/testfiles/t8.shakespeare.txt
jhendersonHDF, line 1: For now these text files have been included as the ros3 and s3comms tests use them. But the ROS3 VFD can't actually read from text files normally, it can only do so indirectly through `H5FDopen()` and `H5FDread()`, which is what those tests are testing. We should revise those tests in the future to do more meaningful testing.

tools/libtest/h5tools_test_utils.c
jhendersonHDF, line 1: The changes in this file mostly just cleaned up stack size warnings from the `H5FD_ros3_fapl_ext_t` structure being too large.

tools/src/h5dump/h5dump.c
```diff
@@ -34,30 +34,11 @@ static h5tools_vfd_info_t vfd_info_g = {0};
...
-    {
-        1,     /* Structure Version */
-        false, /* Authenticate?     */
-        "",    /* AWS Region        */
-        "",    /* Access Key ID     */
-        "",    /* Secret Access Key */
-    },
-    "", /* Session/security token */
-};
+static H5FD_ros3_fapl_ext_t *ros3_fa_g = NULL;
```
jhendersonHDF, line 37: Similarly, I revised `h5dump`, `h5ls` and `h5stat` to allocate these structures to fix stack size warnings.

tools/src/h5dump/h5dump.c
```diff
@@ -891,18 +872,6 @@ parse_command_line(int argc, const char *const *argv)
                 vfd_info_g.type   = VFD_BY_NAME;
                 vfd_info_g.u.name = H5_optarg;
                 use_custom_vfd_g  = true;
-
-#ifdef H5_HAVE_ROS3_VFD
-                if (0 == strcmp(vfd_info_g.u.name, drivernames[ROS3_VFD_IDX]))
```
jhendersonHDF, line 896: Logic like this was moved down in `h5dump`, `h5ls` and `h5stat` to handle overlap between options like `-f ros3` and `--vfd-name ros3` where the former would set the pointer to the VFD info and the latter wouldn't.

tools/src/h5ls/h5ls.c
```diff
@@ -2802,7 +2808,7 @@ main(int argc, char *argv[])
             data_g = true;
         }
         else if (!strcmp(argv[argno], "--enable-error-stack")) {
-            enable_error_stack = 1;
+            enable_error_stack = 2;
```
jhendersonHDF, line 2811: @byrnHDF Since `h5ls` doesn't accept an argument for `--enable-error-stack`, I just changed the default here to 2 so that file open errors aren't hidden since they're often the most interesting errors. At this point, I kind of think 2 should be the default everywhere so that `--enable-error-stack` works and you don't have to specify `--enable-error-stack=2`.

.github/workflows/vfd-ros3.yml
```diff
@@ -0,0 +1,448 @@
...
+    inputs:
+      build_mode:
+        description: "CMake build type (CMAKE_BUILD_TYPE)"
+        required: true
+        type: string
+      aws_c_s3_build_type:
+        description: "Install aws-c-s3 from a package manager ('package') or build from source ('source')"
+        required: true
+        type: string
+      aws_c_s3_tag:
```
byrnHDF, line 14: should add a default value to optional input

config/ConfigureChecks.cmake
```diff
@@ -622,6 +622,36 @@ if (HDF5_ENABLE_ROS3_VFD)
...
+      else ()
+        set (HDF5_ENABLE_ROS3_VFD_DOCKER_PROXY OFF CACHE BOOL "Use docker for ROS3 VFD S3proxy testing" FORCE)
+        message (WARNING "Docker is not installed.")
+      endif ()
+
+      # check if aws-cli is available
+      find_program (AWS_CLI_EXECUTABLE aws)
+      if (NOT AWS_CLI_EXECUTABLE)
+        set (HDF5_ENABLE_ROS3_VFD_DOCKER_PROXY OFF CACHE BOOL "Use docker for ROS3 VFD S3proxy testing" FORCE)
+        message (WARNING "AWS cli is required for ROS3 VFD S3proxy testing, but could not be found.")
```
byrnHDF: Should this be a FATAL_ERROR if it is required?

release_docs/RELEASE.txt
```diff
@@ -573,6 +587,18 @@ New Features
 
     Tools:
     ------
+    - Added AWS endpoint command option to allow specifying an alternate endpoint
```
byrnHDF, line 625: Should this mention the ENV variables?

.github/workflows/vfd-ros3.yml
jhendersonHDF: I'm actually checking for empty input and letting the checkout action use the default branch if so. Probably a bit more future proof than specifying a default (though let's hope the name of the default branch doesn't change in GH again).

config/ConfigureChecks.cmake
jhendersonHDF: I just did the same as the existing logic for the docker option which just turns the option back off. We could make both fatal errors instead though if we want. I just added this because I noticed the test logic expects the aws cli commands to be available as well, but wasn't checking for them. For me it caused a failure with no obvious reason why.

release_docs/RELEASE.txt
jhendersonHDF: Sure, I can add a line for that.

src/H5FDros3_s3comms.c
```diff
@@ -1850,15 +1921,24 @@ H5FD__s3comms_format_http_request_message(const H5FD__s3comms_aws_params_t *aws_
     assert(object_key);
     assert(message_out);
 
+    if (aws_params->force_path_style)
+        use_virtual_style = false;
+
+    /*
+     * If an alternate endpoint URL was specified, use a path-style
```
byrnHDF, line 1926: I've noticed this comment in several places - why is this a problem, because I thought the aws library would deal with this? Is it a hdf design decision?

src/H5Fmodule.h
byrnHDF: The H5?module.h files were convenient (initially) for User Guide entries as they wouldn't be confused/interfere with reference Manuals entries and some UG entries were extensive. Anything should be able to be reworked/relocated as necessary.

config/ConfigureChecks.cmake
byrnHDF: The original may have been wrong, but the more important decision is what is correct? If the user asked for it and it can't be found, will the build/test still give the intended result or should we stop and force the user to fix it. Given the time for some tests to complete (especially on Windows) I would prefer to know immediately and not waste time.

tools/src/h5dump/h5dump.c
byrnHDF: Are they really different options?

byrnHDF (dismissed):

src/H5FDros3_s3comms.c
jhendersonHDF: This is about what a server will accept, not what the aws library can do since it doesnt care too much about that and cant necessarily tell ahead of time what's supported. The only request format I could get to work between the library and s3proxy was the path-style format, which is why this is needed. Virtual-hosted-style requests seem somewhat specific to s3 (though I haven't checked other cloud providers) and since the initial use case for the alternate endpoint is for testing with s3proxy, this seemed reasonable to do. https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html eventually support for path-style requests will be removed from S3, but that's a future problem. I expect we'll still have to maintain support for path-style requests into the future for compatibility.

tools/src/h5dump/h5dump.c
jhendersonHDF: In terms of the tools code they're pretty much all exactly the same now among tools, with the "old" options being replaced by the `--vfd-name` option. The latter was kind of meant for vfd plug-ins, but it evolved to cover builtin library vfds as well, so they really should work the same at this point.

release_docs/INSTALL_S3.txt
```diff
@@ -0,0 +1,183 @@
...
+========================================================================
+
+As of the HDF5 2.0.0 release (or HDF5 commit 75fff3ed0663d3617e3c5cc3c57c1f0c83b38be4),
+HDF5's ROS3 VFD requires the aws-c-s3 library (https://github.com/awslabs/aws-c-s3)
+in order to be built. This library is distributed by several package
+managers, usually as 'aws-c-s3', but consult your particular package
+manager for the name of this package if it's available. HDF5 installs
+this library in the following ways in CI testing:
+
+Linux (Ubuntu):
```
byrnHDF, line 33: On fedora 42 I had to install the "devel" package as well

src/H5FDros3_s3comms.c
byrnHDF, line 1: H5FDros3_s3comms.c:2202:74: warning: unused parameter ‘aws_params’ [-Wunused-parameter]
 2202 | H5FD__s3comms_format_user_agent_header(const H5FD__s3comms_aws_params_t *aws_params,

release_docs/INSTALL_S3.txt
jhendersonHDF: Actually I need to update this line since Ubuntu didn't have a package yet. I don't know how much we want to get into the details for each Linux distribution since it could vary widely.

src/H5FDros3_s3comms.c
jhendersonHDF: Will fix. I turned on the `HDF5_ENABLE_WARNINGS_AS_ERRORS=ON` to try and catch new warnings in the ros3 code, but there appear to be several warning types it doesnt treat as errors, which we should revisit.

release_docs/INSTALL_S3.txt
byrnHDF: At least for linux there always seems to be a devel package and it is needed for CMake to configure.

src/H5FDros3.c
```diff
@@ -707,6 +711,209 @@ H5Pset_fapl_ros3_token(hid_t fapl_id, const char *token)
...
+H5FD__ros3_str_endpoint_copy(const char H5_ATTR_UNUSED *name, size_t H5_ATTR_UNUSED size, void *_value)
+{
+    char **value     = (char **)_value;
+    herr_t ret_value = SUCCEED;
+
+    FUNC_ENTER_PACKAGE
+
+    if (*value)
+        if (NULL == (*value = strdup(*value)))
+            HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "can't copy string property token");
```
lrknox: Should the failure message refer to endpoint instead of property token?

tools/lib/h5tools_utils.c
```diff
@@ -1079,19 +1079,19 @@ h5tools_parse_ros3_fapl_tuple(const char *tuple_str, int delim, H5FD_ros3_fapl_e
...
  *         secret_id,
  *         secret_key,
+ *         session_token (optional)
  *     }
- *     If all three strings are empty (""), the default fapl will be default.
- *     Both aws_region and secret_id values must be both empty or both
- *         populated. If
- *     Only secret_key is allowed to be empty (the empty string, "").
+ *     If all three (or four) strings are empty (""), the default fapl will be
+ *     default. Both aws_region and secret_id values must be both empty or both
```
lrknox: "Both" at the beginning of the sentence seems redundant;  not terribly important, though.

lrknox (dismissed):

src/H5FDros3.c
jhendersonHDF: Yes I think this implementation was copy pasted from the previous implementation for tokens. Will tidy up the error messages.

release_docs/INSTALL_S3.txt
```diff
@@ -0,0 +1,184 @@
...
+Section IV:  Building HDF5 with the ROS3 VFD enabled
+Section V:   Testing the ROS3 VFD
+
+************************************************************************
+
+========================================================================
+I. Preconditions
+========================================================================
+
+- CMake 3.18 or newer is required to build HDF5
```
byrnHDF: Maybe refer to the regular INSTALL file for details and preconditions, this doc should be in addition to that one?

byrnHDF (dismissed):

release_docs/INSTALL_S3.txt
jhendersonHDF: Yes, something like that. I didn't want to essentially reproduce that whole document here.

byrnHDF (approved):

lrknox (approved):

PR #5564, opened 2025-06-03
Move h5repacktst file generation to h5repackgentest

mattjala:
- For consistency with other tools tests, move h5repacktst file generation into h5repackgentest

This is motivated primarily by the need to be able to generate the h5repacktst files for h5repack VOL tests.

Depends on #5622


tools/test/h5repack/h5repackgentest.c
```diff
@@ -295,4 +356,5414 @@ generate_f32le(bool external)
...
+/*-------------------------------------------------------------------------
+ * Function: make_h5repack_testfiles
+ *
+ * Purpose:  make a test file with all types of HDF5 objects,
+ *           datatypes and filters
+ *
+ *-------------------------------------------------------------------------
+ */
+int
+make_h5repack_testfiles(void)
```
mattjala, line 364: These ~5,000 lines are just copied from `h5repacktst.c` and don't need to be closely inspected

tools/test/h5repack/h5repackgentest.c
```diff
@@ -13,8 +13,13 @@
 #include "hdf5.h"
 #include "H5private.h"
 #include "h5repackgentest.h"
+#include "h5tools.h"
+#include "h5tools_utils.h"
+#include "h5test.h"
 
 #define MAX_NAME_SIZE  256
+#define PAGE_SIZE_DEFAULT 4096
```
mattjala, line 21: h5repacktst originally used `FS_PAGESIZE_DEF` from `h5repack.h`. To avoid using h5repack internal symbols, we define a different symbol here with the same value.

tools/test/h5repack/h5repackgentest.h
```diff
@@ -20,5 +20,87 @@ int generate_int32le_2d(bool external);
 int generate_int32le_3d(bool external);
 int generate_uint8be(bool external);
 int generate_f32le(bool external);
+int make_h5repack_testfiles(void);
+int verify_userblock(const char *filename);
+
+/* fill value test */
+#define H5REPACK_FNAME0    "h5repack_fill.h5"
```
mattjala, line 27: These symbols are shared between `h5repackgentest.c` and `h5repacktst.c`

mattjala:
Marked as draft until I fix build issues on Windows

bmribler (dismissed):

vchoi-hdfgroup (dismissed):

tools/test/h5repack/CMakeLists.txt
```diff
@@ -27,19 +27,19 @@ set (REPACK_COMMON_SOURCES
     ${HDF5_TOOLS_SRC_H5REPACK_SOURCE_DIR}/h5repack_verify.c
     ${HDF5_TOOLS_SRC_H5REPACK_SOURCE_DIR}/h5repack.c
 )
+# Add h5repack test executable
 add_executable (h5repacktest ${REPACK_COMMON_SOURCES} ${HDF5_TOOLS_TEST_H5REPACK_SOURCE_DIR}/h5repacktst.c)
 target_include_directories (h5repacktest
     PRIVATE "${HDF5_TOOLS_SRC_H5REPACK_SOURCE_DIR};${HDF5_TOOLS_ROOT_DIR}/lib;${HDF5_SRC_INCLUDE_DIRS};${HDF5_TEST_SRC_DIR};${HDF5_SRC_BINARY_DIR};$<$<BOOL:${HDF5_ENABLE_PARALLEL}>:${MPI_C_INCLUDE_DIRS}>"
 )
-if (BUILD_STATIC_LIBS)
```
jhendersonHDF, line 34: I think this is an issue where we need to determine the full purpose of `HDF5_BUILD_STATIC_TOOLS` and align between it and `BUILD_STATIC_LIBS`. Since this is a regular test program, it should probably be using `BUILD_STATIC_LIBS` as before rather than `HDF5_BUILD_STATIC_TOOLS`. But, when it comes time to link against the tools library we need to check `HDF5_BUILD_STATIC_TOOLS` to link against the correct one (though we should probably solve this higher up and just give lower level CMake code a single target to link against unconditionally).

tools/test/h5repack/CMakeLists.txt
```diff
@@ -27,19 +27,19 @@ set (REPACK_COMMON_SOURCES
...
+if (HDF5_BUILD_STATIC_TOOLS)
   TARGET_C_PROPERTIES (h5repacktest STATIC)
-  target_link_libraries (h5repacktest PRIVATE ${HDF5_TOOLS_LIB_TARGET} ${HDF5_TEST_LIB_TARGET})
+  target_link_libraries (h5repacktest PRIVATE ${HDF5_TOOLS_LIB_TARGET} ${HDF5_TEST_LIB_TARGET} h5repackgentest)
```
jhendersonHDF, line 37: Since the tool gentest object libraries are currently only built when `HDF5_BUILD_GENERATORS` is `ON`, this linking step fails when it's `OFF`. I'm guessing the reason it passes in most of the workflows is because they use the cacheinit file which sets that `ON` by default. I would generally like to avoid the requirement of `HDF5_BUILD_GENERATORS=ON` in cases where we don't need it, but this PR seems like it will need that, even for regular non-VOL testing. Maybe we need to decide if the `HDF5_BUILD_GENERATORS` option is still useful or if we should just always build the generators? Or maybe we should always build the generator _libraries_ but only build the generator binaries when `HDF5_BUILD_GENERATORS=ON`.

jhendersonHDF (dismissed):
CMake logic needs re-worked when it comes to the `HDF5_BUILD_GENERATORS` option, as these changes will currently fail to link when it's `OFF`. CI isn't catching this, likely due to the cacheinit file.

mattjala:
It would be nice to remove h5repacktst entirely since it's superseded by the h5repack CMake tests, but the VFD tests for h5repack still use the script. h5repacktst can still be removed, but it will require some substantial modification to the h5repack tests not dissimilar to the changes needed to set up VOL testing. For now, I think we should leave the script in - reworking h5repack VFD testing should be considered separately.

Marking this as draft since it depends on #5622

brtnfld:
@ellipsis-dev, review the PR

tools/test/h5repack/h5repacktst.c
```diff
@@ -1724,18 +1589,6 @@ main(void)
...
     }
 
-    /*-------------------------------------------------------------------------
-     * clean temporary test files
-     *-------------------------------------------------------------------------
-     */
-    {
-        hid_t fapl;
-
-        fapl = h5_fileaccess();
```
jhendersonHDF, line 1734: These lines should be put back and adapted in some form since `H5REPACK_FILENAMES` doesn't exist here now. Otherwise, all the test files get left behind instead of being cleaned up.

jhendersonHDF (approved):

brtnfld (approved):

lrknox:
I see nothing here that will affect or be affected by Java on arm64-windows, so I think we can override the 2 failing test configurations and merge this.

Actually didn't require override.

PR #5569, opened 2025-06-05
Fix target namespace conflicts in CMake

mattjala:
`HDF_PACKAGE_NAMESPACE` is a CMake variable used to distinguish HDF cmake targets. It's set when the configuration in `cacheinit.cmake` is used, but is left undefined by default. 

This can cause conflicts between certain build targets when the namespace is expected to distinguish them. I ran into a conflict between the `szaec-static` target defined in `EXTERNAL_SZIP_LIBRARY ` and the one defined in `LIBAEC/CMakeLists.txt:179`, and [users have encountered a similar problem for zlib](https://forum.hdfgroup.org/t/hdf5-cmake-szlib-path/12337/8).


byrnHDF:
Yes the namespace was left empty on purpose. It is a user option.

byrnHDF (changes requested):
We need to figure out a solution.

PR #5587, opened 2025-06-09
Correct name of CMake option to HDF5_ENABLE_DEPRECATED_SYMBOLS

### LINKED ISSUE #5591: Using HDF5_ENABLE_DEPRECATED_SYMBOLS to disable deprecated functions fails
Setting HDF5_ENABLE_DEPRECATED_SYMBOLS = OFF caused compile errors because examples used incorrect functions.

byrnHDF:
Fixes #5591


lrknox (changes requested):
The only gcc DBG v2.0.0 default API workflow in main-cmake-spc.yml that fails due to use of a deprecated function is the v2.0.0 default API.  The others fail because src/H5version.h : 54 doesn't allow both H5_USE_*_API and H5_NO_DEPRECATED_SYMBOLS.  v2.0.0 will also fail as soon as there is a newer version.  We'll need to add the HDF5_ENABLE_DEPRECATED_SYMBOLS option to a workflow that doesn't also specify an older version.

brtnfld:
We should implement a YML function that checks when an unknown CMake option is used in our CI testing, to prevent this issue from happening in the future.

.github/workflows/main-cmake-spc.yml
```diff
@@ -333,7 +328,7 @@ jobs:
             -DZLIB_USE_LOCALCONTENT=OFF \
             -DHDF5_ENABLE_MIRROR_VFD:BOOL=ON \
             -DHDF5_ENABLE_DIRECT_VFD:BOOL=ON \
-            -DH5_NO_DEPRECATED_SYMBOLS:BOOL=OFF \
+            -DHDF5_ENABLE_DEPRECATED_SYMBOLS:BOOL=OFF \
```
lrknox, line 331: Leaving this as an exercise when version 3.0 and -DHDF5_DEFAULT_API_VERSION:STRING=v2 come along?

byrnHDF: When that happens then there will be a new latest version and previous one will need to remove the option.

brtnfld (dismissed):

lrknox (dismissed):

lrknox (approved):

.github/workflows/main-cmake-spc.yml
fortnern, line 331: Is this changing the default to disable deprecated symbols?

byrnHDF: It is using the correct name of the option to disable deprecated symbols.

fortnern: Ok but it looks like it's reversing the setting for this workflow.

fortnern: H5_NO_DEPRECATED_SYMBOLS has the opposite meaning to HDF5_ENABLE_DEPRECATED_SYMBOLS. You can see this where the former is set in /CMakeLists.txt

fortnern: I suppose it only affects the workflow, not the code itself, but it doesn't seem like this is what you want to do

byrnHDF: The intention of the job was to build develop branch with no deprecated functions. The default is ON, so yes this is changing the option from the default to OFF so as to disable using deprecated functions.

fortran/src/CMakeLists.txt
```diff
@@ -3,9 +3,9 @@ project (HDF5_F90_SRC C Fortran)
 
 #-----------------------------------------------------------------------------
 if (H5_NO_DEPRECATED_SYMBOLS)
-  set (CMAKE_NO_DEPRECATED_SYMBOLS 0)
-else ()
   set (CMAKE_NO_DEPRECATED_SYMBOLS 1)
+else ()
```
fortnern, line 7: Why does this reverse the option if they're both "no deprecated symbols"?

.github/workflows/main-cmake-spc.yml
fortnern: Why didn't it disable deprecated functions previously?

byrnHDF: Because H5_NO_DEPRECATED_SYMBOLS is not an option name

fortnern: Was the intent always to disable deprecated symbols for this build? If so why wasn't it set to BOOL=ON before?

byrnHDF: Yes because the develop is the latest version and no deprecate only works for it - other API versions disallow using no deprecate

fortran/src/CMakeLists.txt
byrnHDF: Fortran code needs a different define in the *,inc file and the values for for creating the define for the .inc file was incorrect.

.github/workflows/main-cmake-spc.yml
fortnern: You're changing the functionality of this workflow here. If you're going to mix this functionality change in with the name change, at least call it out in the description. Ideally it should be a separate PR.

byrnHDF: No I am not. The wrong name was used! Also the option should not have been in the other jobs.

byrnHDF: It never worked as intended,

fortnern: It may have used the wrong name but by trying to set "no deprecated symbols" to OFF it was trying to do the opposite of what it does now. Even if it didn't work before, the intent in the code was clear, and it has been reversed. If that's what it needs to be that's fine, but it's going beyond a name change.

byrnHDF: No. Had the correct name been used in the first place it would have errored because you cannot use API version and NO_DEPRECATE. Therefore the option would have been removed from the other jobs and left only on the one job that could use it. (In fact that is actually what I did!)

fortnern: But it didn't use no_deprecate. It was set to off.

fortnern: Are you saying that all the jobs were intended to disable deprecated symbols, but they mixed up the variable name and reversed the setting simultaneously?

fortnern (approved):

brtnfld (approved):

PR #5590, opened 2025-06-10
Check if namespace has been defined before creating alias

### LINKED ISSUE #5584: Szip and Zlib builds fail when no HDF package namespace is specified
**Describe the bug**
The library will fail to build with szip and/or zlib due to conflicting CMake target names when HDF5_PACKAGE_NAMESPACE is not defined at configure time.

**Additional context**
See description in #5569

byrnHDF:
Fixes #5584


config/cmake/HDFLibMacros.cmake
```diff
@@ -43,7 +43,9 @@ macro (EXTERNAL_ZLIB_LIBRARY compress_type)
 
   FetchContent_MakeAvailable(HDF5_ZLIB)
 
-  add_library(${HDF_PACKAGE_NAMESPACE}zlib-static ALIAS zlib-static)
+  if (HDF_PACKAGE_NAMESPACE)
+    add_library(${HDF_PACKAGE_NAMESPACE}zlib-static ALIAS zlib-static)
```
mattjala, line 47: Are these aliases necessary for the library's setup when filters are enabled? In other words, if the user has no namespace set and tries to build with zlib/szip, will it work?

lrknox (approved):

config/cmake/HDFLibMacros.cmake
mattjala: I tested the build with szip enabled and without a namespace, and it works.

mattjala (approved):

PR #5597, opened 2025-06-12
Move tool testfiles to shared folder

### LINKED ISSUE #5588: Tool test files should not be tool specific
The test files for tools are currently separated across distinct folders. This complicates the build/test process for files that multiple tools need to use, and increases risk of overlap/duplication of work across multiple tools. Ideally, test files would be generated without reference to specific test cases they are used in. Reorganizing the test files would be a first step in this direction.

Tool test reference files should inhabit a single directory shared between tools instead of tool-specific directories. The same should be true for the copies/generated counterparts during test execution.

mattjala:
Move HDF5 files used by tools tests from individual `tools/test/<tool>/testfiles` folders to a shared `tools/test/testfiles`. This removes several cases where the same reference file existed in different folders. 

Resolves #5588


tools/test/h5repack/CMakeTests.cmake
```diff
@@ -19,29 +19,36 @@
   # --------------------------------------------------------------------
   # Copy all the HDF5 files from the source directory into the test directory
   # --------------------------------------------------------------------
-  set (LIST_REPACK_TEST_FILES
```
mattjala, line 22: This list mixed reference `.dat` files and HDF5 files, so I split it to distinguish the types of files

byrnHDF (approved):

lrknox (approved):

PR #5598, opened 2025-06-13
Resolve spurious linkchecker timeouts

### LINKED ISSUE #5439: Spurious linkchecker failures in CI
The current linkchecker often spuriously fails while checking valid links. We should use the github actions linkchecker (https://github.com/marketplace/actions/check-links-with-linkcheck) instead of the one installed through pip.

mattjala:
- Remove unneeded packages from linkchecker installation

The texlive packages constituted over a gigabyte of data to install, but seem to have no impact on the linkchecker workflow. Removing this drops the installation time from ~9 minutes to ~30 seconds. 

- Reduce number of linkchecker threads 
 
The [current linkchecker](https://github.com/linkchecker/linkchecker) often spuriously fails on valid links due to timeouts (see [here](https://github.com/HDFGroup/hdf5/actions/runs/15586347547/job/43893439400), [here](https://github.com/HDFGroup/hdf5/actions/runs/15637693352/job/44057051203), and [here](https://github.com/HDFGroup/hdf5/actions/runs/15353948295/job/43208633948) as examples).

This seems to be due to the target websites rejecting the connection, potentially due to the linkchecker using multiple threads to send many requests to the same domain.

This PR reduces the number of threads the linkchecker uses from the default (10) to 2. This does not substantially impact the time the linkchecker takes to run.

Resolves #5439

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -26,8 +26,6 @@ jobs:
           sudo apt install -y graphviz
           sudo apt install -y --no-install-recommends doxygen
           sudo apt install -y cmake
-          sudo apt install -y texlive-full
-          sudo apt install -y texlive-latex-extra
           pip3 install linkchecker
 
       - name: build
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -38,6 +38,13 @@ jobs:
       - name: check
         run: |
           cd build/hdf5lib_docs/html
-          linkchecker --no-warnings --ignore-url=/doxygen.css --ignore-url=gnu.org --ignore-url=en.wikipedia.org --ignore-url=help.hdfgroup.org --ignore-url=libpng.org --ignore-url=my.cdash.org --ignore-url=www.oreilly.com --ignore-url=preshing.com --ignore-url=semver.org --ignore-url=sourceforge.net --ignore-url=www.youtube.com --ignore-url=youtu.be --check-extern --threads=2 ./index.html
+          linkchecker --no-warnings --ignore-url=eigen.tuxfamily.org --ignore-url=/doxygen.css \
+          --ignore-url=gnu.org --ignore-url=en.wikipedia.org \
+          --ignore-url=help.hdfgroup.org --ignore-url=libpng.org \
+          --ignore-url=my.cdash.org --ignore-url=www.oreilly.com \
+          --ignore-url=preshing.com --ignore-url=semver.org \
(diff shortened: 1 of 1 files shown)


lrknox (dismissed):

byrnHDF (dismissed):

jhendersonHDF:
Have we verified that this actually resolves #5439 in a meaningful way? Usually, increasing a timeout value is just asking for this to continue happening in the future.

brtnfld:
At least for PR #5506, the addition of the timeout, along with `--no-robots`, did not provide any improvement.

mattjala:
> At least for PR #5506, the addition of the timeout, along with `--no-robots`, did not provide any improvement.

For that PR, it looks like the linkchecker is failing due to real problems with existing links - `https://www.mcs.anl.gov/~thakur/papers/romio-coll.ps.gz` and `http://www.mcs.anl.gov/~thakur/papers/mpio-high-perf.ps.gz` are 404 pages now. So it makes sense that altering the timeout didn't change anything

mattjala:
> Have we verified that this actually resolves #5439 in a meaningful way? Usually, increasing a timeout value is just asking for this to continue happening in the future.

The CI failures are sporadic, so I'm not sure how we can test it directly beyond making the change and seeing if the frequency of failures change. The timeout errors happen when trying to connect to websites that are, to all appearances, online and operational, so increasing the timeout value should have an effect.

brtnfld:
https://web.cels.anl.gov/~thakur/papers/romio-coll.pdf
https://web.cels.anl.gov/~thakur/papers/users-guide.pdf

mattjala:
> https://web.cels.anl.gov/~thakur/papers/romio-coll.pdf https://web.cels.anl.gov/~thakur/papers/users-guide.pdf

Links are updated in #5605

mattjala:
Changed the approach to reducing the number of threads instead of increasing timeout. Also removed unneeded package installs that seriously dropped the time needed for the install step (9 min -> 30 sec)

mattjala:
Added [tuxfamily](https://eigen.tuxfamily.org/index.php?title=Main_Page) to the list of ignored URLs since the site itself seems unreliable and is [causing linkchecker failures](https://github.com/mattjala/hdf5/actions/runs/15714785349/job/44281698403).

We could also simply remove the two links to that domain - they're both in About.dox, and are slightly redundant since we also link to the eigen Gitlab repo, which itself links to that domain.

lrknox (approved):

vchoi-hdfgroup (approved):

PR #5600, opened 2025-06-13
allow to modify or remove the postfix for debug builds

### LINKED ISSUE #4647: Remove the _debug prefixes for CMake debug builds (linux)
When CMake builds in debug mode, it creates library names appended with `_debug`; while this is common for Windows, it is not a common thing to do for Linux.

I propose leaving the same names for a release and debug builds for Linux/macOS.

MarDiehl:
closes #4647

### DIFF
--- a/config/HDFMacros.cmake
+++ b/config/HDFMacros.cmake
@@ -472,15 +472,6 @@ macro (HDF_DIR_PATHS package_prefix)
     set (CMAKE_PREFIX_PATH ${ADDITIONAL_CMAKE_PREFIX_PATH} ${CMAKE_PREFIX_PATH})
   endif ()
 
-  #set the default debug suffix for all library targets
-  if(NOT CMAKE_DEBUG_POSTFIX)
-    if (WIN32)
-      set (CMAKE_DEBUG_POSTFIX "_D")
-    else ()
-      set (CMAKE_DEBUG_POSTFIX "_debug")
-    endif ()
-  endif ()
-
   SET_HDF_BUILD_TYPE()
 
   SET_HDF_OUTPUT_DIRS(${package_prefix})
--- a/config/cmake/cacheinit.cmake
+++ b/config/cmake/cacheinit.cmake
@@ -40,6 +40,15 @@ set (HDF_TEST_EXPRESS "2" CACHE STRING "Control testing framework (0-3)" FORCE)
 
 set (HDF5_MINGW_STATIC_GCC_LIBS ON CACHE BOOL "Statically link libgcc/libstdc++" FORCE)
 
+#set the default debug suffix for all library targets
+if (NOT CMAKE_DEBUG_POSTFIX)
+  if (WIN32)
+    set (CMAKE_DEBUG_POSTFIX "_D")
+  else ()
+    set (CMAKE_DEBUG_POSTFIX "_debug")
+  endif ()
+endif ()
+
 set (HDF5_ALLOW_EXTERNAL_SUPPORT "TGZ" CACHE STRING "Allow External Library Building (NO GIT TGZ)" FORCE)
 set_property (CACHE HDF5_ALLOW_EXTERNAL_SUPPORT PROPERTY STRINGS NO GIT TGZ)
 
--- a/release_docs/RELEASE.txt
+++ b/release_docs/RELEASE.txt
@@ -742,6 +742,12 @@ Bug Fixes since HDF5-2.0.0 release
 
     Configuration
     -------------
+    - Remove default setting of CMAKE_DEBUG_POSTFIX
+
+      Move the default setting of CMAKE_DEBUG_POSTFIX to the cacheinit.cmake file
+      usually used by testing. If CMAKE_DEBUG_POSTFIX is not set with a -D option
+      then CMAKE_DEBUG_POSTFIX will be the default provided by CMake itself.
+    
     - The relative rpaths ($ORIGIN / @loader_path) are appended to the CMAKE_INSTALL_RPATH
 
       The RPATH settings were removed by a pull-request #5271, but the settings are needed


derobins (dismissed):

derobins:
Needs a release note, at the very least

byrnHDF:
I don't like this change as is - first the default can be overridden easiy;
set (CMAKE_DEBUG_POSTFIX "")

Second it is altering a setting that could be relied on due to it being a long standing setting.

byrnHDF (changes requested):
I think this is not the way to do this and why just for NOT win32.

derobins:
> I don't like this change as is - first the default can be overridden easiy;
> set (CMAKE_DEBUG_POSTFIX "")
> 
> Second it is altering a setting that could be relied on due to it being a long standing setting.
> 
> 

The default should be to do what other libraries do and not have a suffix. Anyone who cares about this can change it, but I think most people don't want to change linker commands when building a debug library. Windows is different since the debug library will link to the debug CRT.

derobins:
As for "long standing settings", 2.0 is the time to do the right thing instead of the traditional thing. Also, this isn't this only a long-standing thing on CMake?

config/HDFMacros.cmake
```diff
@@ -476,8 +476,6 @@ macro (HDF_DIR_PATHS package_prefix)
   if(NOT CMAKE_DEBUG_POSTFIX)
```
byrnHDF: I would add an option here "ENABLE_DEBUG_NAME_EXT" that is normally OFF. That is what this code was supposed to accomplish, just to make it easy and optional.
This way one does not need to remember that if I want an extension that I need to set it on Linux but skip WIN32. Easier to change one setting without regard to platform.

MarDiehl: What about making `CMAKE_DEBUG_POSTFIX` accesible via the commandline, i.e.`-DCMAKE_DEBUG_POSTFIX=`?

byrnHDF:
We should just move these settings (the whole if-else) to the cacheinit.cmake file as that is the intended use for that file.

config/cmake/cacheinit.cmake
```diff
@@ -40,6 +40,8 @@ set (HDF_TEST_EXPRESS "2" CACHE STRING "Control testing framework (0-3)" FORCE)
 
 set (HDF5_MINGW_STATIC_GCC_LIBS ON CACHE BOOL "Statically link libgcc/libstdc++" FORCE)
 
+set (CMAKE_DEBUG_POSTFIX "" CACHE STRING "Filename postfix for debug builds" FORCE)
```
byrnHDF: Just replace this with the whole thing:
  if(NOT CMAKE_DEBUG_POSTFIX)
    if (WIN32)
      set (CMAKE_DEBUG_POSTFIX "_D")
    else ()
      set (CMAKE_DEBUG_POSTFIX "_debug")
    endif ()
  endif ()

byrnHDF (dismissed):

jhendersonHDF:
@byrnHDF Does this handle setting of a postfix for developer builds of the library?

byrnHDF:
> @byrnHDF Does this handle setting of a postfix for developer builds of the library?

Yes this affects the $<CONFIG:Developer>>: generator instances

vchoi-hdfgroup (dismissed):

MarDiehl:
Tested locally, works as expected.
Force pushed to update commit message.

lrknox (changes requested):
Needs a release note - probably 2 or 3 lines, one to describe the problem,  one to describe the change, and one saying Fixes or Fixed GitHub issue #4647.  This should be added to the top of the  Bug Fixes / Configuration section at ~ line 743 of release_docs/RELEASE.txt file.

The cause of the netCDF test failures is known and unrelated to this pull request.

lrknox (dismissed):

vchoi-hdfgroup (dismissed):

release_docs/RELEASE.txt
```diff
@@ -742,6 +742,11 @@ Bug Fixes since HDF5-2.0.0 release
 
     Configuration
     -------------
+    - Allow to set postfix of target file names for debug builds
```
byrnHDF: Remove default setting of CMAKE_DEBUG_POSTFIX if not set.

Move the default setting of CMAKE_DEBUG_POSTFIX to the cacheinit,cmake file usually used by testing. If CMAKE_DEBUG_POSTFIX is not set with a -D option then CMAKE_DEBUG_POSTFIX will be the default provided by CMake itself,

byrnHDF (changes requested):
Change release note

release_docs/RELEASE.txt
MarDiehl: I find "Remove default setting of CMAKE_DEBUG_POSTFIX if not set." misleading. `cacheinit.cmake` is part of the release (at least the current), so when building HDF5 the old behavior of having postfixes for debug build is restored.

There is also no such thing as a default provided by CMake.

byrnHDF: > Remove default setting of CMAKE_DEBUG_POSTFIX if not set.

yes should just be:
Remove default setting of CMAKE_DEBUG_POSTFIX

byrnHDF: cacheinit.cmake is provided just for testing and convenience and should only be used if the settings agrree with your expectations. The file can be usedas is with overrides or editing. There are many alternatives. The preferred method now is to use Presets.

This change affects the preferred Presets method by removing the settings. The binaries we build for publication (which do not build debug) use the Presets method. Future changes will be to eliminate the cacheinit.cmake file for a "Testing" CMakeUsersPresets.json file that we would create and use.
Since the default CMake setting for this  CMAKE_DEBUG_POSTFIX isn't implemented the result is that it is the empty string, which is what we want.

byrnHDF (approved):

lrknox (approved):

MarDiehl:
thanks for all the support!

PR #5607, opened 2025-06-17
Correct tools test files ref paths

### LINKED ISSUE #5606: Change for tools testfiles exposed bug in vfd test
When the change to move tools test files into a common folder happened, most all the changes for the files were applied however the CMakeVFDTests.cmake file did not use the convention of `${HDF5_TOOLS_TST_DIR}/{tool}/testfiles` and instead relied on the value of `${PROJECT_SOURCE_DIR}/testfiles` to be the same.

byrnHDF:
Fixes #5606


mattjala (approved):

hyoklee (approved):

brtnfld (approved):

lrknox (approved):

tools/test/h5copy/CMakeTests.cmake
```diff
@@ -46,7 +46,7 @@
   endforeach ()
 
   foreach (listothers ${LIST_OTHER_TEST_FILES})
-    HDFTEST_COPY_FILE("${PROJECT_SOURCE_DIR}/expected/${listothers}" "${PROJECT_BINARY_DIR}/testfiles/${listothers}" "h5copy_files")
+    HDFTEST_COPY_FILE("${HDF5_TOOLS_TST_DIR}/h5copy/expected/${listothers}" "${PROJECT_BINARY_DIR}/testfiles/${listothers}" "h5copy_files")
```
jhendersonHDF, line 49: With the changes in develop none of these paths should have the tool name in them should they?

jhendersonHDF: Ah, the reference files are still under tool-specific sub-directories.

tools/test/h5dump/CMakeVFDTests.cmake
```diff
@@ -35,34 +35,34 @@ set (HDF5_SF2_VFD_H5DUMP_FILES
 foreach (vfdtest ${VFD_LIST})
   if (vfdtest STREQUAL "subfiling")
     foreach (h5_tfile ${HDF5_SF_VFD_H5DUMP_FILES})
-      file(COPY "${PROJECT_SOURCE_DIR}/testfiles/${h5_tfile}" DESTINATION "${PROJECT_BINARY_DIR}/${vfdtest}")
+      file(COPY "${HDF5_TOOLS_TST_DIR}/h5dump/testfiles/${h5_tfile}" DESTINATION "${PROJECT_BINARY_DIR}/${vfdtest}")
```
jhendersonHDF, line 38: The situation's less clear for the paths here though.

jhendersonHDF: We now have a bit of an awkward situation where most of the test files are in the shared location, some are still in tool-specific sub-directories and some are split between the old and new location, as in the case of the subfiling test files.

PR #5622, opened 2025-06-20
Always build test generators

mattjala:
The original motivation for only building the test generators if specifically requested was to avoid issues with the generators being forced to link statically with autotools. Since autotools is no longer support, this is no longer a concern.

- Remove the HDF5_BUILD_GENERATORS option, and always build test generators.


release_docs/AutotoolsToCMakeOptions.md
```diff
@@ -68,7 +68,6 @@ release_docs/INSTALL_CMake.txt file for more information on the CMake build syst
 | instrument[Enable library instrumentation of optimization tracing (only used with parallel builds). [default=yes if a parallel debug build, otherwise no]] | HDF5_ENABLE_INSTRUMENT "Instrument The library" [OFF] |  |
 | dconv-exception[Check exception handling functions during data conversions [default=yes]] | HDF5_WANT_DCONV_EXCEPTION "exception handling functions is checked during data conversions" [ON] |  |
 | dconv-accuracy[Guarantee data accuracy during data conversions [default=yes]] | HDF5_WANT_DATA_ACCURACY "IF data accuracy is guaranteed during data conversions" [ON] |  |
-| build-all[Build helper programs that only developers should need [default=no]] | HDF5_BUILD_GENERATORS "Build Test Generators" [OFF] |  |
```
mattjala, line 71: Is this a historical document that should retain old information? If not, perhaps it should be removed altogether since autotools is deprecated

byrnHDF: Yes this document is for those who used to build hdf5 with autotools before 2.0. As such it will need to be available for the near future.

byrnHDF (approved):

lrknox (approved):

PR #5623, opened 2025-06-20
Fix daily build S3 testing

jhendersonHDF:
Switches over the daily-build workflow jobs for S3 testing to use aws-c-s3 instead of trying to use libcurl.

### DIFF
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -66,6 +66,19 @@ jobs:
       use_tag: snapshot
       use_environ: snapshots
 
+  call-aws-c-s3-build:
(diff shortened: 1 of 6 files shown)


.github/workflows/daily-build.yml
```diff
@@ -66,6 +66,19 @@ jobs:
       use_tag: snapshot
       use_environ: snapshots
 
+  call-aws-c-s3-build:
```
jhendersonHDF, line 69: Build the aws-c-s3 library ahead of time and save an artifact that the builds use later on

lrknox (approved):

byrnHDF (approved):

PR #5640, opened 2025-06-27
Optimizations for repeated VDS source file and dataset names

### LINKED ISSUE #5639: Optimizations for repeated VDS source file and dataset names
We should implement optimizations for the common case where multiple VDS mappings share the same source file name and/or source dataset name. Both in memory and in the file. See this short RFC: https://github.com/HDFGroup/hdf5doc/blob/master/RFCs/HDF5_Library/VDS_Shared_Strings/VDS-Shared-Strings-Design.v2.pdf

fortnern:
Detects when source file and dataset names are shared between multiple VDS mappings in the same layout, and only stores each once in memory, with other instances linking to the first mapping with the same name. Adds new file format for VDS mappings that only stores each repeated string once when beneficial and allowed by libver_bounds setting.

Resolves #5639 

----

> [!IMPORTANT]
> Optimizes VDS mappings by storing repeated source file and dataset names once, updating file format to 4.0, and adding tests for shared string functionality.
> 
>   - **Behavior**:
>     - Optimizes VDS mappings by storing repeated source file and dataset names once and linking others to it.
>     - Updates file format to 4.0 to support shared string storage, used when `libver_bounds` is 2.0 or later.
>   - **Functions**:
>     - Modifies `H5D__virtual_store_layout()` in `H5Dvirtual.c` to calculate and store shared strings.
>     - Updates `H5O__layout_decode()` in `H5Olayout.c` to decode shared strings.
>   - **Tests**:
>     - Adds `test_vds_shared_strings()` in `dsets.c` to verify shared string functionality.
>   - **Misc**:
>     - Updates `RELEASE.txt` to document file format change.
>     - Adds hash table adjustments in `uthash.h` for realloc scenarios.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for aba9a006dd846b320a62a5963fb401a427a76c0d. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/H5Pdcpl.c
```diff
@@ -2080,6 +2082,10 @@ H5Pset_virtual(hid_t dcpl_id, hid_t vspace_id, const char *src_file_name, const
         H5O_storage_virtual_ent_t *x; /* Pointer to the new list */
         size_t new_alloc = MAX(H5D_VIRTUAL_DEF_LIST_SIZE, virtual_layout.storage.u.virt.list_nalloc * 2);
 
+        /* Clear hash tables */
+        HASH_CLEAR(hh_source_file, virtual_layout.storage.u.virt.source_file_hash_table);
```
mattjala: Why clear and rebuild the hash tables each time a mapping is added? I would think you could just check if the new mapping shares a file/dset name with any previous entry and add a key if it doesn't.

fortnern: We need to clear the hash tables before we reallocate the array of entries, since the pointers in the hash table may no longer be valid after the realloc. If it proves to be a performance issue, we can add a feature to uthash that will update these pointers in place (unless it already has that feature and I missed it).

fortnern: Note this clear doesn't happen every time we add a mapping, only when the list is realloced (it doubles every time we realloc).

src/H5Olayout.c
```diff
@@ -575,10 +578,12 @@ H5O__layout_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNU
                                     "ran off end of input buffer while decoding");
                     heap_vers = (uint8_t)*heap_block_p++;
 
-                    if ((uint8_t)H5O_LAYOUT_VDS_GH_ENC_VERS != heap_vers)
-                        HGOTO_ERROR(H5E_OHDR, H5E_VERSION, NULL,
-                                    "bad version # of encoded VDS heap information, expected %u, got %u",
-                                    (unsigned)H5O_LAYOUT_VDS_GH_ENC_VERS, (unsigned)heap_vers);
+                    assert(H5O_LAYOUT_VDS_GH_ENC_VERS_0 == 0);
```
mattjala, line 581: What's the purpose of this assert?

fortnern: If the VERS_0 macro were not binary 0 (unsigned) we would need to check for heap_vers < VERS_0 in the range check below

brtnfld:
@ellipsis-dev, please review this pr

doxygen/dox/H5.format.4.0.dox
```diff
@@ -0,0 +1,12706 @@
...
+  <td>Collection Address</td>
+  <td>This field is the address of the global heap collection where the data object is stored.</td>
+</tr>
+<tr>
+  <td>ID</td>
+  <td>This field is the index of the data object within the global heap collection.</td>
+</tr>
+</table>
+
+\subsection subsec_fmt3_infra_globalheapvds III.F. Disk Format: Level 1F - Global Heap Block for Virtual Datasets
```
mattjala: This is the only section of this document that isn't directly copied from the 3.0 file spec

test/dsets.c
```diff
@@ -16148,108 +16152,1085 @@ test_dcpl_layout_caching(H5D_layout_t layout_type)
...
+        TEST_ERROR;
+
+    /* Set file space strategy to paged aggregation and persisting free-space */
+    if (H5Pset_file_space_strategy(fcpl2, H5F_FSPACE_STRATEGY_PAGE, true, (hsize_t)1) < 0)
+        TEST_ERROR;
+
+    h5_fixname(FILENAME[0], fapl, filename, sizeof filename);
+
+    nerrors += (test_vds_shared_strings(fapl) < 0 ? 1 : 0);
+    exit(0);
```
fortnern: I think this exit(0) needs to be removed

mattjala: Yep, left that in from testing. Just fixed it

src/H5Olayout.c
mattjala: Are you thinking about the possibility that we might want to add a `VERS_ERROR` value later, or something like that? Other than that, I can't think of a reason we'd expect VERS_0 to change, and asserting the value of fixed enum constants is pretty unusual.

test/dsets.c
fortnern: It looks like the exit(0) is still there. It looks like you used a force push instead of a regular commit to attempt to change it, and it must not have worked. I'm not really sure what that message means.

fortnern:
Looks good to me. Thanks for your contributions Matt!

src/H5Olayout.c
fortnern: I simply put that insert in there to document why we were only checking for the decoded version being outside of the range on the high side. Normally when you check for a number being out of range you have to check for both above the maximum and below the minimum. Checking for below the minimum here would be a no-op and would generate a warning about checking  for unsigned < 0 is always false.

brtnfld:
@ellipses-dev, review PR

mattjala (dismissed):

mattjala (approved):

jhendersonHDF (approved):

PR #5666, opened 2025-07-14
Create a file containing the major CMake options

### LINKED ISSUE #5665: Create a file containing the major CMake options that HDF5 can be built with
Recent refactoring of HDF5's CMake logic has highlighted potential issues around the evaluation of specific CMake options, especially in regards to the order of evaluation of the CMake code.

For maintainability, a top-level file which contains all the major CMake options that HDF5 can be built with should be created. Major options would generally be those which are intended to always be available and are not dependent on other features/options of HDF5 being enabled. This would make it easier to find the major options of HDF5 and should help resolve any potential current and future evaluation ordering issues by keeping these options in a single location that is evaluated before other conditional checks.

byrnHDF:
Fixes #5665

----

> [!IMPORTANT]
> Centralizes major CMake options into `CMakeCacheOptions.cmake` for improved organization without changing build behavior.
> 
>   - **CMake Options Centralization**:
>     - Introduces `CMakeCacheOptions.cmake` to centralize major CMake options.
>     - Moves options from `CMakeLists.txt`, `CMakeFilters.cmake`, and `CMakeTests.cmake` to `CMakeCacheOptions.cmake`.
>   - **File Modifications**:
>     - `CMakeLists.txt`: Includes `CMakeCacheOptions.cmake` and removes redundant options.
>     - `CMakeFilters.cmake`: Replaces `cmake_dependent_option` with `option` for `HDF5_USE_ZLIB_NG` and `ZLIB_USE_EXTERNAL`.
>     - `CMakeTests.cmake`: Replaces `cmake_dependent_option` with `option` for various test-related options.
>   - **Behavior**:
>     - No change in build behavior; refactoring for better organization.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for e34ff8698291505df8a0608df94b9a32c93134b4. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


CMakeCacheOptions.cmake
```diff
@@ -0,0 +1,93 @@
+option (HDF5_USE_FOLDERS "Enable folder grouping of projects in IDEs." ON)
```
brtnfld: Could you provide a comment/heading that explains what the file is all about, i.e., why it was created, its functionality in the CMake universe, and its intended uses?

CMakeFilters.cmake
```diff
@@ -9,12 +9,12 @@
 # If you do not have access to either file, you may request a copy from
 # help@hdfgroup.org.
 #
-cmake_dependent_option (HDF5_USE_ZLIB_NG "Use zlib-ng library as zlib library" OFF HDF5_ENABLE_ZLIB_SUPPORT OFF)
+option (HDF5_USE_ZLIB_NG "Use zlib-ng library as zlib library" OFF)
 cmake_dependent_option (HDF5_USE_ZLIB_STATIC "Find static zlib library" OFF HDF5_ENABLE_ZLIB_SUPPORT OFF)
 cmake_dependent_option (HDF5_USE_LIBAEC_STATIC "Find static AEC library" OFF HDF5_ENABLE_SZIP_SUPPORT OFF)
-cmake_dependent_option (ZLIB_USE_EXTERNAL "Use External Library Building for ZLIB" OFF HDF5_ENABLE_ZLIB_SUPPORT OFF)
+option (ZLIB_USE_EXTERNAL "Use External Library Building for ZLIB" OFF)
```
bmribler, line 15: I see that HDF5_ was added to some for consistency but there are still some without it, such as this one, and there are some still have H5_, like H5_ZLIB_FOUND.  Is there a pattern/logic?

CMakeLists.txt
```diff
@@ -681,11 +664,10 @@ endif ()
 #-----------------------------------------------------------------------------
 # Add some definitions for Debug Builds
 #-----------------------------------------------------------------------------
-option (HDF5_ENABLE_TRACE "Enable API tracing capability" OFF)
-mark_as_advanced (HDF5_ENABLE_TRACE)
 if (${HDF_CFG_NAME} MATCHES "Debug" OR ${HDF_CFG_NAME} MATCHES "Developer")
   # Enable instrumenting of the library's internal operations
-  cmake_dependent_option (HDF5_ENABLE_INSTRUMENT "Instrument The library" OFF HDF5_ENABLE_TRACE OFF)
+  option (HDF5_ENABLE_INSTRUMENT "Instrument The library" OFF)
+  mark_as_advanced (HDF5_ENABLE_INSTRUMENT)
```
bmribler: Is there a reason for this (and some others) to be added here instead of the new file?

jhendersonHDF: In general (and we should comment on this in the new file), options that go into the new file should be those that are top-level options which are meant to always be available and aren't dependent on other options/features/configurations. `HDF5_ENABLE_INSTRUMENT` is specific to the configuration type here and also is only really relevant for parallel. Because of that, I would call it not a top-level option.

brtnfld:
I would call the new file CMakeOptions.cmake

CMakeFilters.cmake
byrnHDF: Yes, Ones without HDF5 prefix are externally named options.
H5 prefix is for variables used by the code.

byrnHDF:
> I would call the new file CMakeOptions.cmake
Went with CMakeBuildOptions.cmake instead, the other is too general. The other choice would be CMakeMajorOptions.cmake - but that seemed wrong.

CMakeFilters.cmake
bmribler: Are these differences noted anywhere?  For the next generation. :D

CMakeBuildOptions.cmake
```diff
@@ -0,0 +1,86 @@
+# Put all top-level options build options into one place
```
lrknox, line 1: Should this comment go on to say that dependent and lower level options should not be entered in this file?

byrnHDF: That's not an absolute truth, there may be a good reason to put one here. Depends on the use case.

lrknox: ok

lrknox (dismissed):

CMakeLists.txt
```diff
@@ -695,7 +676,6 @@ if (${HDF_CFG_NAME} MATCHES "Debug" OR ${HDF_CFG_NAME} MATCHES "Developer")
   if (HDF5_ENABLE_INSTRUMENT)
     set (H5_HAVE_INSTRUMENTED_LIBRARY 1)
   endif ()
-  mark_as_advanced (HDF5_ENABLE_INSTRUMENT)
```
jhendersonHDF, line 698: Why the removal of this?

CMakeLists.txt
```diff
@@ -711,7 +691,6 @@ endif ()
 if (CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
   set (H5_HAVE_EMBEDDED_LIBINFO 0)
 else ()
-  option (HDF5_ENABLE_EMBEDDED_LIBINFO "Embed library info into executables" ON)
```
jhendersonHDF, line 714: Moving this option to the top-level build file kind of changes the semantics here since this was previously inside a conditional.

CMakeLists.txt
```diff
@@ -1177,7 +1143,6 @@ endif ()
 # Option to build High Level API's
 #-----------------------------------------------------------------------------
 if (EXISTS "${HDF5_SOURCE_DIR}/hl" AND IS_DIRECTORY "${HDF5_SOURCE_DIR}/hl")
-  option (HDF5_BUILD_HL_LIB "Build HIGH Level HDF5 Library" ON)
```
jhendersonHDF, line 1180: What was the original intention for the check above for several options? Moving these to the top-level file also somewhat changes the semantics since they're moved outside the conditional, but it's not clear whether it makes sense to move or duplicate the check to the top-level file as well.

CMakeTests.cmake
```diff
@@ -15,15 +15,15 @@
 #-----------------------------------------------------------------------------
   set (DART_TESTING_TIMEOUT 1200
       CACHE STRING
-      "Timeout in seconds for each test (default 1200=20minutes)"
```
jhendersonHDF: I think this FORCE should be removed

CMakeTests.cmake
```diff
@@ -33,45 +33,41 @@
...
-    cmake_dependent_option (HDF5_TEST_FHEAP_PASSTHROUGH_VOL "Execute fheap test with different passthrough VOL connectors" ON HDF5_TEST_PASSTHROUGH_VOL OFF)
-    mark_as_advanced (HDF5_TEST_FHEAP_PASSTHROUGH VOL)
-  endif ()
+  cmake_dependent_option (HDF5_TEST_FHEAP_PASSTHROUGH_VOL "Execute fheap test with different passthrough VOL connectors" ON HDF5_TEST_PASSTHROUGH_VOL OFF)
+  mark_as_advanced (HDF5_TEST_FHEAP_PASSTHROUGH VOL)
 
   set (H5_TEST_EXPRESS_LEVEL_DEFAULT "3")
   set (HDF_TEST_EXPRESS "${H5_TEST_EXPRESS_LEVEL_DEFAULT}"
-      CACHE STRING "Control testing framework (0-3) (0 = exhaustive testing; 3 = quicker testing)")
+      CACHE STRING "Control testing framework (0-3) (0 = exhaustive testing; 3 = quicker testing)" FORCE)
```
jhendersonHDF: And this one as well

CMakeFilters.cmake
byrnHDF: No but all the ones that need to be changed have been done. Not sure how to document which is which, kind of just some CMake knowledge.

CMakeLists.txt
byrnHDF: Since it was a cmake_dependent_option, didn't think it was necessary.

CMakeLists.txt
byrnHDF: It wasn't a controlling conditional just a don't do it for emscripten

jhendersonHDF: Right, which changes the semantics for emscripten. It's not a big deal, just noting that it's not the same as before.

CMakeLists.txt
byrnHDF: It was an attempt to be sure the directory exists. Probably not really necessary.

CMakeTests.cmake
byrnHDF: ok

CMakeTests.cmake
byrnHDF: ok

CMakeLists.txt
jhendersonHDF: What gets made advanced is fairly arbitrary so it's not like this is a major change, but since the dependent part is based on `HDF5_ENABLE_TRACE` it has a similar effect of `mark_as_advanced` in only some cases. No big deal though

jhendersonHDF (approved):

lrknox (approved):

PR #5685, opened 2025-07-23
Change default page buffer size to 64MiB when using ROS3 driver

### LINKED ISSUE #5274: Set page buffer size to non-zero value for ROS3 driver
Library's page buffer is available for all files created with page aggregation file space management strategy. The current default page buffer size is zero bytes. Page buffer is shown to significantly improve performance for cloud optimized files so I propose to set its size to a non-zero value when ROS3 driver is used. Suggested page buffer size: 64 MiB.

fortnern:
With this change, H5FDset_fapl_ros3() now sets the page buffer size to 64MiB if it was not set before on the FAPL in question.

Added a new macro, H5F_PAGE_BUFFER_SIZE_DEFAULT, to keep track of when the page buffer size has been set on a FAPL. Setting this value causes it to be "unset". Added test to make sure this value and related values are encoded and decoded properly.

Closes #5274 

----

> [!IMPORTANT]
> Change default page buffer size to 64MB for ROS3 driver and add tests for encoding/decoding this setting.
> 
>   - **Behavior**:
>     - `H5FDset_fapl_ros3()` now sets page buffer size to 64MB if not set.
>     - Introduces `H5F_PAGE_BUFFER_SIZE_DEFAULT` to manage default page buffer size.
>   - **Code Changes**:
>     - `H5FDros3.c`: Sets default page buffer size in `H5Pset_fapl_ros3()`.
>     - `H5Pdapl.c`: Adds encoding/decoding for page buffer size.
>   - **Tests**:
>     - `enc_dec_plist.c`: Adds `test_encode_decode_page_buf_size()` to verify encoding/decoding of default page buffer size.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 99f38b6c164c2ccc2e9a295f62e417bc3e4eec7d. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


release_docs/RELEASE.txt
```diff
@@ -200,6 +200,14 @@ New Features
     Library:
     --------
 
+    - Changed the default page buffer size for the ROS3 driver
+
+      Calling H5FDset_fapl_ros3() now has the side effect of setting the page
+      buffer size in the FAPL to 64MB if it was not previously set. This will
```
ajelenak: Based on the page buffer's value for `ROS3_DEF_PAGE_BUF_SIZE` the units should be MiB, so "64MB" -> "64MiB".

src/H5FDros3.h
```diff
@@ -250,6 +250,14 @@ H5_DLL herr_t H5Pget_fapl_ros3(hid_t fapl_id, H5FD_ros3_fapl_t *fa_out);
  * \fapl_id
  * \param[in] fa Pointer to #H5FD_ROS3 driver configuration structure.
  * \returns \herr_t
+ *
+ * \note    As of HDF5 2.0.0, as a side effect of calling this function, if the
+ *          page buffer size has not been set on fapl_id, it is set to 64 MB. To
```
ajelenak: Same change here: 64MB -> 64MiB. https://simple.wikipedia.org/wiki/Mebibyte

src/H5Fpublic.h
```diff
@@ -71,6 +71,9 @@
  * member is unknown \since 1.8.0 */
 #define H5F_FAMILY_DEFAULT 0 /* (hsize_t) */
 
+/* Macro used to "unset" the page buffer size in a FAPL */
+#define H5F_PAGE_BUFFER_SIZE_DEFAULT SIZE_MAX
```
mattjala, line 75: I think this should probably be named differently to indicate that it's a placeholder/overriding value and not a meaningful default. As it is, it's difficult to discern the difference between H5F_PAGE_BUFFER_SIZE_DEFAULT and H5PB_SIZE_DEFAULT_VALUE.

mattjala (dismissed):
Looks good aside from a minor comment

release_docs/RELEASE.txt
```diff
@@ -200,6 +200,14 @@ New Features
     Library:
     --------
 
+    - Changed the default page buffer size for the ROS3 driver
+
+      Calling H5FDset_fapl_ros3() now has the side effect of setting the page
```
jhendersonHDF: `H5FDset_fapl_ros3()` -> `H5Pset_fapl_ros3()`

src/H5Fpublic.h
fortnern: It is patterned after the existing H5D_CHUNK_CACHE_NBYTES_DEFAULT, but what do you suggest? It would be easier to change H5PB_SIZE_DEFAULT_VALUE since that's internal.

src/H5Pfapl.c
```diff
@@ -5820,6 +5826,142 @@ H5Pget_page_buffer_size(hid_t plist_id, size_t *buf_size /*out*/, unsigned *min_
...
+
+    assert(enc_size < 256);
+
+    if (NULL != *pp) {
+        /* Encode the size */
+        *(*pp)++ = (uint8_t)enc_size;
+
+        /* Encode the value if necessary */
+        if (enc_size != 0) {
+            UINT64ENCODE_VAR(*pp, enc_value, enc_size);
```
jhendersonHDF, line 5904: Instead of casting the value for the page buffer size to a uint64_t and using `UINT64ENCODE_VAR` here, could `ENCODE_VAR` be used instead so the somewhat odd compile assertion of `sizeof(size_t) <= sizeof(uint64_t)` above could be avoided?

fortnern: Possibly, but since this is how H5P__encode_size_t works I think it would be best to address this issue separately in a PR that includes H5P__encode/decode_size_t and others derived from it (including at least H5P__encode/decode chunk_cache_nbytes/nslots)

jhendersonHDF (approved):

mattjala (approved):

PR #5694, opened 2025-07-29
Fix memory leak when decoding a property list with a virtual dataset layout

fortnern:
The code to decode a layout encoded on a serialized DCPL has been updated to properly build the source file and dataset name hash tables, and set the "orig" fields properly so the name buffers are freed if they're not shared. This should fix the memory leaks that occurred because the orig fields were not set, as well as avoid a potential bug if a duplicate source file or dataset were added after decoding.

Addresses #5695 

----

> [!IMPORTANT]
> Fix memory leak in `H5P__dcrt_layout_dec()` by setting "orig" fields and building hash tables for source file and dataset names.
> 
>   - **Memory Management**:
>     - Fix memory leak in `H5P__dcrt_layout_dec()` by properly setting "orig" fields for source file and dataset names.
>     - Ensure name buffers are freed if not shared, preventing memory leaks.
>   - **Hash Tables**:
>     - Build hash tables for source file and dataset names in `H5P__dcrt_layout_dec()` to avoid duplicate entries.
>     - Use `HASH_FIND` and `HASH_ADD_KEYPTR` to manage entries efficiently.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for e870406d6c830b2a0120e543aba5b4b5d0f83363. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5Pdcpl.c
+++ b/src/H5Pdcpl.c
@@ -630,45 +630,61 @@ H5P__dcrt_layout_dec(const void **_pp, void *value)
 
                 /* Decode each entry */
                 for (u = 0; u < (size_t)nentries; u++) {
-                    H5O_storage_virtual_ent_t *tmp_ent;         /* Temporary VDS entry pointer, for hash table lookups */
+                    H5O_storage_virtual_ent_t
+                        *tmp_ent; /* Temporary VDS entry pointer, for hash table lookups */
 
                     /* Check for source file name in hash table */
-                    tmp_ent = NULL;
+                    tmp_ent  = NULL;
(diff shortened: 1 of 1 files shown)


qkoziol (approved):

byrnHDF (approved):

byrnHDF:
This change can be tested by executing the dailybuild on this branch in your fork (change check to ignore)

PR #5717, opened 2025-08-08
Correct VFD h5repack test file cleanup

mattjala:
- Avoid attempting to delete default-driver exclusive test files with non-default drivers

- Use `h5_delete_test_file` for file deletions, which accounts for some requirements that the Family VFD has for filename properties

----

> [!IMPORTANT]
> Improve test file cleanup in `h5repacktst.c` by using `h5_delete_test_file` and avoiding deletion of default-driver exclusive files with non-default drivers.
> 
>   - **Behavior**:
>     - Use `h5_delete_test_file` in `h5repacktst.c` for file deletions, ensuring compatibility with Family VFD filename requirements.
>     - Avoid deleting default-driver exclusive files with non-default drivers in `h5repacktst.c`.
>   - **File Management**:
>     - Reorganize file arrays in `h5repackgentest.h` to separate default-driver files from others.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 1ad0337601a46dd81b4fc479bcce83c428b68fe9. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


tools/test/h5repack/h5repacktst.c
```diff
@@ -1592,17 +1594,36 @@ main(void)
...
+
+    h5_delete_all_test_files(H5REPACK_TEST_H5_FILES, fapl_id);
+
     /* Also clean up H5REPACK_FSPACE_FNAMES array files */
     for (size_t i = 0; i < NELMTS(H5REPACK_FSPACE_FNAMES); i++) {
-        if (H5Fdelete(H5REPACK_FSPACE_FNAMES[i], H5P_DEFAULT) < 0) {
-            printf(" Failed to delete %s", H5REPACK_FSPACE_FNAMES[i]);
-            GOERROR;
+        /* Only the files at these indices are unconditionally created */
+        if (i == 0 || i == 1 || i == 2) {
```
jhendersonHDF: This is going to be fragile at best since new filenames aren't always inserted at the end of lists and often end up somewhere in the middle. It seems that rather than trying to cleanup files at the end and guessing which ones are there, it would be better to either ignore failures (and enclose these in `H5E_BEGIN_TRY`/`H5E_END_TRY` blocks), or separate these into sub-functions that are responsible for cleaning up their own test files.

mattjala: `h5_delete_test_file()` already ignores cleanup failures, so this check was unnecessary in the first place - I've merged the two blocks that clean up `H5REPACK_FSPACE_FNAMES`.

brtnfld:
@greptile

jhendersonHDF (approved):

brtnfld (approved):

PR #5719, opened 2025-08-11
Fix VOL CMake builds

mattjala:
Assuming that this string would never be a list caused build failures with several VOL connectors. We now check that it's not a list before performing the string comparison.

----

> [!IMPORTANT]
> Fixes build failures by ensuring `vol_lib_targets` is not a list before string comparison in `HDF5_GET_VOL_TGT_INFO`.
> 
>   - **Behavior**:
>     - Fixes build failures by checking if `vol_lib_targets` is not a list before string comparison in `HDF5_GET_VOL_TGT_INFO` in `HDF5Macros.cmake`.
>     - Ensures `vol_lib_targets` is not a list by using `string(FIND ...)` to detect semicolons.
>   - **Error Handling**:
>     - If `vol_lib_targets` is not found, a fatal error is triggered with a message indicating the missing targets.
>   - **Misc**:
>     - Minor formatting change in `CMakeTests.cmake` (added a newline).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 6314701b9945a922833b497e4d2bfe7a6b7e4207. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config/HDF5Macros.cmake
+++ b/config/HDF5Macros.cmake
@@ -138,8 +138,12 @@ macro(HDF5_GET_VOL_TGT_INFO vol_tgt vol_name_out vol_env_out)
   # HDF5_PLUGIN_PATH
   get_target_property(vol_lib_targets "${vol_tgt}" HDF5_VOL_TARGETS)
 
-  if (${vol_lib_targets} STREQUAL vol_lib_targets-NOTFOUND)
-    message(FATAL_ERROR "VOL target ${vol_tgt} has no defined HDF5_VOL_TARGETS")
+  # Sanity check
+  string(FIND "${vol_lib_targets}" ";" semicolon_pos)
+  if (semicolon_pos EQUAL -1)
+    if ("${vol_lib_targets}" STREQUAL vol_lib_targets-NOTFOUND)
+      message(FATAL_ERROR "${vol_tgt} has no corresponding targets")
+    endif ()
   endif ()
 
   set(vol_plugin_paths "${CMAKE_BINARY_DIR}/${HDF5_INSTALL_BIN_DIR}")
--- a/config/HDF5Macros.cmake
+++ b/config/HDF5Macros.cmake
@@ -141,7 +141,7 @@ macro(HDF5_GET_VOL_TGT_INFO vol_tgt vol_name_out vol_env_out)
   # Sanity check
   string(FIND "${vol_lib_targets}" ";" semicolon_pos)
   if (semicolon_pos EQUAL -1)
-    if ("${vol_lib_targets}" STREQUAL vol_lib_targets-NOTFOUND)
+    if ("${vol_lib_targets}" STREQUAL "vol_lib_targets-NOTFOUND")
       message(FATAL_ERROR "${vol_tgt} has no corresponding targets")
     endif ()
   endif ()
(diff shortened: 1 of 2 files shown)


lrknox (approved):

vchoi-hdfgroup (approved):

config/HDF5Macros.cmake
```diff
@@ -138,8 +138,12 @@ macro(HDF5_GET_VOL_TGT_INFO vol_tgt vol_name_out vol_env_out)
   # HDF5_PLUGIN_PATH
   get_target_property(vol_lib_targets "${vol_tgt}" HDF5_VOL_TARGETS)
 
-  if (${vol_lib_targets} STREQUAL vol_lib_targets-NOTFOUND)
-    message(FATAL_ERROR "VOL target ${vol_tgt} has no defined HDF5_VOL_TARGETS")
+  # Sanity check
+  string(FIND "${vol_lib_targets}" ";" semicolon_pos)
+  if (semicolon_pos EQUAL -1)
```
jhendersonHDF, line 143: It seems it would be better to simply move this check down into the `foreach` below. This variable should always be a list, so there's no need to check if it is, and rather than only checking if a single target is `STREQUAL "vol_lib_targets-NOTFOUND"`, it's probably best to check all of them before trying to use them.

jhendersonHDF (dismissed):
I believe this may need a slightly different approach

config/HDF5Macros.cmake
mattjala: This is meant as a sanity check to detect if `vol_tgt` has a bad `HDF5_VOL_TARGETS` property. The external passthrough connector actually has this value as a single string - which is why I didn't detect this issue until github CI caught it with other connectors.

I'm not sure that checking each of the individual values in the resulting `vol_lib_targets`  makes sense - the NOTFOUND value that's being checked for here would only result from `get_target_property` not finding a value for the `HDF5_VOL_TARGETS` property. If the property had a list value, then the sanity check has already been passed.

Without this check, if the VOL target somehow was missing the property, then that could lead to the resulting `vol_env_out` not actually providing the plugin path, which could be tricky to trace back to here. That said, if we have a good reason to assume this property will always be populated, this check could be removed entirely.

jhendersonHDF: Apologies, I was misreading the intent of the logic a bit here and conflating checking the property vs checking for unpopulated targets for a particular VOL target. In the current code the `HDF5_VOL_TARGETS` property will always be set for a VOL target, so the check could probably be removed. That said, I do prefer sanity checks in general, especially in our CMake code, so I'm not opposed to leaving one in as it could prevent misuse of this macro on a non-VOL target. Though, even with only one string in it, this variable will still always be a CMake list, so perhaps what you're looking for is simply a `list (LENGTH ...)` comparison rather than trying to find a semicolon? You don't necessarily need that check here either though, as the new form `if ("${vol_lib_targets}" STREQUAL "vol_lib_targets-NOTFOUND")` that you have here should do by itself. By quoting `${vol_lib_targets}`, CMake will treat the list as a string rather than trying to expand the list and causing the error. I'd lean toward just doing that instead of also trying to check whether the variable is a string or list (which are effectively the same thing in CMake). At least for `STREQUAL`, always quoting is a good best practice IMO, even though we sometimes violate that.

My point about moving the check into the `foreach` was more that the actual underlying list of targets is more interesting to check as the normal failure case will probably be that one of them wasn't populated correctly. Checking each element of the list before using it is probably a good idea, though unrelated to this PR and not worth worrying about right now.

jhendersonHDF (approved):
Fine as-is, though the check for a string vs. a list is probably not needed

Issue #5376, opened 2025-03-14
Use After Free in H5FL__blk_gc_list


sae-as-me:
## Affected Projects
hdf5 v1.14.6 ([https://github.com/HDFGroup/hdf5](https://github.com/HDFGroup/hdf5))
## Problem Type
CWE-416: Use After Free
### Summary
A heap-use-after-free vulnerability was discovered in the H5FL__blk_gc_list function within the HDF5 Library. This issue occurs when processing certain files, leading to an out-of-bounds read and potential application crash.
### Details
The vulnerability arises in the H5FL__blk_gc_list function defined in src/H5FL.c at line 985. This function does not check whether the lists in multiple chunks point to the same block address, resulting in duplicated addresses that have already been freed.

```c++
static herr_t
H5FL__blk_gc_list(H5FL_blk_head_t *head)
{
    H5FL_blk_node_t *blk_head; /* Temp. ptr to the free list page node */

    FUNC_ENTER_PACKAGE_NOERR

    /* Loop through all the nodes in the block free list queue */
    blk_head = head->head;
    while (blk_head != NULL) {
        H5FL_blk_node_t *blk_next; /* Temp. ptr to the next free list node */
        H5FL_blk_list_t *list;     /* The free list of native nodes of a particular size */

        /* Sanity check */
        assert((blk_head->onlist && blk_head->list) || (0 == blk_head->onlist && NULL == blk_head->list));

        /* Loop through all the blocks in the free list, freeing them */
        list = blk_head->list;    //The list addresses used in both loops are the same.
        while (list != NULL) {
            H5FL_blk_list_t *next; /* Temp. ptr to the free list list node */

            /* Get the pointer to the next node */
            next = list->next;    //heap-use-after-free                                   <---------
                                                                                                 //|
            /* Free the block */                                                                 //|
            H5MM_free(list);    //freed in this,but used in the next while (blk_head != NULL) loop--

            /* Advance to the next node */
            list = next;
        } /* end while */

        /* 
        ...other codes
        */
        
    } /* end while */

    /* Indicate no free nodes on the free list */
    head->onlist = 0;

    /* Double check that all the memory on this list is recycled */
    assert(0 == head->list_mem);

    FUNC_LEAVE_NOAPI(SUCCEED)
} /* end H5FL__blk_gc_list() */
```

### PoC
**Steps to reproduce:**
1. Clone the hdf5 repository and build it using the following commands : 
```sh
export CC='clang'
[...]

PR #5209, opened 2025-01-07
Fix oss-fuzz vulns

oss-patch:
commit [c792aa1](https://github.com/aled-ua/hdf5/commit/c792aa1108b5774d548235ba8da5dd873b779bba) Fix

- [OSV-2023-76](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61803) - [POC](https://oss-fuzz.com/download?testcase_id=4582680828968960)
Full Sanitizer Report:
```
==13557==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000003640 at pc 0x5627fc3fc76d bp 0x7ffcee65e870 sp 0x7ffcee65e868
READ of size 4 at 0x502000003640 thread T0
    #0 0x5627fc3fc76c in H5SM_delete /root/src/H5SM.c:1542:24
    #1 0x5627fc1bef62 in H5O__msg_write_real /root/src/H5Omessage.c:364:13
    #2 0x5627fc1be718 in H5O_msg_write /root/src/H5Omessage.c:246:9
    #3 0x5627fc0163ab in H5G__stab_valid /root/src/H5Gstab.c:1016:13
    #4 0x5627fc00d689 in H5G_mkroot /root/src/H5Groot.c:235:21
    #5 0x5627fbebb4b8 in H5F_open /root/src/H5Fint.c:2134:13
    #6 0x5627fca30d89 in H5VL__native_file_open /root/src/H5VLnative_file.c:127:9
    #7 0x5627fc9dfe16 in H5VL__file_open /root/src/H5VLcallback.c:3714:25
    #8 0x5627fc9df44d in H5VL_file_open /root/src/H5VLcallback.c:3832:30
    #9 0x5627fbe8d874 in H5F__open_api_common /root/src/H5F.c:780:29
    #10 0x5627fbe8ca70 in H5Fopen /root/src/H5F.c:820:22
    #11 0x5627fbc5ab6c in LLVMFuzzerTestOneInput /root/src/h5_extended_fuzzer.c:29:24
    #12 0x5627fbb65f04 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/root/out/h5_extended_fuzzer+0x188f04) (BuildId: fbd396d65561ef37960c8b3f8a14c0d50bb772c3)
[...]

PR #5722, opened 2025-08-13
Validate page size to refix PR-5209

bmribler:
This PR addresses the root cause of the issue by adding a sanity-check immediately after reading the file space page size from the file.  All the POCs files in the original PR, [PR-5209](https://github.com/HDFGroup/hdf5/pull/5209), were verified.

The same fuzzer in GH-5376 was used to verify that the assert before the vulnerability had occurred and that an error indicating a corrupted file space page size replaced it.

----

> [!IMPORTANT]
> Adds a sanity check for file space page size in `H5O__fsinfo_decode()` to prevent vulnerabilities from corrupted files.
> 
>   - **Sanity Check**:
>     - Adds a sanity check in `H5O__fsinfo_decode()` in `H5Ofsinfo.c` to validate `fsinfo->page_size` is not zero and does not exceed `H5F_FILE_SPACE_PAGE_SIZE_MAX`.
>   - **Verification**:
>     - Verified with fuzzer from GH-5376 to ensure error is caught before vulnerability occurs.
>   - **Misc**:
>     - Initializes `fsinfo` with `memset` in `H5F__super_read()` in `H5Fsuper.c` to ensure clean state.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 54f404b5ad8e63d99e3283646b543b2842a22fd3. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5Fsuper.c
+++ b/src/H5Fsuper.c
@@ -730,6 +730,8 @@ H5F__super_read(H5F_t *f, H5P_genplist_t *fa_plist, bool initial_read)
             if (!(flags & H5O_MSG_FLAG_WAS_UNKNOWN)) {
                 H5O_fsinfo_t fsinfo; /* File space info message from superblock extension */
 
+                memset(&fsinfo, 0, sizeof(H5O_fsinfo_t));
+
                 /* f->shared->null_fsm_addr: Whether to drop free-space to the floor */
                 /* The h5clear tool uses this property to tell the library
                  * to drop free-space to the floor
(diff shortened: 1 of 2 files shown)


mattjala (approved):

src/H5Fsuper.c
```diff
@@ -730,6 +730,8 @@ H5F__super_read(H5F_t *f, H5P_genplist_t *fa_plist, bool initial_read)
             if (!(flags & H5O_MSG_FLAG_WAS_UNKNOWN)) {
                 H5O_fsinfo_t fsinfo; /* File space info message from superblock extension */
 
+                memset(&fsinfo, 0, sizeof(H5O_fsinfo_t));
```
qkoziol, line 733: This is sensible, but are there other structs for messages that could be pro-actively memset() also?

qkoziol (approved):

src/H5Fsuper.c
bmribler: I'm sure there are and there are also other values read from file too, but I'm trying to keep each PR concentrated to one issue, baby step. :D  Once all security issues are done, we can do a final sweep for the undetected ones.

Issue #5383, opened 2025-03-16
Double Free in H5MM_xfree


sae-as-me:
## Affected Projects
hdf5 v1.14.6 ([https://github.com/HDFGroup/hdf5](https://github.com/HDFGroup/hdf5))
## Problem Type
CWE-415: Double Free 
### Summary
A double-free vulnerability was discovered in the `H5MM_xfree` function within the HDF5 Library. This issue occurs when processing certain files, leading to a potential application crash.
### Details
The vulnerability arises in the `H5MM_xfree` function defined in `src/H5MM.c` at line `201`. 
The reason is that the function `H5MM_realloc` defined at line `87` in file `src/H5MM.c` uses `realloc` to release the memory pointed to by pointer `mem` when handling `size` as `0`, but does not set pointer `mem` to NULL afterwards. Causing subsequent operations to result in double free when calling the `H5MM_xfree` function. Similarly, the function `H5MM_xfree` does not set the corresponding pointer to NULL after using `free`.
```c
// src/H5MM.c
void *
H5MM_realloc(void *mem, size_t size)
{
    void *ret_value = NULL;

    /* Use FUNC_ENTER_NOAPI_NOINIT_NOERR here to avoid performance issues */
    FUNC_ENTER_NOAPI_NOINIT_NOERR

    if (NULL == mem && 0 == size)
        /* Not defined in the standard, return NULL */
        ret_value = NULL;
    else {
        ret_value = realloc(mem, size);     //<==>free(mem);ret_value = NULL; when size = 0
                                            //Should the mem pointer be set to NULL here?
        /* Some platforms do not return NULL if size is zero. */
        if (0 == size)
            ret_value = NULL;
    }

    FUNC_LEAVE_NOAPI(ret_value)
} /* end H5MM_realloc() */
```
```c
H5MM_xfree(void *mem)
{
    /* Use FUNC_ENTER_NOAPI_NOINIT_NOERR here to avoid performance issues */
    FUNC_ENTER_NOAPI_NOINIT_NOERR

    free(mem);      //double-free
                    //Should the mem pointer be set to NULL here?
    FUNC_LEAVE_NOAPI(NULL)
} /* end H5MM_xfree() */
```
### PoC
**Steps to reproduce:**
1. Clone the hdf5 repository and build it using the following commands : 
```sh
export CC='clang'
export CXX='clang++'
export CFLAGS='-fsanitize=address -g'
export CXXFLAGS='-fsanitize=address -g'

[...]

PR #5737, opened 2025-08-20
Fix CVE-2025-2925

glennsong09:
This PR fixes issue #5383, which was occurring due to actual_len + H5C_IMAGE_EXTRA_SPACE being 0. When realloc was called, it freed image, but gets sent to done before new_image can be assigned to image. Because the pointer for image isn't null, it attempts to free it here again, causing the double free to occur. This PR removes new_image, which is only used for pointer conversion and directly uses image instead, which then avoids the above situation.

The bug was first reproduced using the fuzzer and the POC file from https://github.com/HDFGroup/hdf5/issues/5383. With this change, the double free no longer occurs.

### DIFF
--- a/src/H5Centry.c
+++ b/src/H5Centry.c
@@ -1053,10 +1053,10 @@ H5C__load_entry(H5F_t *f,
         do {
             if (actual_len != len) {
                 new_image = H5MM_realloc(image, len + H5C_IMAGE_EXTRA_SPACE);
-                image = (uint8_t *)new_image;
+                image     = (uint8_t *)new_image;
                 if (NULL == image)
                     HGOTO_ERROR(H5E_CACHE, H5E_CANTALLOC, NULL, "image null after H5MM_realloc()");
-                
+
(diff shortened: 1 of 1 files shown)


byrnHDF (dismissed):

qkoziol (changes requested):
'new_image' is not here for casting, it's used to preserve the 'image' pointer when realloc fails.  See the man page for realloc for full info, but the relevant portion is: " For realloc(), the input pointer is still valid if reallocation failed."

Another way should be found to address the CVE.

jhendersonHDF:
It would seem a better fix here would be to either check for a realloc size of 0 and throw an error in that case (if an image size of 0 isn't allowed), or simply set `image` to NULL before the goto macro when the realloc call returns NULL. I'm assuming a realloc size of 0 is unexpected here, but would have to read into the logic more. There are two calls to `H5MM_realloc` in that function, make sure to get both of them.

PR #5752, opened 2025-08-29
Merge h5dump test macros

mattjala:
Merge several macros in the h5dump test code into `ADD_H5_TEST` to reduce duplication.

### DIFF
--- a/tools/test/h5dump/CMakeTests.cmake
+++ b/tools/test/h5dump/CMakeTests.cmake
@@ -466,6 +466,10 @@ endmacro ()
 #   TARGET_FILE <filename> - the file to target with h5dump
 #   RESULT_CODE <code>    - expected return code after test execution. 0 is success
 #
+# OPTIONAL FLAG ARGUMENTS:
+#   BINARY_OUTPUT - whether to pass the binary output flag (-b) to h5dump
+#                   requires OUTPUT_FILE
+#
 # OPTIONAL KEYWORD ARGUMENTS:
 #   APPLY_FILTERS <resultvalue> - If provided, test will apply filters to output before comparison.
 #                                 <resultvalue> is used to construct the filter expressions
@@ -477,7 +481,7 @@ endmacro ()
 #
 macro (ADD_H5_TEST testname)
   cmake_parse_arguments(ARG
-    ""
+    "BINARY_OUTPUT"
     "RESULT_CODE;APPLY_FILTERS;TARGET_FILE;OUTPUT_FILE;DDL_FILE"
     "ANY_PATHS"
     ${ARGN}
@@ -563,9 +567,19 @@ macro (ADD_H5_TEST testname)
     set (ARG_DDL_FILE_CMD "")
   endif ()
 
+  if (${ARG_BINARY_OUTPUT})
+    if (NOT DEFINED ARG_OUTPUT_FILE)
+      message(FATAL_ERROR "ADD_H5_TEST: BINARY_OUTPUT flag requires OUTPUT_FILE")
+    endif ()
+
+    set (BINARY_OUTPUT_FLAG "-b")
+  else ()
+    set (BINARY_OUTPUT_FLAG "")
+  endif()
+
   # If using memchecker add tests without using scripts
   if (HDF5_ENABLE_USING_MEMCHECKER)
-    add_test (NAME H5DUMP-${testname} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5dump> ${ARG_ANY_PATHS} ${ARG_UNPARSED_ARGUMENTS} ${ARG_DDL_FILE_CMD} ${ARG_OUTPUT_FILE} ${ARG_TARGET_FILE})
+    add_test (NAME H5DUMP-${testname} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5dump> ${ARG_ANY_PATHS} ${ARG_UNPARSED_ARGUMENTS} ${ARG_DDL_FILE_CMD} ${BINARY_OUTPUT_FLAG} ${ARG_OUTPUT_FILE} ${ARG_TARGET_FILE})
     if (${ARG_RESULT_CODE})
       set_tests_properties (H5DUMP-${testname} PROPERTIES WILL_FAIL "true")
     endif ()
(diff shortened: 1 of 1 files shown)


byrnHDF (dismissed):

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -560,6 +560,10 @@ macro (ADD_H5_TEST testname)
     list(APPEND CLEANUP_FILES "${testname}.txt")
   endif ()
 
+  if (DEFINED ARG_OUTPUT_FILE)
```
mattjala: Even after this change, file cleanup is incomplete - I'm looking into why.

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -1375,120 +1156,120 @@ ADD_H5_TEST (tfloat16_be 0 --enable-error-stack tfloat16_be.h5)
...
 # test to verify github issue#3790: infinite loop closing library
-ADD_H5ERR_MASK_TEST (infinite_loop 1 "unable to open file" 3790_infinite_loop.h5)
+ADD_H5_TEST (infinite_loop RESULT_CODE 1 H5ERRREF "unable to open file" TARGET_FILE 3790_infinite_loop.h5)
 
 # test to verify HDFFV-10333: error similar to H5O_attr_decode in the jira issue
-ADD_H5ERR_MASK_TEST (err_attr_dspace 1 "error getting attribute information" err_attr_dspace.h5)
+ADD_H5_TEST (err_attr_dspace RESULT_CODE 1 H5ERRREF "error getting attribute information" TARGET_FILE err_attr_dspace.h5)
 
 # test to verify HDFFV-9407: long double full precision
-#  ADD_H5_GREP_TEST (t128bit_float 1 "1.123456789012345" -m %.35Lg t128bit_float.h5)
```
byrnHDF, line 1477: Probably should convert this as well. Should this be changed to a SKIP test instead of using a comment

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -454,462 +454,243 @@ macro (ADD_HELP_TEST testname resultcode)
   endif ()
 endmacro ()
 
-macro (ADD_SKIP_H5_TEST skipresultfile skipresultcode testtype)
```
byrnHDF, line 457: Can we retain this SKIP feature as an optional argument?

mattjala: I can add it back in as an argument, but what's the motivation for including it?

byrnHDF: The ability to use a regex at the calling command to skip tests based on command line args/options specific to the test. Yet still indicate that the tests were not run because of user choice. Yes there are other methods, but this would supply another option for the user. Probably wouldn't have noticed it except it was there before. Minor importance.

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -454,462 +454,255 @@ macro (ADD_HELP_TEST testname resultcode)
...
+  # Certain args are fully incompatible with memchecker; skip in these cases
+  set(should_add_test TRUE)
+
+  if (HDF5_ENABLE_USING_MEMCHECKER)
+    if (DEFINED ARG_H5ERRREF OR ${ARG_BINFILE} OR ${ARG_GREP_COMPARE})
+      set(should_add_test FALSE)
+    endif()
+  endif()
+
+  if (${ARG_SKIP_TEST})
```
byrnHDF, line 604: I would have thought this would be first along with a quick exit. Also where is the output?
```
    if (NOT HDF5_USING_ANALYSIS_TOOL)
      add_test (
          NAME H5DUMP-${skipresultfile}
          COMMAND ${CMAKE_COMMAND} -E echo "SKIP ${skipresultfile} ${ARGN}"
      )
      set_property(TEST H5DUMP-${skipresultfile} PROPERTY DISABLED true)
    endif ()

```

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -951,7 +744,7 @@ macro (ADD_H5_TEST_IMPORT conffile resultfile testfile resultcode)
     if ("H5DUMP-IMPORT-h5import-${resultfile}" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")
       set_tests_properties (H5DUMP-IMPORT-h5import-${resultfile} PROPERTIES DISABLED true)
     endif ()
-    add_test (NAME H5DUMP-IMPORT-h5diff-${resultfile} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diff> ${testfile} ${resultfile}.h5 /integer /integer)
+    add_test (NAME H5DUMP-IMPORT- -${resultfile} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5diff> ${testfile} ${resultfile}.h5 /integer /integer)
```
byrnHDF: Test name has a "- -" blank where h5diff was

byrnHDF (changes requested):
Test name has a "- -" blank where h5diff was

tools/test/h5dump/CMakeTests.cmake
mattjala: As far as I'm aware, there is no way to 'quick exit' from a macro. It's delayed in order to come after the processing of various other arg combinations that can also cause it to be skipped (memchecker + errref, memchecker + binfile, ...)

byrnHDF: There are two ways, neither great. Easiest is an if/else the else being the rest of the macro.
The second way would be to split the macro up into multiple sub-macro calss. Which given it's complexity might be the best in the long run?

mattjala: I've added a test skip message

mattjala: > There are two ways, neither great. Easiest is an if/else the else being the rest of the macro. The second way would be to split the macro up into multiple sub-macro calss. Which given it's complexity might be the best in the long run?

I don't think there's much to be gained from re-breaking up this macro. The core operations are simple (call h5diff with arguments, compare output) - the complicated part is all the optional arguments. I'm in favor of using if/else like this to handle the skipping.

byrnHDF: Yes, that is what I'd do.

byrnHDF (dismissed):

lrknox (dismissed):

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -454,462 +454,256 @@ macro (ADD_HELP_TEST testname resultcode)
...
+    endif()
+  endif()
+
+  if (${ARG_SKIP_TEST})
+    set(should_skip_test TRUE)
+  endif()
+
+  if (should_skip_test)
+    # Add a test that just displays a skip message instead
+    add_test(
```
jhendersonHDF: Note that while the `ADD_SKIP_H5_TEST` macro wasn't actually used, what it was doing seems different than what we'd want now. Rather, a test should always be added (with its original name and its original command, not a command that just echoes "SKIP ..."), but when `SKIP_TEST` is specified in this new scheme,

```
set_tests_properties (H5DUMP-{ctest_testname} PROPERTIES DISABLED true)
```

should be called to specify that the test is disabled. That way the real test exists, but isn't run and CI can pick up that the test is being skipped when reporting results. It also keeps the test names consistent without having special test names for how a test is being run. This would also cleanup this particular part of the logic by flattening a level of `if ... else`.

byrnHDF (approved):

jhendersonHDF (approved):

PR #5753, opened 2025-08-29
Remove unused args from h5diff test macro

mattjala:


byrnHDF (approved):

bmribler (approved):

lrknox:
@mattjala, the test MPI_TEST_H5DIFF-h5diff_600 failed on attaway at Sandia today with both gcc/12.3.0 and clang/16.0.6 compiler modules.  I think the test is run with just one file and the output is the help message.  The differences between actual and expected output will be available for a few days at https://my.cdash.org/tests/295831470 and https://my.cdash.org/tests/295856234.  It looks like multiple processes are returning the same output, but the REFER lines do also do that, they're just a line or 2 out of sync.  I 

I put the unused args back in, ran it again, and the tests passed, so take a look when you have a chance.  Note this only failed on the one machine on 2 of 3 builds, and there were 2 previous days since the merge where the tests didn't fail anywhere.  I'll watch for any more failures.

PR #5763, opened 2025-09-02
Generate VOL tests for h5dump

mattjala:
Depends on #5761 and #5752. Will be marked ready for review once those are merged

----

> [!IMPORTANT]
> Add VOL tests for h5dump, update test generation logic, and modify test configurations.
> 
>   - **Tests**:
>     - Add VOL tests for `h5dump` in `CMakeTests.cmake` and `CMakeTestsXML.cmake`.
>     - Modify `ADD_H5_TEST` macro to support VOL-specific tests and filtering.
>     - Add `gent_tvms()` function in `h5dumpgentest.c` and `h5dumpgentest.h` for generating test files.
>   - **Configuration**:
>     - Update `.github/workflows/vol_rest.yml` to exclude `H5DUMP` from certain tests.
>     - Add `NATIVE_ONLY` flag to `ADD_H5_TEST` macro for native VOL connector tests.
>   - **Functions**:
>     - Modify `write_dset()` in `h5dumpgentest.c` to accept `tid_memory` parameter for memory datatype.
>     - Update `gent_fcontents()` and `gent_fvalues()` to use new `write_dset()` signature.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 5dfa416979acbb75b0ebe18a9527e54d6d9d4645. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tools/test/h5dump/CMakeTests.cmake
+++ b/tools/test/h5dump/CMakeTests.cmake
@@ -1218,9 +1218,7 @@ ADD_H5_TEST (texceedsubblock RESULT_CODE 1 H5ERRREF "exceed dataset dims" --enab
 
 # tests for filters
 # SZIP
-if (H5_HAVE_FILTER_SZIP)
-  ADD_H5_TEST (tszip RESULT_CODE 0 APPLY_FILTERS 2 --enable-error-stack -H -p -d szip TARGET_FILE tfilters.h5)
-endif()
+ADD_H5_TEST (tszip RESULT_CODE 0 APPLY_FILTERS 2 --enable-error-stack -H -p -d szip TARGET_FILE tfilters.h5)
 
 # deflate
 ADD_H5_TEST (tdeflate RESULT_CODE 0 APPLY_FILTERS 2 --enable-error-stack -H -p -d deflate TARGET_FILE tfilters.h5)
@@ -1238,9 +1236,7 @@ ADD_H5_TEST (tnbit RESULT_CODE 0 APPLY_FILTERS 1 --enable-error-stack -H -p -d n
 ADD_H5_TEST (tscaleoffset RESULT_CODE 0 APPLY_FILTERS 4 --enable-error-stack -H -p -d scaleoffset  TARGET_FILE tfilters.h5)
 
 # all
-if (H5_HAVE_FILTER_SZIP)
-  ADD_H5_TEST (tallfilters RESULT_CODE 0 APPLY_FILTERS 1 --enable-error-stack -H -p -d all  TARGET_FILE tfilters.h5 APPLY_FILTERS 1)
-endif ()
+ADD_H5_TEST (tallfilters RESULT_CODE 0 APPLY_FILTERS 1 --enable-error-stack -H -p -d all  TARGET_FILE tfilters.h5 APPLY_FILTERS 1)
 
 # user defined
 ADD_H5_TEST (tuserfilter RESULT_CODE 0 --enable-error-stack -H  -p -d myfilter  TARGET_FILE tfilters.h5)
--- a/tools/test/h5dump/CMakeTests.cmake
+++ b/tools/test/h5dump/CMakeTests.cmake
@@ -747,9 +747,6 @@ macro (ADD_H5_TEST testname)
       if (${ARG_RESULT_CODE})
         set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES WILL_FAIL "true")
       endif ()
-      set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES
-          WORKING_DIRECTORY "${workdir}"
-      )
     else ()
       add_test (
           NAME ${vol_prefix}H5DUMP-${ctest_testname}
@@ -775,10 +772,6 @@ macro (ADD_H5_TEST testname)
         WORKING_DIRECTORY "${workdir}"
     )
 
-    if ("${vol_prefix}H5DUMP-${ctest_testname}" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")
-      set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES DISABLED true)
-    endif ()
-    
     # Set VOL-specific properties
     if (NOT "${vol}" STREQUAL "native")
       set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES
--- a/tools/test/h5dump/CMakeTests.cmake
+++ b/tools/test/h5dump/CMakeTests.cmake
@@ -1271,10 +1271,8 @@ ADD_H5_TEST (tldouble RESULT_CODE 0 --enable-error-stack TARGET_FILE tldouble.h5
 ADD_H5_TEST (tldouble_scalar RESULT_CODE 0 -p --enable-error-stack TARGET_FILE tldouble_scalar.h5)
 
 # Add tests for _Float16 type
-if (${${HDF_PREFIX}_HAVE__FLOAT16})
-  ADD_H5_TEST (tfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16.h5)
-  ADD_H5_TEST (tfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16_be.h5)
-endif()
+ADD_H5_TEST (tfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16.h5)
+ADD_H5_TEST (tfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16_be.h5)
 
 # Add tests for bfloat16 type
 ADD_H5_TEST (tbfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16.h5)
--- a/tools/test/h5dump/CMakeTests.cmake
+++ b/tools/test/h5dump/CMakeTests.cmake
@@ -1283,8 +1283,8 @@ ADD_H5_TEST (tfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16.h5
 ADD_H5_TEST (tfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16_be.h5 ${F16_NATIVE_ONLY})
 
 # Add tests for bfloat16 type
-ADD_H5_TEST (tbfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16.h5 ${F16_NATIVE_ONLY})
-ADD_H5_TEST (tbfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16_be.h5 ${F16_NATIVE_ONLY})
+ADD_H5_TEST (tbfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16.h5)
+ADD_H5_TEST (tbfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16_be.h5)
 
 # Add tests for complex numbers. For portability, use a fixed floating-point
 # precision and skip dumping of the "long double _Complex" dataset. The "long
--- a/tools/test/h5dump/CMakeTests.cmake
+++ b/tools/test/h5dump/CMakeTests.cmake
@@ -16,6 +16,15 @@
 ##############################################################################
 ##############################################################################
 
+include(${HDF_CONFIG_DIR}/HDF5Macros.cmake)
+
+# System-independent path separator
+if (WIN32)
+  set (CMAKE_SEP "\;")
+else ()
+  set (CMAKE_SEP ":")
+endif ()
+
 # --------------------------------------------------------------------
 # Copy all the HDF5 files from the test directory into the source directory
 # --------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


tools/test/h5dump/CMakeTestsXML.cmake
```diff
@@ -299,8 +299,10 @@ ADD_XML_H5_TEST (torderattr3.h5 0 -H --sort_by=creation_order --sort_order=ascen
 ADD_XML_H5_TEST (torderattr4.h5 0 -H --sort_by=creation_order --sort_order=descending torderattr.h5)
 
 # Add tests for _Float16 type
-ADD_XML_H5_TEST (tfloat16.h5 0 tfloat16.h5)
-ADD_XML_H5_TEST (tfloat16_be.h5 0 tfloat16_be.h5)
+if (${${HDF_PREFIX}_HAVE__FLOAT16})
```
jhendersonHDF: Were you seeing issues with testing of these files? Generally they should be tested even when support for `_Float16` isn't available (or not enabled), the same as the regular h5dump tests in CMakeTests.cmake, as the output should be the same either way.

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -5071,21 +5073,25 @@ write_attr(hid_t loc_id, int rank, hsize_t *dims, const char *attr_name, hid_t t
...
 static int
-write_dset(hid_t loc_id, int rank, hsize_t *dims, const char *dset_name, hid_t tid, void *buf)
+write_dset(hid_t loc_id, int rank, hsize_t *dims, const char *dset_name, hid_t tid_dset, hid_t tid_memory,
+           void *buf)
 {
     hid_t  did;
     hid_t  sid;
     herr_t status;
 
+    if (tid_memory == H5I_INVALID_HID)
```
jhendersonHDF: Minor really, but it may make more sense just to pass the memory type ID directly (duplicated in most cases) instead of special casing H5I_INVALID_HID in an unusual way.

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -1086,7 +1214,9 @@ ADD_H5_TEST (texceedsubblock RESULT_CODE 1 H5ERRREF "exceed dataset dims" --enab
 
 # tests for filters
 # SZIP
-ADD_H5_TEST (tszip RESULT_CODE 0 APPLY_FILTERS 2 --enable-error-stack -H -p -d szip TARGET_FILE tfilters.h5)
+if (H5_HAVE_FILTER_SZIP)
```
jhendersonHDF: Similar to the other comment about the _Float16 tests, including that section below, generally these tests are meant to be run all the time, even if support for the particular feature isn't available. With the `-H` flag, this shouldn't need the szip filter available as the data won't be printed. Maybe there's an argument parsing issue here?

tools/test/h5dump/CMakeTestsXML.cmake
mattjala: This was something unnecessary that I left in during development - removed.

tools/test/h5dump/CMakeTests.cmake
mattjala: This was another check left over from development - removed

brtnfld (dismissed):

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -507,11 +563,33 @@ macro (ADD_H5_TEST testname)
...
-    set (filters_out "SIZE XXXX (${ARG_APPLY_FILTERS}.XXX:1 COMPRESSION)")
+    set (filters_in "")
+    set (filters_out "")
+
+    list (APPEND filters_in "SIZE [0-9]* \\(${_FILTER_VAL}\\\.[0-9][0-9][0-9]:1 COMPRESSION\\)")
+    list (APPEND filters_out "SIZE XXXX (${_FILTER_VAL}.XXX:1 COMPRESSION)")
+
+    # Mask out the h5dump-assigned anonymous committed datatype numbers
+    # Datatype number may have a forward slash before #
+    list (APPEND filters_in "DATATYPE[ \t]*\"(/?#)[0-9]+\"")
```
jhendersonHDF, line 584: I'm somewhat iffy on this change, only because applying filters to the output for the native connector has the potential to mask real issues.

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -605,101 +683,151 @@ macro (ADD_H5_TEST testname)
...
 
-    set_tests_properties (H5DUMP-${ctest_testname}-clear-objects PROPERTIES
-        WORKING_DIRECTORY "${PROJECT_BINARY_DIR}/testfiles/std"
+    # If using memchecker add tests without using scripts
+    if (HDF5_ENABLE_USING_MEMCHECKER AND NOT ARG_MASK_ERROR AND NOT ARG_GREP_COMPARE AND NOT DEFINED ARG_H5ERRREF)
+      add_test (NAME ${vol_prefix}H5DUMP-${ctest_testname} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:h5dump> ${ARG_ANY_PATHS} ${ARG_UNPARSED_ARGUMENTS} ${ARG_DDL_FILE_CMD} ${BINARY_OUTPUT_FLAG} ${ARG_OUTPUT_FILEARGS} ${ARG_TARGET_FILE})
+      if (${ARG_RESULT_CODE})
+        set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES WILL_FAIL "true")
+      endif ()
+      set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES
```
jhendersonHDF: Looks to be duplicated below

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -605,101 +683,151 @@ macro (ADD_H5_TEST testname)
...
-  endif ()
-  set_tests_properties (H5DUMP-${ctest_testname} PROPERTIES
-      WORKING_DIRECTORY "${PROJECT_BINARY_DIR}/testfiles/std"
-  )
 
-  if ("H5DUMP-${ctest_testname}" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")
-    set_tests_properties (H5DUMP-${ctest_testname} PROPERTIES DISABLED true)
-  endif ()
+    if ("${vol_prefix}H5DUMP-${ctest_testname}" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")
+      set_tests_properties (${vol_prefix}H5DUMP-${ctest_testname} PROPERTIES DISABLED true)
```
jhendersonHDF, line 784: Looks to be duplicated either above or here, depending on order desired; keeping either one works

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -1154,8 +1274,10 @@ ADD_H5_TEST (tldouble RESULT_CODE 0 --enable-error-stack TARGET_FILE tldouble.h5
 ADD_H5_TEST (tldouble_scalar RESULT_CODE 0 -p --enable-error-stack TARGET_FILE tldouble_scalar.h5)
 
 # Add tests for _Float16 type
-ADD_H5_TEST (tfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16.h5)
-ADD_H5_TEST (tfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16_be.h5)
+if (${${HDF_PREFIX}_HAVE__FLOAT16})
```
jhendersonHDF, line 1276: Another instance of "HAVE__FLOAT16" check

jhendersonHDF:
Generally looks good, with just a few comments

tools/test/h5dump/CMakeTests.cmake
mattjala: I agree it's not a perfect solution, but given that verification is done through h5dump output containing these potentially inconsistent elements, I'm not aware of a cleaner solution. The way the masking is done leaves the files on-disk the same, so those can at least be checked directly to see e.g. real sizes and offsets. Only a few specifically marked tests apply these filters.

jhendersonHDF: My issue with this one is more about catching the issues at test time. Sure someone can always manually inspect the files, but they would need to know there is something wrong in the first place before deciding to do that. 

The idea is that at least for the native VOL, the address/size/etc. of something changing in the file would be an interesting point to note. The test failure that would result could signal to the developer that something is wrong in the tools, or at least signal that files should be re-generated. This probably isn't too likely in practice, but it would be nice if we could avoid applying these filters for the native VOL.

jhendersonHDF (dismissed):
Approving as the masking of some values for the native VOL is a fairly minor issue, though it would be nice if it could be resolved. There's also a small conflict in the REST VOL workflow that needs resolved.

tools/test/h5gentest.c
```diff
@@ -275,7 +275,7 @@ gen_h5dump_files(void)
     gent_trefer_reg_1d();
 
     nerrors += gent_test_reference_external();
-
+    nerrors += (gent_tvms() < 0 ? 1 : 0);
```
brtnfld, line 281: What does vms stand for in this context?

mattjala: I'm not sure - tvms.h5 was a pre-existing test file that I just added a generator function to replicate. Since it uses the H5T_VAX_F64 datatype, maybe "Vax Memory Syntax"?

jhendersonHDF: Most likely "Virtual Memory System" from VAX/VMS

jhendersonHDF (dismissed):

brtnfld (commented):
Critical Issues: None Identified

  No major bugs, security vulnerabilities, or breaking changes were found. The code appears to be defensive and follows HDF5's error handling patterns.

  Code Quality Issues: Medium Priority

  1. Inconsistent Function Signature: The write_dset function signature change from 6 to 7 parameters (adding separate dataset and memory datatypes) lacks documentation explaining the
  rationale:
  // Old: write_dset(loc_id, rank, dims, dset_name, tid, buf)
  // New: write_dset(loc_id, rank, dims, dset_name, tid_dset, tid_mem, buf)
  2. Missing Error Handling: In CMakeTests.cmake:1014, the complex VOL connector test generation logic lacks proper error handling for file operations and path manipulations.
  3. Code Duplication: The ADD_H5_TEST macro modifications introduce repeated logic for handling different VOL connectors that could be abstracted into helper functions.
  4. Missing Function Documentation: The new gent_tvms() function lacks documentation explaining its purpose (generating VAX F64 datatype test files).

  Enhancement Opportunities: Low Priority

  1. Performance Optimization: The CMake test generation could be optimized by caching file system operations and reducing redundant path calculations.
  2. Maintainability Improvements:
    - Extract the path separator logic into a dedicated CMake function
    - Add inline comments explaining the VOL connector test matrix approach
    - Consider using CMake generator expressions for cleaner conditional logic
  3. Test Coverage: The PR focuses on h5dump VOL testing but doesn't include corresponding tests for error conditions or edge cases specific to different VOL connectors.
  4. Configuration Management: The test configuration could benefit from a centralized VOL connector configuration file rather than inline conditional logic.

  Specific Technical Recommendations:

  1. Line CMakeTests.cmake:1014: Consider extracting the complex path manipulation into a helper function:
  function(normalize_path_separators input_path output_var)
    # Implementation here
  endfunction()
  2. File h5dumpgentest.c: Add documentation for gent_tvms():
  /*
   * Function: gent_tvms
   * Purpose:  Generate test file with VAX F64 datatype for VOL connector testing
   * Return:   Success: 0, Failure: -1
   */
  3. File CMakeTests.cmake: Add error checking for file operations:
  if(NOT EXISTS "${source_file}")
    message(FATAL_ERROR "Required test file ${source_file} not found")
  endif()

For example, Recommendaton 1

brtnfld (dismissed):

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -118,18 +118,23 @@
 #define FILE95 "tcomplex.h5"
 #define FILE96 "tcomplex_be.h5"
 #endif
-#define FILE97  "tbfloat16.h5"
```
jhendersonHDF, line 121: While it can be annoying to resolve git conflicts around, we should either keep the original naming/numbering scheme here, or convert the existing macros to new names, along with all the associated function-specific value macros. Otherwise, it's easier for the file name and other macros to go out of sync.

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -118,18 +118,23 @@
...
+#define FILE_REFER_OBJ_DEL "trefer_obj_del.h5"
+#define FILE_REFER_OBJ     "trefer_obj.h5"
+#define FILE_REFER_PARAM   "trefer_param.h5"
+#define FILE_REFER_REG     "trefer_reg.h5"
+#define FILE_REFER_REG_1D  "trefer_reg_1d.h5"
+
+#define FILE_REFER_ATTR "trefer_attr.h5"
+#define FILE_ATTR4_BE   "tattr4_be.h5"
+#define FILE_NO_SUBSET  "tno-subset.h5"
+#define F99_DSET        "AHFINDERDIRECT::ah_centroid_t[0] it=0 tl=0"
```
jhendersonHDF: This is now `F101_DSET` (see below) after the bfloat16 changes

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -1158,12 +1271,20 @@ ADD_H5_TEST (tldouble RESULT_CODE 0 --enable-error-stack TARGET_FILE tldouble.h5
...
+  set (F16_NATIVE_ONLY "NATIVE_ONLY")
+endif ()
+
+ADD_H5_TEST (tfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16.h5 ${F16_NATIVE_ONLY})
+ADD_H5_TEST (tfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tfloat16_be.h5 ${F16_NATIVE_ONLY})
 
 # Add tests for bfloat16 type
-ADD_H5_TEST (tbfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16.h5)
-ADD_H5_TEST (tbfloat16_be RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16_be.h5)
+ADD_H5_TEST (tbfloat16 RESULT_CODE 0 --enable-error-stack TARGET_FILE tbfloat16.h5 ${F16_NATIVE_ONLY})
```
jhendersonHDF: These types don't need to be conditional on float16 support as they're distinct types

jhendersonHDF (approved):

brtnfld (approved):

PR #5794, opened 2025-09-08
Revise approach to Unicode filenames on Windows

### LINKED ISSUE #5788: Revise approach to handling unicode filenames on Windows
The handling of Unicode filenames on Windows should be revised to first try converting
filenames to UTF-16 and passing them to `_wopen()`, with a fallback to `_open()` if that
fails. An override mechanism should also be implemented to allow users to prefer `_open()`
to handle ambiguity between UTF-8 and Windows code pages.

<b>Context:</b>
Prior to the HDF5 1.10.6 release, HDF5 opened files on Windows by passing the filename
to the POSIX `open()` function, an alias for Windows' `_open()` function, which interprets
filenames in "ANSI" format according to the current code page. This approach simply
couldn't support Unicode filenames in general.

In the HDF5 1.10.6 release, this was changed in commit 750b5c293076b6a446088fa3020e4e0787d489d7
so that filenames are first converted to UTF-16 using Windows' `MultiByteToWideChar()`
function and then are passed to the `_wopen()` function which supports UTF-16 through
the `wchar_t` type. This allows the creation/opening of files with UTF-8 filenames, but
has the drawback that a user must know to encode filenames as UTF-8 before passing
them to an H5F API, while also knowing that the library is going to convert them (potentially
back again) to UTF-16. Another drawback with the specific implementation of this approach
was that there was no fallback for non-UTF-8 filenames. For certain filenames, or if the
`MultiByteToWideChar()` function failed, the library would simply fail to create/open the
file. This was reported on in https://forum.hdfgroup.org/t/open-create-hdf5-files-with-non-utf8-chars-such-as-shift-jis/11785, where Japanese characters were problematic.

In the HDF5 1.14.4 release, this was changed again in commit eb0351efffe987a9c1882ccc9b03b5b0aec7f2dd
so that an attempt to open the file with the POSIX `open()` function is first
made. If that fails, a fallback path converts the filename to UTF-16 and passes it to the
`_wopen()` function as before. This was a quick attempt at trying to solve both cases in
relation to the HDF Forum post. However, this solution was the reverse of the suggestion
in that post and had the effect of breaking the interpretation of filenames similar to the
situation before the 1.10.6 release.

<b>Related:</b>
https://github.com/HDFGroup/hdf5/issues/5037

jhendersonHDF:
Treat filenames as UTF-8 initially and fall back to ANSI functions if conversion to UTF-16 fails

Adds new HDF5_PREFER_WINDOWS_CODE_PAGE environment variable to prefer interpreting filenames according to the active Windows code page rather than assuming UTF-8 encoding

Fixes #5788

----

> [!IMPORTANT]
> Revise Unicode filename handling on Windows to prefer UTF-8 and add environment variable for code page preference.
> 
>   - **Behavior**:
>     - Treat filenames as UTF-8 initially and fallback to ANSI if UTF-16 conversion fails in `H5system.c`.
>     - Introduces `HDF5_PREFER_WINDOWS_CODE_PAGE` environment variable in `H5public.h` to prefer Windows code page over UTF-8.
>   - **Functions**:
>     - Modify `H5_get_utf16_str()` in `H5system.c` to return `herr_t` and handle errors.
>     - Update `Wopen()` and `Wremove()` in `H5system.c` to use the new environment variable.
>   - **Tests**:
>     - Add tests in `tunicode.c` for UTF-8 filename handling, conversion failures, and code page override.
>   - **Misc**:
>     - Add `unicode_filename_note` alias in `doxygen/aliases`.
>     - Update `RELEASE.txt` with changes to Unicode filename handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 6c484470eb6a5629270e19359ef69a4bdc8c8cce. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


doxygen/aliases
```diff
@@ -254,6 +254,7 @@ ALIASES += cpp_c_api_note="\anchor cpp_c_api_note \attention \Bold{C++ Developer
...
 ALIASES += sa_metadata_ops="\sa \li H5Pget_all_coll_metadata_ops() \li H5Pget_coll_metadata_write() \li H5Pset_all_coll_metadata_ops() \li H5Pset_coll_metadata_write() \li \ref maybe_metadata_reads"
+ALIASES += unicode_filename_note="\note On Windows, HDF5 assumes that a file name string is UTF-8 encoded and will attempt to convert it to UTF-16 before calling wide-character Windows API functions. If a file name string cannot be converted to UTF-16, the equivalent non-wide-character Windows API functions will be used, causing a file name string to be interpreted according to the active Windows code page. If an application desires that the active Windows code page be preferred, the environment variable #HDF5_PREFER_WINDOWS_CODE_PAGE can be set to the value '1' or 'TRUE' (case-insensitive)."
```
jhendersonHDF, line 257: Added documentation to H5F APIs to specify that strings should be UTF-8 encoded and that this can be overridden with the environment variable

src/H5Fpublic.h
```diff
@@ -269,9 +269,15 @@ extern "C" {
  *          container_name can be opened with the file access property list
  *          \p fapl_id.
  *
+ * \parblock
  * \note The H5Fis_accessible() function enables files to be checked with a
  *       given file access property list, unlike H5Fis_hdf5(), which only uses
  *       the default file driver when opening a file.
+ * \endparblock
+ *
+ * \parblock
```
jhendersonHDF, line 278: Surrounded multiple note sections in their own `\parblock` / `\endparblock` sections so that doxygen doesn't merge them together into one

src/H5private.h
```diff
@@ -900,12 +900,13 @@ H5_DLL int HDvasprintf(char **bufp, const char *fmt, va_list _ap);
...
-#define H5_DIR_SEPS              "\\"
-#define H5_CHECK_DELIMITER(SS)   ((SS == H5_DIR_SEPC) || (SS == H5_DIR_SLASH_SEPC))
-#define H5_CHECK_ABSOLUTE(NAME)  ((isalpha(NAME[0])) && (NAME[1] == ':') && (H5_CHECK_DELIMITER(NAME[2])))
-#define H5_CHECK_ABS_DRIVE(NAME) ((isalpha(NAME[0])) && (NAME[1] == ':'))
+#define H5_DIR_SLASH_SEPC      '/'
+#define H5_DIR_SEPC            '\\'
+#define H5_DIR_SEPS            "\\"
+#define H5_CHECK_DELIMITER(SS) ((SS == H5_DIR_SEPC) || (SS == H5_DIR_SLASH_SEPC))
+#define H5_CHECK_ABSOLUTE(NAME)                                                                              \
+    ((isalpha((unsigned char)NAME[0])) && (NAME[1] == ':') && (H5_CHECK_DELIMITER(NAME[2])))
```
jhendersonHDF, line 908: Added cast to `(unsigned char)` to fix an issue where certain characters passed to `isalpha()` would be outside the acceptable range and cause a Debug assertion with MSVC.

src/H5system.c
```diff
@@ -447,43 +447,57 @@ Wflock(int fd, int operation)
...
-H5_get_utf16_str(const char *s)
+herr_t
+H5_get_utf16_str(const char *s, wchar_t **wstring, uint32_t *win_error)
 {
     int      nwchars = -1;   /* Length of the UTF-16 buffer */
     wchar_t *ret_s   = NULL; /* UTF-16 version of the string */
 
     /* Get the number of UTF-16 characters needed */
-    if (0 == (nwchars = MultiByteToWideChar(CP_UTF8, 0, s, -1, NULL, 0)))
+    if (0 == (nwchars = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, s, -1, NULL, 0)))
```
jhendersonHDF, line 468: Add this flag to make `MultiByteToWideChar` fail for invalid UTF-8 characters

test/tunicode.c
```diff
@@ -36,26 +36,436 @@
...
+ * A test to check that the library falls back to code page
+ * "ANSI" functions on Windows when a filename cannot be
+ * converted using MultiByteToWideChar().
+ */
+static void
+test_utf8_conv_failure(void)
+{
+#ifdef H5_HAVE_WIN32_API
+    /* Byte sequence for string 'ハローワールド' or 'Hello World' in Shift JIS. (NOTE: invalid UTF-8) */
+    const char *filename = "\x83\x6E\x83\x8D\x81\x5B\x83\x8F\x81\x5B\x83\x8B\x83\x68"
```
jhendersonHDF, line 245: Used a byte sequence string here so it's independent of the file encoding used

test/tunicode.c
```diff
@@ -796,68 +1207,20 @@ dump_string(const char *string)
...
-/* Main test.
- * Create a string of random Unicode characters, then run each test with
- * that string.
+/*
+ * Main test - test various aspects of unicode support
  */
 void
 test_unicode(void H5_ATTR_UNUSED *params)
 {
-    char         test_string[MAX_STRING_LENGTH];
```
jhendersonHDF, line 807: All this logic was moved to `test_random_utf8_strings()`

brtnfld (approved):
LGTM

byrnHDF (approved):

test/tunicode.c
```diff
@@ -36,26 +36,436 @@
...
+ *
+ * The filenames used for this test are simple examples taken from
+ * a reported regression in the 1.14.4 and 1.14.5 releases.
+ */
+static void
+test_utf8_filenames(void)
+{
+    const char *euro_filename = u8"€.h5";
+    const char *quot_filename = u8"‚.h5"; /* U+201A "Single Low-9 Quotation Mark" */
+    const char *chin_filename = u8"漢字.h5";
```
derobins, line 73: When I originally worked on this feature, I wanted to independently create a test file with a name that used Unicode characters and store it in the repo, so it could be opened later from a test program. That avoids false positive results when the library mis-handles Unicode and creates a gibberish.h5 file and then opens the gibberish.h5 file it created.

I ran into problems, however, when CMake couldn't copy Unicode filenames around, possibly due to missing byte-order marks in the CMake text files.

jhendersonHDF: For this particular test, I used the u8 prefix to ensure that the strings are always encoded as UTF-8 (so will always represent "€", for example) and then made sure to convert the names to UTF-16 and try to access them through the wide-character APIs. This should make sure that if the file names were mistakenly created according to the code page, they won't be accessible with the original name. I could add a check that they aren't accessible through the non-wide-character `access`, I just avoided it for this test because that can fail if the active code page is UTF-8 or specific other ones and the check felt redundant.

For the other tests, I used byte sequence strings to make sure that the filename would be the exact sequence of bytes I expected for the encoding I expected (for example, the Japanese string as encoded in Shift JIS), regardless of the platform, file encoding, etc. Then I check both or a single version of the `access` functions, with some skipping as appropriate based on the active code page.

Having a test file in the repo could be useful, though from all the testing I did on English Windows, English Windows with UTF-8 encoding used everywhere (not the default) and Japanese Windows, this seemed mostly unnecessary and sidesteps any issues CMake might have. Using byte sequence strings also made it so that we don't necessarily need to test on a Japanese Windows machine to catch the relevant issue with Shift JIS that was reported.

derobins: I think that's fine for now, but I'd still feel better about having pre-generated files at some point in the future.

PR #5800, opened 2025-09-10
Add internal spatial tree

mattjala:
Implement an r-tree data structure in a new module. It has three exposed methods: creation, destruction, and search. 

The STR algorithm used during creation is based on the one described [here](https://archive.org/details/nasa_techdoc_19970016975/page/n9/).

This will be used to optimize VDS operations in an upcoming PR.

----

> [!IMPORTANT]
> Introduces an R-tree data structure with creation, search, and destruction methods, integrated into the build system and validated with new tests.
> 
>   - **R-tree Implementation**:
>     - Adds new R-tree data structure in `H5RT.c` with methods `H5RT_create()`, `H5RT_search()`, and `H5RT_free()`.
>     - Implements STR algorithm for efficient R-tree packing.
>   - **CMake Integration**:
>     - Updates `CMakeLists.txt` to include `H5RT.c` and related headers.
>   - **Testing**:
>     - Adds `rtree.c` test file to validate R-tree creation and search functionalities.
>     - Updates `test/CMakeLists.txt` to include `rtree` in test suite.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 1b8ba5888568284bcf06ea26ad8de35c5dc480e9. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


byrnHDF (dismissed):
CMake changes looks good.

src/H5RTprivate.h
```diff
@@ -0,0 +1,50 @@
...
+#define H5RTprivate_H
+
+#include "H5private.h"
+
+/* The overall r-tree structure */
+typedef struct H5RT_t H5RT_t;
+
+/* Struct representing a leaf in the r-tree */
+typedef struct H5RT_leaf_t {
+    uintptr_t record;
```
fortnern: Why uintptr_t instead of just void *?

src/H5RT.c
```diff
@@ -0,0 +1,700 @@
...
+} /* end H5RT_leaf_cleanup() */
+
+bool
+H5RT__leaves_intersect(int rank, hsize_t min1[], hsize_t max1[], hsize_t min2[], hsize_t max2[])
+{
+    FUNC_ENTER_PACKAGE_NOERR
+
+    for (int i = 0; i < rank; i++)
+        if (min1[i] > max2[i] || min2[i] > max1[i])
+            return false; /* No overlap in i-th dimension */
```
fortnern: I don't think it's appropriate to bypass FUNC_ENTER/LEAVE with a return here. I would hope modern compilers would handle HGOTO_DONE(false) just as efficiently.

fortnern: You can maybe add inline to the function declaration

src/H5RT.c
```diff
@@ -0,0 +1,700 @@
...
+    assert(dim);
+
+    l1       = (const H5RT_leaf_t *)leaf1;
+    l2       = (const H5RT_leaf_t *)leaf2;
+    sort_dim = *(int *)dim;
+
+    FUNC_ENTER_PACKAGE_NOERR
+
+    /* Compare based on the midpoint of the specified dimension */
+    if (l1->mid[sort_dim] < l2->mid[sort_dim]) {
```
fortnern: You should probably add an assertion that sort_dim is within the bounds of the array here

src/H5RT.c
```diff
@@ -0,0 +1,700 @@
...
+        new_leaves[i]._coords = coords;
+        new_leaves[i].min     = coords;
+        new_leaves[i].max     = coords + src_leaf->rank;
+        new_leaves[i].mid     = coords + (2 * src_leaf->rank);
+
+        /* Copy coordinate data */
+        memcpy(coords, src_leaf->_coords, 3 * (size_t)src_leaf->rank * sizeof(hsize_t));
+    }
+
+    if ((new_tree = H5RT_create(rtree->rank, new_leaves, rtree->nleaves)) == NULL)
```
fortnern: Could you just deep copy the internal nodes here instead of recreating the tree? Is this location not likely to be important for performance?

src/H5RT.c
```diff
@@ -0,0 +1,700 @@
...
+        node->children.leaves     = leaves;
+    }
+    else {
+        /* Recursive case - there will be child nodes */
+        node->children_are_leaves = false;
+
+        /* Sort hyper-rectangles in this region by the first unsorted coordinate of their midpoints */
+        if (!this_rank_sorted) {
+            assert(prev_sort_dim < rank - 1);
+            sort_dim = prev_sort_dim + 1;
```
fortnern, line 396: So it just sorts dimensions in the order 0,1,2, 3,3,3,3 (for rank 4)?

src/H5RT.c
```diff
@@ -0,0 +1,700 @@
...
+    if (rank < 1 || rank > H5S_MAX_RANK)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "invalid rank");
+    if (!leaves)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "invalid leaves");
+    if (count == 0)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "must have at least one leaf");
+    if (prev_sort_dim < -1)
+        HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "invalid previous sort dimension");
+
+    this_rank_sorted = (prev_sort_dim == (rank - 1));
```
fortnern: This is slightly odd in terms of logic flow - we're saying this dim is sorted before we decide which dim we're operating on. Also rank usually refers to the total number of dimensions not a specific dimension. We could make the code clearer by changing this variable name to maybe something like "use_prev_sort_dim"

src/H5RT.c
fortnern: Would it be hard to use something simple like picking the largest dimension? I'm wondering how much this would help performance

src/H5RT.c
```diff
@@ -0,0 +1,700 @@
...
+        ret_value = 1;
+    }
+    else {
+        ret_value = 0;
+    }
+
+    FUNC_LEAVE_NOAPI(ret_value)
+}
+
+static herr_t
```
fortnern, line 173: Could you create a description for this function?

fortnern: In general can we use the standard function header comment pattern used in the rest of the library?

`/*-------------------------------------------------------------------------
 * Function:    
 *
 * Purpose:     
 *
 * Return:      
 *
 */`

src/H5RT.c
mattjala: Your understanding is correct - it just iterates through the dimensions. We could try picking the largest dimension, though this would require manually keeping tracking of which dimensions have been sorted and which haven't. I'll try it and see if it has an impact.

src/H5RT.c
mattjala: This is meant to check whether the provided prev_sort_dim values indicates that we've already sorted all the dims - and as such, that this dim must already be sorted. I agree this is somewhat confusing and will update it.

src/H5RT.c
mattjala: It's not performance critical, but it won't take long to rework this.

src/H5RT.c
mattjala: What exactly do you mean by largest dimension? I tried an implementation where the dimension in the bbox with the highest span (max - min) was prioritized, but didn't find any improvement in performance.

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+            for (size_t i = 0; i < rtree->nleaves; i++) {
+                if (new_leaves[i]._coords)
+                    free(new_leaves[i]._coords);
+            }
+            free(new_leaves);
+        }
+    }
+
+    FUNC_LEAVE_NOAPI(ret_value);
+}
```
fortnern:
```suggestion
} /* end H5RT_copy() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+            free(rtree->leaves[i]._coords);
+    }
+
+    /* Free the leaves array */
+    free(rtree->leaves);
+    H5FL_FREE(H5RT_t, rtree);
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value);
+}
```
fortnern:
```suggestion
} /* end H5RT_free() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+    if (!node->children_are_leaves)
+        for (int i = 0; i < node->nchildren; i++) {
+            if (node->children.nodes[i]) {
+                H5RT__free_recurse(node->children.nodes[i]);
+                H5FL_FREE(H5RT_node_t, node->children.nodes[i]);
+            }
+        }
+
+    FUNC_LEAVE_NOAPI_VOID
+}
```
fortnern:
```suggestion
} /* end H5RT__free_recurse() */
```

src/H5RTmodule.h
```diff
@@ -0,0 +1,29 @@
...
+
+/* Define the proper control macros for the generic FUNC_ENTER/LEAVE and error
+ *      reporting macros.
+ */
+#define H5RT_MODULE
+// TODO
+#define H5_MY_PKG_ERR  H5E_RTREE
+#define H5_MY_PKG_INIT NO
+
+#endif /* H5RTmodule_H */
```
fortnern:
```suggestion
#endif /* H5RTmodule_H */

```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+                H5FL_FREE(H5RT_node_t, dest_node->children.nodes[i]);
+                dest_node->children.nodes[i] = NULL;
+            }
+        }
+    }
+
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5RT__node_copy() */
+
+static void
```
fortnern, line 726: Can you add a function description here?

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+
+    if (result_set && result_set->results) {
+        free(result_set->results);
+        result_set->results  = NULL;
+        result_set->capacity = 0;
+        result_set->count    = 0;
+    }
+
+    FUNC_LEAVE_NOAPI_VOID
+}
```
fortnern:
```suggestion
} /* end H5RT__result_set_cleanup() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+            HGOTO_ERROR(H5E_INTERNAL, H5E_CANTALLOC, FAIL, "failed to grow result buffer");
+    }
+
+    /* Add the new result */
+    result_set->results[result_set->count] = leaf;
+    result_set->count++;
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5RT__result_set_add() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+    new_results = (H5RT_leaf_t **)realloc(result_set->results, new_capacity * sizeof(H5RT_leaf_t *));
+    if (!new_results)
+        HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "failed to grow result buffer");
+
+    result_set->results  = new_results;
+    result_set->capacity = new_capacity;
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5RT__result_set_grow() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+    result_set->count    = 0;
+
+    /* Allocate initial buffer */
+    result_set->results = (H5RT_leaf_t **)malloc(result_set->capacity * sizeof(H5RT_leaf_t *));
+    if (!result_set->results)
+        HGOTO_ERROR(H5E_RESOURCE, H5E_CANTALLOC, FAIL, "failed to allocate result buffer");
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5RT__result_set_init() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+
+    assert(sort_dim <= l1->rank - 1);
+
+    /* Compare based on the midpoint of the specified dimension */
+    if (l1->mid[sort_dim] < l2->mid[sort_dim])
+        return -1;
+    if (l1->mid[sort_dim] > l2->mid[sort_dim])
+        return 1;
+    return 0;
+}
```
fortnern:
```suggestion
} /* end H5RT__leaf_compare() */
```

src/H5RT.c
```diff
@@ -0,0 +1,895 @@
...
+
+    assert(num_slabs > 0);
+    assert(num_slabs <= node_capacity);
+done:
+    if (ret_value == SUCCEED) {
+        *slab_count_out = num_slabs;
+        *slab_size_out  = slab_size;
+    }
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5RT__compute_slabs() */
```

src/H5RTprivate.h
```diff
@@ -0,0 +1,51 @@
...
+H5_DLL herr_t H5RT_leaf_init(H5RT_leaf_t *leaf, int rank, void *record);
+H5_DLL herr_t H5RT_leaf_cleanup(H5RT_leaf_t *leaf);
+
+/* Main R-tree functions */
+H5_DLL H5RT_t *H5RT_create(int rank, H5RT_leaf_t *leaves, size_t count);
+H5_DLL herr_t  H5RT_search(H5RT_t *rtree, hsize_t min[], hsize_t max[], H5RT_result_set_t **results_out);
+H5_DLL herr_t  H5RT_free_results(H5RT_result_set_t *results);
+H5_DLL herr_t  H5RT_free(H5RT_t *rtree);
+H5_DLL H5RT_t *H5RT_copy(const H5RT_t *rtree);
+#endif /* H5RTprivate_H */
```
fortnern:
```suggestion
#endif /* H5RTprivate_H */

```

test/rtree.c
```diff
@@ -0,0 +1,451 @@
...
+    if (nerrors)
+        goto error;
+
+    printf("All R-tree tests passed.\n");
+    return EXIT_SUCCESS;
+
+error:
+    printf("***** R-TREE TESTS FAILED *****\n");
+    return EXIT_FAILURE;
+}
```
fortnern:
```suggestion
}

```

fortnern:
Looks good other than formatting changes and one function description.

src/H5RTmodule.h
```diff
@@ -0,0 +1,29 @@
...
+ *          is part of the H5RT package.
+ */
+#ifndef H5RTmodule_H
+#define H5RTmodule_H
+
+/* Define the proper control macros for the generic FUNC_ENTER/LEAVE and error
+ *      reporting macros.
+ */
+#define H5RT_MODULE
+// TODO
```
jhendersonHDF: Is this TODO still needed?

src/H5RT.c
```diff
@@ -0,0 +1,905 @@
...
+                              const H5RT_leaf_t *old_leaves_base, H5RT_leaf_t *new_leaves_base);
+static void   H5RT__free_recurse(H5RT_node_t *node);
+
+/* Result buffer helper functions */
+static herr_t H5RT__result_set_init(H5RT_result_set_t *result_set);
+static herr_t H5RT__result_set_add(H5RT_result_set_t *result_set, H5RT_leaf_t *leaf);
+static herr_t H5RT__result_set_grow(H5RT_result_set_t *result_set);
+static void   H5RT__result_set_cleanup(H5RT_result_set_t *result_set);
+
+#if defined(H5_HAVE_DARWIN) || defined(H5_HAVE_WIN32_API)
```
jhendersonHDF: What's the reasoning for swapping the argument ordering based on platform? Is this meant to be a performance optimization?

src/H5RT.c
```diff
@@ -0,0 +1,905 @@
...
+static int H5RT__leaf_compare(void *dim, const void *leaf1, const void *leaf2);
+#else
+static int H5RT__leaf_compare(const void *leaf1, const void *leaf2, void *dim);
+#endif
+
+/* Check if two hyper-rectangles specified by (min1, max1) and (min2, max2) intersect */
```
jhendersonHDF: Seems to be a leftover; presumably from the inline function in H5RTpkg.h

src/H5RT.c
```diff
@@ -0,0 +1,905 @@
...
+    /* Set up pointers to sections of the coordinate block */
+    leaf->min = leaf->_coords;
+    leaf->max = leaf->_coords + rank;
+    leaf->mid = leaf->_coords + (2 * rank);
+
+    leaf->rank   = rank;
+    leaf->record = record;
+
+done:
+    if (ret_value < 0 && leaf && leaf->_coords) {
```
jhendersonHDF: Nitpicking a bit: it seems better to make use of `H5RT_leaf_cleanup()` here to keep the cleanup logic consistent.

src/H5RT.c
mattjala: It's to deal with differences in expected argument ordering in qsort across systems - compare [macOS](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/qsort_r.3.html), [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort-s?view=msvc-170), and [linux](https://linux.die.net/man/3/qsort) signatures.

src/H5RTmodule.h
mattjala: Nope - removed

src/H5RT.c
jhendersonHDF: I see, this seems like something that would be best to hide behind an HD-prefixed macro, as typically only `qsort` has been used in the library.

mattjala: You mean something like `HDqsort_r` that compiles to the OS-specific version? We could do that, but we'd still need the different signatures for the comparison callback.

EDIT: Actually, we can set it up so that the comparison callback is also generic - I've done this in the newest commit

src/H5RT.c
```diff
@@ -0,0 +1,886 @@
...
+                0)
+                HGOTO_ERROR(H5E_INTERNAL, H5E_CANTINIT, FAIL, "failed to fill R-tree");
+
+            /* The next 'child_leaf_count' leaves are now assigned */
+            child_leaf_start += child_leaf_count;
+            leaves_left -= child_leaf_count;
+        }
+    }
+
+done:
```
jhendersonHDF, line 438: Should add some cleanup code here to make sure any nodes that were allocated get freed on failure

src/H5private.h
```diff
@@ -765,6 +765,61 @@ H5_DLL H5_ATTR_CONST int Nflock(int fd, int operation);
...
+    void *gnu_arg;
+} HDqsort_r_wrapper_t;
+
+static inline int
+HDqsort_r_wrapper_func(
+#if defined(H5_HAVE_WIN32_API) || defined(H5_HAVE_DARWIN)
+    void *wrapper_arg,
+#endif
+    const void *a, const void *b
+#if !defined(H5_HAVE_WIN32_API) && !defined(H5_HAVE_DARWIN)
```
jhendersonHDF: A little messy, but seems to be about what we need to do this portably.

It seems you don't need this section of #ifdefs due to this wrapper only being needed for platforms where the ordering is the same (`void *, const void *, const void *`), so this can probably be simplified to just `HDqsort_r_wrapper_func(void *, const void *, const void *)`.

This section of code looks good, though should probably follow our usual conventions for portability wrappers:

- All the actual code here (struct definition, `HDqsort_r_wrapper_func()` and `HDqsort_r()`) should be moved to H5system.c, just to keep as much non-macro code out of H5private.h as possible. `HDqsort_r_wrapper_func()` can remain static there, but `HDqsort_r()` should be made available and probably named differently since we want a single macro called `HDqsort_r` that just wraps `qsort_r()` on GNU systems and points to the `HDqsort_r()` function (named differently) on other systems that need the swapping. Don't worry about adding `inline` to the functions; it probably won't make much of a difference.
- There should be a section in H5win32defs.h which defines a `HDqsort_r` macro and maps it to the differently-named function in H5system.c
- The macro in this file should be surrounded by an `#ifndef HDqsort_r` / `#endif` pair
- The macro in this file can retain the logic for `#ifdef H5_HAVE_DARWIN`, where it will map to the function in H5system.c and otherwise just map to `qsort_r` on GNU systems as it currently does
- Also note that `qsort_r` on some BSDs also appears to have the flipped arguments, which is something we'll probably want to deal with. For example, the argument order is flipped up to FreeBSD 13.5, but appears to have been changed to GNU ordering in FreeBSD 14.0 (see https://cgit.freebsd.org/src/commit/?id=af3c78886fd8). More research would be needed on OpenBSD and others.

jhendersonHDF: Based on that commit, I wonder if we can get away with using generics here since we require C11 now.

src/H5RT.c
```diff
@@ -0,0 +1,886 @@
...
+ *
+ *              Frees both the result set structure and the underlying
+ *              results array buffer.
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5RT_free_results(H5RT_result_set_t *result_set)
```
jhendersonHDF, line 632: It seems like this function has some overlap with `H5RT__result_set_cleanup()`. Especially in the use above in `H5RT_search()` where it seems like a call to `H5RT_free_results()` could replace the call to `H5RT__result_set_cleanup()` + `free(result_set)`. Are there distinct use cases for `H5RT__result_set_cleanup()` vs. `H5RT_free_results()`? I'd consider having `H5RT__result_set_cleanup()` only clear the fields if that's useful and have this function be responsible for all the freeing that is to be done.

mattjala: `H5RT_free_results()` was supposed to be the externally-used free that was independent from how the results were actually handled, as opposed to the internal `result_set_cleanup()`. Given how similar they are and the simplicity of cleanup, I think it makes sense to collapse them into one routine.

src/H5system.c
```diff
@@ -1411,3 +1411,51 @@ H5_strcasestr(const char *haystack, const char *needle)
...
+#if defined(H5_HAVE_WIN32_API)
+    qsort_s(base, nel, size, HDqsort_context_wrapper_func, &wrapper);
+#elif defined(H5_HAVE_DARWIN) || (defined(__FreeBSD__) && __FreeBSD__ < 14)
+    /* Old BSD-style: context parameter comes before comparator function */
+    qsort_r(base, nel, size, &wrapper, HDqsort_context_wrapper_func);
+#endif
+}
+#else
+/* GNU/Linux and FreeBSD >= 14.0: direct mapping to qsort_r with GNU signature */
+#define HDqsort_context(base, nel, size, compar, arg) qsort_r((base), (nel), (size), (compar), (arg))
```
jhendersonHDF: This `#else` section can just be removed, as `HDqsort_context` shouldn't be used directly and this definition of a macro will only be visible in this file (and unused) anyway.

jhendersonHDF (dismissed):
One last minor comment; thanks for the changes to give us a generic `qsort_r`-like macro! We can make improvements with C11 generics later.

jhendersonHDF (approved):

fortnern (approved):

Issue #5810, opened 2025-09-11
Simplify ssize_t guessing on Windows


derobins:
We can't easily eliminate ssize_t from the API, but we can simplify the preprocessor messiness by just using ptrdiff_t, which is signed and will almost certainly have the same size in bytes as size_t.

PR #5811, opened 2025-09-11
Use SSIZE_T for ssize_t on Windows

derobins:
The handstands we do to support ssize_t on Windows are a little baroque. Just use SSIZE_T on Windows (but not MinGW).

The PR adds a compile assert to ensure that size_t and ssize_t have the same size, just to be sure.

Addresses #5810 

----

> [!IMPORTANT]
> Simplify `ssize_t` handling on Windows by using `ptrdiff_t` and add compile-time size check.
> 
>   - **Behavior**:
>     - Use `ptrdiff_t` for `ssize_t` on Windows in `H5public.h`.
>     - Add compile-time assertion in `H5.c` to ensure `sizeof(size_t) == sizeof(ssize_t)`.
>   - **Constants**:
>     - Redefine `SSIZE_MAX` to `PTRDIFF_MAX` in `H5private.h`.
>   - **Misc**:
>     - Simplifies handling of `ssize_t` on Windows, addressing issue #5810.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for e76488833069858be542e14b23e4b24eb84c607d. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


qkoziol (dismissed):
Sensible

jhendersonHDF:
Why the switch away from the `SSIZE_T` type? Seems exactly what it was meant for at least.

derobins:
> Why the switch away from the `SSIZE_T` type? Seems exactly what it was meant for at least.

We could do that but there's no 'max' macro for that (e.g., SSIZE_MAX), so we'd have to come up with some sort of scheme for that. I think it's better to use a C type with a standard max macro.

Something like this:

```
#if defined(_WIN64)
    #define SSIZE_MAX _I64_MAX
#else
    #define SSIZE_MAX LONG_MAX
#endif
```

Which I think is a bit hackier and uglier.

EDIT: Okay, it looks like SSIZE_T_MAX is a thing? I'll update the PR.

jhendersonHDF:
> 
> ```
> #if defined(_WIN64)
>     #define SSIZE_MAX _I64_MAX
> #else
>     #define SSIZE_MAX LONG_MAX
> #endif
> ```

Yes, I was going to suggest that simply because I think at some point you have to do something like this anyway to truly be portable, as it's no guarantee that `ptrdiff_t` and `ssize_t` match up (though generally they will on any modern system).

But if Windows gives us a MAX macro for `SSIZE_T`, all the better (though frustratingly not the POSIX `SSIZE_MAX`...).

derobins:
> > ```
> > #if defined(_WIN64)
> >     #define SSIZE_MAX _I64_MAX
> > #else
> >     #define SSIZE_MAX LONG_MAX
> > #endif
> > ```
> 
> Yes, I was going to suggest that simply because I think at some point you have to do something like this anyway to truly be portable, as it's no guarantee that `ptrdiff_t` and `ssize_t` match up (though generally they will on any modern system).
> 
> But if Windows gives us a MAX macro for `SSIZE_T`, all the better (though frustratingly not the POSIX `SSIZE_MAX`...).

It's such an easy and useful thing to add, yet Microsoft doesn't do it...

jhendersonHDF:
Missing `#include <BaseTsd.h>`

qkoziol (approved):

jhendersonHDF (approved):

byrnHDF (approved):

PR #5843, opened 2025-09-16
Optimize VDS operations with r-tree

mattjala:
Depends on #5800

----

> [!IMPORTANT]
> Optimize VDS operations using R-tree spatial index, adding new API functions and tests for improved performance.
> 
>   - **Behavior**:
>     - Introduces R-tree spatial index for optimizing VDS operations in `H5Dvirtual.c`.
>     - Adds `H5Pset_dset_use_spatial_tree()` and `H5Pget_dset_use_spatial_tree()` to control R-tree usage.
>     - Default behavior uses R-tree for VDS with more than 1000 mappings.
>   - **Implementation**:
>     - Adds `H5RT.c`, `H5RTprivate.h`, and `H5RTpkg.h` for R-tree implementation.
>     - Updates `H5Pdapl.c` and `H5Pdcpl.c` to include R-tree properties.
>     - Modifies `H5Dvirtual.c` to integrate R-tree in VDS I/O operations.
>   - **Testing**:
>     - Adds `rtree.c` for testing R-tree creation, search, and copy operations.
>     - Tests R-tree integration with VDS in `test/dsets.c`.
>     - Verifies R-tree behavior with different dataset access property list settings.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for afc2e518aa2e15d85f49f1ac73891b38de4756ef. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


brtnfld:
1. Memory Management Concerns

  Static vs Dynamic Trade-offs:
```
  // Potential issue: Memory overhead for small mapping counts
  if (storage->list_nused < SPATIAL_TREE_THRESHOLD) {
      // Use linear scan for small datasets
      return H5D__virtual_linear_search();
  }

```
  Recommendations:
  - Threshold-based activation: Only use R-tree when list_nused > N (e.g., N=50)

  2. Dynamic Updates Complexity

  Current VDS allows dynamic mapping additions:
  ```
// If mappings change frequently, R-tree rebuilding is expensive
  H5Pset_virtual(dcpl_id, vspace, src_file, src_dset, src_space);
```

  Potential solutions:
  - Incremental updates: Support R-tree node insertion/deletion
  - Rebuild triggers: Automatic reconstruction when the mapping ratio changes
  - Hybrid approach: Fall back to linear for highly dynamic datasets

  3. Multi-dimensional Optimization

  ```
VDS supports unlimited dimensions:
  // R-tree performance depends on dimensionality
  if (unlimited_dims > 2) {
      // Higher dimensions may reduce R-tree effectiveness
  }
```

  Considerations:
  - Curse of dimensionality: R-tree efficiency decreases with >3-4 dimensions
  - Dimension-specific optimization: Different strategies for 1D, 2D, 3D+ cases
  
1. Adaptive Algorithm Selection

```
  static herr_t
  H5D__virtual_select_search_algorithm(H5O_storage_virtual_t *storage, 
                                      H5S_t *file_space)
  {
      if (storage->list_nused < 50)
          return H5D__virtual_linear_search();

      if (H5S_GET_EXTENT_NDIMS(file_space) > 4)
          return H5D__virtual_linear_search();

      return H5D__virtual_rtree_search();
  }
```

  2. Performance Monitoring

 ```
 // Add performance counters for different access patterns
  typedef struct {
      uint64_t linear_searches;
      uint64_t rtree_searches;
      uint64_t total_intersections_computed;
      double   avg_search_time;
  } H5D_vds_perf_stats_t;
```

  3. Testing Strategy Enhancements

  Comprehensive test coverage needed:
  - Scalability tests: 1, 10, 100, 1000, 10000 mappings
  - Dimensionality tests: 1D through 6D datasets
  - Access pattern tests: Sequential, random, clustered access
  - Memory usage tests: R-tree memory overhead measurement
  - Dynamic update tests: Mapping addition/removal performance

**Comment by mattjala** [comment]
> 1. Memory Management Concerns
> 
> 
> Static vs Dynamic Trade-offs:
> 
> ```
>   // Potential issue: Memory overhead for small mapping counts
>   if (storage->list_nused < SPATIAL_TREE_THRESHOLD) {
>       // Use linear scan for small datasets
>       return H5D__virtual_linear_search();
>   }
> ```
> 
> Recommendations:
> 
>     * Threshold-based activation: Only use R-tree when list_nused > N (e.g., N=50)
> 
> 
>     2. Dynamic Updates Complexity
> 
> 
> Current VDS allows dynamic mapping additions:
> 
> ```
> // If mappings change frequently, R-tree rebuilding is expensive
> H5Pset_virtual(dcpl_id, vspace, src_file, src_dset, src_space);
> ```
> 
> Potential solutions:
> 
>     * Incremental updates: Support R-tree node insertion/deletion
> 
>     * Rebuild triggers: Automatic reconstruction when the mapping ratio changes
> 
>     * Hybrid approach: Fall back to linear for highly dynamic datasets
> 
> 
>     3. Multi-dimensional Optimization
> 
> 
> ```
> VDS supports unlimited dimensions:
> // R-tree performance depends on dimensionality
> if (unlimited_dims > 2) {
>     // Higher dimensions may reduce R-tree effectiveness
> }
> ```
> 
> Considerations:
> 
>     * Curse of dimensionality: R-tree efficiency decreases with >3-4 dimensions
> 
>     * Dimension-specific optimization: Different strategies for 1D, 2D, 3D+ cases
> 
> 
>     1. Adaptive Algorithm Selection
> 
> 
> ```
>   static herr_t
>   H5D__virtual_select_search_algorithm(H5O_storage_virtual_t *storage, 
>                                       H5S_t *file_space)
>   {
>       if (storage->list_nused < 50)
>           return H5D__virtual_linear_search();
> 
>       if (H5S_GET_EXTENT_NDIMS(file_space) > 4)
>           return H5D__virtual_linear_search();
> 
>       return H5D__virtual_rtree_search();
>   }
> ```
> 
>     2. Performance Monitoring
> 
> 
> ```
> // Add performance counters for different access patterns
>  typedef struct {
>      uint64_t linear_searches;
>      uint64_t rtree_searches;
>      uint64_t total_intersections_computed;
>      double   avg_search_time;
>  } H5D_vds_perf_stats_t;
> ```
> 
>     3. Testing Strategy Enhancements
> 
> 
> Comprehensive test coverage needed:
> 
>     * Scalability tests: 1, 10, 100, 1000, 10000 mappings
> 
>     * Dimensionality tests: 1D through 6D datasets
> 
>     * Access pattern tests: Sequential, random, clustered access
> 
>     * Memory usage tests: R-tree memory overhead measurement
> 
>     * Dynamic update tests: Mapping addition/removal performance

1. Yes, I think it makes sense to restrict r-tree usage to some mapping count threshold. I'll implement this.

2. I'm not entirely sure what you mean. Mappings don't change frequently, since they can only be added, and can only be added before the dataset is actually created. Additionally, bulk-creation is faster than incremental updates, and we have no reason to support incremental updates at the moment.

3. Again, I'm not entirely sure what you mean. The performance likely does depend on dimensionality, but there's no specific reason to think it's worse than it could be, and there aren't any specific optimizations I'm aware of that would be in scope for this first implementation. Choosing linear vs. rtree search would be redundant once r-tree are only used for high numbers of mappings anyway, since the only case where linear might be better is when there are very few mappings.

Testing strategy - Scalability and dimensionality tests are already included. The current tests do random access, which should be enough to verify correctness. Everything else you mentioned would fall under benchmarking and should probably be handled elsewhere, if it's something we want to include in the library itself.

**Review by brtnfld — CHANGES_REQUESTED** [review]
You'll need to add a test for the new Fortran APIs; otherwise, it seems to be good.

File: src/H5Dvirtual.c [hunk 1]
```diff
@@ -2776,269 +2815,350 @@ H5D__virtual_io_init(H5D_io_info_t *io_info, H5D_dset_io_info_t H5_ATTR_UNUSED *
...
+    /* Initialize layout if necessary */
+    if (!storage->init)
+        if (H5D__virtual_init_all(dset) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTINIT, FAIL, "can't initialize virtual layout");
+
+    /* Initialize tot_nelmts */
+    *tot_nelmts = 0;
+
+    /* If r-tree use is enabled and no tree currently exists, build the tree */
+    if (NULL == (dapl_plist = (H5P_genplist_t *)H5I_object(dset->shared->dapl_id)))
```
fortnern: Should check if there is already a tree before checking the property - saves the somewhat expensive plist lookup if the tree exists

src/H5Dvirtual.c
```diff
@@ -2776,269 +2815,350 @@ H5D__virtual_io_init(H5D_io_info_t *io_info, H5D_dset_io_info_t H5_ATTR_UNUSED *
...
+        /* Free search results */
+        if (H5RT_free_results(search_results) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTFREE, FAIL, "can't free R-tree search results");
+        search_results = NULL;
+    }
+
+    /* Iterate over the mappings that are not stored in the tree */
+    /* Index of each boolean in the 'is in tree' list = index of the mapping it describes in the mapping list
+     */
+    for (size_t i = 0; i < storage->list_nused; i++) {
```
fortnern: We should have a separate array of mappings that are not in the tree (if there is a tree). We don't want to have to iterate over the entire mapping list on every I/O. Even if it's only checking a flag, it's still an O(N^2) operation (where both number of mappings and number of operations scale with N).

src/H5Dvirtual.c
```diff
@@ -3672,3 +3792,254 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+    herr_t                     ret_value   = SUCCEED;
+    H5S_t                     *vspace      = NULL;
+    H5S_t                     *src_space   = NULL;
+    hsize_t                    virt_nelems = 0;
+    hsize_t                    src_nelems  = 0;
+    H5O_storage_virtual_ent_t *entry       = NULL;
+
+    FUNC_ENTER_PACKAGE_NOERR
+
+    if (!mapping_entry || !should_insert) {
```
fortnern: Should this be an assertion?

src/H5Dvirtual.c
```diff
@@ -3672,3 +3792,254 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+ *
+ * Purpose:     Determine whether the given mapping is valid for
+ *              insertion into a spatial tree
+ *
+ * Return:      Non-negative on success/Negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5D__rtree_should_insert(void *mapping_entry, bool *should_insert)
```
fortnern: This should either be an inline void function, a macro, or it should be converted to handle errors normally. I'm pretty sure the checks below can be assertions, or eliminated if this is turned into a macro.

src/H5Dvirtual.c
```diff
@@ -3672,3 +3792,254 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+
+        /* Initialize leaf with dynamic coordinate allocation */
+        curr_leaf = &leaves_temp[curr_leaf_count];
+        if (H5RT_leaf_init(curr_leaf, rank, (uintptr_t)i) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't initialize R-tree leaf");
+
+        /* Record is already set by H5RT_leaf_init */
+
+        if ((curr_space = mappings[i].source_dset.virtual_select) == NULL)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "mapping has no virtual space");
```
fortnern: I think this can be an assertion

fortnern: Of course you still need to assign curr_space

src/H5Dvirtual.c
```diff
@@ -3672,3 +3792,254 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTINSERT, FAIL, "error checking if mapping should be inserted");
+
+        if (!should_insert_space)
+            continue;
+
+        is_in_tree[i] = true;
+
+        /* Initialize leaf with dynamic coordinate allocation */
+        curr_leaf = &leaves_temp[curr_leaf_count];
+        if (H5RT_leaf_init(curr_leaf, rank, (uintptr_t)i) < 0)
```
fortnern: I'm still pretty sure you could have the r-tree use a void * for records - just pass `curr_mapping` instead of `i` here, and `curr_leaf->record` instead of `&storage->list[mapping_index]` in the invocation of H5D__virtual_pre_io_process_mapping() in the r-tree block of H5D__virtual_pre_io()

brtnfld (dismissed):

java/src/hdf/hdf5lib/H5.java
```diff
@@ -10675,6 +10675,41 @@ public synchronized static native void H5Pset_virtual_prefix(long dapl_id, Strin
...
+     * @exception HDF5LibraryException
+     *            Error from the HDF5 Library.
+     **/
+    public synchronized static native boolean H5Pget_dset_use_spatial_tree(long dapl_id)
+        throws HDF5LibraryException;
+
+    /**
+     * @ingroup JH5P
+     *
+     * H5Pset_dset_use_spatial_tree sets the dapl to not use (or explicitly use) a spatial tree
```
fortnern: The description should specify the purpose of the tree is to accelerate VDS I/O operations. It should also specify the default value.

fortnern: The description in the release notes is good, we should have that level of detail here, since people aren't going to go back and read old release notes to figure out what an API does.

src/H5Dvirtual.c
```diff
@@ -70,6 +71,50 @@
...
+    do {                                                                                                     \
+        H5S_t  *_vspace      = NULL;                                                                         \
+        H5S_t  *_src_space   = NULL;                                                                         \
+        hsize_t _virt_nelems = 0;                                                                            \
+        hsize_t _src_nelems  = 0;                                                                            \
+                                                                                                             \
+        assert(entry);                                                                                       \
+                                                                                                             \
+        /* Get virtual space and element count */                                                            \
+        if ((_vspace = (entry)->source_dset.virtual_select) != NULL)                                         \
```
fortnern: I believe all of these checks except ndims < 1 can be accomplished with simply `if ((entry)->unlim_dim_virtual >= 0)`

fortnern: So the macro could simply be `!(((entry)->unlim_dim_virtual >= 0) || ((entry)->source_dset.virtual_select && H5S_GET_EXTENT_NDIMS((entry)->source_dset.virtual_select) < 1))`

fortnern: And eliminate the should_insert parameter

fortnern: Along with comments explaining the conditions of course

src/H5Dvirtual.c
```diff
@@ -1053,6 +1114,37 @@ H5D__virtual_copy_layout(H5O_layout_t *layout)
...
+        if ((new_tree = H5RT_copy(virt->tree)) == NULL)
+            HGOTO_ERROR(H5E_DATASET, H5E_CANTCOPY, FAIL, "unable to copy spatial tree");
+        virt->tree = new_tree;
+
+        /* Copy not_in_tree_list (pointer array) */
+        if (virt->not_in_tree_nused > 0) {
+            assert(orig_not_in_tree_list);
+
+            /* Allocate new pointer array */
+            if ((virt->not_in_tree_list = (H5O_storage_virtual_ent_t **)H5MM_calloc(
```
fortnern, line 1106: Only need to alloc nused slots in the array, not nalloc.

src/H5Dvirtual.c
```diff
@@ -2776,267 +2884,290 @@ H5D__virtual_io_init(H5D_io_info_t *io_info, H5D_dset_io_info_t H5_ATTR_UNUSED *
...
- * Purpose:     Project all virtual mappings onto mem_space, with the
- *              results stored in projected_mem_space for each mapping.
- *              Opens all source datasets if possible.  The total number
- *              of elements is stored in tot_nelmts.
- *
- * Return:      Non-negative on success/Negative on failure
- *
- *-------------------------------------------------------------------------
- */
+// TODO
```
fortnern: Needs function description

src/H5Dvirtual.c
```diff
@@ -2776,267 +2884,290 @@ H5D__virtual_io_init(H5D_io_info_t *io_info, H5D_dset_io_info_t H5_ATTR_UNUSED *
...
+             * H5Dget_space().  Do not attempt to open it as this might
+             * affect the extent and we are not going to recalculate it
+             * here. */
+            assert(curr_mapping->unlim_dim_virtual >= 0);
+            assert(!curr_mapping->source_dset.dset);
+        } /* end else */
+    }     /* end else */
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5D__virtual_pre_io_process_mapping() */
```

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+        HGOTO_ERROR(H5E_DATASET, H5E_CANTGET, FAIL, "can't get rank of dataspace");
+
+    if (rank == 0)
+        HGOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL, "mapping has zero-dimensional space");
+
+    /* Allocate array of leaf structures and not-in-tree mappings */
+    if ((leaves_temp = (H5RT_leaf_t *)calloc(num_mappings, sizeof(H5RT_leaf_t))) == NULL)
+        HGOTO_ERROR(H5E_DATASET, H5E_CANTALLOC, FAIL, "can't allocate leaves array");
+
+    if ((not_in_tree = (H5O_storage_virtual_ent_t **)H5MM_calloc(
```
fortnern: Can we initially allocate this as a moderate size and grow dynamically, similar to how the results array works?

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+                H5RT_leaf_cleanup(&leaves_temp[j]);
+            }
+            free(leaves_temp);
+        }
+        if (not_in_tree)
+            H5MM_free(not_in_tree);
+    }
+
+    FUNC_LEAVE_NOAPI(ret_value);
+}
```
fortnern:
```suggestion
} /* end H5D__mappings_to_leaves() */
```

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+    else
+        /* Close projected memory space */
+        if (mapping->source_dset.projected_mem_space) {
+            if (H5S_close(mapping->source_dset.projected_mem_space) < 0)
+                HGOTO_ERROR(H5E_DATASET, H5E_CLOSEERROR, FAIL, "can't close temporary space");
+            mapping->source_dset.projected_mem_space = NULL;
+        } /* end if */
+done:
+    FUNC_LEAVE_NOAPI(ret_value);
+}
```
fortnern:
```suggestion
} /* end H5D__virtual_close_mapping() */

```

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+                H5RT_leaf_cleanup(&leaves[i]);
+            }
+            free(leaves);
+        }
+        if (not_in_tree_mappings)
+            H5MM_free(not_in_tree_mappings);
+    }
+
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* H5D__virtual_build_tree() */
```

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+
+    if (!tree_enabled_dapl) {
+        *should_build_tree = false;
+        HGOTO_DONE(SUCCEED);
+    }
+
+    *should_build_tree = true;
+done:
+    FUNC_LEAVE_NOAPI(ret_value);
+}
```
fortnern:
```suggestion
} /* end H5D__should_build_tree() */
```

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+                HGOTO_ERROR(H5E_DATASET, H5E_READERROR, FAIL, "unable to read source dataset");
+    } /* end if */
+    else
+        /* Read from source dataset */
+        if (H5D__virtual_read_one_src(dset_info, &mapping->source_dset) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_READERROR, FAIL, "unable to read source dataset");
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5D__virtual_read_one_mapping() */
```

src/H5Dvirtual.c
```diff
@@ -3672,3 +3902,374 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+                HGOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "unable to write to source dataset");
+    }
+    else
+        /* Write to source dataset */
+        if (H5D__virtual_write_one_src(dset_info, &mapping->source_dset) < 0)
+            HGOTO_ERROR(H5E_DATASET, H5E_WRITEERROR, FAIL, "unable to write to source dataset");
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+}
```
fortnern:
```suggestion
} /* end H5D__virtual_write_one_mapping() */
```

fortnern (changes requested):
See comments

java/src/hdf/hdf5lib/H5.java
```diff
@@ -10675,6 +10675,59 @@ public synchronized static native void H5Pset_virtual_prefix(long dapl_id, Strin
     public synchronized static native void H5Pset_efile_prefix(long dapl_id, String prefix)
         throws HDF5LibraryException, NullPointerException;
 
+    /**
+     * @ingroup JH5P
+     *
+     * H5Pget_dset_use_spatial_tree accesses the flag for whether to use/not use a spatial tree
```
fortnern: These descriptions look good, thanks! Can you add them to the fortran and C RM entries as well?

src/H5Dvirtual.c
```diff
@@ -3672,3 +3893,454 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+    else
+        /* Close projected memory space */
+        if (mapping->source_dset.projected_mem_space) {
+            if (H5S_close(mapping->source_dset.projected_mem_space) < 0)
+                HGOTO_ERROR(H5E_DATASET, H5E_CLOSEERROR, FAIL, "can't close temporary space");
+            mapping->source_dset.projected_mem_space = NULL;
+        } /* end if */
+done:
+    FUNC_LEAVE_NOAPI(ret_value);
+} /* end H5D__virtual_close_mapping() */
```
fortnern:
```suggestion
} /* end H5D__virtual_close_mapping() */

```

fortnern: UNIX like OSs require a newline at the end of the file.

fortnern (changes requested):
Looks good apart from the RM entries and the missing newline

fortnern (dismissed):

src/H5Ppublic.h
```diff
@@ -6493,6 +6527,37 @@ H5_DLL herr_t H5Pset_chunk_opts(hid_t plist_id, unsigned opts);
...
+ *          checked directly.
+ *
+ *          Certain workflows may find that tree creation overhead outweighs the time saved
+ *          on reads. In this case, disabling this property will lead to a performance improvement,
+ *          though it is expected that almost all cases will benefit from the tree on net.
+ *
+ * \since 2.0.0
+ *
+ */
+H5_DLL herr_t H5Pset_dset_use_spatial_tree(hid_t dcpl_id, bool use_tree);
```
jhendersonHDF: A bit nitpicky at this point, but it seems like it would be better if this was called something like `H5Pset_virtual_(dset_)use_spatial_tree` as it's specific to VDS and other VDS property list functions use "virtual" as somewhat of a namespace, which makes it easier for discovery purposes.

test/rtree.c
```diff
@@ -0,0 +1,1063 @@
...
+ * Purpose:     Test the R-tree spatial indexing implementation.
+ */
+
+#include "h5test.h"
+
+/*
+ * This file needs to access private datatypes from the H5RT package.
+ */
+#define H5RT_FRIEND /*suppress error about including H5RTpkg */
+#define H5RT_TESTING
```
jhendersonHDF: Minor note: developer builds show that this macro isn't used. It also seemed like `H5CXprivate.h` and `H5VLprivate.h` below aren't used either.

src/H5Ppublic.h
fortnern: I agree. I think without the dset is better (sorry)

fortran/src/H5Pff.F90
```diff
@@ -4365,6 +4365,96 @@ END FUNCTION h5pget_chunk_cache_c
...
+!! on reads. In this case, disabling this property will lead to a performance improvement,
+!! though it is expected that almost all cases will benefit from the tree on net.
+!!
+!! \param dapl_id  Target dataset access property list identifier.
+!! \param use_tree Value of the setting.
+!! \param hdferr   \fortran_error
+!!
+!! See C API: @ref H5Pget_virtual_dset_use_spatial_tree()
+!!
+  SUBROUTINE h5pget_virtual_dset_use_spatial_tree_f(dapl_id, use_tree, hdferr)
```
fortnern: Sorry to be late on this, but can you change the name of this and the other new public functions to h5pget_virtual_spatial_tree_f() etc.? (no dset or use). This way we're consistent with H5Pset_chunk_cache(), H5Pset_virtual_prefix(), etc.

src/H5Dvirtual.c
```diff
@@ -1053,6 +1092,39 @@ H5D__virtual_copy_layout(H5O_layout_t *layout)
         virt->list_nalloc = 0;
     } /* end else */
 
+    /* Copy spatial tree if it exists */
+    if (virt->tree) {
+        if ((new_tree = H5RT_copy(virt->tree)) == NULL)
```
jhendersonHDF, line 1097: On error, do resources for `new_tree` and the below `virt->not_in_tree_list` need to be freed?

jhendersonHDF: EDIT: I see `H5D__virtual_free_layout_mappings()` is eventually called to free these.

mattjala: They do need to be freed, but they're stored on the layout as soon as their allocation succeeds, and in the event of a failure are cleaned up by `H5D__virtual_reset_layout() -> H5D__virtual_free_layout_mappings()`

src/H5Dvirtual.c
```diff
@@ -3672,3 +3893,454 @@ H5D__virtual_release_source_dset_files(H5D_virtual_held_file_t *head)
...
+    FUNC_ENTER_PACKAGE
+
+    /* Check for "printf" source dataset resolution */
+    if (mapping->psfn_nsubs || mapping->psdn_nsubs) {
+        /* Iterate over sub-source dsets */
+        for (size_t j = mapping->sub_dset_io_start; j < mapping->sub_dset_io_end; j++)
+            /* Close projected memory space */
+            if (mapping->sub_dset[j].projected_mem_space) {
+                if (H5S_close(mapping->sub_dset[j].projected_mem_space) < 0)
+                    HGOTO_ERROR(H5E_DATASET, H5E_CLOSEERROR, FAIL, "can't close temporary space");
```
jhendersonHDF: Previously, in `H5D__virtual_post_io()`, these `H5S_close()` calls used HDONE_ERROR, presumably to make sure all the dataspaces get closed even on error.

jhendersonHDF (dismissed):
Generally looks good; see remaining comment though

fortnern (approved):

jhendersonHDF (approved):

brtnfld (approved):

Issue #5861, opened 2025-09-23
Read errors for some chunks after upgrading to HDF5 1.14.6


stpklp:
After updating to HDF5 1.14.6 I get read errors for some chunks in long existing HDF5 files. I did some debugging and found out that the root cause was introduced in 1.14.5 here:

https://github.com/HDFGroup/hdf5/blob/0fe0459fc24d71be13d5f266513c2832b525671b/src/H5Zscaleoffset.c#L1213-L1214

I think it has to be `if (minbits > p.size * 8)` instead of `>=`. There is an assert `<=` some lines below which contradicts the new check somehow (line 1229) and even a special handling for minbits equal to full precision which is now completely useless (line 1240).

After changing it locally to `>` and rebuilding HDF5 I could read all chunks without problems.

An example dataset that has the problem with 1.14.6 can be found here: http://fly.mpi-cbg.de/~pietzsch/bdv-examples/drosophila.h5

jhendersonHDF:
Hi @stpklp,

I agree. The issue was introduced in https://github.com/HDFGroup/hdf5/pull/4234 and I believe it may have been overlooked in an effort to quickly get all those security fixes out for the release. We'll fix this for the upcoming release due in a few weeks. Thanks for pointing out this issue!

derobins:
I've created a PR with a fix for this. I've confirmed that it works with the drosophila.h5 file.

I'd like to add a test for this, but I'm not sure I'll get a chance to do so before the 2.0 release. I'll see what I can do, though.

mattjala:
Leaving this open until a regression test file is added

- [ ] Add regression test file

PR #5864, opened 2025-09-24
Fix scale-offset filter errors

derobins:
A security fix from 1.14.6 introduced a regression in the scale-offset filter where normal data values could cause the library to emit an error (not a crash/segfault).

Fixes GitHub #5861

----

> [!IMPORTANT]
> Fixes regression in `H5Z__filter_scaleoffset` to prevent errors from normal data values by adjusting `minbits` condition.
> 
>   - **Behavior**:
>     - Fixes regression in `H5Z__filter_scaleoffset` in `H5Zscaleoffset.c` where `minbits` condition was changed from `>=` to `>` to prevent errors when `minbits` equals the size of the type.
>     - Addresses issue where normal data values could trigger a library error.
>   - **Documentation**:
>     - Updates `CHANGELOG.md` to include the fix for the scale-offset filter regression and reference GitHub issue #5861.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 612485eab119578fec2e6f03f756532861a98d69. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -483,6 +483,12 @@ HDF5 release, platforms tested, and known problems in this release.
 
 ## Library
 
+### Fixed a problem with the scale-offset filter
+
+A security fix added to 1.14.6 introduced a regression where certain data values could trigger a library error (not a crash or segfault).
+
+Fixes GitHub issue #5861
+
 ### Fixed security issue CVE-2025-6857
 
    An HDF5 file had a corrupted v1 B-tree that would result in a stack overflow when performing a lookup on it. This has been fixed with additional integrity checks.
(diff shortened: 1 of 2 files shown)


byrnHDF (approved):

mattjala (approved):

jhendersonHDF:
Thanks for the quick fix. It'd be much more ideal to have a test file for regression testing, but based on how many conflicts I know are going to arise in `h5dumpgentest.c`, for example, it will probably make life easier to follow this up with a test file once everything else has settled down. I definitely don't want to forget that though, as this is a very simple issue we should've been able to catch at the time.

PR #5866, opened 2025-09-24
Develop big endianCorrect BE testing issues

### LINKED ISSUE #5736: Fix testing errors on Big Endian machines
Fix the 20+ testing errors being reported on echidna due to the mismatch in testing comparison between LE and BE reference files.

byrnHDF:
Fixes #5736

----

> [!IMPORTANT]
> Fixes big-endian testing issues by adding conditional logic for copying big-endian test files and including new big-endian test files.
> 
>   - **CMakeLists**:
>     - In `HDF5Examples/C/HL/CMakeLists.txt`, added conditional logic to copy big-endian (`BE`) `.ddl` files if `H5_WORDS_BIGENDIAN` is true.
>     - In `tools/test/h5ls/CMakeTests.cmake`, added logic to copy big-endian test files from `testfiles/bigendian` if `H5_WORDS_BIGENDIAN` is true.
>   - **Test Files**:
>     - Added big-endian test files: `h5ex_lite3BE.ddl`, `h5ex_packet_table_FLBE.ddl`, `h5ex_table_02BE.ddl`, `h5ex_table_04BE.ddl`.
>   - **Testing**:
>     - Updated test file copying logic in `CMakeTests.cmake` to handle big-endian files for `LIST_HDF5_END_TESTLS_FILES`, `LIST_HDF5_END_TEST_FILES`, and `LIST_OTHER_TEST_FILES_END`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for caf4df550e45244e293fb494216110aab775f247. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


jhendersonHDF:
The changes look good for resolving the current testing issues we have, but I'm still very much in favor of masking out the datatypes eventually because all of these diff files are going to be dependent on the machine that the testing is run on. The HL test files are generated at runtime and use native types, so the ddl files won't necessarily be reliable. The new BE test files checked in are fixed, but they're tested with h5ls which means that they may or may not print "native int" for attr2 in `tools/test/h5ls/expected/tall-3BE.ls`, for example.

byrnHDF:
There needs to be a review of this testing method for the next release. This is to get through this release. I may need to adjust the ref files if someone ca test the actual h5ls results.

hyoklee (changes requested):
3 errors: https://my.cdash.org/viewBuildError.php?buildid=3108544

hyoklee (changes requested):
11 test failures: https://my.cdash.org/viewTest.php?onlyfailed&buildid=3109734

hyoklee (changes requested):
Same 11 test failures: https://my.cdash.org/viewTest.php?onlyfailed&buildid=3110915

byrnHDF:
> Same 11 test failures: https://my.cdash.org/viewTest.php?onlyfailed&buildid=3110915

Looks like the correct ref files are being used - just not all the datatype instances were changed.

byrnHDF:
There should only be two failing tests remaining - h5_api_test_dataset and H5_H5DUMP-h5ex_image2. They would require on machine debugging most likely. This PR does not address those failures.

.claude/settings.local.json
```diff
@@ -0,0 +1,28 @@
+{
+  "permissions": {
```
mattjala: Looks like this file was included by accident

HDF5Examples/CMakeLists.txt
```diff
@@ -21,7 +21,12 @@ set (H5EX_JAVA_DIR           ${H5EXAMPLES_SOURCE_DIR}/JAVA)
 #-----------------------------------------------------------------------------
 if (NOT EXAMPLES_EXTERNALLY_CONFIGURED)
   set (H5EX_CONFIG_DIR      ${H5EXAMPLES_SOURCE_DIR}/config)
-  set (H5EX_RESOURCES_DIR      ${H5EXAMPLES_SOURCE_DIR}/config/cmake)
+  set (H5EX_RESOURCES_DIR   ${H5EXAMPLES_SOURCE_DIR}/config/cmake)
+
+  include (TestBigEndian)
+  if (NOT WINDOWS)
+    TEST_BIG_ENDIAN (H5EX_WORDS_BIGENDIAN)
```
mattjala, line 28: `TEST_BIG_ENDIAN` doesn't seem to be defined anywhere. Is this supposed to be the (deprecated in favor of `CMAKE_<LANG>_BYTE_ORDER`) `test_big_endian` CMake default?

Also, why is it disabled on Windows?

mattjala (changes requested):
Remove claude settings file

HDF5Examples/CMakeLists.txt
byrnHDF: include (TestBigEndian) enables the CMake function
Windows is always LE

mattjala (approved):

hyoklee (commented):
I can see 2 failures: https://my.cdash.org/viewTest.php?onlyfailed&buildid=3112172

hyoklee (approved):

PR #5872, opened 2025-09-24
Fix CVE-2025-2310

### LINKED ISSUE #5871: Malformed file leads to buffer overflow during attribute metadata decoding
See [CVE-2025-2310](https://nvd.nist.gov/vuln/detail/cve-2025-2310).

mattjala:
Malformed files can have a zero name-length, which when subtracted lead to an overflow and an out-of-bounds read.

Check that name length is not too small in addition to checking for an overflow directly.

----

> [!IMPORTANT]
> Fixes overflow and out-of-bounds read in `H5O__attr_decode()` by ensuring `name_len` is greater than 1 in `H5Oattr.c`.
> 
>   - **Security Fix**:
>     - In `H5O__attr_decode()` in `H5Oattr.c`, added a check to ensure `name_len` is greater than 1 to prevent overflow and out-of-bounds read.
>     - This prevents processing of malformed files with zero-length names.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for a5a0b305acd8fcd2a2d0e5e012072411f7411756. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5Oattr.c
+++ b/src/H5Oattr.c
@@ -167,6 +167,11 @@ H5O__attr_decode(H5F_t *f, H5O_t *open_oh, unsigned H5_ATTR_UNUSED mesg_flags, u
     if (H5_IS_BUFFER_OVERFLOW(p, 2, p_end))
         HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
     UINT16DECODE(p, name_len); /* Including null */
+
+    /* Verify that retrieved name length (including null byte) is valid */
+    if (name_len <= 1)
+        HGOTO_ERROR(H5E_OHDR, H5E_CANTDECODE, NULL, "decoded name length is invalid");
+
     if (H5_IS_BUFFER_OVERFLOW(p, 2, p_end))
         HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
     UINT16DECODE(p, attr->shared->dt_size);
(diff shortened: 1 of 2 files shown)


src/H5Oattr.c
```diff
@@ -190,6 +190,12 @@ H5O__attr_decode(H5F_t *f, H5O_t *open_oh, unsigned H5_ATTR_UNUSED mesg_flags, u
      */
     if (H5_IS_BUFFER_OVERFLOW(p, name_len, p_end))
         HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
+
+    /* Verify that name is null-terminated within its expected length to prevent overrun during strndup */
+    if (memchr((const char *)p, '\0', name_len) == NULL)
```
jhendersonHDF: I don't believe that this is the correct fix, as this is the whole point to using `H5MM_strndup` below so that only the known amount of bytes are read from the buffer, even when the buffer is not a NUL-terminated string. Rather, `name_len` is 0 in this file, causing `name_len - 1` in the call to `H5MM_strndup` to wrap around to `SIZE_MAX` and thus overflow the buffer.

mattjala: Right - I should have checked the man page.

The `name_len` check corrects the specific error that was pointed out in the CVE, but it seems like there may be another, less consistent issue with how the file is decoded. I've seen local crashes with ASan trying to access the nbit and scaleoffset datasets. I'm still testing to see if this is a distinct issue with the library, or a problem with my setup (since it seems ASan itself may be crashing).

bmribler (dismissed):

src/H5Oattr.c
```diff
@@ -190,6 +190,11 @@ H5O__attr_decode(H5F_t *f, H5O_t *open_oh, unsigned H5_ATTR_UNUSED mesg_flags, u
      */
     if (H5_IS_BUFFER_OVERFLOW(p, name_len, p_end))
         HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
+
+    /* Verify that retrieved name length (including null byte) is valid */
+    if (name_len <= 1)
```
bmribler: Shouldn't we find where name_len was decoded?

mattjala: It was decoded earlier in this function - `UINT16DECODE(p, name_len);` at `H5Oattr.c:169`. I wanted to put this check next to the other name_len value check.

jhendersonHDF (changes requested):
Should add a note to the CHANGELOG.md, but otherwise looks good

src/H5Oattr.c
bmribler: Hmm, I think this check is more appropriate right after the value being decoded.

jhendersonHDF (approved):

bmribler (approved):

PR #5874, opened 2025-09-24
Set max POSIX IO size to 0x7ffff000 bytes

### LINKED ISSUE #5873: The maximum POSIX I/O size is wrong for Linux
The maximum POSIX I/O size is set to SSIZE_MAX in H5private.h. While this is correct as far as POSIX is concerned, it is much larger than the maximum size Linux will accept.

Linux defines the max size it will accept as MAX_RW_COUNT in fs.h, which is defined to be INT_MAX & PAGE_CACHE_MASK, or 0x7ffff000. The value in H5private.h should be updated to reflect this.

derobins:
This was previously set to SSIZE_MAX on Linux, which is correct as far as POSIX is concerned, but Linux defines the max as MAX_RW_COUNT, which is set to INT_MAX & PAGE_CACHE_MASK (0x7ffff000) in fs.h. Since this is very close to the INT_MAX value that was previously set for MacOS and Windows, HDF5 will now use the Linux value on all platforms.

Larger I/O sizes will be split into multiple smaller read/write calls in the VFDs.

Fixes #5873

----

> [!IMPORTANT]
> Set maximum POSIX I/O size to `0x7ffff000` bytes for all platforms, splitting larger operations, and updated documentation.
> 
>   - **Behavior**:
>     - Set maximum POSIX I/O size to `0x7ffff000` bytes for all platforms in `H5private.h`.
>     - Larger I/O operations are split into smaller read/write calls.
>   - **Documentation**:
>     - Updated `CHANGELOG.md` to reflect the new maximum POSIX I/O size.
>   - **Issue Fix**:
>     - Fixes issue #5873 by standardizing I/O size limits across platforms.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 3b0b7cfa53c0affa596ca77ec8f553865d225d9f. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5private.h
+++ b/src/H5private.h
@@ -401,22 +401,18 @@
 #endif /* H5_HAVE_PARALLEL */
 
 /*
- * Types and max sizes for POSIX I/O.
- * OS X (Darwin) is odd since the max I/O size does not match the types.
+ * Types and max I/O size for POSIX read/write calls
  */
 #if defined(H5_HAVE_WIN32_API)
-#define h5_posix_io_t         unsigned int
-#define h5_posix_io_ret_t     int
-#define H5_POSIX_MAX_IO_BYTES INT_MAX
-#elif defined(H5_HAVE_DARWIN)
-#define h5_posix_io_t         size_t
-#define h5_posix_io_ret_t     ssize_t
-#define H5_POSIX_MAX_IO_BYTES INT_MAX
+#define h5_posix_io_t     unsigned int
+#define h5_posix_io_ret_t int
 #else
-#define h5_posix_io_t         size_t
-#define h5_posix_io_ret_t     ssize_t
-#define H5_POSIX_MAX_IO_BYTES SSIZE_MAX
+#define h5_posix_io_t     size_t
+#define h5_posix_io_ret_t ssize_t
 #endif
+
+/* Use the Linux limit (MAX_RW_COUNT, which is INT_MAX & PAGE_CACHE_MASK) for everyone */
+#define H5_POSIX_MAX_IO_BYTES 0x7ffff000
 
 /* POSIX I/O mode used as the third parameter to open/_open
  * when creating a new file (O_CREAT is set).
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -172,7 +172,11 @@ HDF5 release, platforms tested, and known problems in this release.
 
    The standard for building the library is now C11. We have updated the build files to set the C standard to C11, though some platforms use gnu11 to get some GNU things to work.
 
-## Library 
+## Library
+
+### The maximum POSIX I/O size has been set to 0x7ffff000 bytes for all platforms
+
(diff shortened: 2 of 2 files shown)


src/H5private.h
```diff
@@ -401,23 +401,19 @@
...
-#define h5_posix_io_ret_t     ssize_t
-#define H5_POSIX_MAX_IO_BYTES INT_MAX
+#define h5_posix_io_t     unsigned int
+#define h5_posix_io_ret_t int
 #else
-#define h5_posix_io_t         size_t
-#define h5_posix_io_ret_t     ssize_t
-#define H5_POSIX_MAX_IO_BYTES SSIZE_MAX
+#define h5_posix_io_t     size_t
+#define h5_posix_io_ret_t ssize_t
```
derobins, line 411: There's no longer a need to handle Darwin/MacOS separately, so we just need to handle Window's weird types.

qkoziol: A bit of a "all the world's Linux" choice, but probably not terrible.

qkoziol (approved):

brtnfld (approved):

jhendersonHDF:
I'm a bit way of setting this to a fixed value not based on a value defined for the platform, as it will ultimately depend on PAGE_MASK which varies according to PAGE_SIZE. For example, PAGE_SIZE seems to be `1 << 12` by default on x86 and `1 << 16`, `1 << 14` or `1 << 12` by default on arm64.

derobins:
> I'm a bit way of setting this to a fixed value not based on a value defined for the platform, as it will ultimately depend on PAGE_MASK which varies according to PAGE_SIZE. For example, PAGE_SIZE seems to be `1 << 12` by default on x86 and `1 << 16` by default on arm64.

Yeah, I'm not super excited about pulling in Linux kernel headers, though, and POSIX didn't deign to give us a handy symbol to use.

src/H5private.h
derobins: More of a "lowest common denominator"

jhendersonHDF:
> Yeah, I'm not super excited about pulling in Linux kernel headers, though, and POSIX didn't deign to give us a handy symbol to use.

Is there a compelling reason to change this from `SSIZE_MAX`? Shouldn't the way that we handle writes until finish be enough to handle this even if the write ends up being broken up for a size of `SSIZE_MAX`?

jhendersonHDF:
I suppose if you wanted a lowest common denominator, 0x7fff0000 might be safer, though that's just splitting hairs a bit really.

derobins:
> > Yeah, I'm not super excited about pulling in Linux kernel headers, though, and POSIX didn't deign to give us a handy symbol to use.
> 
> Is there a compelling reason to change this from `SSIZE_MAX`? Shouldn't the way that we handle writes until finish be enough to handle this even if the write ends up being broken up for a size of `SSIZE_MAX`?

I think you might be right about leaving this alone, though I might rework this to improve the comments and take another look at MacOS. IIRC, MacOS returned real error codes with I/O larger than INT_MAX, but I could be mis-remembering that.

Let me think about it and I'll update tonight.

PR #5927, opened 2025-10-19
Add Java 25 FFM alternates

### LINKED ISSUE #3652: Add HDF5 Java bindings to Maven
User requested that JHI5, <https://support.hdfgroup.org/products/java/JNI3/jhi5/index.html>, be made available on Maven central, <https://central.sonatype.com/>.

byrnHDF:
Enhanced Maven options.
Added Java 25 FFM source.
Created full FFM tests and examples.
Expandedd Presets and CI.

Closes #5847.
Fixes #3652


brtnfld (dismissed):

java/jsrc/org/windows/hdfgroup/javahdf5/FILE.java
jhendersonHDF, line 1: Especially for files like this, it looks like we may be generating interfaces for all sorts of things that don't have an apparent use?

java/jsrc/org/macos/hdfgroup/javahdf5/H5AC_cache_config_t.java
```diff
@@ -0,0 +1,1438 @@
...
+    }
+
+    private static final GroupLayout $LAYOUT =
+        MemoryLayout
+            .structLayout(
+                hdf5_h.C_INT.withName("version"), hdf5_h.C_BOOL.withName("rpt_fcn_enabled"),
+                hdf5_h.C_BOOL.withName("open_trace_file"), hdf5_h.C_BOOL.withName("close_trace_file"),
+                MemoryLayout.sequenceLayout(1025, hdf5_h.C_CHAR).withName("trace_file_name"),
+                hdf5_h.C_BOOL.withName("evictions_enabled"), hdf5_h.C_BOOL.withName("set_initial_size"),
+                MemoryLayout.paddingLayout(6), hdf5_h.C_LONG.withName("initial_size"),
```
jhendersonHDF, line 65: This makes me think that these files are something that needs to be generated at build time instead of being checked in files

java/jsrc/org/windows/hdfgroup/javahdf5/FILE.java
byrnHDF: All the jsrc files are generated from jextract - just pulling out the hdf5_h* files and feature files we actually care about. I suppose you could create a script to remove unneeded files but seems excessive for now.

java/jsrc/org/macos/hdfgroup/javahdf5/H5AC_cache_config_t.java
byrnHDF: The ideal way is to create these files at build time however that was beyond the scope at this time. The process could be modified to do this in the future.

jhendersonHDF: I guess what I mean to say is that even initially it seems needlessly excessive to check in files for other platforms that won't be used during building and also aren't necessarily going to be correct even on the platform they're being built on. These generated files include the particular memory layout of structures for the platform they were generated on, which isn't going to be static. I'd go so far as to say checking these files in is simply the wrong thing to do here.

java/jsrc/org/windows/hdfgroup/javahdf5/FILE.java
jhendersonHDF: Yes, I think the tool is pulling out far more than we actually need, but this is also something that would be mitigated a bit by not checking in these files.

java/src-jni/jni/h5aImp.c
```diff
@@ -18,9 +18,9 @@ extern "C" {
 #include <stdlib.h>
 #include <string.h>
 #include "hdf5.h"
-#include "h5util.h"
-#include "h5jni.h"
-#include "h5aImp.h"
+#include "../../src-jni/jni/h5util.h"
```
jhendersonHDF: Should probably fix all these renames to set the correct header paths in the build system and not use relative include paths here

byrnHDF: Yes must have been claude!

java/jsrc/org/macos/hdfgroup/javahdf5/H5AC_cache_config_t.java
byrnHDF: Yes, but given the time constraints and the other reqs, there just wasn't enough time. Future work could upgrade this.

CMakeBuildOptions.cmake
```diff
@@ -79,6 +79,12 @@ option (HDF5_BUILD_FORTRAN "Build FORTRAN support" OFF)
 option (HDF5_BUILD_CPP_LIB "Build HDF5 C++ Library" OFF)
 
 option (HDF5_BUILD_JAVA "Build Java HDF5 Library" OFF)
+option (HDF5_ENABLE_JNI "Force JNI implementation instead of FFM for Java bindings when Java 24+ is available" OFF)
```
jhendersonHDF: Shouldn't this option be default ON?

java/jsrc/org/windows/hdfgroup/javahdf5/FILE.java
byrnHDF: It was a bit of a shock (but should have been expected) that there would be platform issues. However, I decided that the implementation of the tests, examples and compatibility class should be more important and handled before making things efficient.

CMakeBuildOptions.cmake
byrnHDF: Maybe, I will test it out if the mac CI passes and I know all the CI is working.

java/src-jni/jni/h5aImp.c
byrnHDF: Fixed

byrnHDF:
Superseded by #5957

Issue #5577, opened 2025-06-06
[BUG] A heap-buffer-overflow in `H5FS__sinfo_serialize_node_cb` at `src/H5FScache.c:1441:9`


JJLeo:
## Description

- Version: Latest commit [17c16b6](https://github.com/HDFGroup/hdf5/commit/17c16b659a3ea175356357802ee2a9d69990fa05)
- Environment: Ubuntu 20.04.6 LTS, Clang 18.1.8
- Fuzzing harness: https://github.com/google/oss-fuzz/blob/master/projects/hdf5/h5_extended_fuzzer.c

**Please let me know if you encounter any issues reproducing it — I can upload a Docker image to help.**

## Steps to reproduce

```bash
git clone https://github.com/HDFGroup/hdf5 --depth 1
cd hdf5
export CC="clang"
export CXX="clang++"
export CFLAGS="-fsanitize=address -g -O0 -fno-omit-frame-pointer"
export CXXFLAGS="-fsanitize=address -g -O0 -fno-omit-frame-pointer -stdlib=libc++"
export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
export LDFLAGS="${CFLAGS}"
export CMAKE_C_FLAGS="${CC} ${CFLAGS}"
export CMAKE_CXX_FLAGS="${CXX} ${CXXFLAGS}"

mkdir build-dir
cd build-dir
cmake -G "Unix Makefiles" \
    -DCMAKE_BUILD_TYPE:STRING=Release \
    -DBUILD_SHARED_LIBS:BOOL=OFF \
    -DBUILD_TESTING:BOOL=OFF \
    -DCMAKE_VERBOSE_MAKEFILES:BOOL=ON \
    -DHDF5_BUILD_EXAMPLES:BOOL=OFF \
    -DHDF5_BUILD_TOOLS:BOOL=OFF \
    -DHDF5_ENABLE_SANITIZERS:BOOL=ON \
    -DHDF5_ENABLE_Z_LIB_SUPPORT:BOOL=ON \
    ..

# Make the build verbose for easy logging inspection
cmake --build . --verbose --config Release -j$(nproc)
cd ..
wget https://github.com/google/oss-fuzz/raw/refs/heads/master/projects/hdf5/h5_extended_fuzzer.c
$CC $CFLAGS  -std=c99 -c \
  -I/src/hdf5/src -I/src/hdf5/build-dir/src -I./src/H5FDsubfiling/ \
[...]

PR #5938, opened 2025-10-23
Fix several issues uncovered by CVE-2025-7067

jhendersonHDF:
Fixes several crashes/assertion failures related to CVE-2025-7067

----

> [!IMPORTANT]
> Fixes CVE-2025-7067 issues by updating `H5FS_sect_add` to handle section merging/shrinking and improving error handling.
> 
>   - **Functionality**:
>     - `H5FS_sect_add` in `H5FSsection.c` now accepts an additional `bool *merged_or_shrunk` parameter to track if sections are merged or shrunk.
>     - Updates to `H5MF__add_sect` in `H5MF.c` to handle `merged_or_shrunk` parameter.
>   - **Error Handling**:
>     - Improved error handling in `H5FS_vfd_alloc_hdr_and_section_info_if_needed` in `H5MF.c` to manage allocation failures.
>   - **Tests**:
>     - Updated test cases in `freespace.c` and `mf.c` to include the new parameter in `H5FS_sect_add` calls.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for ef61da4e905d542a72bed8fccfb9fcba59e60d31. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5FSsection.c
+++ b/src/H5FSsection.c
@@ -1297,7 +1297,8 @@ done:
  *-------------------------------------------------------------------------
  */
 herr_t
-H5FS_sect_add(H5F_t *f, H5FS_t *fspace, H5FS_section_info_t *sect, unsigned flags, void *op_data)
+H5FS_sect_add(H5F_t *f, H5FS_t *fspace, H5FS_section_info_t *sect, unsigned flags, void *op_data,
+              bool *merged_or_shrunk)
 {
     H5FS_section_class_t *cls;                      /* Section's class */
     bool                  sinfo_valid    = false;   /* Whether the section info is valid */
@@ -1317,6 +1318,9 @@ H5FS_sect_add(H5F_t *f, H5FS_t *fspace, H5FS_section_info_t *sect, unsigned flag
     assert(sect);
     assert(H5_addr_defined(sect->addr));
     assert(sect->size);
+
+    if (merged_or_shrunk)
+        *merged_or_shrunk = false;
 
     /* Get a pointer to the section info */
     if (H5FS__sinfo_lock(f, fspace, H5AC__NO_FLAGS_SET) < 0)
@@ -1344,9 +1348,12 @@ H5FS_sect_add(H5F_t *f, H5FS_t *fspace, H5FS_section_info_t *sect, unsigned flag
     /* (If section has been completely merged or shrunk away, 'sect' will
      *  be NULL at this point - QAK)
      */
-    if (sect)
+    if (sect) {
         if (H5FS__sect_link(fspace, sect, flags) < 0)
             HGOTO_ERROR(H5E_FSPACE, H5E_CANTINSERT, FAIL, "can't insert free space section into skip list");
+    }
+    else if (merged_or_shrunk)
+        *merged_or_shrunk = true;
 
 #ifdef H5FS_SINFO_DEBUG
     fprintf(stderr, "%s: fspace->tot_space = %" PRIuHSIZE "\n", __func__, fspace->tot_space);
@@ -2314,11 +2321,15 @@ done:
 herr_t
 H5FS_vfd_alloc_hdr_and_section_info_if_needed(H5F_t *f, H5FS_t *fspace, haddr_t *fs_addr_ptr)
 {
-    hsize_t hdr_alloc_size;
-    hsize_t sinfo_alloc_size;
-    haddr_t sect_addr = HADDR_UNDEF; /* address of sinfo */
-    haddr_t eoa       = HADDR_UNDEF; /* Initial EOA for the file */
-    herr_t  ret_value = SUCCEED;     /* Return value */
+    hsize_t hdr_alloc_size    = 0;
+    hsize_t sinfo_alloc_size  = 0;
+    haddr_t sect_addr         = HADDR_UNDEF; /* address of sinfo */
+    haddr_t eoa               = HADDR_UNDEF; /* Initial EOA for the file */
+    bool    allocated_header  = false;       /* Whether a free space header was allocated */
+    bool    inserted_header   = false; /* Whether a free space header was inserted into the metadata cache */
+    bool    allocated_section = false; /* Whether a free space section was allocated */
+    bool    inserted_section  = false; /* Whether a free space section was inserted into the metadata cache */
+    herr_t  ret_value         = SUCCEED; /* Return value */
 
     FUNC_ENTER_NOAPI_NOINIT
 
@@ -2367,10 +2378,12 @@ H5FS_vfd_alloc_hdr_and_section_info_if_needed(H5F_t *f, H5FS_t *fspace, haddr_t 
             /* Allocate space for the free space header */
             if (HADDR_UNDEF == (fspace->addr = H5MF_alloc(f, H5FD_MEM_FSPACE_HDR, hdr_alloc_size)))
                 HGOTO_ERROR(H5E_RESOURCE, H5E_NOSPACE, FAIL, "file allocation failed for free space header");
+            allocated_header = true;
 
             /* Cache the new free space header (pinned) */
             if (H5AC_insert_entry(f, H5AC_FSPACE_HDR, fspace->addr, fspace, H5AC__PIN_ENTRY_FLAG) < 0)
                 HGOTO_ERROR(H5E_FSPACE, H5E_CANTINIT, FAIL, "can't add free space header to cache");
+            inserted_header = true;
 
             *fs_addr_ptr = fspace->addr;
         }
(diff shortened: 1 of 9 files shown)


src/H5FSsection.c
```diff
@@ -1297,7 +1297,8 @@ H5FS__sect_merge(H5FS_t *fspace, H5FS_section_info_t **sect, void *op_data)
  *-------------------------------------------------------------------------
  */
 herr_t
-H5FS_sect_add(H5F_t *f, H5FS_t *fspace, H5FS_section_info_t *sect, unsigned flags, void *op_data)
+H5FS_sect_add(H5F_t *f, H5FS_t *fspace, H5FS_section_info_t *sect, unsigned flags, void *op_data,
+              bool *merged_or_shrunk)
```
jhendersonHDF, line 1301: Callers of this function need to know whether the free space section passed in was merged or shrunk away, even on failure. Otherwise, calling code can incorrectly assume that it needs to free the passed in section if it was allocated by the calling code and `H5FS_sect_add()` failed. In some cases, the passed in section will have already been freed, eventually resulting in a double free situation when the calling code frees the section.

src/H5FSsection.c
```diff
@@ -2453,5 +2468,57 @@ H5FS_vfd_alloc_hdr_and_section_info_if_needed(H5F_t *f, H5FS_t *fspace, haddr_t
     } /* end if */
 
 done:
+    if (ret_value < 0) {
```
jhendersonHDF, line 2471: When this function fails, remove everything that was previously inserted into the metadata cache. Otherwise, crashes or assertion failures can occur later on when trying to flush out half-initialized entries that were inserted.

src/H5MF.c
```diff
@@ -1154,15 +1167,21 @@ H5MF_xfree(H5F_t *f, H5FD_mem_t alloc_type, haddr_t addr, hsize_t size)
...
 
 #ifdef H5MF_ALLOC_DEBUG_MORE
         fprintf(stderr, "%s: Before H5FS_sect_add()\n", __func__);
 #endif /* H5MF_ALLOC_DEBUG_MORE */
 
         /* Add to the free space for the file */
-        if (H5MF__add_sect(f, alloc_type, f->shared->fs_man[fs_type], node) < 0)
+        if (H5MF__add_sect(f, alloc_type, f->shared->fs_man[fs_type], node, &section_merged_or_shunk) < 0) {
+            if (section_merged_or_shunk)
+                node = NULL;
```
jhendersonHDF, line 1182: Avoid double-freeing this section that was allocated above in case `H5MF__add_sect()` already freed it.

src/H5FSsection.c
jhendersonHDF: An alternative approach would be to pass in a `H5FS_section_info_t **`, but that seemed uglier and prone to the caller accidentally losing a pointer.

src/H5MF.c
```diff
@@ -711,8 +713,11 @@ H5MF__find_sect(H5F_t *f, H5FD_mem_t alloc_type, hsize_t size, H5FS_t *fspace, h
 #endif /* H5MF_ALLOC_DEBUG_MORE */
 
             /* Re-add the section to the free-space manager */
-            if (H5MF__add_sect(f, alloc_type, fspace, node) < 0)
+            if (H5MF__add_sect(f, alloc_type, fspace, node, NULL) < 0) {
+                node->sect_info.addr -= size;
+                node->sect_info.size += size;
```
jhendersonHDF, line 718: Reverse the modifications made above on failure.

src/H5MF.c
```diff
@@ -3059,8 +3078,13 @@ H5MF_settle_meta_data_fsm(H5F_t *f, bool *fsm_settled)
         assert(sm_fssinfo_fs_type > H5F_MEM_PAGE_DEFAULT);
         assert(sm_fssinfo_fs_type < H5F_MEM_PAGE_LARGE_SUPER);
 
-        assert(!H5_addr_defined(f->shared->fs_addr[sm_fshdr_fs_type]));
-        assert(!H5_addr_defined(f->shared->fs_addr[sm_fssinfo_fs_type]));
+        if (H5_addr_defined(f->shared->fs_addr[sm_fshdr_fs_type]))
```
jhendersonHDF, line 3082: These assertions can be triggered by a file space info message in the file's superblock extension that was specifically crafted. Throwing an error here causes the library to fail to be able to settle the MDFSM, which eventually leads to issues closing the library and leaked memory due to the metadata cache not being able to be flushed and destroyed. In Debug builds this still results in an assertion failure when specifically-crafted files are opened, but addressing that will require significant changes to the metadata cache shutdown process.

An alternative is to simply wipe out these addresses to `HADDR_UNDEF` and let the following logic take over, but this could end up dropping free space on the floor and also causes far more problems when given a specifically crafted file.

src/H5MFsection.c
```diff
@@ -360,7 +360,6 @@ H5MF__sect_simple_can_merge(const H5FS_section_info_t *_sect1, const H5FS_sectio
     assert(sect1);
     assert(sect2);
     assert(sect1->sect_info.type == sect2->sect_info.type); /* Checks "MERGE_SYM" flag */
-    assert(H5_addr_lt(sect1->sect_info.addr, sect2->sect_info.addr));
```
jhendersonHDF, line 363: With the file given in https://github.com/HDFGroup/hdf5/issues/5577, an attempt is made to place an overlapping section of file free space into the free space manager. That eventually causes this assertion to trigger. While a true fix would likely involve detecting this situation earlier on, this is probably a non-trivial effort. Since these functions are only checking whether two free space sections can be merged and not actually doing the merging, removing these assertions and just letting them return `false` seemed appropriate.

mattjala (dismissed):

src/H5FSsection.c
```diff
@@ -2453,5 +2468,57 @@ H5FS_vfd_alloc_hdr_and_section_info_if_needed(H5F_t *f, H5FS_t *fspace, haddr_t
...
+        /* Remove the free space section that was inserted into the metadata cache,
+         * making sure to free the file space that was allocated for it as well.
+         * Avoid expunging the entry, as the information needs to be kept around
+         * until we finish trying to settle the metadata free space manager(s).
+         */
+        if (inserted_section || allocated_section) {
```
fortnern: Why do we need this line?

src/H5FSsection.c
```diff
@@ -2453,5 +2468,57 @@ H5FS_vfd_alloc_hdr_and_section_info_if_needed(H5F_t *f, H5FS_t *fspace, haddr_t
...
+                                "can't remove file free space section from cache");
+            }
+        }
+
+        /* Remove the free space header that was inserted into the metadata cache,
+         * making sure to free the file space that was allocated for it as well.
+         * Avoid expunging the entry, as the information needs to be kept around
+         * until we finish trying to settle the metadata free space manager(s).
+         */
+        if (inserted_header || allocated_header) {
```
fortnern: Same here, is this one just for the assertion?

src/H5MF.c
fortnern: Could this be detected when the file space info message is decoded?

src/H5MFsection.c
fortnern: We should leave open an issue to add the correct fix, and note that these assertions should be re enabled at that point

fortnern:
Finished reviewing, 2 questions as comments, 2 as replies

src/H5MF.c
jhendersonHDF: In theory yes, however there was no obvious documentation or information I could find on which of the free space managers are and aren't allowed to have a defined address right after decoding and after asking around the opinion seemed to be that any of the managers could be defined. In this specific case however it seems the assumption is that these particular ones won't have a defined address.

src/H5MFsection.c
jhendersonHDF: Will do. Should I instead comment these out for now then?

src/H5FSsection.c
jhendersonHDF: I believe these checks are actually redundant and can be removed at this point. Previously I was removing the cache entries with expunge_entry and the "free file space" flag, in which case I had two branches with different logic where I didn't want to double free the file space with the call to H5MF_xfree followed by expunge_entry. That caused some problems and needed a switch to remove_entry instead, so I think this line and the other line are no longer relevant.

src/H5MFsection.c
fortnern: That sounds good

fortnern (approved):

mattjala (approved):

PR #5939, opened 2025-10-23
Allow use of filters that expand chunks by a large factor

### LINKED ISSUE #108: HDF5 crashes with inefficient compressors
I noticed that HDF5 crashes when I/O filters produce more data than the original dataset size.

When a dataset is created, its declared dimensions + data type are naturally honored when it comes the time to write the data with H5Dwrite. The I/O filter interface, however, allows a compressor to either return a number that’s smaller than that (in which case it successfully compressed the data) or slightly larger (in which case the compressor didn’t do a good job).

Now, let’s say we have a really bad compressor which requires 100x more room than necessary. What I observe is that HDF5 seems to truncate the data, so it’s not possible to retrieve it afterwards. In some cases, HDF5 even crashes when the dataset handle is closed.

Here’s an example I/O filter that reproduces the problem.
```
// build with 'g++ liberror.cpp -C -o libtestcrash.so -shared -fPIC -Wall -g -ggdb'
#include <hdf5.h>
#include <stdlib.h>
#include <sys/types.h>
#include <unistd.h>
#include <string.h>

extern "C" {

size_t callback(unsigned int flags, size_t cd_nelmts, const unsigned int *cd_values, size_t nbytes, size_t *buf_size, void **buf)
{
    if (flags & H5Z_FLAG_REVERSE) {
        return *buf_size;
    } else {
        char *newbuf = (char *) calloc(1000*1000, sizeof(char));
        free(*buf);
        *buf = newbuf;
        *buf_size = 1000*1000;
        return *buf_size;
    }
}

const H5Z_class2_t H5Z_UDF_FILTER[1] = {{
    H5Z_CLASS_T_VERS, 0x2112, 1, 1, "crash_filter", NULL, NULL, callback,
}};

H5PL_type_t H5PLget_plugin_type(void) { return H5PL_TYPE_FILTER; }
const void *H5PLget_plugin_info(void) { return H5Z_UDF_FILTER; }
}
```

The corresponding application code is here:
```
// build with 'g++ mainerror.cpp -o mainerror -g -ggdb -Wall -lhdf5'
// run with 'HDF5_PLUGIN_PATH=$PWD ./mainerror file.h5'
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <hdf5.h>

#define CHECK(hid) if ((hid) < 0) { fprintf(stderr, "failed @line %d\n", __LINE__); exit(1); }

int main(int argc, char **argv)
{
    if (argc != 2) {
        printf("Syntax: %s <file.h5>\n", argv[0]);
        exit(1);
    }
    hsize_t dims[2] = {10, 10};
    hid_t file_id = H5Fopen(argv[1], H5F_ACC_RDWR, H5P_DEFAULT);
    CHECK(file_id);
    hid_t space_id = H5Screate_simple(2, dims, NULL);
    CHECK(space_id);
    hid_t dcpl_id = H5Pcreate(H5P_DATASET_CREATE);
    CHECK(dcpl_id);
    CHECK(H5Pset_filter(dcpl_id, 0x2112, H5Z_FLAG_MANDATORY, 0, NULL));
    CHECK(H5Pset_chunk(dcpl_id, 2, dims));
    hid_t dset_id = H5Dcreate(file_id, "crash_dataset", H5T_STD_I8LE, space_id, H5P_DEFAULT, dcpl_id, H5P_DEFAULT);
    CHECK(dset_id);
    char *data = (char *) calloc(dims[0] * dims[1], sizeof(char));
    CHECK(H5Dwrite(dset_id, H5T_STD_I8LE, H5S_ALL, H5S_ALL, H5P_DEFAULT, data));
    CHECK(H5Dclose(dset_id));
    CHECK(H5Pclose(dcpl_id));
    CHECK(H5Sclose(space_id));
    CHECK(H5Fclose(file_id));
    free(data);
    return 0;
}
```

If you change the I/O filter code so that it allocates 10x10, or even 100x100, the problem won’t kick in.

fortnern:
When using file format 2.0 or above, always encode the size of a filtered chunk using a 64 bit (size of lengths) integer. Chunks are still limited to 2^32 - 1 bytes, but this restriction could be lifted in the future without API or file format changes.

Most of the changes are due to changing `H5D_chk_idx_info_t` to contain an `H5O_layout_t *` instead of an `H5O_layout_chunk_t *` and an `H5O_storage_chunk_t *`. This was done so the index clients could access the version of the layout message.

Fixes #108 

----

> [!IMPORTANT]
> Update chunked dataset file format to use 64-bit encoding for filtered chunk sizes in file format 2.0+, with refactoring and new tests.
> 
>   - **Behavior**:
>     - Update chunked dataset file format to use 64 bits for encoding size of filtered chunks in file format 2.0 or above.
>     - Chunks still limited to `2^32 - 1` bytes.
>   - **Code Refactoring**:
>     - Modify assertions and update macros for chunk size calculations in `H5Dbtree.c`, `H5Dbtree2.c`, `H5Dearray.c`, `H5Dfarray.c`, `H5Dchunk.c`, and `H5Dint.c`.
>     - Refactor `H5D_chk_idx_info_t` to use `H5O_layout_t` instead of separate layout and storage.
>   - **Testing**:
>     - Add `test_chunk_expand2()` in `dsets.c` to test handling of filters that expand chunks too much.
>     - Register and unregister `H5Z_FILTER_EXPAND2` for testing purposes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 06346d375d828664e7144774ddd276914a182981. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fortnern:
The failure looks unrelated to this change and may be a transient issue

jhendersonHDF:
> The failure looks unrelated to this change and may be a transient issue

The MacOS failure is a known issue that happens from time to time

test/dsets.c
```diff
@@ -10907,11 +10924,430 @@ test_chunk_expand(hid_t fapl)
...
+        }
+
+        ret_value = new_nbytes;
+    }
+
+    return ret_value;
+} /* end filter_expand2() */
+
+/*-------------------------------------------------------------------------
+ * Function: test_chunk_expand2
```
mattjala: It looks like the main case - a filter that expands chunks by a large amount that wasn't previously allowed - isn't tested.

src/H5Dpkg.h
brtnfld, line 1: You could replace the similar macros: 
H5D_BT2_COMPUTE_CHUNK_SIZE_LEN
H5D_EARRAY_COMPUTE_CHUNK_SIZE_LEN
H5D_FARRAY_COMPUTE_CHUNK_SIZE_LEN

with a single macro:

```
#define H5D_COMPUTE_CHUNK_SIZE_LEN(chunk_size_len, idx_info) \
    do { \
        if ((idx_info)->pline->nused > 0) { \
            if ((idx_info)->layout->version > H5O_LAYOUT_VERSION_4) \
                (chunk_size_len) = H5F_SIZEOF_SIZE((idx_info)->f); \
            else { \
                (chunk_size_len) = 1 + ((H5VM_log2_gen((uint64_t)(idx_info)->layout->u.chunk.size) + 8) / 8); \
                if ((chunk_size_len) > 8) \
                    (chunk_size_len) = 8; \
            } \
        } \
        else \
            (chunk_size_len) = 0; \
    } while(0)
```

test/dsets.c
fortnern: I don't follow - that's what this does. I verified it failed before I implemented the changes in the library. test_chunk_expand is a preexisting test that tests cases where it's expected to fail, but I didn't want to copy that directly because it would have resulted in files that were too big.

src/H5Dpkg.h
fortnern: The earray and farray ones are necessarily split into two. I can try to unify them on Monday though.

test/dsets.c
mattjala: In that case, I think it's just the comment here that's wrong - it says this test is evaluating error handling on expected failure, but the test itself expects every operation to succeed.

src/H5Dpkg.h
fortnern: I'm actually inclined to leave it how it is, since this only applies to earray, farray, and btree2; not btree1, single, or none. It would then be inelegant to create a shared macro that only applies to some indices. There is a spot in H5Dchunk.c that does the calculation independently that would ideally be unified with the calculations in the index codes, but the proper way to do that would be to add an index callback to retrieve the chunk size encoding length. We don't really have time for that right now though, and it only applies to obsolete file formats so it's probably not worth it.

test/dsets.c
fortnern: Oh sorry I'll fix that

test/dsets.c
```diff
@@ -10907,11 +10924,431 @@ test_chunk_expand(hid_t fapl)
...
+#define FILTER_EXPAND2_FACTOR (256 * 256)
+
+/*-------------------------------------------------------------------------
+ * Function:    filter_expand2
+ *
+ * Purpose:     For testing library's behavior when a filter expands a
+ *              chunk by an amount that would be too much for the old file
+ *              format.
+ *
+ * Note:        Unlike filter_expand, this fitler does actually allocate a
```
fortnern:
```suggestion
 * Note:        Unlike filter_expand, this filter does actually allocate a
```

mattjala (changes requested):
The 2.0+ file format docs should be updated to replace appropriate mentions of `Chunk Size (variable size; at most 8 bytes)`. This will require updating a couple of byte-layout tables.

fortnern:
> The 2.0+ file format docs should be updated to replace appropriate mentions of `Chunk Size (variable size; at most 8 bytes)`. This will require updating a couple of byte-layout tables.

Yes I'm planning to do that, but since IIRC docs changes can go in after the code freeze I'm prioritizing other things right now

mattjala (approved):

jhendersonHDF (approved):

PR #5944, opened 2025-10-24
Semantic versioning update

lrknox:
Add instruction RELEASE_PROCESS to verify library version compliance with sematic
    versioning using link to semantic versioning wiki page.
    Update H5.c and version tests for move of major and minor versions to
    1st and 2nd version numbers.

----

> [!IMPORTANT]
> Update semantic versioning process and adjust version checking logic for minor version compatibility in HDF5.
> 
>   - **Release Process**:
>     - Added step to verify HDF5 library version compliance with semantic versioning in `RELEASE_PROCESS.md`.
>     - Updated instructions for regenerating H5E header files.
>   - **Version Checking**:
>     - Renamed `VERS_RELEASE_EXCEPTIONS` to `VERS_MINOR_EXCEPTIONS` in `H5.c`.
>     - Updated `H5_check_version()` to focus on minor version compatibility instead of release version.
>     - Changed warning messages to reflect minor version mismatches.
>   - **Testing**:
>     - Removed `WILL_FAIL "true"` for minor version check in `CMakeTests.cmake` to allow minor version +1 to pass on non-develop branches.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for de3e556ccc969a6a33a6c4b878b7d51d787ee955. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/H5.c
```diff
@@ -896,17 +896,17 @@ H5_check_version(unsigned majnum, unsigned minnum, unsigned relnum)
...
 
-    } /* end if (H5_VERS_MAJOR != majnum || H5_VERS_MINOR != minnum) */
+    } /* end if (H5_VERS_MAJOR != majnum) */
 
-    /* H5_VERS_RELEASE should be compatible, we will only add checks for exceptions */
-    /* Library develop release versions are incompatible by design */
-    if (H5_VERS_RELEASE != relnum) {
-        for (unsigned i = 0; i < VERS_RELEASE_EXCEPTIONS_SIZE; i++) {
+    /* H5_VERS_MINOR should be compatible, we will only add checks for exceptions */
+    /* Library develop minor versions are incompatible by design */
```
brtnfld, line 917: For semantic versioning, minor version increments should be backward compatible, so I'm not sure why the PR treats minor version differences as exceptions that need special handling

brtnfld: The logic treats ALL minor version differences as potential incompatibilities unless the exceptions list is empty. This is backwards from semantic versioning logic.

byrnHDF: The logic is there because of history mostly. The 1.10 version had incompatible versions because of incorrect programming. The normal expected use is that all minor versions are compatible.

brtnfld: Right, we want to update the schema for semantic versioning,

As is the PR, it does not distinguish between:
 2.0.x -> 2.1.x  is ok
2.1.x -> 2.-.x should warn/fail

it should be 
(1) major must match
(2) minor version backward compatibility check
     2a, check if either version is in the incompatible list (e.g., develop)

test/CMakeTests.cmake
```diff
@@ -654,6 +654,7 @@ set_tests_properties (H5TEST-tcheck_version-major PROPERTIES
 if ("H5TEST-tcheck_version-major" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")
   set_tests_properties (H5TEST-tcheck_version-major PROPERTIES DISABLED true)
 endif ()
+# minor + 1 should pass on non-develop branches
```
brtnfld: But minor test KEEPS: WILL_FAIL "true"
Release test REMOVES: WILL_FAIL "true"

so 
```
# minor + 1 should FAIL on develop branch (minor=0 in exceptions)
# minor + 1 should PASS on release branches (minor!=0 not in exceptions)
# This test expects failure when run on develop
```

test/CMakeTests.cmake
```diff
@@ -663,12 +664,10 @@ set_tests_properties (H5TEST-tcheck_version-minor PROPERTIES
 if ("H5TEST-tcheck_version-minor" MATCHES "${HDF5_DISABLE_TESTS_REGEX}")
   set_tests_properties (H5TEST-tcheck_version-minor PROPERTIES DISABLED true)
 endif ()
-# release + 1 should pass on non-develop branches
```
brtnfld, line 666: ` #release + 1 should always PASS (release version not checked in semver)`

test/CMakeTests.cmake
brtnfld, line 663: `# Expected to fail on develop`

src/H5.c
```diff
@@ -806,8 +806,8 @@ H5get_libversion(unsigned *majnum /*out*/, unsigned *minnum /*out*/, unsigned *r
  *              version numbers compiled into the library.
  *
  *              Within major.minor.release version, the expectation
- *              is that all release versions are compatible, exceptions to
- *              this rule must be added to the VERS_RELEASE_EXCEPTIONS list.
+ *              is that all minor versions are compatible, exceptions to
+ *              this rule must be added to the VERS_MINOR_EXCEPTIONS list.
  *
```
brtnfld, line 816:
```
 *              Within major.minor.release version numbering:
 *              - Major version must match exactly (breaking changes)
 *              - Minor versions within the same major are backward compatible
 *              - Specific minor versions (e.g., develop=0) can be marked as
 *                incompatible in VERS_MINOR_EXCEPTIONS list
 *              - Release (patch) versions are always compatible (not checked)
```

src/H5.c
```diff
@@ -823,8 +823,8 @@ H5get_libversion(unsigned *majnum /*out*/, unsigned *minnum /*out*/, unsigned *r
     "linked with a different version of static or shared HDF5 library.\n"                                    \
     "You should recompile the application or check your shared library related\n"                            \
     "settings such as 'LD_LIBRARY_PATH'.\n"
-#define RELEASE_MISMATCH_WARNING                                                                             \
-    "Warning! ***HDF5 library release mismatched error***\n"                                                 \
+#define MINOR_VERSION_MISMATCH_WARNING                                                                       \
```
brtnfld, line 831: semantic versioning says minor versions should be compatible.

```
"Warning! ***HDF5 library minor version mismatch (incompatible version detected)***\n"
    "The HDF5 header or library version used is marked as incompatible.\n"
    "This typically indicates development/unstable versions are being mixed.\n"
```

brtnfld (changes requested):
See comments

test/CMakeTests.cmake
byrnHDF: Right that is why we do not need to check this condition.

brtnfld: I suggested revising  the comment

lrknox: Comment revised to "release + 1 should always pass (not checked)."

test/CMakeTests.cmake
lrknox: We don't think we can make AI follow this, so we propose to put 999 in the VERS_MINOR_EXCEPTIONS list that will trigger the warning and abort for incompatible minor versions, and test 999 in the tcheck_versions -tm test.  Any future incompatible minor versions can be added to that list (there should never be any).  The release branches can run the same test, and because WILL_FAIL is set to true, the test will pass on both develop and the release branches.  If a truly incompatible branch is added to the list of exceptions and using the incompatible branch is attempted the warning and abort will be triggered as WILL_FAIL is applied only to the test.

src/H5.c
lrknox: Yes.  We should never have occasion to use the exceptions.

test/CMakeTests.cmake
brtnfld: Sounds good, are you working on this? I don't see the changes.

lrknox:
Current status:
 
(1) major must match
(2) minor version backward compatibility check
 
     if minor versions don't match
         2a, check if either version is in the incompatible list (e.g., develop)
             print incompatibility warning and HDF5_DISABLE_VERSION_CHECK options
             abort if version check not disabled
         2b, check if header minor version is newer than library version
             print forward compatibility warning and HDF5_DISABLE_VERSION_CHECK options
             abort if version check not disabled
(3) release (patch) version not checked
(4) VERS_MINOR_EXCEPTIONS list now set to '999', and minor version '999' checked in
    test/tcheck_version.c.  This check will always cause the -tm option to fail and
    can remain in release branches, eliminating need to change or remove WILL_FAIL true
    for release branches.

src/H5.c
```diff
@@ -938,7 +953,40 @@ H5_check_version(unsigned majnum, unsigned minnum, unsigned relnum)
 
         } /* end for */
 
-    } /* end if (H5_VERS_RELEASE != relnum) */
+        /* Check for forward compatibilty usage. */
```
lrknox:
```suggestion
        /* Check for forward compatibility usage. */
```

release_docs/RELEASE_PROCESS.md
```diff
@@ -93,16 +95,17 @@ For more information on the HDF5 versioning and backward and forward compatibili
     - `$ bin/h5vers -s X.Y.Z-{SR+1};` 
     - `$ git commit -m "Updated release preparation branch version number to X.Y.Z-{SR+1}"` 
     - `$ git push` 
-7. ** OBSOLETE CURRENTLY **
+7. Remove 'WILL_FAIL "true"' line for minor version check in test/CMakeTests.cmake (currently line 662).  Minor branches are considered incompatible for develop, but not for release branches.
```
byrnHDF: This should no longer be necessary?

byrnHDF (dismissed):

release_docs/RELEASE_PROCESS.md
```diff
@@ -207,8 +210,9 @@ For more information on the HDF5 versioning and backward and forward compatibili
 [u8]: https://github.com/HDFGroup/hdf5/blob/develop/.github/workflows/release.yml
 [u9]: https://github.com/HDFGroup/hdf5/blob/develop/config/lt_vers.am
 [u10]: https://github.com/HDFGroup/hdf5/blob/develop/bin/h5vers
-[u11]: https://github.com/HDFGroup/hdf5/blob/develop/src/CMakeLists.txt
+[u11]: https://github.com/HDFGroup/hdf5/blob/desvelop/src/CMakeLists.txt
```
hyoklee: @lrknox , is it typo - desvelop?

hyoklee (changes requested):

release_docs/RELEASE_PROCESS.md
lrknox: Correct - removed

release_docs/RELEASE_PROCESS.md
lrknox:
```suggestion
[u11]: https://github.com/HDFGroup/hdf5/blob/develop/src/CMakeLists.txt
```

lrknox: Yes - fixed.  Thanks!

hyoklee:
Release process document refers this link but I don't see it updated for v200 yet as of 2025-10-30.

[u13] https://support.hdfgroup.org/documentation/hdf5/latest/api-compat-macros.html

Will it be updated after the release automatically?

lrknox:
> [u13] https://support.hdfgroup.org/documentation/hdf5/latest/api-compat-macros.html
ttps://support.hdfgroup.org/documentation/hdf5/latest/api-compat-macros.html
I don't think it will be updated automatically, but we should update it.  There were no new API compatibility macros in 1.14, so latest entries are for 1.12. For 2.0.0 there are 3 new ones:  H5Dread_chunk, H5Iregister_type, and H5Tdecode.   

@byrnHDF, I think this is a page created with doxygen?  

I'm thinking about adding a RELEASE_PROCESS.md entry to update it if needed, but it should probably be done before that.  Maybe the RELEASE_PROCESS.md task should be to check that it's updated?

Also,  this is one of 25 technical notes on https://support.hdfgroup.org/documentation/hdf5/latest/_t_n.html.  I wonder if any of the others might go out of date.  I see one named HDF5 Library Release Version Numbers that I intend to check today.

Working on updates now - almost done.  #5968

src/H5.c
```diff
@@ -938,7 +953,40 @@ H5_check_version(unsigned majnum, unsigned minnum, unsigned relnum)
 
         } /* end for */
 
-    } /* end if (H5_VERS_RELEASE != relnum) */
+        /* Check for forward compatibility usage. */
+        if (H5_VERS_MINOR > minnum) {
```
brtnfld: logic is backwards, Library minor > Header minor
Example: Library 2.1.x, Headers 2.0.x
Meaning: App compiled with OLD headers, linked with NEW library
This is backward compatible, so it should be allowed, but it aborts.

It should be `< minnum`

brtnfld (changes requested):

src/H5.c
lrknox: No, this block is for Header minor > library minor, to warn user that for example app compiled with 2.1 headers is linking to 2.0 library.  A message for this particular situation is added in lines 840 - 848.  

Allen's arguments were that there should be a warning for [accidentally] linking to an older minor version library, but that it could still work as long as functions not in the older library were not used in the App.

There is no failure for backward compatibility unless one of the two minor versions is in the list of exceptions to minor version compatibility, which basically should not happen.  Otherwise, if the Header Minor version < library minor version it will always be compatible and there is no warning, and no abort.

There is this comment (from before, just with 'release' changed to 'minor' in line 917 that I question now:  "/* Library develop minor versions are incompatible by design */"

brtnfld: Could you clarify this, and is this information correct? 

H5_VERS_MINOR = The minor version number compiled into the library
minnum = The minor version number from the headers the app was compiled with

"Header minor > library minor, to warn user that app compiled with 2.1 headers is linking to 2.0 library"

App compiled with 2.1 headers, minnum = 1
Linked with 2.0 library, H5_VERS_MINOR = 0
if (H5_VERS_MINOR > minnum) , if (0 > 1) , FALSE
Does not go into the warning block at all.

Instead it warns in the opposite case. Is the intent to warn with headers > library (forward incompat)

lrknox: Yes, you're correct, at least mostly.  I had H5_VERS_MINOR and minnum switched.  I thought minnum was the library minor version.

The conditional in line 957 should be if (minnum > H5_VERS_MINOR), as the intent is to warn that header minor version is > library minor version.  Fixing it shortly.

library minor version > minnum does not return a warning unless there is an incompatible version compared.

byrnHDF (dismissed):

release_docs/RELEASE_PROCESS.md
```diff
@@ -71,9 +72,11 @@ For more information on the HDF5 versioning and backward and forward compatibili
 6. Confirm the necessity of and approve of any interface-breaking changes. If any changes need to be reverted, task the developer who made the change to do so as soon as possible. If a change is reverted, return to the previous step and regenerate the compatibility report after the changes is made. Otherwise, continue to the next step.
 7. Update the .so version numbers in the [config/lt_vers.am][u9] file in the support branch according to [libtool's library interface version](https://www.gnu.org/software/libtool/manual/libtool.html#Versioning) scheme. 
     - See [Updating version info (Libtool)](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info) for rules to help update library version numbers. 
-8. After the release branch has been created, run `./autogen.sh` to regenerate build system files on the release branch and commit the changes.    
+8. After the release branch has been created, run bin/process_source.sh to regenerate the H5E header files on the release branch, and commit the changes.
```
hyoklee: Are H5E header files enough for committing changes?

I can see other modifications:

```
# Run make_vers
# make_vers automatically generates the public headers that define the API version
# macros for HDF5.
echo "Running API version generation script:"
bin/make_vers src/H5vers.txt || exit 1
echo

# Run make_overflow
# make_overflow automatically generates macros for detecting overflows for type
# conversion.
echo "Running overflow macro generation script:"
bin/make_overflow src/H5overflow.txt || exit 1
echo
```

lrknox: process_source.sh runs trace, make_err, make_vers, make_overflow to update H5E headers and macros for the release branch.  This is everything except autoreconf that the autogen.sh script ran before.  I've added the  macro types to be updated to the sentence about running bin/process.sh.

brtnfld (approved):

test/tcheck_version.c
```diff
@@ -70,7 +70,7 @@ parse(int ac, char **av)
                             major++;
                             break;
                         case 'm':
-                            minor++;
+                            minor = 999;
```
hyoklee, line 73: Credit: Claude.ai

## Critical Flaws with Setting Minor Version to 999

### 1. **Violates Semantic Versioning Principles**
The PR proposes adding 999 to a `VERS_MINOR_EXCEPTIONS` list to trigger warnings and abort for incompatible minor versions. This fundamentally breaks semantic versioning because:

- **Semantic versioning uses minor versions to indicate backwards-compatible features**, not incompatibility
- Using 999 as a "special incompatible marker" is a hack that contradicts the entire purpose of semver
- In proper semver (X.Y.Z), minor version changes should be safe upgrades - applications should be able to use any version 2.Y.Z without breaking

### 2. **Confusion Between Development and Release Branches**
The comments in the PR reveal confusion:
- The test expects minor + 1 to FAIL on develop branch (where minor=999) but PASS on release branches (where minor≠999)
- This creates a situation where version 2.999.0 on develop is somehow "compatible" with the exception list, but this is semantically meaningless
- Version 2.999.0 doesn't convey any meaningful information to users about API stability

### 3. **Arbitrary Magic Number**
Using 999 as a sentinel value is problematic:
- **No upper bound validation**: What happens if someone legitimately needs version 2.999.X in the future (unlikely but theoretically possible)?
- **Implicit rather than explicit**: The value 999 has no inherent meaning; it requires developers to remember it's a "magic number"
- **Better alternatives exist**: Proper semver would use major version bumps (3.0.0) for incompatible changes

### 4. **Testing Logic Issues**
The solution proposes setting `WILL_FAIL` to true for the minor version test so it passes on both develop and release branches when testing incompatible versions. This creates:
- Tests that expect failure as success (confusing)
- Reduced ability to catch actual version compatibility bugs
- Maintenance burden as developers must remember this inverted logic

### 5. **Doesn't Address Root Problem**
The HDF5 project has a long history of API breakage between what were labeled "minor" versions (1.8, 1.10, 1.12), which violated semantic versioning principles. Using 999 as a workaround for the develop branch doesn't fix the underlying issue - it just adds another layer of complexity.

## Better Alternatives

1. **Use 0 for develop**: Set minor version to 0 on develop branch (2.0.0-dev) and increment on each release
2. **Use pre-release identifiers**: Use semver pre-release syntax like 2.0.0-alpha, 2.0.0-beta.1
3. **Major version bumps**: For truly incompatible changes, use 3.0.0, 4.0.0, etc.
4. **Development builds**: Use metadata like 2.0.0+dev.20251103 to indicate development versions

The 999 approach is a workaround that adds technical debt rather than following established versioning conventions.

hyoklee (approved):
999 seems too ad-hoc.

I'll approve it with a note for future reference.

PR #5950, opened 2025-10-28
Add new size field to H5D_compact_storage_t to address buffer overflows

glennsong09:
> [!IMPORTANT]
> Add `size` field to `H5D_compact_storage_t` and update related functions to prevent buffer overflows.
> 
>   - **Behavior**:
>     - Add `size` field to `H5D_compact_storage_t` to track buffer size.
>     - Update `H5D__compact_readvv()` in `H5Dcompact.c` to check buffer size before reading.
>     - Modify `H5D__chunk_lock()` in `H5Dchunk.c` to initialize and set `alloc_chunk_size`.
>   - **Functions**:
>     - Update `H5D__chunk_lock()` to accept `size_t *alloc_chunk_size` and set it during chunk allocation.
>     - Modify `H5D__chunk_read()` and `H5D__chunk_write()` in `H5Dchunk.c` to use `alloc_chunk_size`.
>   - **Structures**:
>     - Add `size` field to `H5D_compact_storage_t` in `H5Dpkg.h` to store buffer size.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 72bdcc40e6a6b2390d92eea54ff6735860b6a680. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -307,7 +307,8 @@ static unsigned H5D__chunk_hash_val(const H5D_shared_t *shared, const hsize_t *s
 static herr_t   H5D__chunk_flush_entry(const H5D_t *dset, H5D_rdcc_ent_t *ent, bool reset);
 static herr_t   H5D__chunk_cache_evict(const H5D_t *dset, H5D_rdcc_ent_t *ent, bool flush);
 static void    *H5D__chunk_lock(const H5D_io_info_t *io_info, const H5D_dset_io_info_t *dset_info,
-                                H5D_chunk_ud_t *udata, bool relax, bool prev_unfilt_chunk);
+                                H5D_chunk_ud_t *udata, bool relax, bool prev_unfilt_chunk,
+                                size_t *alloc_chunk_size);
 static herr_t   H5D__chunk_unlock(const H5D_io_info_t *io_info, const H5D_dset_io_info_t *dset_info,
                                   const H5D_chunk_ud_t *udata, bool dirty, void *chunk, uint32_t naccessed);
 static herr_t   H5D__chunk_cache_prune(const H5D_t *dset, size_t size);
@@ -2677,6 +2678,7 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
     H5S_t             *chunk_file_spaces_local[8];  /* Local buffer for chunk_file_spaces */
     haddr_t           *chunk_addrs = NULL;          /* Array of chunk addresses */
     haddr_t            chunk_addrs_local[8];        /* Local buffer for chunk_addrs */
+    size_t             alloc_chunk_size = 0;        /* Size of allocated chunk buffer */
     herr_t             ret_value = SUCCEED;         /*return value        */
 
     FUNC_ENTER_PACKAGE
@@ -2922,11 +2924,12 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
                         (uint32_t)chunk_info->piece_points * (uint32_t)dset_info->type_info.src_type_size;
 
                     /* Lock the chunk into the cache */
-                    if (NULL == (chunk = H5D__chunk_lock(io_info, dset_info, &udata, false, false)))
+                    if (NULL == (chunk = H5D__chunk_lock(io_info, dset_info, &udata, false, false, &alloc_chunk_size)))
                         HGOTO_ERROR(H5E_IO, H5E_READERROR, FAIL, "unable to read raw data chunk");
 
                     /* Set up the storage buffer information for this chunk */
                     cpt_store.compact.buf = chunk;
+                    cpt_store.compact.size = alloc_chunk_size;
 
                     /* Point I/O info at contiguous I/O info for this chunk */
                     chk_io_info = &cpt_io_info;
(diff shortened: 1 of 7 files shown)


src/H5VM.c
```diff
@@ -1371,6 +1371,12 @@ H5VM_memcpyvv(void *_dst, size_t dst_max_nseq, size_t *dst_curr_seq, size_t dst_
     dst = (unsigned char *)_dst + *dst_off_ptr;
     src = (const unsigned char *)_src + *src_off_ptr;
 
+    /* Check if src buffer size is less than expected */
+    if (src_alloc_size > 0) {
+        if (src_alloc_size < tmp_src_len)
```
jhendersonHDF: Just checking `tmp_src_len` isn't enough here, as there may be more than one sequence and depending on which branch is taken below, either `tmp_src_len` or `tmp_dst_len` will need to be checked. You can see that `tmp_src_len` and `tmp_dst_len` are continually updated in the loops below, so you'll need to check `tmp_src_len`/`tmp_dst_len` against the remaining space left in `_src` on each iteration, which will involve keeping track of the end of `_src` and using that alongside the current offset into it (`src`). The checking can probably be done in the usual way with the `H5_IS_BUFFER_OVERFLOW` macro.

src/H5VM.c
```diff
@@ -1371,6 +1371,12 @@ H5VM_memcpyvv(void *_dst, size_t dst_max_nseq, size_t *dst_curr_seq, size_t dst_
...
+            HGOTO_ERROR(H5E_INTERNAL, H5E_CANTOPERATE, FAIL, "src buffer size is less than expected");
```
jhendersonHDF: The values are the same, but since this function returns an `ssize_t` currently, (-1) should be explicitly returned for consistency.

src/H5Dselect.c
```diff
@@ -401,7 +401,7 @@ H5D_select_io_mem(void *dst_buf, H5S_t *dst_space, const void *src_buf, H5S_t *s
 
             /* Perform vectorized memcpy from src_buf to dst_buf */
             if ((bytes_copied = H5VM_memcpyvv(dst_buf, dst_nseq, &curr_dst_seq, dst_len, dst_off, src_buf,
-                                              src_nseq, &curr_src_seq, src_len, src_off)) < 0)
+                                              src_nseq, &curr_src_seq, src_len, src_off, 0)) < 0)
```
jhendersonHDF: I'd maybe consider `SIZE_MAX` instead of 0 for an unknown value, but in this case we should probably make sure we always have a known value to pass to `H5VM_memcpyvv`.

src/H5Dchunk.c
```diff
@@ -4322,7 +4331,7 @@ H5D__chunk_cache_prune(const H5D_t *dset, size_t size)
  */
 static void *
 H5D__chunk_lock(const H5D_io_info_t H5_ATTR_NDEBUG_UNUSED *io_info, const H5D_dset_io_info_t *dset_info,
-                H5D_chunk_ud_t *udata, bool relax, bool prev_unfilt_chunk)
+                H5D_chunk_ud_t *udata, bool relax, bool prev_unfilt_chunk, size_t *alloc_chunk_size)
```
jhendersonHDF, line 4342: In this function, it looks like there are several places where the final value for `alloc_chunk_size` should be set. At least:

- After the chunk entry is retrieved from the cache on line 4372 (which means the `H5D_rdcc_ent_t` structure is probably also going to need to track the size of the allocated chunk buffer)
- After the call to `H5D__chunk_mem_alloc()` on line 4408
- After the call to `H5D__chunk_mem_alloc()` on line 4434
- After the call to `H5D__chunk_mem_alloc()` on line 4515
- After the call to `H5D__chunk_mem_alloc()` on line 4535 (as is done currently in this PR)
- After the call to `H5Z_pipeline()` on line 4553 (as is done currently in this PR, but using the value of `buf_alloc` that comes back rather than setting `alloc_chunk_size` to 0)
- After the call to `H5D__chunk_mem_alloc()` on line 4561
- After the call to `H5D__chunk_mem_alloc()` on line 4582

Due to all these places, it would probably be easiest to keep track of the value in a local variable and assign the value to `alloc_chunk_size` somewhere toward the end of the function.

src/H5VM.c
glennsong09: I made an update to this in my PR.

src/H5VM.c
glennsong09: I addressed this.

src/H5Dselect.c
glennsong09: I addressed this in my update. However, I kept two SIZE_MAX because I looked into finding their size, but does not appear to be a good way to find or easily store their size due to many places messing with the size of dset_info->buf.cvp. Do you mind taking a look and letting me know if you have any good ideas?

src/H5Dchunk.c
glennsong09: I addressed this comment.

fortnern:
Where does the invalid chunk size come from? In H5D__chunk_lock it's just taken from the layout as far as I can see so I'm not sure why that function needs to return it.

fortnern:
The size of the valid data in the chunk should be constant an predictable without needing H5D__chunk_lock to return it. Does the buffer overflow occur because the selection goes beyond the chunk or because the chunk is allocated too small? Is it the case where an unfiltered edge chunk of an otherwise filtered dataset has the wrong size stored in the index?

jhendersonHDF:
> The size of the valid data in the chunk should be constant an predictable without needing H5D__chunk_lock to return it. Does the buffer overflow occur because the selection goes beyond the chunk or because the chunk is allocated too small? Is it the case where an unfiltered edge chunk of an otherwise filtered dataset has the wrong size stored in the index?

I don't remember exactly where the invalid value exists in the file, but if I'm remembering correctly it's the case that file corruption causes `H5D__chunk_lock()` to incorrectly allocate a buffer that is too small without giving back any information about how big that buffer is. Later on, the compact I/O routines run off the buffer since there's an assumption of how big the buffer they're passed is going to be.

fortnern:
H5D__chunk_lock shouldn't need to report the size of the buffer, it should always be the same size.

PR #5956, opened 2025-10-30
Change default nslots value for chunk cache to 8191

fortnern:
In order to reduce hash collisions and take advantage of modern memory capacity, the default hash table size for the chunk cache has been increased from 521 to 8191. This means the hash table will consume approximately 64 KiB per open dataset. This value can be changed with `H5Pset_cache()` or `H5Pset_chunk_cache()`. This value was chosen because it is a prime number close to 8K.

----

> [!IMPORTANT]
> Increase default chunk cache hash table size from 521 to 8191 to reduce hash collisions and improve memory usage.
> 
>   - **Behavior**:
>     - Default hash table size for chunk cache increased from 521 to 8191 in `H5Pfapl.c` and `H5Ppublic.h`.
>     - This change reduces hash collisions and uses approximately 64 KiB per open dataset.
>     - The value can be modified using `H5Pset_cache()` or `H5Pset_chunk_cache()`.
>   - **Documentation**:
>     - Updated `CHANGELOG.md` to document the new default chunk cache hash table size.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 9c0f440c166eaedcb053eabf9aa44bbfda011ffc. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/java/test/TestH5Pfapl.java
+++ b/java/test/TestH5Pfapl.java
@@ -720,7 +720,7 @@ public class TestH5Pfapl {
         double[] rdcc_w0   = {0};
         try {
             H5.H5Pget_cache(fapl_id, null, rdcc_nelmts, rdcc_nbytes, rdcc_w0);
-            assertTrue("H5P_cache default", rdcc_nelmts[0] == 521);
+            assertTrue("H5P_cache default", rdcc_nelmts[0] == 8191);
             assertTrue("H5P_cache default", rdcc_nbytes[0] == (8 * 1024 * 1024));
             assertTrue("H5P_cache default", rdcc_w0[0] == 0.75);
         }
@@ -748,7 +748,7 @@ public class TestH5Pfapl {
         double[] rdcc_w0   = {0};
         try {
             H5.H5Pget_chunk_cache(dapl_id, rdcc_nslots, rdcc_nbytes, rdcc_w0);
-            assertTrue("H5P_chunk_cache default", rdcc_nslots[0] == 521);
+            assertTrue("H5P_chunk_cache default", rdcc_nslots[0] == 8191);
             assertTrue("H5P_chunk_cache default", rdcc_nbytes[0] == (8 * 1024 * 1024));
             assertTrue("H5P_chunk_cache default", rdcc_w0[0] == 0.75);
         }
--- a/test/cache_tagging.c
+++ b/test/cache_tagging.c
@@ -2749,7 +2749,7 @@ check_dataset_write_tags(void)
         dump_cache(fid);
 #endif /* NDEBUG */ /* end debugging functions */
 
-    /* Verify 10 b-tree nodes belonging to dataset  */
+    /* Verify 19 b-tree nodes belonging to dataset  */
     for (i = 0; i < 19; i++)
         if (verify_tag(fid, H5AC_BT_ID, d_tag) < 0)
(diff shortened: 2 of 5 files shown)


ajelenak (dismissed):

test/cache_tagging.c
```diff
@@ -2746,7 +2750,7 @@ check_dataset_write_tags(void)
 #endif /* NDEBUG */ /* end debugging functions */
 
     /* Verify 10 b-tree nodes belonging to dataset  */
```
mattjala: Nitpick: Comment should be changed `10 b-tree nodes` -> `19 b-tree nodes`

test/cache_tagging.c
```diff
@@ -2735,6 +2735,10 @@ check_dataset_write_tags(void)
     if ((H5Dwrite(did, H5T_NATIVE_INT, sid, sid, H5P_DEFAULT, data)) < 0)
         TEST_ERROR;
 
+    /* Flush dataset so all index nodes are created */
+    if (H5Dflush(did) < 0)
```
mattjala, line 2739: Why is this flush only necessary after the change?

mattjala (dismissed):

test/cache_tagging.c
fortnern: The test relied on the chunk cache evicting a specific number of chunks, and since hash table collisions cause eviction, a smaller hash table causes more evictions, and increasing the size (as we do here) causes fewer. The flush makes the test independent of the exact specifics of the chunk cache behavior (this test is for metadata cache tagging).

test/cache_tagging.c
fortnern:
```suggestion
    /* Verify 19 b-tree nodes belonging to dataset  */
```

mattjala (approved):

java/test/TestH5Pfapl.java
```diff
@@ -720,7 +720,7 @@ public void testH5P_cache()
         double[] rdcc_w0   = {0};
         try {
             H5.H5Pget_cache(fapl_id, null, rdcc_nelmts, rdcc_nbytes, rdcc_w0);
-            assertTrue("H5P_cache default", rdcc_nelmts[0] == 521);
+            assertTrue("H5P_cache default", rdcc_nelmts[0] == 8191);
```
bmribler, line 723: Should this value be a constant?  Future change would be simpler.  It's very minor, however, so it can be ignored since we're so close...

bmribler (approved):

java/test/TestH5Pfapl.java
fortnern: I mean it is already a constant in the library. I'm not sure it's worth making a constant for the "expected default value" just for the java test, but I wouldn't be opposed if someone wants to do it. I don't have time to go learn enough java to do so right now.

byrnHDF (approved):

java/test/TestH5Pfapl.java
bmribler: That's ok for now.  It's that way in other files too.  Let's finish this release first.

PR #5963, opened 2025-11-01
Fix CVE-2025-2309

### LINKED ISSUE #5869: Malformed file leads to buffer overflow during type conversion
See [CVE-2025-2309](https://nvd.nist.gov/vuln/detail/cve-2025-2309).

christian-wojek:
A malformed file can trigger bit field type conversions that can (due to missing boundary checks in the conversion step) cause a heap buffer overflow. This PR adds a check on the defined conversion to ensure it does not read beyond the size of a single bit field element. Thus, H5T__bit_copy does not result in a buffer overflow. There are several other calls to H5T__bit_copy which might be subject to a similar issue.

This PR fixes CVE-2025-2309.
Closes #5869.

----

> [!IMPORTANT]
> Adds boundary checks in `H5T__conv_b_b` to prevent heap buffer overflow, fixing CVE-2025-2309.
> 
>   - **Security Fix**:
>     - Adds boundary checks in `H5T__conv_b_b` in `H5Tconv_bitfield.c` to prevent heap buffer overflow during bit field type conversions.
>     - Validates source and destination bit field indices to ensure they do not exceed their respective sizes.
>   - **Misc**:
>     - Fixes CVE-2025-2309.
>     - Closes issue #5869.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 664e5aec5f5702b2ab0bff6dad2c48ff13795127. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fortnern:
Checks like these should go where the datatype is decoded or when it is defined by the API.

christian-wojek:
Moved the check as suggested into `H5T_convert`.

fortnern:
This check needs to take place in H5O__dtype_decode(). We don't want to keep an invalid datatype description in memory if we can avoid it. It's fine to add an assertion at conversion time, but the real check needs to be when it is decoded. I've looked through the API and it looks like those functions should never create atomic datatypes where the precision+offset overflows the size, so we only need to check when decoding from disk.

Unfortunately there doesn't seem to be a great place to put the assertion - it could go in every atomic conversion function, or possibly during the H5T_path_find() operation.

christian-wojek:
> This check needs to take place in H5O__dtype_decode(). We don't want to keep an invalid datatype description in memory if we can avoid it. It's fine to add an assertion at conversion time, but the real check needs to be when it is decoded. I've looked through the API and it looks like those functions should never create atomic datatypes where the precision+offset overflows the size, so we only need to check when decoding from disk.
> 
> Unfortunately there doesn't seem to be a great place to put the assertion - it could go in every atomic conversion function, or possibly during the H5T_path_find() operation.

@fortnern Thanks for the pointer - I finally narrowed the actual issue down to a missing sanity check on offset and precision during decoding for `H5T_BITFIELD`. Other sanity checks, e.g. for `H5T_INTEGER`, seem to be in place. Thus, I would skip the extra check during conversion.

mattjala (approved):

fortnern (approved):

PR #5965, opened 2025-11-02
Add support for 64 bit chunks, including 64 bit chunk dimensions

fortnern:
Various related changes, including refactoring part of dataset creation, and reworking how layout versions are calculated.

Needs more testing of filters that create very large chunks, but that will need code to turn it off in cases where we can't allocate >4GiB buffers. I tested this manually by increasing the expansion ration in the expand2 test in dsets, and it passed everything up until it tried to expand the datasets with H5Dset_extent() and my laptop ran out of disk space.

We should add code and testing to handle the case where the "size of size" is set less than 8 bytes. This is not a new issue, since it can be set to 2.

I will file issues for these, but I don't think they are necessary for the release.

----

> [!IMPORTANT]
> Add support for 64-bit chunk dimensions, refactor dataset creation, and rework layout version calculations.
> 
>   - **Behavior**:
>     - Support for 64-bit chunk dimensions added, removing 32-bit limitations in `H5D.c`, `H5Dbtree.c`, and `H5Dbtree2.c`.
>     - New macro `H5D_CHUNK_ENCODE_SIZE_CHECK` added to `H5Dchunk.c` to check chunk size encoding.
>     - Refactored dataset creation and layout version calculation in `H5Dchunk.c` and `H5Dint.c`.
>   - **Data Types**:
>     - Changed `uint32_t` to `hsize_t` for chunk sizes in `H5Dchunk.c`, `H5Dbtree.c`, and `H5Dbtree2.c`.
>     - Updated function signatures to use `hsize_t` for chunk sizes in `H5D.c` and `H5Dpkg.h`.
>   - **Misc**:
>     - Updated warning flags in `config/gnu-warnings/4.8` and `config/gnu-warnings/cxx-4.8`.
>     - Removed `H5D__layout_set_version` and `H5D__layout_set_latest_indexing` from `H5Dlayout.c`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 9a04ce582133f580073191271d9f5a873d629612. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/H5Dchunk.c
```diff
@@ -772,9 +820,18 @@ H5D__chunk_set_sizes(H5D_t *dset)
     assert(dset);
     assert(dset->shared->layout.u.chunk.ndims > 0);
 
+    /* In this function, some of these sizes may have already been set since they are sometimes stored in the
+     * file. If this is the case, verify the calculated sizes match the stored sizes. */
```
fortnern, line 825: This is a strange thing I found - we are storing these things but not using the stored values. I added these file integrity checks here.

src/H5Dchunk.c
```diff
@@ -811,46 +877,192 @@ H5D__chunk_set_sizes(H5D_t *dset)
...
 
     /* Set chunk sizes */
     if (H5D__chunk_set_sizes(dset) < 0)
         HGOTO_ERROR(H5E_DATASET, H5E_BADVALUE, FAIL, "unable to set chunk sizes");
-    assert((unsigned)(dset->shared->layout.u.chunk.ndims) <= NELMTS(dset->shared->layout.u.chunk.dim));
+    assert((unsigned)(layout->u.chunk.ndims) <= NELMTS(layout->u.chunk.dim));
+
+    /* Calculate layout version */
+    /* Do not downgrade version */
+    version_req = layout->version;
```
fortnern, line 922: This function took over the functions of the deleted H5D__layout_set_latest_indexing(), and now handles the main layout version calculation. We must upgrade to at least version_req, and we will upgrade to version_perf if allowed by the bounds, and up to the low bound if not.

src/H5Dlayout.c
```diff
@@ -271,191 +271,6 @@ H5D__layout_meta_size(const H5F_t *f, const H5O_layout_t *layout, bool include_c
...
-/*-------------------------------------------------------------------------
- * Function:    H5D__layout_set_version
- *
- * Purpose:     Set the version to encode a layout with.
- *
- * Return:      Non-negative on success/Negative on failure
- *-------------------------------------------------------------------------
- */
-herr_t
-H5D__layout_set_version(H5F_t *f, H5O_layout_t *layout)
```
fortnern, line 283: Functionality moved to the construct callbacks and H5D__create (for version bounds check)

src/H5Dlayout.c
```diff
@@ -271,191 +271,6 @@ H5D__layout_meta_size(const H5F_t *f, const H5O_layout_t *layout, bool include_c
...
- * Function:    H5D__layout_set_latest_indexing
- *
- * Purpose:     Set the latest indexing type for a layout message
- *
- * Return:      Non-negative on success/Negative on failure
- *
- *-------------------------------------------------------------------------
- */
-herr_t
-H5D__layout_set_latest_indexing(H5D_t *dset)
```
fortnern, line 320: Functionality moved to H5D__chunk_construct()

src/H5Pdcpl.c
```diff
@@ -434,7 +440,29 @@ H5P__dcrt_layout_enc(const void *value, void **_pp, size_t *size)
     assert(layout);
     assert(size);
 
+    /* Get the file's low_bound and high_bound */
+    if (H5CX_get_libver_bounds(&low_bound, &high_bound) < 0)
+        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, FAIL, "can't get low/high bounds from API context");
+
+    /* Determine version - only upgrade for chunked datasets that have a dimension size >32 bits */
+    if (H5D_CHUNKED == layout->type)
```
fortnern, line 448: The new layout property encoding also needs a test - I will file an issue

test/big.c
```diff
@@ -561,6 +613,48 @@ usage(void)
             (hsize_t)FAMILY_SIZE);
 }
 
+static int
+run_tests(char *filename, hid_t fapl, fsizes_t testsize, int wrt_n)
```
fortnern, line 623: Added many test cases to this test

test/tmisc.c
```diff
@@ -3851,20 +3851,10 @@ test_misc20(void)
         return;
     }
 
-    /* Verify that chunks with dimensions that are too large get rejected */
-
     /* Create a dataset creation property list */
     dcpl = H5Pcreate(H5P_DATASET_CREATE);
     CHECK(dcpl, FAIL, "H5Pcreate");
 
-    /* Try to use chunked storage for this dataset */
```
fortnern, line 3860: This is now allowed. We could cap the high bound to test that it's rejected in that case but that's already tested elsewhere.

test/dsets.c
```diff
@@ -698,6 +700,118 @@ test_simple_io(const char *driver_name, hid_t fapl)
     return FAIL;
 } /* end test_simple_io() */
 
+/*-------------------------------------------------------------------------
```
fortnern, line 703: I wrote this to test the code I wrote for scalar chunk datasets before I realized that was impossible. There was some existing code that seemed to imply it would work.

config/gnu-warnings/4.8
```diff
@@ -1,5 +1,5 @@
 # warning flags added for GCC >= 4.3
--Wlarger-than=2560
```
fortnern, line 2: Increasing these by 24 bytes to make room for the slightly bigger H5O_layout_t since uint32s changed to hsize_t

test/dsets.c
```diff
@@ -16363,8 +16537,8 @@ test_dcpl_layout_caching(H5D_layout_t layout_type)
     hid_t type_id  = H5I_INVALID_HID;
     hid_t dcpl_id  = H5I_INVALID_HID;
 
-    H5O_layout_t layout;
-    H5O_layout_t default_layout;
+    H5O_layout_t *layout         = NULL;
+    H5O_layout_t *default_layout = NULL;
```
fortnern, line 16541: The slightly bigger H5O_layout_t was causing this test to hit the stack size warning of 8192, so I switched to dynamically allocating these.

fortnern:
There appear to be some failures with the 32 bit builds, I will try to fix them tomorrow

src/H5Dcompact.c
```diff
@@ -150,7 +150,8 @@ H5D__compact_fill(const H5D_t *dset)
 /*-------------------------------------------------------------------------
  * Function:	H5D__compact_construct
  *
- * Purpose:	Constructs new compact layout information for dataset
+ * Purpose:	Constructs new compact layout information for dataset and upgrade layout version if
+ *appropriate
```
byrnHDF: whitespace

src/H5Dcontig.c
```diff
@@ -418,7 +418,8 @@ H5D__contig_check(const H5F_t *f, const H5O_layout_t *layout, const H5S_extent_t
 /*-------------------------------------------------------------------------
  * Function:	H5D__contig_construct
  *
- * Purpose:	Constructs new contiguous layout information for dataset
+ * Purpose:	Constructs new contiguous layout information for dataset and upgrade layout version if
+ *appropriate
```
byrnHDF: whitespace

src/H5Dvirtual.c
```diff
@@ -2647,6 +2650,46 @@ H5D__virtual_init_all(const H5D_t *dset)
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5D__virtual_init_all() */
 
+/*-------------------------------------------------------------------------
+ * Function:    H5D__virtual_construct
+ *
+ * Purpose:     Constructs new virtual layout information for dataset and upgrade layout version if
+ *appropriate
```
byrnHDF: whitespace

byrnHDF (dismissed):
looks consistent

src/H5Dpkg.h
```diff
@@ -178,7 +185,7 @@ typedef struct H5D_shared_t H5D_shared_t;
 
 /* Function pointers for I/O on particular types of dataset layouts */
 typedef herr_t (*H5D_layout_construct_func_t)(H5F_t *f, H5D_t *dset);
-typedef herr_t (*H5D_layout_init_func_t)(H5F_t *f, const H5D_t *dset, hid_t dapl_id);
+typedef herr_t (*H5D_layout_init_func_t)(H5F_t *f, H5D_t *dset, hid_t dapl_id, bool open);
```
fortnern: Got rid of the const here because it wasn't really being respected anyways - nested struct are modified by the init callbacks. And it was needed for H5D__chunk_init to call H5D__chunk_set_sizes.

src/H5Dchunk.c
```diff
@@ -883,7 +1103,7 @@ H5D__chunk_construct(H5F_t H5_ATTR_UNUSED *f, H5D_t *dset)
  *-------------------------------------------------------------------------
  */
 static herr_t
-H5D__chunk_init(H5F_t *f, const H5D_t *const dset, hid_t dapl_id)
```
mattjala, line 886: What does the 'open' parameter for chunk initialization mean?

src/H5Dchunk.c
```diff
@@ -2546,8 +2777,11 @@ H5D__chunk_cacheable(const H5D_io_info_t H5_ATTR_PARALLEL_USED *io_info, H5D_dse
              * need to write the fill value, then don't load the chunk into the
              * cache, just write the data to it directly.
              */
-            H5_CHECK_OVERFLOW(dataset->shared->layout.u.chunk.size, uint32_t, size_t);
-            if ((size_t)dataset->shared->layout.u.chunk.size > dataset->shared->cache.chunk.nbytes_max) {
+            /* If the chunk is too big to fit in size_t, assume it is too big to fit in cache */
+            if (!(((hsize_t)((size_t)dataset->shared->layout.u.chunk.size) ==
+                   dataset->shared->layout.u.chunk.size) &&
+                  ((size_t)dataset->shared->layout.u.chunk.size <=
```
mattjala, line 2780: It looks like the comparison between chunk size and `nbytes_max` was switch from `>` to `<=`. Was this intentional? And if I'm understanding the context right, should the logical operator here be OR instead or AND - so that either the size being too big to fit in size_t, or it exceeding nbytes_max, will resolve true?

src/H5Dchunk.c
```diff
@@ -2634,8 +2875,11 @@ H5D__chunk_may_use_select_io(H5D_io_info_t *io_info, const H5D_dset_io_info_t *d
             if (!(io_info->using_mpi_vfd && (H5F_ACC_RDWR & H5F_INTENT(dataset->oloc.file)))) {
 #endif /* H5_HAVE_PARALLEL */
                 /* Check if the chunk is too large to keep in the cache */
-                H5_CHECK_OVERFLOW(dataset->shared->layout.u.chunk.size, uint32_t, size_t);
-                if ((size_t)dataset->shared->layout.u.chunk.size <= dataset->shared->cache.chunk.nbytes_max) {
+                /* If the chunk is too big to fit in size_t, assume it is too big to fit in cache */
+                if (((hsize_t)((size_t)dataset->shared->layout.u.chunk.size) ==
+                     dataset->shared->layout.u.chunk.size) &&
```
mattjala, line 2877: Again, it seems like this should be OR instead

src/H5Dchunk.c
```diff
@@ -5055,13 +5303,14 @@ H5D__chunk_allocate(const H5D_t *dset, bool full_overwrite, const hsize_t old_di
             if (H5Z_pipeline(pline, 0, &filter_mask, err_detect, filter_cb, &orig_chunk_size, &buf_size,
                              &fb_info.fill_buf) < 0)
                 HGOTO_ERROR(H5E_PLINE, H5E_WRITEERROR, FAIL, "output pipeline failed");
-#if H5_SIZEOF_SIZE_T > 4
-            /* Check for the chunk expanding too much to encode in a 32-bit value */
-            if (orig_chunk_size > ((size_t)0xffffffff))
-                HGOTO_ERROR(H5E_DATASET, H5E_BADRANGE, FAIL, "chunk too large for 32-bit length");
```
mattjala, line 5061: Should this error still be thrown in the case that the layout version is < 4?

mattjala: Or more specifically, in the case that the layout version is < 4 and the user set file format version bounds that prevent upgrading.

mattjala (dismissed):
Besides clarifying questions, LGTM

src/H5Dchunk.c
fortnern: Whether it's being called as part of a dataset open operation (as opposed to create)

src/H5Dchunk.c
fortnern: I think that was part of my aborted attempt to align this condition with the later one you also commented on. As it is they're opposite. I'll fix it.

src/H5Dchunk.c
fortnern: I think this is correct - it's:
if (the size doesn't overflow AND the size is less than cacheable maximum)
THEN disable selection I/O (because it will use the chunk cache and the chunk cache isn't supported with selection I/O)

src/H5Dchunk.c
fortnern: Yes I think this needs a H5D_CHUNK_ENCODE_SIZE_CHECK. Thanks!

src/H5Dchunk.c
fortnern: Actually no this is correct - note the whole thing is inverted by the '!' at the beginning

src/H5Ppublic.h
```diff
@@ -6445,14 +6445,20 @@ H5_DLL herr_t H5Pset_alloc_time(hid_t plist_id, H5D_alloc_time_t alloc_time);
...
- *       - The maximum number of elements in a chunk is 2<sup>32</sup>-1 which
- *         is equal to 4,294,967,295. If the number of elements in a chunk is
- *         set via H5Pset_chunk() to a value greater than 2<sup>32</sup>-1,
- *         then H5Pset_chunk() will fail.
- *       - The maximum size for any chunk is 4GB. If a chunk that is larger
- *         than 4GB attempts to be written with H5Dwrite(), then H5Dwrite()
- *         will fail.
+ *       defined with 10x10 chunks.
+ *
+ * \note As of HDF5 version 2.0.0, creation of datasets with arbitrarily large
```
jhendersonHDF: Should we avoid the word "arbitrarily" in several places, as the size is still currently limited to 64-bit sizes?

src/H5Dchunk.c
fortnern: Again the intention was to keep the logic the same between here and that in may_use_select_io, just inverted.

release_docs/CHANGELOG.md
```diff
@@ -193,6 +193,10 @@ All other HDF5 library CMake options are prefixed with `HDF5_`
 
 ## Library
 
+### Added support for large chunks
+
+   The library now supports arbitrarily large chunks (64 bit sizes). Using chunks with size >= 4 GiB will upgrade the file format and prevent the dataset from being opened with earlier versions of the library. 32 bit systems will not be able to use these chunks in all circumstances, such as with data filters or a fill value.
```
jhendersonHDF: Same here as well. "arbitrarily" seems in conflict with "64 bit sizes"

src/H5Dpkg.h
```diff
@@ -105,10 +105,17 @@
...
-                                                                          (DINFO)->type_info.dst_type_size); \
+         * size.  This will only be used if we must allocate a type conversion buffer for the entire I/O.    \
+         * Make sure to check for overflow and disable selection I/O if it happens. */                       \
+        if (!(PIECE_INFO)->in_place_tconv) {                                                                 \
+            hsize_t tconv_buf_hsize;                                                                         \
+            H5_CHECKED_ASSIGN(tconv_buf_hsize, hsize_t, (IO_INFO)->tconv_buf_size, size_t);                  \
+            tconv_buf_hsize += (PIECE_INFO)->piece_points *                                                  \
+                               MAX((DINFO)->type_info.src_type_size, (DINFO)->type_info.dst_type_size);      \
+            (IO_INFO)->tconv_buf_size = (size_t)tconv_buf_hsize;                                             \
+            if (H5_UNLIKELY((hsize_t)(IO_INFO)->tconv_buf_size != tconv_buf_hsize))                          \
```
brtnfld: Does this detect when hsize_t itself overflows during the addition at line 113?

brtnfld: On 64-bit systems where sizeof(hsize_t) == sizeof(size_t) == 8:
The addition overflows first, wrapping tconv_buf_hsize
The wrapped value gets cast to size_t and back
The comparison is wrapped_value != wrapped_value -> FALSE
Overflow goes undetected

brtnfld: Is there some constraint that prevents the multiplication from overflowing?

brtnfld:
```
if (!(PIECE_INFO)->in_place_tconv) {                                                                 \
            hsize_t tconv_buf_hsize;                                                                         \
            hsize_t old_size;                                                                                \
            H5_CHECKED_ASSIGN(tconv_buf_hsize, hsize_t, (IO_INFO)->tconv_buf_size, size_t);                  \
            old_size = tconv_buf_hsize;                                                                      \
            tconv_buf_hsize += (PIECE_INFO)->piece_points *                                                  \
                               MAX((DINFO)->type_info.src_type_size, (DINFO)->type_info.dst_type_size);      \
            if (H5_UNLIKELY(tconv_buf_hsize < old_size || tconv_buf_hsize > SIZE_MAX))                       \
                (IO_INFO)->tconv_buf_overflow = true;                                                        \
            else                                                                                             \
                (IO_INFO)->tconv_buf_size = (size_t)tconv_buf_hsize;                                        \
 }
```

fortnern: It probably wouldn't handle a 64 bit value overflowing correctly, but I'm sure this would be far from the only place in the library that does so.

fortnern: By the time we need to start worrying about that we'll be doing calculations in 128 bit and checking for overflow that way

mattjala (dismissed):

src/H5Dpkg.h
brtnfld: Is that true now that we have complex support? So you don't expect multi-exabyte buffer accumulations until 128-bit systems? the check is two lines, it does not seem like a big deal to check it.

src/H5Dchunk.c
```diff
@@ -3998,11 +4240,7 @@ H5D__chunk_flush_entry(const H5D_t *dset, H5D_rdcc_ent_t *ent, bool reset)
             if (H5Z_pipeline(&(dset->shared->dcpl_cache.pline), 0, &(udata.filter_mask), err_detect,
                              filter_cb, &nbytes, &alloc, &buf) < 0)
                 HGOTO_ERROR(H5E_DATASET, H5E_CANTFILTER, FAIL, "output pipeline failed");
-#if H5_SIZEOF_SIZE_T > 4
-            /* Check for the chunk expanding too much to encode in a 32-bit value */
-            if (nbytes > ((size_t)0xffffffff))
```
jhendersonHDF, line 4003: Similar to the comment above, just highlighting a few places for review in case they also need a `H5D_CHUNK_ENCODE_SIZE_CHECK` check. Here

src/H5Dchunk.c
```diff
@@ -5196,14 +5449,8 @@ H5D__chunk_allocate(const H5D_t *dset, bool full_overwrite, const hsize_t old_di
                                      &fb_info.fill_buf_size, &fb_info.fill_buf) < 0)
                         HGOTO_ERROR(H5E_PLINE, H5E_WRITEERROR, FAIL, "output pipeline failed");
 
-#if H5_SIZEOF_SIZE_T > 4
-                    /* Check for the chunk expanding too much to encode in a 32-bit value */
-                    if (nbytes > ((size_t)0xffffffff))
```
jhendersonHDF, line 5201: Here as well

src/H5Dchunk.c
```diff
@@ -6738,12 +6985,8 @@ H5D__chunk_copy_cb(const H5D_chunk_rec_t *chunk_rec, void *_udata)
         if (H5Z_pipeline(pline, 0, &(udata_dst.filter_mask), H5Z_NO_EDC, filter_cb, &nbytes, &buf_size,
                          &buf) < 0)
             HGOTO_ERROR(H5E_PLINE, H5E_CANTFILTER, H5_ITER_ERROR, "output pipeline failed");
-#if H5_SIZEOF_SIZE_T > 4
-        /* Check for the chunk expanding too much to encode in a 32-bit value */
-        if (nbytes > ((size_t)0xffffffff))
```
jhendersonHDF, line 6743: And here

jhendersonHDF (commented):
Generally looks good, just a few comments around potential checks needed

src/H5Dchunk.c
fortnern: This is not necessary here because the check will be performed when the chunk is allocated on disk in H5D__chunk_file_alloc

src/H5Dchunk.c
fortnern: Now that I look at it again, this is unnecessary since it's checked later on in H5D__chunk_file_alloc

src/H5Dchunk.c
fortnern: Same

src/H5Dchunk.c
fortnern: Same

src/H5Dpkg.h
fortnern: I'm not sure I follow re: complex support. I suppose checks like this could potentially guard against wildly invalid input data. I'll add it if you want me to, just be aware that there are many other places that don't guard against uin64 overflow (I'm not sure I've seen anywhere that does), so the check will likely have no effect until we add checks in most of those other places as well.

fortnern: Check added

src/H5Dchunk.c
fortnern: From the perspective of the layout client open== true means the construct callback was not called

jhendersonHDF (dismissed):

mattjala (dismissed):

fortnern:
I realized the H5D_CHUNK_ENCODE_SIZE_CHECK in H5D__chunk_direct_write is also redundant so I removed that too. I'll leave the macro for now even though it's only used in H5D__chunk_file_alloc

mattjala (approved):

lrknox (approved):

PR #5974, opened 2025-11-04
Update macos-13 to macos-14.

lrknox:


byrnHDF (approved):

byrnHDF:
macos-latest won't work?

hyoklee (approved):

lrknox:
> macos-latest won't work?

Probably would - macos-13 was having a brownout, and macos-14 is supposed to be available.

jhendersonHDF (approved):

PR #6011, opened 2025-11-10
Remove unused macro H5_MY_PKG_ERR

fortnern:
This macro used to be used for the FUNC_ENTER/FUNC_LEAVE machinery but is no longer used anywhere, so this PR removes it from all the module headers,

----

> [!IMPORTANT]
> Remove unused `H5_MY_PKG_ERR` macro from multiple module headers as it is no longer used.
> 
>   - **Removal**:
>     - Remove `#define H5_MY_PKG_ERR` from `H5ACmodule.h`, `H5Amodule.h`, and `H5B2module.h`.
>     - Also removed from `H5Bmodule.h`, `H5CXmodule.h`, and 28 other module headers.
>   - **Purpose**:
>     - `H5_MY_PKG_ERR` was part of the FUNC_ENTER/FUNC_LEAVE machinery but is no longer used anywhere in the codebase.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for a3786da9c07f4c246848687a5474d02834e60f67. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5ACmodule.h
+++ b/src/H5ACmodule.h
@@ -23,7 +23,6 @@
  */
 #define H5AC_MODULE
 #define H5_MY_PKG      H5AC
-#define H5_MY_PKG_ERR  H5E_CACHE
 #define H5_MY_PKG_INIT YES
 
 #endif /* H5ACmodule_H */
--- a/src/H5Amodule.h
+++ b/src/H5Amodule.h
@@ -23,7 +23,6 @@
  */
 #define H5A_MODULE
 #define H5_MY_PKG      H5A
-#define H5_MY_PKG_ERR  H5E_ATTR
 #define H5_MY_PKG_INIT YES
 
 /** \page H5A_UG HDF5 Attributes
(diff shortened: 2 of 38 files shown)


mattjala (approved):

derobins:
If you changed the error lines to use these macros instead of deleting them, you'd get more uniform error messages and eliminate copy-paste errors when people copy across packages.

glennsong09 (approved):

fortnern:
> If you changed the error lines to use these macros instead of deleting them, you'd get more uniform error messages and eliminate copy-paste errors when people copy across packages.

If we're going to go through the trouble of changing all the error messages in the library I'm more inclined to remove the major/minor codes altogether, unless there's a use for them I don't know about. But if we decide we want to go with the H5_MY_PKG_ERR we can always revert this change. Since these do nothing right now I think it's best to remove them.

brtnfld (approved):

Issue #6024, opened 2025-11-11
Bump release version.


brtnfld:
Bump HDF5 version information to 2.1, as new Fortran SWMR APIs were added,.

hyoklee:
#6172 
#6147

PR #6028, opened 2025-11-12
Increment minor version:  set version 2.1.0.

lrknox:
New Fortran SWMR APIs were added;  next release must increment minor version.

#6024 

----

> [!IMPORTANT]
> Incremented HDF5 library version from 2.0.1 to 2.1.0 across documentation, configuration, and Java files.
> 
>   - **Version Update**:
>     - Incremented version from `2.0.1` to `2.1.0` in `README.md`, `HDF5config.cmake`, and `H5.java`.
>     - Updated `LIB_VERSION` array in `H5.java` and `TestH5.java` to `{2, 1, 0}`.
>     - Changed `H5_VERS_MINOR` to `1` and `H5_VERS_RELEASE` to `0` in `H5public.h`.
>   - **Documentation**:
>     - Updated version references in `CHANGELOG.md` to `2.1.0`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 4151e96e8da4403c0f67fc43edd026a639a6876a. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config/examples/HDF5AsSubdirMacros.cmake
+++ b/config/examples/HDF5AsSubdirMacros.cmake
@@ -17,7 +17,7 @@
 # and build it.  The HDF5 options should be set after the FetchContent_Declare command and before
 # the add_subdirectory command..
 macro (EXTERNAL_HDF5_LIBRARY compress_type)
-  set (HDF5_VERSION "2.0.1")
+  set (HDF5_VERSION "2.1.0")
   set (HDF5_VERSEXT "")
   set (HDF5_VERSION_MAJOR "2.0")
   set (HDF5LIB_TGZ_NAME "hdf5.tar.gz" CACHE STRING "Use HDF5LIB from compressed file" FORCE)
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -1,4 +1,4 @@
-HDF5 version 2.0.1 currently under development
+HDF5 version 2.1.0 currently under development
 
 # 🔺 HDF5 Changelog
 All notable changes to this project will be documented in this file. This document describes the differences between this release and the previous
@@ -21,7 +21,7 @@ For releases prior to version 2.0.0, please see the release.txt file and for mor
 * [Platforms Tested](CHANGELOG.md#%EF%B8%8F-platforms-tested)
 * [Known Problems](CHANGELOG.md#-known-problems)
 
-# 🔆 Executive Summary: HDF5 Version 2.0.1
+# 🔆 Executive Summary: HDF5 Version 2.1.0
 
 ## Performance Enhancements:
 
--- a/src/H5public.h
+++ b/src/H5public.h
@@ -65,11 +65,11 @@
 /**
  * For minor interface/format changes
  */
-#define H5_VERS_MINOR 0
+#define H5_VERS_MINOR 1
 /**
  * For tweaks, bug-fixes, or development
  */
-#define H5_VERS_RELEASE 1
+#define H5_VERS_RELEASE 0
 /**
  * For pre-releases like \c snap0. Empty string for official releases.
  */
@@ -77,11 +77,11 @@
 /**
  * Short version string
  */
-#define H5_VERS_STR "2.0.1"
+#define H5_VERS_STR "2.1.0"
 /**
  * Full version string
  */
-#define H5_VERS_INFO "HDF5 library version: 2.0.1"
+#define H5_VERS_INFO "HDF5 library version: 2.1.0"
 
 #define H5check() H5check_version(H5_VERS_MAJOR, H5_VERS_MINOR, H5_VERS_RELEASE)
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-HDF5 version 2.0.1 currently under development
+HDF5 version 2.1.0 currently under development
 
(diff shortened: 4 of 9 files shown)


release_docs/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-HDF5 version 2.0.1 currently under development
+HDF5 version 2.1.0 currently under development
```
brtnfld, line 1: Why do we need the "currently under development"? It is evident from examining the current release.

byrnHDF (approved):

src/H5public.h
```diff
@@ -65,23 +65,23 @@
...
+#define H5_VERS_RELEASE 0
 /**
  * For pre-releases like \c snap0. Empty string for official releases.
  */
 #define H5_VERS_SUBRELEASE ""
 /**
  * Short version string
  */
-#define H5_VERS_STR "2.0.1"
+#define H5_VERS_STR "2.1.0"
```
brtnfld, line 80: Why can't we use a single set of #define H5_VERS_* and utilize those in the remaining #defines instead of the version number being hardcoded in multiple places across the various flavors?

brtnfld:
I would much rather we move to a single source of truth for version numbers, rather than our current situation. A lot of this could be addressed by using templated files when we need version information.

config/examples/HDF5AsSubdirMacros.cmake
```diff
@@ -17,7 +17,7 @@
 # and build it.  The HDF5 options should be set after the FetchContent_Declare command and before
 # the add_subdirectory command..
 macro (EXTERNAL_HDF5_LIBRARY compress_type)
-  set (HDF5_VERSION "2.0.1")
+  set (HDF5_VERSION "2.1.0")
```
brtnfld, line 20: Doesn't CMake already have, or can it get, this information?

config/examples/HDF5AsSubdirMacros.cmake
```diff
@@ -17,7 +17,7 @@
...
   set (HDF5_VERSEXT "")
   set (HDF5_VERSION_MAJOR "2.0")
```
brtnfld, line 22: same here

brtnfld (changes requested):
See comments

src/H5public.h
jhendersonHDF: If we just keep the `H5_VERS_(MAJOR/MINOR/RELEASE)` macros, it should be straightforward to form the others from those

config/examples/HDF5AsSubdirMacros.cmake
jhendersonHDF: It's a bit awkward, but in this case `HDF5_VERSION` may not be available as the library is being obtained with FetchContent. However, the only use seems to be in a status message that could be printed later anyway, so this occurrence could probably be removed.

config/examples/HDF5AsSubdirMacros.cmake
jhendersonHDF: And if it's not removed, it looks like `HDF5_VERSION_MAJOR` needs to be updated anyway. Plus the naming doesn't make much sense, as it includes both the major and minor versions

lrknox: I suspect this is a carryover of CMake code from pre-2.0.0 days.  HDF5_VERSION_MAJOR is set to ${HDF5_PACKAGE_VERSION_MAJOR} which is set to ${H5_VERS_MAJOR}.${H5_VERS_MINOR}.  

There are no occurrences of HDF5_PACKAGE_VERSION_RELEASE nor of HDF5_VERSION_RELEASE.  

Prior to 2.0 the major version was 1.x and there are 101 comparisons of HDF5_VERSION_MAJOR to versions "1.8" through "2.0".  16 of those are with "2.0".  Would it be appropriate and improvements to replace them?

HDF5_PACKAGE_VERSION_MAJOR => HDF5_PACKAGE_VERSION_MAJ_MIN
HDF5_VERSION_MAJOR => HDF5_VERSION_MAJ_MIN


HDF5_VERSION_MINOR is set to ${HDF5_PACKAGE_VERSION_MINOR}, but neither are used except for in config/README.md.cmake.in where it appears they are used incorrectly (once in the line that creates the incorrect link that currently causes linkchecker to fail).  Maybe these 2 should be removed.

HDF5_PACKAGE_VERSION_RELEASE and HDF5_VERSION_RELEASE don't exist.

release_docs/CHANGELOG.md
lrknox: This line appears at the top of the CHANGELOG.md file and the top-level README.md file.  The release script would change it to HDF5 version 2.1.0 released on 2026-05-15.  (whatever the release date would be.  

If the release date is not needed in downloaded code (and maybe binaries) is not needed, and the version is not needed the line can be removed from these files.  (CHANGELOG.md currently has the version in the Executive Summary section.)  

The version number and the release date could also be moved elsewhere in the README.md file.

Alternatively, the first half of the string could be kept and the release date added to it when the release occurs.

PR #6037, opened 2025-11-15
Fix location of EXTERNAL_HDF5_STATUS

dg0yt:
> [!IMPORTANT]
> Fixes the path of `EXTERNAL_HDF5_STATUS` macro in documentation files for correct reference.
> 
>   - **Documentation**:
>     - Corrected the path of `EXTERNAL_HDF5_STATUS` macro in `CHANGELOG.md` and `release_archive.txt` from `config/examples/HDF5SubdirMacros.cmake` to `config/examples/HDF5AsSubdirMacros.cmake`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 6ac7ec4194e3f3e2c2d9eb088bcbf0210a097e48. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -95,7 +95,7 @@ We would like to thank the many HDF5 community members who contributed to HDF5 2
 
    The variables used in hdf5-config.cmake to indicate what options were used to build the installed library have been renamed. All `HDF5_BUILD/ENABLE_{feature}` variables are now `HDF5_PROVIDES_{feature}`. This more clearly indicates that these variables reflect whether the feature is supported by the installed library, instead of whether the feature is an option that can be changed when building an application with the library.
 
-   Created macro `EXTERNAL_HDF5_STATUS` to convert between the old and new names. The macro is in the config/examples/HDF5SubdirMacros.cmake file and can be copied into a project's CMakeLists.txt file to provide backward compatibility.
+   Created macro `EXTERNAL_HDF5_STATUS` to convert between the old and new names. The macro is in the config/examples/HDF5AsSubdirMacros.cmake file and can be copied into a project's CMakeLists.txt file to provide backward compatibility.
 
 ### CMake minimum version is now 3.26
 
--- a/release_docs/release_archive.txt
+++ b/release_docs/release_archive.txt
@@ -57,7 +57,7 @@ New Features
       an application with the library.
 
       Created MACRO EXTERNAL_HDF5_STATUS to convert between the old and new names.
-      The macro is in the config/examples/HDF5SubdirMacros.cmake file and can be
+      The macro is in the config/examples/HDF5AsSubdirMacros.cmake file and can be
       copied into a project's CMakeLists.txt file to provide backward compatibility.
 
     - CMake minimum version is now 3.26


dg0yt:
Apart from the broken file reference:

https://github.com/HDFGroup/hdf5/blob/143f04a0c0cdf7a1ba8ccafe7f62802a813402af/config/examples/HDF5AsSubdirMacros.cmake#L245-L253

Is this macro intented to set the corresponding variables to the *name* or the *value* of the original variables? AFAICS it now provides the *name*, which means it is not enough to call the macro to make code from the other version working.
To provide the values:
~~~diff
-set (${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN     ${HDF5_PACKAGE_NAME}_BUILD_FORTRAN)
+set (${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN     "${${HDF5_PACKAGE_NAME}_BUILD_FORTRAN}")
~~~

byrnHDF (approved):

lrknox (approved):

byrnHDF:
From CMake documentation:
A variable reference has the form ${<variable>} and is evaluated inside a [Quoted Argument]or an [Unquoted Argument]

Therefore those set statements are correct

dg0yt:
~~~cmake
cmake_minimum_required(VERSION 3.10)

set (HDF5_PACKAGE_NAME "DEMO")
message(STATUS "HDF5_PACKAGE_NAME: ${HDF5_PACKAGE_NAME}")

set (${HDF5_PACKAGE_NAME}_BUILD_FORTRAN  "EXPECTED")
message(STATUS "${HDF5_PACKAGE_NAME}_BUILD_FORTRAN: ${${HDF5_PACKAGE_NAME}_BUILD_FORTRAN}")

set (${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN     ${HDF5_PACKAGE_NAME}_BUILD_FORTRAN)
message(STATUS "${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN: ${${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN}")
~~~

~~~
$ cmake -P test.cmake 
-- HDF5_PACKAGE_NAME: DEMO
-- DEMO_BUILD_FORTRAN: EXPECTED
-- DEMO_PROVIDES_FORTRAN: DEMO_BUILD_FORTRAN
~~~
Ouch! Didn't get "EXPECTED".

Hint: It is not inside an `if` expression.

dg0yt:
> From CMake documentation: A variable reference has the form ${} and is evaluated inside a [Quoted Argument]or an [Unquoted Argument]

`${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN` is not a variable reference.

byrnHDF:
> `${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN` is not a variable reference.

OH, it is missing the outer parens!  SB: `${${HDF5_PACKAGE_NAME}_PROVIDES_FORTRAN}`

byrnHDF:
It means that HDF5Examples/config/cmake/HDF5ExamplesMacros.cmake is also wrong! And only for non-2.0 binaries.

lrknox:
Created issue #6041 "Variables set in HDF5AsSubdirMacros.cmake are missing syntax (${ ... } to set correct values" for the missing syntax.

PR #6039, opened 2025-11-17
Minor optimizations of r-tree implementation

fortnern:
Two optimizatins:

1. When bulk loading the rtree, instead of calculating the bounds of each node by iterating over all child leaf nodes (doing this even for nodes whose children aren't leaves), iterate only over direct children after child nodes are created. This does mean the bounds are invalid until later in the bulk loading process, but this info isn't currently used until search. If we change the bulk loading algorithm to need this info (for example when deciding on the dimension to split along). we may need to revert this change.
2. Split the loop in the search algorithm into one for when children are internal nodes and one when they are leaves. This avoids checking the children_are_leaves flag on every iteration.

----

> [!IMPORTANT]
> Optimize R-tree bulk loading and search in `H5RT.c` by refining bounds calculation and splitting search loops.
> 
>   - **Bulk Loading Optimization**:
>     - In `H5RT__bulk_load()`, calculate node bounds by iterating over direct children after child nodes are created, instead of iterating over all child leaf nodes.
>     - Bounds are invalid until later in the bulk loading process, but this is acceptable as bounds are not used until search.
>   - **Search Optimization**:
>     - In `H5RT__search_recurse()`, split the loop into two: one for internal nodes and one for leaves, avoiding the need to check `children_are_leaves` flag on every iteration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 12f744e5699f9f8be0d44aa390373aef97e7653b. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5RT.c
+++ b/src/H5RT.c
@@ -363,27 +363,27 @@ H5RT__bulk_load(H5RT_node_t *node, int rank, H5RT_leaf_t *leaves, size_t count, 
     assert(prev_sort_dim >= -1);
     assert(rank >= 1 && rank <= H5S_MAX_RANK);
 
+    /* Compute the max/min bounds of the provided node */
+    /* Initial values */
+    for (int i = 0; i < rank; i++) {
+        node->min[i] = leaves[0].min[i];
+        node->max[i] = leaves[0].max[i];
+    }
+    /* Compute max/min from leaves */
+    for (size_t i = 0; i < count; i++) {
+        for (int d = 0; d < rank; d++) {
+            if (leaves[i].min[d] < node->min[d])
+                node->min[d] = leaves[i].min[d];
+            if (leaves[i].max[d] > node->max[d])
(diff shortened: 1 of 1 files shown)


mattjala (dismissed):

fortnern:
Surprisingly, I've found that the read time improved with this change (Jordan thought this wouldn't make a difference due to compiler optimization). More surprisingly, I found the open time got worse, indicating that the bulk loading optimization doesn't help, even though it seemingly moves (part of the algorithm) from O(N log N) to O(N ). Maybe something with cache lines? I'll take that part out.

mattjala (approved):

glennsong09 (approved):

Issue #6051, opened 2025-11-23
MSVC build fails on CMake re-configure due to cached C99 complex type size (Cache pollution)


xorespesp:
**Describe the bug:**

When building HDF5 v2.0.0 on Windows with MSVC, the first clean configuration works correctly. However, if `CMakeLists.txt` is modified or a re-configuration is triggered within the same build directory(e.g., touching `CMakeLists.txt` in IDE after a successful first build), the build fails with syntax errors in `H5private.h` regarding `H5_float_complex`.

```
...\src\H5private.h(592): error C2061: syntax error: identifier 'H5_float_complex'
```

**Environment:**

- HDF5 Version: 2.0.0
- OS: Windows 11 24H2 x64
- Compiler: Visual Studio 2022 (MSVC) 17.14.21
- CMake version: 4.1.1 (generator: Ninja)

**Root Cause Analysis:**

This happens because `H5_HAVE_C99_COMPLEX_NUMBERS` is incorrectly set to `1` during CMake re-configuration.
In `config/ConfigureChecks.cmake`, there is a "[safety block](https://github.com/HDFGroup/hdf5/blob/a6ff8aed236ee1e1deff6415e88b16c42b22f17c/config/ConfigureChecks.cmake#L902-L924)" for MSVC to fallback to `_Fcomplex` and unset the C99 complex flag:

```cmake
  ...
  if (MSVC AND NOT ${HDF_PREFIX}_SIZEOF_FLOAT_COMPLEX AND NOT ${HDF_PREFIX}_SIZEOF_DOUBLE_COMPLEX
      AND NOT ${HDF_PREFIX}_SIZEOF_LONG_DOUBLE_COMPLEX)
    # If using MSVC, the _Complex types (if available) are _Fcomplex, _Dcomplex and _Lcomplex.
    # The standard types are checked for first in case MSVC uses them in the future or in case
    # the compiler used is simulating MSVC and uses the standard types.
    cmake_push_check_state ()
    list (APPEND CMAKE_EXTRA_INCLUDE_FILES complex.h)
    HDF_CHECK_TYPE_SIZE ("_Fcomplex" ${HDF_PREFIX}_SIZEOF__FCOMPLEX)
    HDF_CHECK_TYPE_SIZE ("_Dcomplex" ${HDF_PREFIX}_SIZEOF__DCOMPLEX)
[...]

Issue #5940, opened 2025-10-24
Build Failure on Windows-11 Visual Studio 18 2026


hyoklee:
This issue is for the future HDF5 release since VS 18 2026 is still under insider preview. 
I hope this issue can help VS 2026 users and AIs to avoid asking the same issue to HelpDesk.

```
cmake -DCMAKE_TOOLCHAIN_FILE=C:/src/vcpkg.microsoft/scripts/buildsystems/vcpkg.cmake -DBUILD_SHARED_LIBS:BOOL=OFF  -DHDF5_ENABLE_ZLIB_SUPPORT:BOOL=ON -DHDF5_ENABLE_SZIP_SUPPORT:BOOL=OFF  -DCTEST_DROP_SITE_INIT:STRING="my.cdash.org"  -DSITE="win-11" -DBUILDNAME="vs2026/-sh/zl" ..
```

First failure:

```
C:\src\hdf5.HDFGroup\src\H5private.h(592,24): error C2061: syntax error: identifier 'H5_float_complex' [C:\src\hdf5.HDFGroup\vs18\src\hdf5-static.vcxproj]
```

Ref: https://my.cdash.org/viewBuildError.php?buildid=3157693

hyoklee:
From https://www.kitware.com/cmake-4-2-0-available-for-download/ 

```
Some of the most significant changes in CMake 4.2 are:

The “Visual Studio 18 2026” generator was added.
```

hyoklee:
It's official:

https://devblogs.microsoft.com/visualstudio/visual-studio-2026-is-here-faster-smarter-and-a-hit-with-early-adopters/

hyoklee:
https://my.cdash.org/builds/3389258

PR #6052, opened 2025-11-23
Fix(CMake): Resolve MSVC C99 complex check failure on re-configure

xorespesp:
This is a fix for #6051.
I’ve updated the condition to check whether the fallback types (`_FCOMPLEX`) were detected in a previous run, ensuring that the safety block executes even if the standard size variable is cached.

----

> [!IMPORTANT]
> Fixes MSVC C99 complex type check in `ConfigureChecks.cmake` by considering fallback types, resolving re-configuration issue #6051.
> 
>   - **Behavior**:
>     - Fixes MSVC C99 complex type check in `ConfigureChecks.cmake` by updating condition to check fallback types (`_FCOMPLEX`, `_DCOMPLEX`, `_LCOMPLEX`).
>     - Ensures safety block executes if standard size variable is cached.
>   - **Misc**:
>     - Resolves issue #6051 related to re-configuration failure.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 379aaa78456ce9fcad8ba245d0324bfbc98dcb0e. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


hyoklee (dismissed):
@xorespesp , thank you for the fix!

close #5940

jhendersonHDF:
Thanks @xorespesp,

probably a bit of a silly question, but have you verified that MSVC builds enable the complex number types after reconfiguration and don't skip/disable them?

config/ConfigureChecks.cmake
```diff
@@ -899,8 +899,10 @@ if (${HDF_PREFIX}_HAVE_COMPLEX_H)
   HDF_CHECK_TYPE_SIZE ("double _Complex" ${HDF_PREFIX}_SIZEOF_DOUBLE_COMPLEX)
   HDF_CHECK_TYPE_SIZE ("long double _Complex" ${HDF_PREFIX}_SIZEOF_LONG_DOUBLE_COMPLEX)
 
-  if (MSVC AND NOT ${HDF_PREFIX}_SIZEOF_FLOAT_COMPLEX AND NOT ${HDF_PREFIX}_SIZEOF_DOUBLE_COMPLEX
-      AND NOT ${HDF_PREFIX}_SIZEOF_LONG_DOUBLE_COMPLEX)
+  if (MSVC AND 
+      ((NOT ${HDF_PREFIX}_SIZEOF_FLOAT_COMPLEX)       OR ${HDF_PREFIX}_SIZEOF__FCOMPLEX) AND 
+      ((NOT ${HDF_PREFIX}_SIZEOF_DOUBLE_COMPLEX)      OR ${HDF_PREFIX}_SIZEOF__DCOMPLEX) AND 
+      ((NOT ${HDF_PREFIX}_SIZEOF_LONG_DOUBLE_COMPLEX) OR ${HDF_PREFIX}_SIZEOF__LCOMPLEX))
```
bmribler: Is it possible to fix by resetting everything back to the original settings?  That is, similar to "make distclean" for autotools.

byrnHDF: That would be done by deleting the build folder - which is usually the best choice most of the time.

xorespesp:
> have you verified that MSVC builds enable the complex number types after reconfiguration and don't skip/disable them?

Yes, I tested this by integrating the patched HDF5 sources into my project using `FetchContent_Declare(SOURCE_DIR ...)` and triggering a re-configure.

The CMake output confirms that it falls back to the correct MSVC types instead of disabling them. Here is the output log I observed during second run(re-configuration): 

```
...
1> [CMake] -- Checking if complex number support is available
1> [CMake] -- Using MSVC complex number types
...
```
([Ref](https://github.com/HDFGroup/hdf5/blob/a6ff8aed236ee1e1deff6415e88b16c42b22f17c/config/ConfigureChecks.cmake#L940))

config/ConfigureChecks.cmake
xorespesp: While deleting the build folder might be a clean solution when building a library independently, but consider scenarios where HDF5 is consumed via `FetchContent` in large projects.

Modern IDEs often trigger CMake re-configuration automatically in the background.
If users are forced to wipe the entire build directory every time a project CMake script is modified, it would lead to excessive rebuild times. This significantly degrades the developer experience.

byrnHDF: Sorry, not suggesting a "force delete". I don't recall a FetchContent rebuilding something already built, but then I never went looking. My understanding is those Fetched projects should not rebuild without any change to their source. And my understanding is that CMake intention is that those configure cache settings are the intentional function.
Why do you want to force a rebuild, what changed in the main project that would have a bearing on hdf5 being built? The hdf5 configure is inspect system and compiler capabilities, what would be changing in the main project to alter those checks?

byrnHDF:
So regardless of the rebuild discussion, the problem is that the complex check has a unique check on windows?
And the result is that the values or state of configure variables are not cached correctly?

xorespesp:
> the problem is that the complex check has a unique check on windows?

Yes. this issue is specific to the MSVC unique check logic that i mentioned.

> and the result is that the values or state of configure variables are not cached correctly?

it's not that the cached values themselves are incorrect, but rather how the script "interprets" their presence during re-configuration.
- When the script sees the cached `H5_SIZEOF...` variable, it mistakenly assumes standard C99 support was found.
- Consequently, it **skips** the MSVC unique check logic(safety block) entirely.

This behavior is particularly problematic when using HDF5 via `FetchContent`. Even if the fetched HDF5 source is unchanged, touching the main project's CMake scripts triggers an automatic re-configuration in the IDE, triggering this issue and causing the build to fail.

config/ConfigureChecks.cmake
brtnfld: Must also fix this, should be ${HDF_PREFIX}_SIZEOF__LCOMPLEX

brtnfld: Should this be removed?

brtnfld: and then set cached value based on actual detection

`set (H5_HAVE_C99_COMPLEX_NUMBERS 0 CACHE INTERNAL "Using MSVC, no C99 complex" FORCE)`

brtnfld, line 935: Then add (C99 sucess path)
`set (H5_HAVE_C99_COMPLEX_NUMBERS 1 CACHE INTERNAL "Using C99 complex" FORCE)`

In practice:

First configuration:
Standard types fail -> MSVC types detected
Sets H5_HAVE_C99_COMPLEX_NUMBERS = 0 in cache (FORCE)

Re-configuration:
Variable loads from cache as 0
Never gets set to 1 because  optimistic line 896 was removed
Remains 0 (correct)

xorespesp: Thx, I pushed the update.

hyoklee (dismissed):

config/ConfigureChecks.cmake
xorespesp: > Then add (C99 sucess path)
> `set (H5_HAVE_C99_COMPLEX_NUMBERS 1 CACHE INTERNAL "Using C99 complex" FORCE)`

This part is a bit unclear to me.
If I apply the suggested change, I think the code should be modified like this:

```diff
    ...
    if (NOT H5_HAVE_STDC_NO_COMPLEX)
      # Compile simple test program with complex numbers
      HDF_FUNCTION_TEST (HAVE_COMPLEX_NUMBERS)

      if (H5_HAVE_COMPLEX_NUMBERS)
        
+        if (NOT DEFINED H5_HAVE_C99_COMPLEX_NUMBERS)
+          set (H5_HAVE_C99_COMPLEX_NUMBERS 1 CACHE INTERNAL "Using C99 complex" FORCE)
+        endif()

        if (H5_HAVE_C99_COMPLEX_NUMBERS)
          message (STATUS "Using C99 complex number types")
        else ()
          message (STATUS "Using MSVC complex number types")
        endif ()
        ...
```

Is this what you mean?

xorespesp: ... Regarding your suggestion, i completely agree. removing the "optimistic initialization" and using cached variables for both the MSVC unique check and the standard success path makes the logic much clearer.

I've updated the code accordingly. could you please take a look?

brtnfld (approved):
LGTM, thanks

config/ConfigureChecks.cmake
jhendersonHDF: @brtnfld What's the purpose of turning this into a cache variable with `set (H5_HAVE_C99_COMPLEX_NUMBERS 1 CACHE INTERNAL "Using C99 complex" FORCE)`? The intent was to not pollute the cmake cache with more than is necessary and in this case the variable is only needed to populate values in the features header file used by the library, so there's really no need for it to be a cache variable here. I also find the logic like

```
if (H5_HAVE_COMPLEX_NUMBERS)
  if (NOT DEFINED H5_HAVE_C99_COMPLEX_NUMBERS)
    set (H5_HAVE_C99_COMPLEX_NUMBERS 1 CACHE INTERNAL "Using C99 complex" FORCE)
  endif ()
endif ()
```

much more confusing because it's baking in previous assumptions about what the CMake logic did.

brtnfld: I'm also okay with not making it a cache variable.

brtnfld: @jhendersonHDF  The current logic assumes complex type detection is all-or-nothing. Can it be true that on 1 or 2 of the 3 standard types are detected? If so, then the NOT DEFINED hides the complexity

brtnfld:
We can talk about this and get back to you.

hyoklee (approved):
All tests pass with VS 18 2026:

https://my.cdash.org/viewTest.php?onlypassed&buildid=3220270

glennsong09 (approved):

jhendersonHDF:
@xorespesp Could you test my changes over in https://github.com/HDFGroup/hdf5/pull/6080? I decided that the complex configuration logic needed a bit of an update for clarity anyway while looking at the issue here. The setting of the size macro variables in the "MSVC" block was part of the problem due to using the same name for the C99 type size variables as the final size macro variables. I believe using separate names and some local variables made things a bit easier to read and avoided needing more unnecessary CMake cache variables.

lrknox:
We have merged #6080 "Rewrite complex number configuration logic to fix re-configuration issue and improve clarity" to develop, which should have fixed the issues.  If you have a chance, please try develop (with #6080, Dec 5 or later) with the failing configuration.

jhendersonHDF:
Closing this per @lrknox's last comment. @xorespesp please let us know and/or feel free to re-open this if my changes in https://github.com/HDFGroup/hdf5/pull/6080 didn't address the issue for you.

PR #6074, opened 2025-12-01
fixes the GitHub Actions CPack failure on macOS

brtnfld:
* Added conditional logic to detect macOS builds
    * Cleans stale HDF5 volume mounts before running CPack
    * Disables Spotlight indexing which can interfere with disk image operations
    * Implements retry logic (waits 5 seconds and retries once if CPack fails)

----

> [!IMPORTANT]
> Fixes CPack failure on macOS in GitHub Actions by cleaning stale mounts, disabling Spotlight, and adding retry logic.
> 
>   - **GitHub Actions Workflow**:
>     - Adds conditional logic in `main.yml` and `arm-main.yml` to handle macOS builds specifically.
>     - Cleans stale HDF5 volume mounts before running CPack to prevent conflicts.
>     - Disables Spotlight indexing on macOS to avoid interference with disk image operations.
>     - Implements retry logic for CPack execution, waiting 5 seconds and retrying once if it fails initially.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 42d0a88f169451bea5bef19fc168c804d15b6563. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/arm-main.yml
+++ b/.github/workflows/arm-main.yml
@@ -380,8 +380,19 @@ jobs:
         working-directory: ${{ runner.workspace }}/build
 
       - name: Run Package
-        run: cpack -C ${{ inputs.build_mode }} -V
+        run: |
+          # Clean any stale mounts
+          for vol in /Volumes/HDF5*; do
+            [ -d "$vol" ] && hdiutil detach "$vol" -force || true
+          done
+
+          # Disable indexing
+          sudo mdutil -i off / || true
+
+          # Run cpack with retry
+          cpack -C ${{ inputs.build_mode }} -V || (sleep 5 && cpack -C ${{ inputs.build_mode }} -V)
         working-directory: ${{ runner.workspace }}/build
+        shell: bash
 
       - name: List files in the space
         run: |
--- a/.github/workflows/arm-main.yml
+++ b/.github/workflows/arm-main.yml
@@ -381,12 +381,7 @@ jobs:
 
       - name: Run Package
         run: |
-          # Clean any stale mounts
-          for vol in /Volumes/HDF5*; do
-            [ -d "$vol" ] && hdiutil detach "$vol" -force || true
-          done
-
-          # Disable indexing
+          # Disable indexing to prevent file locking during CPack
           sudo mdutil -i off / || true
 
           # Run cpack with retry
--- a/.github/workflows/maven-staging.yml
+++ b/.github/workflows/maven-staging.yml
@@ -306,17 +306,8 @@ jobs:
         id: buildhdf5
         shell: bash
         run: |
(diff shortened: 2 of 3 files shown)


jhendersonHDF:
Was about to mention that this probably needs propagated elsewhere, at least to the Maven staging workflows, right as that workflow happened to fail. There are probably other places as well.

.github/workflows/arm-main.yml
```diff
@@ -380,8 +380,19 @@ jobs:
         working-directory: ${{ runner.workspace }}/build
 
       - name: Run Package
-        run: cpack -C ${{ inputs.build_mode }} -V
+        run: |
+          # Clean any stale mounts
+          for vol in /Volumes/HDF5*; do
```
jhendersonHDF: I'm not sure I understand why there would be stale mounts as each invocation should be a clean runner image.

brtnfld: Proactive cleanup, I can remove it if you want

lrknox: Seems like it would not be necessary, but there were no packaging failures in this run.
@jhendersonHDF, would you prefer trying to remove it?

lrknox (dismissed):

.github/workflows/arm-main.yml
jhendersonHDF: I wouldn't expect it to cause failures, I just wouldn't expect it to ever do anything either. Fine with it either way

.github/workflows/maven-staging.yml
```diff
@@ -306,8 +306,46 @@ jobs:
         id: buildhdf5
         shell: bash
         run: |
+          # macOS: Clean any stale disk image mounts before running CPack
+          if [[ "$RUNNER_OS" == "macOS" ]]; then
```
jhendersonHDF, line 310: However, the logic in this file file is quite different and inconsistent with the logic in the other files (checking `RUNNER_OS` vs. matrix `ostype`, emojis and message outputting, general formatting)

brtnfld: fixed

.github/workflows/arm-main.yml
brtnfld: fixed

lrknox (approved):

jhendersonHDF (approved):
I suspect there are going to be times where a 5 second sleep and a single retry won't be enough, but maybe this will improve things enough that it doesn't matter much

PR #6094, opened 2025-12-08
Fix inconsistent empty postfix handling

### LINKED ISSUE #6087: `-C config/cacheinit.cmake` is required for configuration in hdf5 2.0 / missing `DEBUG_POSTFIX`
Hi! On Windows for hdf5 <2.0 when configuring with `cmake .. -G Ninja -DCMAKE_INSTALL_PREFIX="installation" -DCMAKE_BUILD_TYPE=Debug` it resulted in libraries having debug posfix - e.g. `hdf5_D.lib`, instead of `hdf5.lib`.

But in hdf5 2.0 it's no longer the case and `hdf5.lib` is also produced for Debug build.
Reading through the cmake files, it seems it's started from https://github.com/HDFGroup/hdf5/pull/5600 - when `CMAKE_DEBUG_POSTFIX` was moved to `cacheinit.cmake`.

Does it mean that standard way of configuration is now always requires `cmake .. -C cacheinit.cmake`? It would be kind of awkward, since typically you expect from repo to configure with just `cmake ..`, unless you're using some kind of preset.

If it is required, then it seems `cacheinit.cmake` is missing in the examples below from the installation instructions:

https://github.com/HDFGroup/hdf5/blob/2cee17d8134606edadef07a4991b83faef1ee268/release_docs/INSTALL_CMake.txt#L326-L335

But even providing it and building with `cmake .. -C ../config/cmake/cacheinit.cmake -G Ninja -DCMAKE_INSTALL_PREFIX="L:\oblivion\hdf5-install\" -DCMAKE_BUILD_TYPE=Debug -DDBUILD_TESTING=OFF -DHDF5_ENABLE_Z_LIB_SUPPORT=OFF -DHDF5_BUILD_FORTRAN=OFF -DHDF5_BUILD_JAVA=OFF -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded -DCMAKE_POLICY_CMP0141=NEW -DHDF5_BUILD_TOOLS=OFF -DHDF5_BUILD_EXAMPLES=OFF -DHDF5_BUILD_UTILS=OFF` results in `hdf5.lib`, not `hdf5_d.lib`.

mattjala:
If the `cacheinit.cmake` file is used, the library sets an OS-specific Debug build library postfix if none is provided by the user. However, the way this check is done means that if a user manually provides an empty string value for the postfix, it will be detected the same as an unprovided postfix and replaced. This is likely to lead to confusion, since the user manually specified an empty postfix but instead receives the default postfixes.

This updates the check to distinguish between a user-provided empty postfix and an undefined postfix.

I also removed an old postfix check from an examples counterpart of the `HDF_DIR_PATHS` macro in `HDF5Examples/config/HDFMacros.cmake`, since it appears that this was copied and became out of date with the modern version of the macro in `HDFMacros.cmake`.

Fixes #6087

### DIFF
--- a/HDF5Examples/config/HDFMacros.cmake
+++ b/HDF5Examples/config/HDFMacros.cmake
@@ -173,15 +173,6 @@ macro (HDF_DIR_PATHS package_prefix)
     set (CMAKE_PREFIX_PATH ${ADDITIONAL_CMAKE_PREFIX_PATH} ${CMAKE_PREFIX_PATH})
   endif ()
 
-  #set the default debug suffix for all library targets
-  if(NOT CMAKE_DEBUG_POSTFIX)
-    if (WIN32)
-      set (CMAKE_DEBUG_POSTFIX "_D")
-    else ()
-      set (CMAKE_DEBUG_POSTFIX "_debug")
-    endif ()
-  endif ()
-
   SET_HDF_BUILD_TYPE()
 
   SET_HDF_OUTPUT_DIRS(${package_prefix})
--- a/config/cmake/cacheinit.cmake
+++ b/config/cmake/cacheinit.cmake
@@ -42,9 +42,9 @@ set (HDF5_MINGW_STATIC_GCC_LIBS ON CACHE BOOL "Statically link libgcc/libstdc++"
 #set the default debug suffix for all library targets
 if (NOT DEFINED CMAKE_DEBUG_POSTFIX)
   if (WIN32)
-    set (CMAKE_DEBUG_POSTFIX "_D" CACHE STRING "Debug library postfix" FORCE)
+    set (CMAKE_DEBUG_POSTFIX "_D" CACHE STRING "Debug library postfix")
   else ()
-    set (CMAKE_DEBUG_POSTFIX "_debug" CACHE STRING "Debug library postfix" FORCE)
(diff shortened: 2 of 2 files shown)


Andrej730:
Please consider returning debug postfix for Windows by default - it was convenient and was preventing running into errors, see the reasoning in 

> Okay. That's a bit unforunate then - it makes sense for Unix, where Release and Debug binaries are compatible, but on Windows they're not and not using debug postfix for them is a common source of confusion, leading to linking errors (e.g. Release `libhdf5_cpp.lib` has `/FAILIFMISMATCH:_ITERATOR_DEBUG_LEVEL=0` that will prevent in from being linked to Debug build and vice versa).
> 
> With debug postfix it's possible to install both Release and Debug builds to the same folder and let cmake figure out which ones to use - since it will have both `hdf5-targets-release.cmake` and `hdf5-targets-debug.cmake` configs with separated paths to libraries, you don't really need to change build/link process and provide debug postfix yourself.

lrknox (approved):

bmribler (approved):

PR #6103, opened 2025-12-11
Add h5diff message about file object difference

### LINKED ISSUE #6102: h5diff report mode should indicate when there is a diff due to a missing dataset in one file
Hello,
When running `h5diff -r foo.hdf5 bar.hdf5` I get an output like

```
group  : </> and </>
0 differences found
group  : </data> and </data>
0 differences found
dataset: </data/foo> and </data/foo>
0 differences found
dataset: </data/bar> and </data/bar>
0 differences found
dataset: </data/foofoo> and </data/foofoo>
0 differences found
dataset: </data/boo> and </data/boo>
0 differences found
dataset: </data/baz> and </data/baz>
0 differences found
dataset: </data/biz> and </data/biz>
0 differences found
```

Which looks like it's reporting no diffs, however, the return code is non-zero. I find this very confusing and if I was using a shell configuration that didn't provide any visual indicators for non-zero return codes I wouldn't know any better.

`-v` can be used to show
```
file1     file2
---------------------------------------
    x      x    /              
    x      x    /data          
    x      x    /data/foo   
    x      x    /data/bar   
    x      x    /data/foofoo
    x      x    /data/boo     
           x    /data/booboo
    x      x    /data/baz        
    x      x    /data/biz     
```

But it would be nice if there was some indication from report mode about this situation, e.g. at the end "Missing datasets detected, run with -v to view"

mattjala:
`h5diff` silently returns an error code of 1 whenever the two provided files differ structurally (one contains an object the other does not, different total object count). This PR adds a message indicating that the two files do not contain the same object set. 

Resolves #6102

----

> [!IMPORTANT]
> Add message in `h5diff` to indicate when files do not have the same set of objects, with updates to expected test outputs.
> 
>   - **Behavior**:
>     - Adds message in `print_info()` in `h5diff_common.c` when files differ structurally and `opts->contents` is 0, suggesting to use `-v` for more details.
>   - **Tests**:
>     - Updates `h5diff_11.txt` and `h5diff_13.txt` to include new message about differing object sets.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 2a421957cc94132b57364c2af0382a81ad844210. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tools/src/h5diff/h5diff_common.c
+++ b/tools/src/h5diff/h5diff_common.c
@@ -491,6 +491,7 @@ print_info(diff_opt_t *opts)
         printf("No common objects found. Files are not comparable.\n");
         if (!opts->mode_verbose)
             printf("Use -v for a list of objects.\n");
+        return;
     }
 
     if (opts->not_cmp == 1) {
--- a/tools/src/h5diff/h5diff_common.c
+++ b/tools/src/h5diff/h5diff_common.c
@@ -506,7 +506,7 @@ print_info(diff_opt_t *opts)
     }
 
     if (opts->contents == 0 && !opts->mode_verbose) {
-        printf("Files do not have the same set of objects\n");
+        printf("Files do not have the same set of objects.\n");
         printf("Use -v for a list of objects.\n");
     }
 }
--- a/tools/src/h5diff/h5diff_common.c
+++ b/tools/src/h5diff/h5diff_common.c
@@ -504,6 +504,11 @@ print_info(diff_opt_t *opts)
                 printf("Use -c for a list of objects.\n");
         }
     }
+
+    if (opts->contents == 0 && !opts->mode_verbose) {
+        printf("Files do not have the same set of objects\n");
+        printf("Use -v for a list of objects.\n");
+    }
 }
 
 /*-------------------------------------------------------------------------
--- a/tools/test/h5diff/expected/h5diff_11.txt
+++ b/tools/test/h5diff/expected/h5diff_11.txt
@@ -1,2 +1,4 @@
 dataset: </g1/dset1> and </g1/dset1>
 5 differences found
+Files do not have the same set of objects.
+Use -v for a list of objects.
--- a/tools/test/h5diff/expected/h5diff_13.txt
+++ b/tools/test/h5diff/expected/h5diff_13.txt
@@ -8,3 +8,5 @@ position        dset1           dset1           difference
 [ 1 1 ]          1               1.001           0.001          
 [ 2 1 ]          0               1               1              
 5 differences found
+Files do not have the same set of objects.
+Use -v for a list of objects.


hyoklee (dismissed):
@mattjala , thank you for the quick fix!

tools/src/h5diff/h5diff_common.c
```diff
@@ -504,6 +504,11 @@ print_info(diff_opt_t *opts)
                 printf("Use -c for a list of objects.\n");
         }
     }
+
+    if (opts->contents == 0 && !opts->mode_verbose) {
+        printf("Files do not have the same set of objects\n");
```
hyoklee: Add . before \n for consistency.

hyoklee (dismissed):

tools/src/h5diff/h5diff_common.c
```diff
@@ -504,6 +504,11 @@ print_info(diff_opt_t *opts)
                 printf("Use -c for a list of objects.\n");
         }
     }
+
+    if (opts->contents == 0 && !opts->mode_verbose) {
+        printf("Files do not have the same set of objects.\n");
+        printf("Use -v for a list of objects.\n");
```
bmribler, line 511: Do you know whether "a list of objects" means objects in both files or just one file and, if so, which one and should it be both?  If it should be both, then it can be a separate issue.

bmribler (changes requested):

tools/src/h5diff/h5diff_common.c
mattjala: The `-v` option prints a list of the objects in both of the files and indicates which ones are not shared between the two files. I don't think the `-v` option needs any changes - what specific change are you requesting?

tools/src/h5diff/h5diff_common.c
brtnfld, line 493: Should not this return if true, otherwise it is going to go on?

tools/src/h5diff/h5diff_common.c
```diff
@@ -504,6 +504,11 @@ print_info(diff_opt_t *opts)
                 printf("Use -c for a list of objects.\n");
         }
     }
+
```
brtnfld, line 508: if (opts->contents == 0 && !opts->mode_verbose && opts->cmn_objs != 0) {
}

tools/src/h5diff/h5diff_common.c
bmribler: Your response answered my question completely.  That was what I wanted to know.  Thanks!

bmribler (dismissed):

lrknox (dismissed):

tools/src/h5diff/h5diff_common.c
mattjala: Sure, since this message should supersede the 'some object not comaparable/files contain a different object set' messages.

tools/src/h5diff/h5diff_common.c
mattjala: If the goal of adding the `opts->cmn_objs != 0` check here is to avoid double output with the 'no common objects found' message above, then adding the early `return` should accomplish that same goal, while also preventing the 'some objects are not comparable' message at the same time. So I think this is unnecessary.

lrknox (approved):

bmribler (approved):

PR #6111, opened 2025-12-15
Enable data sieving for chunks that can't be cached

jhendersonHDF:
Fixed an issue that prevented use of a data sieve buffer for I/O on dataset chunks when those chunks couldn't be cached by the library. This issue could result in worst-case behavior of I/O on a single data element at a time when chunks are non-contiguous with respect to memory layout.

Added a test to attempt to catch performance regressions in I/O on dataset chunks that are non-contiguous with respect to memory layout

Updated the External File List logic to set the data sieve buffer size to the smaller of the dataset size and the size set in the FAPL, similar to the logic elsewhere in the library

----

> [!IMPORTANT]
> Re-enable data sieve buffer for non-cached dataset chunks to improve I/O performance and add a test for performance regressions.
> 
>   - **Behavior**:
>     - Re-enable data sieve buffer for I/O on non-cached dataset chunks in `H5D__chunk_read()` and `H5D__chunk_write()` in `H5Dchunk.c`.
>     - Update `H5D__efl_construct()` in `H5Defl.c` to set sieve buffer size to the smaller of dataset size and FAPL size.
>     - Add test `chunk_non_contig_mem_io` in `io_perf.c` to catch performance regressions for non-contiguous memory layout chunks.
>   - **Misc**:
>     - Update `CHANGELOG.md` to document the performance fix.
>     - Add `io_perf` to `H5_EXPRESS_TESTS` in `CMakeLists.txt`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for cb2384235fa9d832a1e13b33d10e1633c9673fa5. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -81,15 +81,15 @@ We would like to thank the many HDF5 community members who contributed to HDF5 2
 
 ### Fixed a performance issue with chunked dataset I/O
 
-   When dataset chunks are too large to be placed in the dataset chunk cache, the library falls back
-   to an alternative approach for I/O on dataset chunks. An issue with the logic in this approach
-   prevented chunked dataset I/O from making use of the library's data sieve buffer I/O optimization
-   functionality. For chunk shapes that are non-contiguous with the memory layout of a buffer, this
-   could result in severely degraded I/O performance, with the worst-case behavior causing I/O to be
-   performed on a single data element at a time.
+   When dataset chunks are unable to be placed in the dataset chunk cache (for example, if a chunk
+   is too large), the library falls back to an alternative approach for I/O on dataset chunks. An
+   issue with the logic in this approach prevented chunked dataset I/O from making use of the library's
+   data sieve buffer I/O optimization functionality. For chunk shapes that are non-contiguous with
+   the memory layout of a buffer, this could result in severely degraded I/O performance, with the
+   worst-case behavior causing I/O to be performed on a single data element at a time.
 
-   The data sieve buffer functionality has been properly re-enabled in these cases and will be used
-   as long as the underlying Virtual File Driver supports data sieving.
+   The data sieve buffer functionality has been extended to cover the case of uncached chunks and
+   will be used as long as the underlying Virtual File Driver supports data sieving.
 
 ## Java Library
 
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -79,6 +79,18 @@ We would like to thank the many HDF5 community members who contributed to HDF5 2
 
 ## Library
 
+### Fixed a performance issue with chunked dataset I/O
+
+   When dataset chunks are too large to be placed in the dataset chunk cache, the library falls back
+   to an alternative approach for I/O on dataset chunks. An issue with the logic in this approach
+   prevented chunked dataset I/O from making use of the library's data sieve buffer I/O optimization
+   functionality. For chunk shapes that are non-contiguous with the memory layout of a buffer, this
+   could result in severely degraded I/O performance, with the worst-case behavior causing I/O to be
+   performed on a single data element at a time.
+
+   The data sieve buffer functionality has been properly re-enabled in these cases and will be used
+   as long as the underlying Virtual File Driver supports data sieving.
+
 ## Java Library
 
 ## Configuration
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -3105,6 +3105,16 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
         ctg_io_info.dsets_info   = &ctg_dset_info;
         ctg_io_info.count        = 1;
 
+        /* Ensure dataset I/O sieve buffer is properly setup for when the contiguous
+         * dataset I/O routines are used for chunks. Force reset of sieve buffer here
+         * on each I/O until more advanced use cases like H5Dset_extent between writes
+         * and subsequent writes or reads can be supported.
+         */
+        dset_info->dset->shared->cache.contig.sieve_buf_size = H5F_SIEVE_BUF_SIZE(dset_info->dset->oloc.file);
+        dset_info->dset->shared->cache.contig.sieve_loc      = HADDR_UNDEF;
+        dset_info->dset->shared->cache.contig.sieve_size     = 0;
+        assert(!dset_info->dset->shared->cache.contig.sieve_dirty); /* Any writes should have been flushed */
+
         /* Initialize temporary contiguous storage info */
         ctg_store.contig.dset_size = dset_info->dset->shared->layout.u.chunk.size;
 
@@ -3262,6 +3272,15 @@ H5D__chunk_write(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
     ctg_dset_info.layout_ops = *H5D_LOPS_CONTIG;
     ctg_io_info.dsets_info   = &ctg_dset_info;
     ctg_io_info.count        = 1;
+
+    /* Ensure dataset I/O sieve buffer is properly setup for when the contiguous
+     * dataset I/O routines are used for chunks. Force reset of sieve buffer here
+     * on each I/O until more advanced use cases like H5Dset_extent between writes
+     * and subsequent writes or reads can be supported.
+     */
+    dset_info->dset->shared->cache.contig.sieve_buf_size = H5F_SIEVE_BUF_SIZE(dset_info->dset->oloc.file);
+    dset_info->dset->shared->cache.contig.sieve_loc      = HADDR_UNDEF;
+    dset_info->dset->shared->cache.contig.sieve_size     = 0;
 
     /* Initialize temporary contiguous storage info */
     ctg_store.contig.dset_size = dset_info->dset->shared->layout.u.chunk.size;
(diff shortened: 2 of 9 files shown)


jhendersonHDF:
Related writeup: 
[HDF5_chunked_dataset_IO_performance_issue.pdf](https://github.com/user-attachments/files/24177132/HDF5_chunked_dataset_IO_performance_issue.pdf)

Note that this is a rather quick fix for the issue and a bit of refactoring in the chunked I/O code would be better in the long run.

src/H5Dchunk.c
```diff
@@ -3105,6 +3105,16 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
         ctg_io_info.dsets_info   = &ctg_dset_info;
         ctg_io_info.count        = 1;
 
+        /* Ensure dataset I/O sieve buffer is properly setup for when the contiguous
+         * dataset I/O routines are used for chunks. Force reset of sieve buffer here
+         * on each I/O until more advanced use cases like H5Dset_extent between writes
+         * / writes and reads can be supported.
+         */
+        dset_info->dset->shared->cache.contig.sieve_buf_size = H5F_SIEVE_BUF_SIZE(dset_info->dset->oloc.file);
```
jhendersonHDF: This line (and the same one in `H5D__chunk_write()`) is the main fix. The library's logic seems to have been setup to account for chunked datasets making use of data sieving, but the sieve buffer size was always 0 so it was unused. In specific cases this could result in I/O being performed element by element.

src/H5Dchunk.c
```diff
@@ -3105,6 +3105,16 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
...
+        dset_info->dset->shared->cache.contig.sieve_loc      = HADDR_UNDEF;
+        dset_info->dset->shared->cache.contig.sieve_size     = 0;
```
jhendersonHDF: These lines are to force re-reads into the sieve buffer on each I/O to account for cases like changing the extent of a dataset between writes and reads

src/H5Defl.c
```diff
@@ -156,8 +156,11 @@ H5D__efl_construct(H5F_t *f, H5D_t *dset)
     tmp_size = (hsize_t)stmp_size * dt_size;
     H5_CHECKED_ASSIGN(dset->shared->layout.storage.u.contig.size, hsize_t, tmp_size, hssize_t);
 
-    /* Get the sieve buffer size for this dataset */
-    dset->shared->cache.contig.sieve_buf_size = H5F_SIEVE_BUF_SIZE(f);
+    /* Get the sieve buffer size for this dataset - the smaller of the dataset size and
+     * the sieve buffer size from the FAPL is used
+     */
+    dset->shared->cache.contig.sieve_buf_size =
```
jhendersonHDF: Updated the EFL code here to set the sieve buffer size to the smaller of the dataset size and the FAPL-specified size. This just matches the logic elsewhere in the library and what's currently documented for `H5Pset_sieve_buf_size()`:

> Internally, the library checks the storage sizes of the datasets in
> the file. It picks the smaller one between the size from the file
> access property and the size of the dataset to allocate the sieve
> buffer for the dataset in order to save memory usage.

test/io_perf.c
jhendersonHDF, line 1: When a sieve buffer is not used, the test in this file is designed to run a long time to try and cause a timeout in CMake for TestExpress of 0, but run very quickly for a TestExpress of 3. We currently do testing with a TestExpress of 0 for every PR, but those tests should probably be moved to a scheduled run as originally intended since tests like this can be problematic for testing at that level for every PR.

mattjala (dismissed):

src/H5Dchunk.c
```diff
@@ -3263,6 +3273,15 @@ H5D__chunk_write(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
     ctg_io_info.dsets_info   = &ctg_dset_info;
     ctg_io_info.count        = 1;
 
+    /* Ensure dataset I/O sieve buffer is properly setup for when the contiguous
+     * dataset I/O routines are used for chunks. Force reset of sieve buffer here
+     * on each I/O until more advanced use cases like H5Dset_extent between writes
+     * / writes and reads can be supported.
```
lrknox:
```suggestion
     * and reads can be supported.
```

jhendersonHDF: This was intentional to try to highlight the difference between write/write and write/read cycles, but could probably be written differently.

lrknox: I thought it was a repeated word typo.  Maybe a ; between the two writes, or a . and W?

jhendersonHDF: Reworded this slightly

mattjala (dismissed):

lrknox (dismissed):

fortnern:
I would consider this a new feature, not a fix

fortnern:
Has data sieving ever been used in this case before? The AI seems to think so but I think it's wrong.

src/H5Dchunk.c
fortnern: Ideally the "contig" struct should be renamed to sieve or something similar since it's no longer specific to contiguous datasets, but this should be fine for now

src/H5Dchunk.c
fortnern: So the sieve buffer isn't persistent for chunked datasets?

fortnern: Should we free the sieve buffer after I/O since the data in it is useless?

src/H5Dint.c
```diff
@@ -1991,6 +1991,11 @@ H5D_close(H5D_t *dataset)
                         H5FL_FREE(H5D_piece_info_t, dataset->shared->cache.chunk.single_piece_info);
                     dataset->shared->cache.chunk.single_piece_info = NULL;
                 } /* end if */
+
+                /* Free the data sieve buffer, if it's been allocated */
+                if (dataset->shared->cache.contig.sieve_buf)
```
fortnern: I'm inclined to move this out of the switch statement (and remove the H5D_CONTIGUOUS case)

fortnern (changes requested):
Marking "request changes", really just requesting discussion/responses to my comments, then we'll decide if changes are needed

jhendersonHDF:
> I would consider this a new feature, not a fix

I can see that, but it was very clearly a bug due to oversight in the way that the chunking code was written which led to drastic performance problems.

> Has data sieving ever been used in this case before?

I haven't gone through the history much, but randomly picking 1.10.3 as a test branch shows the same problem.

src/H5Dchunk.c
jhendersonHDF: The sieve buffer is currently freed when the dataset is closed, but we could free it after I/O if you think that's better; it just seemed more efficient to keep it around. Setting `sieve_loc` to `HADDR_UNDEF` and `sieve_size` to 0 is just to make sure that the library doesn't read stale data from the sieve buffer in the case where the dataset extent changes. Note the `sieve_size` here is just the value tracking how much data is currently in the buffer.

release_docs/CHANGELOG.md
```diff
@@ -79,6 +79,18 @@ We would like to thank the many HDF5 community members who contributed to HDF5 2
...
+### Fixed a performance issue with chunked dataset I/O
+
+   When dataset chunks are too large to be placed in the dataset chunk cache, the library falls back
+   to an alternative approach for I/O on dataset chunks. An issue with the logic in this approach
+   prevented chunked dataset I/O from making use of the library's data sieve buffer I/O optimization
+   functionality. For chunk shapes that are non-contiguous with the memory layout of a buffer, this
+   could result in severely degraded I/O performance, with the worst-case behavior causing I/O to be
+   performed on a single data element at a time.
+
+   The data sieve buffer functionality has been properly re-enabled in these cases and will be used
```
fortnern: This implies it was previously enabled, which it wasn't. This should read something like "The data sieve buffer functionality has been extended to work in these cases  and will be used..." In general the change log should make it clear that the sieve buffer was previously only for contiguous datasets and has been extended to this case so the performance of uncompressed uncached chunked datasets matches that of contiguous datasets.

src/H5Dint.c
jhendersonHDF: Seems reasonable. I just modeled this as if the `contig` struct were a union.

release_docs/CHANGELOG.md
fortnern: Though I suppose with repeated operations the performance is still potentially slower since the sieve buffer isn't persistent with chunked

src/H5Dchunk.c
fortnern: Since we don't need the data in the buffer keeping it around is effectively treating it like a buffer on the free list. The sieve buffer is already integrated with the free list however, so I think we should just free it after I/O (it will be held open by the H5FL code if appropriate).

release_docs/CHANGELOG.md
jhendersonHDF: Re-worded to not imply the sieve buffer was enabled before and also to not imply that a chunk being too large is the only time it won't be placed in the cache.

src/H5Dchunk.c
jhendersonHDF: I'd say that's reasonable to do while already looking at this specific code.

src/H5Dint.c
jhendersonHDF: Kept the `H5D_CONTIGUOUS` case as developer builds currently warn about missing switch cases, but left a comment similar to `H5D_COMPACT`.

src/H5Dchunk.c
jhendersonHDF: Done

src/H5Dchunk.c
```diff
@@ -409,6 +409,9 @@ H5FL_BLK_DEFINE_STATIC(chunk);
 /* Declare extern free list to manage the H5S_sel_iter_t struct */
 H5FL_EXTERN(H5S_sel_iter_t);
 
+/* Declare the external PQ free list for the sieve buffer information */
+H5FL_BLK_EXTERN(sieve_buf);
```
jhendersonHDF, line 413: With the sieve buffer being managed by H5Dchunk.c as well, it needs a reference to the free list. This is a bit messy as the sieve buffer is still almost completely controlled by H5Dcontig.c, but refactoring that seems outside the scope of this PR.

src/H5Dchunk.c
```diff
@@ -3201,6 +3214,14 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
     }     /* end else */
 
 done:
+    /* Free dataset sieve buffer and reset cached fields */
+    if (dset_info->dset->shared->cache.sieve.sieve_buf) {
```
jhendersonHDF, line 3218: While the resetting of fields in these sections is mostly redundant with the same above, I wanted to keep both so the context (comment) isn't lost in potential future refactoring.

fortnern (approved):

brtnfld (approved):

lrknox (approved):

PR #6113, opened 2025-12-17
docs: change .sh to .rpm or .deb for Linux binary distribution

hyoklee:
There is no .sh file.

Ref: 
1. https://github.com/HDFGroup/hdf5/releases/tag/2.0.0
2. https://www.hdfgroup.org/download-hdf5/

----

> [!IMPORTANT]
> Update `Using_CMake.txt` to correct Linux binary distribution file extension from `.sh` to `.rpm`.
> 
>   - **Documentation**:
>     - Update `Using_CMake.txt` to change Linux binary distribution file extension from `.sh` to `.rpm`.
>     - Reflects actual distribution format as per HDF5 release notes and download page.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for c67ffca4e9fbf1fd74e42b2b2367c57ae17d20ca. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


HDF5Examples/Using_CMake.txt
```diff
@@ -32,7 +32,7 @@ I. Preconditions
 
    2. You have installed the HDF5 library built with CMake, by executing
       the HDF Install Utility (the *.msi file in the binary package for
-      Windows or the *.sh on Linux). You can obtain pre-built binaries
+      Windows or the *.rpm on Linux). You can obtain pre-built binaries
```
byrnHDF: Actually, this would depend on which binary you use, since there are three types, sh. rpm and deb.

hyoklee: @byrnHDF , I don't see `.sh` type. Where can I download it?

byrnHDF: There are three linux (ubuntu files) the tar.gz file when uncompressed should have a sh file.

hyoklee: @byrnHDF , I couldn't find .sh file when I uncompressed the tar.gz.

byrnHDF:
Well I guess I might have it confused with HDFView.

lrknox:
The most appropriate binary for Ubuntu would be the .deb file.  I tried it on an Ubuntu24.04 EC2 instance.  It did install, though the output misled me to believe it was already installed.  The files are installed in /HDFGroup/HDF5/2.0.0/.  At Claud's suggestion I ran "sudo ln -s /HDF_Group/HDF5/2.0.0/lib/* /usr/local/lib/" and similar commands for include and bin.  That seems to work, but I think it would be better to use something like CMAKE_INSTALL_PREFIX=/usr/local for Ubuntu?

The rpm file can potentially be used for ubuntu but requires additional processing to create a .deb file.  It should be usable on linux machines with yum (centos, rocky9, probably fedora).  I tried it on rocky9, but it complained that nothing provides libc.so.6 and libstdc++.so.6, even though they are present on the machine in /lib64.  I'm not sure whether the version is incorrect, or the location.  I'll investigate further, and maybe try having CMake build the rpm binary on one of those machines instead of on ubuntu, with a different CMAKE_INSTALL_PREFIX.

I remember the .sh ffile;  Allen said it was created in the past, but maybe was dropped in the workflow files at some point.  I'll look for that, too.

jhendersonHDF (commented):
Personally I'd like if we brought back the .sh installer, but this will require a bit of work on the CMake side, so approving this for now.

jhendersonHDF (approved):

lrknox (approved):

Issue #6107, opened 2025-12-12
Duplicate definition of ALIAS target "libaec::sz" when building other projects that use HDF5


malcolm-vivosa:
**Describe the bug**
Since https://github.com/HDFGroup/hdf5/commit/f2d2d45d2620c07abdecaa1d35e5a680825a2a8f I am seeing errors when building some projects that use casacore (https://github.com/casacore/casacore) which in turn uses hdf5.

The errors look like this:
```
CMake Error at /usr/lib/cmake/libaec/libaec-config.cmake:52 (add_library):
  add_library cannot create ALIAS target "libaec::aec" because another target
  with the same name already exists.
Call Stack (most recent call first):
  /development/built/hdf5/cmake/hdf5-config.cmake:147 (find_package)
  /usr/share/cmake/Modules/FindHDF5.cmake:677 (find_package)
  cmake/FindCasacore.cmake:234 (find_package)
  CMakeLists.txt:142 (find_package)
CMake Error at /usr/lib/cmake/libaec/libaec-config.cmake:53 (add_library):
  add_library cannot create ALIAS target "libaec::sz" because another target
  with the same name already exists.
Call Stack (most recent call first):
  /development/built/hdf5/cmake/hdf5-config.cmake:147 (find_package)
  /usr/share/cmake/Modules/FindHDF5.cmake:677 (find_package)
  cmake/FindCasacore.cmake:234 (find_package)
  CMakeLists.txt:142 (find_package)
```

In the past this doesn't seem to be an issue because hdf5 used to protect against duplicate export but with recent changes this is no longer the case.

I was able to fix this locally by changing `hdf5-config.cmake.in` to use `find_dependency` instead of `find_package` which my understanding is what all .cmake files should do anyway. 
Interested to hear the thoughts of others if this would be the correct thing to do, I'm happy to submit a MR with these changes if theres some consensus that this is what should change.

**Expected behavior**
Should be able to build cmake based programs that have worked in the past with hdf5 without encountering multiple alias errors.

**Platform (please complete the following information)**
 - HDF5 version - master branch
 - Cmake version - 4.2.1

**Additional context**

mattjala:
The change you propose should be fine. Our pull request CI should have scenarios that build SZIP both inline and externally, which should catch any issues that change would cause.

lrknox:
Fixed by pull request #6139

PR #6139, opened 2026-01-02
Use 'find_dependency' instead of 'find_package'

malcolm-vivosa:
It is considered best practice for cmake package configuration files to use "find_dependency" instead of "find_package". This is because it can propagate flags properly, gives more clear error messages and is able to avoid duplicate dependency include issues.

This change fixes an issue I am seeing in some external cmake projects when including hdf5. As per issue report here https://github.com/HDFGroup/hdf5/issues/6107

----

> [!IMPORTANT]
> Replace `find_package` with `find_dependency` in `hdf5-config.cmake.in` to improve dependency handling.
> 
>   - **Behavior**:
>     - Replace `find_package` with `find_dependency` in `hdf5-config.cmake.in` for dependencies: `MPI`, `Threads`, `ZLIB`, and `LIBAEC`.
>     - Improves flag propagation, error messages, and avoids duplicate dependency issues.
>   - **Misc**:
>     - Add `include(CMakeFindDependencyMacro)` at the beginning of `hdf5-config.cmake.in`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 253d3835d423741b945a921f11b2990f7913a9d3. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/config/install/hdf5-config.cmake.in
+++ b/config/install/hdf5-config.cmake.in
@@ -16,6 +16,8 @@
 #-----------------------------------------------------------------------------
 @PACKAGE_INIT@
 
+include(CMakeFindDependencyMacro)
+
 string(TOUPPER @HDF5_PACKAGE@ HDF5_PACKAGE_NAME)
 
 set (${HDF5_PACKAGE_NAME}_VALID_COMPONENTS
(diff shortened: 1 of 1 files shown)


lrknox (approved):

jhendersonHDF (approved):
Thanks @malcolm-vivosa, LGTM

PR #6140, opened 2026-01-02
☀️  Fix memory safety vulnerabilities in high-level and VFD code

brtnfld:
Address multiple CWE-415 (double-free), CWE-416 (use-after-free), and CWE-122 (buffer overflow) vulnerabilities identified by static analysis:

- hl/src/H5DS.c: Fix double-free in H5DSis_scale() by setting buf to NULL after free and adding NULL check in cleanup path

- hl/src/H5LT.c: Fix multiple memory issues:
  * Set myinput to NULL after free in H5LTtext_to_dtype()
  * Add NULL check in realloc_and_append() to prevent use-after-free
  * Refactor duplicated stmp handling by creating H5LT_append_dtype_super_text() helper function, eliminating ~50 lines of repeated code across 4 case blocks

- hl/src/H5TB.c: Replace unsafe strcpy() with strncpy() in H5TBget_field_info() using HLTB_MAX_FIELD_LEN constant to prevent buffer overflow

- hl/src/H5TBpublic.h: Document buffer size requirements for field_names parameter

- src/H5FDstdio.c: Fix inconsistent resource cleanup in H5FD_stdio_open() by using file->fp instead of f throughout error paths

- src/H5VLnative.c: Add assert checks for obj and file parameters in H5VL_native_get_file_struct() following internal API conventions

SAFE project work.

----

> [!IMPORTANT]
> Fixes memory safety vulnerabilities in HDF5 codebase, addressing double-free, use-after-free, and buffer overflow issues across multiple files.
> 
>   - **Memory Safety Fixes**:
>     - `H5DS.c`: Fix double-free in `H5DSis_scale()` by setting `buf` to NULL after free and adding NULL check in cleanup.
>     - `H5LT.c`: Set `myinput` to NULL after free in `H5LTtext_to_dtype()`, add NULL check in `realloc_and_append()`, refactor `H5LT_append_dtype_super_text()` to reduce code duplication.
>     - `H5TB.c`: Replace `strcpy()` with `strncpy()` in `H5TBget_field_info()` to prevent buffer overflow.
>   - **Documentation**:
>     - `H5TBpublic.h`: Document buffer size requirements for `field_names` parameter.
>   - **Resource Management**:
>     - `H5FDstdio.c`: Use `file->fp` consistently in `H5FD_stdio_open()` for error paths.
>   - **Assertions**:
>     - `H5VLnative.c`: Add assert checks for `obj` and `file` parameters in `H5VL_native_get_file_struct()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 7b22833a25806cfc4beb08a6725150fab3cf3d4e. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/hl/src/H5LT.c
+++ b/hl/src/H5LT.c
@@ -1889,13 +1889,18 @@ H5LTtext_to_dtype(const char *text, H5LT_lang_t lang_type)
 
     input_len = strlen(text);
     myinput   = strdup(text);
+    if (!myinput) {
+        H5Epush_ret(__func__, H5E_ERR_CLS, H5E_RESOURCE, H5E_NOSPACE, "memory allocation failed", -1);
+    }
 
     if ((type_id = H5LTyyparse()) < 0) {
         free(myinput);
+        myinput = NULL;
         goto out;
     }
 
     free(myinput);
+    myinput   = NULL;
     input_len = 0;
 
     return type_id;
@@ -1909,7 +1914,17 @@ out:
  *
  * Purpose:     Expand the buffer and append a string to it.
  *
- * Return:      void
+ * Parameters:
+ *   _no_user_buf: Operating mode flag
+ *                 - true:  Library-managed buffer (can reallocate)
+ *                 - false: User-provided buffer (fixed size, no realloc)
+ *   len:         Pointer to current buffer size (updated if reallocated)
+ *   buf:         Buffer to append to. If NULL, returns NULL immediately.
+ *   str_to_add:  String to append. If NULL:
+ *                  - Library-managed mode: reallocates to ensure extra space, no append
+ *                  - User-provided mode: no-op, returns buf unchanged (space cannot be ensured)
+ *
+ * Return:      Updated buffer pointer on success, NULL on failure
  *
  *-------------------------------------------------------------------------
  */
@@ -1918,11 +1933,12 @@ realloc_and_append(bool _no_user_buf, size_t *len, char *buf, const char *str_to
 {
     size_t size_str_to_add, size_str;
 
+    if (!buf)
+        goto out;
+
+    /* Mode 1: Library-managed buffer - can reallocate as needed */
     if (_no_user_buf) {
         char *tmp_realloc;
-
-        if (!buf)
-            goto out;
 
         /* If the buffer isn't big enough, reallocate it.  Otherwise, go to do strcat. */
         if (str_to_add && ((ssize_t)(*len - (strlen(buf) + strlen(str_to_add) + 1)) < LIMIT)) {
@@ -1941,6 +1957,8 @@ realloc_and_append(bool _no_user_buf, size_t *len, char *buf, const char *str_to
         else
             buf = tmp_realloc;
     }
+    /* Mode 2: User-provided buffer - fixed size, no reallocation
+     * (else case is implicit - no action needed) */
 
     if (str_to_add) {
         /* find the size of the buffer to add */
@@ -2169,6 +2187,58 @@ out:
     free(text_str);
 
     return FAIL;
+}
+
+/*-------------------------------------------------------------------------
+ * Function:    append_dtype_super_text
+ *
+ * Purpose:     Helper function to get super type text and append it to dt_str.
+ *              This encapsulates the common pattern of: allocate buffer,
+ *              convert dtype to text, append to string, free buffer.
+ *
+ * Return:      Success: updated dt_str pointer, Failure: NULL
+ *
+ *-------------------------------------------------------------------------
+ */
+static char *
+append_dtype_super_text(hid_t super, char *dt_str, H5LT_lang_t lang, size_t *slen, bool no_user_buf)
+{
+    size_t super_len;
+    char  *stmp = NULL;
+
+    /* Get required buffer size for super type text */
+    if (H5LTdtype_to_text(super, NULL, lang, &super_len) < 0)
+        return NULL;
+    /* Treat zero-length result as failure rather than calling calloc(0,...) */
+    if (super_len == 0)
+        return NULL;
+
+    /* Allocate buffer for super type text */
+    stmp = (char *)calloc(super_len, sizeof(char));
+    if (!stmp) {
+        H5Epush_ret(__func__, H5E_ERR_CLS, H5E_RESOURCE, H5E_NOSPACE, "memory allocation failed", NULL);
+    }
+
+    /* Convert super type to text */
+    if (H5LTdtype_to_text(super, stmp, lang, &super_len) < 0) {
+        free(stmp);
+        return NULL;
+    }
+
+    /* Append super type text to dt_str.  Use a temporary pointer so that
+     * dt_str is not overwritten before we confirm success.  Note:
+     * realloc_and_append frees dt_str itself on failure in library-managed
+     * mode, so on the failure path dt_str must not be used. */
+    {
+        char *tmp = realloc_and_append(no_user_buf, slen, dt_str, stmp);
+        free(stmp);
+        stmp = NULL;
+        if (!tmp)
+            return NULL;
+        dt_str = tmp;
+    }
+
+    return dt_str;
 }
 
 /*-------------------------------------------------------------------------
@@ -2536,8 +2606,7 @@ next:
             tag = H5Tget_tag(dtype);
             if (tag) {
                 snprintf(tmp_str, TMP_LEN, "OPQ_TAG \"%s\";\n", tag);
-                if (tag)
-                    H5free_memory(tag);
+                H5free_memory(tag);
                 tag = NULL;
             }
             else
@@ -2556,9 +2625,7 @@ next:
             break;
         }
         case H5T_ENUM: {
-            hid_t  super;
-            size_t super_len;
-            char  *stmp = NULL;
+            hid_t super;
 
             /* Print lead-in */
             snprintf(dt_str, *slen, "H5T_ENUM {\n");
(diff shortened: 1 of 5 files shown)


hl/src/H5DS.c
```diff
@@ -2301,7 +2302,8 @@ H5DSis_scale(hid_t did)
     }
 out:
     if (is_ds < 0) {
-        free(buf);
+        if (buf)
```
jhendersonHDF: Nitpick but there's generally never a need to have an `if (buf)` statement like this; `free(buf)` is fine when `buf` is `NULL`.

hl/src/H5TBpublic.h
```diff
@@ -457,7 +457,10 @@ H5HL_DLL herr_t H5TBget_table_info(hid_t loc_id, const char *dset_name, hsize_t
  *
  * \fg_loc_id
  * \param[in] dset_name         The name of the dataset to read
- * \param[out] field_names      An array containing the names of the fields
+ * \param[out] field_names      An array containing the names of the fields.
+ *                              Each element must be allocated by the caller to at least
+ *                              255 bytes. Field names longer than 254 characters will be
```
jhendersonHDF: Rather than using specific numbers here, it seems like it would be better if we call out the macro `HLTB_MAX_FIELD_LEN` representing the value and `HLTB_MAX_FIELD_LEN - 1` for 254 here. If that macro's currently private it should probably be public to prevent possible future issues if the buffer size requirement changes.

src/H5FDstdio.c
```diff
@@ -393,8 +393,8 @@ H5FD_stdio_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxaddr
 
         /* Use the value in the property list */
         if (H5Pget_file_locking(fapl_id, &unused, &file->ignore_disabled_file_locks) < 0) {
+            fclose(file->fp);
```
jhendersonHDF, line 396: What was the issue with these close calls? `file->fp` should be the same as `f` at this point

hl/src/H5TBpublic.h
brtnfld: Yes, it is private. I'll make it public

src/H5FDstdio.c
brtnfld: Snyk flags it for clarity on resource ownership. The concern is that it's not explicit which pointer "owns" the resource after the assignment. Once you've assigned file->fp = f, the FILE* is conceptually owned by the file structure, and using file->fp in cleanup makes this ownership clear.

hl/src/H5DS.c
brtnfld: fixed.

hl/src/H5TBpublic.h
brtnfld: done

hl/src/H5LT.c
```diff
@@ -1942,6 +1946,9 @@ realloc_and_append(bool _no_user_buf, size_t *len, char *buf, const char *str_to
             buf = tmp_realloc;
     }
 
+    if (!buf)
```
jhendersonHDF: The intent of the `_no_user_buf` parameter isn't really obvious, but it seems like this check overlaps with the same check inside that block, which seems like it would imply `buf` being allowed to be passed in as `NULL` in the false case. But I'm guessing this check was added due to the `strlen(buf)` below. This seems like we should determine whether it was ever intended for `buf` to be allowed as `NULL`.

hl/src/H5TBprivate.h
```diff
@@ -19,9 +19,9 @@
 /* public TB prototypes			*/
 #include "H5TBpublic.h"
 
-#define TABLE_CLASS        "TABLE"
-#define TABLE_VERSION      "3.0"
-#define HLTB_MAX_FIELD_LEN 255
+#define TABLE_CLASS   "TABLE"
+#define TABLE_VERSION "3.0"
+/* HLTB_MAX_FIELD_LEN is now defined in H5TBpublic.h */
```
jhendersonHDF: Harmless, but it's probably unnecessary to document that a macro used to be in this file

hl/src/H5LT.c
brtnfld: added comments, side-stepped the question if we should assert/error more clearly.

hl/src/H5TBprivate.h
brtnfld: fixed

hl/src/H5LT.c
```diff
@@ -2171,6 +2195,50 @@ H5LTdtype_to_text(hid_t dtype, char *str, H5LT_lang_t lang_type, size_t *len)
...
+static char *
+H5LT_append_dtype_super_text(hid_t super, char *dt_str, H5LT_lang_t lang, size_t *slen, bool no_user_buf)
+{
+    size_t super_len;
+    char  *stmp = NULL;
+
+    /* Get required buffer size for super type text */
+    if (H5LTdtype_to_text(super, NULL, lang, &super_len) < 0)
+        return NULL;
+
```
bmribler, line 2220: Could super_len still be invalid somehow even though H5LTdtype_to_text didn't fail?

hl/src/H5TBpublic.h
```diff
@@ -457,7 +471,13 @@ H5HL_DLL herr_t H5TBget_table_info(hid_t loc_id, const char *dset_name, hsize_t
...
- * \param[out] field_names      An array containing the names of the fields
+ * \param[out] field_names      An array containing the names of the fields.
+ *                              Each element must be allocated by the caller with sufficient
+ *                              space to hold the field name. Field names are copied as-is if
+ *                              shorter than #HLTB_MAX_FIELD_LEN. Names longer than
+ *                              #HLTB_MAX_FIELD_LEN - 1 characters will be truncated and
+ *                              require buffers of at least #HLTB_MAX_FIELD_LEN bytes.
+ *                              Pass NULL to skip retrieving field names.
  * \param[out] field_sizes      An array containing the size of the fields
  * \param[out] field_offsets    An array containing the offsets of the fields
```
bmribler, line 482: Not from the changes in this PR, but this should be "offset of the fields" or the above should be "sizes of the fields"

hl/src/H5TBpublic.h
```diff
@@ -472,6 +492,13 @@ H5HL_DLL herr_t H5TBget_table_info(hid_t loc_id, const char *dset_name, hsize_t
  *          named \p dset_name attached to the object specified
  *          by the identifier \p loc_id.
  *
+ * \note    When retrieving field names, callers should allocate each field_names[i]
+ *          buffer with sufficient space for the expected field name length.
```
bmribler: "for the expected field name."?

hl/test/test_table.c
```diff
@@ -1488,6 +1488,201 @@ test_table(hid_t fid, int do_write)
...
+        goto out;
+
+    /* Verify field names are correctly retrieved */
+    for (i = 0; i < NFIELDS; i++) {
+        if ((strcmp(field_names[i], names_out[i]) != 0)) {
+            goto out;
+        }
+    }
+
+    /* release */
```
bmribler: Release?  (for consistency :)

hl/test/test_table.c
```diff
@@ -1488,6 +1488,201 @@ test_table(hid_t fid, int do_write)
...
+        memset(field_253, 'A', 253);
+        field_253[253] = '\0';
+        memset(field_254, 'B', 254);
+        field_254[254] = '\0';
+        memset(field_255, 'C', 255);
+        field_255[255] = '\0';
+        memset(field_1000, 'D', 1000);
+        field_1000[1000] = '\0';
+
+        const char *boundary_field_names[5] = {field_253, field_254, field_255, field_1000, "ShortName"};
```
bmribler: Should "ShortName" be defined as a constant?  Same for file names, i.e., "test_boundary.h5".
Definitely "5" should be a constant.

bmribler (changes requested):

brtnfld:
@bmribler	

Added if (super_len == 0) return NULL; guard after the probe call
changed size of the fields" to "sizes of the fields"
H5TBpublic.h, changed "for the expected field name length." to "for the expected field name."
test_table.c: changed /* release */ to  /* Release */ (both instances)
	               Defined BOUNDARY_NFIELDS, BOUNDARY_NRECORDS, BOUNDARY_FILE, BOUNDARY_SHORT_NAME and replaced all magic literals

bmribler (approved):

hl/src/H5LT.c
```diff
@@ -1918,12 +1935,17 @@ realloc_and_append(bool _no_user_buf, size_t *len, char *buf, const char *str_to
 {
     size_t size_str_to_add, size_str;
 
+    /* Precondition: buf must be allocated (verified in debug builds) */
+    assert(buf != NULL && "Buffer must not be NULL");
```
jhendersonHDF: If `_no_user_buf` is `false` and `str_to_add` is `NULL`, `buf` could be `NULL` since it's never touched. Based on the added header comment it seems like this may have been intentional behavior as well, but this function never "ensures space is available" in that case, it just returns `buf` (or `NULL`).

brtnfld: updated

jhendersonHDF: There's still a difference between debug and release builds here. debug builds will crash on the assertion failure, whereas release builds will return NULL

brtnfld: Addressed

hl/src/H5TB.c
```diff
@@ -3021,11 +3021,29 @@ H5TBget_field_info(hid_t loc_id, const char *dset_name, char *field_names[], siz
...
+            char  *member_name;
+            size_t name_len;
 
             if (NULL == (member_name = H5Tget_member_name(tid, (unsigned)i)))
                 goto out;
-            strcpy(field_names[i], member_name);
+
+            name_len = strlen(member_name);
+
+            /* Only enforce HLTB_MAX_FIELD_LEN limit when truncation is necessary.
```
jhendersonHDF: I'm not sure this comment really makes sense, as it doesn't seem to have anything to do with backward compatibility. Regardless of whether `memcpy()` copies the whole string or `HLTB_MAX_FIELD_LEN - 1` bytes, the NUL terminator needs to come at the end of the string either way, so the behavior should be the same. It's just that copying less than `HLTB_MAX_FIELD_LEN - 1` bytes could be more efficient.

brtnfld: Agreed,  the 'backward compatibility' rationale was wrong and has been removed. The comment now explains this is simply an efficiency optimization (copy only name_len + 1 bytes when the name fits, rather than the full HLTB_MAX_FIELD_LEN - 1). The H5TB.c changes have already landed in develop via PR #6371,

brtnfld:
:sunny:  SHINES project-related.

hl/src/H5LT.c
```diff
@@ -1909,7 +1914,17 @@ H5LTtext_to_dtype(const char *text, H5LT_lang_t lang_type)
...
- * Return:      void
+ * Parameters:
+ *   _no_user_buf: Operating mode flag
+ *                 - true:  Library-managed buffer (can reallocate)
+ *                 - false: User-provided buffer (fixed size, no realloc)
+ *   len:         Pointer to current buffer size (updated if reallocated)
+ *   buf:         Buffer to append to. If NULL, returns NULL immediately.
+ *   str_to_add:  String to append. If NULL:
+ *                  - Library-managed mode: reallocates to ensure extra space, no append
+ *                  - User-provided mode: no-op, returns buf unchanged (space cannot be ensured)
```
fortnern, line 1925: We should add a comment that this function must ALWAYS leave buf untouched on failure in order to maintain realloc-like behavior. In particular, the block starting on line 2232 requires this.

fortnern (changes requested):
Looks good other than a comment (or two) I'd like to see added

hl/src/H5LT.c
fortnern, line 1966: The function cannot fail after this point

brtnfld: added additional comment

hl/src/H5LT.c
brtnfld: updated the comment

fortnern:
I guess my comment was not entirely correct. Thanks for catching it. Actual realloc-like behavior might be ideal for a function called realloc(_and_append), but this is fine for now.

fortnern (approved):

PR #6145, opened 2026-01-07
Simplify MPI configuration in CMake

dutkalex:
To the best of my understanding, hdf5 only needs the C API for MPI (and in particular does not need the C++ binding which are a recurring source of configuration/build problems in practice). Therefore, the `find_package(MPI)` call can be modified to only look for the C library, which also simplifies a little bit the error checking (if the find package call succeeds, then `MPI_C_FOUND` is true).


hyoklee (dismissed):
@dutkalex , thank you for PR!

CMakeLists.txt
```diff
@@ -781,52 +781,48 @@ endif ()
...
-      set (LARGE_PARALLEL_IO OFF)
-    endif ()
+  # Require MPI standard 3.0 and greater
+  if (MPI_VERSION LESS 3)
+    message (FATAL_ERROR "HDF5 requires MPI standard 3.0 or greater")
+  endif ()
 
-    # Used by Subfiling VFD feature
-    CHECK_SYMBOL_EXISTS (MPI_Comm_split_type "mpi.h" H5_HAVE_MPI_Comm_split_type)
+  # MPI checks, only do these if MPI_C_FOUND is true, otherwise they always fail
```
brtnfld: This 2-line comment is not needed, it's explaining a guard that no longer exists against a situation that can no longer happen.

brtnfld:
Since these are related, would you mind also removing the dead code (HDF5 C++ never uses MPI C++ directly, and MPI C++ bindings are deprecated/removed from MPI implementations) in HDFCompilerCXXFlags.cmake:145

```
if (HDF5_ENABLE_PARALLEL AND MPI_CXX_FOUND)
  set (_HDF5_H5CC_CXX_COMPILER ${MPI_CXX_COMPILER})
else ()
  set (_HDF5_H5CC_CXX_COMPILER ${CMAKE_CXX_COMPILER})
endif ()
```

to

`set (_HDF5_H5CC_CXX_COMPILER ${CMAKE_CXX_COMPILER}`

Also, for MPI_C_FOUND in src/CMakeLists.txt:1435 could be cleaned as you did in your PR, if you don't mind. This would make everything consistent. Otherwise, looks good, thanks!

CMakeLists.txt
dutkalex: Indeed, this comment is no longer needed. I'll remove it for clarity.

config/flags/HDFCompilerCXXFlags.cmake
```diff
@@ -139,14 +139,7 @@ if (CMAKE_CXX_COMPILER_LOADED)
...
 
-  #-----------------------------------------------------------------------------
-  # Option to override the compiler for h5c++
-  #-----------------------------------------------------------------------------
-  if (HDF5_ENABLE_PARALLEL AND MPI_CXX_FOUND)
-    set (_HDF5_H5CC_CXX_COMPILER ${MPI_CXX_COMPILER})
-  else ()
-    set (_HDF5_H5CC_CXX_COMPILER ${CMAKE_CXX_COMPILER})
-  endif ()
+  set (_HDF5_H5CC_CXX_COMPILER ${CMAKE_CXX_COMPILER})
```
dutkalex, line 142: I'm not sure if this intermediate variable is still needed. It seems like it is only used in the next line, the two could easily be collapsed.

dutkalex:
@brtnfld I applied the changes you suggested

jhendersonHDF:
> and MPI C++ bindings are deprecated/removed from MPI implementations

The bindings may be deprecated, but the compiler wrapper is still there, which is what that variable is used for.

brtnfld:
> The bindings may be deprecated, but the compiler wrapper is still there, which is what that variable is used for.

True, still, the C++ API is a serial-only wrapper around the C API.

brtnfld (approved):

lrknox (approved):

PR #6166, opened 2026-01-21
[WIP] Updates to Digital Signature Verification for HDF5 Plugins

brtnfld:
### Appended Signature Format
Uses a platform-independent appended signature approach that works on Linux, Windows, and macOS:

> [!NOTE]
>This approach is still under discussion

###  src/H5PLsig.c
- H5PL__verify_signature_appended(): Main verification function

src/H5PLint.c

New function called: H5PL__verify_plugin_signature()
 - Isolated signature verification logic from the main plugin loading
 - Parallel MPI support: Intelligently chooses verification strategy:
  - Collective mode: Root rank verifies, broadcasts result to all ranks
  - Independent mode: Each rank verifies independently (fallback)
Serial mode: Direct verification

- Queries API context (DXPL) to determine MPI transfer mode
- Set HDF5_REQUIRE_SIGNED_PLUGINS (default: OFF)

----

> [!IMPORTANT]
> Adds digital signature verification for HDF5 plugins using an appended signature format, enhancing security by ensuring only verified plugins are loaded.
> 
>   - **Behavior**:
>     - Introduces digital signature verification for HDF5 plugins using an appended signature format.
>     - `H5PL__verify_signature_appended()` in `H5PLsig.c` verifies RSA signatures appended to plugin binaries.
>     - `H5PL__verify_plugin_signature()` in `H5PLint.c` handles verification logic, supporting both collective and independent MPI modes.
>     - Plugins are verified before loading in `H5PL__open()` in `H5PLint.c`.
>   - **Build Configuration**:
>     - Adds `H5PLsig.c` and `H5PLsig.h` to `CMakeLists.txt` for build integration.
>   - **Misc**:
>     - Removes old OpenSSL-based signature verification functions from `H5PLint.c` and `H5PLprivate.h`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 50af792c4a90a5a2c578a4b9428bc0296c2233da. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/H5PLsig.c
```diff
@@ -0,0 +1,233 @@
...
+    if (footer.magic != H5PL_SIG_MAGIC)
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
+                    "invalid signature magic number (not a signed HDF5 plugin or corrupted)");
+
+    /* Validate signature length */
+    if (footer.signature_length == 0 || footer.signature_length > 8192)
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL, "invalid signature length");
+
+    /* Calculate binary data size (file - signature - footer) */
+    binary_size = (size_t)(file_size - footer.signature_length - sizeof(H5PL_sig_footer_t));
```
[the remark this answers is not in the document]
brtnfld: fixed

.github/workflows/signed-plugins.yml
jhendersonHDF, line 1: Are there any signed and invalidly-signed plugins that can be used for this testing? Without that, I don't think any of this testing is particularly helpful since it looks like it's only mocking what we intend to test.

src/H5PLint.c
```diff
@@ -279,6 +290,84 @@ H5PL_load(H5PL_type_t type, const H5PL_key_t *key)
...
+    /* Check args */
+    assert(path);
+
+#ifdef H5_HAVE_PARALLEL
+    /* Try to get DXPL from API context to determine transfer mode */
+    dxpl_id = H5CX_get_dxpl();
+
+    /* Check if collective mode is active and we have a valid communicator */
+    if (dxpl_id != H5I_INVALID_HID) {
+        if (H5Pget_dxpl_mpio(dxpl_id, &xfer_mode) >= 0) {
```
jhendersonHDF: This should be converted to internal H5P calls, as public API calls shouldn't be made from within the library.

src/H5PLint.c
```diff
@@ -279,6 +290,84 @@ H5PL_load(H5PL_type_t type, const H5PL_key_t *key)
...
+            if (xfer_mode == H5FD_MPIO_COLLECTIVE) {
+                if (H5F_mpi_retrieve_comm(H5I_INVALID_HID, H5I_INVALID_HID, &comm) >= 0 &&
```
jhendersonHDF: This seems like it wouldn't work. Either the first or second parameter need to be valid; both can't be `H5I_INVALID_HID`.

src/H5PLsig.c
```diff
@@ -0,0 +1,237 @@
...
+#ifdef H5_REQUIRE_DIGITAL_SIGNATURE
+
+#include <openssl/rsa.h>
+#include <openssl/pem.h>
+#include <openssl/sha.h>
+#include <openssl/bio.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#ifdef H5_HAVE_WIN32_API
```
jhendersonHDF: Everything below here (and probably a few includes above) should all come from H5private.h and so are probably unnecessary here.

src/H5PLsig.c
```diff
@@ -0,0 +1,237 @@
...
+    RSA              *public_key    = NULL;
+    int               verify_result = 0;
+    herr_t            ret_value     = SUCCEED;
+
+    FUNC_ENTER_PACKAGE
+
+    assert(plugin_path);
+
+    /* Open plugin file for reading */
+    if ((fd = open(plugin_path, O_RDONLY | O_BINARY)) < 0)
```
jhendersonHDF: `HDopen` would be more appropriate here, as well as using the HD-prefixed versions of several other system call functions.

src/H5PLsig.c
```diff
@@ -0,0 +1,237 @@
...
+    /* Allocate buffer for binary data */
+    if (NULL == (binary_data = (unsigned char *)H5MM_malloc(binary_size)))
+        HGOTO_ERROR(H5E_PLUGIN, H5E_CANTALLOC, FAIL, "cannot allocate binary data buffer");
+
+    /* Seek to start of file */
+    if (lseek(fd, 0, SEEK_SET) < 0)
+        HGOTO_ERROR(H5E_PLUGIN, H5E_CANTSEEK, FAIL, "cannot seek to file start");
+
+    /* Read binary data */
+    if (read(fd, binary_data, binary_size) != (ssize_t)binary_size)
```
jhendersonHDF: Similar to above, `HDread` would probably be better here, but also there's more to consider here like making sure all the data was read since it could potentially take more than one call to `(HD)read`.

.github/workflows/signed-plugins.yml
glennsong09: I have some created that I use for testing, I can create some new fake ones for testing.

glennsong09 (approved):

PR #6167, opened 2026-01-22
Feature: Implement multithreaded reads in hdf5

OmkarAgnihotri:
This Pull Request implements a Proof-Of-Concept for multithreaded chunk 
reads for hyperslab selections over chunked datasets in hdf5.

The high level idea is to bypass the thread unsafe chunk cache (`H5D_rdcc_t`) 
and the page buffer (`H5PB_t`).

This works in two phases:
1. Serial phase
    - Accumulate chunk addresses/offsets
2. Parallel phase
    - Read raw data for the given chunks
    - Apply filter pipelines
    - Copy data to appropriate locations in the memory space

----

> [!IMPORTANT]
> Implements multithreaded chunk reads for hyperslab selections in HDF5, bypassing thread-unsafe components and using OpenMP for parallel processing.
> 
>   - **Behavior**:
>     - Implements multithreaded chunk reads for hyperslab selections in `H5D__chunk_read()` in `H5Dchunk.c`.
>     - Bypasses thread-unsafe `H5D_rdcc_t` and `H5PB_t`.
>     - Uses OpenMP for parallel processing.
>   - **Structures**:
>     - Adds `H5D_chunk_info_light_t` in `H5Dpkg.h` for parallel I/O operations.
>   - **Parallel Processing**:
>     - Serial phase: Accumulates chunk addresses/offsets.
>     - Parallel phase: Reads raw data, applies filters, and copies data to memory.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 8cc195a5f4414a57959a75c9f94c882c4cf0420d. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/H5Dchunk.c
+++ b/src/H5Dchunk.c
@@ -58,6 +58,9 @@
 #include "H5PBprivate.h" /* Page Buffer	                             */
 #include "H5SLprivate.h" /* Skip Lists                               */
 #include "H5VMprivate.h" /* Vector and array functions               */
+#include "H5Zprivate.h"  /* Data filters                             */
+#include "H5FDprivate.h" /* File drivers                             */
+#include <omp.h>         /* OpenMP                                   */
 
 /****************/
 /* Local Macros */
@@ -408,6 +411,12 @@ H5FL_BLK_DEFINE_STATIC(chunk);
 
 /* Declare extern free list to manage the H5S_sel_iter_t struct */
 H5FL_EXTERN(H5S_sel_iter_t);
+
+/* Declare extern free list to manage sequences of size_t */
+H5FL_SEQ_EXTERN(size_t);
+
+/* Declare extern free list to manage sequences of hsize_t */
+H5FL_SEQ_EXTERN(hsize_t);
 
 /*-------------------------------------------------------------------------
  * Function:    H5D__chunk_direct_write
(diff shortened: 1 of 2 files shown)


gheber:
Great work @OmkarAgnihotri . 

Just to clarify the use case/assumptions:

1. You are **not** looking for multiple application threads calling the public API (`H5Dread`), i.e., your proposal is something that would be supported in non-thread-safe library builds.
2. Fixed‑size datatype, no conversion/xform.
3. Simple hyperslab selections (no point selections).
4. Read‑only.
5. No filters (or only filters that are known re‑entrant and don’t touch global state).
6. Bypass the chunk cache and metadata cache where possible.

Is that a fair description?

OmkarAgnihotri:
> Just to clarify the use case/assumptions:
> 
> 1. You are **not** looking for multiple application threads calling the public API (`H5Dread`), i.e., your proposal is something that would be supported in non-thread-safe library builds.
> 2. Fixed‑size datatype, no conversion/xform.
> 3. Simple hyperslab selections (no point selections).
> 4. Read‑only.
> 5. No filters (or only filters that are known re‑entrant and don’t touch global state).
> 6. Bypass the chunk cache and metadata cache where possible.
> 
> Is that a fair description?

Yes, all the 6 points exactly describe the use case. Thanks @gheber for the clear summary.

gheber:
I'd add:

7. The VFD used is known to be re‑entrant for concurrent reads (which `sec2` is).
8. The OMP parallel region does only read + memcpy on disjoint regions.

PR #6174, opened 2026-01-26
Fix CMake config file install paths to follow platform conventions

lrknox:
Update HDF_DIR_PATHS macro to install hdf5-config.cmake and related files to platform-appropriate locations:
- Windows (not Cygwin): <prefix>/cmake/
- Linux/Unix/Cygwin: <prefix>/<lib_dir>/cmake/hdf5/
- macOS Frameworks: <framework_prefix>/hdf5.framework/Resources/CMake/


config/HDFMacros.cmake
```diff
@@ -442,18 +450,24 @@ macro (HDF_DIR_PATHS package_prefix)
       set (${package_prefix}_INSTALL_DATA_DIR ${CMAKE_INSTALL_DATADIR})
     endif ()
     if (NOT ${package_prefix}_INSTALL_CMAKE_DIR)
-      set (${package_prefix}_INSTALL_CMAKE_DIR ${CMAKE_INSTALL_LIBDIR}/cmake)
+      # CMake config install directory:
+      # - Frameworks: <framework>/Resources/CMake/
+      # - Windows (not Cygwin): cmake/
+      # - Linux/Unix/Cygwin: ${CMAKE_INSTALL_LIBDIR}/cmake/<package_name>/
+      if (${package_prefix}_BUILD_FRAMEWORKS)
```
jhendersonHDF: I believe only the Unixy changes need to be put in this particular spot. It seems unlikely that `HDF5_USE_GNU_DIRS` is going to be ON on Windows or Apple. There may be a use case for Apple, but that would seem to go against the frameworks approach anyway.

mattjala:
Superseded by #6410

Issue #6161, opened 2026-01-19
filters: `can_apply` does nothing for optional filters


crusaderky:
From the [documentation of H5Zregister](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_z.html#title6):

> If this filter is enabled through [H5Pset_filter()](https://support.hdfgroup.org/documentation/hdf5/latest/group___o_c_p_l.html#gae7ac4c110e9ec3fcf38966928b96cda3) as optional and the can apply function returns 0, the library will skip the filter in the filter pipeline.

The above does not seem to be implemented.
The current behaviour is that
1. `H5Pset_filter(..., H5Z_FLAG_MANDATORY, ...)` later causes the call to `H5Dcreate2` to execute `can_apply` and then fail if it returned 0, as expected.
2. `H5Pset_filter(..., H5Z_FLAG_OPTIONAL, ...)` later causes
   a. the call to `H5Dcreate2` to execute `can_apply` and, even after it returns 0, to execute `set_local`.
  b. the call to `H5DWrite` / `H5DRead` to invoke the `filter` callback for every chunk, even if `can_apply` earlier returned 0.

This behaviour can be seen [here](https://github.com/HDFGroup/hdf5/blob/476c60961c4148c31e738cd1e3e9a253c6917cf8/src/H5Z.c#L799-L818), where the return value of 0 is disregarded.

## Reproducer
#6162

## Workaround
A filter developer could change `set_local` to internally rerun `can_apply`. if it returns 0, `set_local` would set a guard value in `cd_values` and return 0. The `filter` callback would then need to perform an early check for the guard in cd_values.

## Context
I found this issue while investigating #5942.

## Version
hdf5 git tip as of today (476c60961c4148c31e738cd1e3e9a253c6917cf8)

crusaderky:
[test_can_apply2](https://github.com/HDFGroup/hdf5/blob/dd3080a58cc6bb86f3b34284399915da9e513262/test/dsets.c#L6439) in theory tests this functionality.
What it actually tests, however, is that when a filter is optional than `filter()` can quietly return 0.
Adding `assert(0)` to `filter_bogus3` demonstrates that the filter callback is still being invoked.

PR #6178, opened 2026-01-28
Apply filters to variable-length data

### LINKED ISSUE #5942: `can_apply` and `set_local` filter callbacks skipped for `H5T_VARIABLE` dtypes
Reproduced on
- hdf5 1.14.6, Linux x64
- hdf5 2.0.0.4 git tip (cd2414b212c33fa847c2be94740172aa9746de1a)

# Issue description

If I create a variable string dtype

```C
dtype = H5Tcopy(H5T_C_S1);
assert(dtype >= 0);
r = H5Tset_size(dtype, H5T_VARIABLE);
assert(r >= 0);
```
and then I create a dataset with a filter
```C
r = H5Pset_filter(plist, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 0, NULL);
assert(r >= 0);
dset = H5Dcreate(fid, "dset", dtype, sid, H5P_DEFAULT, plist, H5P_DEFAULT);
assert(dset >= 0);
```
then the `can_apply` and `set_local` callbacks previously defined by `H5Zregister` are not invoked by `H5DCreate`. Nothing in the documentation hints at the potential of these callbacks to be skipped.

# Impact
Both [hdf5-blosc](https://github.com/Blosc/hdf5-blosc) and [hdf5-BLOSC2](https://github.com/Blosc/hdf5-BLOSC2) rely on `set_local` to automatically amend `cd_values`. If `set_local` doesn't run, `filter` crashes.

What follows is two examples of typical user configuration:

```C
unsigned int cd_values[7];
/* 0 to 3 (inclusive) param slots are reserved. */
cd_values[4] = 4;       /* compression level */
cd_values[5] = 1;       /* 0: shuffle not active, 1: shuffle active */
cd_values[6] = BLOSC_BLOSCLZ; /* the actual compressor to use */
r = H5Pset_filter(plist, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 7, cd_values);
assert(r >= 0);
```
in the above example, `blosc_set_local` fills in the previously uninitialised `cd_values[0:4]`, so that `blosc_filter` can use them. Crucially, this includes the size of the data type, which is not retrievable from the filter callback itself.

```C
r = H5Pset_filter(plist, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 0, NULL);
assert(r >= 0);
```
In the above example, `blosc_set_local` bumps up `cd_nelmts` to 4 and fills `cd_values`.

# Workaround
## hdf5-blosc
The final user can explicitly initialise `cd_values[0:4]` to work around the problem:
```C
unsigned int cd_values[7] = {
    FILTER_BLOSC_VERSION, BLOSC_VERSION_FORMAT, 1, 0,
    4, 1, BLOSC_BLOSCLZ
};
r = H5Pset_filter(plist, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 7, cd_values);
assert(r >= 0);
```
or just the bare minimum:
```C
unsigned int cd_values[4] = {FILTER_BLOSC_VERSION, BLOSC_VERSION_FORMAT, 1, 0};
r = H5Pset_filter(plist, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 4, cd_values);
assert(r >= 0);
```
## hdf5-BLOSC2
There is no simple workaround for Blosc2, short of manually invoking `blosc2_set_local`, because the `cd_values` are heavily data-dependent and non-trivial.

# Reproducers
## Internal
#6162

## In hdf5-blosc
https://github.com/Blosc/hdf5-blosc/pull/38

```bash
$ git clone https://github.com/crusaderky/hdf5-blosc.git 
$ cd hdf5-blosc
$ git checkout strings
$ pixi r build
$ pixi r test
[...]
 9/10 Test  #9: test_strings ...........................Subprocess aborted***Exception:   1.11 sec
test_strings: /home/crusaderky/github/tmp/hdf5-blosc/src/blosc_filter.c:176: blosc_filter: Assertion `cd_nelmts >= 4' failed.
```

crusaderky:
- Closes #5942. A more detailed explanation of the rationale for this PR is in the linked issue.
- *Mandatory* filters can be now applied to variable-length data. This notably includes h5py object strings and NpyStrings.
- *Optional* filters applied to variable-length data no longer skip `can_apply` and `set_local`, only to jump directly to `filter`. This fixes hdf5_blosc and HDF5_Blosc2, which rely on `set_local`. Note that all filters which have neither `can_apply` nor `set_local` have always worked fine with vlen data, as long as they were tagged as optional. Notable examples are deflate, lzf, bzip2 (from hdf5plugin), and lz4 (also from hdf5plugin).
- Lossy filters, such as scale-offset, are and remain untested and most likely broken, as I expect them to treat the vlen metadata as numbers and corrupt it. The right place to ensure proper behaviour is their `can_apply` function, which however is broken (#6161).

----

> [!IMPORTANT]
> Enhance filter application to variable-length data, fixing optional filter handling and adding tests.
> 
>   - **Behavior**:
>     - Mandatory filters can now be applied to variable-length data, including h5py strings.
>     - Optional filters no longer skip `can_apply` and `set_local` steps, fixing issues with hdf5_blosc and HDF5_Blosc2.
>     - Lossy filters remain untested and potentially broken for vlen data.
>   - **Code Changes**:
>     - Modify `H5Z_ignore_filters()` in `H5Z.c` to handle optional filters correctly.
>     - Add tests in `dsets.c` for scalar and null datasets with optional filters, and for `set_local` updates with vlen data.
>   - **Tests**:
>     - Add `test_optional_filters_scalar`, `test_optional_filters_null`, `test_set_local_updates_cd`, `test_set_local_updates_cd_vlen`, and `test_deflate_vlen` in `dsets.c`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 70ab78aa874d0b9d8bed9240db6975d23f9ed5b5. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


test/dsets.c
```diff
@@ -6635,35 +6781,87 @@ test_optional_filters(hid_t file)
...
+        H5Dclose(dsid);
+        H5Sclose(sid);
+        H5Pclose(dcplid);
+    }
+    H5E_END_TRY
+    return FAIL;
+} /* end test_optional_filters_scalar() */
+
+/*-------------------------------------------------------------------------
+ * Function:    test_optional_filters_null
```
crusaderky, line 6896: this was previously untested.

test/dsets.c
```diff
@@ -17752,6 +17948,345 @@ test_vds_shared_strings(hid_t fapl)
...
+        H5Dclose(dsid);
+        H5Sclose(sid);
+        H5Pclose(dcpl);
+    }
+    H5E_END_TRY
+    return FAIL;
+} /* end test_set_local_updates_cd() */
+
+/*-------------------------------------------------------------------------
+ * Function:    test_set_local_updates_cd_vlen
```
crusaderky, line 18151: This test fails in the main branch, complaining that filters can't be applied to vlen data.
If I replace H5Z_FLAG_MANDATORY with H5Z_FLAG_OPTIONAL below, it crashes on `assert(0)` inside the `filter` callback, highlighting that `set_local` has not been called.

test/dsets.c
```diff
@@ -17752,6 +17948,345 @@ test_vds_shared_strings(hid_t fapl)
...
+        H5Sclose(sid);
+        H5Tclose(dtype);
+        H5Pclose(dcpl);
+    }
+    H5E_END_TRY
+    return FAIL;
+} /* end test_set_local_updates_cd_vlen() */
+
+/*-------------------------------------------------------------------------
+ * Function:    test_deflate_vlen
```
crusaderky, line 18272: This tests what happens when `filter` actually processes the contiguous raw chunk data.

This test fails in the main branch, complaining that filters can't be applied to vlen data.
However, if I replace `H5Z_FLAG_MANDATORY` with `H5Z_FLAG_OPTIONAL` below, it passes in the main branch, thanks to the fact that the deflate filter has no `set_local` callback.

Running deflate on more substantial data shows that the output file size is much smaller than the uncompressed version.

crusaderky:
Downstream tests in hdf5-blosc: https://github.com/Blosc/hdf5-blosc/pull/38

test/dsets.c
```diff
@@ -6581,50 +6712,66 @@ test_can_apply2(hid_t file)
...
-    /* The filter is optional. */
-    if (H5Pset_filter(dcplid, H5Z_FILTER_DEFLATE, H5Z_FLAG_OPTIONAL, cd_nelmts, cd_values) < 0)
+    /* Create dataset with optional filter */
+    if ((dsid = H5Dcreate2(file, DSET_OPTIONAL_SCALAR, H5T_NATIVE_INT, sid, H5P_DEFAULT, dcplid,
+                           H5P_DEFAULT)) < 0)
         TEST_ERROR;
 
-    /* Create dataset with optional filter */
-    if ((dsid = H5Dcreate2(file, DSET_OPTIONAL_SCALAR, strtid, sid, H5P_DEFAULT, dcplid, H5P_DEFAULT)) < 0)
+    /* Write data to the dataset */
```
crusaderky, line 6842: This checks that the `filter` callback is never invoked, which the previous version of this test left untested.

crusaderky:
@hyoklee is it possible to avoid having CI re-authorised by a maintainer every time I need to fix a trivial error? This is hurting velocity quite badly.

jhendersonHDF:
@crusaderky I believe this is simply a requirement due to our current GitHub configuration. I've approved our workflows to run again, but I believe we'll need to continue re-approving them as necessary. I'll keep an eye on this PR, but feel free to reach out again as needed.

crusaderky:
@jhendersonHDF ready for another go

crusaderky:
CI is green; this is ready for review and potentially for merging.

crusaderky:
@jhendersonHDF @hyoklee gentle ping

jhendersonHDF:
Thanks for the ping @crusaderky. We're currently working on getting together the next release of HDF5, but should be able to get to this after that. I think we'll want to discuss internally for a bit, mostly around the historical reason for why filters were disabled for variable-length data and if that's still relevant.

fortnern:
I'll try to review this soon, but I want to point out that the filters will not operate on the actual variable length data without much more work. They will only operate on the global heap IDs, which are analogous to pointers to the variable length data.

fortnern:
I didn't realize they're disabled, I'll try to find out the reason they were disabled in 2019. It makes sense you might want to compress a compound containing variable length data as well as fixed length data.

crusaderky:
> the filters will not operate on the actual variable length data without much more work. They will only operate on the global heap IDs, which are analogous to pointers to the variable length data.

This is true, but there is still value for short data - which is very common in real world databases.

as of h5py git tip, hdf5 git tip:

```python
import os
import h5py
import numpy as np

datas = {
    "long": "x" * 100_000,
    # ~average length of first name + surname
    # https://www.researchgate.net/figure/First-names-and-last-names-lengths-distributions_fig1_328894441
    "short": "xxxxxxx yyyyyyyyy",
    "very_short": "xxx",
}

for prefix, data in datas.items():
    data = np.tile(np.asarray([data], dtype="T"), 10_000)
    for suffix, kwargs in [("plain", {}), ("compressed", {"compression": "gzip", "compression_opts": 9})]:
        fname = f"{prefix}.{suffix}.h5"
        with h5py.File(fname, "w") as f:
            f.create_dataset("data", data=data, chunks=(1000, ), **kwargs)
        print(f"{os.path.getsize(fname):-10d} {fname}")
```
Output:
```
1000483496 long.plain.h5
1000356163 long.compressed.h5
    564904 short.plain.h5
    423290 short.compressed.h5
    425640 very_short.plain.h5
    283976 very_short.compressed.h5
```

gzip gives you -25% space savings on a <name surname> field, and a 33% saving on what could be a 3-letter tag (which was saved as variable-length on the assumption that it may rarely be much larger than that).

> It makes sense you might want to compress a compound containing variable length data as well as fixed length data.

Absolutely! I would however leave it as a follow-up.

test/dsets.c
```diff
@@ -1914,6 +1926,105 @@ filter_bogus3(unsigned int H5_ATTR_UNUSED flags, size_t H5_ATTR_UNUSED cd_nelmts
...
+/*------------------------------------------------------------------------
+ * Function:  can_apply_crash
+ *
+ * Purpose:   A 'can apply' callback that crashes if ever called.
+ *------------------------------------------------------------------------
+ */
+static htri_t
+can_apply_crash(hid_t H5_ATTR_UNUSED dcpl_id, hid_t H5_ATTR_UNUSED type_id, hid_t H5_ATTR_UNUSED space_id)
+{
+    assert(0);
```
fortnern: I'm not a fan of using assertions to trigger a test failure. Can you change this to perhaps modify a global variable to indicate if it was called?

test/dsets.c
```diff
@@ -1914,6 +1926,105 @@ filter_bogus3(unsigned int H5_ATTR_UNUSED flags, size_t H5_ATTR_UNUSED cd_nelmts
...
+/*---------------------------------------------------------------------
+ * Function:  set_local_crash
+ *
+ * Purpose:   A 'set local' callback that crashes if ever called.
+ *---------------------------------------------------------------------
+ */
+static herr_t
+set_local_crash(hid_t H5_ATTR_UNUSED dcpl_id, hid_t H5_ATTR_UNUSED type_id, hid_t H5_ATTR_UNUSED space_id)
+{
+    assert(0);
```
fortnern: Same here

test/dsets.c
```diff
@@ -1914,6 +1926,105 @@ filter_bogus3(unsigned int H5_ATTR_UNUSED flags, size_t H5_ATTR_UNUSED cd_nelmts
...
+ *
+ * Purpose:   A 'filter' callback that crashes if ever called.
+ *-------------------------------------------------------------------
+ */
+static size_t
+filter_crash(unsigned int H5_ATTR_UNUSED flags, size_t H5_ATTR_UNUSED cd_nelmts,
+             const unsigned int H5_ATTR_UNUSED *cd_values, size_t H5_ATTR_UNUSED nbytes,
+             size_t H5_ATTR_UNUSED *buf_size, void H5_ATTR_UNUSED **buf)
+{
+    assert(0);
```
fortnern: Same here

test/dsets.c
```diff
@@ -1914,6 +1926,105 @@ filter_bogus3(unsigned int H5_ATTR_UNUSED flags, size_t H5_ATTR_UNUSED cd_nelmts
...
+ *            Failure:    0
+ *-------------------------------------------------------------------------
+ */
+static size_t
+filter_update_cd(unsigned int H5_ATTR_UNUSED flags, size_t cd_nelmts, const unsigned int *cd_values,
+                 size_t nbytes, size_t H5_ATTR_UNUSED *buf_size, void H5_ATTR_UNUSED **buf)
+{
+    // Note: don't return 0, as 0 causes hdf5 to quietly skip
+    // the filter when it is optional.
+    assert(cd_nelmts == 1 && cd_values[0] == 42);
```
fortnern: Again please remove the assertion here and have the filter return failure and/or modify a globla variableif there is a failure. You could also place the error status in buf I suppose

test/dsets.c
```diff
@@ -6581,50 +6712,66 @@ test_can_apply2(hid_t file)
...
         TEST_ERROR;
 
     /* Create the data space */
     if ((sid = H5Screate(H5S_SCALAR)) < 0)
         TEST_ERROR;
 
-    /* The filter is optional. */
-    if (H5Pset_filter(dcplid, H5Z_FILTER_DEFLATE, H5Z_FLAG_OPTIONAL, cd_nelmts, cd_values) < 0)
+    /* Create dataset with optional filter */
+    if ((dsid = H5Dcreate2(file, DSET_OPTIONAL_SCALAR, H5T_NATIVE_INT, sid, H5P_DEFAULT, dcplid,
```
fortnern: Can you reinstate the test with the deflate filer and and add this test as a separate case (either later in this function or in a new function)? We should generally not remove test cases unless there's a good reason to do so

test/dsets.c
```diff
@@ -6581,50 +6712,66 @@ test_can_apply2(hid_t file)
...
+    TESTING("scalar dataset with optional filters");
 
     /* Create dcpl with special filter */
     if ((dcplid = H5Pcreate(H5P_DATASET_CREATE)) < 0)
         TEST_ERROR;
 
-    /* Create the datatype */
-    if ((strtid = H5Tcreate(H5T_STRING, H5T_VARIABLE)) < 0)
+    /* The filter is optional. Callbacks will crash if ever invoked. */
+    if (H5Pset_filter(dcplid, H5Z_FILTER_CRASH, H5Z_FLAG_OPTIONAL, 0, NULL) < 0)
```
fortnern: Where is this filter registered?

fortnern (changes requested):
Library changes look good, some comments about the tests

fortnern:
I've confirmed that disabling can_apply and set_local for vlens was a mistake

fortnern:
> > It makes sense you might want to compress a compound containing variable length data as well as fixed length data.
> 
> Absolutely! I would however leave it as a follow-up.

This should actually work fine with just this change. We can write a test for it at some point if it's not already tested (it's possible the test passes because it uses a filter that doesn't use can_apply or set_local).

test/dsets.c
crusaderky: It wasn't. I think the test worked because hdf5 skips the whole filter stack due to the data type of the dataset. This highlights that H5Pset_filter is lacking input validation (out of scope for this PR).

crusaderky:
> > > It makes sense you might want to compress a compound containing variable length data as well as fixed length data.
> > 
> > 
> > Absolutely! I would however leave it as a follow-up.
> 
> This should actually work fine with just this change. We can write a test for it at some point if it's not already tested (it's possible the test passes because it uses a filter that doesn't use can_apply or set_local).

My demo above does not change its output after this PR, so the bit that feeds the actual data through the filter is missing.

test/dsets.c
crusaderky: I've reinstated the old test, minus the part that creates a VLEN dtype which is conceptually incorrect now, as VLEN is a valid dtype for filters.

However I still think that that test is unsalvageable. _Before and after this PR_ the test passes as the apply callback in the DEFLATE filter is never executed, neither for scalar types, nor for vlen, because nothing writes/reads data on the dataset, and the filter does not have the other two callbacks that would be executed by H5S_CREATE. You could pass to the test a filter where the apply callback points to 0xDEADBEEF (segfault on call) and it would still pass!

## develop branch:
- `test_optional_filters` tests creating a scalar type dataset, without writing or reading on it, with DEFLATE and verifies it doesn't crash
- `test_optional_filters` then tests creating a VLEN type dataset, without writing or reading on it, with DEFLATE and verifies it doesn't crash

## this PR:
- `test_optional_filters` tests creating a scalar type dataset, without writing or reading on it, with DEFLATE and verifies it doesn't crash *[reinstated in last commit]*
- `test_optional_filters_scalar` tests creating a scalar type dataset, reading, and writing on it with a custom filter, and tests that the filter callbacks are never invoked
- `test_optional_filters_null` tests creating a null type dataset, reading, and writing on it with a custom filter, and tests that the filter callbacks are never invoked

crusaderky:
@fortnern all comments have been addressed. This is ready for another round of review.

src/H5Z.c
```diff
@@ -1058,27 +1058,27 @@ H5Z_set_local_direct(const H5O_pline_t *pline)
...
+ *      of can_apply.
  *
  * Return:   Non-negative(true/false) on success
  *           Negative on failure
  *
  *-------------------------------------------------------------------------
  */
 htri_t
-H5Z_ignore_filters(hid_t dcpl_id, const H5T_t *type, const H5S_t *space)
+H5Z_ignore_filters(hid_t dcpl_id, const H5_ATTR_UNUSED H5T_t *type, const H5S_t *space)
```
fortnern: Sorry I missed this the first time, but you can just remove the type parameter

fortnern (dismissed):

src/H5Z.c
crusaderky:
```suggestion
H5Z_ignore_filters(hid_t dcpl_id, void H5_ATTR_UNUSED *type, const H5S_t *space)
```

crusaderky:
@fortnern I applied the nit you suggested but it removed your approval

src/H5Z.c
```diff
@@ -1058,27 +1058,27 @@ H5Z_set_local_direct(const H5O_pline_t *pline)
...
+ *      of can_apply.
  *
  * Return:   Non-negative(true/false) on success
  *           Negative on failure
  *
  *-------------------------------------------------------------------------
  */
 htri_t
-H5Z_ignore_filters(hid_t dcpl_id, const H5T_t *type, const H5S_t *space)
+H5Z_ignore_filters(hid_t dcpl_id, void H5_ATTR_UNUSED *type, const H5S_t *space)
```
fortnern:
```suggestion
H5Z_ignore_filters(hid_t dcpl_id, const H5S_t *space)
```

You will also need to change the prototype declaration in H5Zprivate.h

fortnern: And the invocation in H5D__create()

fortnern: I went ahead and pushed the changes myself, I hope that's ok. We're trying to get this in the release.

fortnern (approved):

jhendersonHDF (approved):

PR #6191, opened 2026-02-03
Fix incorrect libsz/libaec entries in hdf5.pc when using libaec to provide SZIP

Growl1234:
When HDF5 is built with SZIP support enabled and use system-installed libaec to provide SZIP, the generated hdf5.pc file will contain invalid library entries such as:

```
Libs.private:  -lm -ldl -lz -llibname:NOTFOUND -llibname:NOTFOUND
```

This occurs because the imported targets for libsz and libaec do not have a meaningful OUTPUT_NAME, causing CMake to fail to resolve the correct library names when generating Libs.private.

This PR manually sets the OUTPUT_NAME property for the corresponding imported targets in `CMakeFilters.cmake`, like what's done for ZLIB. As a result, the generated hdf5.pc correctly reports:

```
Libs.private:  -lm -ldl -lz -lsz -laec
```

instead of invalid placeholders.

----

> [!IMPORTANT]
> Fixes incorrect library entries in `hdf5.pc` by setting `OUTPUT_NAME` for libaec targets in `CMakeFilters.cmake`.
> 
>   - **Behavior**:
>     - Fixes incorrect library entries in `hdf5.pc` when using libaec for SZIP support.
>     - Sets `OUTPUT_NAME` for `libaec::sz` and `libaec::aec` in `CMakeFilters.cmake` to ensure correct library names.
>   - **Files**:
>     - Modifies `CMakeFilters.cmake` to set `OUTPUT_NAME` for imported targets `libaec::sz` and `libaec::aec`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 7478c3cc97d457280ed1ffd4674cc9278129ee1c. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


CMakeFilters.cmake
```diff
@@ -229,6 +229,12 @@ if (HDF5_ENABLE_SZIP_SUPPORT)
           set (HDF5_ENABLE_SZIP_SUPPORT OFF CACHE BOOL "" FORCE)
           message (FATAL_ERROR "SZIP support in HDF5 was enabled and libaec was found, but no libsz compatibility library was found")
         endif ()
+        get_filename_component (libname ${libsz_LIBRARIES} NAME_WLE)
```
jhendersonHDF: Similar to the above for zlib, this probably needs to be wrapped in a

```CMake
if (NOT WIN32)
    ...
endif ()
```

section so that an improper `OUTPUT_NAME` isn't set for Windows.

Growl1234: Thanks for reviewing! This if condition has been added.

brtnfld (dismissed):

CMakeFilters.cmake
```diff
@@ -229,6 +229,14 @@ if (HDF5_ENABLE_SZIP_SUPPORT)
           set (HDF5_ENABLE_SZIP_SUPPORT OFF CACHE BOOL "" FORCE)
           message (FATAL_ERROR "SZIP support in HDF5 was enabled and libaec was found, but no libsz compatibility library was found")
         endif ()
+        if (NOT WIN32)
+          get_filename_component (libname ${libsz_LIBRARIES} NAME_WLE)
+          string (REGEX REPLACE "^lib" "" libname ${libname})
+          set_target_properties (libaec::sz PROPERTIES OUTPUT_NAME ${libname})
```
jhendersonHDF: I tested this a bit both with libaec built with Autotools and with CMake and I think we'll need a slightly more comprehensive solution. This works fine for Autotools-built libaec, but fails for CMake-built libaec because their config file creates `libaec::sz` and `libaec::aec` as ALIAS targets to, for example, `libaec::aec-shared` and `set_target_properties` can't be called on ALIAS targets. Rather than carrying forward this approach of setting the `OUTPUT_NAME` of a target (which is a bit odd IMO), we should probably just try to determine the correct flags from `INTERFACE_LINK_LIBRARIES` and use those, which is what I mostly have done at https://github.com/HDFGroup/hdf5/blob/develop/src/CMakeLists.txt#L1374-L1451. I think the same just needs to be applied to https://github.com/HDFGroup/hdf5/blob/develop/src/CMakeLists.txt#L1453-L1458 below it that I left out at the time.

Growl1234: Sorry that I may not be very familiar with this area, so I ended up adding a simple additional step: checking whether the target is an `ALIASED_TARGET`, and if so, resetting it to the actual one. In my subsequent tests, both cases — libaec built with CMake and libaec built with Autotools — worked as expected.

I still look forward to future improvements in `src/CMakeLists.txt`, though, since above is just a tempory fix anyway. If such changes are done, the special handling currently required for ZLIB may no longer be necessary as well.

jhendersonHDF (approved):
Thanks for the PR @Growl1234! It appears this will work for now while I update the pkgconfig generation logic to handle this without needing to overwrite `OUTPUT_NAME` on targets.

lrknox (approved):

PR #6195, opened 2026-02-03
Fixes the "multiple @param documentation sections" doxygen warnings

brtnfld:
Addresses the "multiple @ param documentation sections" warnings are caused by the interaction of three things:

1. MACRO_EXPANSION = YES with EXPAND_ONLY_PREDEF = NO in Doxyfile.in

2. Versioned API macros in H5version.h like:

\def documentation in APIVersions.dox like:

3. \def H5Tarray_create \api_vers_2{H5Tarray_create,H5Tarray_create1,H5Tarray_create2}

Because Doxygen expands all macros, it sees H5Tarray_create as an alias for H5Tarray_create1. It then associates documentation from both the \def block (for the macro) and the function's own doc block (with \param entries) with the same function — resulting in "multiple @ param documentation sections."

----

> [!IMPORTANT]
> Fix Doxygen warnings by adding predefined macros in `CMakeLists.txt` and adjusting Fortran code for conditional compilation.
> 
>   - **Doxygen Configuration**:
>     - Add versioned API macros to `DOXYGEN_PREDEFINED` in `CMakeLists.txt` to prevent Doxygen from expanding macros in `H5version.h`, avoiding multiple `@param` warnings.
>   - **Fortran Source**:
>     - Add `#ifdef H5_DOXYGEN` condition in `H5FDff.F90` to handle Doxygen-specific documentation sections.
>     - Adjust `h5fdsubfiling_get_file_mapping_f` to conditionally compile based on `H5_DOXYGEN` and `H5_FORTRAN_HAVE_CHAR_ALLOC`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for c65e24bf67414090ce2e7358dc4e8dd50b64f8a2. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


fortran/src/H5FDff.F90
```diff
@@ -151,7 +151,9 @@ END FUNCTION h5free_string_array_memory_c
 SUBROUTINE h5fdsubfiling_get_file_mapping_f(file_id, filenames, num_files, hdferr)
   IMPLICIT NONE
   INTEGER(HID_T), INTENT(IN) :: file_id
-#ifdef H5_FORTRAN_HAVE_CHAR_ALLOC
+#ifdef H5_DOXYGEN
+  CHARACTER(LEN=:), ALLOCATABLE, DIMENSION(:), INTENT(OUT) :: filenames
+#elif defined(H5_FORTRAN_HAVE_CHAR_ALLOC)
```
bmribler, line 156: Is the purpose of having #ifdef and #elif separately for clarity?

doxygen/CMakeLists.txt
```diff
@@ -32,7 +32,64 @@ if (DOXYGEN_FOUND)
...
+    H5_HAVE_MAP_API
+    H5_HAVE_PARALLEL
+    H5_HAVE_ROS3_VFD
+    H5_DOXYGEN
+    H5_HAVE_SUBFILING_VFD
+    H5_HAVE_IOC_VFD
+    H5_HAVE_MIRROR_VFD
+    "H5std_string=std::string"
+    "H5G_link_t=H5L_type_t"
+    "H5Acreate=H5Acreate2"
```
jhendersonHDF: Will these need to be maintained going forward?

brtnfld: only if addition versioned APIs get added, as they will also probably trigger warnings

fortran/src/H5FDff.F90
brtnfld: yes

doxygen/CMakeLists.txt
jhendersonHDF: Does it make a difference if the aliases point to the latest version of a function or an earlier version though? What if `H5Acreate3` is introduced?

brtnfld: You would need to do "H5Acreate=H5Acreate3"

brtnfld: I could autogenerate the list with bin/make_vers

brtnfld: done.

fortran/src/H5FDff.F90
brtnfld: Also for documenation generation, gets rid of oxygen warnings

1. #ifdef H5_DOXYGEN - When Doxygen processes the code, it only sees the clean, ideal interface signature. 

2. #elif defined(H5_FORTRAN_HAVE_CHAR_ALLOC) - The actual compile-time check for Fortran 2003+ compilers that support allocatable character strings.

3. #else - Fallback for older compilers using fixed-length strings.

bmribler (dismissed):

jhendersonHDF (dismissed):

lrknox (dismissed):

lrknox (approved):

CMakeLists.txt
```diff
@@ -1147,7 +1147,7 @@ if (HDF5_BUILD_DOC AND EXISTS "${HDF5_DOXYGEN_DIR}" AND IS_DIRECTORY "${HDF5_DOX
     else ()
       set (HDF5_DOXY_WARNINGS "NO")
     endif ()
-    message (VERBOSE "Doxygen version: ${DOXYGEN_VERSION}")
+    message (STATUS "Found Doxygen version: ${DOXYGEN_VERSION}")
```
jhendersonHDF, line 1150: Do we want this printed out on every build?

lrknox: Wouldn't it be only for builds where documentation HDF5_BUILD_DOC was ON (default is off)?

jhendersonHDF: Yes, I assumed ON would be the default. I'm fine with either way, we just have several STATUS-level messages in the build currently that are really just noise.

jhendersonHDF (approved):

PR #6198, opened 2026-02-04
✨[Feature] Digital Signature Verification for HDF5 Plugins

### LINKED ISSUE #5116: Allow only digitally signed plugins to be loaded
We'd like to add a feature that checks if plugins are digitally signed before loading them. This would apply to any plugin: filter, VFD, and VOL connector.

brtnfld:
## Description
This PR introduces a robust security framework for digitally signing and verifying HDF5 dynamic plugins (filters). This feature ensures that the HDF5 library loads only trusted, unmodified plugins, significantly enhancing security in environments where plugins are distributed dynamically.

### 1. Plugin Signature Verification
*   **Mechanism:** Implements an append-style signature verification where RSA signatures are attached to the plugin binary footer. This allows signed plugins to remain compatible with standard system loaders while enabling cryptographic verification by HDF5.
*   **Crypto:** Utilizes **OpenSSL (1.1.0+)** for RSA signature verification (SHA-256, SHA-384, SHA-512 supported).
*   **Key Management:** Introduces a "KeyStore" concept. Public keys can be loaded from:
    *   A configured directory (`HDF5_PLUGIN_KEYSTORE_DIR`).
    *   An environment variable (`HDF5_PLUGIN_KEYSTORE`).
    *   A compile-time embedded key (fallback).
*   **Performance:** Implements an internal signature cache to prevent redundant verification of the same plugin file, ensuring minimal performance impact after the first load.
*   **Revocation:** Adds support for a revocation list (blocklist) to reject specific known-bad signatures.

### 2. New Tool: `h5sign`
*   Added a new command-line tool, `h5sign`, located in `tools/src/h5sign`.
*   Allows plugin developers to sign their binaries using a private key.
*   Supports verification mode to check existing signatures.

### 3. Build & Configuration Changes
*   **New CMake Option:** `HDF5_REQUIRE_SIGNED_PLUGINS` (Default: `OFF`). When enabled, HDF5 enforces signature verification and refuses to load unsigned or invalid plugins.
*   **Dependencies:** Adds `OpenSSL` as a required dependency when signed plugins are enabled.

### 4. CI/CD & Infrastructure Updates
*   **New Workflow:** Added `.github/workflows/signed-plugins.yml` to specifically test the new signature verification logic in serial and parallel configurations.

### New Files
*   `src/H5PLsig.c` / `.h`: Core logic for reading signatures, managing the KeyStore, and performing OpenSSL verification.
*   `tools/src/h5sign/`: Source code for the signing utility.
*   `release_docs/PLUGIN_SIGNATURE_README.md`: Comprehensive documentation for developers and users regarding key generation, signing, and air-gapped environment handling.
*   `test/test_plugin_signature.c`: New test suite covering valid signatures, tampered binaries, and invalid keys.

### Modified Files
*   `src/H5PLint.c`: Hooked into the plugin loading process to trigger verification before `dlopen`/`LoadLibrary`.
*   `CMakeLists.txt` & `src/CMakeLists.txt`: build logic for OpenSSL linking and new source files.
*   `.github/workflows/*`:  CI updates

## Testing
*   Added `h5signverifytest` and `test_plugin_signature` to the test suite.
*   Verified that unsigned plugins are rejected when the feature is enabled.
*   Verified that tampered plugins (bit-flipped) are rejected.
*   Verified support for multiple trusted keys in a single KeyStore.
*   Validated caching mechanisms to ensure performance.

## Documentation
*   Added `release_docs/PLUGIN_SIGNATURE_README.md` detailing the security model, usage instructions, and best practices.

Fixes #5116


brtnfld:
@jhendersonHDF @glennsong09, this is ready for review when you get the chance.

src/H5Z.c
```diff
@@ -1323,8 +1323,34 @@ H5Z_find(bool try, H5Z_filter_t id, H5Z_class2_t **cls)
...
+        if (!try) {
+            /* Filter not in registered table; attempt dynamic loading so that any
+             * real load failure (e.g. signature verification) appears in the error
+             * stack rather than being silently replaced by "filter not registered". */
+            H5PL_key_t          key;
+            const H5Z_class2_t *filter_info;
+
+            key.id = (int)id;
+            if (NULL != (filter_info = (const H5Z_class2_t *)H5PL_load(H5PL_TYPE_FILTER, &key))) {
+                if (H5Z_register(filter_info) < 0)
```
jhendersonHDF: I don't think that trying to load and register a filter here is necessarily the right approach, as it should have already been loaded/registered at this point and is not really what this function is for. In the best case, I think this is going to hide design errors when it comes to where filters should have been dynamically loaded at. In the general case, I think this is just going to add an additional failure to dynamically load a plugin in cases where a failure to load one already occurred.

tools/test/h5copy/CMakeTests.cmake
```diff
@@ -392,6 +392,8 @@ macro (ADD_H5_UD_TEST testname resultcode infile sparam srcname dparam dstname c
             -D "TEST_ENV_VAR=HDF5_PLUGIN_PATH"
             -D "TEST_ENV_VALUE=${ud_search_path}${CMAKE_SEP}${vol_plugin_path}"
             -D "TEST_LIBRARY_DIRECTORY=${CMAKE_TEST_OUTPUT_DIRECTORY}"
+            $<$<BOOL:${HDF5_REQUIRE_SIGNED_PLUGINS}>:-D>
```
jhendersonHDF: A pretty minor nitpick but you should be able to specify these everywhere in a single expression as just `$<$<BOOL:${HDF5_REQUIRE_SIGNED_PLUGINS}>:-DTEST_KEYSTORE_DIR=${CMAKE_BINARY_DIR}/test_keystore>`

tools/test/h5sign/h5signgentest.h
jhendersonHDF, line 1: This file can probably just be deleted since there's nothing in it.

docs/PLUGIN_SIGNATURE_README.md
jhendersonHDF, line 1: A lot of the information in this file seems like stuff we shouldn't be trying to instruct users on. For example, anybody who has an air-gapped environment requirement absolutely should not need or take advice from this document.

I also am concerned about whether any of the steps/processes mentioned in this document have been tested at this point?

release_docs/PLUGIN_SIGNATURE_README.md
```diff
@@ -0,0 +1,1542 @@
...
+1. Check the [Troubleshooting](#troubleshooting) section above
+2. Review HDF5 logs with debug output enabled
+3. Contact your plugin developer for plugin-specific issues
+4. Report HDF5 library issues: <https://github.com/HDFGroup/hdf5/issues>
+
+---
+
+**Document Version**: 1.0
+**Last Updated**: 2026-01-28
+**HDF5 Version**: 1.15.0+
```
jhendersonHDF: Out of date / incorrect

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+/*-------------------------------------------------------------------------
+ * Function:    H5PL_sig_encode_footer
+ *
+ * Purpose:     Encode a signature footer struct into a 12-byte little-endian
+ *              buffer suitable for appending to a signed plugin file.
+ *
+ * Note:        Requires H5encode.h for UINT32ENCODE / UINT16ENCODE.
+ *-------------------------------------------------------------------------
+ */
+static H5_ATTR_UNUSED void
```
jhendersonHDF: Was an unused attribute really intended here or was inline intended? The function only seems to be used in the h5sign tool, so it should probably go there instead.

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+H5PL_sig_encode_footer(uint8_t buf[H5PL_SIG_FOOTER_SIZE], const H5PL_sig_footer_t *footer)
```
jhendersonHDF: A buffer size parameter should still be added here and used to check for overflow in each step

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+/*-------------------------------------------------------------------------
+ * Function:    H5PL_sig_decode_footer
+ *
+ * Purpose:     Decode a 12-byte little-endian buffer into a footer struct.
+ *
+ * Note:        Requires H5encode.h for UINT32DECODE / UINT16DECODE.
+ *-------------------------------------------------------------------------
+ */
+static H5_ATTR_UNUSED void
+H5PL_sig_decode_footer(const uint8_t buf[H5PL_SIG_FOOTER_SIZE], H5PL_sig_footer_t *footer)
```
jhendersonHDF: The same comments for `H5PL_sig_encode_footer` apply here too.

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+static H5_ATTR_UNUSED void
+H5PL_sig_encode_footer(uint8_t buf[H5PL_SIG_FOOTER_SIZE], const H5PL_sig_footer_t *footer)
+{
+    uint8_t *p = buf;
+
+    UINT32ENCODE(p, footer->signature_length);
+    *p++ = footer->algorithm_id;
+    *p++ = footer->format_version;
+    UINT16ENCODE(p, footer->reserved);
+    UINT32ENCODE(p, footer->magic);
```
jhendersonHDF: If a magic value is going to be used, it should probably be encoded first so that it can be decoded first and verified before further decoding happens.

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * Note:        Requires H5encode.h for UINT32DECODE / UINT16DECODE.
+ *-------------------------------------------------------------------------
+ */
+static H5_ATTR_UNUSED void
+H5PL_sig_decode_footer(const uint8_t buf[H5PL_SIG_FOOTER_SIZE], H5PL_sig_footer_t *footer)
+{
+    const uint8_t *p = buf;
+
+    UINT32DECODE(p, footer->signature_length);
```
jhendersonHDF: Related to the above, it seems like there should be at least minimal verification performed here

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+    footer->algorithm_id   = *p++;
+    footer->format_version = *p++;
+    UINT16DECODE(p, footer->reserved);
+    UINT32DECODE(p, footer->magic);
+} /* end H5PL_sig_decode_footer() */
+
+#ifdef H5_REQUIRE_DIGITAL_SIGNATURE
```
jhendersonHDF: This section can just be deleted it looks like

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+#define H5PL_SIG_ALGO_SHA512     0x03 /* SHA-512 with RSA-PKCS1 (default) */
+#define H5PL_SIG_ALGO_SHA256_PSS 0x11 /* SHA-256 with RSA-PSS */
+#define H5PL_SIG_ALGO_SHA384_PSS 0x12 /* SHA-384 with RSA-PSS */
+#define H5PL_SIG_ALGO_SHA512_PSS 0x13 /* SHA-512 with RSA-PSS */
+#define H5PL_SIG_ALGO_SHA3_256   0x20 /* SHA3-256 (future) */
+#define H5PL_SIG_ALGO_BLAKE3     0x30 /* BLAKE3 (future) */
+
+/* Byte offset of the magic field within the footer buffer.
+ * Layout: [sig_len: 4][algo_id: 1][format_ver: 1][reserved: 2][magic: 4] */
+#define H5PL_SIG_FOOTER_MAGIC_OFFSET (4U + 1U + 1U + 2U)
```
jhendersonHDF: If the magic value comes first this won't be needed

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+    /* Check if directory exists and get permissions */
+    if (HDstat(dir_path, &st) < 0)
+        HGOTO_ERROR(H5E_PLUGIN, H5E_CANTGET, FAIL, "cannot stat keystore directory: %s", dir_path);
+
+    /* Verify it's a directory */
+    if (!S_ISDIR(st.st_mode))
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL, "keystore path is not a directory: %s", dir_path);
+
+    /* Reject world-writable directories */
+    if (st.st_mode & S_IWOTH)
```
jhendersonHDF: Maybe there's value in this check, but it seems a like a whole lot of complication, especially on the Windows side of things, to verify something that I don't necessarily believe we should be concerning ourselves with. What exactly is meant by unprivileged users adding malicious keys here and why should it be a problem for HDF5 to deal with?

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+
+    FUNC_ENTER_PACKAGE
+
+    assert(file_path);
+
+    /* Try to load key */
+    if (NULL != (key = H5PL__create_public_RSA_from_file(file_path))) {
+        /* Add to keystore */
+        if (H5PL__add_key_to_keystore(key, file_path) < 0) {
+            EVP_PKEY_free(key);
```
jhendersonHDF: Since this function only makes two calls, anywhere `H5PL__process_key_file` is called could probably just be replaced with these two calls. Otherwise, `EVP_PKEY_free()` should be moved to the `done:` section so future expansions don't accidentally skip cleanup of the key on failure.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+    char   line[256];
+    size_t path_len;
+    herr_t ret_value = SUCCEED;
+
+    FUNC_ENTER_PACKAGE
+
+    assert(keystore_dir);
+
+    /* Build path to revoked signatures file */
+    path_len = strlen(keystore_dir) + strlen("/revoked_signatures.txt") + 1;
```
jhendersonHDF: The assumption about the name of this file for revoked signatures doesn't seem to be documented anywhere. Revoked signatures also don't appear to be tested anywhere.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+
+    /* Try to open revoked signatures file (optional - not an error if missing) */
+    if (NULL == (fp = fopen(filepath, "r"))) {
+        /* File doesn't exist - not an error, just means no revoked signatures */
+        HGOTO_DONE(SUCCEED);
+    }
+
+    /* Read file line by line */
+    while (fgets(line, sizeof(line), fp) != NULL) {
+        unsigned char hash[H5PL_SIGNATURE_HASH_SIZE];
```
jhendersonHDF, line 806: Are revoked signatures always sha-256 or do they match the encoding algorithm, in which case `H5PL_SIGNATURE_HASH_SIZE` would be the wrong size for sha-512, for example.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+            trimmed[line_len - 1] = '\0';
+            line_len--;
+        }
+
+        /* Skip empty lines and comments */
+        if (line_len == 0 || trimmed[0] == '#')
+            continue;
+
+        /* Parse hex string (must be exactly 64 hex characters for SHA-256) */
+        if (line_len != H5PL_SIGNATURE_HASH_SIZE * 2) {
```
jhendersonHDF: Shouldn't this just be `H5PL_SIGNATURE_HASH_SIZE` (not * 2) for sha-256?

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+
+    if (1 != EVP_DigestFinal_ex(mdctx, hash, NULL))
+        HGOTO_DONE(false);
+
+    /* Check if hash is in revoked list using binary search
+     * (array is sorted in H5PL__load_revoked_signatures)
+     */
+    if (H5PL_revoked_sigs_count_g > 0) {
+        H5PL_revoked_signature_t key;
+        memcpy(key.hash, hash, H5PL_SIGNATURE_HASH_SIZE);
```
jhendersonHDF: It looks like the hash could be compared directly here without the need for a `memcpy`

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+/*-------------------------------------------------------------------------
+ * Function:    H5PL__check_signature_cache
+ *
+ * Purpose:     Check signature verification cache
+ *
+ * Return:      SUCCEED/FAIL
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5PL__check_signature_cache(const char *plugin_path, bool *cached_result)
```
jhendersonHDF: We're already caching loaded plugins and they should stay cached until the H5PL package is terminated, so we should generally already be verifying a signature only a single time. Adding another layer of caching that could potentially return incorrect results seems like additional design complexity for little benefit.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+    /* Read footer from end of file */
+    if (H5PL__read_file_data(fd, file_size - (HDoff_t)H5PL_SIG_FOOTER_SIZE, footer_buf, H5PL_SIG_FOOTER_SIZE,
+                             plugin_path) < 0)
+        HGOTO_ERROR(H5E_PLUGIN, H5E_READERROR, FAIL, "cannot read signature footer");
+
+    /* Decode footer (little-endian to native byte order) */
+    H5PL_sig_decode_footer(footer_buf, footer_out);
+
+    /* Validate magic number */
+    if (footer_out->magic != H5PL_SIG_MAGIC)
```
jhendersonHDF: All of this validation logic should be moved into `H5PL_sig_decode_footer`. The verification should happen as decoding occurs and not after.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL, "unsupported signature format version %u (expected %u)",
+                    (unsigned)footer_out->format_version, (unsigned)H5PL_SIG_FORMAT_VERSION_CURRENT);
+
+    /* Validate algorithm ID */
+    if (NULL == H5PL__get_hash_algorithm(footer_out->algorithm_id)) {
+        if (footer_out->algorithm_id == H5PL_SIG_ALGO_SHA3_256 ||
+            footer_out->algorithm_id == H5PL_SIG_ALGO_BLAKE3)
+            HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
+                        "algorithm ID 0x%02X is reserved for a future HDF5 release - "
+                        "upgrade HDF5 to verify this plugin",
```
jhendersonHDF: This message seems like designing ahead for something that may never happen

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+
+    /* Validate magic number */
+    if (footer_out->magic != H5PL_SIG_MAGIC)
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
+                    "invalid signature magic number (expected 0x%08X, got 0x%08X) - "
+                    "not a signed HDF5 plugin or corrupted",
+                    (unsigned)H5PL_SIG_MAGIC, (unsigned)footer_out->magic);
+
+    /* Validate format version */
+    if (footer_out->format_version != H5PL_SIG_FORMAT_VERSION_CURRENT)
```
jhendersonHDF: Are plugins signed with older versions of the format not going to be supported by newer versions of the library?

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+
+        /* Practical size limit: 1GB for plugin files */
+        if (binary_size_off > H5PL_MAX_PLUGIN_SIZE)
+            HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
+                        "plugin binary size %llu exceeds maximum allowed size (%llu bytes) - "
+                        "file too large to verify",
+                        (unsigned long long)binary_size_off, (unsigned long long)H5PL_MAX_PLUGIN_SIZE);
+
+        /* Check for overflow when casting to size_t */
+        if (binary_size_off < 0 || (uint64_t)binary_size_off > (uint64_t)SIZE_MAX)
```
jhendersonHDF: This check seems pretty much redundant since the comment above states `binary_size_off` is non-negative and we've also already checked it against `H5PL_MAX_PLUGIN_SIZE`.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+            diagnostic = "\n"
+                         "  DIAGNOSIS: Unknown verification failure\n"
+                         "  - Enable debug output with: export HDF5_DEBUG=PL\n";
+        }
+
+        keystore_path = getenv("HDF5_PLUGIN_KEYSTORE");
+        if (keystore_path == NULL)
+            keystore_path = H5PL_SIG_KEYSTORE_DIR_STR;
+
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
```
jhendersonHDF, line 1176: Rather than using the error stack as a diagnosis mechanism like this PR does in several places, I think this is something that needs to be returned as error codes and documented. Dumping a huge amount of output to a terminal may not be all that useful and we're just guessing at what the problem could be rather than returning something from SSL, for example. In the parallel case the amount of output would be drastic.

src/H5PLsig.c
```diff
@@ -0,0 +1,1831 @@
...
+
+    /* Initialize keystore on first use */
+    if (!H5PL_keystore_initialized_g) {
+        if (H5PL__init_keystore() < 0)
+            HGOTO_ERROR(H5E_PLUGIN, H5E_CANTINIT, FAIL, "cannot initialize keystore");
+    }
+
+    /* Check if signature is revoked */
+    if (H5PL__is_signature_revoked(signature, footer.signature_length))
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
```
jhendersonHDF: There are more places that I won't call out, but this one seems like a good example where we could return an error code that a plugin wasn't loaded specifically because of a revoked signature and then document everything below for that code.

src/H5PLsig.h
```diff
@@ -0,0 +1,137 @@
...
+/* Magic number to identify HDF5 signed plugins */
+#define H5PL_SIG_MAGIC 0x48444635 /* "HDF5" in hex */
+
+/* Current signature format version */
+#define H5PL_SIG_FORMAT_VERSION_CURRENT 1
+
+/* Hash Algorithm Identifiers */
+#define H5PL_SIG_ALGO_SHA256     0x01 /* SHA-256 with RSA-PKCS1 */
+#define H5PL_SIG_ALGO_SHA384     0x02 /* SHA-384 with RSA-PKCS1 */
+#define H5PL_SIG_ALGO_SHA512     0x03 /* SHA-512 with RSA-PKCS1 (default) */
```
jhendersonHDF: It seems a bit strange here to use overlapping bit patterns for distinct algorithms. An enum seems like it would be better than the current `uint8_t` with overlapping bits.

CMakeLists.txt
```diff
@@ -781,6 +781,130 @@ if (HDF5_ENABLE_HDFS)
...
+      "      CentOS 7 users should install openssl11 package")
+  endif ()
+
+  # Informational message for OpenSSL 3.0+ (APIs are compatible, not deprecated)
+  if (OPENSSL_VERSION VERSION_GREATER_EQUAL "3.0.0")
+    message(STATUS "OpenSSL 3.0+ detected - all EVP_* APIs are compatible (not deprecated)")
+  endif ()
+
+  # KeyStore directory must be configured
+  if (NOT HDF5_PLUGIN_KEYSTORE_DIR)
```
jhendersonHDF: If the keystore can be set by an environment variable, why require one to be set here unless `HDF5_LOCK_PLUGIN_KEYSTORE` is `ON`?

tools/test/h5sign/h5signverifytest.c
```diff
@@ -0,0 +1,441 @@
...
+    if (time2 <= time1) {
+        printf("PASSED\n");
+        printf("  First verification: %.1f seconds\n", time1);
+        printf("  Second verification (cached): %.1f seconds\n", time2);
+        tests_passed++;
+        return 0;
+    }
+    else {
+        printf("PASSED (with note)\n");
+        printf("  NOTE: Cache hit not faster (timing may be too coarse)\n");
```
jhendersonHDF: If the test is going to pass regardless of the timing, I believe this message will never be seen or acted on (I do still think the caching layer is unnecessary though).

tools/test/h5sign/h5signverifytest.c
```diff
@@ -0,0 +1,441 @@
...
+    if (ret2 != SUCCEED) {
+        printf("FAILED\n");
+        printf("  Cached verification failed\n");
+        tests_failed++;
+        remove(temp_plugin);
+        return 1;
+    }
+
+    /* Modify the file (touch to update mtime) */
+    HDsleep(2); /* Ensure mtime changes */
```
jhendersonHDF: Is this sleep really needed here? Especially since the file isn't modified until after this point.

tools/src/h5sign/h5sign.c
```diff
@@ -0,0 +1,895 @@
...
+ *
+ *           File format after signing:
+ *             [ Plugin Binary ] [ RSA Signature ] [ Footer ]
+ *
+ *           Footer (12 bytes, little-endian):
+ *             - Signature length  (4 bytes)
+ *             - Algorithm ID      (1 byte)
+ *             - Format version    (1 byte)
+ *             - Reserved          (2 bytes)
+ *             - Magic number 0x48444635 "HDF5" (4 bytes)
```
jhendersonHDF: If the magic number is placed at the front of the footer, make sure to update this (and probably a few other places).

tools/src/h5sign/h5sign.c
```diff
@@ -0,0 +1,895 @@
...
+         * to at check time (which could differ if a symlink is swapped between stat and fopen). */
+        {
+            h5_stat_t key_stat;
+            if (HDfstat(fileno(fp), &key_stat) == 0) {
+                if (!S_ISREG(key_stat.st_mode)) {
+                    fprintf(rawerrorstream, "Error: '%s' is not a regular file\n", keyfile);
+                    fclose(fp);
+                    goto done;
+                }
+                if (key_stat.st_mode & (S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH)) {
```
jhendersonHDF: I wouldn't necessarily consider group r/w insecure permissions, which is why I don't think we should be concerning ourselves with a system's setup.

tools/src/h5sign/h5sign.c
```diff
@@ -0,0 +1,895 @@
...
+    file_size = (hsize_t)st.st_size;
+
+    /* Sanity check file size */
+    if (file_size == 0) {
+        fprintf(rawerrorstream, "Error: Plugin file '%s' is empty\n", plugin_path);
+        ret_value = FAIL;
+        goto done;
+    }
+
+    if (file_size > MAX_PLUGIN_SIZE) {
```
jhendersonHDF: This should use the `H5PL_MAX_PLUGIN_SIZE` macro directly (which should probably be moved into H5PLsig.h) so that it's kept in sync with the main library code. Related, though not that it's likely to happen, this seems like it should check if `file_size` + `bytes_added` > `H5PL_MAX_PLUGIN_SIZE` so that a load error doesn't occur later. If `bytes_added` isn't known at this point, the file size should be checked after being signed.

tools/src/h5sign/h5sign.c
```diff
@@ -0,0 +1,895 @@
...
+        fprintf(rawoutstream, "Signature created successfully\n");
+        fprintf(rawoutstream, "Signature length: %zu bytes\n", sig_len);
+    }
+
+    /* Truncate the file to exactly file_size bytes before appending.
+     * If any concurrent writer extended the file during hashing, SEEK_END would
+     * place us past the extra data, producing a signature that covers fewer bytes
+     * than the binary.  Truncating to the size we hashed guarantees that the
+     * signature is appended at precisely the byte we expect. */
+    if (HDftruncate(fd, (HDoff_t)file_size) < 0) {
```
jhendersonHDF: This seems like solving for a problem that won't exist. Also, if there are concurrent writers then it seems like the file could also be touched immediately after being truncated.

test/test_plugin_signature.c
```diff
@@ -0,0 +1,1443 @@
...
+     */
+    {
+        unsigned char  footer_bytes[H5PL_SIG_FOOTER_SIZE];
+        unsigned char *p = footer_bytes;
+
+        UINT32ENCODE(p, footer.signature_length);
+        *p++ = footer.algorithm_id;
+        *p++ = footer.format_version;
+        UINT16ENCODE(p, footer.reserved);
+        UINT32ENCODE(p, footer.magic);
```
jhendersonHDF: This should probably use functions from H5PL so that if/when the signature format changes (such as placing the magic number first) this doesn't need updating as well.

test/test_plugin_signature.c
```diff
@@ -0,0 +1,1443 @@
...
+    }
+
+    /* Write footer with wrong magic number
+     * On-disk layout (12 bytes): [sig_len:4][algo_id:1][format_ver:1][reserved:2][magic:4]
+     */
+    UINT32ENCODE(p, (uint32_t)256); /* Signature length */
+    *p++ = H5PL_SIG_ALGO_SHA256;    /* Algorithm ID */
+    *p++ = 1;                       /* Format version */
+    UINT16ENCODE(p, (uint16_t)0);   /* Reserved */
+    UINT32ENCODE(p, 0xDEADBEEF);    /* Wrong magic */
```
jhendersonHDF: The same here as well.

CMakeLists.txt
```diff
@@ -781,6 +781,130 @@ if (HDF5_ENABLE_HDFS)
...
+        message(VERBOSE "KeyStore directory: ${HDF5_PLUGIN_KEYSTORE_DIR} (${KEYSTORE_KEY_COUNT} key(s) found)")
+      endif ()
+    endif ()
+
+    # Define KeyStore directory as compile definition
+    add_compile_definitions(H5PL_KEYSTORE_DIR="${HDF5_PLUGIN_KEYSTORE_DIR}")
+  endif ()
+
+  # Enable digital signature verification
+  add_compile_definitions(H5_REQUIRE_DIGITAL_SIGNATURE)
```
jhendersonHDF: This should probably be added to H5pubconf.h instead of directly added as a compile definition

CMakeLists.txt
```diff
@@ -781,6 +781,130 @@ if (HDF5_ENABLE_HDFS)
...
+        message(WARNING
+          "KeyStore directory exists but contains no .pem files: ${HDF5_PLUGIN_KEYSTORE_DIR}\n"
+          "Add public key files before using HDF5.")
+      else ()
+        message(VERBOSE "KeyStore directory: ${HDF5_PLUGIN_KEYSTORE_DIR} (${KEYSTORE_KEY_COUNT} key(s) found)")
+      endif ()
+    endif ()
+
+    # Define KeyStore directory as compile definition
+    add_compile_definitions(H5PL_KEYSTORE_DIR="${HDF5_PLUGIN_KEYSTORE_DIR}")
```
jhendersonHDF, line 829: I'm not sure if there are any implications to this, but storing the keystore directory in the library's strings seems problematic or at least revealing to me. Of course, without that (and the environment variable) the keystore logic would need a different way of determining where to load keys from.

.github/workflows/signed-plugins.yml
```diff
@@ -0,0 +1,141 @@
+name: Test Signed Plugins
+
+on:
+  push:
+    branches: [ develop, feature/dig_sig_ver, feature/* ]
```
jhendersonHDF: At least `feature/dig_sig_ver` should be removed pre- or post-merge

src/H5PLsig.h
brtnfld: Encode is called in 3 places (h5sign, 2x in test_plugin_signature). Decode is called in 3 places (H5PLsig.c, 2x in h5sign). Both functions are genuinely shared. H5_ATTR_UNUSED is used the same way in H5VMprivate.h on 10+ static header functions.

src/H5PLsig.h
brtnfld: Buffer is always H5PL_SIG_FOOTER_SIZE (compile-time constant 12). Every call site passes buf[H5PL_SIG_FOOTER_SIZE], so I don't think a runtime size check on a fixed-size buffer adds any safety.

src/H5PLsig.h
brtnfld: same earlier comment applies

src/H5PLsig.h
brtnfld: Values like 0x01 and 0x11 share bits, but they're never used as bitmasks — always discrete values in switch/case and == comparisons. The IS_PSS macro uses range comparison (>=/<=), not bitwise ops. Is not using #define  consistent with the standard pattern for on-disk format constants? An enum wouldn't change the on-disk representation and could complicate the on-disk format.   These must remain fixed integer values regardless of whether they're defined as #define or enum

src/H5PLsig.c
brtnfld: A SHA-256 hash is 32 bytes = 64 hex characters. H5PL_SIGNATURE_HASH_SIZE * 2 = 32 * 2 = 64. Each byte is represented by 2 hex chars, so * 2 is right.

src/H5PLsig.c
brtnfld: The revocation list hashes the signature bytes with SHA-256 (line 965: EVP_DigestInit_ex(mdctx, EVP_sha256(), NULL)). This is independent of the plugin's signing algorithm. Whether the plugin was signed with SHA-256 or SHA-512, the revocation system always hashes the signature blob with SHA-256 to produce a 32-byte hash for lookup. So H5PL_SIGNATURE_HASH_SIZE = 32 is correct. However, the comment on line 94 should be clearer that this is the hash of the signature bytes for the revocation list, not the signing algorithm hash.

src/H5PLsig.c
jhendersonHDF: But line_len comes from strlen which is going to be in terms of bytes, not characters, and two hex characters represent one byte.

src/H5PLsig.h
jhendersonHDF: That just seems like a misuse of the unused attribute to me. Static inline alone should be fine and is a somewhat common pattern.

src/H5PLsig.h
jhendersonHDF: My point is having the array size in the parameter guarantees very little, checks very little and can be easily "defeated". Rather than trying to predict how the code may or may not change in the future, always passing the size of the buffer just seems like a much better option.

src/H5PLsig.h
jhendersonHDF: Both #defines and enums are used during encoding the on-disk format in different places. There's not much difference but I prefer enums just because they have more typing to them than #defines do, which is nice for passing around as parameters and can sometimes catch accidental bad values. I think it also helps a bit in removing the potential ambiguity between 0x3 meaning sha-512 or an accidental bitwise OR of sha-256 and sha-384 when passing around `uint8_t`. It's unlikely but I'm considering how the code might change in the future and an accidental bitwise OR of two `uint8_t` values doesn't stick out as much as ORing enum values. `uint8_t` passed around as a parameter usually reads as "flags" to me at a first glance.

release_docs/PLUGIN_SIGNATURE_README.md
```diff
@@ -0,0 +1,1164 @@
...
+
+# Verbose output (recommended for first-time)
+h5sign -p my_filter_plugin.so -k my_private_key.pem -v
+```
+
+**What happens**:
+
+1. h5sign computes a SHA-512 hash of your plugin binary
+2. Signs the hash with your private key using RSA
+3. Appends the signature and metadata to the end of the plugin file
```
glennsong09: In regards to what Jordan mentioned, I think it's possible we shouldn't give this information out.

brtnfld: Which information are you referring to? I removed a lot of the airgap material.

brtnfld:
## Category 1: Footer Format — Magic First (5 comments)
- Reordered footer layout to `[magic][sig_len][algo_id][format_ver][reserved]`
- Removed `H5PL_SIG_FOOTER_MAGIC_OFFSET` macro (no longer needed)
- Updated encode/decode functions, tests, h5sign, and shell scripts

## Category 2: Footer Encode/Decode Improvements (4 comments)
- Changed `H5_ATTR_UNUSED` → `static inline` on both functions
- Added `buf_size` parameter with `assert` to both encode/decode
- Converted algorithm ID `#define`s to `H5PL_sig_algo_t` enum
- Updated all call sites and function signatures across 4 files

## Category 3: Dead Code Removal (4 comments)
- Deleted empty `h5signgentest.h`
- Reverted `H5Z_find` plugin load/register fallback in `H5Z.c`
- Removed unnecessary `ftruncate` in h5sign
- Removed dead `#ifdef H5_REQUIRE_DIGITAL_SIGNATURE` section

## Category 4: Security Check Concerns (3 comments)
- Removed world-writable directory check (`S_IWOTH`) and Windows ACL checks
- Removed private key permission check from h5sign
- Removed ~370 lines of air-gapped environment content from README
- Added "Keystore Security Requirements" section documenting expectations

## Category 5: Error Handling & Diagnostics (3 comments)
- Replaced verbose diagnostic dumps with concise single-line error messages
- Simplified revoked signature error
- Removed "upgrade HDF5" message for reserved algorithm IDs

## Category 6: Revocation Logic (3 comments + 1 related)
- Defined `H5PL_REVOKED_SIGS_FILENAME` macro and documented it
- Added comment clarifying SHA-256 is always used for revocation hashing
- Confirmed `* 2` is correct (bytes → hex chars)
- Removed `memcpy` in hash lookup (compare directly via bsearch)

## Category 7: Caching (2 comments)
- Removed entire signature verification cache (~180 lines)
- Removed cache timing test and `HDsleep(2)` from test suite

## Category 8: Build/CMake (4 comments)
- Moved `H5_REQUIRE_DIGITAL_SIGNATURE` to `H5pubconf.h.in`
- Made `HDF5_PLUGIN_KEYSTORE_DIR` optional (env var fallback)
- Simplified CMake generator expressions in 5 test CMake files

## Category 9: Miscellaneous (5 comments)
- Moved `EVP_PKEY_free` to `done:` section in `H5PL__process_key_file`
- Removed redundant `SIZE_MAX` overflow check
- Moved `H5PL_MAX_PLUGIN_SIZE` to `H5PLsig.h` (shared between lib and h5sign)
- Added format version compatibility comment

## Category 10: Test Code (2 comments)
- Updated `test_plugin_signature.c` to use `H5PL_sig_encode_footer` instead of manual encoding

## Category 11: Workflow/Doc Cleanup (2 comments)
- Removed `feature/dig_sig_ver` from workflow branch triggers
- Fixed version info in README

## Category 12: Design Questions (2 comments)
- Added comment about format version forward-compatibility intent

release_docs/PLUGIN_SIGNATURE_README.md
jhendersonHDF: Outside of the now-removed air-gap information, I think this comes down to gut feeling. My point is, let's not try to be security experts and give people information that may be wrong aecurity-wise for their environment. Information on how to sign a plugin is good and information on how to use that plugin with HDF5 is good. Anything beyond that which is prescriptive, such as stating where/how to store your keys and signed plugins, is something I think we need to assume users need to know on their own. Maybe a recommendations section could be fine, but I don't want to get into the details too much.

jhendersonHDF: Kind of the difference between something like "make sure to store your keys in a safe location others can't access" vs "store your keys in your home directory, set the permissions with chmod go-rw".

brtnfld: Updated

CMakeLists.txt
```diff
@@ -781,6 +781,121 @@ if (HDF5_ENABLE_HDFS)
...
+# KeyStore directory for multiple trusted public keys (recommended)
+set(HDF5_PLUGIN_KEYSTORE_DIR "" CACHE PATH
+    "Directory containing trusted public keys (.pem files) for plugin verification")
+# Security Note: For production deployments where environment variables may be
+# controllable by untrusted users (e.g., HPC clusters, multi-tenant systems),
+# you can prevent runtime override of the keystore directory by enabling
+# HDF5_LOCK_PLUGIN_KEYSTORE at compile time (requires rebuild).
+
+# Security: Disable environment variable override (prevents runtime keystore redirection)
+option(HDF5_LOCK_PLUGIN_KEYSTORE "Disable HDF5_PLUGIN_KEYSTORE environment variable override (security hardening)" OFF)
```
jhendersonHDF: `cmake_dependent_option (HDF5_LOCK_PLUGIN_KEYSTORE "Disable HDF5_PLUGIN_KEYSTORE environment variable override (security hardening)" OFF "HDF5_REQUIRE_SIGNED_PLUGINS" OFF)`

would be nice here. Due to the way dependent options work it should stay outside of the `if (HDF5_REQUIRE_SIGNED_PLUGINS)` if converted though.

CMakeLists.txt
```diff
@@ -781,6 +781,121 @@ if (HDF5_ENABLE_HDFS)
   endif ()
 endif ()
 
+#-----------------------------------------------------------------------------
+# Option to Require Digitally Signed plugins
+#-----------------------------------------------------------------------------
+option (HDF5_REQUIRE_SIGNED_PLUGINS "Require digitally signed plugins" OFF)
+
+# KeyStore directory for multiple trusted public keys (recommended)
+set(HDF5_PLUGIN_KEYSTORE_DIR "" CACHE PATH
```
jhendersonHDF: This could probably be moved inside the `if (HDF5_REQUIRE_SIGNED_PLUGINS)` for the time being

CMakeLists.txt
```diff
@@ -781,6 +781,121 @@ if (HDF5_ENABLE_HDFS)
...
+      "set HDF5_PLUGIN_KEYSTORE environment variable at runtime.")
+  endif ()
+
+  # Configure KeyStore directory if provided
+  if (HDF5_PLUGIN_KEYSTORE_DIR)
+    if (NOT EXISTS "${HDF5_PLUGIN_KEYSTORE_DIR}")
+      message(WARNING
+        "KeyStore directory does not exist: ${HDF5_PLUGIN_KEYSTORE_DIR}\n"
+        "Keys will be loaded at runtime if directory is created.")
+    elseif (NOT IS_DIRECTORY "${HDF5_PLUGIN_KEYSTORE_DIR}")
```
jhendersonHDF: Won't this conflict with the check above that seems to be allowing the directory to not exist until runtime?

CMakeLists.txt
```diff
@@ -781,6 +781,121 @@ if (HDF5_ENABLE_HDFS)
...
+  # environment variable can be used at runtime instead.  Require a compile-time
+  # directory only when the environment variable override is locked out.
+  if (HDF5_LOCK_PLUGIN_KEYSTORE AND NOT HDF5_PLUGIN_KEYSTORE_DIR)
+    message(FATAL_ERROR
+      "HDF5_LOCK_PLUGIN_KEYSTORE=ON requires a compile-time KeyStore directory:\n"
+      "  -DHDF5_PLUGIN_KEYSTORE_DIR=/etc/hdf5/trusted_keys")
+  endif ()
+
+  if (NOT HDF5_PLUGIN_KEYSTORE_DIR)
+    message(STATUS "No compile-time KeyStore directory configured; "
```
jhendersonHDF: If you want to call this out, maybe use `NOTICE` or even `WARNING` instead. Otherwise, it will likely get buried.

CMakeLists.txt
```diff
@@ -781,6 +781,121 @@ if (HDF5_ENABLE_HDFS)
...
+  endif ()
+
+  # Add OpenSSL to link libraries for the HDF5 library
+  # Only libcrypto is needed (EVP, PEM, BIO, ERR APIs); libssl (TLS) is not used
+  list(APPEND LINK_LIBS OpenSSL::Crypto)
+
+  # Add Windows security libraries for ACL checking (GetNamedSecurityInfoA, etc.)
+  # These are used in H5PLsig.c for KeyStore directory permission validation
+  if (WIN32)
+    list(APPEND LINK_LIBS Advapi32)
```
jhendersonHDF: Not needed at the moment

src/H5PLsig.c
```diff
@@ -0,0 +1,1340 @@
...
+        /* Only try if directory was configured */
+        h5_stat_t st;
+        if (HDstat(H5PL_KEYSTORE_DIR, &st) == 0) {
+            /* Directory exists, try to load */
+            if (H5PL__load_keys_from_directory(H5PL_KEYSTORE_DIR) < 0) {
+                /* Not immediately fatal - log the specific reason and fall through
+                 * to the generic "no valid public keys" error which provides guidance. */
+                H5PL_SIG_DEBUG_PRINT("WARNING: Failed to load keys from configured keystore: %s\n",
+                                     H5PL_KEYSTORE_DIR);
+                H5E_clear_stack(); /* Clear so the generic error below is the top-level error */
```
jhendersonHDF: Rather than clear the stack, it would probably be cleaner to use `H5E_pause_stack()` / `H5E_resume_stack()` for something that could potentially fail.

src/H5PLsig.h
```diff
@@ -0,0 +1,139 @@
...
+/* Signature footer on-disk size (12 bytes) */
+#define H5PL_SIG_FOOTER_SIZE 12
+
+/* True when algo id selects an RSA-PSS padding variant */
+#define H5PL_SIG_ALGO_IS_PSS(id) ((id) >= H5PL_SIG_ALGO_SHA256_PSS && (id) <= H5PL_SIG_ALGO_SHA512_PSS)
+
+/* Maximum RSA signature size in bytes.
+ * A 4096-bit RSA key produces a 512-byte signature; 1024 bytes allows
+ * headroom for 8192-bit keys.  Used by both the signer and verifier. */
+#define H5PL_MAX_SIGNATURE_SIZE 1024
```
jhendersonHDF, line 70: Is there a particular reason for limiting the signature size? Not that I necessarily think a larger size will be needed in most cases, but it seems somewhat arbitrary.

brtnfld: RSA signatures are always equal to the key size in bytes (4096-bit key -> 512 bytes), so 1024 covers up to 8192-bit keys, which is well beyond current practice. The actual allocation uses the sig_len value from the footer, so smaller keys work fine. Happy to increase it if you think a larger ceiling is warranted, or we could read the key size from OpenSSL and use that as the bound instead.

src/H5PLsig.h
```diff
@@ -0,0 +1,154 @@
...
+ * Note:        Requires H5encode.h for UINT32DECODE / UINT16DECODE.
+ *              buf_size must be >= H5PL_SIG_FOOTER_SIZE (12).
+ *-------------------------------------------------------------------------
+ */
+static inline bool
+H5PL_sig_decode_footer(const uint8_t *buf, size_t buf_size, H5PL_sig_footer_t *footer)
+{
+    const uint8_t *p = buf;
+
+    assert(buf_size >= H5PL_SIG_FOOTER_SIZE);
```
jhendersonHDF: Using only an `assert()` for encode is maybe fine since we're more in control, but this one for sure should be checked with a regular error check.

src/H5PLsig.h
```diff
@@ -0,0 +1,154 @@
...
+    if (footer->magic != H5PL_SIG_MAGIC)
+        return false;
+
+    UINT32DECODE(p, footer->signature_length);      /* bytes 4-7  */
+    footer->algorithm_id   = (H5PL_sig_algo_t)*p++; /* byte  8    */
+    footer->format_version = *p++;                  /* byte  9    */
+    UINT16DECODE(p, footer->reserved);              /* bytes 10-11 */
+
+    /* Verify format version */
+    if (footer->format_version != H5PL_SIG_FORMAT_VERSION_CURRENT)
```
jhendersonHDF: Moving this comment as the decoding logic was moved: Are plugins signed with older versions of the format not going to be supported by newer versions of the library?

tools/src/h5sign/h5sign.c
```diff
@@ -0,0 +1,851 @@
...
+    file_size = (hsize_t)st.st_size;
+
+    /* Sanity check file size */
+    if (file_size == 0) {
+        fprintf(rawerrorstream, "Error: Plugin file '%s' is empty\n", plugin_path);
+        ret_value = FAIL;
+        goto done;
+    }
+
+    if (file_size > H5PL_MAX_PLUGIN_SIZE) {
```
jhendersonHDF: Moving part of the old comment as half is outdated now: related, though not that it's likely to happen, this seems like it should check if `file_size` + `bytes_added` > `H5PL_MAX_PLUGIN_SIZE` so that a load error doesn't occur later. If `bytes_added` isn't known at this point, the file size should be checked after being signed.

CMakeLists.txt
brtnfld: keystore only contains public keys, and is only there if -DHDF5_PLUGIN_KEYSTOR_DIR is present

docs/PLUGIN_SIGNATURE_README.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Re-signing a Plugin
+
+To update the signature (e.g., after rebuilding or to change the algorithm):
+
+```bash
+h5sign -p my_plugin.so -k my_private_key.pem -f
+```
+
+The `-f` / `--force` flag strips the existing signature before re-signing.
```
fortnern, line 120: Seems a strange name for this flag

docs/PLUGIN_SIGNATURE_README.md
```diff
@@ -0,0 +1,277 @@
...
+
+- **Unsigned malicious plugins**: Blocked (signature required)
+- **Tampered plugins**: Detected (signature invalidates)
+- **Untrusted sources**: Rejected (keystore verification)
+
+### Signature Revocation
+
+Individual signatures can be revoked without removing the entire public key.
+Place a file named `revoked_signatures.txt` in the keystore directory. Each
+line is the 64-character hex-encoded SHA-256 hash of the raw signature bytes
```
fortnern, line 159: Is this hash the same as the signature that's appended to the plugin (except hex-coded)?

docs/PLUGIN_SIGNATURE_README.md
brtnfld: mimics the -f in cp mv rm commands, in that it forces the signature to be overwritten, do you an alternative suggestion?

docs/PLUGIN_SIGNATURE_README.md
brtnfld: No.
I added a clarification that the revocation entry is the SHA-256 digest of the signature bytes, not the signature itself hex-encoded.

src/H5PLint.c
```diff
@@ -340,6 +346,12 @@ H5PL__open(const char *path, H5PL_type_t type, const H5PL_key_t *key, bool *succ
     if (plugin_type)
         *plugin_type = H5PL_TYPE_ERROR;
 
+#ifdef H5_REQUIRE_DIGITAL_SIGNATURE
```
fortnern, line 349: So there's no run-time switch for signature verification? Is this a deliberate decision for security reasons? I can see not wanting to allow disabling it if the switch is compiled in, but it seems to me we should allow the opposite.

brtnfld: Yes, this is a deliberate security design decision. The compile-time gate ensures that the entity building/deploying the library controls the security policy, not individual applications at runtime. A runtime "enable" switch would mean that every application must remember to opt in, and a missed call would mean that unsigned plugins load silently. This follows the same pattern as Linux CONFIG_MODULE_SIG_FORCE and FIPS-mode compilation: the security invariant is set at build time and cannot be weakened (or forgotten) at runtime.

src/H5PLsig.c
```diff
@@ -0,0 +1,1311 @@
...
+                    /* Non-fatal - continue even if revoked signatures fail to load */
+                }
+            }
+            H5E_resume_stack();
+        }
+    }
+#endif
+
+    /* Must have at least one key */
+    if (!keys_loaded || H5PL_keystore_count_g == 0) {
```
fortnern: Does this need to be an error? It seems to be there could be cases where the sysadmin requires HDF5 to be compiled with plugin verification required, but the user doesn't need to load any plugins.

src/H5PLsig.c
```diff
@@ -0,0 +1,1311 @@
...
+        const char *attempted_source    = env_keystore ? env_keystore : H5PL_SIG_KEYSTORE_DIR_STR;
+        bool        keystore_configured = (env_keystore != NULL);
+
+#ifdef H5PL_KEYSTORE_DIR
+        keystore_configured = true;
+#endif
+
+        HGOTO_ERROR(H5E_PLUGIN, H5E_BADVALUE, FAIL,
```
fortnern: Seems this message could be misleading - if there is no H5PL_KEYSTORE_DIR, the user set the env var, but there are no keys in the env dir, it will say "no keystore configured" which is wrong.

src/H5PLsig.h
```diff
@@ -0,0 +1,157 @@
...
+    uint8_t *p = buf;
+
+    assert(buf_size >= H5PL_SIG_FOOTER_SIZE);
+    (void)buf_size; /* used only by assert */
+
+    UINT32ENCODE(p, footer->magic);            /* bytes 0-3  */
+    UINT32ENCODE(p, footer->signature_length); /* bytes 4-7  */
+    *p++ = (uint8_t)footer->algorithm_id;      /* byte  8    */
+    *p++ = footer->format_version;             /* byte  9    */
+    UINT16ENCODE(p, footer->reserved);         /* bytes 10-11 */
```
fortnern: What is the purpose of the reserved bytes? I don't think it will be a problem to change the size of the footer in the future if necessary, as long as we put the version field last. And it's not like adding the reserved bytes will allow this version of the library to understand later versions of the footer.

fortnern: I suppose it makes some of the logic for calculating the binary size etc. a bit more complicated, but I still don't think it's necessary, unless we have something specific in mind for those two bytes. If we don't, and we need to add something, there's a good chance it'll need more than two bytes and we'll need the variable footer size logic anyways.

tools/src/h5sign/h5sign.c
```diff
@@ -0,0 +1,855 @@
...
+                            (hsize_t)existing_sig_len + (hsize_t)H5PL_SIG_FOOTER_SIZE > file_size) {
+                            fprintf(rawerrorstream, "Error: Corrupt signature footer in '%s' (sig_len=%u)\n",
+                                    plugin_path, existing_sig_len);
+                            ret_value = FAIL;
+                            goto done;
+                        }
+
+                        binary_size = file_size - (hsize_t)existing_sig_len - (hsize_t)H5PL_SIG_FOOTER_SIZE;
+
+                        if (HDftruncate(fd, (HDoff_t)binary_size) < 0) {
```
fortnern, line 534: Why is it necessary to truncate the file here? Can't we just only read the first binary_size bytes then overwrite the signature?

docs/PLUGIN_SIGNATURE_README.md
fortnern: Do we know how similar signing utilities work? I'm also wondering if we even need this protection - I don't think anything bad would happen if we just signed the whole thing including the old signature. But maybe users want to be alerted if they're signing something that's already signed. Again, do we know how other signing utilities handle this? There's also the 1 in 4 billion chance of a false positive making it impossible to sign a plugin without breaking it (overwriting the binary).

src/H5PLint.c
fortnern: Right, but why not allow a run-time switch to enable it for when it's not enabled at compile time?

brtnfld: Good point, and not unreasonable as a future enhancement. The current H5_REQUIRE_DIGITAL_SIGNATURE flag serves a dual purpose: it both gates the OpenSSL dependency and enforces it. Without it, the OpenSSL code is entirely absent from the binary — adding a runtime "enable" switch would require always compiling in the OpenSSL code and always linking against OpenSSL, even for users who never want plugin signing. That felt like too large a default footprint for a security-optional feature. If there's demand for "compile support in but leave it opt-in at runtime", that could be done as a follow-on with a separate HDF5_BUILD_PLUGIN_SIGNING_SUPPORT CMake option alongside the existing HDF5_REQUIRE_SIGNED_PLUGINS. For now, keeping it simple.

src/H5PLsig.c
brtnfld: The keystore is only initialized lazily, H5PL__init_keystore is only called from the plugin load path, so if the user never loads a plugin, this code is never reached. If they do try to load a plugin and there are no keys, verification is impossible, so failing hard is intentional. A plugin loaded without a way to verify its signature would silently defeat the security guarantee provided by the compiled-in flag.

src/H5PLsig.c
brtnfld: Tracing through the logic: if HDF5_PLUGIN_KEYSTORE is set, env_keystore != NULL so keystore_configured = true, and the message prints "Attempted to load from: [path]". The "No keystore configured" branch only fires when env_keystore == NULL and H5PL_KEYSTORE_DIR is not defined, i.e., truly nothing was configured. Could you give a specific scenario you're concerned about? I may be misreading the case you have in mind.

src/H5PLsig.h
brtnfld: Fair point. The reserved field doesn't actually help a future version of the library parse a newer footer that it doesn't understand. That's what format_version is for. And you're right that if we ever need more space, 2 bytes probably won't be enough anyway. Happy to drop the reserved field and shrink the footer to 10 bytes. Just let me know.

src/H5PLsig.c
fortnern: Couldn't we just move the check for num_keys==0 to the code that verifies a plugin?

tools/src/h5sign/h5sign.c
brtnfld: The truncate is needed because the signature is appended to the file, not stored at a fixed offset. After stripping, file_size = binary_size so the rest of the signing logic treats the file as a plain binary and appends the new signature at the end. If we sought to binary_size and wrote from there instead, we'd need a truncate at the end anyway to remove any remaining old bytes (e.g., if the new signature is shorter than the old one). Truncating upfront to clean binary size is simpler and avoids that trailing-bytes edge case.

docs/PLUGIN_SIGNATURE_README.md
brtnfld: The 1-in-4-billion false positive is real but unlikely and recoverable, the user runs with --force and it works. As for signing the whole file, including the old signature: that would work (the library reads the last footer, which correctly covers the full preceding bytes), but each re-sign would permanently bloat the file by (old_sig_len + 12) bytes. For plugins distributed frequently or in size-constrained environments, that's undesirable. Other signing tools (e.g., codesign, apksigner) also replace rather than stack signatures for the same reason. The --force approach is consistent with that.

src/H5PLsig.c
brtnfld: Good idea. Moved the check — H5PL__init_keystore now succeeds with an empty keystore, and the "no valid public keys" error (with configuration guidance) is reported from H5PL__verify_signature_appended at the single enforcement point. This also fixes an inconsistency where H5PL_keystore_initialized_g was set before the old error, causing different messages on first vs. subsequent plugin loads.

src/H5PLsig.h
fortnern: Yes I think that's best. I'm thinking we should order the fields:
algorithm_id
signature_length
magic
version

So that the magic is always in the same place relative to the end and will always be included in a speculative read, and similarly signature_length will always be available after the speculative read so we will always know exactly how much to read for the second read. I'm certainly open to arguments otherwise.

tools/src/h5sign/h5sign.c
fortnern: It would still be possible to remove a truncate most of the time, but I suppose the way it is now is OK since this code isn't performance critical, and since truncate isn't likely to be expensive in serial

src/H5PLsig.h
brtnfld: The current magic-first ordering was requested by Jordan so that the magic can be decoded and verified before interpreting any remaining fields, thereby avoiding parsing untrusted data from an unsigned file. 
@jhendersonHDF, thoughts?

brtnfld: Magic-first (current, Jordan's request):

Naturally enforces "verify before parsing" in code — the security property is structural, not just procedural
Downside: if the footer grows in a future version, a v1 reader can't find magic and reports "not signed" instead of "unsupported version"

Magic-at-fixed-offset-from-EOF (Neil's proposal: [algo_id][sig_len][magic][version]):

Guarantees that any reader can always find magic + version regardless of footer changes, producing meaningful errors on version mismatch
sig_len is also at a fixed offset, so the two-phase read always works
Downside: magic isn't the first thing parsed in natural byte order; the decode function must deliberately check magic out of order (though this is easily enforced with a comment + assert)

Since the full footer is always read in one 10-byte I/O, the security property (check magic before trusting other fields) can be enforced in code regardless of storage order. The question is whether we value structural enforcement (magic first makes bad parsing order harder to write) or forward compatibility (magic at fixed EOF offset gives better version-mismatch diagnostics).

@jhendersonHDF  — given these tradeoffs, do you still prefer magic-first, or does the forward-compatibility argument for magic-near-end change your view?

fortnern: There's no reason you can't parse the magic first with it at the end. You'll just need to move the decoding pointer afterwards

fortnern: If we put the magic first and a future version of the file format changes or removes the algorithm_id field, then the library will need to decode the version field before it knows where to look for the magic, and attempting to decode and interpret information before verifying the magic field is what we're trying to avoid here.

jhendersonHDF: I guess I don't see why the library would need to decode the version field before knowing where to look for the magic field. The point is to always decode and validate magic first before attempting to decode any other field, as that's the whole purpose for it in the first place (even though it's a weak guarantee). If the fields are arranged as `magic` -> `version` -> `...` forever, then we know magic is always the first field even if the footer changes in the future (assuming we continue to keep the arrangement `magic` -> `...`). Not that it's a big deal, but it also saves us from having to reset the decoding pointer.

fortnern: The problem is if a new version changes the size of the footer we don't know where the footer begins until we decode the version. The only way to verify the magic first is to put it at a fixed location relative to the end of the file.

docs/PLUGIN_SIGNATURE_README.md
fortnern: It's not recoverable - if the binary has the magic number in the right (wrong) place, and the user uses --force, it will overwrite part of the binary

fortnern: It's recoverable only by changing the code of h5sign, which maybe is fine for 1 in 4 billion

src/H5PLsig.h
jhendersonHDF: I was envisioning starting from the beginning of the buffer that was read from the end of the file and locating the `magic` bytes. That would work as long as the size of the footer only grows in the future, but would be an issue if the footer shrinks unless we always speculatively read the largest size the footer has ever been. In either case we may have to do a bit of jumping around in the buffer, so I'm fine with either way. I'm just more used to magic bytes coming first in an encoding, but this is a bit of a special case.

jhendersonHDF: Also just wanting to note that unless we really think the footer may grow huge one day, always speculatively reading the largest size the footer has ever been and following it with a `memchr()` or something similar could allow us to always get the footer in a single read.

fortnern: Also, if the magic number is not at a fixed position relative to the end of the file, if an older library tries to read (or sign) a plugin signed with a newer version, the library won't be able to tell the difference between an unsigned plugin and one signed with a newer version of the library

fortnern: Finally  the magic number is just "HDF5" which is more likely to show up than random number (especially in an HDF5 plugin!). HDF5 files use an 8 byte magic number with some non-ASCII bytes to get around this. I think we should probably use a similar scheme here. The analogous fields in metadata pieces aren't comparable since there's no situation where a spurious "OHDR" in the file, for example, would cause a problem

jhendersonHDF: > the library won't be able to tell the difference between an unsigned plugin and one signed with a newer version of the library

I think that's a decent argument for putting `magic` toward the end, but it seems this depends on what we want the forward compatibility guarantees to be. I assume older versions of the library shouldn't necessarily be able to load newer plugins, but maybe that can be allowable.

If a new version of the library grows the footer size such that the older library reads an amount of footer bytes that doesn't include the `magic` field (or only part of it), then it won't be able to locate those bytes and will fail with a potentially misleading error message, whereas we could've given a more specific error about a version mismatch.

If a new version of the library shrinks the footer size such that the older library over-reads, it will still be able to locate those bytes and then decode the version and give a mismatch error, if necessary.

> Finally the magic number is just "HDF5" which is more likely to show up than random number (especially in an HDF5 plugin!)

Yes, regardless of this discussion I was going to suggest we probably want to consider something more specific for the magic bytes.

fortnern: If a new version shrinks the footer in a way that moves the magic, the old version won't be able to tell it was signed.

Also sorry my last comment was supposed to go in the other thread but I suppose it's fine here.

jhendersonHDF: > If a new version shrinks the footer in a way that moves the magic, the old version won't be able to tell it was signed.

Yes, to me that seems acceptable _iff_ we're assuming that an older version of the library should fail to load newer format plugins. In that case, unsigned or signed with a newer version shouldn't matter; it should fail in both cases. If we assume older versions _can_ load newer format plugins, then it's an issue but also can be solved with a fixed upper bound speculative read size not to ever be exceeded.

That said, I think we should aim for using a speculative read size that is at least as big as the footer has ever been regardless of the encoding format and make sure that gets adjusted upward as the footer grows (if we don't use a fixed upper bound speculative read size). It seems to me that we should always be able to capture the entire footer in a single read for versions of the library using the newest plugin format, again assuming that the footer shouldn't ever be more than a handful of bytes.

fortnern: A speculative read large enough for the largest footer + possibly the largest (reasonable?) signature is fine with me. I still think we should have the magic, version, and probably the signature length at the end of the file/footer.

jhendersonHDF: I think either encoding format works well, with the main benefit to having these fields at the end being that we can give more reasonable error messages when an older version of the library loads a newer format signed plugin.

@brtnfld let's go ahead and put the fields as @fortnern suggested. That is,

```
algorithm_id
signature_length
magic
version
```

While still making sure `magic` is decoded and validated first.

fortnern: @jhendersonHDF , do you agree that we should switch to an 8-byte magic value that includes non-ASCII characters (but not exactly the same as the current HDF5 signature)?

jhendersonHDF: Yes, I think it makes sense to use something that's less likely to be in a file, though 8 bytes seems specific. Is that just to match the existing length of the HDF5 magic value?

fortnern: I suppose it doesn't have to be 8 but that seems like a logical choice, and I think we want around the same level of certainty in this case that we do with the HDF5 file signature

brtnfld: Replace the 4-byte uint32_t magic (0x48444635 / "HDF5") with an 8-byte
non-ASCII magic sequence {0x89, 'H', 'P', 'S', '\r', '\n', 0x1A, '\n'},
modeled on the HDF5 file signature but distinct. This increases the
footer from 10 to 14 bytes.

The non-ASCII bytes reduce false positives in arbitrary binary data
(especially HDF5 plugins that may contain "HDF5" as ASCII), and the
8-byte length matches the HDF5 file signature convention.

Footer field order remains: algo_id(1) | sig_len(4) | magic(8) | ver(1),
keeping magic and version at fixed offsets from EOF for forward
compatibility with future footer versions.

brtnfld: "HPS" for HDF5 Plugin Signature

fortnern: That sounds good, but it doesn't look like that change has made it to this PR yet

brtnfld: @fortnern, I forgot to push earlier. It's uploaded now.

fortnern (approved):

bmribler (approved):

release_docs/CHANGELOG.md
```diff
@@ -97,6 +97,10 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## Library
 
+### Added optional digital signature verification for dynamically loaded plugins
+
+   When built with `-DHDF5_REQUIRE_SIGNED_PLUGINS=ON` and OpenSSL, HDF5 will cryptographically verify each plugin before loading it. Plugins are signed with the new `h5sign` tool, which appends an RSA signature and a compact footer to the plugin binary. Verification uses a keystore directory of trusted public keys, configurable at compile time (`-DHDF5_PLUGIN_KEYSTORE_DIR=<path>`) or at runtime via the `HDF5_PLUGIN_KEYSTORE` environment variable. Individual signatures can be revoked without removing the entire public key by listing their SHA-256 hashes in a `revoked_signatures.txt` file in the keystore directory. Supported algorithms include SHA-256, SHA-384, and SHA-512 with both PKCS#1 v1.5 and PSS padding. See `release_docs/PLUGIN_SIGNATURE_README.md` for details.
```
jhendersonHDF, line 102:
```suggestion
   When built with `-DHDF5_REQUIRE_SIGNED_PLUGINS=ON` and OpenSSL, HDF5 will cryptographically verify each plugin before loading it. Plugins are signed with the new `h5sign` tool, which appends an RSA signature and a compact footer to the plugin binary. Verification uses a keystore directory of trusted public keys, configurable at compile time (`-DHDF5_PLUGIN_KEYSTORE_DIR=<path>`) or at runtime via the `HDF5_PLUGIN_KEYSTORE` environment variable. Individual signatures can be revoked without removing the entire public key by listing their SHA-256 hashes in a `revoked_signatures.txt` file in the keystore directory. Supported algorithms include SHA-256, SHA-384, and SHA-512 with both PKCS#1 v1.5 and PSS padding. See `docs/PLUGIN_SIGNATURE_README.md` for details.
```

test/test_plugin_signature.c
```diff
@@ -0,0 +1,1307 @@
...
+
+    snprintf(full_path, sizeof(full_path), "%s/%s", base_dir, dir_name);
+
+    if (HDmkdir(full_path, permissions) < 0) {
+        fprintf(stderr, "Failed to create keystore directory: %s\n", full_path);
+        return NULL;
+    }
+
+#ifdef H5_HAVE_WIN32_API
+    /* Restrict ACLs so the keystore permission check passes.
```
jhendersonHDF, line 339: Not needed anymore

jhendersonHDF (approved):
Minor changes but otherwise I think it looks good

test/test_plugin_signature.c
brtnfld: fixed

hyoklee (changes requested):
Add `--version` option that matches `-V`.

fortnern (approved):

jhendersonHDF (approved):

hyoklee:
Did anyone succeed signing ZFP plugin on OpenBSD?

hyoklee (changes requested):
Fix ZFP plugin first and run h5sign test on ZFP later again:

```
[ 78%] Building C object ZFP/example/CMakeFiles/h5ex_d_zfp.dir/h5ex_d_zfp.c.o
/home/runner/work/actions/actions/hdf5_plugins/ZFP/example/h5ex_d_zfp.c:179:73: warning: format specifies type 'int' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
  179 |             printf("   Number of parameters is %d with the value %u\n", nelmts, values_out[0]);
      |                                                ~~                       ^~~~~~
      |                                                %zu
1 warning generated.
[ 79%] Linking C executable ../../bin/h5ex_d_zfp
ld: error: unable to find library -ldl
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error 1 in . (ZFP/example/CMakeFiles/h5ex_d_zfp.dir/build.make:102 'bin/h5ex_d_zfp')
*** Error 2 in . (CMakeFiles/Makefile2:2538 'ZFP/example/CMakeFiles/h5ex_d_zfp.dir/all')
*** Error 2 in /home/runner/work/actions/actions/plugins-build (Makefile:166 'all': /usr/bin/make -s -f CMakeFiles/Makefile2 all)
```

lrknox:
I don't think the OpenBSD test configuration for push/pull request enables plugins, sp it's probably not tested there.  I'll take a closer look later on.

jhendersonHDF (approved):
Approving as admin to proceed with merge

hyoklee (commented):
I can't make this PR work on OpenBSD with [plugins pr248](https://github.com/HDFGroup/hdf5_plugins/pull/248).

```
[ 99%] Building C object _deps/blosc2-build/blosc/CMakeFiles/blosc2_static.dir/blosc2.c.o
[ 99%] Linking C static library ../../../bin/libblosc2.a
Merging bundled dependency archives into libblosc2.a
/usr/bin/ar: error: script line 4: could not parse library: The file was not recognized as a valid object file
CMake Error at /home/runner/work/actions/actions/plugins-build/_deps/blosc2-src/cmake/merge_static_libs.cmake:75 (message):
  Failed to merge static libraries into
  /home/runner/work/actions/actions/plugins-build/bin/libblosc2.a
  merge_blosc2_static_Release.cmake:7 (include)


*** Error 1 in . (_deps/blosc2-build/blosc/CMakeFiles/blosc2_static.dir/build.make:439 'bin/libblosc2.a')
*** Error 2 in . (CMakeFiles/Makefile2:2177 '_deps/blosc2-build/blosc/CMakeFiles/blosc2_static.dir/all')
*** Error 2 in /home/runner/work/actions/actions/plugins-build (Makefile:166 'all': /usr/bin/make -s -f CMakeFiles/Makefile2 all)
```

hyoklee (approved):

PR #6214, opened 2026-02-12
consistent extension for C++

MarDiehl:
except for this file, all use cpp instead of cxx. No obvious reason why the same file types should have different extensions. Might be helpful to automatically reject files with cxx extension

----

> [!IMPORTANT]
> Rename `h5_api_test_driver.cxx` to `h5_api_test_driver.cpp` and update `CMakeLists.txt` for consistency.
> 
>   - **File Renaming**:
>     - Rename `h5_api_test_driver.cxx` to `h5_api_test_driver.cpp` for consistency.
>   - **CMakeLists Update**:
>     - Update `CMakeLists.txt` to use `h5_api_test_driver.cpp` instead of `h5_api_test_driver.cxx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 42dcf257a44fc9c0f786060b1b7e323188f67d1f. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/test/API/driver/CMakeLists.txt
+++ b/test/API/driver/CMakeLists.txt
@@ -30,5 +30,5 @@ endif()
 
 include_directories(${hdf5_kwsys_BINARY_DIR})
 
-add_executable(h5_api_test_driver h5_api_test_driver.cxx)
+add_executable(h5_api_test_driver h5_api_test_driver.cpp)
 target_link_libraries(h5_api_test_driver h5_api_test_sys)
--- a/test/API/driver/h5_api_test_driver.cpp
+++ b/test/API/driver/h5_api_test_driver.cpp
@@ -21,16 +21,16 @@
 #include <cstdlib>
 
 #if !defined(_WIN32) || defined(__CYGWIN__)
-# include <unistd.h>
-# include <sys/wait.h>
+#include <unistd.h>
+#include <sys/wait.h>
 #endif
 
 #include <h5_api_test_sys/RegularExpression.hxx>
 #include <h5_api_test_sys/SystemTools.hxx>
 
-using std::vector;
-using std::string;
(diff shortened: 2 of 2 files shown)


MarDiehl:
Looks like the format check fails because it was not active for `cxx`. I can fix this, but it makes only sense if you agree that this renaming makes sense.

jhendersonHDF:
@MarDiehl you'll need to rebase off of/merge in the latest develop branch as we had to fix a CI issue.

As far as the naming goes, it's inconsistent because the file was copied and modified from another project where `.cxx` is used and the file is unrelated to our C++ wrappers, so the file extension wasn't ever considered. No reason it can't be renamed; it should just be excluded from clang-format in that case.

hyoklee (dismissed):
Thank you for your patch, @MarDiehl !

MarDiehl:
@jhendersonHDF: So you don't want the style adjustments because the file was copied? Otherwise, getting it fixed by clang would be easier than setting an exception.

jhendersonHDF:
> @jhendersonHDF: So you don't want the style adjustments because the file was copied? Otherwise, getting it fixed by clang would be easier than setting an exception.

Right, I think it makes it easier to patch in changes if needed when the formatting is the same as the original. However, this file hasn't been modified or directly used in our testing in a while, so I'm OK with either direction you choose to go.

MarDiehl:
from my side, ready to review.

jhendersonHDF (approved):

bmribler (approved):

PR #6241, opened 2026-02-25
docs: add 2.

hyoklee:
2.0 is released.

----

> [!IMPORTANT]
> Update `file-locking.dox` to include version 2.0 in the file locking scheme documentation.
> 
>   - **Documentation**:
>     - Update `file-locking.dox` to include version 2.0 in the list of versions carrying forward the file locking scheme.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 514de80641788158241e83055742f6f219c896b6. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/doxygen/dox/file-locking.dox
+++ b/doxygen/dox/file-locking.dox
@@ -55,7 +55,7 @@ it wrong could result in corrupt files or crashed readers, we decided to add
 a file locking scheme to help users get it right. Since this would also help
 prevent harmful accesses when SWMR is not in use, we decided to switch the
 file locking scheme on by default. This scheme has been carried forward into
-HDF5 1.12 and 1.14 (soon to be 2.0).
+HDF5 1.12, 1.14, and 2.0.
 
 Note that the current implementation of SWMR is only useful for appending to chunked
 datasets. Creating file objects like groups and datasets is not supported
--- a/doxygen/dox/file-locking.dox
+++ b/doxygen/dox/file-locking.dox
@@ -55,7 +55,7 @@ it wrong could result in corrupt files or crashed readers, we decided to add
 a file locking scheme to help users get it right. Since this would also help
 prevent harmful accesses when SWMR is not in use, we decided to switch the
 file locking scheme on by default. This scheme has been carried forward into
-HDF5 1.12, 1.14, and 2.0.
+HDF5 1.12, 1.14, and 2.
 
 Note that the current implementation of SWMR is only useful for appending to chunked
 datasets. Creating file objects like groups and datasets is not supported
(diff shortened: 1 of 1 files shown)


doxygen/dox/file-locking.dox
```diff
@@ -55,7 +55,7 @@ it wrong could result in corrupt files or crashed readers, we decided to add
 a file locking scheme to help users get it right. Since this would also help
 prevent harmful accesses when SWMR is not in use, we decided to switch the
 file locking scheme on by default. This scheme has been carried forward into
-HDF5 1.12 and 1.14 (soon to be 2.0).
+HDF5 1.12, 1.14, and 2.0.
```
bmribler: Should it be 2.x or something to indicate the series?  @lrknox?

bmribler (changes requested):

mattjala (dismissed):

doxygen/dox/file-locking.dox
lrknox: The root of the 2.0 vs 2.x question is the change for major versions from 2 version fields to 1, i.e 1.12 and 1.14 to 2.  So for a series of major versions this would be HDF5 1.12, 1.14, and 2.  What we have here is major, major, and minor versions.  The statement is true with any of 2, 2.0, or 2.x. 

I can see reasons for either 2.0 or 2.x and I think either will work here.

@bmribler, thank you for raising the question.  If you feel strongly that it should be 2.x we'll continue the discussion.  If you're ok with 2.0 here, you'll need to approve to clear the change request.

mattjala (dismissed):

hyoklee:
@bmribler , please finish review.

doxygen/dox/file-locking.dox
```diff
@@ -55,7 +55,7 @@ it wrong could result in corrupt files or crashed readers, we decided to add
 a file locking scheme to help users get it right. Since this would also help
 prevent harmful accesses when SWMR is not in use, we decided to switch the
 file locking scheme on by default. This scheme has been carried forward into
-HDF5 1.12 and 1.14 (soon to be 2.0).
+HDF5 1.12, 1.14, and 2.
```
lrknox:
```suggestion
HDF5 1.12 and later versions.
```

bmribler (approved):

lrknox (approved):

PR #6244, opened 2026-02-26
Fix POSIX header guard to include _WIN32 for Intel icx on Windows

brtnfld:
Intel's icx compiler defines _WIN32 (with underscore) but not WIN32, causing unistd.h and other POSIX headers to be incorrectly included on Windows builds.

----

> [!IMPORTANT]
> Fixes header guards to include `_WIN32` for Intel icx on Windows, preventing incorrect POSIX header inclusion.
> 
>   - **Behavior**:
>     - Fixes header guard to include `_WIN32` in `test/atomic_reader.c`, `test/atomic_writer.c`, and `tools/test/perform/direct_write_perf.c`.
>     - Prevents incorrect inclusion of POSIX headers on Windows when using Intel's icx compiler.
>   - **Files Affected**:
>     - `test/atomic_reader.c`: Adds `_WIN32` to the header guard.
>     - `test/atomic_writer.c`: Adds `_WIN32` to the header guard.
>     - `direct_write_perf.c`: Adds `_WIN32` to the header guard.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 52194da730902d17aa2356bb6b36385ff29e96c9. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/test/atomic_reader.c
+++ b/test/atomic_reader.c
@@ -31,7 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if !defined(WIN32) && !defined(__MINGW32__)
+#if !defined(WIN32) && !defined(__MINGW32__) && !defined(_WIN32)
 
 #include <fcntl.h>
 #include <unistd.h>
--- a/test/atomic_writer.c
+++ b/test/atomic_writer.c
@@ -34,7 +34,7 @@
 #include <stdint.h>
 #include <string.h>
 
-#if !defined(WIN32) && !defined(__MINGW32__)
+#if !defined(WIN32) && !defined(__MINGW32__) && !defined(_WIN32)
 
 #include <fcntl.h>
 #include <unistd.h>
--- a/tools/test/perform/direct_write_perf.c
+++ b/tools/test/perform/direct_write_perf.c
@@ -25,7 +25,7 @@
 #include H5_ZLIB_HEADER /* "zlib.h" */
 #endif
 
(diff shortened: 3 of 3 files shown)


mattjala (approved):

test/atomic_reader.c
```diff
@@ -31,7 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if !defined(WIN32) && !defined(__MINGW32__)
+#if !defined(WIN32) && !defined(__MINGW32__) && !defined(_WIN32)
```
jhendersonHDF, line 34: It seems like it would be more reliable to use the results of configure checks that we already have for including some of these headers such as H5_HAVE_UNISTD_H. Having said that, it looks like these files are only meant to be compiled on Unix, so maybe we should specifically check for that or make them more portable. Any indication why our current icx testing works? It appears to be building these test files.

brtnfld: I've not looked into why the builds pass, so I'm not sure. I know it passes for CGNS CI because I don't build HDF5 tests.

jhendersonHDF (approved):
We should look into these test programs more as it looks like it should be fairly straightforward to build them for WIN32 as long as they're tested on appropriate filesystems only.

PR #6272, opened 2026-03-05
Improve performance of H5Ovisit with deeply nested groups

### LINKED ISSUE #6225: H5Ovisit complexity grows quadratically with recursion depth
This issue was first reported by [HO9](https://github.com/HO-9) in the [H5py repo](https://github.com/h5py/h5py/issues/2741). The H5py team quickly determined that this is not an `h5py` issue, but a problem in the HDF5 library.

A C reproducer was provided by [HO9](https://github.com/h5py/h5py/issues/2741#issuecomment-3545067675).

It clearly shows a linear increase in complexity for `H5Lvisit`, whereas `H5Ovisit` exhibits quadratic growth. The root cause is that the current object visitor implementation traverses links and visits objects while re-resolving each object from the root path.

Food for thought on how to fix this can be found [here](https://github.com/gheber/hdf5/blob/bug/visitors/src/H5Oint.c).

fortnern:
Improve performance of H5Ovisit (and H5Ocopy, and functions that retrieve an object name) with deeply nested groups by passing more information about the visited object from the underlying H5G_visit routine to these callbacks. This eliminates the need to look up the object by the path name in the callback. Introduced an internal object callback for H5G_visit to facilitate this. H5Ovisit1 and potentially H5Ovisit2 are still slow with deeply nested groups due to the way these deprecated functions interact with the VOL layer.

Resolves #6225

### DIFF
--- a/src/H5Gint.c
+++ b/src/H5Gint.c
@@ -69,7 +69,8 @@ typedef struct {
     char           *path;          /* Path name of the link */
     size_t          curr_path_len; /* Current length of the path in the buffer */
     size_t          path_buf_size; /* Size of path buffer */
-    H5L_iterate2_t  op;            /* Application callback */
+    H5L_iterate2_t  link_op;       /* Application callback for link operations */
+    H5G_obj_iterate_t obj_op;      /* Library internal callback for object operations */
     void           *op_data;       /* Application's op data */
 } H5G_iter_visit_ud_t;
 
@@ -981,7 +982,6 @@ static herr_t
 H5G__visit_cb(const H5O_link_t *lnk, void *_udata)
 {
     H5G_iter_visit_ud_t *udata = (H5G_iter_visit_ud_t *)_udata; /* User data for callback */
-    H5L_info2_t          info;                                  /* Link info */
     H5G_loc_t            obj_loc;                               /* Location of object */
     H5G_name_t           obj_path;                              /* Object's group hier. path */
     H5O_loc_t            obj_oloc;                              /* Object's object location */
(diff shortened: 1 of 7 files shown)


mattjala (dismissed):

src/H5Gint.c
```diff
@@ -1051,72 +1056,79 @@
...
-                                             H5G__visit_cb, udata);
-
-                /* Restore location */
-                udata->curr_loc = old_loc;
-            } /* end if */
-        }     /* end if */
-    }         /* end if */
+
+                    /* Point to this group's location info */
+                    udata->curr_loc = &obj_loc;
```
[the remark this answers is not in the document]
fortnern: This is fine because obj_loc will be in scope during the call to H5G__obj_iterate, and udata->curr_loc is reset afterwards.

src/H5Oint.c
```diff
@@ -2535,85 +2513,32 @@ H5O__free_visit_visited(void *item, void H5_ATTR_UNUSED *key, void H5_ATTR_UNUSE
...
     FUNC_ENTER_PACKAGE
 
     /* Sanity check */
     assert(name);
-    assert(linfo);
+    assert(obj_oloc);
     assert(udata);
 
-    /* Check if this is a hard link */
-    if (linfo->type == H5L_TYPE_HARD) {
```
jhendersonHDF, line 2555: It's a bit difficult to tell from the diff, but is this behavior accounted for somewhere else now?

mattjala (approved):

src/H5Oint.c
fortnern: Yes it's handled in H5G__visit_cb(). This callback (H5O__visit_cb) is now invoked for each object, not each link, so the H5G_visit() machinery has already checked the link type.

jhendersonHDF (approved):

PR #6273, opened 2026-03-06
chore: fix typos

hyoklee:
`mouting` -> `mounting`

### DIFF
--- a/c++/src/H5Location.cpp
+++ b/c++/src/H5Location.cpp
@@ -1723,7 +1723,7 @@ H5Location::mount(const H5std_string &name, const H5File &child, const PropList 
 void
 H5Location::unmount(const char *name) const
 {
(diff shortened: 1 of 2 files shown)


c++/src/H5Location.cpp
```diff
@@ -1723,7 +1723,7 @@ H5Location::mount(const H5std_string &name, const H5File &child, const PropList
 void
 H5Location::unmount(const char *name) const
 {
-    // Call C routine H5Fmount to do the mouting
+    // Call C routine H5Fmount to do the mounting
```
lrknox: @hyoklee should this be "unmounting", since the function is unmount?  Probably the  comment was copy/pasted from the mount function

lrknox (approved):

bmribler (approved):

Issue #5975, opened 2025-11-04
3 Test Failures with MinGW Static GCC Configuration (HDF5_MINGW_STATIC_GCC_LIBS=ON)


hyoklee:
## Description

Three tests are failing when building HDF5 with the MinGW Static GCC configuration on Windows. The failures were detected in CDash build ID 3178953.

**CDash Build Report:** https://my.cdash.org/viewTest.php?onlyfailed&buildid=3178953

## Failed Tests

1. **H5TESTXPR-cache_api** - Execution time: 40ms
2. **H5TEST-dt_arith** - Execution time: 250ms  
3. **H5TEST-err_compat** - Execution time: 40ms

## Build Configuration

The failures occurred with the following CMake configuration:

```cmake
cmake -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DHDF5_MINGW_STATIC_GCC_LIBS:BOOL=ON \
  -DBUILD_SHARED_LIBS:BOOL=ON \
  -DHDF5_BUILD_CPP_LIB:BOOL=ON \
  -DHDF5_BUILD_FORTRAN:BOOL=OFF \
  -DHDF5_BUILD_JAVA:BOOL=OFF \
  -DHDF5_BUILD_HL_LIB:BOOL=ON \
  -DHDF5_ENABLE_ZLIB_SUPPORT:BOOL=OFF \
  -DHDF5_ENABLE_SZIP_SUPPORT:BOOL=OFF
```

## Environment

- **Platform:** Windows Server 2022 (Build 26100)
- **Architecture:** AMD64
- **Compiler:** MinGW-w64 GCC (D:/a/_temp/msys64/mingw64/bin/c++.exe)
- **Build System:** Ninja
- **MSYS2 System:** MINGW64
- **Build Name:** mw/cpp
- **Site:** win-22
- **Test Time:** 2025-11-04 19:29:49 UTC

## Workflow

This configuration is tested via the GitHub Actions workflow: `win_mgw_sta.yml`

The key configuration flag being tested is:
- `HDF5_MINGW_STATIC_GCC_LIBS:BOOL=ON` - This enables static linking of GCC runtime libraries on MinGW

## Additional Information

All three tests completed execution but failed their validation checks. The total test execution time was approximately 5 minutes 42 seconds for the entire test suite.

## Expected Behavior

All tests should pass with the MinGW Static GCC configuration, as this is a supported build configuration for HDF5 on Windows.

## Reproduction

[...]

Issue #5067, opened 2024-11-04
HDF5 test configuration with Cygwin fails 7 tests


lrknox:
These 7 tests failed 11/4/2024
	H5IMPORT-STR-H5DMP_CMP,0.16, (Failed)
	H5REPACK-deflate_file,0.11, (Failed)
	H5TESTXPR-cache_api,0.1, (Failed)
	H5TEST-accum,0.27, (Failed)
	H5TEST-dtypes,0.88, (Failed)
	H5REPACK-deflate_file_DFF,0.05, (Failed)
	H5TEST-err_compat,0.13, (Failed)

Error messages for the H5TEST-dtypes test:
Testing complex number datatype conversions                           *FAILED*
Conversion path for H5T_NATIVE_FLOAT_COMPLEX -> H5T_NATIVE_FLOAT_COMPLEX was not a no-op
Testing complex number datatype <-> array datatype compatibility conversions*FAILED*
Conversion path for complex type -> array type was not a no-op
Testing complex number datatype <-> compound datatype compatibility conversions*FAILED*
Conversion path for complex type -> compound type was not a no-op

The other 6 failed for various other reasons that may not be related to the H5TEST-dtypes test failure.

**Platform (please complete the following information)**
 - HDF5 version :  develop commit 35df4fb6
 - CYGWIN_NT-10.0-20348-3.5.4-1.x86_64-x86_64
 - Compiler and version:  GNU 12.4.0
 - Build system: CMake
 - Any configure options you specified:  -DCMAKE_BUILD_TYPE:STRING=Release  --log-level=VERBOSE -DHDF5_GENERATE_HEADERS=OFF -DHDF5_BUILD_JAVA:BOOL=OFF -DHDF5_BUILD_CPP_LIB:BOOL=OFF -DHDF5_BUILD_FORTRAN:BOOL=OFF -DHDF5_ENABLE_Z_LIB_SUPPORT:BOOL=ON -DHDF5_ENABLE_SZIP_SUPPORT:BOOL=ON -DHDF5_ENABLE_SZIP_ENCODING:BOOL=ON -DHDF5_ENABLE_PLUGIN_SUPPORT:BOOL=ON

matzeri:
FYI:

The current HDF5 1.14.5 cygwin package build log is available at  
https://github.com/cygwin/scallywag/actions/runs/12731686354/job/35486017261
https://github.com/cygwin/scallywag/actions/runs/12731686354

Currently built with Autotools
[...]

PR #6274, opened 2026-03-06
test/cache_api: avoid false rpt_fcn pointer mismatch on Windows

phfix:
## Summary
On Windows, raw function-pointer address comparison can differ because of import thunks even when callbacks are semantically equivalent.

This patch updates cache_api test config comparison to avoid false negatives:
- On _WIN32: compare rpt_fcn nullness
- On non-Windows: keep strict pointer equality

## Validation
- Rebuilt cache_api in MinGW
- Ran: ctest --test-dir build-mingw -R "^H5TESTXPR-cache_api$" --output-on-failure
- Result: pass

### DIFF
--- a/test/cache_common.c
+++ b/test/cache_common.c
@@ -4990,14 +4990,9 @@ resize_configs_are_equal(const H5C_auto_size_ctl_t *a, const H5C_auto_size_ctl_t
 {
     if (a->version != b->version)
         return (false);
-#if defined(_WIN32)
-    /*
-     * On Windows, function pointer addresses can differ because of import
-     * thunks even when they represent the same callback symbol.
-     */
     else if ((a->rpt_fcn == NULL) != (b->rpt_fcn == NULL))
         return (false);
-#else
+#ifndef H5_HAVE_MINGW
     else if (a->rpt_fcn != b->rpt_fcn)
         return (false);
 #endif
--- a/test/cache_common.c
+++ b/test/cache_common.c
@@ -4990,8 +4990,17 @@ resize_configs_are_equal(const H5C_auto_size_ctl_t *a, const H5C_auto_size_ctl_t
 {
     if (a->version != b->version)
         return (false);
+#if defined(_WIN32)
+    /*
+     * On Windows, function pointer addresses can differ because of import
+     * thunks even when they represent the same callback symbol.
+     */
+    else if ((a->rpt_fcn == NULL) != (b->rpt_fcn == NULL))
+        return (false);
+#else
     else if (a->rpt_fcn != b->rpt_fcn)
         return (false);
+#endif
     else if (compare_init && (a->set_initial_size != b->set_initial_size))
         return (false);
     else if (compare_init && (a->initial_size != b->initial_size))
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -83,11 +83,11 @@ jobs:
 
       - name: Run Tests
         run: |
-          ctest . --parallel 2 -C ${{ inputs.build_mode }} -E "cache_api|dt_arith|err_compat"
+          ctest . --parallel 2 -C ${{ inputs.build_mode }} -E "dt_arith|err_compat"
         working-directory: ${{ runner.workspace }}/build
 
       - name: Run Expected to Fail Tests
         run: |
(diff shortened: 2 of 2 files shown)


lrknox:
Codespell check failed due to typos unrelated to this PR.  The typos  have been fixed in PR #6273, now merged to develop.

hyoklee (dismissed):
@phfix , thank you for the fix!

[Test Results](https://my.cdash.org/index.php?project=HDF5&date=2026-03-06)

We'd appreciate if you could update `.github/workflows/msys2.yml` as well in this PR.

phfix:
I updated .github/workflows/msys2.yml as requested and force-pushed the branch. The key thing to check is that cache_api now runs normally in the MSYS2 jobs instead of being treated as an expected failure.

hyoklee (dismissed):

test/cache_common.c
```diff
@@ -4990,8 +4990,17 @@ resize_configs_are_equal(const H5C_auto_size_ctl_t *a, const H5C_auto_size_ctl_t
 {
     if (a->version != b->version)
         return (false);
+#if defined(_WIN32)
+    /*
+     * On Windows, function pointer addresses can differ because of import
+     * thunks even when they represent the same callback symbol.
+     */
+    else if ((a->rpt_fcn == NULL) != (b->rpt_fcn == NULL))
```
jhendersonHDF, line 4993: @phfix I'm thinking that it's probably not very useful to check that both `rpt_fcn`s are non-NULL or NULL here, especially since it's logically not the same thing as `a->rpt_fcn != b->rpt_fcn`. In this case I'd prefer either only performing the check for `#ifndef _WIN32`, or just dropping the function pointer comparison check and the platform-specific #ifdef altogether.

lrknox (dismissed):

test/cache_common.c
phfix: I think H5_HAVE_MINGW would be a better guard than _WIN32 here. The issue is really about MinGW’s DLL import/auto-import handling and thunk indirection, not Windows in general. Even though checking NULL/non-NULL is not the same as strict pointer equality, I still think it is better than skipping the test entirely, because it preserves the meaningful part we can still verify here: whether a report function is assigned or not.
What do you prefer?

jhendersonHDF: In general I try to avoid platform-specific #ifdefs in testing if possible since it complicates the logic and what needs to be maintained. I guess my main point is that, if desired, the `(a->rpt_fcn == NULL) != (b->rpt_fcn == NULL)` check can simply be added for all platforms without needing to be behind an #ifdef, whereas it makes more sense to me to put `a->rpt_fcn != b->rpt_fcn` behind something like `#ifndef H5_HAVE_MINGW` if we want to keep it. Otherwise, I'd just drop that check to simplify things.

phfix: That makes sense. I can change it to keep the unconditional NULL/non-NULL check and put only the strict a->rpt_fcn != b->rpt_fcn comparison behind #ifndef _WIN32,  or H5_HAVE_MINGW e.g.

```
else if ((a->rpt_fcn == NULL) != (b->rpt_fcn == NULL))
    return (false);
#ifndef _WIN32
else if (a->rpt_fcn != b->rpt_fcn)
    return (false);
#endif
```

That keeps the meaningful assigned-vs-unassigned test on all platforms while limiting the platform-specific guard to the raw pointer equality check. There are also a couple of related platform-specific issues in this area, such as [#5975](https://github.com/HDFGroup/hdf5/issues/5975) and [#5067](https://github.com/HDFGroup/hdf5/issues/5067). If that looks right to you, I’ll update the branch accordingly.

jhendersonHDF: Yes that looks reasonable to me. It makes sense to me restrict to only `H5_HAVE_MINGW` if that passes.

phfix: I updated this to keep the unconditional NULL/non-NULL check and to guard only the strict pointer comparison with #ifndef H5_HAVE_MINGW. I also rebuilt cache_api in the MinGW environment and reran H5TESTXPR-cache_api, which passed:

cmake --build build-mingw --target cache_api
ctest --test-dir build-mingw -R "cache_api|H5TESTXPR-cache_api" --output-on-failure

hyoklee (approved):

jhendersonHDF (approved):

PR #6275, opened 2026-03-06
Correct spellings in comments causing spellcheck failures.

lrknox:


lrknox:
sharability is actually an alternate spelling for shareability.  Adding it to the list of words to ignore would be an alternative to changing the spelling in the comment, if that is preferred.

hyoklee:
Duplicate of #6273

lrknox:
Duplicate of #6273

PR #6280, opened 2026-03-10
 Default versioned API functions to earliest version for older API settings

### LINKED ISSUE #6278: Change default API version of functions introduced after global API version
Currently, when setting a global API version, for example with `#define H5_USE_16_API`, if the application invokes a function that was introduced after that API version (1.6 in the example), the library uses the *latest* version of that function. We should change it to use the earliest version of that function for maximum compatibility and minimum user headaches. Here is an example:

1. Version 1.6: Application written.
2. Version 1.8: Application developers insert `#define H5_USE_16_API` to maintain compatibility without rewriting their app
3. Version 1.10: HDF5 introduces `H5new_function()`, app developers integrate it into their app
4. Version 1.12: HDF5 versions `H5new_function()`, splitting it into `H5new_function1()` and `H5new_function2()`, with `H5new_function()` now being a macro that maps to one of the other two. At this point the app no longer works because the `H5new_function()` defaults to `H5new_function2()` while the app is coded for `H5new_function1()`

The proper usage would be to, at step 3, update API usage to version 1.10, since they have already broken compatibility with earlier versions of HDF5 by using `H5new_function()`. However, we believe there is no downside to making things work by defaulting to `H5new_function1()` at step 4, and it will help reduce the workload for users of HDF5.

brtnfld:
Fixes #6278 

The fix in bin/make_vers:

```
my $intro_vers = $vers_list[0];
foreach my $hash_vers (@versions) {
    if ($hash_vers < $intro_vers) {
        $api_vers_to_function_vers{$hash_vers}{$name} = 1;
    }
}
```

For every API version older than a function's introduction version, it now explicitly sets _vers = 1 (earliest). This means:

H5_USE_16_API -> H5Sencode_vers = 1 -> maps to H5Sencode1 (instead of falling through to H5Sencode2)
H5_USE_16_API -> H5Dread_chunk_vers = 1 -> maps to H5Dread_chunk1 (introduced v1.10, versioned in v2.0.0)
Functions already introduced at or before the configured API version are unchanged.

The auto-generated test in test/tapi_version_default.c validates this for every function × API-version combination by compiling the test six times (once per API level: v16, v18, v110, v112, v114, v200) and asserting each _vers macro equals the expected value.


brtnfld:
The netCDF tst_h_files4.c is calling H5Oopen_by_token and accessing info->u.token, which requires H5O_info2_t (version 2 of the struct, introduced in v1.12). But the build is getting H5O_info1_t (version 1), which has info->u.addr instead of info->u.token.

netcdf needs to:
1. Update the API version to at least H5_USE_112_API since the code uses v1.12 features (H5O_info2_t / .token)
2. Explicitly set #define H5O_info_t_vers 2 before including HDF5 headers
3. Use the versioned name directly: H5O_info2_t instead of H5O_info_t

brtnfld:
NetCDF fix, https://github.com/Unidata/netcdf-c/pull/3310

brtnfld:
NetCDF has been patched.

lrknox (dismissed):

fortnern (approved):

bmribler (approved):

Issue #6204, opened 2026-02-06
2.0.0 fail to build with zlibng 2.3.3


braindevices:
***
 ⛔ **IMPORTANT**  ⛔
 Security issues should not be reported here! Instead, please
 report them privately as security vulnerabilities; see for more
 information https://github.com/HDFGroup/hdf5/security/policy
***

**Describe the bug**
build fail with zlib-ng 2.3.3
```
_deps/hdf5_zlib-src/gzread.c.in does not exist.
```
it has introduced changes that gzread.c.in does not exist.
current patch strategy is probably quite bad instead of really using patch file, we copy our own cmakelist into the zlibng package, which is very vulnerable to any cmake change made by zlib-ng

zlib-ng v2.3.3
HDF5 v2.0.0

hyoklee:
Hi, @braindevices!

Thank you for testing zlilb-ng and your bug report.

Would you please test with develop branch?
I'm getting the following error with develop branch on Win-2025 / VS-2026 / vcpkg-installed zlib-ng:

`D:\a\actions\actions\src\H5Zdeflate.c(26,1): error C1083: Cannot open include file: 'zlib-ng.h': No such file or directory [D:\a\actions\actions\build\src\hdf5-static.vcxproj]`

I'd appreciate if you could give more information on your system (OS, compilers) and how you built it (e.g., cmake options).

jhendersonHDF:
The build of external zlib-ng has been updated to use 2.3.3 without patching it with the changes in https://github.com/HDFGroup/hdf5/pull/6287.

PR #6284, opened 2026-03-11
build: support vcpkg/brew zlib-ng 2.3.3

hyoklee:
CMake error occurs when a user installed zlib-ng 2.3.3 via vcpkg or brew and use it as zlib replacement.

```
CMake Error at src/CMakeLists.txt:1262 (target_link_libraries):
  Target "hdf5-static" links to:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at test/CMakeLists.txt:510 (target_link_libraries):
  Target "chunk_info" links to:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at test/CMakeLists.txt:538 (target_link_libraries):
  Target "direct_chunk" links to:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at tools/test/perform/CMakeLists.txt:130 (target_link_libraries):
  Target "zip_perf" links to:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



-- Generating done (40.4s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

This PR fixes the above error.
Tested on Win-11/2025+VS2026 (vcpkg), Mac-12 (brew), and Ubuntu-24.04 (vcpkg)

ref: #6204


jhendersonHDF (changes requested):
I already have fixes for zlib-ng in my branch https://github.com/jhendersonHDF/hdf5/tree/cmake_filters_changes

PR #6294, opened 2026-03-17
Document CMake options in a central location

jhendersonHDF:


.github/workflows/netcdf.yml
```diff
@@ -45,7 +45,8 @@ jobs:
             -G Ninja \
             --log-level=VERBOSE \
             -DCMAKE_BUILD_TYPE=Release \
-            -DHDF5_ONLY_SHARED_LIBS:BOOL=ON \
```
jhendersonHDF, line 56: `HDF5_ONLY_SHARED_LIBS` is an option mostly used internally and I'd like to avoid continuing to promote setting it since it confuses things and has odd interactions with cmake GUI programs.

CMakeBuildOptions.cmake
```diff
@@ -42,9 +42,16 @@ mark_as_advanced (HDF5_MINGW_STATIC_GCC_LIBS)
 
 option (HDF5_ENABLE_TRACE "Enable API tracing capability" OFF)
 mark_as_advanced (HDF5_ENABLE_TRACE)
+if (HDF5_ENABLE_TRACE)
```
jhendersonHDF, line 45: This option no longer does anything.

CMakeBuildOptions.cmake
```diff
@@ -42,9 +42,16 @@ mark_as_advanced (HDF5_MINGW_STATIC_GCC_LIBS)
...
+  message (DEPRECATION "HDF5_ENABLE_TRACE has been deprecated and may be removed in a future release of HDF5")
+endif ()
 
-option (HDF5_ENABLE_EMBEDDED_LIBINFO "Embed library info into executables" ON)
+option (HDF5_ENABLE_EMBEDDED_LIBINFO "Embed library info into executables" OFF)
 mark_as_advanced (HDF5_ENABLE_EMBEDDED_LIBINFO)
+if (HDF5_ENABLE_EMBEDDED_LIBINFO)
```
jhendersonHDF, line 51: This option doesn't do anything and it's not clear that it ever did.

release_docs/INSTALL_CMake_options.md
jhendersonHDF, line 1: While the CMake options were mostly documented in `INSTALL_CMAKE.md`, several were missing or were otherwise poorly documented. This documents most of them in a single place, with the main exception being that the filter options are discussed more in `INSTALL_Filters.md` since there is some complication/confusion around the interactions of the different options.

brtnfld:
I'm also wondering if these options should be moved to advanced?

HDF5_ENABLE_DEPRECATED_SYMBOLS
HDF5_ENABLE_NONSTANDARD_FEATURES
HDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16

niche features
HDFS, DIRECT, ROS3, MIRROR VFDs
HDF5_ENABLE_PLUGIN_SUPPORT

developer/CI
HDF5_DISABLE_COMPILER_WARNINGS
HDF5_ENABLE_WARNINGS_AS_ERRORS
HDF5_ENABLE_DEV_WARNINGS
HDF5_ENABLE_ALL_WARNINGS
HDF5_ENABLE_PROFILING
HDF5_ENABLE_OPTIMIZATION
HDF5_BUILD_DOC

jhendersonHDF:
> I'm also wondering if these options should be moved to advanced?
> 
> HDF5_ENABLE_DEPRECATED_SYMBOLS HDF5_ENABLE_NONSTANDARD_FEATURES HDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16
> 
> niche features HDFS, DIRECT, ROS3, MIRROR VFDs HDF5_ENABLE_PLUGIN_SUPPORT
> 
> developer/CI HDF5_DISABLE_COMPILER_WARNINGS HDF5_ENABLE_WARNINGS_AS_ERRORS HDF5_ENABLE_DEV_WARNINGS HDF5_ENABLE_ALL_WARNINGS HDF5_ENABLE_PROFILING HDF5_ENABLE_OPTIMIZATION HDF5_BUILD_DOC

I think a reasonable case could be made for all of them except maybe `HDF5_BUILD_DOC` I'd leave visible.

brtnfld:
> I think a reasonable case could be made for all of them except maybe `HDF5_BUILD_DOC` I'd leave visible.

Were you planning to do that in this PR?

jhendersonHDF:
> > I think a reasonable case could be made for all of them except maybe `HDF5_BUILD_DOC` I'd leave visible.
> 
> Were you planning to do that in this PR?

I'll do this in a separate PR since it's mostly unrelated

brtnfld (dismissed):

release_docs/INSTALL_CMake_options.md
```diff
@@ -0,0 +1,453 @@
...
+## General options
+
+These options concern the general build process of the main HDF5 libraries, utilities, documentation, etc.
+
+| CMake option | Type | Default | Description |
+|:-------------|:-----|:--------|:------------|
+| `BUILD_SHARED_LIBS` | `BOOL` | `ON` | If `ON`, builds shared HDF5 libraries. |
+| `BUILD_STATIC_LIBS` | `BOOL` | `ON` | If `ON`, builds static HDF5 libraries. |
+| `BUILD_STATIC_EXECS` | `BOOL` | `OFF` | If `ON`, builds statically-linked executables. **NOTE:** The `BUILD_STATIC_EXECS` option is only valid on some UNIX operating systems. It adds the `-static` flag to `CMAKE_EXE_LINKER_FLAGS`. This flag is not available on Windows and some modern Linux systems will ignore the flag. |
+| `HDF5_DEFAULT_API_VERSION` | `STRING` | `v200` | Specifies the default HDF5 API version to use when compiling HDF5 libraries. Valid values are `v200` (2.0.0 API), `v114` (1.14 API), `v112` (1.12 API), `v110` (1.10 API), `v18` (1.8 API) and `v16` (1.6 API). See [API Compatibility Macros](https://support.hdfgroup.org/documentation/hdf5/latest/api-compat-macros.html#title5) for more information on this option. |
```
lrknox: I suppose 2.0.0 API is ok, since the next opportunity for versioning functions should not come until version 3.  v200 (2.0.0 API) became awkward to think about because when the API series of major versions went from 1.14 to 2 for semantic versioning, the v symbol for major version 2 was made v200 so that v200 would be the (numerically) larger version compared to v110, v112 and v114.  Note that version 2.1.x has already been released, 2.2.0 will come next, and v200 will probably apply to all versions 2.x.  I assume that since the minor version number changes when functions are added, version 2.2 will be  2.2.0 API?

jhendersonHDF: Version 2.2 should still be 2.0.0 API, but I agree this is a bit awkward to specify as the API technically may change in a minor release. For the purpose of this variable it's still going to be the 2.0.0 API for all 2.x releases though. Maybe I should specify as 2.x instead of 2.0.0?

lrknox: Whichever is least likely to be confusing to users.  The option for 2.x will be HDF5_DEFAULT_API_VERSION=v200, and in future v300, v400, etc.  PR #6280 as described will set any functions versioned after the specified default to version 1.  Otherwise functions will be as defined in the current API and if any functions from the default API have since been deprecated/removed they will be unavailable.  I think 2.x is slightly more accurate;  If you think it is less likely to be confusing I would use it.

release_docs/INSTALL_CMake_options.md
```diff
@@ -0,0 +1,453 @@
...
+|:-------------|:-----|:--------|:------------|
+| `HDF5_BUILD_CPP_LIB` | `BOOL` | `OFF` | If `ON`, enables building of the HDF5 C++ wrapper interface. |
+
+### High-level C library options
+<a name="hl_lib"></a>
+
+| CMake option | Type | Default | Description |
+|:-------------|:-----|:--------|:------------|
+| `HDF5_BUILD_HL_LIB` | `BOOL` | `ON` | If `ON`, enables building of the HDF5 high-level C wrapper interface. |
+| `HDF5_DIMENSION_SCALES_NEW_REF` | `BOOL` | `OFF` | If `ON`, the high-level C dimension scale API (H5DS) will use a newer format for reference objects which is incompatible with versions of HDF5 before 1.12.0. |
```
lrknox: How incompatible, and incompatible how?  Does incompatible mean if the option is 'ON' it will produce dimension scales in .h5 files that won't work when read by HDF5 versions before 1.12, or that won't show dimension scales from files produced by HDF 5 older that 1.12, or something else?

jhendersonHDF: The former here (the references will be unreadable). I'm open to suggestions for better wording to make that more clear.

release_docs/INSTALL_CMake_options.md
```diff
@@ -0,0 +1,453 @@
...
+| `MPI_C_COMPILER` | `FILEPATH` | Varies | Path to the program to be used for compiling MPI C source code. Usually auto-detected by CMake. |
+| `MPIEXEC_EXECUTABLE` | `FILEPATH` | Varies | Path to the program to be used for running MPI programs. Usually auto-detected by CMake. |
+| `MPIEXEC_MAX_NUMPROCS` | `STRING` | Varies | Maximum number of MPI processes to use when running MPI programs. Usually auto-detected by CMake. |
+| `MPIEXEC_NUMPROC_FLAG` | `STRING` | Varies | Command-line flag to pass to `MPIEXEC_EXECUTABLE` for specifying the maximum number of MPI processes. Usually auto-detected by CMake. |
+| `MPIEXEC_PREFLAGS` | `STRING` | `""` (empty string) | Additional user-defined flags to pass to `MPIEXEC_EXECUTABLE` when running MPI programs. Any flags specified for this option will come directly before the name of the executable being run. |
+| `MPIEXEC_POSTFLAGS` | `STRING` | `""` (empty string) | Additional user-defined flags to pass to `MPIEXEC_EXECUTABLE` when running MPI programs. Any flags specified for this option will come after all flags passed to `MPIEXEC_EXECUTABLE`. |
```
lrknox: Do they also come after the name of the executable being run?  If so, is that implied?

jhendersonHDF: Yes, the command that gets run is roughly:

```
${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MAX_NUMPROCS} ${MPIEXEC_PREFLAGS} ${EXECUTABLE} ${MPIEXEC_POSTFLAGS}
```

So `MPIEXEC_POSTFLAGS` is meant to be things passed to the executable, but we could make that more clear.

release_docs/INSTALL_CMake_options.md
lrknox: If `ON`, the high-level C dimension scale API (H5DS) will use a newer format for reference objects.  The dimension scales created in data files produced cannot be read by HDF5 library versions below HDF5 1.12.0.
?

(I assume the file can be read except for the dimension scales, though if you are producing datasets with dimension scales you probably want the dimension scales to be readable.  When JPSS files came out they had datasets with data in an All_Data group and datsets with object and region references to those datasets in All_Data in a Data_Products group.  NetCDF at the time didn't handle datasets with references, and wouldn't open the file at all when it encountered a dataset that it couldn't handle.)

release_docs/INSTALL_CMake_options.md
lrknox: Maybe - 
Additional user-defined flags to pass to `MPIEXEC_EXECUTABLE` to pass as arguments for the MPI programs that it runs. Any flags specified for this option will come directly after the name of the executable being run.

lrknox (approved):

brtnfld (approved):

PR #6298, opened 2026-03-17
Update paths-ignore for 'push' events in workflows

jhendersonHDF:


.github/workflows/call-workflows.yml
```diff
@@ -4,6 +4,14 @@ name: hdf5 dev cmake CI
 on:
   workflow_dispatch:
   push:
+    paths-ignore:
+      - '.github/CODEOWNERS'
```
jhendersonHDF, line 8: `paths-ignore:` should be specified for both `push:` and `pull_request:` events. Otherwise, CI that should be ignored runs after commits are merged to develop and can slow down other CI.

lrknox (approved):

brtnfld:
Were the other workflows correct, or not an issue?

jhendersonHDF:
> Were the other workflows correct, or not an issue?

I just did a quick search for paths-ignore, but may have missed some workflows. Or those workflows didn't have a paths-ignore or used a different form of the field.

brtnfld (approved):

PR #6299, opened 2026-03-18
Fix some filters testing issues

jhendersonHDF:


.github/workflows/cygwin.yml
```diff
@@ -103,6 +103,7 @@ jobs:
             set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_ENABLE_ZLIB_SUPPORT:BOOL=ON")
             set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_ENABLE_SZIP_SUPPORT:BOOL=ON")
             set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_ENABLE_PLUGIN_SUPPORT:BOOL=ON")
+            set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DPLUGIN_USE_EXTERNAL:BOOL=ON")
```
jhendersonHDF, line 106: The logic in CMakePlugins.cmake was previously forcing `PLUGIN_USE_EXTERNAL` to `ON` if `HDF5_ALLOW_EXTERNAL_SUPPORT` was `GIT` or `TGZ`. In https://github.com/HDFGroup/hdf5/pull/6287 this was fixed in the same way as https://github.com/HDFGroup/hdf5/commit/5de837660fd5726d8dc3b1b0a7eb456ce8e5640d since it could be problematic as well. So several workflows now have to explicitly set `PLUGIN_USE_EXTERNAL` to `ON` instead of assuming it.

lrknox (approved):

brtnfld (approved):

PR #6302, opened 2026-03-18
Security hardening

brtnfld:
Addresses CodeQL issues (Critical and High).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -85,6 +85,12 @@ jobs:
           #   These are bounded by snprintf, guarded by H5_WARN_FORMAT_NONLITERAL_OFF,
           #   and the format strings originate from application code, not untrusted input.
           #
+          # - cpp/path-injection: All flagged sites are CLI tools (h5dump, h5repack,
+          #   h5import, h5jam, h5ls, h5perf) that take file paths from command-line
+          #   arguments or documented environment variables. Accepting user-provided
+          #   paths is the core purpose of these tools. The two core library hits
+          #   (H5FDsubfiling, H5FDioc) use documented env vars for VFD configuration.
+          #
           # Code style rules (non-security, no severity rating):
           # These are structural patterns inherent to HDF5's C codebase and
           # do not represent security vulnerabilities.
@@ -110,6 +116,8 @@ jobs:
                    id: cpp/uncontrolled-format-string
                - exclude:
                    id: cpp/tainted-format-string
+               - exclude:
+                   id: cpp/path-injection
                - exclude:
                    id: cpp/long-switch
                - exclude:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -90,6 +90,16 @@ jobs:
           #   arguments or documented environment variables. Accepting user-provided
           #   paths is the core purpose of these tools. The two core library hits
           #   (H5FDsubfiling, H5FDioc) use documented env vars for VFD configuration.
+          # - cpp/type-confusion: HDF5 object header messages use a void* dispatch
+          #   pattern where H5O_shared_t is embedded as the first member of message
+          #   structs (C "base class" idiom). Runtime type guards prevent invalid casts.
+          #   H5TSpool thread arg is always passed as the correct type.
+          # - cpp/uncontrolled-allocation-size: CLI tools and benchmarks intentionally
+          #   accept allocation sizes from command-line arguments (block size, dataset
+          #   dimensions). This is expected for performance testing and file utilities.
+          # - cpp/world-writable-file-creation: CLI tools create files using fopen()
+          #   which defaults to 0666 & ~umask. File permissions are the user's
+          #   responsibility via umask, consistent with standard POSIX utilities.
           #
           # Code style rules (non-security, no severity rating):
           # These are structural patterns inherent to HDF5's C codebase and
@@ -118,6 +128,12 @@ jobs:
                    id: cpp/tainted-format-string
                - exclude:
                    id: cpp/path-injection
+               - exclude:
+                   id: cpp/type-confusion
+               - exclude:
+                   id: cpp/uncontrolled-allocation-size
+               - exclude:
+                   id: cpp/world-writable-file-creation
                - exclude:
                    id: cpp/long-switch
                - exclude:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -73,9 +73,15 @@ jobs:
           
           # Excluded rules justification:
           #
+          # ── Global exclusions (legitimate across the entire codebase) ──
+          #
           # Security false positives:
           # - cpp/toctou-race-condition: HDF5 file operations inherently involve
           #   check-then-act patterns that cannot be avoided in a file I/O library.
+          # - cpp/type-confusion: HDF5 object header messages use a void* dispatch
+          #   pattern where H5O_shared_t is embedded as the first member of message
+          #   structs (C "base class" idiom). Runtime type guards prevent invalid casts.
+          #   H5TSpool thread arg is always passed as the correct type.
           # - cpp/non-constant-format, cpp/uncontrolled-format-string,
           #   cpp/tainted-format-string: All flagged sites are intentional
           #   format-string-as-template patterns:
@@ -84,22 +90,6 @@ jobs:
           #   H5FDmulti.c uses API-configured member name templates with %s.
           #   These are bounded by snprintf, guarded by H5_WARN_FORMAT_NONLITERAL_OFF,
           #   and the format strings originate from application code, not untrusted input.
-          #
-          # - cpp/path-injection: All flagged sites are CLI tools (h5dump, h5repack,
-          #   h5import, h5jam, h5ls, h5perf) that take file paths from command-line
-          #   arguments or documented environment variables. Accepting user-provided
-          #   paths is the core purpose of these tools. The two core library hits
-          #   (H5FDsubfiling, H5FDioc) use documented env vars for VFD configuration.
-          # - cpp/type-confusion: HDF5 object header messages use a void* dispatch
-          #   pattern where H5O_shared_t is embedded as the first member of message
-          #   structs (C "base class" idiom). Runtime type guards prevent invalid casts.
-          #   H5TSpool thread arg is always passed as the correct type.
-          # - cpp/uncontrolled-allocation-size: CLI tools and benchmarks intentionally
-          #   accept allocation sizes from command-line arguments (block size, dataset
-          #   dimensions). This is expected for performance testing and file utilities.
-          # - cpp/world-writable-file-creation: CLI tools create files using fopen()
-          #   which defaults to 0666 & ~umask. File permissions are the user's
-          #   responsibility via umask, consistent with standard POSIX utilities.
           #
           # Code style rules (non-security, no severity rating):
           # These are structural patterns inherent to HDF5's C codebase and
@@ -114,12 +104,30 @@ jobs:
           #   reference during ongoing development.
           # - cpp/use-of-goto: HDF5 uses goto for centralized cleanup (HGOTO_ERROR /
           #   HGOTO_DONE macros), a well-established C resource management pattern.
+          #
+          # ── Tools-only exclusions (enforced for core library via filter-sarif) ──
+          #
+          # The following rules are excluded only for CLI tools and benchmarks
+          # via path-scoped filter-sarif patterns (see filter-sarif step below),
+          # but remain ENFORCED for the core library (src/, hl/src/):
+          #
+          # - cpp/path-injection: CLI tools (h5dump, h5repack, h5import, h5jam,
+          #   h5ls, h5perf) accept file paths from command-line arguments by design.
+          #   Kept enforced for src/ to surface env-var path usage in VFDs
+          #   (e.g., H5FDsubfiling, H5FDioc) for review.
+          # - cpp/uncontrolled-allocation-size: CLI tools and benchmarks intentionally
+          #   accept allocation sizes from command-line arguments (block size, dataset
+          #   dimensions). Kept enforced for src/ to catch library allocation issues.
+          # - cpp/world-writable-file-creation: CLI tools create files using fopen()
+          #   which defaults to 0666 & ~umask. Kept enforced for src/ to review
+          #   library file creation (VFD logging, cache logging).
+          #
           config: |
              query-filters:
                - exclude:
                    id: cpp/toctou-race-condition
                - exclude:
-                   id: cpp/short-global-name
+                   id: cpp/type-confusion
                - exclude:
                    id: cpp/non-constant-format
                - exclude:
@@ -127,13 +135,7 @@ jobs:
                - exclude:
                    id: cpp/tainted-format-string
                - exclude:
-                   id: cpp/path-injection
-               - exclude:
-                   id: cpp/type-confusion
-               - exclude:
-                   id: cpp/uncontrolled-allocation-size
-               - exclude:
-                   id: cpp/world-writable-file-creation
+                   id: cpp/short-global-name
                - exclude:
                    id: cpp/long-switch
                - exclude:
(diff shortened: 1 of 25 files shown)


hyoklee (dismissed):

src/H5win32defs.h
```diff
@@ -60,6 +60,23 @@ struct timezone {
...
+    return gmtime_s(result, timep) == 0 ? result : NULL;
+}
+#define HDgmtime_r(T, R) H5_gmtime_r(T, R)
+
+static __inline struct tm *
+H5_localtime_r(const time_t *timep, struct tm *result)
+{
+    return localtime_s(result, timep) == 0 ? result : NULL;
+}
+#define HDlocaltime_r(T, R) H5_localtime_r(T, R)
```
jhendersonHDF, line 83: Why are these getting redefined? The `#ifndef HDlocaltime_r` in H5private.h should already be dealing with this.

tools/lib/h5diff.c
```diff
@@ -1037,8 +1037,8 @@ h5diff(const char *fname1, const char *fname2, const char *objname1, const char
             MPI_Abort(MPI_COMM_WORLD, 0);
         } /* end if */
 
-        strcpy(filenames[0], fname1);
```
jhendersonHDF, line 1040: Fine, but `strcpy` was perfectly fine to use here since the string length was already checked above

tools/src/h5dump/h5dump_xml.c
```diff
@@ -388,22 +388,22 @@ xml_dump_all_cb(hid_t group, const char *name, const H5L_info2_t *linfo, void H5
...
-                        t_link_path = (char *)malloc(strlen(prefix) + linfo->u.val_size + 1);
+                        char   linkxid[100];
+                        char   parentxid[100];
+                        char   targetxid[100];
+                        char  *t_prefix   = xml_escape_the_name(strcmp(prefix, "") ? prefix : "/");
+                        char  *t_name     = xml_escape_the_name(name);
+                        char  *t_targbuf  = xml_escape_the_name(targbuf);
+                        char  *t_obj_path = xml_escape_the_name(obj_path);
+                        char  *t_link_path;
+                        size_t t_link_path_len = strlen(prefix) + linfo->u.val_size + 2;
```
jhendersonHDF, line 400: It would probably be helpful to set the size next to the `malloc` call and leave a comment that the +2 is for "/" and NUL.

tools/src/h5import/h5import.c
```diff
@@ -131,7 +131,8 @@ main(int argc, char *argv[])
 
             case 1: /* counting input files */
                 if (opt->fcount < 29) {
-                    (void)strcpy(opt->infiles[opt->fcount].datafile, argv[i]);
+                    (void)snprintf(opt->infiles[opt->fcount].datafile,
```
jhendersonHDF: Use `MAX_PATH_NAME_LENGTH` as the size for all of these instead of `sizeof()` in case the type of `datafile` changes. Also probably remove the (void) casting and potentially add error checking

tools/src/h5import/h5import.c
```diff
@@ -160,7 +162,7 @@ main(int argc, char *argv[])
                     (void)fprintf(rawerrorstream, err10, argv[i]);
                     goto err;
                 }
-                (void)strcpy(opt->outfile, argv[i]);
+                (void)snprintf(opt->outfile, sizeof(opt->outfile), "%s", argv[i]);
```
jhendersonHDF: Instead of `sizeof()`, use a constant in case the type of `outfile` changes. It's currently hard-coded to 256 in h5import.h but should probably use a macro.

tools/src/h5import/h5import.c
```diff
@@ -3973,8 +3975,7 @@ getExternalFilename(struct Input *in, FILE *strm)
 
     temp_len           = strlen(temp);
     in->externFilename = (char *)malloc((temp_len + 1) * sizeof(char));
-    (void)strcpy(in->externFilename, temp);
```
jhendersonHDF, line 3976: Another case where converting to `snprintf` really isn't needed

tools/src/h5repack/h5repack_main.c
```diff
@@ -725,7 +725,7 @@ parse_command_line(int argc, const char *const *argv, pack_opt_t *options)
             case 'S': {
                 char strategy[MAX_NC_NAME];
 
-                strcpy(strategy, H5_optarg);
+                snprintf(strategy, sizeof(strategy), "%s", H5_optarg);
```
jhendersonHDF: Use `MAX_NC_NAME`, not `sizeof()`.

tools/src/h5repack/h5repack_parse.c
```diff
@@ -94,7 +94,7 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t
                 else
                     sobj[k + 1] = '\0';
 
-                strcpy(obj_list[n].obj, sobj);
+                snprintf(obj_list[n].obj, sizeof(obj_list[n].obj), "%s", sobj);
```
jhendersonHDF: `MAX_NC_NAME` for the size of these two instead of `sizeof(obj_list[n].obj)`

tools/src/h5repack/h5repack_main.c
```diff
@@ -680,7 +680,7 @@ parse_command_line(int argc, const char *const *argv, pack_opt_t *options)
                 else {
                     char msgType[10];
 
-                    strcpy(msgType, msgPtr + 1);
+                    snprintf(msgType, sizeof(msgType), "%s", msgPtr + 1);
```
jhendersonHDF, line 683: Ideally, use a constant for the size of `msgType` and for `snprintf`, but this is localized enough to not likely be a problem in the future

brtnfld:
Addressed all the comments, I will have to dismiss the CodeQL issues that will probably remain in the cases where the  length was already validated.

jhendersonHDF:
> Addressed all the comments, I will have to dismiss the CodeQL issues that will probably remain in the cases where the length was already validated.

It's fine to use `snprintf` instead of `strcpy` if CodeQL is going to complain about it, just noting that it's not really necessary.

hyoklee:
> Addressed all the comments, I will have to dismiss the CodeQL issues that will probably remain in the cases where the length was already validated.

Please do not dismiss CodeQL issues.

E.g., OpenBSD will warn whenever strcpy() is used:

```
`warning: strcpy() is almost always misused, please use strlcpy()`
```

hyoklee (dismissed):
Great effort!

tools/src/h5import/h5import.c
```diff
@@ -3973,8 +3981,7 @@ getExternalFilename(struct Input *in, FILE *strm)
...
-    in->externFilename[temp_len] = '\0';
```
jhendersonHDF, line 3977: If reverting back to `strcpy`, this should probably be put back

tools/src/h5perf/pio_engine.c
```diff
@@ -429,22 +429,26 @@ pio_create_filename(iotype iot, const char *base_name, char *fullname, size_t si
             fullname[size - 1] = '\0';
         }
 
-        if ((strlen(fullname) + strlen(base_name) + 1) < size) {
```
jhendersonHDF, line 432: This and the check in sio_engine seem a bit fragile since the size seems like it may be +1 or +2, depending.

src/H5FDcore.c
```diff
@@ -387,16 +387,17 @@ H5FD__core_write_to_bstore(H5FD_core_t *file, haddr_t addr, size_t size)
...
+            char time_str[32];
 
+            H5_get_localtime_str(time_str, sizeof(time_str));
             offset = HDlseek(file->fd, 0, SEEK_CUR);
 
             HGOTO_ERROR(H5E_IO, H5E_WRITEERROR, FAIL,
                         "write to backing store failed: time = %s, filename = '%s', file descriptor = %d, "
                         "errno = %d, error message = '%s', ptr = %p, total write size = %llu, bytes this "
                         "sub-write = %llu, bytes actually written = %llu, offset = %llu",
-                        ctime(&mytime), file->name, file->fd, myerrno, strerror(myerrno), (void *)ptr,
```
jhendersonHDF, line 399: Why all the changes to the time functions in this PR?

brtnfld: Use of potentially dangerous functions is marked as a critical issue; I can dismiss those instead.

brtnfld: I'm fine with reverting those changes and excluding them. Tools don't have to worry about thread-safty. Your call.

tools/src/h5perf/pio_engine.c
```diff
@@ -429,22 +429,27 @@ pio_create_filename(iotype iot, const char *base_name, char *fullname, size_t si
...
                      * Default to PREFIX's original prefix value. */
-                    strcpy(fullname, prefix);
+                    snprintf(fullname, size, "%s", prefix);
                 }
 
-            strcat(fullname, "/");
-            strcat(fullname, base_name);
+            /* Re-check length after potential prefix fallback */
+            cur_len = strlen(fullname);
+            if ((cur_len + strlen(base_name) + 2) >= size)
```
jhendersonHDF: Does this +2 include the NUL? If so, either it should be comparing with +1 or comparing `> size` as `== size` should be fine when the compared size includes the NUL. It's even simpler to just let `snprintf` print as many characters as possible and then check the result rather than trying to do potentially incorrect calculations here.

tools/src/h5perf/sio_engine.c
jhendersonHDF, line 1: The same comments for pio_engine apply here as well

src/H5FDcore.c
jhendersonHDF: If the time output format still looks reasonable I'm fine with it

tools/src/h5perf/sio_engine.c
brtnfld: updated

tools/src/h5perf/pio_engine.c
brtnfld: updated

hl/tools/h5watch/h5watch.c
```diff
@@ -728,13 +728,15 @@ parse_command_line(int argc, const char *const *argv)
...
                         leave(EXIT_FAILURE);
                     }
-                    if ((g_list_of_fields = (char *)realloc(g_list_of_fields, strlen(g_list_of_fields) +
-                                                                                  strlen(str) + 2)) == NULL) {
-                        error_msg("memory allocation failed (file %s:line %d)\n", __FILE__, __LINE__);
-                        leave(EXIT_FAILURE);
+                    {
+                        size_t cur_len = strlen(g_list_of_fields);
+                        size_t new_len = cur_len + strlen(FIELD_SEP) + strlen(str) + 1;
+                        if ((g_list_of_fields = (char *)realloc(g_list_of_fields, new_len)) == NULL) {
```
jhendersonHDF: Unrelated to this PR, but this realloc call should be made using a temporary variable assignment instead of assigning back to `g_list_of_fields`.

src/H5FDcore.c
brtnfld: The debug functions (H5Odbg, H5Omtime) use a fixed "%Y-%m-%d %H:%M:%S %Z" format which is locale-independent and more readable than the original. The VFD error paths use strftime with %c, which in the default C locale produces output equivalent to the original ctime minus the trailing newline.

hl/tools/h5watch/h5watch.c
brtnfld: rewritten, so addressing it now.

tools/src/h5ls/h5ls.c
```diff
@@ -2285,14 +2285,15 @@ list_obj(const char *name, const H5O_info2_t *oinfo, const char *first_seen, voi
...
+                struct tm *tm_result;
 
                 if (simple_output_g)
-                    tm = gmtime(&(oinfo->mtime));
+                    tm_result = HDgmtime_r(&(oinfo->mtime), &tm_buf);
                 else
-                    tm = localtime(&(oinfo->mtime));
-                if (tm) {
-                    strftime(buf, sizeof(buf), "%Y-%m-%d %H:%M:%S %Z", tm);
+                    tm_result = HDlocaltime_r(&(oinfo->mtime), &tm_buf);
```
jhendersonHDF, line 2294: I think it's worth noting that `gmtime_r` and `locatime_r` seem to only have come as recently as POSIX 2024 or C23. I think we may need to be careful about using these functions without having checked for them first.

.github/workflows/codeql.yml
```diff
@@ -110,6 +126,14 @@ jobs:
                    id: cpp/uncontrolled-format-string
                - exclude:
                    id: cpp/tainted-format-string
+               - exclude:
+                   id: cpp/path-injection
+               - exclude:
+                   id: cpp/type-confusion
+               - exclude:
+                   id: cpp/uncontrolled-allocation-size
```
jhendersonHDF: With this list of exclusions, I'm thinking we need to determine if we can use different configurations for different parts of HDF5. `cpp/uncontrolled-allocation-size` is something I'd be more inclined to ignore for a tool program, but less likely to ignore for the library itself.

brtnfld: # CodeQL Exclusion Split: Global vs Tools-Only

### Rules moved from global to tools-only (via path-scoped filter-sarif)

| Rule | Excluded For | Enforced For | Expected Library Hits |
|------|-------------|-------------|----------------------|
| `cpp/path-injection` | `tools/**`, `hl/tools/**` | `src/`, `hl/src/` | H5FDsubfiling, H5FDioc env-var path usage |
| `cpp/uncontrolled-allocation-size` | `tools/**`, `hl/tools/**` | `src/`, `hl/src/` | Dataset dimension-driven allocations (likely false positives) |
| `cpp/world-writable-file-creation` | `tools/**`, `hl/tools/**` | `src/`, `hl/src/` | H5FDlog, H5FDstdio, H5Clog_json, H5Clog_trace `fopen()` calls |

### Rules remaining as global exclusions

**Security false positives:**
- `cpp/toctou-race-condition` — inherent to file I/O library design
- `cpp/type-confusion` — C "base class" idiom (`H5O_shared_t` as first struct member)
- `cpp/non-constant-format`, `cpp/uncontrolled-format-string`, `cpp/tainted-format-string` — bounded by `snprintf`, format strings from API config (includes `H5FDmulti.c` in `src/`)

**Code style (non-security):**
- `cpp/short-global-name` — intentional `H5F_`/`H5D_` namespace convention
- `cpp/long-switch` — standard C dispatch pattern for type enums
- `cpp/guarded-free` — defensive NULL-guarded free convention
- `cpp/commented-out-code` — legacy reference code
- `cpp/use-of-goto` — centralized cleanup via `HGOTO_ERROR`/`HGOTO_DONE` macros

### Implementation

- **Global exclusions** are handled in the CodeQL `config:` block (`query-filters`)
- **Tools-only exclusions** are handled in the `filter-sarif` step with path-scoped patterns:
  ```yaml
  -tools/**:cpp/path-injection
  -tools/**:cpp/uncontrolled-allocation-size
  -tools/**:cpp/world-writable-file-creation
  -hl/tools/**:cpp/path-injection
  -hl/tools/**:cpp/uncontrolled-allocation-size
  -hl/tools/**:cpp/world-writable-file-creation
Test directories continue to be excluded entirely (**/test/**, **/testpar/**, **/tools/test/**)

tools/src/h5ls/h5ls.c
brtnfld: We could add this check to CMake if you think it would be important for platforms without POSIX 1-2001 or C23 support to fail gracefully

jhendersonHDF: I'm mostly wondering if it's relevant. When we added support for `qsort_r`, we didn't catch the issues on BSDs because we weren't testing them at the time. There may be platforms we're missing from our current testing. I'm just wary since POSIX 2024/C23 is very new generally speaking.

hyoklee (approved):

tools/src/h5ls/h5ls.c
brtnfld: Just to be clear,  POSIX.1-2001 (SUSv3) & Earlier: gmtime_r and localtime_r were included as part of the Thread-Safe Functions option in POSIX.1-2001 and were widely available in systems compliant with SuspV2. They have been universally available on Linux (via glibc), BSDs, and Unix systems for over 20 years. So not exactly the same as qsort_r, which was a signature issue.

We could fall back to non-thread-safe if not available.

jhendersonHDF: Ah that's my mistake for not reading further into the man page for these functions. No issue with usage of them then.

jhendersonHDF (approved):

hyoklee:
@brtnfld , 6 critical remaining: https://github.com/HDFGroup/hdf5/security/code-scanning

PR #6303, opened 2026-03-19
Fix unbounded writes in h5repack_opttable.c:

brtnfld:
Replace strcpy/strcat with snprintf

Replace all strcpy calls copying into fixed-size path buffers with snprintf to prevent buffer overflows.  Also simplify the path prefix check and the empty-string initialization.

Fixes 6 critical CodeQL cpp/unbounded-write alerts.

### DIFF
--- a/tools/src/h5repack/h5repack_opttable.c
+++ b/tools/src/h5repack/h5repack_opttable.c
@@ -104,12 +104,17 @@ aux_inctable(pack_opttbl_t *table, unsigned n_objs)
     int      ret_value = 0;
 
     table->size += n_objs;
-    table->objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
-    if (table->objs == NULL) {
-        H5TOOLS_INFO("not enough memory for options table");
-        ret_value = -1;
+    {
+        pack_info_t *new_objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
+        if (new_objs == NULL) {
+            H5TOOLS_INFO("not enough memory for options table");
+            ret_value = -1;
+        }
+        else {
+            table->objs = new_objs;
+        }
     }
-    else {
+    if (ret_value == 0) {
         for (u = table->nelems; u < table->size; u++)
             init_packobject(&table->objs[u]);
     }
@@ -193,6 +198,7 @@ options_add_layout(obj_list_t *obj_list, unsigned n_objs, pack_info_t *pack, pac
     if (table->nelems > 0) {
         /* go through the supplied list of names */
         for (j = 0; j < n_objs; j++) {
+            found = false;
             /* linear table search */
             for (i = 0; i < table->nelems; i++) {
                 /*already on the table */
@@ -272,6 +278,7 @@ options_add_filter(obj_list_t *obj_list, unsigned n_objs, filter_info_t filt, pa
     if (table->nelems > 0) {
         /* go through the supplied list of names */
         for (j = 0; j < n_objs; j++) {
+            found = false;
             /* linear table search */
             for (i = 0; i < table->nelems; i++) {
                 /*already on the table */
--- a/tools/src/h5repack/h5repack_opttable.c
+++ b/tools/src/h5repack/h5repack_opttable.c
@@ -100,26 +100,21 @@ aux_tblinsert_layout(pack_opttbl_t *table, unsigned int idx, pack_info_t *pack)
 static int
 aux_inctable(pack_opttbl_t *table, unsigned n_objs)
 {
-    unsigned u;
-    int      ret_value = 0;
+    unsigned     u;
+    pack_info_t *new_objs;
 
     table->size += n_objs;
-    {
-        pack_info_t *new_objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
-        if (new_objs == NULL) {
-            H5TOOLS_INFO("not enough memory for options table");
-            ret_value = -1;
-        }
-        else {
-            table->objs = new_objs;
-        }
+    new_objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
+    if (new_objs == NULL) {
+        H5TOOLS_INFO("not enough memory for options table");
+        return -1;
     }
-    if (ret_value == 0) {
-        for (u = table->nelems; u < table->size; u++)
-            init_packobject(&table->objs[u]);
-    }
+    table->objs = new_objs;
 
-    return ret_value;
+    for (u = table->nelems; u < table->size; u++)
+        init_packobject(&table->objs[u]);
+
+    return 0;
 }
 
 /*-------------------------------------------------------------------------
@@ -198,7 +193,6 @@ options_add_layout(obj_list_t *obj_list, unsigned n_objs, pack_info_t *pack, pac
     if (table->nelems > 0) {
         /* go through the supplied list of names */
         for (j = 0; j < n_objs; j++) {
-            found = false;
             /* linear table search */
             for (i = 0; i < table->nelems; i++) {
                 /*already on the table */
@@ -278,7 +272,6 @@ options_add_filter(obj_list_t *obj_list, unsigned n_objs, filter_info_t filt, pa
     if (table->nelems > 0) {
         /* go through the supplied list of names */
         for (j = 0; j < n_objs; j++) {
-            found = false;
             /* linear table search */
             for (i = 0; i < table->nelems; i++) {
                 /*already on the table */
--- a/tools/src/h5repack/h5repack_opttable.c
+++ b/tools/src/h5repack/h5repack_opttable.c
@@ -28,7 +28,7 @@ init_packobject(pack_info_t *obj)
 {
     int j, k;
 
-    strcpy(obj->path, "\0");
+    obj->path[0] = '\0';
     for (j = 0; j < H5_REPACK_MAX_NFILTERS; j++) {
         obj->filter[j].filtn     = -1;
         obj->filter[j].cd_nelmts = DEFAULT_CDELEMTS;
@@ -215,7 +215,7 @@ options_add_layout(obj_list_t *obj_list, unsigned n_objs, pack_info_t *pack, pac
                 /* keep the grow in a temp var */
                 idx = table->nelems + added;
                 added++;
-                strcpy(table->objs[idx].path, obj_list[j].obj);
+                snprintf(table->objs[idx].path, MAX_NC_NAME, "%s", obj_list[j].obj);
                 aux_tblinsert_layout(table, idx, pack);
             }
             /* cases where we have an already inserted name but there is a new name also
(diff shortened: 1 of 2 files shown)


brtnfld:
The ` File created without restricting permissions` is in the src; should we exclude those globally or dismiss them individually?

tools/src/h5repack/h5repack_opttable.c
```diff
@@ -338,12 +338,11 @@ options_get_object(const char *path, pack_opttbl_t *table)
 
     for (i = 0; i < table->nelems; i++) {
         /* make full path (start with "/") to compare correctly  */
-        if (strncmp(table->objs[i].path, "/", 1) != 0) {
-            strcpy(tbl_path, "/");
-            strcat(tbl_path, table->objs[i].path);
+        if (table->objs[i].path[0] != '/') {
+            snprintf(tbl_path, sizeof(tbl_path), "/%s", table->objs[i].path);
```
jhendersonHDF: Just use `MAX_NC_NAME + 1` for both of these. IMO, `sizeof()` is almost never a good choice to use for `snprintf()`.

jhendersonHDF:
> The ` File created without restricting permissions` is in the src; should we exclude those globally or dismiss them individually?

This may be a broader question of whether the library should be handling this by using a combination of `open()` and `fdopen()` in these places in order to set more restrictive permissions on the files.

jhendersonHDF:
I think it's probably best to fix the file permissions issues instead of dismissing them. I'll go ahead and do that.

tools/src/h5repack/h5repack_opttable.c
brtnfld: fixed

jhendersonHDF (dismissed):

hyoklee:
Claude.ai's review:

## Residual Issues (Not Fixed by PR)

### Finding 1 Off-by-one in `tbl_path` buffer

**Severity:** Low-Medium
**Location:** `options_get_object`, line ~337

```c
char tbl_path[MAX_NC_NAME + 1];  // 257 bytes unchanged by PR
snprintf(tbl_path, MAX_NC_NAME + 1, "/%s", table->objs[i].path);
```

`tbl_path` is 257 bytes. The `"/%s"` format with a 255-char path produces 255 + 1 (`/`) + 1 (NUL) = 257 bytes total but `snprintf` cap of 257 means the NUL lands at byte 257 leaving zero margin. If `path` is ever allowed to reach its declared capacity, this silently truncates.

**Fix:**
```c
char tbl_path[MAX_NC_NAME + 2];  // '/' + MAX_NC_NAME chars + NUL
snprintf(tbl_path, sizeof(tbl_path), "/%s", table->objs[i].path);
snprintf(tbl_path, sizeof(tbl_path), "%s",  table->objs[i].path);
```

---

### Finding 2 Hardcoded `MAX_NC_NAME` instead of `sizeof` in `snprintf` calls

**Severity:** Low
**Location:** 6 call sites in `options_add_layout` and `options_add_filter`

```c
snprintf(table->objs[idx].path, MAX_NC_NAME, "%s", obj_list[j].obj);
```

Using a hardcoded `MAX_NC_NAME` instead of `sizeof(table->objs[idx].path)` means the bound won't track future changes to the field declaration. `snprintf` truncation is also unchecked at all 6 sites.

**Fix:**
```c
snprintf(table->objs[idx].path, sizeof(table->objs[idx].path), "%s", obj_list[j].obj);
```

---

## Pre-existing Issues (Not in Scope of This PR)

### Finding 3  `realloc` misuse in `aux_inctable`

**Severity:** Medium
**Location:** `aux_inctable`, line ~107

```c
table->objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
if (table->objs == NULL) {
    H5TOOLS_INFO("not enough memory for options table");
    ret_value = -1;
}
```

If `realloc` returns `NULL`, the original pointer is overwritten and leaked. Any caller that ignores `ret_value` will then dereference NULL.

**Fix:**
```c
pack_info_t *new_objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
if (new_objs == NULL) {
    H5TOOLS_INFO("not enough memory for options table");
    ret_value = -1;
} else {
    table->objs = new_objs;
}
```

---

### Finding 4  `found` not reset between outer loop iterations

**Severity:** Low
**Location:** `options_add_layout` (~line 195), `options_add_filter` (~line 274)

`found` is initialized to `false` once before the outer `for (j = 0; j < n_objs; j++)` loop and never reset at the top of each iteration. After the first match sets `found = true`, all subsequent `j` iterations skip the `if (!found)` insertion branch, causing missed insertions and a stale-index read (`table->objs[i]` with `i` from the previous inner loop run).

**Fix:** Add `found = false;` as the first statement inside the outer `j` loop body.

---

## Summary

| # | Severity | Status | Description | Location |
|---|----------|--------|-------------|----------|
| 1 | Low-Medium | **Residual not fixed** | `tbl_path[MAX_NC_NAME+1]` too small for `"/%s"` + NUL when path is max length | `options_get_object`, line ~337 |
| 2 | Low | **Residual  not fixed** | `snprintf` uses hardcoded `MAX_NC_NAME` instead of `sizeof`; truncation unchecked | 6 sites in `options_add_layout`, `options_add_filter` |
| 3 | Medium | Pre-existing, out of scope | `realloc` result overwrites original pointer  memory leak on failure | `aux_inctable`, line ~107 |
| 4 | Low | Pre-existing, out of scope | `found` not reset in outer loop stale state, missed insertions, stale index read | `options_add_layout`/`options_add_filter` |

**Recommendation:** Address Finding 1 (change `tbl_path` to `MAX_NC_NAME + 2` and use `sizeof`) before merging. Findings 3 and 4 should be tracked as follow-up issues.

jhendersonHDF:
> Finding 1 Off-by-one in tbl_path buffer
> Finding 2 Hardcoded MAX_NC_NAME instead of sizeof in snprintf calls

For both of these, I'm very much against using `sizeof()` in `snprintf()` except in very rare cases. There has been more than one occasion where a `char[]` variable was converted to `char *` when it became dynamically-allocated and the `sizeof()` condition wasn't changed, leading to problems.

brtnfld:
Just addressed 3 and 4, 1 and 2 are already correct

tools/src/h5repack/h5repack_opttable.c
```diff
@@ -104,12 +104,17 @@ aux_inctable(pack_opttbl_t *table, unsigned n_objs)
     int      ret_value = 0;
 
     table->size += n_objs;
-    table->objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
-    if (table->objs == NULL) {
-        H5TOOLS_INFO("not enough memory for options table");
-        ret_value = -1;
+    {
+        pack_info_t *new_objs = (pack_info_t *)realloc(table->objs, table->size * sizeof(pack_info_t));
```
jhendersonHDF: A bit of a nitpick, but can you just move `pack_info_t *new_objs;` to the top of the function and avoid starting a new scope block with "{" here? It's just a bit harder to read and not really needed here.

tools/src/h5repack/h5repack_opttable.c
```diff
@@ -104,12 +104,17 @@ aux_inctable(pack_opttbl_t *table, unsigned n_objs)
...
+        if (new_objs == NULL) {
+            H5TOOLS_INFO("not enough memory for options table");
+            ret_value = -1;
+        }
+        else {
+            table->objs = new_objs;
+        }
     }
-    else {
+    if (ret_value == 0) {
```
jhendersonHDF: Instead of needing to check `ret_value` here it would be a lot simpler to just `H5TOOLS_INFO(...)` and then `return -1` if the realloc call fails

tools/src/h5repack/h5repack_opttable.c
```diff
@@ -272,6 +278,7 @@ options_add_filter(obj_list_t *obj_list, unsigned n_objs, filter_info_t filt, pa
     if (table->nelems > 0) {
         /* go through the supplied list of names */
         for (j = 0; j < n_objs; j++) {
+            found = false;
```
jhendersonHDF: I don't think this is correct. The logic looks like it was intending for `found` to remain true once set.

hyoklee:
Is this unit(code coverage) test by Claude invalid?

```
TESTING("    options_get_object needs MAX_NC_NAME+2 buffer (PR #6303 Finding 1)");
    {
        pack_opttbl_t *tbl2   = NULL;
        pack_info_t   *found2 = NULL;
        /* '/' + MAX_NC_NAME chars + NUL = MAX_NC_NAME + 2 bytes */
        char query_path2[MAX_NC_NAME + 2];

        if (options_table_init(&tbl2) < 0)
            GOERROR;

        /* Zero the entire first entry so filter[0].filtn's first byte (the
         * byte immediately after path[MAX_NC_NAME-1] in memory) is 0x00.
         * Then fill path with MAX_NC_NAME non-NUL bytes.  The string read by
         * strcat/snprintf through path is therefore exactly MAX_NC_NAME = 256
         * characters long. */
        memset(&tbl2->objs[0], 0, sizeof(pack_info_t));
        memset(tbl2->objs[0].path, 'a', MAX_NC_NAME);
        tbl2->nelems = 1;

        /* Absolute query path: '/' + MAX_NC_NAME 'a's + NUL */
        query_path2[0] = '/';
        memset(query_path2 + 1, 'a', MAX_NC_NAME);
        query_path2[MAX_NC_NAME + 1] = '\0';

        /* With tbl_path[MAX_NC_NAME + 1] = 257 bytes, strcat writes the NUL
         * to tbl_path[257]—one byte past the end—and strcmp fails (or
         * AddressSanitizer aborts).  With tbl_path[MAX_NC_NAME + 2] = 258
         * bytes the NUL lands at tbl_path[257], which is within bounds, and
         * strcmp succeeds. */
        found2 = options_get_object(query_path2, tbl2);
        options_table_free(tbl2);

        if (found2 == NULL)
            GOERROR;
    }
    PASSED();
  
```
Your PR fails on this test.

tools/src/h5repack/h5repack_opttable.c
```diff
@@ -193,6 +198,7 @@ options_add_layout(obj_list_t *obj_list, unsigned n_objs, pack_info_t *pack, pac
...
```
jhendersonHDF: The same here as below. I don't think the intention is for `found` to be reset once it's set

jhendersonHDF:
Since `tbl_path` in `options_get_object` is only `MAX_NC_NAME + 1` in size, it can't be compared against a fully-populated buffer of size `MAX_NC_NAME + 2` like that test is doing.

tools/src/h5repack/h5repack_opttable.c
brtnfld: you're correct, removed.

tools/src/h5repack/h5repack_opttable.c
brtnfld: updated

hyoklee:
```
 One caveat: h5repack_filters.c:60 has strcpy(objout->path, name) with no bounds check — that's a separate, real
  unbounded-write issue the PR doesn't address.
```

jhendersonHDF (approved):

hyoklee (approved):
Another great work! I hope it will make 100% for OpenSSF Best Practices.

PR #6305, opened 2026-03-19
Restrict permissions for some created files

jhendersonHDF:


src/H5Defl.c
```diff
@@ -431,7 +431,7 @@ H5D__efl_write(const H5O_efl_t *efl, const H5D_t *dset, haddr_t addr, size_t siz
             HGOTO_ERROR(H5E_EFL, H5E_OVERFLOW, FAIL, "external file address overflowed");
         if (H5_combine_path(dset->shared->extfile_prefix, efl->slot[u].name, &full_name) < 0)
             HGOTO_ERROR(H5E_EFL, H5E_NOSPACE, FAIL, "can't build external file name");
-        if ((fd = HDopen(full_name, O_CREAT | O_RDWR, H5_POSIX_CREATE_MODE_RW)) < 0) {
+        if ((fd = HDopen(full_name, O_CREAT | O_RDWR, H5_POSIX_CREATE_MODE_URWGROR)) < 0) {
```
jhendersonHDF: It should be noted that this could be an unexpected change on some systems. On my system the default `umask` is `0022`, which would mean the file would get created with the same permissions as with the previous `H5_POSIX_CREATE_MODE_RW` (`0666`), resulting in User r/w, Group r and Other r. With different `umask` values the permissions may have been different from this (even r/w for everyone), whereas now they should always be User r/w, Group r and Other r.

jhendersonHDF:
I believe https://github.com/HDFGroup/hdf5/security/code-scanning/2235 should be excluded, as the stdio VFD is meant to use standard I/O calls and is explicitly stated as not being meant for production use.

src/H5Defl.c
brtnfld: maybe we should leave it alone, the user should probably have control over the permissions

jhendersonHDF: I tend to agree, though we don't have any infrastructure for setting that currently and anything currently or previously created with `H5_POSIX_CREATE_MODE_RW` would use the default permissions of `0666` (everyone r/w) and then would be modified by the system's `umask`.

brtnfld: Right, I think that is what the user would expect. These seem different from log-type files. I would leave it and dismiss the alert.

jhendersonHDF: I think whether or not someone expects files they write to be writable by everyone by default will likely vary, but I'm happy to let people deal with this with `umask` for now.

brtnfld (approved):

lrknox (approved):

PR #5746, opened 2025-08-27
Fix CVE-2025-2915

glennsong09:
This PR fixes issue https://github.com/HDFGroup/hdf5/issues/5380, which has a heap based buffer overflow after H5MF_xfree is called on an address of 0 (file superblock). This PR changes an assert making sure addr isn't 0 to an if check.

The bug was first reproduced using the fuzzer and the POC file from https://github.com/HDFGroup/hdf5/issues/5380. With this change, the heap based buffer overflow no longer occurs.

PR #6311, opened 2026-03-20
Fix cache image testing issues and re-enable testing

jhendersonHDF:
Fix issue where chunked datasets could get setup with an incorrect chunking index type in parallel HDF5

Fix issue where metadata cache images with an undefined address and size of 0 couldn't be properly decoded

Fix issue where a flag in H5Cimage.c wasn't getting set correctly for release builds of the library, leading to incorrect error checking when reconstructing metadata cache entries

### DIFF
--- a/src/H5Cimage.c
+++ b/src/H5Cimage.c
@@ -2744,11 +2744,11 @@ H5C__reconstruct_cache_entry(const H5F_t *f, H5C_t *cache_ptr, hsize_t *buf_size
     flags = *p++;
     if (flags & H5C__MDCI_ENTRY_DIRTY_FLAG)
         is_dirty = true;
+    if (flags & H5C__MDCI_ENTRY_IS_FD_PARENT_FLAG)
+        is_fd_parent = true;
 #ifndef NDEBUG /* only used in assertions */
     if (flags & H5C__MDCI_ENTRY_IN_LRU_FLAG)
         in_lru = true;
-    if (flags & H5C__MDCI_ENTRY_IS_FD_PARENT_FLAG)
-        is_fd_parent = true;
     if (flags & H5C__MDCI_ENTRY_IS_FD_CHILD_FLAG)
         is_fd_child = true;
 #endif
--- a/src/H5Dint.c
+++ b/src/H5Dint.c
@@ -1326,19 +1326,6 @@ H5D__create(H5F_t *file, hid_t type_id, const H5S_t *space, hid_t dcpl_id, hid_t
     if (H5O_fill_set_version(file, &new_dset->shared->dcpl_cache.fill) < 0)
         HGOTO_ERROR(H5E_DATASET, H5E_CANTSET, NULL, "can't set latest version of fill value");
 
-    /* Check if the file driver would like to force early space allocation */
-    if (H5F_HAS_FEATURE(file, H5FD_FEAT_ALLOCATE_EARLY))
-        new_dset->shared->dcpl_cache.fill.alloc_time = H5D_ALLOC_TIME_EARLY;
-
-    /*
-     * Check if this dataset is going into a parallel file and set space allocation time.
-     * If the dataset has filters applied to it, writes to the dataset must be collective,
-     * so we don't need to force early space allocation. Otherwise, we force early space
-     * allocation to facilitate independent raw data operations.
-     */
-    if (H5F_HAS_FEATURE(file, H5FD_FEAT_HAS_MPI) && (new_dset->shared->dcpl_cache.pline.nused == 0))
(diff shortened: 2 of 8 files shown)


src/H5Cimage.c
```diff
@@ -2744,11 +2744,11 @@ H5C__reconstruct_cache_entry(const H5F_t *f, H5C_t *cache_ptr, hsize_t *buf_size
     flags = *p++;
     if (flags & H5C__MDCI_ENTRY_DIRTY_FLAG)
         is_dirty = true;
+    if (flags & H5C__MDCI_ENTRY_IS_FD_PARENT_FLAG)
+        is_fd_parent = true;
 #ifndef NDEBUG /* only used in assertions */
     if (flags & H5C__MDCI_ENTRY_IN_LRU_FLAG)
         in_lru = true;
-    if (flags & H5C__MDCI_ENTRY_IS_FD_PARENT_FLAG)
-        is_fd_parent = true;
```
jhendersonHDF, line 2751: After the changes in https://github.com/HDFGroup/hdf5/commit/3914bb7f7ec7105d8bfbeb3aebd92e867cff5b70, this flag needs setting all the time so that an error check below is performed correctly.

src/H5Dint.c
```diff
@@ -1326,19 +1326,6 @@ H5D__create(H5F_t *file, hid_t type_id, const H5S_t *space, hid_t dcpl_id, hid_t
     if (H5O_fill_set_version(file, &new_dset->shared->dcpl_cache.fill) < 0)
         HGOTO_ERROR(H5E_DATASET, H5E_CANTSET, NULL, "can't set latest version of fill value");
 
-    /* Check if the file driver would like to force early space allocation */
-    if (H5F_HAS_FEATURE(file, H5FD_FEAT_ALLOCATE_EARLY))
```
jhendersonHDF, line 1330: After the changes in https://github.com/HDFGroup/hdf5/commit/1093a18c181aa84458180693f3779e7b9cb1c0b6, setting this field needs to be moved down. Previously, the old function `H5D__layout_set_latest_indexing()` was being called before setting these fields, which correctly set a chunked dataset's layout to "fixed array" when appropriate. Since that now occurs in the `construct` callback for dataset layouts below, these fields being set early means that a chunked dataset with fixed dimensions and no filters applied was accidentally being set to use the "none" index type since early allocation was set. Early allocation needs to be set after the layout's chunk index type is set, but before `H5D__update_oh_info()` is called so that file space is allocated properly.

When the index type was set to "none", this had the effect that free space in the file for chunks wouldn't be freed and reclaimed when a dataset is deleted, leading to wasted file space and a failure in t_cache_image testing.

src/H5Ocache_image.c
```diff
@@ -116,12 +116,12 @@ H5O__mdci_decode(H5F_t *f, H5O_t H5_ATTR_UNUSED *open_oh, unsigned H5_ATTR_UNUSE
         HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "ran off end of input buffer while decoding");
     H5F_DECODE_LENGTH(f, p, mesg->size);
 
-    if (mesg->addr >= (HADDR_UNDEF - mesg->size))
-        HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "address plus size overflows");
-    if (mesg->addr == HADDR_UNDEF)
-        HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "address is undefined");
-    if ((mesg->addr + mesg->size) > H5F_get_eoa(f, H5FD_MEM_SUPER))
-        HGOTO_ERROR(H5E_OHDR, H5E_OVERFLOW, NULL, "address plus size exceeds file eoa");
+    if (H5_addr_defined(mesg->addr) && mesg->size) {
```
jhendersonHDF, line 119: The changes in https://github.com/HDFGroup/hdf5/pull/5746 didn't account for the case where a metadata cache image has an undefined address and a size of 0.

test/cache_image.c
```diff
@@ -6477,6 +6477,12 @@ cache_image_api_error_check_4(bool single_file_vfd)
             pass         = false;
             failure_mssg = "h5_fileaccess() failed.\n";
         }
+
+        if (H5Pset_libver_bounds(fapl_id, H5F_LIBVER_EARLIEST, H5F_LIBVER_V18) < 0) {
```
jhendersonHDF, line 6481: This test was previously assuming the default library version bounds setting. It needs to be set explicitly now.

lrknox (dismissed):

bmribler (approved):

mattjala (approved):

PR #6319, opened 2026-03-25
Fix several undefined behavior issues noted by UBsan

jhendersonHDF:
Disable float16 support for undefined sanitizer workflow for now as it causes a crash in UBSan


src/H5Aint.c
```diff
@@ -72,7 +72,7 @@ typedef struct {
 /* Local Prototypes */
 /********************/
 
-static herr_t H5A__close_cb(H5VL_object_t *attr_vol_obj, void **request);
+static herr_t H5A__close_cb(void *attr_vol_obj, void **request);
```
jhendersonHDF, line 75: There were several instances in the library where an `H5I_free_t` callback was declared with some type parameter instead of `void *` and then casted to `H5I_free_t`, causing UBSan to complain. The typing is lost for these callbacks, but several other `H5I_free_t` callbacks in the library already use the correct `void *` parameter anyway.

src/H5Dscatgath.c
```diff
@@ -1264,12 +1264,21 @@ H5D__scatgath_write_select(H5D_io_info_t *io_info)
...
                 !H5D__SCATGATH_USE_CMPD_OPT_WRITE(dset_info, io_info->sel_pieces[i]->in_place_tconv)) {
-                /* Non-const write_buf[i].  Use pointer math here to avoid const warnings.  When
-                 * there's a background buffer write_buf[i] always points inside the non-const tconv
-                 * buf so this is OK. */
-                void *tmp_write_buf =
-                    (void *)((uint8_t *)io_info->tconv_buf +
-                             ((const uint8_t *)write_bufs[i] - (const uint8_t *)io_info->tconv_buf));
+                void *tmp_write_buf;
+
+                if (io_info->sel_pieces[i]->in_place_tconv) {
```
jhendersonHDF, line 1269: @fortnern do these changes look correct? UBSan was complaining because in this particular case (where modifying of write buffers is enabled), the type conversion buffer wasn't allocated, so the previous logic was adding a non-zero offset to a `NULL` pointer. I'm not entirely certain how the testing passes currently given that, but it was definitely wrong. In the `in_place_tconv` case, it seemed like the logic above in the first loop already set each `write_bufs[i]` to point to the correct location in the write buffer. In fact, it also looks like each `write_bufs[i]` gets set to the correct place in `io_info->tconv_buf` above as well, so both cases might be able to be simplified to just the two lines below without needing all the casting in the `in_place_tconv == false` case.

.github/workflows/analysis.yml
```diff
@@ -414,6 +414,7 @@ jobs:
               set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_BUILD_JAVA:BOOL=OFF")
               set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_BUILD_CPP_LIB:BOOL=ON")
               set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_BUILD_FORTRAN:BOOL=OFF")
+              set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_ENABLE_NONSTANDARD_FEATURE_FLOAT16:BOOL=OFF")
```
jhendersonHDF, line 417: The `_Float16` type currently crashes UBSan, preventing further analysis in dt_arith.c.

src/H5EA.c
```diff
@@ -368,7 +368,7 @@ H5EA__lookup_elmt(const H5EA_t *ea, hsize_t idx, bool will_extend, unsigned thin
         *thing             = iblock;
         *thing_elmt_buf    = (uint8_t *)iblock->elmts;
         *thing_elmt_idx    = idx;
-        *thing_unprot_func = (H5EA__unprotect_func_t)H5EA__iblock_unprotect;
```
jhendersonHDF, line 371: Similar to `H5I_free_t`, these callbacks were declaring in an incompatible way with `H5EA__unprotect_func_t` and then casted into it, causing UBSan to complain.

src/H5Odtype.c
```diff
@@ -1490,7 +1490,7 @@ H5O__dtype_decode(H5F_t *f, H5O_t *open_oh, unsigned H5_ATTR_UNUSED mesg_flags,
 {
     bool           skip;
     H5T_t         *dt        = NULL;
-    const uint8_t *p_end     = p + p_size - 1;
```
jhendersonHDF, line 1493: There were a handful of cases in the library where this type of calculation was performed when `p_size` is `SIZE_MAX` (the case where we can't know how big the buffer is for checking during decoding). This would cause `p` to overflow and cause UBSan to complain. Not really an issue, but easy to fix by checking `p_size` before doing the addition.

src/H5Tref.c
```diff
@@ -641,11 +641,10 @@ H5T__ref_mem_write(H5VL_object_t *src_file, const void *src_buf, size_t src_size
                    H5VL_object_t H5_ATTR_UNUSED *dst_file, void *dst_buf,
                    size_t H5_ATTR_NDEBUG_UNUSED dst_size, void H5_ATTR_UNUSED *bg_buf)
 {
-    H5F_t          *src_f   = NULL;
-    hid_t           file_id = H5I_INVALID_HID;
-    H5R_ref_priv_t *dst_ref = (H5R_ref_priv_t *)dst_buf;
```
jhendersonHDF, line 646: This casting can form an unaligned pointer which could potentially be problematic and which UBSan complains about. As the only place it is actually used is in a `memcpy()` from a temporary, aligned local reference object, just `memcpy()` directly into `dst_buf` instead of trying to go through `dst_ref`.

src/H5Tref.c
```diff
@@ -655,7 +654,6 @@ H5T__ref_mem_write(H5VL_object_t *src_file, const void *src_buf, size_t src_size
     assert(src_size);
     assert(dst_buf);
     assert(dst_size == H5T_REF_MEM_SIZE);
-    HDcompile_assert(sizeof(*dst_ref) == sizeof(tmp_ref));
```
jhendersonHDF, line 658: This assertion isn't really needed because `*dst_ref` would always be `H5R_ref_priv_t` in the previous logic, and `tmp_ref` is a `H5R_ref_priv_t`, making this trivially true. For catching future size issues, using `dst_size` would probably be more appropriate, but that was a bit outside the scope of this fix.

brtnfld:
Not part of PR, but minor observation:

At H5T.c:2502, the error line reads:

HGOTO_ERROR(H5E_ATTR, H5E_CANTDEC, FAIL, "unable to free VOL object");
This should be H5E_DATATYPE , every other HGOTO_ERROR in that function uses H5E_DATATYPE.

brtnfld (dismissed):

brtnfld (approved):

mattjala (approved):

PR #6324, opened 2026-03-26
Fix data alignment requirements check in direct I/O VFD

jhendersonHDF:
During file creation, the direct I/O VFD only checked data alignment requirements for writes and assumed they were the same for reads. Checks are now performed separately for writes and reads. In order to avoid EINVAL errors, the VFD is also now slightly more conservative about when it tries to avoid performing extra work when data alignment isn't required.

### DIFF
--- a/src/H5FDdirect.c
+++ b/src/H5FDdirect.c
@@ -520,7 +520,7 @@ H5FD__direct_check_alignment_reqs(H5FD_direct_t *file, int o_flags)
 {
     HDoff_t orig_file_size = 0;
     ssize_t io_bytes       = 0;
-    bool    changed_size   = 0;
+    bool    changed_size   = false;
     void   *buf            = NULL;
     herr_t  ret_value      = SUCCEED;
 
--- a/src/H5FDdirect.c
+++ b/src/H5FDdirect.c
@@ -551,15 +551,16 @@ H5FD__direct_check_alignment_reqs(H5FD_direct_t *file, int o_flags)
                 HSYS_GOTO_ERROR(H5E_IO, H5E_SEEKERROR, FAIL, "unable to seek to proper position");
         }
 
+        /* Attempt what should be an unaligned write. If this fails and anything
+         * other than EINVAL is returned, make no assumptions about alignment
+         * requirements (default is to always ensure alignment). If it succeeds,
+         * assume alignment isn't required.
+         */
         io_bytes = HDwrite(file->fd, (void *)((uint8_t *)buf + 1), file->fa.fbsize - 1);
         if (io_bytes >= 0)
             file->fa.must_align_writes = false;
         else if (EINVAL == errno) {
             HDoff_t align_offset = 0;
-
-            /* If anything other than EINVAL was returned, make no assumptions
-             * about alignment requirements
-             */
 
             /* Seek to aligned offset past end of file if necessary */
             if ((size_t)orig_file_size % file->fa.mboundary != 0)
@@ -582,14 +583,16 @@ H5FD__direct_check_alignment_reqs(H5FD_direct_t *file, int o_flags)
     if (HDlseek(file->fd, 1, SEEK_SET) < 0)
         HSYS_GOTO_ERROR(H5E_IO, H5E_SEEKERROR, FAIL, "unable to seek to proper position");
 
-    /* Make no assumptions about alignment requirements when bytes_read == 0 */
+    /* Attempt what should be an unaligned read. If this fails and anything
+     * other than EINVAL is returned, make no assumptions about alignment
+     * requirements (default is to always ensure alignment). If it succeeds,
+     * assume alignment isn't required. Make no assumptions about alignment
+     * requirements when io_bytes == 0.
+     */
     io_bytes = HDread(file->fd, (void *)((uint8_t *)buf + 1), file->fa.fbsize - 1);
     if (io_bytes > 0)
         file->fa.must_align_reads = false;
     else if (io_bytes < 0 && EINVAL == errno) {
-        /* If anything other than EINVAL was returned, make no assumptions
-         * about alignment requirements
-         */
         if (HDlseek(file->fd, 0, SEEK_SET) < 0)
             HSYS_GOTO_ERROR(H5E_IO, H5E_SEEKERROR, FAIL, "unable to seek to proper position");
 
--- a/src/H5FDdirect.c
+++ b/src/H5FDdirect.c
@@ -40,10 +40,11 @@ hid_t H5FD_DIRECT_id_g = H5I_INVALID_HID;
 
 /* Driver-specific file access properties */
 typedef struct H5FD_direct_fapl_t {
-    size_t mboundary;  /* Memory boundary for alignment    */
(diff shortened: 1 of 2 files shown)


jhendersonHDF:
As of Linux 6.15, 6.16 or 6.17, the requirements for direct I/O appear to have been made a little more strict. The Ubuntu runners were updated in https://github.com/actions/runner-images/commit/015d15c63b9df61029d3795f181a2544a8b8d360 and the direct VFD has been failing with `EINVAL` because it was assuming the alignment requirements for writes were the same as for reads.

src/H5FDdirect.c
```diff
@@ -482,51 +485,8 @@ H5FD__direct_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxad
...
-    else {
-        if (HDread(file->fd, buf1, sizeof(int)) < 0) {
-            if (HDread(file->fd, buf2, file->fa.fbsize) < 0)
-                HGOTO_ERROR(H5E_FILE, H5E_READERROR, NULL, "file system may not support Direct I/O");
-            else
-                file->fa.must_align = true;
-        }
-        else {
-            if (o_flags & O_RDWR) {
-                if (HDlseek(file->fd, 0, SEEK_SET) < 0)
```
jhendersonHDF, line 514: This case was previously seeking to offset 0 and writing some bytes, which doesn't seem wise. The logic now always writes to the end of the file then truncates back to the original file size.

src/H5FDdirect.c
```diff
@@ -535,11 +495,126 @@ H5FD__direct_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxad
     if (ret_value == NULL) {
         if (fd >= 0)
             HDclose(fd);
+        if (file)
```
jhendersonHDF, line 498: Fix an infinite loop when file open fails

src/H5FDdirect.c
```diff
@@ -535,11 +495,126 @@ H5FD__direct_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxad
...
+ *            or I/O sizes for I/O. The `must_align_writes` and
+ *            `must_align_reads` of the passed in H5FD_direct_t structure
+ *            will be modified as appropriate.
+ *
+ * Return:    Non-negative on success / negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5FD__direct_check_alignment_reqs(H5FD_direct_t *file, int o_flags)
```
jhendersonHDF, line 519: This function keeps the same overall logic as before where writes are directly made to the file to check the alignment requirements, which isn't ideal. This is just a quick fix; a better fix would be to use the more modern `statx(2)` to check alignment requirements and only fall back to this function if necessary, and probably switch this function to use a temporary file located in the same directory being written to in order to avoid possible overwriting of existing bytes in the file.

src/H5FDdirect.c
```diff
@@ -805,7 +881,8 @@ H5FD__direct_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_U
      * read it directly from the file.  If not, read a bigger
      * and aligned data first, then copy the data into memory buffer.
      */
-    if (!_must_align || ((addr % _fbsize == 0) && (size % _fbsize == 0) && ((size_t)buf % _boundary == 0))) {
+    data_is_aligned = ((addr % _fbsize == 0) && (size % _fbsize == 0) && ((size_t)buf % _boundary == 0));
```
jhendersonHDF, line 887: Split the logic into a separate variable just for readability

lrknox (dismissed):

src/H5FDdirect.c
```diff
@@ -535,11 +495,126 @@ H5FD__direct_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxad
...
+             * file currently from seek above
+             */
+            if (HDlseek(file->fd, 1, SEEK_CUR) < 0)
+                HSYS_GOTO_ERROR(H5E_IO, H5E_SEEKERROR, FAIL, "unable to seek to proper position");
+        }
+
+        io_bytes = HDwrite(file->fd, (void *)((uint8_t *)buf + 1), file->fa.fbsize - 1);
+        if (io_bytes >= 0)
+            file->fa.must_align_writes = false;
+        else if (EINVAL == errno) {
```
mattjala, line 562: It looks like if the write operation fails for any reason other than `EINVAL`, that error would silently fall through.

jhendersonHDF: Yes, that's the intended purpose from before. If it failed for any reason other than `EINVAL` we aren't trying to guess why it failed. Just trying to probe whether an unaligned write fails due to `EINVAL` and a following aligned write succeeds, which would for the most part tell us that data alignment is required.

jhendersonHDF: I should add that by default alignment is required, so setting it to `true` in these cases isn't strictly necessary. The setting to `false` is what's more interesting. When this fails for non-`EINVAL`, alignment will still be required.

src/H5FDdirect.c
```diff
@@ -535,11 +495,126 @@ H5FD__direct_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxad
...
+ * Return:    Non-negative on success / negative on failure
+ *
+ *-------------------------------------------------------------------------
+ */
+static herr_t
+H5FD__direct_check_alignment_reqs(H5FD_direct_t *file, int o_flags)
+{
+    HDoff_t orig_file_size = 0;
+    ssize_t io_bytes       = 0;
+    bool    changed_size   = 0;
```
brtnfld: Should this be `false`

jhendersonHDF: Good catch, yes. Leftover from changed datatype

src/H5FDdirect.c
```diff
@@ -535,11 +495,126 @@ H5FD__direct_open(const char *name, unsigned flags, hid_t fapl_id, haddr_t maxad
...
+                HSYS_GOTO_ERROR(H5E_IO, H5E_SEEKERROR, FAIL, "unable to seek to proper position");
+        }
+
+        io_bytes = HDwrite(file->fd, (void *)((uint8_t *)buf + 1), file->fa.fbsize - 1);
+        if (io_bytes >= 0)
+            file->fa.must_align_writes = false;
+        else if (EINVAL == errno) {
+            HDoff_t align_offset = 0;
+
+            /* If anything other than EINVAL was returned, make no assumptions
```
brtnfld: Nit-pick,  wouldn't this be clearer if comment was moved to before the else if?

brtnfld (dismissed):

src/H5FDdirect.c
jhendersonHDF: I did that originally, but it felt a bit less readable to me than where it's at now and gets worse the longer the comment gets

jhendersonHDF: I could also just put it before the read/write calls

brtnfld (approved):

lrknox (approved):

PR #6326, opened 2026-03-27
Fix JS errors in doxygen navtree

### LINKED ISSUE #5612: HDF5 File Format Specification links are not copyable
The index to the [HDF5 File Format Specification](https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t3.html) on the right side of the page, the links take you to the correct section. 

But you cant cut and paste the link, for example into this document, so that someone can click on the link and get to that section.

For example https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t3.html#subsubsec_fmt3_dataobject_hdr_msg_layout takes you to the top of the document, instead of to the subsection IV.A.2.i. The Data Layout Message.

This is a pretty important functionality. Hopefully its an easy fix.

jhendersonHDF:
Fixes anchor links in some browsers and restores the navtree alongside the Table of Contents for several pages


jhendersonHDF:
This needs more testing before being merged to make sure the behavior is correct, but it should fix anchor links in some browsers and allow the navtree to exist alongside the created TOC.

Sample page before:
Docs Before

After:
Docs After

doxygen/hdf5_navtree_hacks.js
```diff
@@ -1,3 +1,524 @@
+/*
+ @licstart  The following is the entire license notice for the JavaScript code in this file.
```
jhendersonHDF: Most of the functions that were being overridden before are nested functions now, so we're effectively just modifying the original code. Included the license in this file.

However, the note below of "The above is the entire license notice for the JavaScript code in this file" is not necessarily the case for code that we added, so it may need to be put in a separate file.

doxygen/hdf5_navtree_hacks.js
jhendersonHDF, line 1: The vast majority of the changes in this file consist of just copying over navtree.js from a build directory and modifying just a few spots from the previous logic.

doxygen/hdf5_navtree_hacks.js
```diff
@@ -74,169 +580,47 @@ function resizeHeight()
...
-        node.expanded = false;
-      } else {
-        expandNode(o, node, false, false);
-      }
-    }
-    node.expandToggle.appendChild(imgNode);
-    domNode.appendChild(node.expandToggle);
+  // Each item in the nav tree is currently 30px. This may change, but
+  // this knowledge is needed to get a reasonable size for the TOC.
+  if (NAVTREE) {
```
jhendersonHDF: There were many cases where the height of the TOC was larger than the calculated window height, which would cause weirdness with resizing. Doxygen appears to give us navtreedata.js, which contains this variable to tell us how many items are in the navtree. Given that those items are 30px in height currently, this gives a reasonable limit on the size of the navtree. This is somewhat fragile, but the most practical way I could think of to determine how to size the navtree against the TOC.

doxygen/hdf5_navtree_hacks.js
```diff
@@ -74,169 +580,47 @@ function resizeHeight()
...
-      if(window.location.href.indexOf('unsupported')==-1)
+      if (window.location.href.indexOf('unsupported')==-1)
         expandNode(global_navtree_object, global_navtree_object.node.children[0].children[2], true, true);
       else
         expandNode(global_navtree_object, global_navtree_object.node.children[0].children[1], true, true);
 
       // Hide the root node "HDF5"
       $(document.getElementsByClassName('index.html')[0]).parent().parent().css({display:"none"});
 
+      resizeHeight();
```
jhendersonHDF: Needed to make sure things get resized at an appropriate time.

doxygen/hdf5_navtree_hacks.js
```diff
@@ -1,3 +1,524 @@
...
+    }
+    a.className = stripPath(link.replace('#',':'));
+    if (link.indexOf('#')!=-1) {
+      const aname = '#'+link.split('#')[1];
+      const srcPage = stripPath(pathName());
+      const targetPage = stripPath(link.split('#')[0]);
+      a.href = srcPage!=targetPage ? url : aname;
+      a.onclick = function() {
+        storeLink(link);
+        aPPar = $(a).parent().parent();
```
brtnfld: This creates an accidental global. Should be const aPPar = ....

brtnfld:
The top tabs already provide the global navigation (Getting Started, User Guide, Reference Manual, Cookbook, etc.). So the sidebar navtree is duplicating that.

jhendersonHDF:
> The top tabs already provide the global navigation (Getting Started, User Guide, Reference Manual, Cookbook, etc.). So the sidebar navtree is duplicating that.

Im happy to simply have the TOC take up all the space and overwrite the navtree, but then the question becomes what point is there in having the navtree?

doxygen/hdf5_navtree_hacks.js
glennsong09: Do you mean that we need to add an additional license due to us adding code, or that because we've modified parts of this code, the license should be moved? From reading the license information, it seems like the license holds even with modifications to the code.

jhendersonHDF: I think it's a bit ambiguous since previously we were just overriding the original code with our own functions without having directly modified the original code. It's possible that some of those modifications, such as the new functions we added, could have been covered under our BSD license, but the rest of the code was similar to the original code from what I could tell, so that may have been more appropriately covered under the missing MIT license.

Since then, the functions that we were overriding have been moved to being nested functions, meaning that overriding them won't work as intended and we now have to modify the original code. The same new functions we added could be covered under the MIT license or our BSD license.

jhendersonHDF:
Using the PAGE_OUTLINE_PANEL from Doxygen 14+ looks like below. Seems like this might be preferable to trying to hack our own together.

Screenshot_2026-04-07_14-25-22

PR #6330, opened 2026-03-29
Fix more undefined behavior issues

jhendersonHDF:


jhendersonHDF:
With these changes, the default Ubuntu sanitizer build only has a handful of issues left in dt_arith testing, which will be a bit more involved to fix correctly.

src/H5Ztrans.c
```diff
@@ -106,21 +106,27 @@ static void       H5Z__xform_reduce_tree(H5Z_node *tree);
...
                                                                                                              \
         if (((RESL).type == H5Z_XFORM_SYMBOL) && ((RESR).type != H5Z_XFORM_SYMBOL)) {                        \
-            TYPE  *p;                                                                                        \
-            double tree_val;                                                                                 \
+            uint8_t *p;                                                                                      \
+            double   tree_val;                                                                               \
                                                                                                              \
             tree_val =                                                                                       \
                 ((RESR).type == H5Z_XFORM_INTEGER ? (double)(RESR).value.int_val : (RESR).value.float_val);  \
-            p = (TYPE *)(RESL).value.dat_val;                                                                \
```
jhendersonHDF, line 114: Casting `p` from `void *` to `TYPE *` here can create misaligned pointers. Use `memcpy()` from/to `p` and intermediate `TYPE` variables to avoid this.

test/dsets.c
```diff
@@ -14357,7 +14359,7 @@ typedef struct scatter_info_t {
                 }
 
 static herr_t
-scatter_cb(void **src_buf /*out*/, size_t *src_buf_bytes_used /*out*/, void *_scatter_info)
```
jhendersonHDF, line 14360: The prototype for an `H5Dscatter()` callback uses `const void **`, which is trivial to do correctly here.

test/dt_arith.c
```diff
@@ -157,11 +157,11 @@ static int without_hardware_g = 0;
             }                                                                                                \
                                                                                                              \
             if (n < SRC_PREC - 2) {                                                                          \
-                value1 = (TYPE)(value1 << 1);                                                                \
+                value1 = (TYPE)((uint64_t)value1 << 1);                                                      \
```
jhendersonHDF, line 160: Perform shifts in unsigned types before casting back to avoid UBSan warnings about left shifting negative values.

test/dtransform.c
```diff
@@ -54,7 +54,7 @@ const float windchillFfloat[ROWS][COLS] = {
     {25.0F, 18.0F, 11.0F, 4.0F, -3.0F, -11.0F, -18.0F, -25.0F, -32.0F, -39.0F, -46.0F, -54.0F, -61.0F, -68.0F,
      -75.0F, -82.0F, -89.0F, -97.0F},
     {25.0F, 17.0F, 10.0F, 3.0F, -4.0F, -11.0F, -19.0F, -26.0F, -33.0F, -40.0F, -48.0F, -55.0F, -62.0F, -69.0F,
-     -76.0F, -84.0F, -91.0F, -98.0F}};
+     -76.0F, -84.0F, -91.0F, -97.0F}};
```
jhendersonHDF, line 57: Since intermediate data transformation operations are casted to the final type, `-98.0F` ends up just outside the range of `char` during a transformation operation. Cap the value to `-97.0F`.

test/hyperslab.c
```diff
@@ -762,12 +762,12 @@ test_transpose(size_t nx, size_t ny)
     size[0]       = nx;
     size[1]       = ny;
     src_stride[0] = 0;
-    src_stride[1] = sizeof(*src);
-    dst_stride[0] = (hsize_t)((1 - nx * ny) * sizeof(*src));
```
jhendersonHDF, line 766: This stride value, being negative, was being converted to a huge unsigned value when casted to `hsize_t`. As the intention is for the stride to be negative, use `hssize_t` values and `H5VM_stride_copy_s()` instead.

test/hyperslab.c
```diff
@@ -1071,8 +1073,13 @@ test_array_offset_n_calc(size_t n, size_t x, size_t y, size_t z)
         off = H5VM_array_offset(ARRAY_OFFSET_NDIMS, dims, coords);
 
         /* Check offset of coordinate */
-        if (a[off] != off)
+        if (a[off] != off) {
```
jhendersonHDF, line 1076: This is just for debugging a current CI failure

.github/workflows/analysis.yml
```diff
@@ -214,7 +213,6 @@ jobs:
               set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DHDF5_ENABLE_PLUGIN_SUPPORT:BOOL=OFF")
               set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DLIBAEC_USE_LOCALCONTENT:BOOL=OFF")
               set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DZLIB_USE_LOCALCONTENT:BOOL=OFF")
-              set (ADD_BUILD_OPTIONS "${ADD_BUILD_OPTIONS} -DPLUGIN_USE_LOCALCONTENT:BOOL=OFF")
```
jhendersonHDF, line 217: This CMake option isn't used in these builds and gives a configuration warning when set due to that

brtnfld (approved):

glennsong09 (approved):

PR #6331, opened 2026-03-30
Make JAR dep paths modifiable

mattjala:
The build system for the java libraries currently forces dependency JARs (e.g. SLF4j) to always be pulled from their bundled location in `java/lib`.

This PR changes several JAR path variables to CACHE variables, so that users interested in providing their own path to different JARs can do so via `-D` overrides. Specifically, the altered variables are
- HDF5_JAVA_LOGGING_JAR
- HDF5_JAVA_LOGGING_NOP_JAR
- HDF5_JAVA_LOGGING_SIMPLE_JAR

This PR also moves the JUnit and Hamcrest JAR paths to similarly modifiable cache variables (`HDF5_JAVA_JUNIT_JAR` and `HDF5_JAVA_HAMCREST_JAR`) instead of building them inline in a couple different places.


hyoklee (dismissed):

CMakeLists.txt
```diff
@@ -325,9 +325,11 @@ set (HDF5_JAVA_SRCJNI_PATH         ${HDF5_SOURCE_DIR}/java/src-jni)
 set (HDF5_JAVA_SRCJNI_JNI_SRC_DIR  ${HDF5_SOURCE_DIR}/java/src-jni/jni)
 set (HDF5_JAVA_LIB_DIR             ${HDF5_SOURCE_DIR}/java/lib)
 set (HDF5_JAVA_JSRC_DIR            ${HDF5_SOURCE_DIR}/java/jsrc)
-set (HDF5_JAVA_LOGGING_JAR         ${HDF5_SOURCE_DIR}/java/lib/slf4j-api-2.0.16.jar)
-set (HDF5_JAVA_LOGGING_NOP_JAR     ${HDF5_SOURCE_DIR}/java/lib/ext/slf4j-nop-2.0.16.jar)
-set (HDF5_JAVA_LOGGING_SIMPLE_JAR  ${HDF5_SOURCE_DIR}/java/lib/ext/slf4j-simple-2.0.16.jar)
+set (HDF5_JAVA_LOGGING_JAR         ${HDF5_SOURCE_DIR}/java/lib/slf4j-api-2.0.16.jar CACHE FILEPATH "Path to SLF4J API jar")
```
jhendersonHDF, line 328: If these are going to be changed to being cache variables, they should get set further down in the top-level Java CMakeLists.txt so that they only exist when Java is enabled. It should also be documented somewhere how they should be used (possibly in INSTALL_Cmake_options.md).

hyoklee (dismissed):

jhendersonHDF:
Two other comments:

- A change like this should probably have a CHANGELOG entry
- Copilot has a point in that `CMakeInstallation.cmake` is hard-coding these JAR names, which can result in the wrong paths being put/used in `hdf5-config.cmake`. See https://github.com/HDFGroup/hdf5/blob/develop/CMakeInstallation.cmake#L107-L118 and https://github.com/HDFGroup/hdf5/blob/develop/config/install/hdf5-config.cmake.in#L108-L125. There may be other cases being missed; we should quickly test (just manually) with alternative JAR paths to see what the resulting installation looks like and whether it works.

mattjala:
The most recent changes propagate the user-specified path changes down to `hdf5-config.cmake`. I also added the CHANGELOG, and tested a local build using a custom path to dependency JARs, which passed the java ctests.

glennsong09 (dismissed):

hyoklee (dismissed):
Proof: https://github.com/hyoklee/actions/actions/runs/23868895539/workflow

jhendersonHDF (dismissed):

jhendersonHDF:
Just needs conflict resolving

hyoklee (dismissed):

glennsong09 (dismissed):

jhendersonHDF (approved):

lrknox (approved):

PR #6332, opened 2026-03-30
Add test for H5VLget_wrap_ctx API

muhannad-ibrahim:
## Summary

This PR adds a standalone test (`test/vol_wrap_ctx.c`) that exercises the `H5VLget_wrap_ctx` API through a simulated VOL pass-through connector pattern.

The test validates the following:
1. Creates an HDF5 file and obtains a valid VOL connector ID and underlying object via `H5VLget_connector_id` and `H5VLobject`.
2. Increments the connector reference count with `H5Iinc_ref`, then calls `H5VLget_wrap_ctx` to retrieve the underlying wrap context.
3. Verifies the returned wrap context is valid and non-NULL.
4. Properly cleans up all resources (wrap context, VOL connector references, file handle) in both success and error paths.

## Coverage Impact

| Metric | Before | After | Change |
|---|---|---|---|
| Covered Lines | 343,738 | 343,758 | **+20** |

## Test Details

- **Target API:** `H5VLget_wrap_ctx`
- **Generated by:** Code Sa TestGeneration System

## How to build and run

```bash
h5cc -o vol_wrap_ctx test/vol_wrap_ctx.c
./vol_wrap_ctx
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -438,6 +438,7 @@ set (H5_TESTS
     cork
     swmr
     vol
+    vol_wrap_ctx
     timer
     cmpd_dtransform
     event_set
--- a/test/CMakePassthroughVOLTests.cmake
+++ b/test/CMakePassthroughVOLTests.cmake
@@ -75,6 +75,7 @@ add_custom_target (HDF5_VOLTEST_LIB_files ALL COMMENT "Copying files needed by H
       fheap
       big
       vol
+      vol_wrap_ctx
       error_test
       err_compat
       tcheck_version
--- a/test/vol_wrap_ctx.c
+++ b/test/vol_wrap_ctx.c
@@ -1,0 +1,310 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * Copyright by The HDF Group.                                               *
+ * All rights reserved.                                                      *
+ *                                                                           *
+ * This file is part of HDF5.  The full HDF5 copyright notice, including     *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the LICENSE file, which can be found at the root of the source code       *
+ * distribution tree, or in https://www.hdfgroup.org/licenses.               *
+ * If you do not have access to either file, you may request a copy from     *
+ * help@hdfgroup.org.                                                        *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+#include "hdf5.h"
+#include "H5Ipublic.h"
+#include "H5VLpublic.h"
+#include "H5VLconnector.h"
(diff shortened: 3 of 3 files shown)


jhendersonHDF:
Hi @muhannad-ibrahim,

tests should not be added as standalone source files; instead they should be added to CMakeLists.txt and CMakeTests.cmake as necessary. That said, `H5VLget_wrap_ctx()` is not generally meant to be called directly by applications and should be tested through use of a passthrough VOL connector, which is currently tested when testing the passthrough VOL connector in `src/H5VLpassthru.c`.

muhannad-ibrahim:
Hi @jhendersonHDF, thanks for the feedback, that's helpful context.

I looked into this further and wanted to share what I found. The passthrough VOL test matrix (CMakePassthroughVOLTests.cmake) is gated behind HDF5_TEST_PASSTHROUGH_VOL, which defaults to OFF in CMakeTests.cmake (line 86). In a standard build and ctest run (2999/3009 tests passing), no VOL- prefixed passthrough tests execute, which means the H5VLget_wrap_ctx() entry point in src/H5VLcallback.c (lines 811–843) — including argument validation, connector lookup via H5I_object_verify, and the callback dispatch — is not exercised.

Our coverage data confirms this: these lines show as uncovered in the default test configuration, and this test adds coverage for +20 of those lines.

That said, I understand that when HDF5_TEST_PASSTHROUGH_VOL=ON is enabled, the passthrough connector would exercise this path indirectly. So I have two questions:

> 1. Would it be valuable to have explicit coverage for this API in the default test configuration?
> 2. If so, would you prefer this be integrated into test/vol.c rather than as a separate file?

Happy to rework the approach based on your guidance.

jhendersonHDF:
Hi again @muhannad-ibrahim,

this PR is among others that we should get around to discussing internally soon, but I can go ahead and answer your questions:

> Would it be valuable to have explicit coverage for this API in the default test configuration?

I'd generally say yes, but also note that we could get coverage for the function by simply changing the default of the CMake variable `HDF5_TEST_PASSTHROUGH_VOL` to `ON`, which wouldn't be a bad idea anyway. It would be indirectly testing `H5VLget_wrap_ctx()` as you mentioned, but that API call really isn't intended to be called directly by an HDF5 application anyway, though it wouldn't really hurt to make sure it doesn't behave poorly when an application does call it.

In this test, the native VOL connector is being used by default and it has no `get_wrap_ctx` callback, so `H5VLget_wrap_ctx()` will effectively do nothing and always return a `NULL` wrap ctx pointer, so not much is being tested there. To get the test to return a real wrap ctx pointer, the default VOL connector the file FAPL would need to be changed.

> If so, would you prefer this be integrated into test/vol.c rather than as a separate file?

If explicit testing of the API is going to be added, then yes it would make the most sense to be added to `test/vol.c`.

Issue #6325, opened 2026-03-27
Error in the file format specification: Incorrect 'Page Bits' field size in Extensible Array Indexing Information


gheber:
The file format specification incorrectly describes the 'Page Bits' field size in Extensible Array Indexing Information as 2 bytes. The correct size is 1 byte.

The [RFC: File Format Changes in HDF5 1.10.0](https://support.hdfgroup.org/releases/hdf5/documentation/rfc/2015-09-28-RFC-HDF5-1.10.0-File-Format-Superblock-Changes-EP.docx.pdf) stated the size correctly as 5 bytes for all five fields of the layout. However, the error was introduced/is present in the [specification updates](https://support.hdfgroup.org/archive/support/HDF5/doc/H5.format.html) that shipped with HDF5 1.10.0.

The evidence from the code is clear: 

- `H5D_EARRAY_CREATE_PARAM_SIZE = 5` at `H5Dpkg.h:56`, and by the struct  definition in `H5Oprivate.h` where all five fields are `uint8_t`.
- The actual on-disk format uses 1 byte, consistent across encode (`H5Olayout.c:697`), decode (`H5Olayout.c:467`), and the size constant (`H5Dpkg.h:56`).

The error was first reported by Mr. Kazuyoshi Furutaka on the [HDF forum](https://forum.hdfgroup.org/t/the-number-of-bytes-used-for-page-bits-in-extensible-array-indexing-information/13725).

vchoi-hdfgroup:
I click on the link above "specification updates", the file format spec for Extensible Array Header shows 1 byte each.  I didn't see the error as indicated.

gheber:
The error is in section **IV.A.2.i. The Data Layout Message** of the file format specification. 

- Go to [https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t4.html](https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t4.html)
- Search for 'extensible array indexing'

You will see this:

<img width="1070" height="571" alt="Image" src="https://github.com/user-attachments/assets/2031118b-1c4b-4366-b0cf-1eb15d00f869" />

This suggests that the 'Page Bits' field is two bytes, which is wrong.

gheber:
Earlier versions have the same problem:
- [https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t3.html](https://support.hdfgroup.org/documentation/hdf5/latest/_f_m_t3.html) has the same problem, and so does
[...]

PR #6333, opened 2026-03-30
Fix for issue #6325: fix 'Page Bits' field size in Extensible Array I…

vchoi-hdfgroup:
…ndexing information

which is embedded within the data layout message.


gheber (approved):
Looks good!

jhendersonHDF (approved):

PR #6344, opened 2026-04-03
Java JNI library changes

opoplawski:
Fedora Java packaging guidelines https://docs.fedoraproject.org/en-US/packaging-guidelines/Java/#_guideline state that:
*  JNI shared objects MUST be placed in %{_libdir}/%{name}.
The first commit here allows changing the install location.

The second removes the library versioning from the JNI library.  It is not necessary and discouraged.  The Java loader will simply look for the `.so` file.  The JAR and JNI library are always paired so versioning is not relevant.


config/HDFMacros.cmake
```diff
@@ -435,6 +435,9 @@ macro (HDF_DIR_PATHS package_prefix)
     if (NOT ${package_prefix}_INSTALL_LIB_DIR)
       set (${package_prefix}_INSTALL_LIB_DIR ${CMAKE_INSTALL_LIBDIR})
     endif ()
+    if (NOT ${package_prefix}_INSTALL_JNI_LIB_DIR)
```
jhendersonHDF, line 438: I believe there are more cases that will need to be dealt with below, as this only sets `HDF5_INSTALL_JNI_LIB_DIR` when `HDF5_USE_GNU_DIRS=ON` (which is `OFF` by default). `HDF5_INSTALL_JNI_LIB_DIR` would be empty in most default cases and I believe CMake would just install to `lib` by default, which might not always be right.

It would also be great if an entry could be added to docs/INSTALL_CMake_options.md under the Installation options section so people could be aware of this option.

opoplawski: I think this should be resolved now.

jhendersonHDF (approved):
Thanks @opoplawski. Did some basic testing and this seems fine.

brtnfld (approved):

PR #6350, opened 2026-04-04
Fix integer overflow in array datatype nelem computation

### LINKED ISSUE #6349: Integer overflow in array datatype nelem computation (H5Odtype.c)
**Description**
  The loop in `H5O__dtype_decode_helper()` at `src/H5Odtype.c:811-814` computes `nelem` by multiplying array dimension sizes with no per-step overflow check. The post-loop overflow check at line 830 validates the already-wrapped value, making it ineffective.                                                                                                                                    

On 64-bit, three dimensions of 2^22 (4194304) each produce a true element count of 2^66, which wraps `size_t` to 4. The wrapped value passes both the overflow check and the size-mismatch check at line 838.

  The practical impact is limited, because `dt->shared->size` is consistently derived from the wrapped `nelem`, allocations and copies remain internally consistent. This is a silent data corruption issue (wrong element count) rather than a memory safety vulnerability.  Probably still worth fixing as the existing overflow check is ineffective.

  **Expected behavior**
  The decode should reject array datatypes whose dimension sizes overflow `size_t` when multiplied together, rather than silently wrapping to an incorrect element count.

  **Platform**
  - HDF5 version: 2.2.0
  - OS and version: Ubuntu 22.04 (Docker)
  - Compiler and version: GCC 14.3.0
  - Build system: CMake 3.28, Unix Makefiles
  - Any configure options: default (no special flags)
  - MPI: N/A (serial build)

  **Additional context**

  I have a fix ready on my fork: https://github.com/simons-hub/hdf5/tree/fix/array-nelem-overflow
  
  Will open a PR.

simons-hub:
Fixes #6349 

The loop in `H5O__dtype_decode_helper()` that computes `nelem` by multiplying array dimension sizes has no per-step overflow check. 

On 64-bit, three dimensions of 2^22 each wrap `nelem` from 2^66 to 4, bypassing the post-loop overflow check which validates the already-wrapped value. This produces silently incorrect element counts that propagate through type conversion, vlen iteration, and size calculations.

  ## Fix

Add a per-step overflow guard (`nelem > SIZE_MAX / dim[u]`) inside the multiplication loop at `src/H5Odtype.c:813` so the wrap is caught before it happens. 

The `dim[u] != 0` guard prevents division by zero — zero dimensions fall through to the existing size-mismatch check at line 838.

  ## Testing
  - Built with CMake (GCC 14.3, Ubuntu 22.04), no warnings from H5Odtype.c
  - Full test suite passes (2,842 tests, 0 failures)
  - CHANGELOG.md entry included

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -130,6 +130,10 @@ We would like to thank the many HDF5 community members who contributed to this r
 
    The direct I/O VFD attempts to determine data alignment requirements for a file on file open to try and avoid extra work when data alignment isn't required. Depending on the file access flags used when opening a file, the VFD could incorrectly determine these requirements for either writes or reads, eventually leading to a possible EINVAL return value on write or read. This has been fixed by separately determining the requirements for writes and reads and being more conservative about trying to avoid data alignment requirements.
 
+### Fixed integer overflow in array datatype element count computation
+
+   Fixed a bug in H5O__dtype_decode_helper() where the loop computing the total number of elements in an array datatype had no per-step overflow check. On 64-bit systems, large dimension sizes could cause the element count to wrap around, bypassing the post-loop overflow check and producing silently incorrect results in downstream type conversion and size calculations.
+
 ### Fixed an issue with chunked datasets using the wrong index type with parallel HDF5
 
    Fixed a bug in parallel HDF5 that would cause chunked datasets with fixed dimensions and without filters applied to use the "none" index type instead of the "fixed array" index type.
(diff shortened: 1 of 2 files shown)


simons-hub:
It seems likely the Clang format error was due to the HGOTO_ERROR log being short enough to fit into 1 line rather than 2.  Confirmed locally and just pushed the commit to log on 1 line.

jhendersonHDF (approved):

lrknox (approved):

PR #6354, opened 2026-04-06
Fix Java build ordering when building with parallel jobs

jhendersonHDF:


java/hdf/hdf5lib/CMakeLists.txt
```diff
@@ -233,6 +233,9 @@ if (HDF5_ENABLE_MAVEN_DEPLOY)
   # Export both JARs for installation
   SET_GLOBAL_VARIABLE (HDF5_JAVA_JARS_TO_EXPORT "${HDF5_JAVA_JARS_TO_EXPORT};${HDF5_JAVA_UNIVERSAL_JAR_FILE}")
   SET_GLOBAL_VARIABLE (HDF5_MAVEN_PLATFORM_JAR_FILE "${HDF5_JAVA_PLATFORM_JAR_FILE}")
+
+  # Set HDF5_JAVA_LIBRARY for examples to depend on
+  SET_GLOBAL_VARIABLE (HDF5_JAVA_LIBRARY ${HDF5_JAVA_UNIVERSAL_TARGET})
```
jhendersonHDF, line 238: When building the Java examples with `HDF5_ENABLE_MAVEN_DEPLOY` on, the custom command above to copy the platform-specific JAR to a new name needs to be completed first. Otherwise, build failures can occur since the Java classpath is set to include a JAR file by the "universal" name instead of the platform-specific name.

mattjala (approved):

config/cmake/HDF5ExampleCache.cmake
```diff
@@ -147,7 +147,7 @@ else ()
       set (H5EXAMPLE_JAVA_LIBRARY ${HDF5_LIBSH_TARGET})
     endif ()
     message (STATUS "HDF5 Example java lib: ${H5EXAMPLE_JAVA_LIBRARY} jars: ${HDF5_JAVA_INCLUDE_DIRS}")
-    set (H5EXAMPLE_JAVA_LIBRARIES ${HDF5_JAVA_HDF5_LIB_TARGET} ${H5EXAMPLE_JAVA_LIBRARY})
+    set (H5EXAMPLE_JAVA_LIBRARIES ${HDF5_JAVA_LIBRARY} ${H5EXAMPLE_JAVA_LIBRARY})
     set (HDF5_LIBRARY_PATH ${CMAKE_TEST_OUTPUT_DIRECTORY})
     message (STATUS "HDF5 Example java jar: ${HDF5_JAVA_HDF5_LIB_TARGET} path: ${CMAKE_TEST_OUTPUT_DIRECTORY}")
```
brtnfld, line 152: says ${HDF5_JAVA_HDF5_LIB_TARGET} while the actual dependency now uses ${HDF5_JAVA_LIBRARY}.

jhendersonHDF: This is correct really. They should eventually have the same name, it just depends on where at in the build process you are at the time.

brtnfld (approved):

PR #6356, opened 2026-04-07
Add hdf5-native Maven JARs and SciJava native-lib-loader integration

### LINKED ISSUE #6355: Publish hdf5-native Maven artifacts (bundled libhdf5 for classpath loading)
## Is your feature request related to a problem? Please describe.

HDF5 already publishes Java API JARs (`jarhdf5` / `hdf5-java-jni`, `hdf5-java-ffm`, etc.) and, for JNI, **`hdf5-jni-native`** so the JNI bridge can be delivered via Maven. Even so, running the bindings still requires a **separate, correctly versioned `libhdf5`** on the machine. Users often hit `UnsatisfiedLinkError` or linker issues until they install HDF5 or fix `LD_LIBRARY_PATH` / `PATH` / `DYLD_LIBRARY_PATH`. That blocks “Maven resolves what I need to run” in CI, containers, and other minimal environments, and encourages version skew between Java artifacts and the C library.

Example documentation still states that Maven does not supply system-level libraries—for instance [HDF5Examples/JAVA/README-MAVEN.md](HDF5Examples/JAVA/README-MAVEN.md) emphasizes manual native setup for `libhdf5`.

---

## Describe the solution you'd like

Add **`org.hdfgroup:hdf5-native`** (platform classifiers, e.g. `linux-x86_64`): a Maven JAR that bundles **`libhdf5`** built by the same CMake tree, using the [SciJava native-lib-loader](https://github.com/scijava/native-lib-loader) layout under `natives/<platform>/`.

CMake should produce `hdf5-native-*.jar` and matching POMs when Maven deployment is enabled; published POMs should integrate with the **existing** Java artifacts so resolution can place **`libhdf5`** on the classpath together with the API and (where used) **`hdf5-jni-native`**. Runtime loading should continue to use **`hdf.hdf5lib.Hdf5NativeLoader`** (or equivalent) so bundled `hdf5` loads before the JNI bridge when both are present, with escape hatches (`-Dhdf.hdf5lib.NativeLibraryBootstrap.skip=true`, existing `H5` properties) for system installs.

CI should **validate and stage/deploy** `hdf5-native` alongside existing JARs.

*This work is scoped to the new native bundle and wiring; it does not introduce `jarhdf5` or `hdf5-jni-native`—those remain as today.*

---

## Describe alternatives you've considered

1. **Status quo** — Keep requiring a system or custom-path `libhdf5` for all Maven consumers.
2. **Document heavier** — More README for OS packages only; does not fix headless/CI scenarios.
3. **Fat / multi-platform JAR** — One artifact for all platforms; awkward versus Maven classifiers and download size.
4. **Custom extraction** — Prefer native-lib-loader’s standard layout over bespoke code.

---

## Additional context

- **Licensing / redistribution:** Bundled `libhdf5` may depend on other vendors’ DLLs/`.so` files (e.g. zlib on some Windows builds); release notes should mention that when applicable ([docs/INSTALL_CMake.md](docs/INSTALL_CMake.md)).
- **Branch scope (for reviewers):** CMake targets and POMs for **`hdf5-native`**, loader/bootstrap integration with existing artifacts, native-lib-loader vendoring path if needed, test/CI classpaths, staging/deploy workflow updates for `hdf5-native-*.jar` and `pom-hdf5-native.xml` (exact paths as in the PR).

matteodg:
Fixes #6355 

I created this with Cursor, as my knowledge of CMake is basically zero: I tried to do it myself, but I was not able. I tested this in Linux x86_64 and Windows x86_64 and it is creating the proper jar and pom.xml file, that I can install in my local Maven repository and depend on them from an example application. This was there is not need of installing HDF5 library or setting up `LD_LIBRARY_PATH` anymore.

This will be useful for the future of HDFView as well.


brtnfld:
Must Fix:
FFM H5.java static block lacks try-catch around Hdf5NativeLoader call -- NoClassDefFoundError will permanently break H5 if the native-lib-loader JAR is missing

The FFM [H5.java:314] calls Hdf5NativeLoader.loadBundledNativesIfPresent(false) with no try-catch in the static block. If native-lib-loader-2.5.0.jar is missing from the classpath, loading Hdf5NativeLoader will throw NoClassDefFoundError (because it imports org.scijava.nativelib.NativeLoader), which will cause an ExceptionInInitializerError that permanently prevents H5 from being used in that JVM.

The JNI [H5.java:366-374] correctly wraps the call in catch (Throwable err), which handles this. The FFM version does not. The FFM path should have the same protection.

Nice to have:
- Partial-load when JNI bridge fails but hdf5 succeeds -- works correctly but worth a comment
- [docs] Transitive native deps (zlib, szip) not bundled -- "zero-install" doesn't hold if HDF5 was built with external filters

Moderate:
- Duplicate POM profiles across two pom.xml.in files -- The 5 OS-activated profiles (~70 lines each) in pom.xml.in and pom-jni.xml.in) are nearly identical (JNI version adds hdf5-jni-native dependency). This is a lot of duplicated XML that will need to be kept in sync. Not easily avoidable with Maven profiles though.
HDF5JavaNativeBundles.cmake is include()'d, not add_subdirectory() -- This means all its variables leak into the parent scope. The leading underscores help, but it's still a code smell. Consider using add_subdirectory() or wrapping in a function.

Low Severity

- Regex in validate-maven-artifacts.sh is getting unwieldy ([line 30]


`grep -qE "<artifactId>hdf5-java(-ffm|-jni)?</artifactId>|<artifactId>hdf5-native</artifactId>|<artifactId>hdf5-jni-native</artifactId>|<artifactId>hdf5-java-examples</artifactId>"`

Consider using a simpler pattern like <artifactId>hdf5-(java|native|jni-native) or an array-based check.

- find with -exec cp {} may clobber files -- In ctest.yml and maven-staging.yml, multiple POM files with different names (pom.xml, pom-hdf5-native.xml, etc.) are copied to a flat directory. If there are multiple pom.xml files from different build directories, later ones overwrite earlier ones. This was a pre-existing issue but is now slightly more likely to bite.

- The _np and _jp variable names in maven-deploy.yml ([lines 112-116] are cryptic. Minor, but NATIVE_POM / JNI_NATIVE_POM would be clearer.

- RELEASE_PROCESS.md change crams a lot of information into a single bullet point that's now very long and hard to parse.

Verify:

- [ ] Vendored JAR should be checksum-verified against Maven Central; license `acknowledged`

brtnfld (changes requested):
"Must Fix" needs to be addressed.

brtnfld:
-  Replace err.printStackTrace() with proper logging

In both java/hdf/hdf5lib/H5.java (FFM) and java/src-jni/hdf/hdf5lib/H5.java (JNI), replace:

```

catch (Throwable err) {
    err.printStackTrace();
}
```
with:
```

catch (Throwable err) {
    log.debug("Bundled HDF5 native library not loaded: " + err.getMessage());
}
```

- Remove the unused _HDF5JAVA_NATIVE_POM_DATE variable

In java/cmake/HDF5JavaNativeBundles.cmake:

```

string (TIMESTAMP _HDF5JAVA_NATIVE_POM_DATE "%Y-%m-%d %H:%M:%S UTC" UTC)
This is computed at configure time but never used anywhere in the file
```

Nice to have:

- Add trailing newline to both new POM template files

java/cmake/pom-native.xml.in and java/cmake/pom-jni-native.xml.in both end without a newline (\ No newline at end of file). Add a newline at the end of each.

- Clarify isHdf5LibraryLoaded dual semantics
java/src-jni/hdf/hdf5lib/H5.java

The problem: isHdf5LibraryLoaded is set in two unrelated places with different meanings:

In the bundled loader block — means "bundled hdf5 loaded"
At H5.H5dont_atexit() — means "hdf5 is confirmed loaded by any means"
A reader maintaining this code later will think the second assignment is a copy-paste error and delete it.

— split the flag's responsibilities):

Replace the two flags with three, each with a single clear meaning:

```

private static boolean isHdf5LibraryLoaded     = false; // bundled hdf5 load succeeded
private static boolean isHdf5JavaLibraryLoaded = false; // hdf5_java load succeeded
private static boolean isInitialized           = false; // full loadH5Lib() completed
```
Change the early-exit guard from:

```
if (isHdf5LibraryLoaded && isHdf5JavaLibraryLoaded)
    return;
```
to:

```
if (isInitialized)
    return;
```
Change the H5.H5dont_atexit() block from:

```
        try {
            H5.H5dont_atexit();
            isHdf5LibraryLoaded = true;
        }
        catch (HDF5LibraryException e) {
            System.exit(1);
        }
```
to:

```
        try {
            H5.H5dont_atexit();
            isInitialized = true;
        }
        catch (HDF5LibraryException e) {
            System.exit(1);
        }
```

This way isHdf5LibraryLoaded tracks only the bundled load result, isHdf5JavaLibraryLoaded tracks only the JNI bridge load result, and isInitialized guards against redundant re-entry.

brtnfld (changes requested):
See comments

matteodg:
I added to the branch the changes needed to create, not only the `hdf5-native` Maven artifact, but also the `hdf5-szip-native`, `hdf5-zlib-native`. So adding them as dependency on a Java app is sufficient to be able to manage .h5 files, including szip/zlib compressed datasets. 

Moreover I made optional the dependencies of the existing `hdf5-java-ffm`, `hdf5-java-jni` on the `hdf5*native` ones, so the user could always provide the native libraries via `LD_LIBRARY_PATH` and similar.

matteodg:
I updated to always use jextract 25, as hdf5 FFM Maven artifacts can only be built with JDK 25+.

matteodg:
I'm trying to fix the failing builds due to this branch.

matteodg:
This is now ready and it should be merged to develop after PR #6504 as that fixes some javadoc errors that prevent the builds to succeed.

Thanks in advance for any review and suggestion.

mkitti (commented):
The `Formatting Check` CI job is failing due to clang-format line-wrapping in the new `Hdf5NativeLoader.java` doc comments (unrelated to the javadoc/#6504 issue discussed elsewhere on this PR). Posted the two suggested reflows below — they match exactly what `clang-format17 --style=file` produces for this file.

java/hdf/hdf5lib/Hdf5NativeLoader.java
```diff
@@ -0,0 +1,173 @@
...
+import org.scijava.nativelib.NativeLoader;
+
+/**
+ * Loads bundled HDF5 shared libraries from {@code hdf5-native}, {@code hdf5-zlib-native},
+ * {@code hdf5-szip-native}, and {@code hdf5-jni-native} Maven JARs using SciJava native-lib-loader
+ * ({@code natives/&lt;platform&gt;/} layout). Extraction uses
+ * {@link org.scijava.nativelib.NativeLibraryUtil#loadNativeLibrary} so only classpath native JARs are consulted — not
+ * {@code java.library.path} (see {@link org.scijava.nativelib.NativeLoader#loadLibrary(String, String...)}). No
+ * Panama/JNI types are referenced here so this class can initialize before jextract-generated classes (FFM) or
+ * {@code System.loadLibrary("hdf5_java")} (JNI).
```
mkitti: Updating this suggestion: rather than letting clang-format's greedy fill wrap mid-clause, splitting into short sentences separated by `<p>` (matching the style already used in `HDFArray.java`/`package-info.java`) reads more naturally and is stable under `clang-format -style=file` (verified locally with clang-format 18).

```suggestion
 * Loads bundled HDF5 shared libraries from the {@code hdf5-native}, {@code hdf5-zlib-native},
 * {@code hdf5-szip-native}, and {@code hdf5-jni-native} Maven JARs using SciJava
 * native-lib-loader ({@code natives/&lt;platform&gt;/} layout).
 * <p>
 * Extraction uses {@link org.scijava.nativelib.NativeLibraryUtil#loadNativeLibrary}, so only
 * classpath native JARs are consulted, not {@code java.library.path} as
 * {@link org.scijava.nativelib.NativeLoader#loadLibrary(String, String...)} would do.
 * <p>
 * No Panama/JNI types are referenced here, so this class can initialize before
 * jextract-generated classes (FFM) or {@code System.loadLibrary("hdf5_java")} (JNI).
```

java/hdf/hdf5lib/Hdf5NativeLoader.java
```diff
@@ -0,0 +1,173 @@
...
+    private static boolean skipBundledLoad()
+    {
+        return "true".equalsIgnoreCase(System.getProperty(SKIP_PROPERTY, ""));
+    }
+
+    /**
+     * Load {@code libName} only from bundled native Maven JARs ({@code natives/&lt;platform&gt;/} on the
+     * classpath). Unlike {@link org.scijava.nativelib.NativeLoader#loadLibrary(String, String...)}, does not call
+     * {@code System.loadLibrary} first, so in-tree CI with {@code java.library.path} does not preload
+     * build-tree libs before {@code hdf5_java}.
```
mkitti: Same reasoning as the other comment — splitting into two sentences with a `<p>` break avoids the mid-clause "does not\ncall" split and stays clang-format-stable.

```suggestion
     * Load {@code libName} only from bundled native Maven JARs ({@code natives/&lt;platform&gt;/}
     * on the classpath).
     * <p>
     * Unlike {@link org.scijava.nativelib.NativeLoader#loadLibrary(String, String...)}, this does
     * not call {@code System.loadLibrary} first, so in-tree CI with {@code java.library.path} does
     * not preload build-tree libs before {@code hdf5_java}.
```

hyoklee (approved):

mkitti:
Via Copilot:

> CI is failing in Maven Staging Repository Test during CPack install, not compilation. The packaging step tries to install java/src-jni/hdf/hdf5lib/javadoc/hdf5_java_doc, but that file doesn’t exist, so file INSTALL aborts.

> Minimal fix: guard the Javadoc install rule with if(EXISTS ...) or ensure the Javadoc output is generated before packaging. This should unblock CPack without affecting the native JAR packaging.

```cmake
if (EXISTS "${CMAKE_CURRENT_BINARY_DIR}/javadoc/hdf5_java_doc")
  install(
    FILES "${CMAKE_CURRENT_BINARY_DIR}/javadoc/hdf5_java_doc"
    DESTINATION ${HDF5_INSTALL_JAR_DIR}
    COMPONENT maven
  )
endif()
```

I will take a look and see if I make the fix into a suggestion.

java/src-jni/hdf/hdf5lib/CMakeLists.txt
```diff
@@ -263,5 +268,10 @@ if (HDF5_ENABLE_FORMATTERS)
   clang_format (HDF5_JAVA_SRC_FORMAT ${HDF5_JAVA_HDF_HDF5_CALLBACKS_SOURCES} ${HDF5_JAVA_HDF_HDF5_EXCEPTIONS_SOURCES} ${HDF5_JAVA_HDF_HDF5_STRUCTS_SOURCES} ${HDF5_JAVA_HDF_HDF5_SOURCES})
 endif ()
 
+include (${HDF5_SOURCE_DIR}/java/cmake/HDF5JavaJavadoc.cmake)
+hdf5_java_add_javadoc_target (
+  "${HDF5_JAVA_LOGGING_JAR};${HDF5_JAVA_NATIVE_LIB_LOADER_JAR}"
+)
+
```
mkitti, line 275:
```suggestion
include (${HDF5_SOURCE_DIR}/java/cmake/HDF5JavaJavadoc.cmake)
hdf5_java_add_javadoc_target (
  "${HDF5_JAVA_LOGGING_JAR};${HDF5_JAVA_NATIVE_LIB_LOADER_JAR}"
)

if (EXISTS "${CMAKE_CURRENT_BINARY_DIR}/javadoc/hdf5_java_doc")
  install(
    FILES "${CMAKE_CURRENT_BINARY_DIR}/javadoc/hdf5_java_doc"
    DESTINATION ${HDF5_INSTALL_JAR_DIR}
    COMPONENT maven
  )
endif()
```

Issue #6357, opened 2026-04-10
Extend the H5P[get,set]_fapl_core API to specify an initial core buffer size (different from the increment)


gheber:
The current approach is too limited, and channels everything through the increment argument.

PR #6358, opened 2026-04-10
Let users specify an initial size for the core VFD backing buffer.

gheber:
This PR addresses #6357.


jhendersonHDF:
While making changes to the core VFD it may be worth considering fixing the issue noted in https://github.com/HDFGroup/hdf5/issues/3275. That said, these changes would have to wait for a 3.0.0 release.

lrknox:
> While making changes to the core VFD it may be worth considering fixing the issue noted in #3275. That said, these changes would have to wait for a 3.0.0 release.

Given that it needs to wait, should this be made a draft PR?

PR #6361, opened 2026-04-11
chore: fix typo in external link message

hyoklee:
`n` in `n external` is unnecessary.


lrknox:
> `n` in `n external` is unnecessary.

The 'n' makes the printed statement phonetically  correct:  "Link <name> is an external link."  I didn't write it, and I haven't seen the actual output, but it looks to me that it is likely correct with the 'n'.

gheber:
Since this is non-technical, even I can contribute. I'm with @lrknox. See [Indefinite Articles: A and An](https://www.grammarly.com/blog/grammar/indefinite-articles-a-and-an/).

It would have taken someone at my level of pedantry to write that example.

hyoklee:
I missed line 13.

PR #6367, opened 2026-04-15
ci: add big-endian (s390x) build and test workflow via QEMU

brtnfld:
## Summary

- Adds `.github/workflows/bigendian.yml`: a reusable `workflow_call` job that builds and tests HDF5 on Alpine Linux `s390x` (IBM Z, big-endian) using `jirutka/setup-alpine` with QEMU user-space emulation
- Wires it into `call-workflows.yml` as `call-release-cmake-bigendian` alongside the existing `i386` job
- Includes a sanity assertion (`python3 -c "assert sys.byteorder == 'big'"`) to fail fast if endianness is wrong

## Design notes

- Modeled directly after the existing `i386.yml` pattern — same action, same pinned hash, same CMake flags
- Fortran, Java, C++, ZLIB, SZIP, and plugin support are disabled to keep the emulated build lean; the goal is endianness correctness, not full feature coverage
- Parallelism capped at `--parallel 2` / `ctest --parallel 2` since s390x emulation is slower than native x86

## Test plan

- [x] Verify the `bigendian.yml` job appears in the Actions tab on a push/PR to `develop`
- [x] Confirm the "Verify big-endian" step prints `Confirmed big-endian`
- [ ] Confirm build and `ctest` pass without failures


brtnfld:
It takes too long to run for a PR check; it should be moved to a daily test.

.github/workflows/bigendian.yml
```diff
@@ -0,0 +1,68 @@
...
+          mkdir build
+          cd build
+          cmake -C ../config/cmake/cacheinit.cmake -G "Unix Makefiles" \
+          --log-level=VERBOSE \
+          -DCMAKE_BUILD_TYPE=${{ inputs.build_mode }} \
+          -DHDF5_ENABLE_ZLIB_SUPPORT:BOOL=OFF \
+          -DHDF5_ENABLE_SZIP_SUPPORT:BOOL=OFF \
+          -DHDF5_ENABLE_PLUGIN_SUPPORT:BOOL=OFF \
+          -DHDF5_BUILD_CPP_LIB:BOOL=OFF \
+          -DHDF5_BUILD_FORTRAN:BOOL=OFF \
```
hyoklee: Enable FORTRAN. It used to pass:

https://github.com/hdfeos/hdf5/blob/develop/WATCHME.md

hyoklee (changes requested):

hyoklee:
```
	2600 - FORTRAN_testhdf5_fortran (Failed)
```

hyoklee (dismissed):

hyoklee (approved):

hyoklee:
@brtnfld , can you patch this and include image test?

```
--- a/HDF5Examples/C/HL/tfiles/h5ex_image2BE.ddl
+++ b/HDF5Examples/C/HL/tfiles/h5ex_image2BE.ddl
@@ -33878,7 +33878,7 @@ GROUP "/" {
       }
    }
    DATASET "image8bit" {
-      DATATYPE  H5T_STD_U8LE
+      DATATYPE  H5T_STD_U8BE
       DATASPACE  SIMPLE { ( 400, 300 ) / ( 400, 300 ) }
       DATA {
       (0,0): 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
@@ -41397,7 +41397,7 @@ GROUP "/" {
       }
    }
    DATASET "palette" {
-      DATATYPE  H5T_STD_U8LE
+      DATATYPE  H5T_STD_U8BE
       DATASPACE  SIMPLE { ( 256, 3 ) / ( 256, 3 ) }
       DATA {
       (0,0): 0, 0, 255,
```

bmribler (approved):

hyoklee (changes requested):
2 more changes are desired:

1. See https://github.com/HDFGroup/hdf5/pull/6367#issuecomment-4263507001 .
2. Use `uraimo/run-on-arch-action` instead of `jirutka/setup-alpine`. Then, fortran test will also pass.

Gemini's explanation:

Recent issues in the jirutka/setup-alpine repository suggest that while s390x is supported, there have been sporadic "segmentation faults" and "Illegal instruction" errors tied to how it installs apk-tools during the initial setup on certain architectures.

Recommendation
If you need stability for s390x: Stick with uraimo/run-on-arch-action. Its Docker-based approach is better at isolating the emulation layer from the host's quirks.

If you need speed and only use x86/ARM: jirutka/setup-alpine is significantly faster because it doesn't have the overhead of Docker image layers, but it is more sensitive to the host environment.

brtnfld:
QEMU instability on GitHub-hosted runners seems to be an issue. Neither Alpine nor Ubuntu made a difference.

At this point, uraimo/run-on-arch-action is not reliable enough for compiling HDF5 on s390x via QEMU on these runners.

hyoklee:
I see a red again if you revert to alpine after 2 greens with `uraimo`.

brtnfld:
Looking into other options that are not emulation, closing for now.

PR #6371, opened 2026-04-16
HL: Fix prefix-match bugs in H5TB field lookup and H5DS/H5IM CLASS checks (#5633)

### LINKED ISSUE #5633: H5TBwrite_fields_name and H5TBread_fields_name can select the wrong field if one name is a prefix of another
### Describe the bug

When using `H5TBwrite_fields_name` and `H5TBread_fields_name`, inputting e.g. `field_names="somestr"` may end up operating on a field named `somestring_but_longer`.

This is due to a poorly written strncmp in `H5TB_find_field`.

### Suggested fix

https://github.com/HDFGroup/hdf5/blob/13706b8e93138bd3fc20b1641c1df49135cbf793/hl/src/H5TB.c#L3114-L3115

```diff
-     if (strncmp(start, field, strlen(field)) == 0)
+     if (strncmp(start, field, strlen(field) + 1) == 0)
          return true;
```

Either that or just change it back to `strcmp` because (a) using `strlen` inside `strncmp` is pretty smelly, and (b) it isn't actually preventing any out-of-bounds access anyways.  (if `field_names` has no null then *`strstr` will have already read out of bounds by the time we get here.*)

### Additional context

I looked into git blame and got quite an amusing history lesson.

* This bug has existed in the past!  Sort of.
* Because `field_names` is a comma-separated list, there are two separate places in `H5TB_find_field` that check if a field name matches.  The first is in the loop, for names followed by a comma. The second is for the last name, followed by a null.
* **[Once upon a time, many years ago](https://github.com/HDFGroup/hdf5/blob/dde381add4e4bd4e03812ce8a5b678185d6296b7/hl/src/H5TB.c#L3464)**, this bug was present in the first check, meaning it would occur **for all BUT the last name in `field_names`.**  E.g. if `field_names = "foo,bear";` then it could correctly distinguish `bear` from `beard`, but not `foo` from `foobar`.
* 19 years ago, in **[\[svn-r13546\]](https://github.com/HDFGroup/hdf5/commit/7aaf40e9f9c2d200f5a9c8bee646c0741b9cea49)**, this bug was fixed by adding a "length equals" check.  All was good.
* 11 years ago, in **[\[svn-r25614\]](https://github.com/HDFGroup/hdf5/commit/0324181acc#diff-e10ef6e8e88f08ad791552a68ce5ca301097e2c1cdf9e786d7fb10553fe5017dR3267)**, the bug was reintroduced, this time into the second check, by replacing `strcmp` with `strncmp`.  **Now, ONLY the last name in the list is affected.** :rofl:

### Aside

This is not the only strncmp in **[\[svn-r25614\]](https://github.com/HDFGroup/hdf5/commit/0324181acc#diff-e10ef6e8e88f08ad791552a68ce5ca301097e2c1cdf9e786d7fb10553fe5017dR3267)** that uses strlen without adding 1.  **There may be other manifestations of this same logic error!!!**  Might wanna give that commit a good look over.  Or seriously, just get rid of any `strn*` that uses `strlen` because, like, ......kinda defeats the point. 😛 

---

(most amusingly among the other instances of this code smell in that commit is a place where they let `n=MIN(strlen(a), strlen(b))`.  Which invokes undefined behavior in *strictly more cases* than the original `strcmp`. And has the prefix bug, in both directions.

I wonder if that's still--- yep it is

https://github.com/HDFGroup/hdf5/blob/13706b8e93138bd3fc20b1641c1df49135cbf793/hl/src/H5DS.c#L2291

brtnfld:
Fixes #5633 and related prefix-match bugs across the HL library.

H5TB

H5TB_find_field() compared only strlen(field) bytes of a tokenized user field name, so a user-supplied name like "PressureExtra" would match a real column "Pressure". Changed to strcmp against the NUL-terminated token parsed out of the comma-separated list. Applies to both H5TBread_fields_name() and H5TBwrite_fields_name(). The write path also silently accepted input with zero matching fields; it now returns an error, matching the read path's existing behavior.

H5DS

H5DSis_scale() and H5DS_is_reserved() used strncmp(buf, CLASS, MIN(strlen(CLASS), strlen(buf))), which matched any value whose prefix equaled the expected class name. Replaced with strcmp. Also:

Fixed a latent double-free in H5DSis_scale() on the no-match path, which was previously unreachable because the buggy compare always matched.
Added a variable-length string guard: a VLEN CLASS attribute passes the H5T_STR_NULLTERM check, but H5Aread into a char* buffer would write an hvl_t, so it is now explicitly rejected.
Allocate one extra byte and explicitly NUL-terminate after H5Aread as defense-in-depth.

H5IM

Same class of bug in H5IMis_image() (IMAGE class) and H5IMis_palette() (PALETTE class). Additionally:

Initialized did/aid/atid/attr_data at declaration so the out: error handler is safe from any failure point.
Added VLEN string guard (same rationale as H5DS).
Fixed pre-existing resource leaks in the out: handler: aid, atid, and attr_data were not closed/freed on mid-function failures.
Allocate one extra byte and explicitly NUL-terminate after H5Aread.

Regression tests

hl/test/test_table.c: read and write tests that request "PressureExtra" against a table with a real "Pressure" column and expect failure.
hl/test/test_image.c: test_class_prefix covers CLASS values "IMAGE_EXTRA", "I", "PALETTE_EXTRA", and "PAL" — all must return 0.
hl/test/test_ds.c: test_is_scale_class_prefix writes a 16-byte CLASS attribute holding "DIMENSION_S" (null-padded), the exact path H5DSis_scale() walks for the scale class length, and asserts 0. test_is_reserved_class_prefix exercises H5DS_is_reserved() indirectly via H5DSattach_scale() on a dataset tagged CLASS="IMAGE_EXTRA", which must not be treated as reserved.

All new tests were verified to fail against the unpatched sources and pass with the fix.

### DIFF
--- a/hl/src/H5DS.c
+++ b/hl/src/H5DS.c
@@ -2221,6 +2221,7 @@ H5DSis_scale(hid_t did)
     hid_t       aid = H5I_INVALID_HID; /* attribute ID */
     htri_t      attr_class;            /* has the "CLASS" attribute */
     htri_t      is_ds = -1;            /* set to "not a dimension scale" */
+    htri_t      isvl;                  /* result of H5Tis_variable_str */
     H5I_type_t  it;                    /* type of identifier */
     char       *buf = NULL;            /* buffer to read name of attribute */
     size_t      string_size;           /* size of storage for the attribute */
@@ -2261,54 +2262,71 @@ H5DSis_scale(hid_t did)
             is_ds = 0;
             goto out;
         }
-        /* check to make sure string is null-terminated;
-           if not, then it is not dimension scale */
-        if ((strpad = H5Tget_strpad(tid)) < 0)
+        if ((isvl = H5Tis_variable_str(tid)) < 0)
             goto out;
-        if (H5T_STR_NULLTERM != strpad) {
-            is_ds = 0;
-            goto out;
+
+        if (isvl > 0) {
+            /* Variable-length string: H5Aread allocates the string and writes
+             * its address into the buffer; use H5Treclaim to free it after use. */
+            char *vl_str = NULL;
+            hid_t asid   = H5I_INVALID_HID;
+
+            if ((asid = H5Aget_space(aid)) < 0)
+                goto out;
+            if (H5Aread(aid, tid, &vl_str) < 0) {
+                H5Sclose(asid);
+                goto out;
+            }
+            is_ds = (vl_str && strcmp(vl_str, DIMENSION_SCALE_CLASS) == 0) ? 1 : 0;
+            H5Treclaim(tid, asid, H5P_DEFAULT, &vl_str);
+            H5Sclose(asid);
         }
+        else {
+            /* check to make sure string is null-terminated;
+               if not, then it is not dimension scale */
+            if ((strpad = H5Tget_strpad(tid)) < 0)
+                goto out;
+            if (H5T_STR_NULLTERM != strpad) {
+                is_ds = 0;
+                goto out;
+            }
 
-        /* According to Spec string is ASCII and its size should be 16 to hold
-           "DIMENSION_SCALE" string */
-        if ((string_size = H5Tget_size(tid)) == 0)
-            goto out;
-        if (string_size != 16) {
-            is_ds = 0;
-            goto out;
+            /* According to Spec string is ASCII and its size should be 16 to hold
+               "DIMENSION_SCALE" string */
+            if ((string_size = H5Tget_size(tid)) == 0)
+                goto out;
+            if (string_size != 16) {
+                is_ds = 0;
+                goto out;
+            }
+
+            /* Allocate one extra byte and force null termination so strcmp never
+             * reads past the buffer even if some tool wrote the attribute without
+             * honoring the H5T_STR_NULLTERM invariant. */
+            buf = (char *)malloc((size_t)string_size * sizeof(char) + 1);
+            if (buf == NULL)
+                goto out;
+
+            /* Read the attribute */
+            if (H5Aread(aid, tid, buf) < 0)
+                goto out;
+            buf[string_size] = '\0';
+
+            /* compare strings */
+            if (strcmp(buf, DIMENSION_SCALE_CLASS) == 0)
+                is_ds = 1;
+            else
+                is_ds = 0;
         }
-
-        buf = (char *)malloc((size_t)string_size * sizeof(char));
-        if (buf == NULL)
-            goto out;
-
-        /* Read the attribute */
-        if (H5Aread(aid, tid, buf) < 0)
-            goto out;
-
-        /* compare strings */
-        if (strncmp(buf, DIMENSION_SCALE_CLASS, MIN(strlen(DIMENSION_SCALE_CLASS), strlen(buf))) == 0)
-            is_ds = 1;
-
-        free(buf);
-
-        if (H5Tclose(tid) < 0)
-            goto out;
-
-        if (H5Aclose(aid) < 0)
-            goto out;
     }
 out:
-    if (is_ds < 0) {
-        free(buf);
-        H5E_BEGIN_TRY
-        {
-            H5Aclose(aid);
-            H5Tclose(tid);
-        }
-        H5E_END_TRY
+    free(buf);
+    H5E_BEGIN_TRY
+    {
+        H5Aclose(aid);
+        H5Tclose(tid);
     }
+    H5E_END_TRY
     return is_ds;
 }
 
@@ -2441,6 +2459,7 @@ static herr_t
 H5DS_is_reserved(hid_t did, bool *is_reserved)
 {
     htri_t has_class;
+    htri_t isvl;
     hid_t  tid = H5I_INVALID_HID;
     hid_t  aid = H5I_INVALID_HID;
     char  *buf = NULL;  /* Name of attribute */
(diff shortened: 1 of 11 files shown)


hl/src/H5DS.c
```diff
@@ -2463,23 +2485,31 @@ H5DS_is_reserved(hid_t did, bool *is_reserved)
...
+
     /* Check to make sure string is null-terminated */
     if (H5T_STR_NULLTERM != H5Tget_strpad(tid))
         goto error;
 
     /* Allocate buffer large enough to hold string */
     if ((string_size = H5Tget_size(tid)) == 0)
         goto error;
-    if (NULL == (buf = malloc(string_size * sizeof(char))))
+    /* Allocate one extra byte and force NUL termination so strcmp never
```
bmribler: typo NUL?

hl/src/H5TB.c
```diff
@@ -3021,11 +3026,20 @@ H5TBget_field_info(hid_t loc_id, const char *dset_name, char *field_names[], siz
...
-            strcpy(field_names[i], member_name);
+
+            name_len = strlen(member_name);
+            if (name_len >= HLTB_MAX_FIELD_LEN) {
+                memcpy(field_names[i], member_name, HLTB_MAX_FIELD_LEN - 1);
+                field_names[i][HLTB_MAX_FIELD_LEN - 1] = '\0';
+            }
+            else {
+                memcpy(field_names[i], member_name, name_len + 1);
+            }
```
bmribler, line 3045: Do we know field_names[i] is large enough?  Is it a guarantee that it is HLTB_MAX_FIELD_LEN?

bmribler:
General cosmetic comment: I think the use of 0 vs. SUCCEED and -1 vs. FAIL should be more consistent.

bmribler (changes requested):

hl/src/H5DS.c
brtnfld: NUL is the name of the first character in the ASCII character set, literal as '\0'.

but since HDF5 uses NULL (pointer meaning) instead, I've changed it to NULL.

hl/src/H5TB.c
brtnfld: if branch (name_len >= HLTB_MAX_FIELD_LEN): copies exactly HLTB_MAX_FIELD_LEN - 1 bytes then writes '\0' -> correctly truncates any overlong name to 254 chars + null, always within a 255-byte buffer.

else branch (name_len < HLTB_MAX_FIELD_LEN): copies name_len + 1 bytes. Since name_len <= 254, the maximum copy is 255 bytes. That fits in a HLTB_MAX_FIELD_LEN-sized buffer exactly.

The else branch is not worse than the old strcpy for any caller who actually follows the API contract. With strcpy, callers had to know the name length in advance to size the buffer, which defeats the purpose of calling H5TBget_field_info. In practice, callers don't know the field names before calling this function, so they must pre-allocate a fixed-size buffer. HLTB_MAX_FIELD_LEN is now the documented contract for that size.

There is no stronger runtime enforcement; both old and new code rely on the caller's buffer being large enough. The improvement is:

- Long names no longer overflow (they're truncated)
- The maximum required buffer size is now bounded and documented

bmribler: Understood.  I was thinking about "stronger runtime enforcement", but this is fine.

bmribler (dismissed):

lrknox (dismissed):

hl/src/H5IM.c
```diff
@@ -967,102 +967,113 @@ H5IMget_palette(hid_t loc_id, const char *image_name, int pal_number, unsigned c
...
-        /* check to make sure attribute is a string */
-        if (H5T_STRING != H5Tget_class(atid))
-            goto out;
+    if (H5T_STRING != H5Tget_class(atid))
+        goto out;
 
-        /* check to make sure string is null-terminated */
-        if (H5T_STR_NULLTERM != H5Tget_strpad(atid))
-            goto out;
+    /* Reject VLEN strings; H5Aread into a char* buffer would write an hvl_t.
```
jhendersonHDF: Does the specification for these attributes disallow variable-length strings? What will be read is `char **`, _not_ `hvl_t`.

brtnfld: Corrected comment,  char * pointer overwrite rather than hvl_t. Fixed occurrences elsewhere as well.

jhendersonHDF: My main concern is whether or not variable-length strings should actually be rejected. The specifications don't seem to directly say that they should. `PALETTE_CLASS` and `IMAGE_CLASS` don't appear in the current specification and there is a line of text that says "Optionally, String valued attributes may be stored in a String longer than the minimum, in which case it must be zero terminated or null padded."

brtnfld: changed to accept VL strings

hl/test/test_table.c
```diff
@@ -1480,14 +1506,161 @@ test_table(hid_t fid, int do_write)
...
+        hid_t   fid_boundary = H5I_INVALID_HID;
+        hid_t   boundary_field_types[BOUNDARY_NFIELDS];
+        size_t  boundary_field_offsets[BOUNDARY_NFIELDS];
+        char  **boundary_names_out = NULL;
+        size_t  boundary_sizes_out[BOUNDARY_NFIELDS];
+        size_t  boundary_offset_out[BOUNDARY_NFIELDS];
+        size_t  boundary_size_out;
+        hsize_t boundary_nfields;
+        hsize_t boundary_nrecords;
+        char   *field_253  = (char *)malloc(254);
```
mattjala: These malloc calls should have error checks

hl/test/test_table.c
```diff
@@ -1480,14 +1506,161 @@ test_table(hid_t fid, int do_write)
...
+        int boundary_data[BOUNDARY_NRECORDS][BOUNDARY_NFIELDS] = {{1, 2, 3, 4, 5}, {6, 7, 8, 9, 10}};
+
+        if (H5TBmake_table("Boundary Test", fid_boundary, "boundary_table", BOUNDARY_NFIELDS,
+                           BOUNDARY_NRECORDS, BOUNDARY_NFIELDS * sizeof(int), boundary_field_names,
+                           boundary_field_offsets, boundary_field_types, 10, NULL, 0, boundary_data) < 0)
+            goto boundary_out;
+
+        boundary_names_out = (char **)malloc(sizeof(char *) * BOUNDARY_NFIELDS);
+        for (i = 0; i < BOUNDARY_NFIELDS; i++) {
+            boundary_names_out[i] = (char *)malloc(HLTB_MAX_FIELD_LEN);
```
mattjala, line 1615: These memory allocations should have error checks as well

brtnfld: Fixed this and other instances as well.

hl/test/test_table.c
```diff
@@ -1480,14 +1506,172 @@ test_table(hid_t fid, int do_write)
...
+        char  **boundary_names_out = NULL;
+        size_t  boundary_sizes_out[BOUNDARY_NFIELDS];
+        size_t  boundary_offset_out[BOUNDARY_NFIELDS];
+        size_t  boundary_size_out;
+        hsize_t boundary_nfields;
+        hsize_t boundary_nrecords;
+        char   *field_253  = (char *)malloc(254);
+        char   *field_254  = (char *)malloc(255);
+        char   *field_255  = (char *)malloc(256);
+        char   *field_1000 = (char *)malloc(1001);
```
jhendersonHDF: These values and those used below should probably be calculated in terms of `HLTB_MAX_FIELD_LEN` so that the test doesn't need updating if that value changes.

jhendersonHDF (dismissed):
Mostly just a minor comment about maintainability but looks good either way

hl/test/test_table.c
brtnfld: updated, along with similar issues.

mattjala (approved):

jhendersonHDF (approved):

PR #6372, opened 2026-04-17
Add exclude attribute by name option to h5diff

### LINKED ISSUE #6364: [h5diff] add abilty to exclude specific attributes
This is a request to extend or add a feature to h5diff to exclude specific attributes.  Currently it can only exclude from comparison all the attributes attached to an object.

The background to this is that we use h5diff for the verification stage of our regression testing.  We compare the HDF5 file produced to a 'known good' file using h5diff.

Some objects whose value is specific to the actual run (such as execution date/time) will not match what's in 'known good' file even though that shouldn't cause the regression test to fail, and thus we don't want h5diff to report them as not equal.  Often these objects are specific attributes.

mattjala:
Added new h5diff option `--exclude-attr-name` (short form `-B`) which allows users to exclude specific attribute names from the comparison.  It can be specified multiple times to skip multiple attribute names. If multiple attributes share the provided name(s), then every one will be skipped.

This PR also includes new tests (h5diff_488-491), test file generation in h5diffgentest, and modifies the pre-existing test files which involve usage text to include the new usage text for the new option.

Resolves #6364


mattjala:
There is an issue in ph5diff where pointer fields (`exclude`, `exclude_attr`, `exclude_attr_names`) were broadcast directly into worker processes with where the addresses pointed to were undefined memory. This most recent commit adds a couple helper functions to pack these lists for the transfer and unpack them afterwards in each worker process.

There may be other fields in `diff_opts_t` which require a similar fix in order to be safe for use by worker processes, but I think that's out of scope for this PR.

tools/src/h5diff/ph5diff_main.c
```diff
@@ -22,6 +22,107 @@
 
 static void ph5diff_worker(int);
 
+#ifdef H5_HAVE_PARALLEL
```
jhendersonHDF, line 25: This file only gets compiled for parallel HDF5, no real need to #ifdef here

tools/src/h5diff/ph5diff_main.c
```diff
@@ -78,9 +179,19 @@ main(int argc, char *argv[])
...
             parse_command_line(argc, (const char *const *)argv, &fname1, &fname2, &objname1, &objname2,
                                &opts);
 
+            /* Send exclude lists to each worker via point-to-point so that
+             * workers can restore the pointer fields in diff_opt_t after
+             * receiving it as a flat byte copy over MPI.  Using point-to-point
+             * (rather than MPI_Bcast) means workers stay in their probe loop
+             * and can still receive MPI_TAG_END if we exit early. */
+            for (i = 1; i < g_nTasks; i++)
+                send_opts_to_worker(&opts, i);
```
jhendersonHDF, line 193: This isn't going to scale particularly well and seems to point to a design issue. There's no reason you should need to use point-to-point operations rather than a broadcast, but you will need to do some extra work to get things in the right form for a broadcast. This is probably a case for https://www.open-mpi.org/doc/v3.1/man3/MPI_Pack.3.php, but I think more discussion is needed on the design for this feature.

That said, I think this can all be handled with `MPI_TAG_ARGS` which would fix the problem with error handling in the first place. See the other comment about this.

tools/src/h5diff/ph5diff_main.c
```diff
@@ -115,16 +226,21 @@ main(int argc, char *argv[])
...
+    hid_t      file2_id    = H5I_INVALID_HID;
+    diff_opt_t worker_opts = {0}; /* exclude lists populated via MPI_TAG_OPTS */
 
     while (1) {
         MPI_Status Status;
 
         MPI_Probe(0, MPI_ANY_TAG, MPI_COMM_WORLD, &Status);
 
+        /* Receive exclude list options from manager (sent before MPI_TAG_PARALLEL) */
+        if (Status.MPI_TAG == MPI_TAG_OPTS) {
```
jhendersonHDF, line 239: Let's not introduce a separate tag just for this one option, as others may be tempted to do the same in the future for each new option. This can be handled with `MPI_TAG_ARGS`, it's just going to take a bit of extra work.

Currently, `ph5diff` can get away with broadcasting the arguments for the most part because the object names use fixed-length fields in `diff_mpi_args`. However, the subsetting and attribute exclusion options are problematic due to their use of dynamic allocation, where an address on one MPI rank will almost certainly be invalid on another. This was probably overlooked at the time because `ph5diff` tries to share code with `h5diff` and the dynamically-allocated fields weren't specifically thought about for parallel in the shared `diff()` function.

This will take a decent bit of refactoring to handle "correctly", but I'd suggest creating a packed MPI type from everything currently sent in `diff_mpi_args`, along with the dynamically-allocated fields and then have the parallel workers unpack that type after the `MPI_Recv` in `ph5diff_worker()`, ideally setting the pointers in the `opts` struct after doing so. The benefit would be that we could expand to new dynamically-allocated options in the future and also could have arbitrary object name lengths if needed, as opposed to failing in `ph5diff` for object names > 255 characters.

I'd say some sort of refactoring needs to happen anyway, as it can be unclear whether a field from the `opts` struct is safe to use in parallel or not without having an overall understanding of that code.

tools/src/h5diff/h5diff_common.c
```diff
@@ -800,6 +837,15 @@ usage(void)
     PRINTVALSTREAM(rawoutstream, "         excluded.\n");
     PRINTVALSTREAM(rawoutstream, "         This option can be used repeatedly to exclude multiple paths.\n");
     PRINTVALSTREAM(rawoutstream, "\n");
+    PRINTVALSTREAM(rawoutstream, "   -B \"attr_name\", --exclude-attr-name \"attr_name\"\n");
```
jhendersonHDF, line 840: Matching all the attributes with the specified name is almost certain to cause frustration. This seems like it would benefit more from a path-style specification like `h5dump` has, perhaps with the ability to specify "all attributes matching this name".

jhendersonHDF (changes requested):
Just noting a couple things while I'm looking over this

PR #6377, opened 2026-04-21
Fix NULL pointer dereferences in H5HL_protect and H5HL_unprotect

tbeu:
When reading corrupted HDF5 files, `H5HL_protect` can encounter a heap where `prfx->heap` is NULL or `heap->dblk_image` is NULL. Similarly, `H5HL_unprotect` can encounter a heap where `heap->prfx` or `heap->dblk` has been set to NULL by `H5HL__prfx_dest` during cache eviction of corrupted entries.

In both cases the code proceeded to dereference NULL pointers, causing a segfault. This was found by [OSS-Fuzz](https://github.com/google/oss-fuzz) via the [matio](https://github.com/tbeu/matio) project's fuzzer.

### Changes

- **`H5HL_protect`**: Add NULL checks for `prfx->heap` and `heap->dblk_image` after loading the local heap prefix from the cache. Return an error via `HGOTO_ERROR` instead of crashing.
- **`H5HL_unprotect`**: Replace the `assert(heap->dblk)` with a proper NULL check and `HGOTO_ERROR`. Add an equivalent NULL check for `heap->prfx` in the `single_cache_obj` branch.

### ASAN report (from OSS-Fuzz)

https://issues.oss-fuzz.com/504827191

> =================================================================
> 	==250==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x57025b1b892d bp 0x7ffea4182170 sp 0x7ffea4182150 T0)
> 	==250==The signal is caused by a READ memory access.
> 	==250==Hint: address points to the zero page.
	    # 0 0x57025b1b892d in H5AC_unpin_entry [hdf5/src/H5AC.c:1438](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5AC.c#L1438):28
	    # 1 0x57025b51aed5 in H5HL_unprotect [hdf5/src/H5HL.c:436](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5HL.c#L436):25
	    # 2 0x57025b49f6d3 in H5G__stab_lookup [hdf5/src/H5Gstab.c:832](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5Gstab.c#L832):17
	    # 3 0x57025b496916 in H5G__obj_lookup [hdf5/src/H5Gobj.c:1086](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5Gobj.c#L1086):13
	    # 4 0x57025b4a356b in H5G__traverse_real [hdf5/src/H5Gtraverse.c:572](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5Gtraverse.c#L572):13
	    # 5 0x57025b4a26e9 in H5G_traverse [hdf5/src/H5Gtraverse.c:845](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5Gtraverse.c#L845):9
	    # 6 0x57025b47ab2e in H5G_loc_find [hdf5/src/H5Gloc.c:423](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5Gloc.c#L423):9
	    # 7 0x57025b2ae07d in H5D__open_name [hdf5/src/H5Dint.c:1474](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5Dint.c#L1474):9
	    # 8 0x57025bd53dd3 in H5VL__native_dataset_open [hdf5/src/H5VLnative_dataset.c:331](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5VLnative_dataset.c#L331):25
	    # 9 0x57025bd07a64 in H5VL__dataset_open [hdf5/src/H5VLcallback.c:2053](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5VLcallback.c#L2053):25
	    # 10 0x57025bd074a2 in H5VL_dataset_open [hdf5/src/H5VLcallback.c:2088](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5VLcallback.c#L2088):30
	    # 11 0x57025b27e0c5 in H5D__open_api_common [hdf5/src/H5D.c:361](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5D.c#L361):25
	    # 12 0x57025b27da7e in H5Dopen2 [hdf5/src/H5D.c:400](https://github.com/HDFGroup/hdf5/blob/4fde4fd787744b6a19b12032f4dad0383fbe6fbd/src/H5D.c#L400):22


bmribler:
@tbeu Thank you for the PR.  Please see my comment in PR #6378.  Similar to that PR, we need to investigate why/where these values are NULL and flag the errors there, in addition to these checks.

tbeu:
The issue occurs with corrupted HDF5 files on fuzzing. It is not possible to detect / prevent the error outside hdf5, i.e. as consumer of hdf5 lib.

bmribler:
Hi @tbeu, how would you like to proceed with this PR?

tbeu:
> Hi @tbeu, how would you like to proceed with this PR?

My idea is that the three assigned reviewers assess the changes and then merge or reject.

src/H5HL.c
```diff
@@ -325,6 +325,10 @@ H5HL_protect(H5F_t *f, haddr_t addr, unsigned flags)
 
     /* Get the pointer to the heap */
     heap = prfx->heap;
+    if (NULL == heap)
```
mattjala: Because of how the unprotection in the `done` block at the end of this function is setup, this check creates the possibility for a leak. Specifically, it creates a path for `if (prfx && heap && H5AC_unprotect(f, H5AC_LHEAP_PRFX, heap->prfx_addr, prfx, H5AC__NO_FLAGS_SET) < 0)` to be executed when heap is NULL and prfx is not, causing `H5AC_unprotect` to not be called.  

Additionally, this check probably doesn't really catch any dangerous cases, since `prfx->heap` is set in `H5HL__prfx_new` and cleared only in `H5HL__prfx_dset`, which frees the prefix in the same call. I'd say it should just be removed.

src/H5HL.c
```diff
@@ -325,6 +325,10 @@ H5HL_protect(H5F_t *f, haddr_t addr, unsigned flags)
...
+        HGOTO_ERROR(H5E_HEAP, H5E_CANTPROTECT, NULL, "local heap is NULL");
+    if (NULL == heap->dblk_image)
```
mattjala: The CI failures are due to the fact that `heap->dblk_image` is NULL by design here whenever the data block isn't contiguous with the prefix. `H5HL__cache_prefix_deserialize` allocates it only when `single_cache_obj = true` (src/H5HLcache.c:436-464). Otherwise it's allocated later by `H5HL__cache_datablock_deserialize`. This check should be removed.

mattjala (changes requested):
In addition to my other comments, this should include a regression test case using the file from the OSS report and a CHANGELOG.md entry describing the problem that was fixed.

tbeu:
Review remarks are addressed.

tbeu:
> https://issues.oss-fuzz.com/504827191

It has been published meanwhile.

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+ *-------------------------------------------------------------------------
+ */
+/* Visitor: recurse into groups and open datasets by name. Best-effort smoke
+ * test for the fuzzer file (must not crash on open/traverse). */
+static herr_t
+corrupt_heap_visit(hid_t group, const char *name, const H5L_info_t *info, void *op_data)
+{
+    H5O_info_t oinfo;
+
+    H5E_BEGIN_TRY
```
mattjala, line 65: The entire block should not be under `H5E_BEGIN_TRY`. Each operation should be error-checked immediately. Even if a failure is expected, the try block should be narrowly scoped to the specific operation(s) which are expected to fail.

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+    {
+        if (H5Lget_info(group, name, (H5L_info_t *)info, H5P_DEFAULT) >= 0 &&
+            info->type == H5L_TYPE_HARD &&
+            H5Oget_info_by_name3(group, name, &oinfo, H5O_INFO_BASIC, H5P_DEFAULT) >= 0) {
+            if (oinfo.type == H5O_TYPE_GROUP) {
+                hid_t g = H5Gopen2(group, name, H5P_DEFAULT);
+                if (g >= 0) {
+                    hsize_t i = 0;
```
mattjala: This `i` variable being declared and unused is tripping warnings and build failures

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+
+    TESTING("corrupted local heap unprotect (OSS-Fuzz 504827191)");
+
+    /* Best-effort smoke test: the minimized fuzzer file must open and traverse
+     * without crashing. The deterministic check below forces the exact buggy
+     * condition that the fuzzer file triggers inside HDF5. */
+    {
+        const char *tf = H5_get_srcdir_filename(CORRUPT_HEAP_FILE);
+
+        H5E_BEGIN_TRY
```
mattjala, line 143: See above comment about `H5E_BEGIN_TRY`

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+        {
+            file = H5Fopen(tf, H5F_ACC_RDONLY, H5P_DEFAULT);
+            if (file >= 0) {
+                hsize_t i = 0;
```
mattjala: This unused var is also tripping build warnings/errors.

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+
+    /* Deterministic regression test for the NULL-pointer dereference in
+     * H5HL_unprotect(). A corrupted entry whose local heap prefix/data block
+     * pointer became NULL (as set by H5HL__prfx_dest during cache eviction)
+     * used to crash. Simulate that by nulling the pointers on a protected heap
+     * and verify H5HL_unprotect() returns an error instead of dereferencing
+     * NULL.
+     */
+    fapl = h5_fileaccess();
+    h5_fixname(CORRUPT_HEAP_TESTFILE, fapl, filename, sizeof filename);
```
mattjala: `h5_fixname` appends `.h5`, so the final filename here would be `heap_corrupt_prfx.h5.h5`

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
 #include "H5VLprivate.h" /* Virtual Object Layer                     */
 
 static const char *FILENAME[] = {"lheap", NULL};
 
 #define TESTFILE "tsizeslheap.h5"
 
+/* File from OSS-Fuzz (matio fuzzer, issue 504827191): a corrupted local
+ * heap whose prefix/data block pointer becomes NULL during cache eviction.
+ * Opening and traversing it used to crash in H5HL_unprotect(). */
+#define CORRUPT_HEAP_FILE "heap_corrupt_prfx.h5"
```
mattjala, line 31: If `h5_fixname` is to be used on this, then it should be added to the `FILENAME[]` array to be cleaned up.

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+    }
+
+    /* Deterministic regression test for the NULL-pointer dereference in
+     * H5HL_unprotect(). A corrupted entry whose local heap prefix/data block
+     * pointer became NULL (as set by H5HL__prfx_dest during cache eviction)
+     * used to crash. Simulate that by nulling the pointers on a protected heap
+     * and verify H5HL_unprotect() returns an error instead of dereferencing
+     * NULL.
+     */
+    fapl = h5_fileaccess();
```
mattjala: This properly list is never cleaned up.

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+/* Visitor: recurse into groups and open datasets by name. Best-effort smoke
+ * test for the fuzzer file (must not crash on open/traverse). */
+static herr_t
+corrupt_heap_visit(hid_t group, const char *name, const H5L_info_t *info, void *op_data)
+{
+    H5O_info_t oinfo;
+
+    H5E_BEGIN_TRY
+    {
+        if (H5Lget_info(group, name, (H5L_info_t *)info, H5P_DEFAULT) >= 0 &&
```
mattjala: It would be better to use `H5Lget_info2()` to keep this test consistent, which would remove the need for the cast here (which is also triggering warnings).

That said, this entire get info calls seems redundant, since the link iteration process (`H5G_link_to_info`) should already have populated `info` anyway.

test/lheap.c
```diff
@@ -15,18 +15,179 @@
...
+ *
+ * Return:      Success:        0
+ *              Failure:        1
+ *
+ *-------------------------------------------------------------------------
+ */
+/* Visitor: recurse into groups and open datasets by name. Best-effort smoke
+ * test for the fuzzer file (must not crash on open/traverse). */
+static herr_t
+corrupt_heap_visit(hid_t group, const char *name, const H5L_info_t *info, void *op_data)
```
mattjala: Since this callback is passed to `H5Lvisit2`, this `H5L_info_t` should be `H5L_info2_t`.

mattjala (changes requested):

test/lheap.c
```diff
@@ -37,53 +37,65 @@ static const char *FILENAME[] = {"lheap", NULL};
...
+    (void)op_data;
 
-    H5E_BEGIN_TRY
-    {
-        if (H5Lget_info(group, name, (H5L_info_t *)info, H5P_DEFAULT) >= 0 &&
-            info->type == H5L_TYPE_HARD &&
-            H5Oget_info_by_name3(group, name, &oinfo, H5O_INFO_BASIC, H5P_DEFAULT) >= 0) {
+    /* info is already populated by the link iteration; only follow hard links */
+    if (info->type == H5L_TYPE_HARD) {
+        H5O_info_t oinfo;
```
mattjala: Should be `H5O_info2_t`

test/lheap.c
```diff
@@ -15,18 +15,197 @@
...
+
+    /* Save the valid pointers, then simulate a corrupted/evicted entry whose
+     * local heap prefix/data block pointer became NULL (as set by
+     * H5HL__prfx_dest during cache eviction), and verify that H5HL_unprotect()
+     * returns an error instead of dereferencing NULL.
+     */
+    saved_prfx = heap->prfx;
+    saved_dblk = heap->dblk;
+    heap->prfx = NULL;
+    heap->dblk = NULL;
```
mattjala: It would be preferable to test against a real corrupt file using the workflow that produced the issue original instead of hard coding these pointers to be NULL.

release_docs/CHANGELOG.md
```diff
@@ -83,6 +83,10 @@ We would like to thank the many HDF5 community members who contributed to this r
 
    Fixes GitHub issue #6531
 
+### Fixed NULL pointer dereferences when unprotecting a corrupted local heap
+
+   Fixed a crash (segfault / assertion failure) that could occur when reading a corrupted file whose local heap prefix or data block pointer became NULL during cache eviction. `H5HL_unprotect()` now validates these pointers and returns an error via the normal HDF5 error mechanism instead of dereferencing NULL. Found by OSS-Fuzz via the matio project's fuzzer.
```
hyoklee: Remove space: ` (segfault / assertion failure)` -> ` (segfault/assertion failure)`
Change: `normal HDF5 error mechanism` -> `standard HDF5 error mechanism`

hyoklee (dismissed):

bmribler:
Hi @tbeu, please resolve the CHANGELOG conflict and after @mattjala's approval, this can be merged.  Thanks so much for your patience.

tbeu:
Feel free to push the proposed changes to my branch. GitHub allows editing of PR branches from maintainers.

mattjala (approved):

hyoklee (approved):

PR #6380, opened 2026-04-22
Pin BLOSC2 git fetch to versioned tag

brtnfld:
BLOSC2_GIT_BRANCH was "main" causing the inline plugin build to fetch unreleased c-blosc2 that breaks test output comparisons. Pin to the tagged release matching HDF5_BLOSC2_VERSION so the git and tgz paths are consistent.


hyoklee (approved):

bmribler (approved):

mattjala (approved):

config/CacheURLs.cmake
jhendersonHDF, line 1: @brtnfld Every plugin in this file uses the respective "development" branch (except for blosc2 now), presumably because releases are typically built with the tgz option rather than git. Either every URL should use the version branches (which may be harder to maintain) or none of them should. I don't think this one change should be made for blosc2 specifically because it seems like the intention was for the git option to point to the latest development by default and if you had a workflow that was failing you should specify the git branch rather than leaving the default.

That is, I'm thinking this change should probably be reverted. Or at least every other URL should be updated to match.

Also worth noting that this is now out of date with the CacheURLs file in hdf5_plugins.

PR #6381, opened 2026-04-23
To match the values in Table 1 to those of the original pdf.

furutaka:
The value for `File_with_compression.h5` should read `345 seconds` instead of `0.37 seconds`.
according to [the original? document.](https://support.hdfgroup.org/documentation/hdf5-docs/hdf5_topics/HDF5ImprovingIOPerformanceCompressedDatasets.pdf)


hyoklee (approved):
@furutaka , thank you for your PR!

bmribler (approved):

PR #6385, opened 2026-04-28
Fix NULL file pointer dereferences in vlen disk operations

tbeu:
When reading corrupted HDF5 files, the file pointer passed to the disk-based vlen type operations (`H5T__vlen_disk_read`, `H5T__vlen_disk_write`, `H5T__vlen_disk_isnull`, `H5T__vlen_disk_setnull`, `H5T__vlen_disk_delete`) and `H5T__vlen_set_loc` can be NULL. The existing `assert(file)` causes a crash in debug builds, while release builds proceed to dereference the NULL pointer in `H5VL_blob_put` / `H5VL_blob_get` / `H5VL_blob_specific`, causing a SEGV.

### Changes

Replace `assert(file)` with proper NULL checks and `HGOTO_ERROR` so that the error is propagated through the normal HDF5 error handling mechanism.

### ASAN report (from OSS-Fuzz)

https://issues.oss-fuzz.com/issues?q=5366895365914624

> =================================================================
> ==1150==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x5d09af33f82f bp 0x7ffd030797b0 sp 0x7ffd03079770 T0)
> ==1150==The signal is caused by a READ memory access.
> ==1150==Hint: address points to the zero page.
>     # 0 H5VL_blob_put hdf5/src/H5VLcallback.c:7453:48
>     # 1 H5T__vlen_disk_write hdf5/src/H5Tvlen.c:879:9
>     # 2 H5T__conv_vlen hdf5/src/H5Tconv_vlen.c:474:29
>     # 3 H5T_convert_with_ctx hdf5/src/H5T.c:6545:14
>     # 4 H5T_convert hdf5/src/H5T.c:6474:9
>     # 5 H5A__read hdf5/src/H5Aint.c:787:21

### DIFF
--- a/src/H5Tvlen.c
+++ b/src/H5Tvlen.c
@@ -282,7 +282,8 @@ H5T__vlen_set_loc(H5T_t *dt, H5VL_object_t *file, H5T_loc_t loc)
                 H5VL_file_cont_info_t cont_info = {H5VL_CONTAINER_INFO_VERSION, 0, 0, 0};
                 H5VL_file_get_args_t  vol_cb_args; /* Arguments to VOL callback */
 
-                assert(file);
+                if (NULL == file)
+                    HGOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "file pointer is NULL for disk VL type");
 
                 /* Mark this type as being stored on disk */
                 dt->shared->u.vlen.loc = H5T_LOC_DISK;
@@ -749,7 +750,8 @@ H5T__vlen_disk_isnull(const H5VL_object_t *file, void *_vl, bool *isnull)
     FUNC_ENTER_PACKAGE
 
     /* Check parameters */
-    assert(file);
+    if (NULL == file)
+        HGOTO_ERROR(H5E_DATATYPE, H5E_BADVALUE, FAIL, "file pointer is NULL");
     assert(vl);
     assert(isnull);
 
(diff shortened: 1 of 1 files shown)


qkoziol (approved):

vchoi-hdfgroup (approved):

bmribler:
Please do not merge this yet.  The comment in https://github.com/HDFGroup/hdf5/pull/6377 also applies here.

tbeu:
The issue occurs with corrupted HDF5 files on fuzzing. It is not possible to detect / prevent the error outside hdf5, i.e. as consumer of hdf5 lib.

jhendersonHDF:
Hi @tbeu,

I just wanted to clarify a bit. What @bmribler is saying is that we currently believe it would be more appropriate to fix this at a higher level rather than converting these `assert()`s to regular error checks. That is, we expect that these deeper internal interfaces should always be getting a non-`NULL` pointer (the reason for the `assert()`s being there) and checking with a regular error check elsewhere would be better.

src/H5Tvlen.c
```diff
@@ -282,7 +282,8 @@ H5T__vlen_set_loc(H5T_t *dt, H5VL_object_t *file, H5T_loc_t loc)
                 H5VL_file_cont_info_t cont_info = {H5VL_CONTAINER_INFO_VERSION, 0, 0, 0};
                 H5VL_file_get_args_t  vol_cb_args; /* Arguments to VOL callback */
 
-                assert(file);
```
jhendersonHDF, line 285: For example, it may make sense to leave this as an `assert()` and instead add a regular error check to `H5T_set_loc()` in H5T.c when the `loc` param is `H5T_LOC_DISK`. Even better would be to figure out where the `file` pointer initially gets set to `NULL` and fix the issue there.

tbeu:
Closing in favor of a fix at the decode level (validating vlen.type in H5O__dtype_decode_helper). See replacement PR.

PR #6388, opened 2026-05-01
ci: remove EOL OS versions from workflows

brtnfld:
- freebsd.yml: drop FreeBSD 13.5 (EOL Jan 2026); keep 14.3 and 15.0
- openbsd.yml: bump 7.5 → 7.8 (7.5 EOL ~Nov 2024); drop pinned gcc version
- i386.yml: replace Alpine 3.16 x86 (EOL May 2024) with i386/debian:bookworm (EOL Jun 2028)

### DIFF
--- a/.github/workflows/openbsd.yml
+++ b/.github/workflows/openbsd.yml
@@ -45,7 +45,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        openbsd-version: ['7.5']
+        openbsd-version: ['7.8']
 
     steps:
       - name: Checkout repository
@@ -58,7 +58,7 @@ jobs:
           usesh: true
           prepare: |
             echo "https://ftp.openbsd.org/pub/OpenBSD" > /etc/installurl
-            pkg_add cmake gmake pkgconf curl gcc-11.2.0p11
+            pkg_add cmake gmake pkgconf curl gcc
           run: |
             set -e
 
--- a/.github/workflows/openbsd.yml
+++ b/.github/workflows/openbsd.yml
@@ -52,18 +52,15 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Build and test on OpenBSD
-        uses: vmactions/openbsd-vm@271a1ba62300483cfc58345ff9f425f1349a2cab # v1.3.4
+        uses: vmactions/openbsd-vm@d7d892b7b9ba97ed2747b0fc201be65037d64c3e # v1.4.0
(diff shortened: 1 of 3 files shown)


.github/workflows/openbsd.yml
```diff
@@ -45,26 +45,23 @@ jobs:
...
       - name: Build and test on OpenBSD
-        uses: vmactions/openbsd-vm@271a1ba62300483cfc58345ff9f425f1349a2cab # v1.3.4
+        uses: vmactions/openbsd-vm@d7d892b7b9ba97ed2747b0fc201be65037d64c3e # v1.4.0
         with:
           release: ${{ matrix.openbsd-version }}
           usesh: true
           prepare: |
             echo "https://ftp.openbsd.org/pub/OpenBSD" > /etc/installurl
-            pkg_add cmake gmake pkgconf curl gcc-11.2.0p11
+            pkg_add cmake gmake pkgconf curl
```
hyoklee: Remove `gmake`.

hyoklee (changes requested):

.github/workflows/openbsd.yml
brtnfld: gmake is used as the cmake build, you want to use ninja instead?

.github/workflows/openbsd.yml
```diff
@@ -75,8 +72,6 @@ jobs:
             cmake -C ../config/cmake/cacheinit.cmake \
               --log-level=VERBOSE \
               -DCMAKE_BUILD_TYPE=Release \
-              -DCMAKE_C_COMPILER=egcc \
-              -DCMAKE_CXX_COMPILER=eg++ \
               -DCMAKE_MAKE_PROGRAM=gmake \
```
hyoklee: Remove `-DCMAKE_MAKE_PROGRAM=gmake \`

hyoklee (changes requested):

.github/workflows/openbsd.yml
brtnfld: why use Ninja instead?

.github/workflows/openbsd.yml
hyoklee: cmake will be sufficient.

.github/workflows/openbsd.yml
hyoklee: It won't work.

hyoklee (approved):

lrknox (approved):

lrknox:
nvhpc workflow failed compiling test/dtarith, as expected.  This workflow will be disabled by PR #6390.

Issue #4481, opened 2024-05-13
Reading a hyperslab of a limited dataset after an unlimited dataset increases the execution time of the former after multiple runs.


abhibaruah:
**Seen only in Windows**

I have 2 h5 files with 2 datasets. The first dataset is of size 1999 x 512 x 512, with the first dimension being unlimited. It is compressed using deflate compression (level 6) and has chunk size of 20 x 10 x 10.

The second dataset is of size 2000 x 512 x 512 (no unlimited dimensions).  It is compressed using deflate compression (level 6) and has chunk size of 70 x 20 x 50.

In my reproduction code, I read the whole of the first dataset (then close the identifiers of the first file). Then, I open the second dataset and read a hyperslab of the dataset (start = {0,0,0} and stride = {2,2,2}). Then I close all identifiers.

I have noticed that the execution time of the hyperslab read increases in successive iterations. 
This happens only on Windows, and I could not reproduce the issue in Linux.

- HDF5 version (if building from a maintenance branch, please include the commit hash) : HDF5 1.10.11
 - OS and version : Windows 11

You can find the h5 files used inthe reproduction here: 
https://mathworks-my.sharepoint.com/:f:/p/abaruah/Ekr_sEmqx1hFlgPM4tWx55UBS7bzUtxAmHPM33bT6t51CQ?e=ChoCCJ

Reproduction Code:
```
#include "hdf5.h"
#include <stdio.h>
#include <stdlib.h>
#include <iostream>
#include <chrono>
#include <array>
#include <random>

#define FILE1 "unlimDim_smallChunkSize_deflate_2.h5"
#define DATASET1        "/lrg_unl_dset2_double_dset"
#define DIM10            1999
#define DIM11            512
#define DIM12            512


#define FILE2 "largeGroupNesting.h5"
#define DATASET2         "/lrg_dset2_double_dset"
#define DIM20            1000
#define DIM21            256
#define DIM22            256


void unlimDsetRead()
{
    hid_t           file, space, dset, dcpl;    /* Handles */
    herr_t          status;
    

    double* dset_read = new double[DIM10 * DIM11 * DIM12];
    /*
     * Create a new file using the default properties.
     */
    file = H5Fopen(FILE1, H5F_ACC_RDONLY, H5P_DEFAULT);

    /*
     * Create the chunked dataset.
     */
    dset = H5Dopen(file, DATASET1, H5P_DEFAULT);
    hid_t my_file_space_id = H5Dget_space(dset);
[...]

Issue #3858, opened 2023-11-16
H5Dwrite execution time in Windows increases in subsequent runs when a variable is written with chunking and compression


abhibaruah:
Hello all, 

I am trying to write a dataset of size 2000 x 512 x 512 with chunking and deflate compression (Chunk Size = 20 x 10 x 10, Deflate Level = 3).
I am timing the call to H5Dwrite.
I run this code in a for loop of 10 iterations and at the end of every iteration, delete the created h5 file.

I have noticed that for each successive run, the time taken by H5Dwrite increases. This issue is not seen if I remove the chunking and compression. 
I also tried the reproduction code to create a dataset of smaller size, but I could not reproduce the issue.

The issue is seen only on Windows and I could see the issue in HDF5 1.10.10 and 1.10.11.
Please find the reproduction C++ code below.
Here is the output from the program:

In main
Index: 0
Before H5Dwrite
After H5Dwrite
Execution time: 11.0488 seconds
Index: 1
Before H5Dwrite
After H5Dwrite
Execution time: 11.9102 seconds
Index: 2
Before H5Dwrite
After H5Dwrite
Execution time: 13.3122 seconds
Index: 3
Before H5Dwrite
After H5Dwrite
Execution time: 15.4624 seconds
Index: 4
Before H5Dwrite
After H5Dwrite
Execution time: 21.7152 seconds
Index: 5
Before H5Dwrite
After H5Dwrite
Execution time: 25.9505 seconds
Index: 6
Before H5Dwrite
After H5Dwrite
Execution time: 27.9015 seconds
Index: 7
Before H5Dwrite
After H5Dwrite
Execution time: 32.6089 seconds
Index: 8
Before H5Dwrite
After H5Dwrite
Execution time: 35.6063 seconds
Index: 9
Before H5Dwrite
After H5Dwrite
Execution time: 37.548 seconds

P.S: I saw a similar behavior in netCDF as well and think that it might be related (since netCDF4 uses HDF5 under the hood). I have created a netCDF ticket: https://github.com/Unidata/netcdf-c/issues/2750 but no one has responded there yet.

Kindly take a look at my reproduction script and let me know if there is anything I am doing wrong, or if this is a bug with HDF5.

**Platform (please complete the following information)**
 - HDF5 version (if building from a maintenance branch, please include the commit hash) : 1.10.10, 1.10.11
 - OS and version : Windows 10
 - Compiler and version : Visual Studio 2022 v17
[...]

PR #6389, opened 2026-05-01
pre-size chunk buffer to avoid realloc fragmentation; fix nbit/scaleoffset no-op return values

brtnfld:
1. Pre-size chunk buffer before the filter pipeline (H5Dchunk.c)

When reading a filtered chunk, the decompression buffer was initialized at the compressed on-disk size. Filters like deflate grow this buffer via repeated realloc() calls, doubling its size when the output exceeds capacity. On Windows, this fragments the heap and causes read times to increase steadily across successive iterations.

The fix pre-sizes the buffer to the full uncompressed chunk_size when a filter pipeline is active, so the output buffer is large enough from the start and no realloc is needed.

2. Return nbytes instead of *buf_size in nbit and scaleoffset no-op paths (H5Znbit.c, H5Zscaleoffset.c)

Per the HDF5 filter API, nbytes is the count of valid data bytes in the buffer; *buf_size is the allocated buffer capacity. These two values were historically always equal on the read path, so either worked in practice. The pre-sizing change above makes *buf_size > nbytes on compressed reads for the first time, exposing two latent bugs:

H5Znbit.c: the no-compress pass-through path (cd_values[1] == 1) returned *buf_size instead of nbytes. The downstream filter (e.g., Fletcher32) then treated the full pre-sized buffer as valid input and computed a checksum over uninitialized bytes, causing filter pipeline failures.
H5Zscaleoffset.c: the no-process path (when the data already spans the full bit range and no scaling is needed) had the same *buf_size vs nbytes mistake. No current test exercises this path through a multi-filter pipeline where the size mismatch would be observable, but the fix is correct by the same API reasoning.


brtnfld:
I will need to do a PR to netcdf for a fix.

src/H5Dchunk.c
```diff
@@ -4825,9 +4825,12 @@ H5D__chunk_lock(const H5D_io_info_t H5_ATTR_NDEBUG_UNUSED *io_info, const H5D_ds
...
                 H5_CHECKED_ASSIGN(buf_alloc, size_t, chunk_disk_size, hsize_t);
 
-                /* Ideally we should allocate a buffer at least as large as chunk_size, to give the filter the
-                 * opportunity to avoid doing a realloc when uncompressing. This causes problems now, however,
-                 * and needs more investigation. -NAF */
+                /* When filters are present, pre-size the buffer to the uncompressed chunk size.
+                 * This avoids repeated realloc() inside the filter pipeline (e.g., deflate's
+                 * doubling strategy when the output buffer fills), which on Windows fragments
+                 * the heap and causes read time to increase steadily across iterations. */
+                if ((old_pline && old_pline->nused) && buf_alloc < chunk_size)
```
fortnern: I'm not sure you need the first condition - you can probably assert that chunk_nbytes == chunk_size if there are no filters (I think it was checked earlier):

```
if (buf_alloc < chunk_size) {
    assert(old_pline && old_pline->nused);
    buf_alloc = chunk_size;
}
```

brtnfld: Revised it.

fortnern (dismissed):

fortnern (dismissed):

brtnfld:
Will benchmark on Windows before merging

brtnfld:
## Windows benchmark: PR #6389 chunk-buffer pre-sizing fix

**Setup:** Windows 11, Intel oneAPI 2026.1 — `icx` compiled the HDF5 C library under test (Release, CMake+Ninja, zlib fetched via `FetchContent`); `icx-cl` compiled the C++ benchmark harness. HDF5 built twice from git worktrees:
- **Baseline (pre-fix):** `efe6bd838b` (PR base commit)
- **Fix:** `ac6520ca21` (PR #6389 head)

**Mechanism, verified at the source level:** `H5Z_filter_deflate` in `H5Zdeflate.c` starts its internal `nalloc` at `*buf_size` (the `buf_alloc` computed in `H5D__chunk_lock()`) and only doubles+`realloc`s if that runs out mid-`inflate()`. Pre-fix, `buf_alloc` = `chunk_disk_size` (the small on-disk compressed size), so the loop grows repeatedly. Post-fix, `buf_alloc` = `MAX(chunk_disk_size, chunk_size)`, so it's already sufficient and the loop never reallocates.

**Benchmark:** Mirrors issue #4481's repro shape — each iteration alternates between a full read of dataset A and a strided (stride-2) hyperslab read of a differently-chunked dataset B, both deflate-compressed with highly compressible (~constant) data so `chunk_disk_size ≪ chunk_size`:

| Dataset | Chunks | Chunk shape | Access pattern |
|---|---|---|---|
| A | 30,000 | 8×8 doubles (512B) | full read |
| B | 20,000 | 12×12 doubles (1152B) | stride-2 hyperslab (10,000 chunks touched) |

Run sequentially, 1000 iterations each (open file → read A → read B → close, per iteration):

| | Baseline (pre-fix) | Fix (PR #6389) | Improvement |
|---|---|---|---|
| Avg A+B time/iter (all 1000 iters) | 0.2581 s | 0.2039 s | **21.0% faster** |
| Avg A+B time/iter (steady-state, iters 50–999) | 0.2581 s | 0.2029 s | **21.4% faster** |
| Max single iteration | 0.3507 s | 0.2562 s | |
| Steady-state working set | ~55.5 MiB | ~176.3 MiB | (see note) |

Prior, simpler benchmarks (single dataset, few large chunks) showed a similar, consistent effect at smaller scale:

| Scale | Chunks | Dataset | Iterations | Speedup |
|---|---|---|---|---|
| Small | 64 × 512×512 | 128 MB | 300 | 27.5% |
| Larger | 128 × 512×512 | 256 MB | 600 | 25.9% |

**~21–28% faster chunked reads, consistently, across every scale tested** — directly attributable to PR #6389 pre-sizing the chunk buffer to the full uncompressed `chunk_size` up front in `H5D__chunk_lock()`.

**Memory note (hypothesis, not confirmed):** the fix's process working set ramps from ~47 MiB to a steady ~176 MiB within the first ~50 iterations, then stays flat through iteration 1000 (not a leak — a one-time plateau, confirmed by sampling working set at 100-iteration intervals across the full run). A plausible explanation is Windows' Low-Fragmentation Heap growing a dedicated bucket in response to the fix's steady, fixed-size allocation pattern, while baseline's varying realloc-growth sizes may not trigger the same bucket promotion — but this wasn't verified with a heap profiler, so treat it as a working theory rather than a demonstrated mechanism. Real trade-off either way: worth being aware of on memory-constrained Windows deployments, though the speed win is the clear, verified headline.

I did not reproduce the original issue's escalating *multi-second* slowdown across iterations (#4481/#4513 saw read time climb from 17s→22s over 30 iterations on datasets orders of magnitude larger) — regression slope restricted to the steady-state window (iters 50–999) shows both variants drifting <2% of their average, i.e. essentially flat/noise. That escalating effect likely needs a much larger, longer-running dataset to visibly fragment the heap to the point of measurably slower allocation, rather than just a larger steady-state reservation.

brtnfld:
## Attempted reproduction of #4481/#4513's escalating-slowdown symptom

Following up on the benchmark results above: I attempted a dedicated reproduction of the *specific* symptom reported in #4481 (read time climbing over successive iterations, 17s→22s over 30 iterations) — not just the general throughput improvement.

**Methodology:** scaled a synthetic two-dataset benchmark to match the original repro's actual regime as closely as possible without the original files (hosted externally, inaccessible):
- Chunk sizes matched to the original's chunk dimensions: ~16KB (`20×10×10` doubles) and ~550KB (`70×20×50` doubles) — spanning both the LFH small-block and large-block/`VirtualAlloc` Windows heap regimes.
- Same structural pattern: full read of one dataset + stride-2 hyperslab read of a second, fresh file open/close every iteration.
- Ran 500 iterations (~11M total chunk allocate/free cycles), *exceeding* the original repro's estimated ~8.1M cycles (30 iterations × ~270K chunks/iteration).

**Result: no escalation in either variant.**

| | Baseline (pre-fix) | Fix (PR #6389) |
|---|---|---|
| Average | 3.7831 s | 3.5181 s (**7% faster**) |
| Full-run drift (regression slope × 500 iters) | 0.7% (noise) | 0.2% (noise) |
| Working-set growth over 500 iters | +1 MB | +26 MB |
| 10×50-iteration windowed trend | oscillates 3.75–3.82s, no direction | flat, 3.51–3.53s band |

An initial smaller run (150 iterations, ~3.3M cycles) showed a modest baseline drift (+3.7%) that looked like a promising signal — but it did not hold up or compound at the larger scale; it was noise from an undersized sample.

**Conclusion:** PR #6389 gives a consistent, real speedup here too (7%, in line with the 21–28% seen at other chunk sizes elsewhere in this thread), but I was not able to reproduce the *specific* escalating-degradation symptom from #4481/#4513, even after scaling chunk sizes and total cycle count to match or exceed the original report's regime. Plausible reasons this doesn't rule out the bug being real and fixed:
- Our "pre-fix" baseline is current `develop` minus only this PR's 3-file diff — not literally HDF5 1.10.11 (the version in the original report). Years of unrelated changes sit between them.
- Windows' heap manager itself has changed since 2024; the exact fragmentation pattern reported may not manifest identically on current Windows builds.
- We don't have the reporter's actual files, only a size-matched synthetic approximation.

**Net assessment:** the fix is well-motivated (the realloc/copy churn it removes is real, confirmed at the source level in `H5Zdeflate.c` and the other filters) and delivers a consistent, verified performance improvement across every scale tested. Whether it specifically resolves the reported escalating-slowdown behavior remains empirically unconfirmed on my end — I'd frame the "Fixes #4481/#4513" claim as well-reasoned but not reproduced-and-verified, unless someone can test against the original files or a closer-matched environment.

brtnfld:
## Faithful reproduction of #4481's exact repro code — escalation confirmed and measurably reduced

Following up further: I rewrote the benchmark to use the **issue's actual repro code** (not just a size-matched synthetic approximation) — `unlimDsetRead()`/`hypersladDsetRead()` copied near-verbatim, including details a synthetic rewrite easily loses:
- Reading `H5T_NATIVE_FLOAT` from `H5T_NATIVE_DOUBLE`-stored data (triggers HDF5's type-conversion path)
- A stride-`{2,2,2}` hyperslab (all three dimensions, not just one)
- Dataset 1's unlimited first dimension (forces the Extensible Array chunk index, not Fixed Array)

Files were synthesized locally (same chunk shapes/dims as the repro: `20×10×10` and `70×20×50`. 100 iterations against both baseline (pre-fix) and fix (PR #6389):

| | Baseline (pre-fix) | Fix (PR #6389) |
|---|---|---|
| `hypersladDsetRead` rise (iter 0-4 avg → plateau) | **+22.6%** (5.29s → 6.49s) | **+10.9%** (4.25s → 4.71s) |
| `unlimDsetRead` rise | +7.8% | +5.1% |
| Windowed trend (10-iter blocks, `hypersladDsetRead`) | 5.72, 6.21, 6.56, 6.33, 6.30, 6.43, 6.42, 6.48, 6.52, 6.47 — climbs then plateaus | 4.34, 4.52, 4.58, 4.58, 4.61, 4.64, 4.66, 4.70, 4.72, 4.72 — climbs more gently, plateaus lower |

**This is a real reproduction of the reported effect** (unlike an earlier, simpler synthetic benchmark that showed no escalation at all) — both variants climb early then plateau, consistent with a heap-fragmentation-style effect rather than noise. **PR #6389 roughly halves the escalation but doesn't eliminate it.**

### Why it's reduced, not eliminated
This PR changed the *size* passed in, not the *allocator*. Every chunk-cache eviction during these reads is still a raw allocator round-trip. The fix removes the 4-9 reallocs-per-chunk it used to cost; it doesn't remove the fact that it's still raw `malloc`/`free`, still on the one allocator (Windows') where this apparently matters. That predicts exactly what we see: escalation reduced by roughly half, not eliminated.

**Net assessment:** PR #6389 is a real, verified, substantial improvement — both in raw throughput (21-28% faster reads across every benchmark scale) and in reducing (not eliminating) the specific escalating-slowdown behavior from #4481/#4513.

brtnfld:
## Full investigation summary: PR #6389, the escalating-slowdown symptom, and what's actually happening

This is a complete write-up of an extended investigation into whether PR #6389 resolves the specific behavior reported in #4481/#4513 (read time escalating over successive iterations), beyond the general performance benchmarks posted earlier in this thread.

### 1. Baseline performance win (high confidence, independently verified)

PR #6389's chunk-buffer pre-sizing gives a real, consistent **21-28% faster** read throughput for compressed chunked datasets, verified across multiple scales and confirmed at the source level: `H5Zdeflate.c`'s decompression loop starts its buffer at `*buf_size` and only grows it if insufficient — the fix makes it sufficient from the start, eliminating the realloc-doubling loop entirely. This holds regardless of anything below.

### 2. Reproducing the escalating-slowdown symptom specifically

A first synthetic benchmark (matched read/storage types, single-axis stride) showed no escalation in either baseline or fix — flat in both. It was missing key ingredients.

A **faithful reproduction**, using the issue's actual `unlimDsetRead`/`hypersladDsetRead` code (including `H5T_NATIVE_FLOAT` reads from `H5T_NATIVE_DOUBLE` data, and a 3D stride-`{2,2,2}` hyperslab), did reproduce real escalation:

| | Baseline (pre-fix) | Fix (PR #6389 alone) |
|---|---|---|
| `hypersladDsetRead` (B) rise | +22.6% | +10.9% |
| `unlimDsetRead` (A) rise | +7.8% | +5.1% |

**PR #6389 roughly halves the escalation but doesn't eliminate it**, in this repeated-`H5Fopen`/`H5Fclose` scenario matching the original repro's structure.

### 3. Isolating where the residual escalation actually lives

Ruled out two plausible culprits by reading the code: type-conversion buffers (`H5Dio.c`, uses `H5FL_BLK_MALLOC` — free-list) and the Extensible Array chunk index (`H5EAhdr.c`, uses `H5FL_FAC_MALLOC` — also free-list). Neither explained it.

An experimental patch extending the free list to filtered chunk buffers (not shippable — it breaks third-party filter plugins that manage their own buffers via raw `malloc`/`realloc`/`free`) gave only a small improvement (10.9%→9.4%), confirming raw allocation of filtered chunks is *a* factor but not the dominant one.

**The decisive diagnostic**: open the file once and read repeatedly, vs. reopening every iteration.

| Scenario | Baseline | Fix (PR #6389 alone) |
|---|---|---|
| Reopen every iteration | +22.6% / +7.8% | +10.9% / +5.1% (residual) |
| **File opened once** | +57.8% / +12.7% | **~0% / ~0%** (fully eliminated) |

This proved PR #6389 **fully** resolves the chunk-read-path escalation — the residual only appears when repeatedly reopening, meaning it's tied to something that runs on every `H5Fopen`/`H5Dopen`, not on repeated chunk reads.

### 4. The real remaining driver: speculative metadata loads, same pattern, different subsystem

`H5C__load_entry` (`H5Centry.c`) implements a "speculative load" mechanism for metadata cache entries whose on-disk size isn't known upfront: read an initial guessed size, then `H5MM_realloc` (raw, not free-list) if the guess was wrong. This is architecturally the *same bug pattern* as the original chunk-buffer issue, just in the metadata cache (`H5AC`/`H5C`) instead of the chunk cache (`H5D`) — a subsystem PR #6389 never touches.

Two fixed, small initial-size guesses were found and tested by simply increasing them:

| Constant | Location | Purpose | Original |
|---|---|---|---|
| `H5O_SPEC_READ_SIZE` | `H5Opkg.h` | Object header initial read (every `H5Dopen`) | 512 bytes |
| `H5HL_SPEC_READ_SIZE` | `H5HLcache.c` | Local heap initial read (link-name resolution) | 512 bytes |

Both are plausibly undersized for any dataset with real complexity (chunked + filtered + named + grouped, as ours are). Increasing both to 4096 bytes progressively closed the gap:

| Configuration | A rise | B rise |
|---|---|---|
| Pure fix (PR #6389 only) | +5.1% | +10.9% |
| + object header fix (512→4096) | +0.8% | +7.7% |
| + local heap fix (512→4096) | **-3.0%** (flat/noise) | **+5.0%** |

Dataset A (unlimited-dim, Extensible Array index) is now fully resolved. Dataset B (no unlimited dim, simpler structure, but larger ~547KB chunks and a 3D strided selection) continues to shrink with each fix but retains a smaller residual (~5%), suggesting at least one more small contributor exists, not yet identified — the pattern of roughly halving with each fix suggests diminishing but real remaining sources of the same kind.

### Net assessment

- **PR #6389 is a real, substantial, verified fix**, independently confirmed to (a) deliver 21-28% faster reads regardless of the escalation question, and (b) **fully** eliminate the escalating-slowdown behavior in the chunk-read path itself (proven decisively by the open-once test).
- **The residual escalation that survives in the reopen-every-iteration scenario is a second, distinct issue** — the same "small fixed guess + raw `realloc`" pattern recurring independently in HDF5's metadata cache (object header and local heap loading), unrelated to filters or compression, affecting essentially any dataset opened repeatedly on Windows. This is out of scope for PR #6389 and would need its own issue/fix — likely either enlarging these speculative-read defaults or (better, but bigger) routing metadata cache image buffers through a free-list the way `H5FL` already does for other hot allocation paths in this codebase.
- Both metadata-cache constant increases here were **quick experimental verification changes**, not vetted for correctness/compatibility (e.g., interaction with very small files, non-default page-buffering configs) — they're evidence of the mechanism, not a proposed patch.

brtnfld:
## Write-path repeated trials: confirms no measurable difference (issue #3858)

Following up on the single-run write-path comparison — ran 5 repeated trials each (baseline and pure PR #6389, same repro as before: 2000×512×512 chunked+deflate, 10 iterations per trial) to check whether the earlier ~5.5% gap was real or noise.

| | Baseline | Fix (pure PR #6389) |
|---|---|---|
| Trial totals (s) | 67.19, 66.71, 67.51, 66.61, 68.30 | 67.52, 67.91, 67.12, 67.66, 68.76 |
| **Mean total** | **67.26s** | **67.79s** |
| Stdev | 0.69s | 0.61s |
| Mean iter0 → iter9 | 4.015s → 9.162s | 4.060s → 9.259s |
| **Mean escalation (iter0→iter9)** | **+128.2%** | **+128.1%** |

**The gap is noise.** Mean difference is 0.53s (0.8% of baseline) — well inside one standard deviation, t-stat ≈ 1.29 (not significant at any conventional threshold with n=5 per group). The escalation curves are effectively identical between baseline and fix (128.2% vs 128.1% rise — indistinguishable).

### Conclusion

**PR #6389 has no measurable effect, positive or negative, on this write-side escalation scenario.** The earlier single-run ~5.5% gap was exactly what I flagged it as: run-to-run noise, not a real regression. This is consistent with the mechanism argument: #6389's `buf_alloc` change is in the read-side decompression path (`H5D__chunk_lock`), which a one-shot write of a brand-new dataset shouldn't exercise.

**Net effect on the two related-but-distinct issues:**
- **#4481 (read escalation):** PR #6389 substantially reduces it (and fully eliminates it in the open-file-once scenario) — real, verified improvement.
- **#3858 (write escalation):** PR #6389 is neutral — doesn't fix it, doesn't worsen it. The write-side escalation is a separate, unaddressed bug with its own (not yet identified) mechanism, likely unrelated to filter-pipeline decompression since a single `H5Dwrite` call doesn't involve `H5Zdeflate.c`'s read path at all.

"does it make Windows performance worse"? at least for pure PR #6389 without the H5O/H5HL follow-on experiments — on this machine/toolchain, no.

fortnern:
Maybe you could try to isolate the slowdown by unrolling the loop in the benchmark, and creating a small wrapper function for each write call (write1(), write2(), etc.). That way you can easily see in a profiler what contributes to the time for each iteration and you should be able to isolate what's causing the slowdown.

fortnern:
At any rate, is there a problem with this change itself, even if it doesn't fix the issue it was supposed to fix? If not, I think we should merge it and keep the original issue open.

jhendersonHDF:
In the last few rounds of testing done by @bmribler before @brtnfld's investigations, it appeared that the changes made Windows `H5Dwrite()` performance roughly twice as slow as before the changes, but potentially eliminated the increasing `H5Dwrite()` time issue in https://github.com/HDFGroup/hdf5/issues/3858. The latest testing seems to show no difference in Windows `H5Dwrite()` performance but also potentially no change to the increasing time issue either. If the current changes in this PR are now separate from that issue and also show an improvement with no performance regression, then it should be merged.

src/H5Zscaleoffset.c
```diff
@@ -1187,7 +1187,7 @@ H5Z__filter_scaleoffset(unsigned flags, size_t cd_nelmts, const unsigned cd_valu
 
         /* no need to process data */
         if (scale_factor == (int)(cd_values[H5Z_SCALEOFFSET_PARM_SIZE] * 8)) {
-            ret_value = *buf_size;
+            ret_value = nbytes;
```
bmribler: HGOTO_DONE(nbytes) for consistency?

bmribler (changes requested):

src/H5Zscaleoffset.c
brtnfld: updated

bmribler:
Can #3858 be closed as a side-effect fix?

bmribler (approved):

fortnern (approved):

fortnern:
> Can #3858 be closed as a side-effect fix?

Jordan's comment suggests that issue isn't fixed by this PR

brtnfld:
@jhendersonHDF, from Binh-Minh:

From 4481    475cf6cea0 Use HGOTO_DONE(nbytes) for the scaleoffset no-op passthrough
Execution time: 18.6551 seconds
Execution time: 21.231 seconds
Execution time: 21.921 seconds
Execution time: 22.7077 seconds
Execution time: 21.5929 seconds
Execution time: 21.3189 seconds
Execution time: 20.8187 seconds
Execution time: 20.3429 seconds
Execution time: 20.0046 seconds
Execution time: 20.2479 seconds
 
From develop branch
Execution time: 14.1031 seconds
Execution time: 20.0717 seconds
Execution time: 21.3375 seconds
Execution time: 24.5474 seconds
Execution time: 25.5938 seconds
Execution time: 24.3789 seconds
Execution time: 24.7894 seconds
Execution time: 25.5692 seconds
Execution time: 24.2109 seconds
Execution time: 23.1343 seconds

PR #6400, opened 2026-05-06
Improve performance of h5trav interfaces for links to objects

### LINKED ISSUE #6399: HDF5 tools performance issue for multiply-linked objects
The h5trav interfaces used by various HDF5 tools (`h5dump`, `h5ls`, `h5diff`, `h5repack`, `h5stat` and `h5format_convert`) to traverse file structures keep an internal array of objects with a reference count greater than 1 in order to avoid redundant processing on already-visited objects linked to elsewhere with hard links. Checking if an object has already been visited performs a linear scan over this array that compares object "tokens" with `H5Otoken_cmp()` until a matching object is found. For file structures that contain many objects with at least two hard links to them (including the original hard link when the object was created), this can result in behavior that is quadratic or near-quadratic with the number of objects.

Additionally, `h5repack` uses an h5trav interface which keeps a separate internal array of hard link name aliases for already-visited objects. This array is used so that processing can avoid accidentally expanding a hard link into a duplicate object in a repacked file. Checking for visited objects here also performs a linear scan over the array, comparing object "tokens" with `H5Otoken_cmp()` until a matching object is found. For files with structures similar to the one mentioned above, this can approach 2n<sup>2</sup> complexity when combined with the linear scan from above.

Quick initial profiling of repacking a file with ~500,000 objects, where most of the objects are linked with hard links at least twice, using no options for `h5repack` showed ~95% of the runtime being spent in `H5Otoken_cmp()` operations. Replacing these arrays with hash tables greatly improves the performance of the HDF5 tools above for specific file structures.

Note that a similar performance issue may exist for soft/external links in the `symlink_visit_add()` and `symlink_is_visited()` functions in `tools/lib/h5trav.c`, but this path hasn't been tested yet.

Running `h5repack` on the above file with no options:

<b>Current</b>

Manually aborted after 3 hrs, the entirety of which was spent setting up the table of objects to traverse (no actual repacking had occurred yet)

<b>With hash table replacement:</b>

real	5m23.073s
user	3m10.633s
sys	0m47.585s

Running `h5stat` on the same file:

<b>Current</b>

Manually aborted after 1 hr, the entirety of which was spent setting up the table of objects to traverse

<b>With hash table replacement:</b>

real	0m32.472s
user	0m29.419s
sys	0m3.042s

jhendersonHDF:
For objects with multiple hard links, use hash table to map between object tokens and pathnames during traversal to avoid linear scan over all previous objects for each hard link seen

Use separate hash table for h5trav "table" interface to map between object tokens and an index into the table of visited objects. This facilitates quick lookups of objects when adding hard link name aliases for h5repack processing.

See the linked issue for context

### DIFF
--- a/tools/lib/h5trav.c
+++ b/tools/lib/h5trav.c
@@ -169,7 +169,7 @@ trav_token_visited(hid_t loc_id, trav_seen_hash_t *objects_seen, H5O_token_t *to
 
     HASH_FIND(hh, objects_seen, token, sizeof(H5O_token_t), entry);
 
-    if (visited_obj_ret)
+    if (entry && visited_obj_ret)
         *visited_obj_ret = &entry->obj;
 
     return (entry != NULL);
--- a/tools/lib/h5trav.c
+++ b/tools/lib/h5trav.c
@@ -14,20 +14,34 @@
 #include "h5tools.h"
 #include "H5private.h"
 
+/* Replace uthash's default key comparison function with a wrapper around H5Otoken_cmp */
+#undef HASH_KEYCMP
+#define HASH_KEYCMP(a, b, len) trav_token_visited_cmp(loc_id, (const H5O_token_t *)a, (const H5O_token_t *)b)
+
 /*-------------------------------------------------------------------------
  * local typedefs
  *-------------------------------------------------------------------------
  */
-typedef struct trav_addr_path_t {
-    H5O_token_t token;
-    char       *path;
-} trav_addr_path_t;
+/* Structure for tracking visited objects in a hash table for
+ * quicker lookups to determine when an object has already been
+ * visited
+ */
+typedef struct trav_seen_hash_t {
+    trav_seen_t obj;
 
-typedef struct trav_addr_t {
-    size_t            nalloc;
-    size_t            nused;
-    trav_addr_path_t *objs;
-} trav_addr_t;
+    UT_hash_handle hh; /* Hash table handle */
+} trav_seen_hash_t;
+
+/* Structure for tracking the index into the table of objects
+ * where a visited object was placed to facilitate quicker
+ * lookups when adding path aliases
+ */
+typedef struct trav_table_hash_t {
+    H5O_token_t token;
+    size_t      index;
+
+    UT_hash_handle hh; /* Hash table handle */
+} trav_table_hash_t;
 
 typedef struct {
     h5trav_obj_func_t visit_obj; /* Callback for visiting objects */
@@ -36,7 +50,7 @@ typedef struct {
 } trav_visitor_t;
 
 typedef struct {
-    trav_addr_t          *seen;          /* List of addresses seen already */
+    trav_seen_hash_t     *objects_seen;  /* Hash table of objects seen already */
     const trav_visitor_t *visitor;       /* Information for visiting each link/object */
     bool                  is_absolute;   /* Whether the traversal has absolute paths */
     const char           *base_grp_name; /* Name of the group that serves as the base
@@ -56,9 +70,12 @@ typedef struct trav_path_op_data_t {
  * local functions
  *-------------------------------------------------------------------------
  */
-static void trav_table_add(trav_table_t *table, const char *objname, const H5O_info2_t *oinfo);
+static herr_t trav_table_add(trav_table_t *table, const char *objname, const H5O_info2_t *oinfo);
 
-static void trav_table_addlink(trav_table_t *table, const H5O_token_t *obj_token, const char *path);
+static herr_t trav_table_addlink(trav_table_t *table, const H5O_token_t *obj_token, const char *path,
+                                 const char *orig_path);
+
+static int trav_token_visited_cmp(hid_t loc_id, const H5O_token_t *token1, const H5O_token_t *token2);
 
 /*-------------------------------------------------------------------------
  * local variables
@@ -108,25 +125,32 @@ h5trav_set_verbose(int print_verbose)
  *
  * Purpose:  Add an object token to visited data structure
  *
- * Return:   void
+ * Return:   Non-negative on success/negative on failure
  *-------------------------------------------------------------------------
  */
-static void
-trav_token_add(trav_addr_t *visited, H5O_token_t *token, const char *path)
+static herr_t
+trav_token_add(trav_seen_hash_t **objects_seen_ptr, H5O_token_t *token, const char *path,
+               trav_seen_t **visited_obj_ret)
 {
-    size_t idx; /* Index of address to use */
+    trav_seen_hash_t *entry = NULL;
 
-    /* Allocate space if necessary */
-    if (visited->nused == visited->nalloc) {
-        visited->nalloc = MAX(1, visited->nalloc * 2);
-        visited->objs =
-            (trav_addr_path_t *)realloc(visited->objs, visited->nalloc * sizeof(trav_addr_path_t));
-    } /* end if */
+    if (NULL == (entry = malloc(sizeof(*entry))))
+        return FAIL;
+    if (NULL == (entry->obj.path = strdup(path))) {
+        free(entry);
+        return FAIL;
+    }
+    memcpy(&entry->obj.token, token, sizeof(H5O_token_t));
 
-    /* Append it */
-    idx = visited->nused++;
-    memcpy(&visited->objs[idx].token, token, sizeof(H5O_token_t));
-    visited->objs[idx].path = strdup(path);
+    /* HASH_ADD modifies what's pointed to by objects_seen_ptr when it
+     * initializes the hash table after being called for the first time
+     */
+    HASH_ADD(hh, (*objects_seen_ptr), obj.token, sizeof(H5O_token_t), entry);
+
+    if (visited_obj_ret)
+        *visited_obj_ret = &entry->obj;
+
+    return SUCCEED;
 } /* end trav_token_add() */
 
 /*-------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


tools/lib/h5trav.h
```diff
@@ -15,11 +15,6 @@
 
 #include "hdf5.h"
 
-/* Typedefs for visiting objects */
-typedef herr_t (*h5trav_obj_func_t)(const char *path_name, const H5O_info2_t *oinfo, const char *first_seen,
```
jhendersonHDF, line 19: These typedefs had to be moved down for the new `trav_seen_t` parameter to be available

tools/lib/h5trav.h
```diff
@@ -111,8 +115,16 @@ typedef struct trav_table_t {
     size_t      size;
     size_t      nobjs;
     trav_obj_t *objs;
+
+    /* Private data for this trav_table_t */
+    void *priv_data;
```
jhendersonHDF, line 121: This is really a pointer to a structure with a `UT_hash_handle`, but exposing the uthash API at the h5trav.h level is problematic due to its header-only nature and already being used in H5private.h. This is just a quick hack to hide the implementation details.

tools/lib/h5trav.c
```diff
@@ -14,20 +14,34 @@
...
+    trav_seen_t obj;
+
+    UT_hash_handle hh; /* Hash table handle */
+} trav_seen_hash_t;
+
+/* Structure for tracking the index into the table of objects
+ * where a visited object was placed to facilitate quicker
+ * lookups when adding path aliases
+ */
+typedef struct trav_table_hash_t {
```
jhendersonHDF, line 40: This structure is mostly for h5repack, which now has its own hash table separate from the main one used by other h5trav interfaces. I considered modifying the h5trav interfaces to allow h5repack to share a single hash table, but it would have been fairly awkward to do so and would have imposed unnecessary memory overhead for other tools that wouldn't need the extra information that would have been stored, so I instead just used a separate hash table specifically for the h5trav "table" interface.

tools/lib/h5trav.c
```diff
@@ -137,25 +161,39 @@ trav_token_add(trav_addr_t *visited, H5O_token_t *token, const char *path)
...
+ * Function: trav_token_visited_cmp
+ *
+ * Purpose:  Wrapper around H5Otoken_cmp for comparing objects in
+ *           trav_token_visited()
+ *
+ * Return:   -1/0/1 (similar to memcmp())
+ *-------------------------------------------------------------------------
+ */
+static int
+trav_token_visited_cmp(hid_t loc_id, const H5O_token_t *token1, const H5O_token_t *token2)
```
jhendersonHDF, line 249: This function is called by HASH_FIND and is just used to wrap around H5Otoken_cmp() instead of a plain `memcmp` of object token bytes

tools/lib/h5trav.c
```diff
@@ -244,37 +288,31 @@ static int
...
         /* Set up user data structure */
-        udata.seen          = &seen;
+        udata.objects_seen  = NULL;
         udata.visitor       = visitor;
         udata.is_absolute   = (*grp_name == '/');
         udata.base_grp_name = grp_name;
         udata.fields        = fields;
 
+        /* Check for multiple links to top group */
+        if (oinfo.rc > 1)
```
jhendersonHDF: Moved this down below the udata initialization in case the hash table gets initialized by the call to `trav_token_add()`.

tools/lib/h5trav.c
```diff
@@ -137,25 +161,39 @@ trav_token_add(trav_addr_t *visited, H5O_token_t *token, const char *path)
...
-        /* Check for token already in array */
-        if (H5Otoken_cmp(loc_id, &visited->objs[u].token, token, &token_cmp) < 0)
-            return NULL;
-        if (!token_cmp)
-            return (visited->objs[u].path);
-    }
+    HASH_FIND(hh, objects_seen, token, sizeof(H5O_token_t), entry);
+
+    if (visited_obj_ret)
+        *visited_obj_ret = &entry->obj;
```
mattjala, line 234: It doesn't really matter, since `trav_token_add` overwrites this value right after, but it would be slightly safer to have a check here to avoid populating `visited_obj_ret` with a bad value when `entry` is NULL on a hash table miss.

tools/lib/h5trav.c
```diff
@@ -244,37 +288,31 @@ static int
...
-        udata.seen          = &seen;
+        udata.objects_seen  = NULL;
         udata.visitor       = visitor;
         udata.is_absolute   = (*grp_name == '/');
         udata.base_grp_name = grp_name;
         udata.fields        = fields;
 
+        /* Check for multiple links to top group */
+        if (oinfo.rc > 1)
+            trav_token_add(&udata.objects_seen, &oinfo.token, grp_name, NULL);
```
mattjala: Should check return value here

mattjala (dismissed):
Only a couple minor issues

tools/lib/h5trav.c
jhendersonHDF: Good catch; I should have added an `if (entry)` check anyway

tools/lib/h5trav.h
fortnern: I wonder if it makes sense to just include H5private.h here

jhendersonHDF: Due to the header-only nature of uthash.h, it being included in H5private.h meant that having different settings for it in different parts of the library became somewhat difficult. In h5trav.c only, I needed to change change `HASH_KEYCMP` to use `H5Otoken_cmp()` to keep compatibility, but it became a first-to-include-uthash.h-wins scenario if I included H5private.h in h5trav.h since H5private.h is also including uthash.h with its own settings. The scope of including the uthash.h header should probably be reduced from H5private.h to where it's actually needed.

tools/lib/h5trav.c
```diff
@@ -108,25 +125,32 @@ h5trav_set_verbose(int print_verbose)
...
+    memcpy(&entry->obj.token, token, sizeof(H5O_token_t));
 
-    /* Append it */
-    idx = visited->nused++;
-    memcpy(&visited->objs[idx].token, token, sizeof(H5O_token_t));
-    visited->objs[idx].path = strdup(path);
+    /* HASH_ADD modifies what's pointed to by objects_seen_ptr when it
+     * initializes the hash table after being called for the first time
+     */
+    HASH_ADD(hh, (*objects_seen_ptr), obj.token, sizeof(H5O_token_t), entry);
```
fortnern: Are we certain that tokens never contain garbage bytes? I suspect so just want to make sure, since they are stored as "MAX_TOKEN_SIZE", implying that some connectors don't use all of the bytes.

fortnern: I see that H5VL_native_addr_to_token zeroes the unused bytes, just curious how strong our documentation is that third party VOLs should do the same.

tools/lib/h5trav.c
```diff
@@ -244,37 +288,33 @@ static int
...
         /* Set up user data structure */
-        udata.seen          = &seen;
+        udata.objects_seen  = NULL;
         udata.visitor       = visitor;
         udata.is_absolute   = (*grp_name == '/');
         udata.base_grp_name = grp_name;
         udata.fields        = fields;
 
+        /* Check for multiple links to top group */
+        if (oinfo.rc > 1) {
```
fortnern, line 388: Is this fixing a bug?

tools/lib/h5trav.c
```diff
@@ -642,76 +690,129 @@ h5trav_getindext(const char *name, const trav_table_t *table)
...
+    HASH_FIND(hh, ((trav_table_hash_t *)table->priv_data), obj_token, sizeof(H5O_token_t), entry);
+    if (entry) {
+        i = entry->index;
+
+        /* Make sure objects are the same */
+        if (orig_path && (0 != strcmp(table->objs[i].name, orig_path)))
+            return FAIL;
+    }
+    else {
         for (i = 0; i < table->nobjs; i++) {
```
fortnern, line 866: Would it be worth adding a hash table for the path names to accelerate this?

tools/lib/h5trav.c
jhendersonHDF: This is the same logic as above, just had to be moved down a bit.

fortnern (dismissed):

tools/lib/h5trav.c
jhendersonHDF: That's true, I did make an assumption here and I haven't found any specific documentation mentioning that the bytes should be zeroed out or set to a specific value yet. We do provide `H5O_TOKEN_UNDEF` for undefined token values which in theory would be an initializer, but is not really documented as such. Without that assumption and without knowing how many valid bytes there are, I suppose you can't do much with an object token other than pass it to the token APIs. To be 100% sure here, we would need to either document that, or may need to have a callback for returning a hashed version of the token.

tools/lib/h5trav.c
jhendersonHDF: Probably yes. I mostly tried to limit the scope of this to the performance problem I directly observed, but there's also likely a similar issue with soft links, as well as possibly with the path names here.

tools/lib/h5trav.c
```diff
@@ -99,34 +118,94 @@ h5trav_set_verbose(int print_verbose)
...
+ * Function: trav_token_get_size
+ *
+ * Purpose:  Given an object ID, retrieves the size of an object token from
+ *           the VOL connector the object is opened with
+ *
+ * Return:   Non-negative on success/negative on failure
  *-------------------------------------------------------------------------
  */
+static herr_t
+trav_token_get_size(hid_t obj_id, size_t *token_size)
```
jhendersonHDF, line 130: A bit of an ugly hack, but about the only way that the tools code can ask a VOL connector for the size of object tokens, similar to what the reference API does

tools/lib/h5trav.c
```diff
@@ -762,10 +947,22 @@ trav_table_init(hid_t fid, trav_table_t **tbl)
...
+        table->fid       = fid;
+        table->size      = 0;
+        table->nobjs     = 0;
+        table->objs      = NULL;
+        table->priv_data = NULL;
+
+        /* Determine how large an object token is so we can properly hash them */
+        H5E_BEGIN_TRY
+        {
+            status = trav_token_get_size(fid, &table->obj_token_size);
```
jhendersonHDF, line 961: This does result in a second call to `trav_token_get_size()` where `traverse()` would already have called it, but it keeps the usage of the token size to just the `trav_table_` interfaces where it's used, as opposed to having to pass it through the `h5trav_obj_func_t` and `h5trav_lnk_func_t` callbacks where the other `trav_` interfaces wouldn't be using the parameter.

mattjala (approved):

lrknox (approved):

jhendersonHDF:
Created https://github.com/HDFGroup/hdf5/issues/6514 as a follow on issue for documenting the expectations around object tokens

PR #6237, opened 2026-02-25
Update AOCC workflow concurrency group

jhendersonHDF:
Prevent AOCC workflows from being cancelled when pushes are made to develop branch on merge of PRs

----

> [!IMPORTANT]
> Update concurrency group in `aocc.yml` to prevent workflow cancellation on develop branch pushes during PR merges.
> 
>   - **Concurrency Group Update**:
>     - In `.github/workflows/aocc.yml`, update concurrency group to `${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}`.
>     - Prevents workflow cancellation on pushes to the develop branch during PR merges.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=HDFGroup%2Fhdf5&utm_source=github&utm_medium=referral)<sup> for 15c0f92270e24ee4b17a2c90a799e1d80025b45b. You can [customize](https://app.ellipsis.dev/HDFGroup/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #6405, opened 2026-05-13
CI: fix concurrency group to cancel stale runs on branch push

brtnfld:
## Summary

- Fixes the `concurrency` group expression in 8 workflow files so that stale runs are cancelled when a new push arrives on the same branch or PR
- Uses `${{ github.event.pull_request.number || github.ref }}` instead of the previous `${{ github.head_ref && github.ref || github.run_id }}`

## Root cause

`github.head_ref` is only populated for `pull_request` events — it is empty for direct branch pushes. The old ternary therefore fell back to `github.run_id` (unique per run) for every branch push, making each run its own group and preventing cancellation entirely.

## Why not just `github.ref`?

Using `github.ref` alone would group all pushes to `develop` (or any shared branch) together, causing one contributor's push to cancel another contributor's in-flight run on the same branch.

## The fix

`${{ github.event.pull_request.number || github.ref }}`

| Event | Group key | Effect |
|---|---|---|
| `pull_request` | PR number (e.g. `6405`) | New pushes to the same PR cancel old runs; different PRs never interfere |
| Branch push | `refs/heads/my-branch` | New pushes to the same branch cancel old runs; different branches never interfere |

## Test plan

- [ ] Push to a branch with an open PR — confirm the old PR run is cancelled
- [ ] Push directly to a branch without a PR — confirm the old branch run is cancelled
- [ ] Verify pushes to different branches do not cancel each other's runs

### DIFF
--- a/.github/workflows/call-workflows.yml
+++ b/.github/workflows/call-workflows.yml
@@ -25,7 +25,7 @@ on:
 
 # Using concurrency to cancel any in-progress job or run
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 permissions:
--- a/.github/workflows/call-workflows.yml
+++ b/.github/workflows/call-workflows.yml
@@ -25,7 +25,7 @@ on:
 
 # Using concurrency to cancel any in-progress job or run
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
   cancel-in-progress: true
 
 permissions:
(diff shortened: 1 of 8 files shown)


jhendersonHDF (changes requested):
Just using `github.ref` is not adequate because it can cause unintentional cancellation of other people's actions. See https://github.com/HDFGroup/hdf5/pull/6237/files#r2850205074. `github.head_ref && github.ref` is intentional for various reasons and should be left.

jhendersonHDF:
`github.event.pull_request.number` also isn't correct either because it covers the `pull_request` event and not following pushes to an existing PR.

brtnfld:
Closing — the original concurrency group expression `github.head_ref && github.ref || github.run_id` is intentional. It correctly cancels in-progress runs for PR events while deliberately avoiding cancellation of direct branch pushes (to prevent unintentional cross-contributor cancellation). See #6237 for context.

PR #6408, opened 2026-05-15
config/HDF5UseZLIB.cmake: resolve ALIASED_TARGET before re-aliasing

brtnfld:
## Summary

- Fix CMake configure failure when both `HDF5_ENABLE_ZLIB_SUPPORT=ON` and `HDF5_USE_ZLIB_NG=ON` are set.

## Reproducing

```bash
cmake -S . -B build \
  -C config/cmake/cacheinit.cmake \
  -DHDF5_ENABLE_ZLIB_SUPPORT=ON \
  -DHDF5_USE_ZLIB_NG=ON \
  -DZLIB_USE_LOCALCONTENT=OFF
```

aborts with:

```
CMake Error at config/HDF5UseZLIB.cmake:476 (add_library):
  add_library cannot create ALIAS target "hdf5::zlib-ng-static" because
  target "zlib-ng-static" is itself an ALIAS.
  config/HDF5UseZLIB.cmake:538 (external_zlib_ng_library)
  CMakeLists.txt:1030 (include)
```

## Root cause

Upstream zlib-ng's CMake exports public target names (notably `zlib-ng-static`) as aliases of the real underlying target. In `external_zlib_ng_library()` we walk the exported target names and create a namespaced `hdf5::<name>` alias for each:

```cmake
add_library (${HDF_PACKAGE_NAMESPACE}${zlib_ng_target} ALIAS ${zlib_ng_target})
```

CMake refuses `add_library(NEW ALIAS EXISTING)` when `EXISTING` is itself an alias — aliases cannot chain.

## Fix

Query `ALIASED_TARGET` before creating the namespaced alias, and point the new alias at the real underlying target:

```cmake
get_target_property (_aliased_target ${zlib_ng_target} ALIASED_TARGET)
if (_aliased_target)
  add_library (${HDF_PACKAGE_NAMESPACE}${zlib_ng_target} ALIAS ${_aliased_target})
else ()
  add_library (${HDF_PACKAGE_NAMESPACE}${zlib_ng_target} ALIAS ${zlib_ng_target})
endif ()
```

When the target is not an alias, behavior is unchanged.

## Why we don't see this on `develop`

`HDF5_ENABLE_ZLIB_SUPPORT` defaults to OFF; the `external_zlib_ng_library()` path at `HDF5UseZLIB.cmake:522` is only entered when both `HDF5_ENABLE_ZLIB_SUPPORT` *and* `HDF5_USE_ZLIB_NG` are ON. The "gcc use zlib-ng filter" CI job currently passes only the latter, dodging the alias path. Any user explicitly enabling both — or any future workflow change that turns `HDF5_ENABLE_ZLIB_SUPPORT` ON in that job — runs into this.


config/HDF5UseZLIB.cmake
```diff
@@ -469,11 +469,21 @@ function (external_zlib_ng_library)
...
+  # underlying target; CMake refuses add_library(NEW ALIAS EXISTING)
+  # when EXISTING is itself an alias (aliases cannot chain).  Resolve
+  # through ALIASED_TARGET before creating the namespaced alias.
   if (HDF_PACKAGE_NAMESPACE)
     foreach (zlib_ng_target ${zlib_ng_targets})
       if (NOT TARGET ${HDF_PACKAGE_NAMESPACE}${zlib_ng_target})
-        add_library (${HDF_PACKAGE_NAMESPACE}${zlib_ng_target} ALIAS ${zlib_ng_target})
+        get_target_property (_aliased_target ${zlib_ng_target} ALIASED_TARGET)
+        if (_aliased_target)
+          add_library (${HDF_PACKAGE_NAMESPACE}${zlib_ng_target} ALIAS ${_aliased_target})
```
jhendersonHDF, line 488: I think there's a question of whether or not these ALIAS targets are actually used by anything, but it's also probably worth noting that the same logic exists for both regular zlib and libaec, where this should probably be applied as well.

brtnfld: The aliases are stored in H5_ZLIB_LIBRARY/H5_SZIP_LIBRARY, folded into LINK_COMP_LIBS, and passed to target_link_libraries for the main HDF5 library, tests, and tools.

I've updated the other places.

jhendersonHDF (approved):

lrknox (approved):

PR #6411, opened 2026-05-19
fix(H5Zdeflate): suppress C4267 warning

hyoklee:
On MSVC ARM64, `size_t` is 64-bit while zlib's `uLong` is 32-bit, 
causing C4267 (conversion from 'size_t' to 'uLong', possible loss of data).


src/H5Zdeflate.c
```diff
@@ -176,7 +176,7 @@ H5Z__filter_deflate(unsigned flags, size_t cd_nelmts, const unsigned cd_values[]
         size_t z_dst_buf_size = zng_compressBound(nbytes); /* 5730 */
         size_t z_dst_nbytes   = z_dst_buf_size;
 #else
-        uLongf z_dst_buf_size = (uLongf)compressBound(nbytes);
+        uLongf z_dst_buf_size = (uLongf)compressBound((uLong)nbytes);
```
jhendersonHDF, line 181: A check should be added to make sure that casting `nbytes` to `uLong` doesn't actually lose data

```
if ((size_t)(uLong)nbytes != nbytes)
    error
```

jhendersonHDF (approved):

fortnern (approved):

PR #6413, opened 2026-05-20
Fix bad reads of nested cmpd/vlen types in JNI

### LINKED ISSUE #6467: JNI mishandles compound and variable-length datatypes
The HDF5 Java JNI layer has several related bugs in how it translates variable-length (VLEN) and compound data between native buffers and Java objects. The most visible issue is that calling H5DreadVL on a compound-of-vlen dataset returns empty data when the caller pre-allocates each row slot as an ArrayList (the pattern HDFView's H5Datatype.allocateArray uses), even though the read reports no error. The same read succeeds if the row slots are left null.

The root cause is in translate_rbuf's H5T_COMPOUND case, which branches on a found_jList flag and, when a slot is pre-allocated, tries to set elements using built-in Java array syntax on what is actually an ArrayList, causing a silent no-op.

The same found_jList confusion affects the H5T_VLEN read case, where reading a vlen sequence into a null (non-pre-allocated) slot incorrectly invokes .add() on a plain object array. On the write side, translate_atomic_wbuf's H5T_VLEN case treats an incoming ArrayList as a Java array, producing a garbage length and causing a SIGSEGV when writing compound-of-vlen sequences.

Additionally, translate_atomic_rbuf/translate_atomic_wbuf mishandle nested compounds. The read path discards recursive return values (yielding empty lists), and both paths compute an incorrect member offset by including a spurious i * typeSize term.

Resolved by #6413

mattjala:
- Fixed H5DreadVL failing to read data for pre-allocated compound-of-vlen datasets

When H5DreadVL is called with a `H5T_COMPOUND` memory type whose members include variable-length data, and the caller pre-allocates each row slot in the outer `Object[]` buffer, then the read completes without throwing an error, but each pre-allocated row ArrayList comes back empty.

If the caller follows the alternate calling pattern, leaving each row slot null and letting the native code allocate the per-row record, then the same read works correctly.

The root cause is the `H5T_COMPOUND` case in `translate_rbuf`. This function walks each row's compound, decodes each member into a Java object via `translate_atomic_rbuf`, and stores the result in the per-row ArrayList. It branches on a `found_jList` flag based on the result of `GetObjectArrayElement()`, i.e. whether the caller pre-allocated the row slot.

Later on, this flag is used to decide between using the `.add()` method of ArrayList (if `found_jList` is false), or trying to directly set an element on a builtin java array (if found_jList is true).

The latter case is wrong in two ways.

1. Type mismatch. `jList` is always an ArrayList, either fetched from ret_buf[i] (where each slot is an ArrayList) or newly created a few lines earlier. The attempt to directly set an element using the built-in array syntax silently no-ops.
2. Wrong index variable. Even if jList were a Java array, the index supplied is i (the outer row counter), not x (the inner member counter).

There are currently no tests covering `H5DreadVL()` for H5T_COMPOUND in the src-jni test tree. This was discovered in the context of HDFView attempting to read compound-of-sequence datasets. HDFView's H5Datatype.allocateArray pre-fills each row slot with new ArrayList<>() before handing the buffer to H5DreadVL, which triggered the bug.

My fix is to replace the use the ArrayList.add() method in every case, and drop the found_jList tracking since jList is always an ArrayList and we always want to use the .add() method. This also causes it to match the pattern of the H5T_VLEN branch, which works for both pre-allocated and non-pre-allocated slots.

To test this case, I added `test.TestH5D.testH5Dread_compound_of_vlen_preallocated`. For the sake of regression testing, I also added testH5Dread_compound_of_vlen which tests the same thing with non-pre-allocated row slots, although this test passed even before I made any changes to the source code.

I also fixed a very similar problem in H5DwriteVL that led to a SIGSEGV when writing compound of variable-length sequences. Specifically, `translate_atomic_wbuf` treats `in_obj` as a java array when it is an ArrayList, producing a garbage length value that eventually leads to a crash.

I changed the H5T_VLEN case to behave similarly to the H5T_ARRAY and H5T_COMPOUND cases and use `mToArray` on the provided ArrayList to get an array object for the array operations. The test verifying the fix is `testH5Dwrite_compound_of_vlen`.

---

`translate_rbuf`'s H5T_VLEN case had a similar bug where when `found_jList` was set to false due to an entry in `ret_buf` being null. `ret_buf.add()` would then be invoked on an array of objects without the list .add() method. This would occur whenever a read was invoked of a vlen sequence with a null (non-preallocated) entry. The pre-existing tests only tested the pre-allocated cases.

I removed the use of the `found_jList` flag, since it conflated the passing of an unallocated slot with `ret_buf` not being an array. Instead, it now uses `ret_buflen == 0` as the check to match the pattern in H5T_INTEGER and other branches.

The test for this fix is `testH5Dread_vlen_of_compound_nullslot`.

---

`translate_atomic_rebuf` had two issues related to handling of nested compounds. First, it discarded recursive return values, resulting in the construction of empty lists. Secondly, its member offset (`char_buf + i * typeSize + memb_offset`) was incorrect. In this case, `i` was the member index and `memberSize` was the entire cmpd size, so the offset would be erroneously large. It seems like this came from copying of the offset computation from `translate_rbuf`, which had to advance over entire elements of compound data. This error was duplicated on the write side in `translate_atomic_wbuf`'s H5T_COMPOUND case (h5util.c:4611).

I changed `translate_atomic_rbuf` to capture the resultant object, and dropped the `i * typeSize` term in both routines.
The new test verifying the fix is `testH5Dread_vlen_of_nested_compound`.

Fixes #6467


jhendersonHDF:
After looking over this for a while, I have several comments, some related to the changes in this PR and some unrelated:

- I believe support for the pre-allocated variable-length sequence case should be entirely removed. It was clearly not implemented correctly, it isn't advertised anywhere in example code as far as I can tell and it simply complicates the logic with a pattern that isn't used anywhere else in the API (callers don't pre-allocate buffers for reads of variable-length data in the C API). I believe trying to support this pattern came about in https://github.com/HDFGroup/hdfview/commit/06e888c481b886069c157e5ba5748c3c7cabcfed when changing attributes to be represented with real byte data rather than all strings. If HDFView is making use of this pattern, it should probably be changed.
- `translate_rbuf()` and `translate_wbuf()` should have included the size of the buffer to prevent mistakes such as the indexing issue with `char_buf + i * typeSize + memb_offset`.
- Due to many of these cases working previously in HDFView 3.1.1 (the only one I've tested so far), I believe this is a case of the data model contract between the JNI (HDF5 side of things) and the object library (HDFView/Java side of things) becoming broken at some point. From that, it seems to me that the JNI code should be doing a much more thorough job of verifying, immediately at the API level, that the buffer given to it for reading/writing is in the expected (and documented somewhere) format. These mismatches are responsible for display issues/crashes in HDFView the vast majority of the time. While complex compound types could incur some overhead for this processing, I believe it's worth it to pay that price in order to be able to catch problems before they cause a crash.
- It's apparent that automated UI testing of HDFView with SWTBot needs to be setup somewhere once again, as it would have caught all of these problems the moment things broke. This testing stopped when testing was moved to GH CI, but there are many test files under https://github.com/HDFGroup/hdfview/tree/master/hdfview/src/test/resources/uitest that were previously used to catch problems.

nbagha1:
@jhendersonHDF  - 10294

java/src-jni/jni/h5util.c
```diff
@@ -4576,22 +4613,33 @@ translate_atomic_wbuf(JNIEnv *env, jobject in_obj, jlong mem_type_id, H5T_class_
             if (!(vlSize = H5Tget_size(memb)))
                 H5_LIBRARY_ERROR(ENVONLY);
 
-            /* Convert element to a vlen element */
-            hvl_t vl_elem;
-
-            jsize jnelmts = ENVPTR->GetArrayLength(ENVONLY, in_obj);
```
jhendersonHDF, line 4582: As a future minor optimization, if it's known that this will be a flat ArrayList, just calling `.size()` would be a little better than converting to an array just to get the length. This also looks like another case of losing track of / not validating the expected format of the buffer before operating on it.

java/src-jni/jni/h5util.c
```diff
@@ -4644,18 +4700,25 @@ translate_atomic_wbuf(JNIEnv *env, jobject in_obj, jlong mem_type_id, H5T_class_
             /* invoke the toArray method */
             if (mToArray == NULL)
                 CHECK_JNI_EXCEPTION(ENVONLY, JNI_FALSE);
-            jobjectArray array   = (jobjectArray)ENVPTR->CallObjectMethod(ENVONLY, in_obj, mToArray);
-            jsize        jnelmts = ENVPTR->GetArrayLength(ENVONLY, array);
```
jhendersonHDF, line 4648: Another future note, this particular case looks like the data model should be changed, as Object[] should cover this case without needing to involve ArrayList.

java/src-jni/hdf/hdf5lib/H5.java
```diff
@@ -194,6 +194,21 @@
  * and the parameter <i>data</i> can be any multi-dimensional array of numbers, such as float[][], or
  * int[][][], or Double[][].
  * <p>
+ * <b>Buffer data model</b>
```
jhendersonHDF, line 197: Could you also leave a similar header comment on the `translate_wbuf()` / `translate_rbuf()` functions so that the data model is documented right where the data is transformed? For those working in the JNI that will probably be easier to find.

java/src-jni/jni/h5util.c
```diff
@@ -4816,9 +4885,12 @@ translate_rbuf(JNIEnv *env, jobjectArray ret_buf, jlong mem_type_id, H5T_class_t
...
 
-    ret_buflen = ENVPTR->GetArrayLength(ENVONLY, ret_buf);
-    if (ret_buflen < 0)
-        H5_JNI_FATAL_ERROR(ENVONLY, "ret_buflen: Array length cannot be negative");
+    /* Top-level calls pass a Java Object[] array; recursive calls pass an
+     * ArrayList. Detect which so the append-vs-set decision is unambiguous.
+     *
+     * Caller-side pre-allocation of per-element slots is not supported.
+     * Any object already present in a Java array slot is overwritten. */
+    retIsList = ENVPTR->IsInstanceOf(ENVONLY, ret_buf, arrCList);
```
jhendersonHDF, line 4921: Future note: ideally, this should be split out so that `translate_rbuf()` isn't called recursively and doesn't add to the type confusion; separate methods should exist for separate types, even if those only shape the data into something appropriate for a common helper function.

java/src-jni/jni/h5util.c
```diff
@@ -5263,6 +5339,357 @@ translate_wbuf(JNIEnv *env, jobjectArray in_buf, jlong mem_type_id, H5T_class_t
...
+                H5Tclose(memb);
+                memb = H5I_INVALID_HID;
+            }
+
+            ENVPTR->DeleteLocalRef(ENVONLY, array);
+            break;
+        }
+        case H5T_INTEGER:
+        case H5T_ENUM:
+        case H5T_BITFIELD:
```
jhendersonHDF: Opaque types probably need to be left out of this particular verification or the checks need changed for them. Since they can be arbitrary size, you're not likely to find a matching Java type for them (they're just written as a byte[] for each element). It looks like `translate_atomic_wbuf()` is handling this incorrectly as well. I believe the logic should be the same as for `H5T_REFERENCE`.

jhendersonHDF (changes requested):
A couple minor comments but I think this is a solid improvement over the previous logic. The comments about future changes can be ignored for now, but should probably be dealt with eventually. There are also a few minor CI failures to be fixed.

jhendersonHDF (approved):

PR #6390, opened 2026-05-01
Disable nvhpc workflow

lrknox:
Disable calling nvhpc.yml to test hdf5 with nvhpc while dt_arith test fails.

PR #6415, opened 2026-05-22
Add flag -Mnovect for compiling test/dt_arith.c 

lrknox:
to work around LLVM crash in nvhpc.yml workflow.

Posted issue on https://forums.developer.nvidia.com/t/hdf5-develop-testing-on-github-with-nvhpc-26-3-0-fails-to-compile-test-dt-arith-c-crashing-llvm/371564.

We opted to add the flag to CMAKE_C-FLAGS in the workflow instead of adding it to the CMake target code for test/dt_arith.c to avoid forcing it on users for all versions of nvhpc.


hyoklee:
https://github.com/HDFGroup/hdf5/pull/6390

hyoklee (dismissed):

jhendersonHDF (approved):

hyoklee (approved):

PR #6352, opened 2026-04-05
Change h5repack low library bound to H5F_LIBVER_V18

ajelenak:
Closes #6348.

PR #6427, opened 2026-06-01
Cover all libver low bounds in h5repack metadata_block_size test (#6352 follow-up)

ajelenak:
After PR #6352 made `h5repack` default to H5F_LIBVER_V18 as the low bound,
the existing metadata block size option check broke because with V18+
the larger `meta_block_size` shrinks the output file rather than growing
it. Loop the check over every `H5F_LIBVER_*` low bound and assert the
size ordering appropriate to each regime.


hyoklee (approved):
s390x tests [passed](https://my.cdash.org/builds/3582987/tests?filters=%7B%22all%22%3A%5B%7B%22eq%22%3A%7B%22status%22%3A%22PASSED%22%7D%7D%5D%7D).

PR #6428, opened 2026-06-01
Improve HDF5_BUILD_PARALLEL_TOOLS docs and MFU not-found message

### LINKED ISSUE #2265: [Cmake Building Question] Parallel tools or Parallel ?
Hi,

I'm having trouble building the parallel HDF5 in order to use parallel NetCDF functionalities. It seems when I turn on both `HDF5_ENABLE_PARALLEL` and `HDF5_BUILD_PARALLEL_TOOLS` then it will try to find a MFU package.

My questions are:

1. What is the MFU library?
2. Do I need `HDF5_ENABLE_PARALLEL` or `HDF5_BUILD_PARALLEL_TOOLS` or both ?
3. What is the difference between `HDF5_ENABLE_PARALLEL` and `HDF5_BUILD_PARALLEL_TOOLS`?

Thanks

brtnfld:
- `CMakeBuildOptions.cmake`: expand the `HDF5_BUILD_PARALLEL_TOOLS` option description to name the tool it builds (`h5dwalk`), list all three required external libraries (MFU, CIRCLE, DTCMP), link to the mpiFileUtils project, and note that `HDF5_ENABLE_PARALLEL=ON` is also required.
- `config/cmake/FindMFU.cmake`: replace the dead/logic-inverted error block (wrong `VERBOSE`/`FATAL_ERROR` branching, reference to non-existent `MFU_INSTALL` variable) with a `STATUS` message that fires before `find_package_handle_standard_args` terminates configure. The message names all three required libraries, provides their GitHub URLs, and explains that `MFU_ROOT` is the environment variable to set for a standard mpiFileUtils install.Fixes #2265

### DIFF
--- a/CMakeBuildOptions.cmake
+++ b/CMakeBuildOptions.cmake
@@ -73,7 +73,7 @@ mark_as_advanced (HDF5_ENABLE_MAP_API)
 
 option (HDF5_BUILD_DOC "Build documentation" OFF)
 
-option (HDF5_BUILD_PARALLEL_TOOLS "Build parallel HDF5 file walker tool (h5dwalk); requires MFU, CIRCLE, and DTCMP HPC libraries (see https://github.com/hpc/mpifileutils). Also requires HDF5_ENABLE_PARALLEL=ON." OFF)
+option (HDF5_BUILD_PARALLEL_TOOLS "Build h5dwalk parallel tool. Requires HDF5_ENABLE_PARALLEL=ON and MFU/CIRCLE/DTCMP (mpiFileUtils)." OFF)
 mark_as_advanced (HDF5_BUILD_PARALLEL_TOOLS)
 
 option (HDF5_BUILD_TOOLS "Build HDF5 Tools" ON)
--- a/CMakeBuildOptions.cmake
+++ b/CMakeBuildOptions.cmake
@@ -73,7 +73,7 @@ mark_as_advanced (HDF5_ENABLE_MAP_API)
 
 option (HDF5_BUILD_DOC "Build documentation" OFF)
 
-option (HDF5_BUILD_PARALLEL_TOOLS "Build Parallel HDF5 Tools" OFF)
+option (HDF5_BUILD_PARALLEL_TOOLS "Build parallel HDF5 file walker tool (h5dwalk); requires MFU, CIRCLE, and DTCMP HPC libraries (see https://github.com/hpc/mpifileutils). Also requires HDF5_ENABLE_PARALLEL=ON." OFF)
 mark_as_advanced (HDF5_BUILD_PARALLEL_TOOLS)
 
 option (HDF5_BUILD_TOOLS "Build HDF5 Tools" ON)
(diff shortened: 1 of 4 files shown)


CMakeBuildOptions.cmake
```diff
@@ -73,7 +73,7 @@ mark_as_advanced (HDF5_ENABLE_MAP_API)
 
 option (HDF5_BUILD_DOC "Build documentation" OFF)
 
-option (HDF5_BUILD_PARALLEL_TOOLS "Build Parallel HDF5 Tools" OFF)
+option (HDF5_BUILD_PARALLEL_TOOLS "Build h5dwalk parallel tool. Requires HDF5_ENABLE_PARALLEL=ON and MFU/CIRCLE/DTCMP (mpiFileUtils)." OFF)
```
jhendersonHDF: For the sake of line width and if there may be other parallel-enabled tools in the future, I think it makes sense to keep this description something short like "Build parallel-specific HDF5 tools". Arguably, `ph5diff` and part of `h5perf` should have been moved under this option when it was created.

config/cmake/FindMFU.cmake
```diff
@@ -85,14 +85,12 @@ endif()
 
 # Report the results.
 if (NOT MFU_FOUND)
-  set (MFU_DIR_MESSAGE
-      "Mfu was not found. Make sure MFU_LIBRARY and MFU_INCLUDE_DIR are set or set the MFU_INSTALL environment variable."
+  message (STATUS "MFU (mpiFileUtils) was not found. "
```
jhendersonHDF: It seems like this should go back to being a `FATAL_ERROR` message unless we expect to be able to continue without these libraries. Alternatively, force the option back `OFF`, but I'm less fond of that pattern.

CMakeBuildOptions.cmake
brtnfld: what about "Build MPI-enabled HDF5 tools"

config/cmake/FindMFU.cmake
brtnfld: changed to FATEL_ERROR

CMakeBuildOptions.cmake
jhendersonHDF: Seems reasonable to me. May be more misleading as that would include `ph5diff` and `h5perf` as well, but I think that's maybe a separate discussion.

brtnfld: changed, it seems that this option was not fully thought through regarding its final intent.

jhendersonHDF (dismissed):

lrknox:
docs/AutotoolsToCMakeOptions.md line 62 also has the description "Build Parallel HDF5 Tools", and similarly docs/INSTALL_CMake_options.md line 198 has " enables building of parallel HDF5 tools".   Should those also be updated to "Build MPI-enabled HDF5 tools"?

jhendersonHDF:
> docs/AutotoolsToCMakeOptions.md line 62 also has the description "Build Parallel HDF5 Tools", and similarly docs/INSTALL_CMake_options.md line 198 has " enables building of parallel HDF5 tools". Should those also be updated to "Build MPI-enabled HDF5 tools"?

Yes, those should probably be updated as well

brtnfld:
done

jhendersonHDF (approved):

PR #6434, opened 2026-06-05
Fix stack buffer overflows in h5repack parse_filter (fixes #6433)

### LINKED ISSUE #6433: h5repack_parse.c: stack buffer overflows in filter string parser (scomp/stype[16] unbounded writes)
## Summary

`parse_filter()` in `tools/src/h5repack/h5repack_parse.c` contains four sites where a fixed 16-byte stack buffer (`scomp[16]` or `stype[16]`) is written without a bounds check. A malformed `-f` / `--filter` argument longer than 15 characters in the relevant field causes a stack buffer overflow (undefined behaviour, likely crash or stack corruption).

These are pre-existing bugs in the legacy parser code. The new TOML path for `UD=` filters added in branch 6153 is **not affected** — it added explicit length checks. The legacy path was left as-is.

## Affected sites

All in `tools/src/h5repack/h5repack_parse.c`:

### 1. `scomp[k]` — filter name before `=` (line ~115)

```c
for (i = (size_t)(end_obj + 1), k = 0, j = 0; i < len; i++, k++) {
    c        = str[i];
    scomp[k] = c;          /* written before any bounds check on k */
    if (c == '=' || i == len - 1) {
```

A filter name token longer than 15 chars (e.g. `--filter "AVERYLONGFILTERNAME=..."`) overflows `scomp[16]`.

### 2. `stype[m]` — SZIP pixels-per-block digits (line ~142)

```c
for (m = 0, u = i + 1; u < len; u++, m++) {
    ...
    if (l == -1)
        stype[m] = c;   /* m unbounded; stype is 16 bytes */
```

`h5repack -f "SZIP=12345678901234567,NN" input.h5 output.h5` overflows after 15 digits.

### 3. `stype[m]` — SOFF scale_factor digits (line ~192)

Same pattern as site 2, in the `SOFF=` branch:

```c
for (m = 0, u = i + 1; u < len; u++, m++) {
    ...
    if (l == -1)
        stype[m] = c;   /* m unbounded; stype is 16 bytes */
```

`h5repack -f "SOFF=12345678901234567,IN" input.h5 output.h5` overflows after 15 digits.

### 4. `stype[q]` — UD legacy numeric fields (line ~362)

```c
for (m = 0, q = 0, u = i + 1; u < len; u++, m++, q++) {
    if (str[u] == ',') {
        stype[q] = '\0';
        ...
        q = 0;
    }
    ...
    stype[q] = c;    /* q unbounded between commas; stype is 16 bytes */
}
```

Any individual field (filter id, flags, nelmts, or a cd_value) longer than 15 digits overflows `stype`.

## Severity

**Low.** Input comes from CLI arguments supplied by the local user; this is not reachable from file content. However, overflows can cause crashes and undefined behaviour in scripted pipelines, and the fix is straightforward.

Note: `smask[l]` in the same function is **not** affected — it has an explicit `if (l == 2)` stop guard.

brtnfld:
## Summary

Fixes #6433, three stack buffer overflow vulnerabilities in `parse_filter()` in `tools/src/h5repack/h5repack_parse.c`, reported in #6433.

- All five write sites for the `scomp[16]`, `stype[16]`, and `smask[16]` parse buffers are now guarded by a `PARSE_BUF_WRITE` macro that performs a bounds check before each character write.
- The macro takes an explicit `buf_sz` argument (passed as `sizeof(buf)` at every call site) to prevent silent pointer-decay if the buffers are ever refactored to heap allocations.
- In the SZIP and SOFF inner `u` loops, the prior code wrote bare `smask[l] = c` with no bounds check and used `i = len - 1` to stop the *outer* loop without breaking the *inner* loop. A trailing-garbage input such as `-f SZIP=16,NNxxx` could therefore write past `smask[2]`. Both sites now use `PARSE_BUF_WRITE` and add an explicit `break` to exit the inner loop as soon as the two-character mask is complete.
- The redundant `if (obj_list_ptr)` guard before `free()` is removed; `free(NULL)` is a safe no-op per C99 §7.20.3.2.

### DIFF
--- a/tools/src/h5repack/h5repack_parse.c
+++ b/tools/src/h5repack/h5repack_parse.c
@@ -161,7 +161,6 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
                             l++;
                             if (l == 2) {
                                 smask[l] = '\0';
-                                i        = len - 1; /* end outer loop */
                                 if (strcmp(smask, "NN") == 0)
                                     filt->cd_values[j++] = H5_SZIP_NN_OPTION_MASK;
                                 else if (strcmp(smask, "EC") == 0)
@@ -170,7 +169,8 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
                                     error_msg("szip mask must be 'NN' or 'EC' \n");
                                     exit(EXIT_FAILURE);
                                 }
-                                break; /* exit inner u loop; i=len-1 ends the outer loop */
+                                filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
+                                goto done_filter_params;
                             }
                         }
                     } /* u */
@@ -212,7 +212,6 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
                             l++;
                             if (l == 2) {
                                 smask[l] = '\0';
-                                i        = len - 1; /* end outer loop */
                                 if (strcmp(smask, "IN") == 0)
                                     filt->cd_values[j++] = H5Z_SO_INT;
                                 else if (strcmp(smask, "DS") == H5Z_SO_FLOAT_DSCALE)
@@ -221,7 +220,8 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
                                     error_msg("scale type must be 'IN' or 'DS' \n");
                                     exit(EXIT_FAILURE);
                                 }
-                                break; /* exit inner u loop; i=len-1 ends the outer loop */
+                                filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
+                                goto done_filter_params;
                             }
                         }
                     } /* u */
@@ -304,6 +304,7 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
                     filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
 
                 i += m; /* jump */
+done_filter_params:;
             }
             else if (i == len - 1) { /*no more parameters */
                 scomp[k + 1] = '\0';
--- a/tools/src/h5repack/h5repack_parse.c
+++ b/tools/src/h5repack/h5repack_parse.c
@@ -301,8 +301,6 @@ done_filter_params:;
                     j = 0;
                 else
                     filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
-
-                i += m; /* jump */
             }
             else if (i == len - 1) { /*no more parameters */
                 scomp[k + 1] = '\0';
--- a/tools/src/h5repack/h5repack_parse.c
+++ b/tools/src/h5repack/h5repack_parse.c
@@ -453,7 +453,7 @@ done_filter_params:;
              *-------------------------------------------------------------------------
              */
         case H5Z_FILTER_SZIP:
-            pixels_per_block = filt->cd_values[0];
+            pixels_per_block = filt->cd_values[1];
             if ((pixels_per_block % 2) == 1) {
                 if (obj_list)
                     free(obj_list);
--- a/tools/src/h5repack/h5repack_parse.c
+++ b/tools/src/h5repack/h5repack_parse.c
@@ -453,7 +453,7 @@ done_filter_params:;
              *-------------------------------------------------------------------------
              */
         case H5Z_FILTER_SZIP:
-            pixels_per_block = filt->cd_values[1];
+            pixels_per_block = filt->cd_values[H5Z_SZIP_PARM_PPB];
             if ((pixels_per_block % 2) == 1) {
                 if (obj_list)
                     free(obj_list);
--- a/tools/src/h5repack/h5repack_parse.c
+++ b/tools/src/h5repack/h5repack_parse.c
@@ -14,15 +14,16 @@
 #include "h5tools.h"
 #include "h5tools_utils.h"
 
-/* Bounds-check a running index against a fixed parse buffer before writing.
- * All three parse buffers (scomp, stype, smask) are the same size, so a
- * single macro covers every site.  Frees obj_list and exits on overflow,
- * consistent with all other error handling in parse_filter. */
-#define PARSE_BUF_WRITE(buf, idx, ch, obj_list_ptr, input_str)                                               \
+/* Bounds-check a running index against a parse buffer before writing.
+ * buf_sz must be passed explicitly (sizeof at the call site) to avoid the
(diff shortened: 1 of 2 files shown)


tools/src/h5repack/h5repack_parse.c
```diff
@@ -154,6 +170,7 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t
                                     error_msg("szip mask must be 'NN' or 'EC' \n");
                                     exit(EXIT_FAILURE);
                                 }
+                                break; /* exit inner u loop; i=len-1 ends the outer loop */
```
jhendersonHDF: As far as I can tell adding these breaks makes sense, but a simple `goto` might make more sense. That would avoid the addition to `i` on line 306 in case it would cause bad behavior in the future (such as being used to index into the string after the loop).

jhendersonHDF (dismissed):

tools/src/h5repack/h5repack_parse.c
```diff
@@ -438,7 +453,7 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t
              *-------------------------------------------------------------------------
              */
         case H5Z_FILTER_SZIP:
-            pixels_per_block = filt->cd_values[0];
+            pixels_per_block = filt->cd_values[1];
```
jhendersonHDF: Use H5Z_SZIP_PARM_PPB

tools/src/h5repack/h5repack_parse.c
```diff
@@ -274,18 +290,17 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t
...
                     stype[m] = '\0';
                 } /*if */
 
+done_filter_params:;
                 if ((strcmp(scomp, "UD") == 0) && (filt->cd_nelmts == 0))
                     j = 0;
                 else
                     filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
-
-                i += m; /* jump */
```
jhendersonHDF, line 288: The parsing logic here is somewhat complicated with unintuitive variable names, but it seems like this addition is still useful in some cases, just not the ones where `i` was previously being set to `len - 1`.

brtnfld: Traced this through carefully, including the pre-PR code on `develop` (before this PR touched the file at all): `i += m` was already dead in every branch, not just the cases that used to set `i = len - 1`.

The whole parameter-parsing section (SZIP/SOFF/UD/"all other filters") plus the filter-name translation chain sits inside one `if (c == '=' || i == len - 1) { ... }` body that always ends in an unconditional `break;` on that same outer-loop iteration, regardless of which branch was taken. So the new value of `i` is never read again afterward — confirmed with `grep -n '\bi\b'` across the rest of the function, including the post-loop validation `switch`. That was true on `develop` before this PR existed too, so the removal isn't tied to the bug fix and doesn't change behavior for any input I can construct.

jhendersonHDF: If anything, that just makes me suspicious of the unconditional break considering that it clearly seems like the `i += m` statement was intended to advance past parsed parameters in the filter string. Seems that break statement changed the previous behavior of the function at the time it was added.

brtnfld:
While digging into the `i += m` discussion above, I built h5repack from this branch and ran the actual regression cases #5480 added (`plugin_zero`, `plugin_test_less`, `plugin_test_ex`, `plugin_zero_extra`) to confirm this PR doesn't regress that fix. All four still pass with the expected exit codes, and the colon-prefix behavior (`-f :NONE` vs `-f NONE`) is still identical.

However, while checking *why* `plugin_test_less` (`UD=257,0,1`) fails, I found it's not for the reason the test name implies. `parse_filter` itself accepts that spec — the failure actually happens later, inside the real filter callback (`H5Z_pipeline(): filter returned failure`), because filter 257 does its own internal cd_values count check. I confirmed this by running the identical malformed spec against filter 250 instead (which has no such internal check): `UD=250,0,1` exits 0, silently treating the declared `nelmts=1` as `0` and discarding it.

Root cause: when `cd_nelmts` is the *last* token in a UD spec with no values following it, it's only ever committed inside the comma-triggered branch of the inner loop, which requires a comma *after* it. Since there isn't one, `filt->cd_nelmts` stays at its `memset` default of `0`, and the tail check (`if (cd_nelmts == 0) j = 0;`) can't distinguish "user wrote 0" from "user's declared count was never recorded." So `h5repack`'s own "incorrect number of compression parameters" validation never fires for this case — it's only caught in CI because filter 257 happens to validate itself.

This logic is unchanged from what #5480 introduced — this PR doesn't touch it (only the bounds-checking macro and the `goto` replacing the inner loop's `i = len - 1`/missing break touch nearby code). So it's not a regression from this PR, but flagging it since it's adjacent to what we've been discussing and seems worth a follow-up.

brtnfld:
Filed the UD `cd_nelmts` gap mentioned above as a separate issue, since it's pre-existing and not a regression from this PR: #6462

jhendersonHDF (approved):

tools/src/h5repack/h5repack_parse.c
brtnfld: Went and actually simulated the pre-2017 code (the parent of d627cd29bd6e, "HDFFV-10046 Fix windows parse error"), which is where `git blame` traces the `break;` to.

You're right that `i += m` mattered before that commit, but not in the way that is obvious, and the `break` doesn't actually change observable behavior for any input I could construct:

- For SZIP/SOFF, the old code set `i = len - 1` directly inside the inner loop when the mask completed. `i += m` only ever adds to that afterward, so `i` lands at `len - 1` or past it either way — natural termination, with or without the explicit `break`.
- For the single-digit "all other filters" branch, the inner loop has no comma-skip logic, so `m` always equals exactly `len - i - 1`, and `i += m` lands at `len - 1` — same result.
- For `UD` with 2+ commas, `m` does undershoot `len - i - 1` (each comma makes the inner loop's `u` advance one step ahead of `m`), so `i += m` can leave `i` short of `len - 1`, and the loop would run another outer iteration without the `break`. I confirmed this with a literal trace of `UD=307,1,9`: without the `break`, the loop re-enters once more near the end of the string.

But that extra iteration is harmless: the first `'='` trigger already wrote `scomp[k] = '\0'` at the filter-name boundary, so `scomp` as a C string is permanently "UD" from that point on — later iterations only write garbage after that embedded null, which `strcmp` never sees. So the spurious re-trigger just redundantly re-runs the same `UD` validation against the same `filt->cd_nelmts`/`j`, which haven't changed, and passes again. No difference in `filt->filtn`, `cd_nelmts`, `cd_values`, or `j` with or without the `break`.

So: the `break` was a real, intentional-looking behavior change in the sense that it removes that extra (always-harmless) iteration, not just dead-code cleanup. But I can't find an input where it changes the final result.

PR #6438, opened 2026-06-10
Fix issues with MSVC and OneAPI CI due to VS 18 2026

jhendersonHDF:
Visual Studio 17 2022 seems to no longer be available from actions run on windows-latest from forks

Ref: https://github.com/actions/runner-images/issues/14017

### DIFF
--- a/.github/workflows/call-workflows.yml
+++ b/.github/workflows/call-workflows.yml
@@ -173,6 +173,9 @@ jobs:
             thread_safety: ""
             build_mode: "Release"
             force_java_implementation: "jni"
+            java_version: 11
+            win_runner: "windows-2022"
+            win_vs_generator: "Visual Studio 17 2022"
 
     call-jni-latest-java:
         name: "JNI Latest Java Testing"
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -71,7 +71,7 @@ jobs:
 
   Intel_oneapi_windows:
     name: "windows-oneapi ${{ inputs.build_mode }}"
-    runs-on: windows-latest
+    runs-on: windows-2025
     steps:
       - name: Get Sources (Windows)
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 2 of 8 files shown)


.github/workflows/intel.yml
```diff
@@ -71,7 +71,7 @@ jobs:
 
   Intel_oneapi_windows:
     name: "windows-oneapi ${{ inputs.build_mode }}"
-    runs-on: windows-latest
```
jhendersonHDF, line 74: The fortran-lang/setup-fortran action that's currently being used to setup OneAPI hasn't been updated yet for the latest windows, so this can only run on windows-2025 and not windows-latest (depending on where windows-latest gets mapped to, either windows-2025 or windows-2025-vs2026).

glennsong09 (dismissed):

.github/workflows/call-workflows.yml
```diff
@@ -173,6 +173,9 @@ jobs:
             thread_safety: ""
             build_mode: "Release"
             force_java_implementation: "jni"
+            java_version: 11
+            win_runner: "windows-2022"
+            win_vs_generator: "Visual Studio 17 2022"
```
jhendersonHDF, line 178: CMake 3.26 doesn't support VS 18 2026, so this action has to use windows-2022 with VS 17 2022

.github/workflows/intel.yml
jhendersonHDF: Correction, `windows-2022` is needed currently until the OneAPI setup can be updated

.github/workflows/intel.yml
```diff
@@ -86,6 +86,12 @@ jobs:
           compiler: intel
           version: '2025.2'
 
+      - name: Setup Java
+        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
+        with:
+          java-version: 11
```
jhendersonHDF, line 92: `windows-2022` uses Java 8 by default whereas `windows-latest` uses 17. Since Java 11 is the minimum required, that has to be specified for `windows-2022` runs

brtnfld (approved):

PR #6445, opened 2026-06-11
CI: fix template injection in workflow run: blocks

brtnfld:
## Summary

- `${{ inputs.* }}` and `${{ steps.*.outputs.* }}` expressions used directly inside `run:` shell blocks are a template-injection vector: the expression is expanded as literal text into the shell script before the shell parses it, so values containing shell metacharacters could alter the command
- Move all such expressions to `env:` blocks on the same step and reference them as plain shell variables (`$VAR`) or PowerShell variables (`$env:VAR`)
- Expressions in `env:` are expanded safely by the Actions runner before the shell sees them, eliminating the injection path
- Fixes zizmor `template-injection` findings across 6 files: `setup-jextract/action.yml`, `symlink-ctest-scripts/action.yml`, `abi-report.yml`, `ctest.yml`, `maven-deploy.yml`, `maven-staging.yml`


brtnfld:
Closing in favor of a targeted fix. The reusable workflow files fixed here receive their inputs from the calling workflow (call-workflows.yml), which is clean — so those fixes were unnecessary. A new PR will address only the genuine risks: workflow_dispatch inputs used directly in run: blocks in java-implementation-test.yml, maven-staging.yml, maven-build-test.yml, test-maven-packages.yml, publish-branch.yml, and test-binary-installation.yml.

Issue #6440, opened 2026-06-11
OSS-Fuzz: Proposal: move the OSS-Fuzz fuzzers and build scripts into this repository for upstream maintenance


arthurscchan:
At present, all of the OSS-Fuzz fuzzing for HDF5 lives solely in the OSS-Fuzz repository (https://github.com/google/oss-fuzz/tree/master/projects/hdf5). Because of this, new fuzzer development happens entirely outside this project, and the upstream maintainers have no straightforward way to review or monitor it.

I would like to propose moving all of the fuzzers, together with their build configuration, into a dedicated directory under test/ in this repository, so that the maintainers have the opportunity to review, maintain, and evolve the fuzzers alongside the rest of the codebase.

The suggested arrangement would be:

1. The fuzzers and their build script live upstream (here), and the OSS-Fuzz side simply checks them out and builds them. New fuzzers are added and existing ones are modified in this repository.
2. This keeps the fuzzers in step with the library: if an API or function signature changes, or a new target worth fuzzing appears, the harnesses can be updated in the same change, with quicker fixes and better long-term maintenance.
3. It is also designed so that no further changes are needed on the OSS-Fuzz side when fuzzers are added or changed upstream while OSS-Fuzz would discover and build whatever is present in the upstream fuzzing directory.

I would be happy to carry out this migration myself: moving the existing fuzzers and their build configuration into this repository and wiring them into the CMake build behind an opt-in option (e.g. HDF5_ENABLE_FUZZERS), so that it imposes no maintenance burden on you to set up. I already have a working prototype of this and would gladly open a pull request if this direction is agreeable.

I am also working on expanding the fuzzing to cover more of the HDF5 code, and I would like that work to land here once the existing fuzzers have been moved upstream.

Of course, this is only a suggestion. If the maintainers would prefer the fuzzers to remain in OSS-Fuzz, I am happy to keep the current setup and to ping the relevant maintainers for review whenever I add new fuzzers later in the process.

Thank you for considering this, and I would welcome any feedback on the preferred approach.

PR #6451, opened 2026-06-12
OSS-Fuzz: Move fuzzers upstream

arthurscchan:
This PR is proposed for #6440. Which add the OSS-Fuzz fuzzing build flags and copy the OSS-Fuzz fuzzers for hdf5 project to this upstream repository. After this fixes has been accepted and merged in, the OSS-Fuzz build will also be changed (https://github.com/google/oss-fuzz/pull/15718) to build from the new approach.


jhendersonHDF:
Thanks for the PR @arthurscchan; I believe these changes make sense to do, but we're going to discuss this internally as well.

PR #6456, opened 2026-06-15
Fix some undefined behavior from casting in dt_arith.c

jhendersonHDF:
Fixes some more undefined behavior in dt_arith.c due to casting floating-point type values into integer types that can't represent the values. Emulates what the library is doing internally in conversion routines. There is still some remaining undefined behavior in the library itself (exposed by dt_arith.c) from not checking for NaN values before casting them to integer types.

### DIFF
--- a/test/dt_arith.c
+++ b/test/dt_arith.c
@@ -5972,7 +5972,12 @@ test_conv_int_fp_conv_to_schar(void *hw_p, unsigned char *src_buf, size_t idx, d
 #ifdef H5_HAVE__FLOAT16
             H5__Float16 f16;
             memcpy(&f16, src_buf + idx * sizeof(H5__Float16), sizeof(H5__Float16));
-            aligned = (signed char)f16;
+            if (f16 > (H5__Float16)(SCHAR_MAX))
+                aligned = SCHAR_MAX;
+            else if (f16 < (H5__Float16)(SCHAR_MIN))
+                aligned = SCHAR_MIN;
+            else
+                aligned = (signed char)f16;
             break;
 #else
             H5_FAILED();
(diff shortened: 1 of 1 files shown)


test/dt_arith.c
```diff
@@ -6547,6 +6684,7 @@ test_conv_int_fp_conv_to_uint(void *hw_p, unsigned char *src_buf, size_t idx, dt
 #ifdef H5_HAVE__FLOAT16
             H5__Float16 f16;
             memcpy(&f16, src_buf + idx * sizeof(H5__Float16), sizeof(H5__Float16));
+            /* No overflow/underflow checking needing here */
```
brtnfld: "needed"

test/dt_arith.c
```diff
@@ -6904,50 +7117,74 @@ test_conv_int_fp_conv_to_llong(void *hw_p, unsigned char *src_buf, size_t idx, d
         case FLT_FLOAT: {
             float f;
             memcpy(&f, src_buf + idx * sizeof(float), sizeof(float));
-            aligned = (long long)f;
+            if (f > (float)(LLONG_MAX))
```
brtnfld, line 7120: Should this be >=

test/dt_arith.c
```diff
@@ -6087,7 +6116,12 @@ test_conv_int_fp_conv_to_uchar(void *hw_p, unsigned char *src_buf, size_t idx, d
 #ifdef H5_HAVE__FLOAT16
             H5__Float16 f16;
             memcpy(&f16, src_buf + idx * sizeof(H5__Float16), sizeof(H5__Float16));
-            aligned = (unsigned char)f16;
+            if (f16 > (H5__Float16)(UCHAR_MAX))
+                aligned = UCHAR_MAX;
+            else if (f16 < (H5__Float16)0)
```
brtnfld, line 6121: Don't you also need this for to_ushort/to_uint/to_ulong/to_ullong

jhendersonHDF: In this case no, because the conversion logic in the library doesn't perform any checks since we're guaranteed that an f16 will fit into ushort and larger

test/dt_arith.c
jhendersonHDF: There are possibly some comparison issues here with certain types due to rounding in specific compilers, but this is currently just matching the logic in the library so the testing passes

brtnfld (approved):

jhendersonHDF:
Merging as this fixed a few issues but there are still problems remaining with NVHPC. However, it seems to be some weird interaction between nvc and cmake. dt_arith passes testing with a Developer build and fails with all other build types. But if a build tree is first built as a Developer build and then switched over to any other build type, suddenly dt_arith passes with all build types.

Further, dt_arith fails for Debug builds even though the only difference between Developer and Debug builds in terms of build flags is `-DH5_NO_FREE_LISTS -DH5_STRICT_FORMAT_CHECKS` and specifying either or both of these for a Debug build still fails.

Finally, changing the `ALIGNMENT` macro in dt_arith.c from 1 to 0 causes dt_arith to pass for all build types.

Serial vs. parallel compilation does not appear to make a difference.

PR #6460, opened 2026-06-16
Exclude www.doxygen.org/index.html from checking in linkchecker workflow

lrknox:
Doxygen adds a footer link to doxygen.org on every generated HTML page, causing lychee to check the same URL 31 times. Fix by excluding www.doxygen.org from the HTML glob scan and adding the URL to a small text file that lychee reads as a separate input, checking it once.

However, www.doxygen.org apparently blocks requests run directly from github actions, so we exclude it from checking.

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -79,16 +79,6 @@ jobs:
           # Let the action fail naturally - we'll handle it in the summary step
           fail: true
 
-      - name: Setup to check Doxygen footer URL (once)
-        run: printf 'https://www.doxygen.org/index.html\n' > /tmp/doxygen-url.txt
-
-      - name: Check Doxygen footer URL (once)
-        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411
-        with:
-            args: >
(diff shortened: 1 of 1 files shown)


brtnfld:
The central problem with this approach is that --exclude "www.doxygen.org" is global, the URL is never checked at all 
```


Split into two workflow steps — one with the exclusion for the HTML scan, one without for the single URL check:

- name: Run Comprehensive Link Check
  uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411
  with:
    args: |
      --exclude "www.doxygen.org"
      "${{ steps.build_docs.outputs.html_path }}/**/*.html"
    output: lychee-report.md
    format: markdown

- name: Check Doxygen footer URL (once)
  uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411
  with:
    args: "https://www.doxygen.org/index.html"
```
This guarantees the URL is checked exactly once and --exclude cannot interfere.

brtnfld (changes requested):
see comment

brtnfld (approved):

PR #6466, opened 2026-06-19
h5repack: fix silent loss of a declared cd_nelmts for UD filters

### LINKED ISSUE #6462: h5repack: UD filter parser silently drops a declared cd_nelmts when no values follow it
## Description

\`parse_filter()\` in \`tools/src/h5repack/h5repack_parse.c\` silently coerces a user-declared, nonzero \`cd_nelmts\` to \`0\` when it's the last token in a \`-f UD=filtn,filt_flag,cd_nelmts[,val1,...]\` spec and no values follow it. \`h5repack\`'s own "incorrect number of compression parameters" validation never fires for this malformed input.

## Reproduction

\`\`\`
h5repack -f UD=250,0,1 in.h5 out.h5
\`\`\`

Filter 250 has no internal cd_values count check, so this exits 0 and silently repacks with \`cd_nelmts=0\` instead of rejecting the spec (the user wrote \`cd_nelmts=1\` but supplied no value).

Contrast with the existing CMake test \`plugin_test_less\` (\`UD=257,0,1\`, added in #5480), which does exit 1 — but for the wrong reason. \`parse_filter\` accepts it the same way; the failure actually happens later inside the real filter callback (\`H5Z_pipeline(): filter returned failure\`) because filter 257 happens to validate its own cd_values count internally. So the existing test passes by coincidence, not because \`h5repack\`'s parser caught the error.

## Root cause

In the \`UD\` branch's inner loop (around \`tools/src/h5repack/h5repack_parse.c\`, \`parse_filter()\`), \`filt->cd_nelmts\` is only ever assigned inside the comma-triggered branch, which requires a comma *after* the \`cd_nelmts\` token. When \`cd_nelmts\` is the last token in the string (no values follow), that branch never runs, so \`filt->cd_nelmts\` stays at its \`memset\` default of \`0\`. The post-loop tail check:

\`\`\`c
if ((strcmp(scomp, "UD") == 0) && (filt->cd_nelmts == 0))
    j = 0;
else
    filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
\`\`\`

can't distinguish "user explicitly wrote 0" from "the declared count was never recorded," so it always takes the \`j = 0\` branch and discards the trailing token instead of treating it as the (unfulfilled) \`cd_nelmts\` value.

## Scope note

This logic predates #6434 (the stack-buffer-overflow fix for #6433) and is unchanged by it — #6434 only added bounds-checking and replaced the inner loop's \`i = len - 1\`/missing \`break\` with a \`goto\`. Found and discussed during review of #6434: https://github.com/HDFGroup/hdf5/pull/6434#issuecomment-4733185017

## Suggested fix direction

Track whether \`cd_nelmts\` was explicitly committed (e.g. a separate flag, mirroring \`l\`/\`f\`/\`p\` used for the other UD fields) instead of relying on its value still being the zero-initialized default. Add a CMake test for \`UD=<filtn>,<flag>,1\` with no value against a filter that does *not* do its own internal validation, so the gap can't be masked again.

brtnfld:
## Summary

Fixes #6462.

A declared `cd_nelmts` with no values following it (e.g. `UD=<filtn>,<flag>,1`) was never committed to `filt->cd_nelmts` in `parse_filter()`: the trailing token (the one with no comma after it) was always treated as a *value*, never checked against the `l`/`f`/`p` pending flags the way the comma-triggered branch already does. `h5repack`'s own "incorrect number of compression parameters" validation then silently passed, since `filt->cd_nelmts` stayed at its `memset` default of `0` regardless of what the user actually wrote. The mismatch was only ever caught by coincidence, when the underlying filter plugin happened to do its own internal `cd_values` count check (see discussion on #6434).

- Commits the trailing token to whichever UD field (`filtn`/`filt_flag`/`cd_nelmts`/value) is still pending, mirroring the comma-triggered branch.
- Initializes `l`/`f`/`p` at declaration to clear a real `-Wmaybe-uninitialized` warning the change exposed (the compiler can't otherwise prove the UD branch that sets them was taken).
- Adds a regression test (`plugin_test_nelmts_no_values`, `UD=250,0,1`) using filter 250, which has no internal validation, so the parser's own check is what's actually being exercised — the existing `plugin_test_less` test (`UD=257,0,1`) passed for the wrong reason (filter 257's internal check, not `parse_filter()`'s), and still does, but now for the right reason too.

### DIFF
--- a/tools/src/h5repack/h5repack_parse.c
+++ b/tools/src/h5repack/h5repack_parse.c
@@ -58,7 +58,7 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
     size_t      i, m, u;
     char        c;
     size_t      len = strlen(str);
-    int         f, k, l, p, q, end_obj = -1, no_param = 0;
+    int         f = -1, k, l = -1, p = -1, q, end_obj = -1, no_param = 0;
     unsigned    j, n;
     char        sobj[MAX_NC_NAME];
     char        scomp[16] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
@@ -297,8 +297,19 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t 
                 } /*if */
 
 done_filter_params:;
-                if ((strcmp(scomp, "UD") == 0) && (filt->cd_nelmts == 0))
-                    j = 0;
+                if (strcmp(scomp, "UD") == 0) {
+                    /* the trailing token (no comma after it) was never committed inside
+                     * the loop above; commit it to whichever UD field is still pending,
+                     * mirroring the comma-triggered branch exactly */
+                    if (l == -1)
+                        filt->filtn = atoi(stype);
+                    else if (f == -1)
+                        filt->filt_flag = (unsigned)strtoul(stype, NULL, 0);
+                    else if (p == -1)
+                        filt->cd_nelmts = strtoull(stype, NULL, 0);
+                    else if (filt->cd_nelmts > 0)
+                        filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
+                }
                 else
                     filt->cd_values[j++] = (unsigned)strtoul(stype, NULL, 0);
             }
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -190,6 +190,15 @@ The `h5repack` tool now obtains its default low and high library version bounds 
 
 ## Tools
 
+### Fixed h5repack silently dropping a declared cd_nelmts for user-defined filters
+
+   `h5repack -f UD=<filtn>,<flag>,<cd_nelmts>` with no values following `cd_nelmts` silently
+   treated the declared count as 0 instead of validating it, because the trailing token (with
+   no comma after it) was never committed to `cd_nelmts` inside the parser. `parse_filter()` now
+   commits the trailing token to whichever UD field is still pending, so a declared, unfulfilled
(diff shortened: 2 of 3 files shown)


tools/src/h5repack/h5repack_parse.c
```diff
@@ -297,8 +297,19 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t
                 } /*if */
 
 done_filter_params:;
-                if ((strcmp(scomp, "UD") == 0) && (filt->cd_nelmts == 0))
-                    j = 0;
+                if (strcmp(scomp, "UD") == 0) {
```
jhendersonHDF, line 300: Since the UD filter processing code doesn't appear to have any `goto` statements that jump to `done_filter_params`, it seems like this is all logic that should just be in the `if (strcmp(scomp, "UD") == 0)` block further up. Then, there wouldn't be a "trailing" token to deal with here.

tools/src/h5repack/h5repack_parse.c
```diff
@@ -297,8 +297,19 @@ parse_filter(const char *str, unsigned *n_objs, filter_info_t *filt, pack_opt_t
...
+                    /* the trailing token (no comma after it) was never committed inside
+                     * the loop above; commit it to whichever UD field is still pending,
+                     * mirroring the comma-triggered branch exactly */
+                    if (l == -1)
+                        filt->filtn = atoi(stype);
```
jhendersonHDF, line 305: These calls should really have error checking and then overflow checking before casting to unsigned, but this parsing code is already fairly messy

jhendersonHDF (approved):
Fine as a bugfix but I think this parsing code needs some refactoring

lrknox (approved):

Issue #6365, opened 2026-04-13
Corrections to the chunked storage property description in the layout message


vchoi-hdfgroup:
According to the library coding:
For version 1 and 2 of the layout message:
--there is no "Dataset Element Size" field for chunked storage
For version 3 of the layout message:
--there is no "Dataset Element Size" field in the chunked storage property description

For version 3 and 4 of the layout message:
--in the chunked storage property description table: correct the field "Dimension 0...#n Size" to "Dimension 1 ... #n Size"

PR #6469, opened 2026-06-19
Fix for issue #6365

vchoi-hdfgroup:
Clarify the description for the chunked dataset's dimensions.

### DIFF
--- a/docs/doxygen/dox/H5.format.4.0.dox
+++ b/docs/doxygen/dox/H5.format.4.0.dox
@@ -7654,7 +7654,7 @@ another datatype) and the dataspace for a dataset describes the location of the 
   <td>An array has a fixed dimensionality. This field specifies the number of dimension size fields later
       in the message. 
       <br />For chunked storage, the value stored is one greater than the dataset&rsquo;s actual number of
-      dimemsions. For example, a 1-dimensional dataset stores a value of 2.
+      dimensions. For example, a 1-dimensional dataset stores a value of 2.
       This increment accounts for the dataset element size which occupies the final dimension entry.</td>
 </tr>
 <tr valign=top>
--- a/docs/doxygen/dox/H5.format.4.0.dox
+++ b/docs/doxygen/dox/H5.format.4.0.dox
@@ -8134,7 +8134,7 @@ information for the virtual layout class as well as indexing information for the
 </tr>
 <tr>
   <td>Dataset Element Size</td>
-  <td>The size of a dataset elemnt, in bytes.  
+  <td>The size of a dataset element, in bytes.  
       This field is only present for chunked data storage
       and is stored as the last dimension entry in the array of dimensions.
       See previous field description.</td>
--- a/docs/doxygen/dox/H5.format.4.0.dox
+++ b/docs/doxygen/dox/H5.format.4.0.dox
@@ -7584,29 +7584,32 @@ another datatype) and the dataspace for a dataset describes the location of the 
   <th width="25%">byte</th>
   <th>byte</th>
 </tr>
-<tr align="center">
+<tr>
   <td>Version</td>
   <td>Dimensionality</td>
   <td>Layout Class</td>
   <td>Reserved <em>(zero)</em></td>
 </tr>
-<tr align="center">
+<tr>
   <td colspan="4">Reserved <em>(zero)</em></td>
 </tr>
-<tr align="center">
+<tr>
   <td colspan="4"><br />Data Address<sup>O</sup> <em>(optional)</em><br /><br /></td>
 </tr>
-<tr align="center">
+<tr>
+  <td colspan="4">Dimension 0 Size</td>
+</tr>
+<tr>
   <td colspan="4">Dimension 1 Size</td>
 </tr>
-<tr align="center">
-  <td colspan="4">Dimension 2 Size</td>
-</tr>
-<tr align="center">
+<tr>
   <td colspan="4">...</td>
 </tr>
-<tr align="center">
-  <td colspan="4">Dimension \#n Size</td>
+<tr>
+  <td colspan="4">Dimension \#(n - 1) Size</td>
+</tr>
+<tr>
+  <td colspan="4">Dataset Element Size</td>
 </tr>
 <tr>
   <td colspan="4">Compact Data Size <em>(optional)</em></td>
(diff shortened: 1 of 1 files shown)


docs/doxygen/dox/H5.format.4.0.dox
```diff
@@ -7597,16 +7597,16 @@ another datatype) and the dataspace for a dataset describes the location of the
   <td colspan="4"><br />Data Address<sup>O</sup> <em>(optional)</em><br /><br /></td>
 </tr>
 <tr align="center">
-  <td colspan="4">Dimension 0 Size</td>
+  <td colspan="4">Dimension 1 Size</td>
```
jhendersonHDF: Shouldn't these stay 0-based and instead "Dimension \#n" below should be changed to n - 1? Otherwise, this seems to conflict with the text below that uses "(n - 1)" when referring to dimensions stored.

gheber: I think there are several problems here:

1. `n` is not defined anywhere in the context. The `Dimension #n Size` field name is just sloppy.
2. Whatever we settle on, (0- or 1-based), it must be consistent throughout the document and explicitly stated near the beginning of the document.
3. I find the term 'Dimension Size' problematic. Throughout the document, we refer to 'Size of Offsets', etc., but what's meant here is **not** the Dimension Size, but the *Dimension Extent*.
4. The parsing code for the layout message, lines 1501-1526 in [messages.pk](https://github.com/HDFGroup/hdf5-pickles/blob/main/pickles/messages.pk) shows another logical problem: The description of the `Dimensionality` field states, "This field specifies the number of dimension size fields later in the message. The value stored for chunked storage is 1 greater than the number of dimensions in the dataset’s dataspace." This begs the obvious question: what does that mysterious additional element represent? Well, it's actually the `Dataset Element Size`, which, according to the table, is not part of the array.

I think we need to set the logic straight, first, i.e., what's in the array, and what's not, and then we can deal with the 0- vs. 1-based indexing.

vchoi-hdfgroup: If you look at the field "dimensionality" description, "For chunked storage, the value stored is one greater than the dataset’s actual number of dimemsions. For example, a 1-dimensional dataset stores a value of 2. This increment accounts for the dataset element size which occupies the final dimension entry."

jhendersonHDF: But assuming indices are still 0-based here, what's stored is the sizes for dimension 0..(n-1) and the dataset element size just happens to be an extra element in that "array".

vchoi-hdfgroup: When you mentioned above conflicting with the text "(n-1)" which line are you referring to?

jhendersonHDF: In this diff it's line 7700. That text pretty much states what I'd expect: For an n-dimensional dataset (for n >= 2), dimensions 0..(n-1) are stored and then the dataset element size is stored. The dataset element size is part of the dimensions "array", but it's not really a dimension itself.

vchoi-hdfgroup: So actually I should have the following:
In the layout table:
"Dimension 0 Size, Dimension 1 size,.... Dimension (n-1) size, Dataset element size"

And in the description table:

For chunked storage, the dimensions define the size of a single chunk. They are in units of array elements (not bytes). The first dimension stored in the list of dimensions is the slowest changing dimension and the (n - 1) dimension stored is the fastest changing dimension. 
For an n-dimensional dataset (for n >= 2), dimensions 0..(n-1) are stored and then the dataset element size is stored. The dataset element size is part of the dimensions "array", but it's not really a dimension itself.

jhendersonHDF: > "Dimension 0 Size, Dimension 1 size,.... Dimension (n-1) size, Dataset element size"

Yes, this part is what I think the layout table should say.

> For an n-dimensional dataset (for n >= 2), dimensions 0..(n-1) are stored and then the dataset element size is stored. The dataset element size is part of the dimensions "array", but it's not really a dimension itself.

These were just my own words summarizing the encoding; I think the text currently in the description table in this diff is fine as is and you don't need to add these words.

gheber: I like where this is going, but it must be consistent across versions of a primitive. There is a lot of repetition in "IV.A.2.i. The Data Layout Message". The price we have to pay is that updating `Dimension #n Size`, `Dataset Element Size` needs to be updated in 3 (4?) places to remain consistent. Do we want to cut out some of that redundancy?

vchoi-hdfgroup: I have made changes accordingly.

jhendersonHDF (dismissed):

gheber (dismissed):
Looking much better.

docs/doxygen/dox/H5.format.4.0.dox
vchoi-hdfgroup: > I like where this is going, but it must be consistent across versions of a primitive. There is a lot of repetition in "IV.A.2.i. The Data Layout Message". The price we have to pay is that updating `Dimension #n Size`, `Dataset Element Size` needs to be updated in 3 (4?) places to remain consistent. Do we want to cut out some of that redundancy?

I replaced those duplicate field descriptions with cross-reference to the version 3 layout message.

vchoi-hdfgroup: > I think there are several problems here:
> 
>     1. `n` is not defined anywhere in the context. The `Dimension #n Size` field name is just sloppy.
> 
>     2. Whatever we settle on, (0- or 1-based), it must be consistent throughout the document and explicitly stated near the beginning of the document.
> 
>     3. I find the term 'Dimension Size' problematic. Throughout the document, we refer to 'Size of Offsets', etc., but what's meant here is **not** the Dimension Size, but the _Dimension Extent_.
> 
>     4. The parsing code for the layout message, lines 1501-1526 in [messages.pk](https://github.com/HDFGroup/hdf5-pickles/blob/main/pickles/messages.pk) shows another logical problem: The description of the `Dimensionality` field states, "This field specifies the number of dimension size fields later in the message. The value stored for chunked storage is 1 greater than the number of dimensions in the dataset’s dataspace." This begs the obvious question: what does that mysterious additional element represent? Well, it's actually the `Dataset Element Size`, which, according to the table, is not part of the array.
> 
> 
> I think we need to set the logic straight, first, i.e., what's in the array, and what's not, and then we can deal with the 0- vs. 1-based indexing.

I will file an issue to address problems #1, #2 and #3 as they also occur in other parts of the document.  Need to investigate the whole spec to identify the areas to be corrected.

gheber (dismissed):
Looking good.

docs/doxygen/dox/H5.format.4.0.dox
vchoi-hdfgroup: I have pushed more changes to this PR, hopefully will give a better description.

gheber (approved):
Still looking good.

PR #6470, opened 2026-06-19
✨ RFC-HDFG-2026-001: String-based filter configuration API

### LINKED ISSUE #6153: ✨ [Feature Request] String-Based Configuration Interface for Filters
### 1. Problem Description
Currently, the generic interface for configuring HDF5 filters relies exclusively on `cd_values`, an array of `unsigned int`. While this works for simple integer parameters, it creates significant friction for modern filters (like **H5Z-ZFP**) that require floating-point parameters (rates, precision, thresholds).

**Pain Points:**
*   **Type Punning:** Users must manually bit-cast `double` or `float` values into integer arrays. This is error-prone, endian-sensitive, and difficult to implement in high-level languages (Python, Java, Matlab) without dedicated wrappers.
*   **CLI Usability:** Tools like `h5repack` require users to pass "magic numbers" (e.g., `-f UD=32013,0,4,1,0,0,1074921472`) instead of human-readable settings.
*   **Opaque Configuration:** It is difficult to introspect a file's creation properties and understand what compression settings were used (e.g., `cd_values[2]` is meaningless compared to `"rate=4.0"`).

### 2. Proposed Solution
Introduce a mechanism to configure filters using a string-based key-value interface, mirroring modern capabilities recently added for Virtual File Drivers (VFDs) and VOL connectors.

The underlying storage format (the `cd_values` array stored in the object header) **should remain unchanged**. This feature would simply be a usability layer that allows the filter plugin to parse a string and populate the `cd_values` array itself.

#### C API Comparison
**Current (Manual Type Punning):**
```c
// User must know internal layout and bit-casting
unsigned int cd_vals[4];
double rate = 3.5;
cd_vals[0] = H5Z_ZFP_MODE_RATE;
memcpy(&cd_vals[2], &rate, sizeof(double)); // Complex & unsafe
H5Pset_filter(plist, H5Z_FILTER_ZFP, flags, 4, cd_vals);
```

**Proposed:**
```c
// User specifies intent; Plugin handles the bits
H5Pset_filter_config(plist, H5Z_FILTER_ZFP, "mode=rate; rate=3.5");
```

#### CLI Impact (`h5repack`)
This would allow `h5repack` to support a generic string flag, removing the need for external helper scripts like `print_h5repack_farg`.

**Current:**
```bash
./print_h5repack_farg rate=3.5
# Output: -f UD=32013,0,4,1,0,0,1074528256
h5repack -f UD=32013,0,4,1,0,0,1074528256 in.h5 out.h5
```

**Proposed:**
```bash
h5repack -f ZFP="mode=rate; rate=3.5" in.h5 out.h5
```

### 3. References
*   Related Issue: [llnl/H5Z-ZFP#159](https://github.com/llnl/H5Z-ZFP/issues/159) - Discusses the complexity of generic usage.
*   Precedent: HDF5 1.14 VFD configuration strings (`H5Pset_driver_by_name` configuration argument).

brtnfld:
## Summary
- Adds a human-readable key=value parameter string API for HDF5 filters, alongside the existing integer `cd_values` arrays: `H5Pappend_filter`, `H5Pmodify_filter_by_idx`, `H5Pget_filter_params_by_idx`, `H5Zconfig_get_int/double/bool/str`, `H5Z_filter_id_by_name`, `H5Zget_filter_class_info`.
- Extends `H5Z_class3_t` with `name` (canonical, validated at registration) and `description`, plus `set_config`/`get_config` callbacks, and threads `dxpl_id`/`scaled[]`/`ndims` through `H5Z_pipeline` so v3 filter callbacks have full context. All six built-in filters (deflate, shuffle, fletcher32, nbit, szip, scaleoffset) implement `set_config`/`get_config`.
- New on-disk pipeline v3 format: adds an `H5F_LIBVER_V300` file-format bound (the new `LATEST`) under which the canonicalized parameter string is persisted verbatim alongside each filter's `cd_values` in the pipeline message, so a file/tool without the original plugin can still show a human-readable configuration. Older bounds are unaffected; read compatibility with pre-3.0 files/libraries is preserved.
- Vendors a TOML subset parser (tomlc17, MIT) in `src/tomlc17/`, built unconditionally into libhdf5 with hidden visibility to avoid symbol collisions; updated to tag R260821 to pick up upstream's own subnormal-handling fix.
- Adds Fortran, C++, and Java (JNI + FFM) bindings, plus `h5dump`/`h5repack` support for the new parameter strings, including a `UD=id,flags,key=value` h5repack CLI form and hex-float annotation for binary-exact floats in h5dump output.

**Resolves the review feedback from the earlier iteration of this PR:** @mattjala noted that the `filter_title`/`cd_values` mechanism described in RFC §4.3/5.2/10.1 didn't match the implementation. RFC-HDFG-2026-001 has since been amended to describe the mechanism actually implemented: `H5Z_class3_t.name` is a canonical, uniqueness-checked identifier validated at `H5Zregister()` time and persisted on disk via the pipeline v3 message (satisfying the "survives without the plugin" goal from a different angle than originally spec'd); `description` remains registry-only and is documented as best-effort. The reserved in-file blob-callback placeholders that had no implementation have been removed from `H5Z_class3_t`.

This branch also went through a full comprehensive-review pass: 3 Critical, 10 High, 13 Medium, and 7 Low findings were identified and all but two have been fixed and verified (including a self-inflicted file-corruption path from an unbounded canonicalized parameter string, and a stack buffer over-read in h5dump). The two exceptions are called out as deliberate, tracked follow-ups rather than being folded in here:
- `H5P_ENCODE_VERS` was not bumped — it's shared by every encoded property list, not just the filter pipeline, and deserves its own focused review.
- The 10 new Java JNI test methods are not yet ported to the FFM test suite (no JDK available in this environment to verify a port).

Fixes #6153.


release_docs/CHANGELOG.md
```diff
@@ -97,6 +97,42 @@ We would like to thank the many HDF5 community members who contributed to this r
...
 ## Library
 
+### Added string-based filter configuration API (RFC-HDFG-2026-001)
+
+   A new API allows filters to be configured using human-readable `key=value`
+   parameter strings (TOML-subset syntax) in addition to the existing integer
+   `cd_values` arrays.
+
+   **New C API functions:**
+   - `H5Pappend_filter(plist, filter_id, flags, params)` — appends a filter to
```
hyoklee: Replace — with -.

hyoklee (changes requested):

hyoklee (dismissed):

src/H5Z.c
```diff
@@ -1793,6 +2002,67 @@ H5Zget_filter_info(H5Z_filter_t filter, unsigned *filter_config_flags /*out*/)
...
+ *           String fields in `info` point into library-owned storage; they
+ *           are valid until the filter is unregistered.
+ *
+ * Return:   Non-negative on success / Negative on failure
+ *
+ * Since:    2.2.0
+ *-------------------------------------------------------------------------
+ */
+herr_t
+H5Zget_filter_info2(H5Z_filter_t filter, H5Z_class_info_t *info /*out*/)
```
jhendersonHDF: By using 2 here, this would seem to imply a deprecation (and also versioning) of the symbol `H5Zget_filter_info`, rather than just adding a new API function.

tools/test/h5dump/CMakeTests.cmake
```diff
@@ -1222,22 +1222,20 @@ ADD_H5_TEST (texceedsubblock RESULT_CODE 1 H5ERRREF "exceed dataset dims" --enab
...
 # SZIP
 # For VOL connectors, test files are generated during testing, which requires the filter itself to be available.
+# The PARAMS_STRING line in the reference is the SZIP get_config form
+# (e.g. coding = "nn", pixels_per_block = 4); without SZIP the library
+# falls back to the raw cd_values form, so only run when SZIP is available.
 if (HDF5_ENABLE_SZIP_SUPPORT)
-  set (SZIP_NATIVE_ONLY "")
-else ()
-  set (SZIP_NATIVE_ONLY "NATIVE_ONLY")
+  ADD_H5_TEST (tszip RESULT_CODE 0 APPLY_FILTERS 2 --enable-error-stack -H -p -d szip TARGET_FILE tfilters.h5)
```
jhendersonHDF, line 1243: This and the other lines below should use two different test reference files to test both cases instead of being converted to only testing when support for the particular filter is available

src/H5Zdevelop.h
```diff
@@ -26,10 +26,29 @@
...
 /**
- * Current version of the H5Z_class_t struct
+ * Current version of the H5Z_class_t struct (used by H5Z_class2_t)
  */
 #define H5Z_CLASS_T_VERS (1)
 
+/**
+ * Version constant for the new H5Z_class3_t struct \since 2.2.0
+ */
+#define H5Z_CLASS3_T_VERS (2)
```
jhendersonHDF, line 39: I still think this split doesn't make sense. Presumably, there's never going to be a new version of `H5Z_class2_t` and `H5Z_class3_t` is just an evolution (v3 of `H5Z_class_t`, but really v2 of `H5Z_class2_t`) and should just share the version number with `H5Z_CLASS_T_VERS`, with `H5Z_CLASS_T_VERS` remaining at v1 by default (for `H5Z_class2_t`) while making sure that version checks don't reject v2 (for `H5Z_class3_t`). As a split, it also doesn't make much sense for `H5Z_CLASS3_T_VERS` to start at v2.

mattjala:
The RFC's `filter_title` mechanism doesn't seem to exist at all in the implementation. RFC 4.3 defines `filter_title` as a short label packed into `cd_values` as a reserved key at registration, which then gets persisted in the file and extracted by H5Pget_filter2, with the goal being a human-readable label that survives without the plugin (see 10.1).

The new H5Z_class3_t carries name and description as registry fields instead of in cd_values. While name gets persisted to the file via the pipeline message, description stays in the registry.
 
The RFC's persistence goal is still met, but through a different mechanism than 4.3 specifies. The plugin's H5Z_class3_t.name is written into the H5O_pline message at encode time (when id ≥ 256) and read back by H5Pget_filter2. It seems like the filter_title/cd_values design in mentioned in 4.3/5.2/10.1 was dropped in favor of this. If so, we should update the RFC to describe the mechanism that was actually implemented.

tools/lib/h5tools_dump.c
```diff
@@ -3553,32 +3583,134 @@ h5tools_dump_dcpl(FILE *stream, const h5tool_format_t *info, h5tools_context_t *
...
+                            }
+                            ebuf[out_i] = '\0';
+                            params_str  = ebuf;
+                        }
+                    }
+
+                    {
+                        H5Z_class_info_t finfo;
+                        if (H5Zget_filter_info2(filtn, &finfo) >= 0 && finfo.description != NULL)
+                            filter_descr = finfo.description;
```
mattjala, line 3955: The -p DESCRIPTION line in h5dump's output comes from H5Zget_filter_info2().description, a registry lookup that fails for any filter that isn't currently registered. Since that string lives only in the in-memory registry, from the user-perspective, DESCRIPTION silently disappears when the plugin isn't loaded.

That produces a confusing asymmetry in the FILTERS block. COMMENT (the filter name) has a file-persistence path via the pipeline message and always displays, while DESCRIPTION (registry prose) does not. `h5dump -p` on the same file on two machines with/without the plugin will get inconsistent DESCRIPTION output.  And DESCRIPTION isn't mentioned in RFC 8.2 at all.

I think DESCRIPTION should be documented in 8.2 and the DDL grammar as a registry-sourced, best-effort field that is omitted for unregistered filters, so that the difference from the persisted COMMENT is at least explained somewhere.

If the intention was for DESCRIPTION to serve as the persisted, plugin-independent label, then that is achieved by the name/COMMENT path, and it might make sense to drop DESCRIPTION to avoid implying it's stored in the file.

tools/src/h5repack/h5repack_filters.c
```diff
@@ -363,6 +363,21 @@ apply_filters(const char    *name,    /* object name from traverse list */
             if (obj->filter[i].filtn < 0)
                 H5TOOLS_GOTO_ERROR((-1), "invalid filter");
 
+            /* RFC-HDFG-2026-001 §9: TOML parameter string from
+             * UD=id,flags,key=value... — dispatch through H5Pappend_filter so
+             * the filter's set_config callback translates the string into
+             * cd_values.  Bypasses the per-filter H5Pset_* switch below. */
+            if (obj->filter[i].params_str[0] != '\0') {
```
mattjala, line 369: This is new behavior, and should have a new CLI-level test covering it to avoid regression issues. 

Specifically, there should be a new h5repack test in tools/test/h5repack/CMakeTests.cmake that repacks a dataset with the new form (e.g. -f 'UD=<id>,0,mode=rate,rate=3.5' against a registered v3 test plugin) and verifies via h5dump -p that the resulting pipeline/params round-trip without any issues. It would also be a good idea to include a case for the old form (e.g. UD=307,0,1,9) in the same test to guard against inconsistencies in the rewritten parser.

src/H5Zcd.c
jhendersonHDF, line 1: Unless there's a use case that motivates having this functionality, I believe all of the functions in this file are trivial for filter authors to implement and I'm not sure I see the value in them. It should be assumed that filter authors know what they're doing rather than trying to force conventions like little-endian ordering that may not always make sense.

src/H5Zdevelop.h
```diff
@@ -171,6 +190,112 @@ typedef struct H5Z_class2_t {
...
+    const char  *name;               /**< Canonical string identifier (e.g., "zfp"); must not be NULL; used as
+                                        display name */
+    const char *description;         /**< Human-readable description of the filter (e.g., "Deflate (zlib)
+                                        general-purpose compression"); may be NULL */
+    H5Z_can_apply_func_t  can_apply; /**< The "can apply" callback for a filter      */
+    H5Z_set_local_func_t  set_local; /**< The "set local" callback for a filter      */
+    H5Z_func2_t           filter;    /**< Extended filter callback: dxpl_id + scaled */
+    H5Z_set_config_func_t set_config; /**< String configuration callback; may be NULL */
+    H5Z_get_config_func_t get_config; /**< Parameter string reconstruction; may be NULL */
+    void                 *write_blob; /**< Reserved: set to NULL; defined by RFC-HDFG-2026-* */
```
jhendersonHDF: Before reserving these a signature should be decided if anything, because the type for these pointers / prototype for these functions will have to be changed anyway at the point that they become used, so reserving them this way is largely pointless as a major release will still be needed.

src/H5Zdevelop.h
```diff
@@ -183,6 +308,164 @@ typedef struct H5Z_class2_t {
...
+ * \return Positive if found, 0 if not found, negative on error.
+ *
+ * \details Accepts "true" or "false" (lowercase only, per TOML).
+ *          Bare keys (boolean flags with no '=' sign) are treated as TRUE.
+ *          Returns negative if the key exists but its value is not a TOML
+ *          boolean (type mismatch).
+ *
+ * \since 2.2.0
+ */
+H5_DLL htri_t H5Zconfig_get_bool(const char *params, const char *key, hbool_t *out);
```
jhendersonHDF: `hbool_t` shouldn't be used in public API functions anymore, just use `bool`.

src/H5Zprivate.h
```diff
@@ -71,19 +75,24 @@ struct H5O_pline_t; /*forward decl*/
...
 H5_DLL herr_t H5Z_register(const H5Z_class2_t *cls);
+H5_DLL herr_t H5Z_register3(const H5Z_class3_t *cls);
 H5_DLL herr_t H5Z_append(struct H5O_pline_t *pline, H5Z_filter_t filter, unsigned flags, size_t cd_nelmts,
                          const unsigned int cd_values[]);
 H5_DLL herr_t H5Z_modify(const struct H5O_pline_t *pline, H5Z_filter_t filter, unsigned flags,
                          size_t cd_nelmts, const unsigned int cd_values[]);
 H5_DLL herr_t H5Z_pipeline(const struct H5O_pline_t *pline, unsigned flags, unsigned *filter_mask /*in,out*/,
                            H5Z_EDC_t edc_read, H5Z_cb_t cb_struct, size_t *nbytes /*in,out*/,
-                           size_t *buf_size /*in,out*/, void **buf /*in,out*/);
+                           size_t *buf_size /*in,out*/, void **buf /*in,out*/, hid_t dxpl_id,
```
jhendersonHDF: Minor nitpick, but as all the new params are IN-type params, it would be nice to have them after `flags` so that all the IN-type parameters come first and then the OUT- or IN/OUT-type ones follow

src/H5Zprivate.h
```diff
@@ -71,19 +75,24 @@ struct H5O_pline_t; /*forward decl*/
...
+                           const hsize_t *scaled, size_t ndims);
 H5_DLL herr_t H5Z_find(bool attempt, H5Z_filter_t id, H5Z_class2_t **cls);
-H5_DLL herr_t H5Z_can_apply(hid_t dcpl_id, hid_t type_id);
-H5_DLL herr_t H5Z_set_local(hid_t dcpl_id, hid_t type_id);
-H5_DLL herr_t H5Z_can_apply_direct(const struct H5O_pline_t *pline);
-H5_DLL herr_t H5Z_set_local_direct(const struct H5O_pline_t *pline);
-H5_DLL htri_t H5Z_ignore_filters(hid_t dcpl_id, const H5S_t *space);
+/* Find filter entry; returns pointer into internal table (cast-compatible with H5Z_class2_t *) */
+struct H5Z_entry_t; /* forward decl; full def in H5Zpkg.h */
```
jhendersonHDF: Needing to forward-declare `H5Z_entry_t` here seems to point to a design issue. Is there something H5Zpkg.h-wide which needs access to `H5Z_entry_t` such that the typedef can't be here in H5Zprivate.h?

test/chunk_info.c
```diff
@@ -49,13 +49,9 @@
...
-static const char *FILENAME[] = {"tchunk_info_earliest",
-                                 "tchunk_info_v18",
-                                 "tchunk_info_v110",
-                                 "tchunk_info_v112",
-                                 "tchunk_info_v114",
-                                 "tchunk_info_v200",
-                                 NULL};
+static const char *FILENAME[] = {
+    "tchunk_info_earliest", "tchunk_info_v18",  "tchunk_info_v110", "tchunk_info_v112",
+    "tchunk_info_v114",     "tchunk_info_v200", "tchunk_info_v300", NULL};
```
jhendersonHDF, line 56: This is something that should get updated when a new library version bound is added

tools/src/h5dump/h5dump.h
```diff
@@ -283,9 +283,10 @@ typedef struct {
     int include_attrs;          /* Display attributes */
     int display_vds_first;      /* vds display to all by default */
     int vds_gap_size;           /* vds skip missing files default is none */
+    int display_filter_params;  /* emit PARAMS_STRING in FILTERS block */
```
jhendersonHDF: Is there any way to actually turn this off individually? It seems it just uses the same logic as `display_dcpl` which makes this a useless field.

tools/test/h5dump/expected/tscaleoffset.ddl
```diff
@@ -7,7 +7,11 @@ DATASET "scaleoffset" {
       SIZE XXXX (4.XXX:1 COMPRESSION)
    }
    FILTERS {
-      COMPRESSION SCALEOFFSET { MIN BITS 2 }
+      COMPRESSION SCALEOFFSET {
+         MIN BITS 2
+         PARAMS_STRING "scale_type = \"int\", scale_factor = 4"
```
jhendersonHDF: From a display perspective, it probably makes sense to leave escaping characters out of the output given here

test/dsets.c
```diff
@@ -2741,16 +2741,19 @@ test_get_filter_info(void)
...
 
-    if (H5Z_SZIP->encoder_present) {
-        if (((flags & H5Z_FILTER_CONFIG_ENCODE_ENABLED) == 0) ||
-            ((flags & H5Z_FILTER_CONFIG_DECODE_ENABLED) == 0))
+    /* DECODE must always be available; ENCODE presence determines the other flag */
+    if ((flags & H5Z_FILTER_CONFIG_DECODE_ENABLED) == 0)
+        TEST_ERROR;
+    if (flags & H5Z_FILTER_CONFIG_ENCODE_ENABLED) {
+        /* Both encode and decode should be reported when the encoder is present */
+        if ((flags & H5Z_FILTER_CONFIG_DECODE_ENABLED) == 0)
```
jhendersonHDF: These changes don't make sense to me and seem to miss the point that the flags returned by `H5Zget_filter_info` were being checked for the right values against the value of `H5Z_SZIP->encoder_present`. The first branch also uselessly checks `if ((flags & H5Z_FILTER_CONFIG_DECODE_ENABLED) == 0)` again after already being check above.

src/H5version.h
```diff
@@ -1683,6 +1683,8 @@
   #define H5Z_class_t H5Z_class2_t
 #elif H5Z_class_t_vers == 1
   #define H5Z_class_t H5Z_class1_t
+#elif H5Z_class_t_vers == 3
```
fortnern, line 1684: Is this line added directly here? H5version.h is an auto-generated file, if someone runs make_vers this change will be overridden. We might need to modify make_vers to allow H5Z_class_t to be versioned to 3 by default by using "v300" in H5vers.txt. See the note in H5vers.txt about a similar situation in 1.10.

src/H5Pocpl.c
```diff
@@ -1761,3 +1757,345 @@ H5Pget_filter_by_id1(hid_t plist_id, H5Z_filter_t id, unsigned int *flags /*out*
...
+    if (params == NULL || params->type == H5Z_PARAMS_CDVALUES) {
+        /* Raw cd_values path – behaves identically to H5Pset_filter */
+        if (params) {
+            cd_nelmts = params->u.raw.cd_nelmts;
+            cd_values = params->u.raw.cd_values;
+            if (cd_nelmts > 0 && cd_values == NULL)
+                HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, FAIL, "cd_values is NULL but cd_nelmts > 0");
+        }
+    }
+    else if (params->type == H5Z_PARAMS_BLOB) {
```
fortnern: Why even have this option if it's not supported?

src/H5version.h
brtnfld: Extended bin/make_vers so a version tag outside the real @versions list (e.g., v300) can register a new symbol version without becoming the default selected version. Added v300 to the H5Z_class entry in H5vers.txt with an explanatory note.

tools/src/h5repack/h5repack_filters.c
brtnfld: Added a new v3 test plugin (tools/test/h5repack/dynlib_rpk3.c, filter ID 261) with set_config/get_config implementing mode=rate,rate=<N>, wired it into CMake, and added two h5repack tests (plugin_test_cfg for the new string form, plugin_test_cfg_legacy for the old raw form as a regression guard)

tools/lib/h5tools_dump.c
brtnfld: documented rather than dropped: added a \verysubsection{DESCRIPTION} to the RFC's h5dump section explaining it's best-effort/registry-sourced with no on-disk fallback (unlike COMMENT), added the corresponding grammar tokens to docs/doxygen/dox/DDLBNF200.dox, and added a source comment at the emission site in tools/lib/h5tools_dump.c.

src/H5Zdevelop.h
brtnfld: Agreed, removed. write_blob/read_blob/close_blob and the corresponding H5Z_PARAMS_BLOB/has_blob_callbacks fields are gone from H5Z_class3_t, H5Z_params_type_t, and H5Z_class_info_t (and from all the built-in filters, the test plugin, and the Java bindings that mirrored them). In-file blob configuration data can come back as its own RFC with a real struct/version bump once the callback signatures are actually designed.

src/H5Pocpl.c
brtnfld: Good point, removed. The H5Z_PARAMS_BLOB case (and the reserved write_blob/read_blob/close_blob fields it was for) is gone entirely rather than kept as a stub that always returns H5E_UNSUPPORTED.

src/H5vers.txt
```diff
@@ -95,5 +95,14 @@ FUNCTION: H5Tdecode; ; v18, v200
...
-TYPEDEF: H5Z_class; v16, v18
+# Note:  v300 is allowed in make_vers for a similar reason to the v111
+#        exception above, but for the opposite purpose.  H5Z_class3_t must
+#        NOT become the version selected by H5Z_class_t by default --
+#        existing code using H5Z_class_t would otherwise silently get the
+#        larger struct with uninitialized callback fields.  Tagging the
+#        version as "v300" (rather than a real, supported API version)
+#        registers H5Z_class3_t as symbol version 3 without it ever being
+#        reachable through @versions, so it is only selected when an
+#        application explicitly defines H5Z_class_t_vers to 3.
```
fortnern: Maybe a note about how it will become the default when 3.0.0 is released

brtnfld: Added a note above the TYPEDEF entry: when HDF5 3.0.0 is released, v300 should be added to @versions so H5Z_class3_t becomes the default, and the note removed.

src/H5Ppublic.h
```diff
@@ -850,7 +850,7 @@ H5_DLL hid_t H5Pcreate(hid_t cls_id);
  * \details H5Pcreate_class() registers a new property list class with the
  *          library. The new property list class can inherit from an
  *          existing property list class, \p parent, or may be derived
- *          from the default “empty” class, NULL. New classes with
+ *          from the default "empty" class, NULL. New classes with
```
fortnern: Are we changing the quotation marks from left quote (147) and right quote (148) to double quote (34)?

src/H5private.h
```diff
@@ -1445,6 +1453,26 @@ extern char H5_lib_vers_info_g[];
...
+ * Use this macro for public API functions that may be re-entered from inside
+ * a library callback that already holds the API lock (e.g. filter v3
+ * H5Z_set_config_func_t callbacks invoked from H5Pappend_filter).  Performs
+ * the same error-handling and function-name setup as FUNC_ENTER_API_NOINIT
+ * but skips H5_API_LOCK, so the function is safe to call while another
+ * public API frame already holds the lock.  Only appropriate for pure
+ * helpers that touch no global library state.  Examples: H5Zconfig_get_int,
+ * H5Zconfig_get_double, etc.
+ */
+#define FUNC_ENTER_API_NOINIT_NOLOCK                                                                         \
```
fortnern, line 1466: Why is this necessary? There is already machinery to allow re--entry within a single thread

bin/format_source
```diff
@@ -20,7 +20,7 @@ echo "parameter is optional and can be used to force a specific"
 echo "installed version of clang-format to be used."
 echo ""
 
-find . \( -type d -path ./config -prune -and -not -path ./config \) \
+find . \( -type d \( -path ./config -o -path ./src/tomlc17 \) -prune \) \
```
hyoklee, line 23: Is `/src/tomlc17/` necessary?

hyoklee (changes requested):

src/H5Ppublic.h
brtnfld: That was unintentional scope creep — reverted. The non-ASCII-punctuation cleanup should have stayed scoped to this PR's own new content; it accidentally swept through pre-existing, unrelated docstrings (H5Pcreate_class, H5Pget_driver, H5Pset_file_image, H5Pget_virtual_dsetname/filename/srcspace, H5Pset_scaleoffset, H5Pset_szip, H5Pset_local_heap_size_hint, H5Pget_mcdt_search_cb). Restored the original curly quotes/em-dash/≤ in all of those.

src/H5private.h
brtnfld: Fair point. The API mutex is initialized recursive (H5TS_mutex_init(..., H5TS_MUTEX_TYPE_RECURSIVE) in H5TSint.c), so same-thread re-entry is already handled by the regular lock -- the NOLOCK bypass was unnecessary and added a second, easy-to-misuse code path. Removed FUNC_ENTER_API_NOINIT_NOLOCK/FUNC_LEAVE_API_NOINIT_NOLOCK entirely and switched H5Zconfig_has_key/get_int/get_double/get_bool/get_str to the standard FUNC_ENTER_API_NOINIT/FUNC_LEAVE_API_NOINIT.

bin/format_source
brtnfld: Yes, we are vendoring it,  so we don't want/need to format it.

brtnfld:
Closing this for now until the implementation of the expanded scope matures.

brtnfld:
Superseded by #6663, same content but from `feature/filter_config_string` (an HDFGroup/hdf5 branch) instead of my fork's `6153` branch.

PR #6475, opened 2026-06-19
Add flag -tp=x86-64-v3 to CMAKE C, Fortran, and CXX flags to avoid crashing …

lrknox:
…llvm in nvhpc tests.


jhendersonHDF:
"-tp=px" seems safe. "-tp=x86-64-v2" or "-tp=x86-64-v3" would probably be safe as well (just to include some optimization/alternative instructions).

lrknox:
@jhendersonHDF "-tp=x86-64-v3"  also works.  Do you prefer that one?

jhendersonHDF:
> @jhendersonHDF "-tp=x86-64-v3" also works. Do you prefer that one?

For this particular case on this runner I think it makes sense and should generally be fine

jhendersonHDF (dismissed):

jhendersonHDF (approved):

PR #6476, opened 2026-06-20
validate free space section type during sinfo decode

naruto-lgtm:
## Describe your changes

Repro: open a file whose free space section info block carries a section type byte >= the number of registered section classes (the file FSM registers 3: simple/small/large).
Cause: H5FS__cache_sinfo_deserialize() reads the one-byte type and indexes fspace->sect_cls[sect_type], then calls the class deserialize callback through it. The only guard is an assert(), which is compiled out in release builds, so the type is unvalidated there.
Fix: reject sect_type that is not less than fspace->nclasses before the array is indexed, the same way the header decode already rejects a class count past fspace->nclasses.

Without the check a corrupted file drives an out-of-bounds read of the class array and an indirect call through whatever function pointer lands at that offset. Keeping the validation at the decode site is the right layer because that byte is the trust boundary; every later sect_cls[type] use operates on sections whose type was already accepted here.


fortnern (dismissed):

release_docs/CHANGELOG.md
```diff
@@ -151,6 +151,9 @@ The `h5repack` tool now obtains its default low and high library version bounds
 
 ## Library
 
+### Validate free space section type during decode
+
+   When loading a free space section info block, the per-section type byte read from the file was used directly to index the free space manager's section class array and to call the class `deserialize` callback, guarded only by an assertion that is removed in release builds. A corrupted or fuzzed file could supply a type beyond the number of registered classes, causing an out-of-bounds read of the class array and an indirect call through a bogus function pointer. `H5FS__cache_sinfo_deserialize()` now rejects a section type that is not less than the number of section classes.
 ### Fixed a heap buffer overflow when decoding a shared message list
```
lrknox, line 158:
```suggestion

### Fixed a heap buffer overflow when decoding a shared message list
```

lrknox (approved):

fortnern (approved):

PR #6478, opened 2026-06-22
ROS3 VFD block cache feature

jhendersonHDF:
Implements a minimal I/O block cache for files without paged allocation enabled to better optimize I/O and reduce requests to S3

I/O is cached in fixed-size blocks (16MiB blocks by default) and are kept in a simple LRU cache that evicts the oldest used block when a new block needs to be cached

Reading and caching of the initial bytes of a file has been delayed from file open to the first read for a file instead

API functions have been added for setting/getting the I/O block caching parameters to be used


hyoklee (changes requested):
I don't see any significant performance improvement backed by real data and numbers in release_docs.

jhendersonHDF:
> I don't see any significant performance improvement backed by real data and numbers in release_docs.

"Significant performance improvement" is not the main focus of this PR, reducing the number of requests to S3 is the main focus. Depending on the file in question and the I/O access pattern, there may be no actual benefit. Data is still being gathered on the improvements for files that do not use paged file space allocation, but listing any concrete numbers in CHANGELOG.md will not offer much insight, as any improvements will depend on I/O access pattern.

src/H5FDros3.c
```diff
@@ -1022,23 +1255,49 @@ H5FD__ros3_open(const char *url, unsigned flags, hid_t fapl_id, haddr_t maxaddr)
...
+        if (file->page_cache.page_cache_size != H5F_PAGE_BUFFER_SIZE_DEFAULT)
+            page_buf_size = file->page_cache.page_cache_size;
+        else {
+            if (H5P_get(plist, H5F_ACS_PAGE_BUFFER_SIZE_NAME, &page_buf_size) < 0)
+                HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, NULL, "can't get page buffer size");
+            if (page_buf_size == H5F_PAGE_BUFFER_SIZE_DEFAULT)
+                page_buf_size = ROS3_DEF_PAGE_BUF_SIZE;
+        }
+
+        file->page_cache.max_num_pages = (page_buf_size / file->page_cache.page_size);
```
mattjala: If `H5Pset_page_buffer_size()` is used to set the page buffer below 16 MiB and `H5Pset_fapl_ros3_paging()` is not used, this computation will yield zero.

It seems like this can cause a crash during a later read. In the main caching loop, `make_space`'s `HASH_COUNT < max_num_pages` check will resolve false, so it would attempt an eviction on an empty LRU list. With the default `lock_super_page = true`, the superblock page isn't in the LRU, so LRU_tail would be NULL and `ROS3_PAGE_CACHE_LRU_REMOVE` would try to dereference a null page.

Just clamping max_num_pages to a minimum of 1 here should resolve this.

src/H5FDros3.c
```diff
@@ -1367,12 +1690,339 @@ H5FD__ros3_read(H5FD_t *_file, H5FD_mem_t H5_ATTR_UNUSED type, hid_t H5_ATTR_UNU
...
+    }
+
+    assert(io_size == 0);
+
+    *io_reqs_out     = io_reqs;
+    *num_io_reqs_out = num_pages;
+
+done:
+    FUNC_LEAVE_NOAPI(ret_value)
+} /* end H5FD__ros3_determine_io_pages() */
```
mattjala: Ending comment here has the wrong function name

src/H5FDros3.c
jhendersonHDF: This is true; the page buffer size should act as the final say of how many bytes can be allocated, so the page size should be rounded down to the page buffer size in that case, similar to what `H5Pset_fapl_ros3_paging()` does.

mattjala: I'm talking about the value for `max_num_pages`, not the size of the pages themselves. It seems like storing zero for the max page number could lead to the crash I described later on.

jhendersonHDF: Yes, when the chosen page buffer size is smaller than the default page size, the page size should be rounded down to the page buffer size, resulting in 1 for `max_num_pages`. Just rounding `max_num_pages` up to 1 without adjusting the page size would result in a 16MiB (by default) page being allocated even though the "caller" asked for an upper limit on the page buffer size that's smaller than 16MiB.

mattjala (dismissed):

hyoklee:
> > I don't see any significant performance improvement backed by real data and numbers in release_docs.
> 
> "Significant performance improvement" is not the main focus of this PR, reducing the number of requests to S3 is the main focus. Depending on the file in question and the I/O access pattern, there may be no actual benefit. Data is still being gathered on the improvements for files that do not use paged file space allocation, but listing any concrete numbers in CHANGELOG.md will not offer much insight, as any improvements will depend on I/O access pattern.

Did you finish gathering data?

jhendersonHDF:
@hyoklee, as provided by @ajelenak:

[cohdf5_S3_requests_report.html](https://github.com/user-attachments/files/29711989/cohdf5_S3_requests_report.html)

hyoklee:
@jhendersonHDF , please update `release_docs` performance section according to the report.

hyoklee (dismissed):

jhendersonHDF:
Updated the terminology from 'page' -> 'block' and changed the default cache size from 64MiB to 128MiB

release_docs/CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ For releases prior to version 2.0.0, please see the release.txt file and for mor
 
 ## Performance Enhancements:
 
+- Added an I/O block cache to the ROS3 VFD to reduce requests to S3 for files not using paged allocation
```
hyoklee: `Added an I/O block cache to the ROS3 VFD to reduce S3 requests for files that do not use paged allocation.`

release_docs/CHANGELOG.md
```diff
@@ -98,6 +100,10 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## Library
 
+### Added an I/O block cache to the ROS3 VFD
+
+   Added an I/O block cache to the ROS3 VFD to reduce requests to S3 for files that don't use paged allocation. This is a simple LRU cache that performs I/O in fixed-size blocks and serves I/O requests from the in-memory cached buffers. By default, the ROS3 VFD now performs I/O in 16MiB (see new macro `HDF5_ROS3_VFD_DEFAULT_BLOCK_SIZE`) blocks, caching up to a total of 128MiB (see new macro `HDF5_ROS3_VFD_DEFAULT_BLOCK_CACHE_SIZE`) of data at a time. The new `H5Pset_fapl_ros3_block_caching()` / `H5Pget_fapl_ros3_block_caching()` API functions can be used to modify or retrieve the caching parameters set on a File Access Property List, respectively. Additionally, caching of the initial bytes of a file has been delayed from file open to the first read of a file instead to reduce the overhead of file opens.
```
hyoklee: requests to S3 -> S3 requests
serves I/O requests -> servers requests (I/O is redundant)
16MiB -> 16 MiB
128MiB -> 128 MiB

hyoklee (dismissed):

release_docs/CHANGELOG.md
jhendersonHDF: Updated

release_docs/CHANGELOG.md
jhendersonHDF: Added "the number of" to clarify a bit more, but "requests to S3" and "S3 requests" read the same to me

hyoklee (approved):

mattjala (approved):

PR #6499, opened 2026-07-01
reject SOHM list message count exceeding list_max

naruto-lgtm:
## Describe your changes

When a shared object header message index is stored as a list, both `H5SM__cache_list_verify_chksum()` and `H5SM__cache_list_deserialize()` take the `num_messages` count straight from the on-disk index header. The image buffer and the `list->messages` array are both sized for `list_max` entries, so a file whose `num_messages` is larger than `list_max` makes verify_chksum size the checksum region past the buffer and makes the deserialize loop write past the array and read past the buffer. A list holds at most `list_max` messages on disk before it is promoted to a B-tree, so any file that violates that is corrupt.

Reject `num_messages > list_max` at both callbacks before the count is used to size a read or index the array. That header field is the trust boundary here, and the rest of the SOHM code already assumes the count fits within `list_max`, so validating it once at load time keeps every later `list->messages` access sound.

#6501 

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -144,6 +144,10 @@ The `h5repack` tool now obtains its default low and high library version bounds 
 
 ## Library
 
+### Fixed a heap buffer overflow when decoding a shared message list
+
+   When reading a shared object header message (SOHM) list from the metadata cache, `H5SM__cache_list_deserialize()` allocated the message array for `list_max` entries but drove the decode loop with the `num_messages` count read from the on-disk index header. A corrupted or malicious file whose `num_messages` exceeds `list_max` caused writes past the end of the array and reads past the end of the input buffer. The count is now validated against `list_max` before the loop runs.
+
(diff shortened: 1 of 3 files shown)


hyoklee (dismissed):
`release_docs` looks good.

lrknox (changes requested):
A test should be included if possible.

mattjala (dismissed):

naruto-lgtm:
Added a test in tsohm.c (test_sohm_reject_bad_count). It creates a file with a list-index SOHM, pushes the on-disk message count past list_max and repairs the table checksum, then reopens and shares a message so the list gets protected. With the fix that load is rejected cleanly; on the pre-fix code the same path is the heap-buffer-overflow read in H5F_get_checksums via H5SM__cache_list_verify_chksum, so it trips under ASan. It's guarded to the default sec2 driver since it edits the file image directly.

mattjala (approved):

lrknox (approved):

glennsong09 (approved):

PR #6504, opened 2026-07-04
Fix javadoc errors

matteodg:
The branch tries to resolve at least the javadoc errors (there are also a lot of warnings, but not addressed in this branch).
This helps the GitHub builds to pass.


java/hdf/hdf5lib/HDF5Constants.java
```diff
@@ -33,7 +33,7 @@
  * <P>
  * <B>Do not edit this file!</b>
  *
- * @see @ref HDF5LIB
+ * @ref HDF5LIB
```
jhendersonHDF, line 36: @matteodg I believe the reason `@see @ref ref` is used here is because this is intended to be for Doxygen and not javadoc, so this change would remove the "See also" at https://support.hdfgroup.org/documentation/hdf5/latest/_h_d_f5_c_o_n_s_t.html. That's a very minor change that wouldn't affect much, but I think it brings up the point that we have to consider how to make the two work together.

java/hdf/hdf5lib/H5.java
```diff
@@ -19309,7 +19302,7 @@ public static String H5Rget_attr_name(byte[] ref_ptr)
      *
      * @see H5S, C-API
```
mattjala, line 19303: The @defgroup headers for H5S, H5T, H5VL, and H5Z still use `@see H5S, C-API [...]`, while every other interface here was converted to @ref. These should be updated to to `@ref H5x, C-API [...]` for consistency.

mkitti:
Opened a follow-up PR against this branch addressing the two open review comments (from @jhendersonHDF and @mattjala on the `@see`/`@ref` conversions): matteodg#2

Feel free to merge that in whenever convenient.

matteodg:
Thanks @mkitti!
Sorry for the delay: I'm reviewing the whole branch and there are other places where I removed the `@see` and kept only `@ref`, I guess I should add "See also" before `@ref` there as well, right?

For example in `java/src-jni/hdf/hdf5lib/HDFNativeData.java:32-34`:
```java
 * <P>
 * @ref HDFARRAY.
 */
```

PR #6506, opened 2026-07-04
Validate metadata cache load sizes before allocation

Alb3e3:
## Describe your changes

This hardens malformed-file handling for metadata cache entry loads:

- Check non-speculative metadata cache entry image sizes against the file EOA before allocating the on-disk image buffer. Existing speculative load behavior is preserved: speculative reads can still be trimmed to EOA before retry.
- Validate decoded object header continuation chunk address/size pairs before queuing continuation chunks.
- Add an `ohdr` regression that creates an object header continuation chunk, corrupts the encoded continuation size on disk, and verifies the malformed object header is rejected before the cache allocates that size.

This prevents malformed metadata from driving very large allocations before the later file-read/EOA checks run.

AI assistance disclosure: this patch was prepared with assistance from OpenAI Codex. I reviewed the changes and ran the verification below locally.

N/A

## Local verification

- `git diff --check origin/develop..HEAD`
- `cmake --build build-asan-o0 --target ohdr h5dump -j$(nproc)`
- `ASAN_OPTIONS=detect_leaks=0:allocator_may_return_null=0:abort_on_error=1 timeout 180s build-asan-o0/bin/ohdr`
- 17 malformed h5dump crash-corpus files under ASan: no AddressSanitizer/UBSan/OOM output and no timeouts; all exit as ordinary open failures.
- Valid-file sanity check: `build-asan-o0/bin/h5dump -H test/testfiles/tmtimeo.h5`


Alb3e3:
Pushed a follow-up for the CI failures. The first issue was formatting in `test/ohdr.c`. The Linux SWMR failures came from the new metadata load-size EOA guard rejecting reads against a SWMR reader's stale local EOA before the existing lower read path could handle writer-flushed metadata.

The update keeps the non-SWMR metadata size checks in place, removes the too-early continuation-message EOA check, and skips this cache-entry EOA validation only for `H5F_ACC_SWMR_READ`, matching the existing VFD/SWMR read behavior.

Local verification:
- `git diff --check`
- `cmake --build /tmp/hdf5build-pr6506 --target flushrefresh swmr_check_compat_vfd ohdr use_disable_mdc_flushes twriteorder use_append_chunk use_append_mchunks -j2`
- `ctest -V -R '^H5SHELL-test_flush_refresh$'`
- `ctest -V -R '^H5SHELL-test_use_cases$'`
- `./bin/ohdr`

hyoklee (dismissed):
`release_docs` looks good.

src/H5Ocont.c
```diff
@@ -23,9 +23,11 @@
  */
 
 #include "H5Omodule.h" /* This source code file is part of the H5O module */
+#define H5F_FRIEND     /*suppress error about including H5Fpkg   */
```
mattjala: The `H5F_FRIEND` here and the `H5Fpkg.h` inclusion below are unnecessary and should be removed

src/H5Centry.c
```diff
@@ -954,6 +954,14 @@ H5C__verify_len_eoa(H5F_t *f, const H5C_class_t *type, haddr_t addr, size_t *len
 
     FUNC_ENTER_PACKAGE
 
+    /*
+     * SWMR readers can observe metadata flushed by a writer before their local
+     * EOA cache is refreshed.  Lower VFD read paths already bypass EOA
+     * validation for SWMR reads and rely on read retries / EOF handling.
+     */
+    if (H5F_INTENT(f) & H5F_ACC_SWMR_READ)
```
jhendersonHDF: I'm somewhat suspicious about the need for this addition; does this become an issue only after the addition of the call to `H5C__verify_len_eoa()` in `H5C__load_entry()`? That sticks out to me as indicating that `H5C__load_entry()` may not be the place this issue needs to be solved at.

mattjala: This early-return is in a shared helper, so in addition to the new non-speculative check, it also affects two pre-existing speculative call sites (the actual=false trim at the H5C__CLASS_SPECULATIVE_LOAD_FLAG check, and the actual=true post-read check on actual_len). So for SWMR readers, speculative loads are no longer trimmed to EOA, and speculative growth past EOA no longer throws an error. This is a change to long-standing paths that this PR is mostly unrelated to.

I think the bypass is only actually needed for the new non-speculative check. If it were moved to that call site instead of the shared helper, it would preserve the pre-existing SWMR behavior.

If a global SWMR bypass is intended, a SWMR regression test should be added. The current test is single-file only and skips non-default VFDs.

src/H5Ocont.c
Alb3e3: Removed both `H5F_FRIEND` and the `H5Fpkg.h` include. The existing H5O package headers provide what this decoder needs. The Debug build and full `ohdr` test binary pass after the removal.

src/H5Centry.c
Alb3e3: Yes—the conflict was specific to applying the new non-speculative EOA check to SWMR readers. I kept the validation in `H5C__load_entry()` because this is the common point after the client computes `len` and immediately before allocating `len + H5C_IMAGE_EXTRA_SPACE`; validating later in a decoder would be after the allocation we need to bound. The SWMR exception is now scoped only to that new non-speculative call site, while the shared helper retains its pre-existing behavior for speculative clients. The focused `cache` and full `ohdr` binaries pass in a Debug build.

Alb3e3: Agreed. I moved the SWMR bypass to the new non-speculative call site. `H5C__verify_len_eoa()` now always trims or rejects speculative lengths as before, including for SWMR readers, so neither pre-existing speculative path is bypassed. A global SWMR bypass is not intended; the existing `ohdr` SWMR cases and the focused `cache` test both pass with the corrected scope.

mattjala (dismissed):

fortnern:
Are there later checks that could be changed into asserts with this change? Or are there other paths to reach those later checks besides through H5C__load_entry()?

Alb3e3:
Good questions — I looked at both.

**Could later checks become asserts?**

I'd recommend against it for the checks inside the client `deserialize()` callbacks, because this change doesn't make them unconditional:

- The SWMR carve-out means a non-speculative load under `H5F_ACC_SWMR_READ` still reaches `H5MM_malloc()` and `deserialize()` without an EOA check. Anything downgraded to an assert would become a release-build gap on exactly that path.
- For speculative loads the first pass still only gets the clamping form (`actual = false`); the strict form only runs afterwards if `get_final_load_size()` changes the length.

The only assert that would be sound is one in `H5C__load_entry()` itself right before the `H5MM_malloc()`, and that is tautological with the check I just added, so it buys nothing.

**Are there other paths to those later checks?**

Yes — `H5C__deserialize_prefetched_entry()` is a second `type->deserialize()` call site that never goes through `H5C__load_entry()`. It is the cache-image restore path, and it deliberately skips the load path's sizing logic; the comment above it says so:

> Since the size of the on disk image is known exactly, there is no need for either a call to the `get_initial_load_size()` callback, or retries if the `H5C__CLASS_SPECULATIVE_LOAD_FLAG` flag is set. Similarly, there is no need to clamp possible reads beyond EOF.

There `len = pf_entry_ptr->size`, which is decoded from the cache image block — i.e. it comes from the file.

While checking that, I noticed the "known exactly" premise doesn't hold for a malformed file. In `H5C__decode_cache_image_entry()` on current `develop`:

```c
H5F_addr_decode(f, &p, &pf_entry_ptr->addr);

/* Validate address range */
eoa = H5F_get_eoa(f, H5FD_MEM_DEFAULT);
if (!H5_addr_defined(pf_entry_ptr->addr) || H5_addr_overflow(pf_entry_ptr->addr, pf_entry_ptr->size) ||
    H5_addr_ge(pf_entry_ptr->addr + pf_entry_ptr->size, eoa))
    HGOTO_ERROR(H5E_CACHE, H5E_BADVALUE, NULL, "invalid entry address range");

/* Decode entry length */
H5F_DECODE_LENGTH(f, p, pf_entry_ptr->size);
```

The range check runs *before* the length is decoded. `pf_entry_ptr` comes from `H5FL_CALLOC()`, so `->size` is still 0 there — the overflow and EOA terms are evaluated against zero and only the address is meaningfully validated. The real, file-supplied size is decoded on the next line and never re-checked against EOA.

That is not an out-of-bounds read in the decoder itself: the following `H5MM_malloc()` and image copy are bounded by `H5_IS_BUFFER_OVERFLOW()` against the cache image block. But it does mean a prefetched entry can carry a size that was never validated against EOA straight into `type->deserialize()` — which is the concrete reason I'd keep the client-side checks as real error checks rather than asserts.

Moving that validation to after the length decode looks like a small, self-contained fix. Happy to fold it into this PR or open a separate one — whichever you prefer.

release_docs/CHANGELOG.md
```diff
@@ -221,6 +221,11 @@ The `h5repack` tool now obtains its default low and high library version bounds
 ### Hardened decoding of serialized dataspace selections against malformed buffers
 
    `H5S_select_deserialize()` and the per-selection-type deserialize callbacks (all, hyperslab, none, and point) previously computed the pointer to the last valid buffer byte as `buffer + size - 1` without first checking the buffer size. A buffer shorter than the 4-byte selection-type header, or a zero-length selection-info buffer, would underflow this computation and produce an out-of-bounds end pointer, defeating subsequent overflow checks. The deserialize routines now reject a buffer that is too small to hold the selection type, and they reject an empty selection-info buffer before deriving the end pointer. Hyperslab decoding additionally now rejects a serialized rank of 0 or greater than `H5S_MAX_RANK`. As a companion fix, `H5S__hyper_serialize()` now returns an error when asked to serialize a hyperslab selection on a rank-0 (scalar or null) dataspace, a state that can arise when a dataspace extent is collapsed to a scalar after a hyperslab selection has already been made.
+### Added pre-allocation validation for metadata cache entry image sizes
```
hyoklee, line 97: Please add an extra line before new heading.

release_docs/CHANGELOG.md
```diff
@@ -221,6 +221,11 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
+
+   Metadata cache entries with non-speculative load sizes are now checked against the file's end of
```
hyoklee: Please do not split lines.

hyoklee (dismissed):

release_docs/CHANGELOG.md
lrknox: Addressed format comments for CHANGELOG.md entry while resolving conflicts

release_docs/CHANGELOG.md
lrknox: Addressed format comments for CHANGELOG.md entry while resolving conflicts

hyoklee (dismissed):

Alb3e3:
@mattjala Friendly ping — this PR has hyoklee's approval and the CI is green. Could you check off the src/test areas in the review checklist when you have a moment? Happy to address any remaining concerns.

mattjala:
I think this can go in as it is now, though there's still a CHANGELOG conflict to resolve.

mattjala (dismissed):

Alb3e3:
Rebased this branch onto current `origin/develop` at `51b92ccee5` and resolved the remaining CHANGELOG overlap while keeping the existing metadata-cache/object-header hardening intact.

Local verification on `1349ae70d118fcc85e8e41bded79a457a3b922fc`:
- `git diff --check origin/develop..HEAD`
- `cmake --build build-asan-o0 --target ohdr h5dump -j8`
- `ASAN_OPTIONS=detect_leaks=0:allocator_may_return_null=0:abort_on_error=1 timeout 180s build-asan-o0/bin/ohdr`
- `for f in ../crashes/h5dump/*.h5; do printf '== %s ==\\n' "$(basename "$f")"; env ASAN_OPTIONS=detect_leaks=0:allocator_may_return_null=0:abort_on_error=1 timeout 30s build-asan-o0/bin/h5dump -H "$f"; printf 'rc=%s\\n' "$?"; done`\n- `build-asan-o0/bin/h5dump -H test/testfiles/tmtimeo.h5`\n\nResults: `git diff --check` was clean, `ohdr` passed, all 17 malformed `h5dump` corpus files exited as ordinary failures with no ASan/UBSan/OOM/timeout output, and the valid-file `h5dump` sanity check passed.

Alb3e3:
Friendly ping — this PR has been approved by both @hyoklee and @mattjala for a while. Happy to address any remaining concerns or rebase again if the branch is behind. Just want to make sure it doesn't fall through the cracks.

release_docs/CHANGELOG.md
```diff
@@ -94,6 +94,78 @@ We would like to thank the many HDF5 community members who contributed to this r
 
    Fixes GitHub issue #6531
 
+### Fixed bug that prevented internal library filters from printing error messages
```
mattjala: It looks like the `CHANGELOG.md` conflict got merged in a way that brought in a bunch of unrelated messages from `2.2.0`.

mattjala (changes requested):

Alb3e3:
Good catch — the CHANGELOG conflict resolution on 2026-08-20 accidentally pulled in a large block of unrelated 2.2.0 entries (62 lines, from the "Fixed bug that prevented internal library filters" section down through the "incorrect file format validation check" entry). Force-pushed `f90df7b317` to trim those out. The diff against `origin/develop` now adds only the two entries that belong to this PR's work.

release_docs/CHANGELOG.md
```diff
@@ -94,6 +94,16 @@ We would like to thank the many HDF5 community members who contributed to this r
 
    Fixes GitHub issue #6531
 
+### Hardened decoding of serialized dataspace selections against malformed buffers
```
mattjala: This CHANGELOG message is from #6525 and should not be included in this PR

mattjala (changes requested):

Alb3e3:
Good catch — removed. That entry was for #6525 and got pulled in during the CHANGELOG conflict resolution.

Force-pushed `38180fb065`. The CHANGELOG diff against `origin/develop` is now exactly one entry, the one that belongs to this PR:

```
+### Added pre-allocation validation for metadata cache entry image sizes
+
+   Metadata cache entries with non-speculative load sizes are now checked against the file's end of
+   allocation before allocating the on-disk image buffer. Object header continuation messages also
+   validate their decoded address and size fields before queuing continuation chunks.
```

No code changes in this push — `src/H5Centry.c`, `src/H5Ocont.c`, and `test/ohdr.c` are untouched since your last review.

release_docs/CHANGELOG.md
```diff
@@ -94,6 +94,12 @@ We would like to thank the many HDF5 community members who contributed to this r
 
    Fixes GitHub issue #6531
 
+### Added pre-allocation validation for metadata cache entry image sizes
+
+   Metadata cache entries with non-speculative load sizes are now checked against the file's end of
+   allocation before allocating the on-disk image buffer. Object header continuation messages also
```
hyoklee: `file's end of allocation` -> `file's end-of-allocation address`

hyoklee (dismissed):

Alb3e3:
Rebased onto current `develop` and resolved the `CHANGELOG.md` conflict. The changelog now contains only this PR's single entry ("Added pre-allocation validation for metadata cache entry image sizes"); the unrelated 2.2.0/#6525 messages that a prior conflict resolution had pulled in are gone. Also applied the review wording fixes ("file's end-of-allocation address", single-line entry body, blank line before the heading). No source changes in this update — the H5Centry.c / H5Ocont.c fixes are unchanged.

mattjala (dismissed):

fortnern:
Is there a possibility that a speculative load in SWMR read mode could inappropriately run afoul of this check, since the reader might not see the updated EOA before reading a reference to a new piece of metadata? Or are metadata clients that use speculative loads not allowed to be written in SWMR mode by the current SWMR implementation? I realize this would be a problem in the current code (before this PR) as well.

Alb3e3:
Yes — for a hypothetical speculative cache client in SWMR read mode, the existing final-size check could reject an entry when the reader has a stale EOA. `H5C__load_entry()` already validates the speculative initial size with `actual = false`, then validates the resolved size with `actual = true` after `get_final_load_size()`; neither of those paths changed in this PR.

I checked the current cache client definitions as well: no production HDF5 metadata client sets `H5C__CLASS_SPECULATIVE_LOAD_FLAG`; its only use is the synthetic `variable_class` in `test/cache_common.c`. So this is not reachable in current SWMR metadata I/O.

The new pre-allocation guard applies only to non-speculative loads and explicitly skips SWMR readers. A future production client that adopts speculative loads would need to decide how its final-size validation should synchronize with SWMR EOA visibility, but this PR preserves the pre-existing speculative behavior.

hyoklee (approved):

Issue #6430, opened 2026-06-03
Format specification: missing format description for datatype encoding


vchoi-hdfgroup:
***
 
**Describe the bug**
Format description for datatype encoding is missing.

**Expected behavior**
The datatype encoding is done via the API H5Tencode/H5Tdecode.
Add a section "Datatype Encoding" to the Appendix.
This will be similar to the section "Dataspace Encoding" in Appendix VIII.A.

vchoi-hdfgroup:
Fix is done with PR #6510

PR #6510, opened 2026-07-07
Fix for issue #6430: add decription for H5Tencode/H5Tdecode to Appendix.

vchoi-hdfgroup:
Put the description to the Appendix.


gheber (approved):
Looking good.

PR #6517, opened 2026-07-10
Fix write-path integer overflow in scale-offset filter

### LINKED ISSUE #5861: Read errors for some chunks after upgrading to HDF5 1.14.6

mattjala:
This started as a regression test for #5864, `test_scaleoffset_int_fullprec` in `test/dsets.c`. It writes integer data spanning the entire type range, which forces the filter into its full-precision fallback and exercises the read path that #5864 fixed.

The test ran into a different bug, a signed-integer overflow on the write path. The filter was computing `max - min` in signed int which, for full-range data, would overflow a signed int.

That subtraction is now performed in unsigned arithmetic instead, which will be safe for any range of input values.

I also added a pre-emptive test for a the analgous full-precision for float datatypes (`test_scaleoffset_float_fullprec`), though there's no existing problems in that case.

Resolves #5861


fortnern (dismissed):

release_docs/CHANGELOG.md
```diff
@@ -145,6 +145,10 @@ The `h5repack` tool now obtains its default low and high library version bounds
 
 ## Library
 
+### Fixed signed integer overflow in the scale-offset filter for full-range data
+
+   When a signed-integer dataset filtered with scale-offset contained values spanning the full range of the datatype, the filter's calculation of the difference between the maximum and minimum value would overflow, resulting in undefined behavior. The difference is now computed in the unsigned domain, producing the same result without triggering an overflow.
```
hyoklee: `value` -> `values`

`the same result`  -> `the correct result`

hyoklee (changes requested):

hyoklee (dismissed):

hyoklee (dismissed):

hyoklee (dismissed):

hyoklee (dismissed):

PR #6518, opened 2026-07-10
Detect filtered chunk size overflow with small size of sizes

### LINKED ISSUE #6023: Test setting size of sizes to less than 8 with chunked datasets
The dataset chunk code uses the configurable "size of sizes" set by H5Pset_sizes for encoding various quantities. However, we do not currently test chunekd datasets with any value other than the default 8. We should add testing for other sizes of sizes (2 and 4), making sure to cover all relevant cases and ensure overflow is detected and handled correctly, and fix any issues that are uncovered by this testing.

mattjala:
For all chunk index types in version-5 chunk layout messages, and for the single chunk index in any version of the chunk layout message, the on-disk size of a filtered chunk is encoded in a fixed-width field equal to the file's "size of sizes" (set via H5Pset_sizes()). The encode check assumes this field is always 8 bytes, so when the size of sizes is set to 2 or 4 an oversized chunk has its encoded size silently truncated (e.g. 160000 encoded in 2 bytes becomes 160000 & 0xFFFF = 28928), silently corrupting the chunk.

The library now verifies that a filtered chunk's size fits in the file's size of sizes and reports an error at write time instead, via a check in the H5D_CHUNK_ENCODE_SIZE_CHECK macro (src/H5Dchunk.c), which runs from H5D__chunk_file_alloc() before any chunk storage is allocated.

The in-range case, when the filtered chunk is small enough to be encoded by a small size of sizes, is tested by `test_chunk_size_of_sizes`, and the case where a filtered chunk is too large to be encoded is tested by `test_chunk_size_of_sizes_overflow`.

Resolves #6023


hyoklee (approved):

hyoklee (dismissed):

release_docs/CHANGELOG.md
```diff
@@ -222,6 +222,10 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
 
+### Fixed silent truncation of filtered chunk sizes with a small file "size of sizes"
+
+   For all chunk index types in version-5 chunk layout messages, and for the single chunk index in any version of the chunk layout message, the on-disk size of a filtered chunk is encoded in a fixed-width field equal to the file's "size of sizes" (set via `H5Pset_sizes()`). The encode check assumed this field was always 8 bytes, so when the size of sizes was set to 2 or 4 an oversized chunk had its encoded size silently truncated (e.g. 160000 encoded in 2 bytes became 160000 & 0xFFFF = 28928), silently corrupting the chunk. The library now verifies that a filtered chunk's size fits in the file's size of sizes and reports an error at write time instead.
```
hyoklee: `e.g.` -> `e.g.,`

hyoklee (changes requested):

hyoklee (dismissed):

hyoklee (dismissed):

hyoklee (dismissed):
Copilot's suggestion for CHANGELOG.md: 

```
### Fixed silent truncation of filtered chunk sizes when the file "size of sizes" is small

For all chunk index types in version‑5 chunk layout messages, and for the single chunk index in any version of the chunk layout message, the on‑disk size of a filtered chunk is encoded in a fixed‑width field equal to the file's "size of sizes" (set via `H5Pset_sizes()`). The encode check previously assumed this field was always 8 bytes. When the size of sizes was set to 2 or 4, an oversized chunk's encoded size could be silently truncated (for example, 160000 encoded in 2 bytes becomes 160000 & 0xFFFF = 28928), corrupting the chunk. The library now verifies that a filtered chunk's size fits within the file's size of sizes and reports an error at write time.

Fixes GitHub issue #6023
```

PR #6509, opened 2026-07-07
Harden serialized selection decoding

Alb3e3:
## Summary

This hardens serialized dataspace selection decoding against malformed or truncated selection buffers.

The patch adds explicit length validation before decoding the common serialized selection type, passes the remaining selection-type-specific payload size to the concrete deserializers, rejects zero-length concrete selection payloads, and validates the decoded hyperslab rank before it is used to patch or allocate a dataspace.

## Motivation and context

Several selection deserializers computed `*p + p_size - 1` before proving that `p_size` was non-zero. A malformed serialized selection can therefore underflow the end pointer and let later bounds checks reason about an invalid buffer range. Hyperslab selection decoding also accepted an unchecked serialized rank before patching the dataspace rank.

## Testing

Built and tested with Clang 18 ASan/UBSan on `develop`:

- `cmake -S . -B build-asan -G "Unix Makefiles" -DCMAKE_C_COMPILER=clang-18 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTING=ON -DHDF5_BUILD_TOOLS=ON -DHDF5_BUILD_CPP_LIB=OFF -DHDF5_BUILD_FORTRAN=OFF -DHDF5_BUILD_JAVA=OFF -DHDF5_ENABLE_PARALLEL=OFF -DCMAKE_C_FLAGS_RELWITHDEBINFO="-O1 -g -DNDEBUG -fsanitize=address,undefined -fno-omit-frame-pointer" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined"`
- `cmake --build build-asan --target testhdf5 h5dump h5ls -j2`
- `ASAN_OPTIONS=allocator_may_return_null=1:detect_leaks=0:abort_on_error=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1 build-asan/bin/testhdf5 -only select`
- 57-file `test/testfiles` tool baseline across 456 `h5dump`/`h5ls` invocations: 0 sanitizer failures. The existing slow `tlayouto.h5` default `h5dump` case still timed out and was treated separately from sanitizer findings.
- Targeted mutation fuzz pass over the same 57-file corpus: 6,526 runs, 0 new sanitizer findings, 0 skipped known-submitted findings.

AI assistance was used to find and prepare this patch. I reviewed the changes and verified them locally with the commands above.

PR #6525, opened 2026-07-13
Harden selection decoding

mattjala:
Updated version of #6509 which is clang-formatted and has a couple extra cases covered in the new test


src/H5Spoint.c
```diff
@@ -1375,6 +1375,8 @@ H5S__point_deserialize(H5S_t **space, const uint8_t **p, const size_t p_size, bo
 
     if (skip)
         p_end = *p;
+    else if (p_size == 0)
```
jhendersonHDF, line 1378: Is there an example of how this caused a problem? `p_end` should end up as `*p - 1`, which I'd think should immediately fail the precondition in `H5_IS_BUFFER_OVERFLOW()` / `H5_IS_KNOWN_BUFFER_OVERFLOW()` of checking `(ptr) > (buffer_end)`.

jhendersonHDF: I suppose being that this falls under undefined behavior, some or most compilers might just be removing the `(ptr) > (buffer_end)` check anyway

mattjala: Not specifically. This should already be guarded against by the changes in the outer routines, and was added here mostly as a precaution.

src/H5Shyper.c
```diff
@@ -4328,6 +4330,9 @@ H5S__hyper_deserialize(H5S_t **space, const uint8_t **p, const size_t p_size, bo
     if (H5_IS_KNOWN_BUFFER_OVERFLOW(skip, pp, sizeof(uint32_t), p_end))
         HGOTO_ERROR(H5E_DATASPACE, H5E_OVERFLOW, FAIL, "buffer overflow while decoding selection rank");
     UINT32DECODE(pp, rank);
+    if (0 == rank || rank > H5S_MAX_RANK)
```
jhendersonHDF, line 4341: I'm not 100% certain if it makes sense to have a hyperslab selection with a rank of 0 (which in HDF5 would mean a scalar point), but if not it's probably worth protecting against this in the serialize case as well. Consider this program:

```C
#include "hdf5.h"

int
main(int argc, char **argv)
{
    hsize_t dims[1]    = { 10 };
    hsize_t start[1];
    hsize_t stride[1];
    hsize_t count[1];
    hsize_t block[1];
    uint8_t buf[2048];
    size_t  buf_size   = 2048;
    hid_t   dspace_id  = H5I_INVALID_HID;
    hid_t   dspace_id2 = H5I_INVALID_HID;

    dspace_id = H5Screate_simple(1, dims, NULL);

    start[0]  = 0;
    stride[0] = 2;
    count[0]  = 5;
    block[0]  = 1;
    H5Sselect_hyperslab(dspace_id, H5S_SELECT_SET, start, stride, count, block);

    H5Sset_extent_simple(dspace_id, 0, NULL, NULL);

    H5Sencode2(dspace_id, buf, &buf_size, H5P_DEFAULT);

    dspace_id2 = H5Sdecode(buf);

    H5Sclose(dspace_id);
    H5Sclose(dspace_id2);
}
```

On my machine, this generates a `Bus error` due to negative array indexing after assigning `fast_dim` to `ndims - 1` in `H5S__hyper_serialize()`. `H5Sselect_hyperslab()` can't be called on scalar dataspaces, but a dataspace can currently be changed to scalar after the selection has already been made.

jhendersonHDF (commented):
Should include a changelog entry of some sort, but otherwise looks good

release_docs/CHANGELOG.md
```diff
@@ -185,6 +185,10 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
 
+### Hardened decoding of serialized dataspace selections against malformed buffers
+
+   `H5S_select_deserialize()` and the per-selection-type deserialize callbacks (all, hyperslab, none, and point) previously computed the pointer to the last valid buffer byte as `buffer + size - 1` without first checking the buffer size. A buffer shorter than the 4-byte selection-type header, or a zero-length selection-info buffer, would underflow this computation and produce an out-of-bounds end pointer, defeating the subsequent overflow checks. The deserialize routines now reject a buffer that is too small to hold the selection type and reject an empty selection-info buffer before deriving the end pointer. Hyperslab decoding additionally now rejects a serialized rank of 0 or greater than `H5S_MAX_RANK`. As a companion fix, `H5S__hyper_serialize()` now errors out when asked to serialize a hyperslab selection on a rank-0 (scalar or null) dataspace, a state that can arise when a dataspace extent is collapsed to a scalar after a hyperslab selection was already made.
```
hyoklee: `defeating the subsequent` -> `defeating subsequent`
`hold the selection type and reject` -> `hold the selection type, and they reject` 
`now errors out` -> `now returns an error`
`was already made` -> `has already been made`

hyoklee (changes requested):

hyoklee (approved):

PR #6529, opened 2026-07-15
Avoid converting fill value from memory to file type when reading vlen data from a nonexistent chunk

### LINKED ISSUE #6187: Problem reading vlen dataset in read only mode
When reading a non-allocated chunk from a dataset containing a variable length data type and a non-default fill value from a file opened in read only mode, the library attempts to write data to the file and therefore issues a failure. This is happening because, internally, when the library reads a non-allocated chunk it creates a chunk with the fill value in the dataset cache, and because the chunks are cached with the file datatype, the library converts the fill value first to memory, duplicates it, then converts the chunk  to file type. The conversion routine for variable length datatypes to disk then writes the variable length portion of the data to the file (in a global heap).

We should find a way to stop this write from happening, possibly by bypassing the chunk cache in cases like this and duplicating the memory format fill value to the user's buffer directly. We could possibly do this even with the file opened with write access to prevent unexpected file writes.

fortnern:
This prevents unnecessary writes to the file when reading data, and prevents errors from trying to write to a read-only file (prevents surprising writes triggered by a read operation). This is done by eliminating the "nonexistent chunk" temporary layout lass that caused the library to create copies of the file type fill value, each of which needed a global heap entry written to disk, only to convert these back to memory type. Instead the chunk code calls H5Dfill() directly and bypasses the layout/io op callback abstractions. Also prevents trying to cache nonexistent chunks that are read from.

This is an aggressive solution that applies even to non-vlen datasets. There may be cases where repeatedly reading data from a nonexistent (non-vlen) chunk with a non-default fill value could benefit from caching these chunks. If this is desired we could re-enable the caching behavior only for non-vlen datasets, while keeping the changes to the nonexistent chunk read pathway.

The rare PR that actually reduces the amount of library code!

Fixes #6187


test/dsets.c
```diff
@@ -18869,6 +18870,190 @@ test_bad_buf_size(hid_t file)
     return FAIL;
 } /* end test_bad_buf_size() */
 
+/*-------------------------------------------------------------------------
+ * Function:    test_readonly_chunk_vlen_fill
+ *
+ * Purpose:     Test that chunked vlen datasets with a no file is opened in read only mode.
```
mattjala: This is hard to parse, maybe reword to `Test reading a chunked vlen dataset with a fill value from a read-only file.`

mattjala (dismissed):

hyoklee (changes requested):
Please rewrite CHANGELOG.md:

```
# Fixed error when reading variable-length chunked datasets in read-only mode

When reading from a chunked dataset with a variable-length data type, a non-default fill value, and unwritten chunks, the library would internally attempt to write data to the file. This caused a failure when the file was opened in read-only mode.

The I/O code has been reworked to prevent these unnecessary writes. As a benefit, this change may also improve performance and file space efficiency for similar scenarios where files are opened with write access.
```

fortnern:
> Please rewrite CHANGELOG.md:
> 
> ```
> # Fixed error when reading variable-length chunked datasets in read-only mode
> 
> When reading from a chunked dataset with a variable-length data type, a non-default fill value, and unwritten chunks, the library would internally attempt to write data to the file. This caused a failure when the file was opened in read-only mode.
> 
> The I/O code has been reworked to prevent these unnecessary writes. As a benefit, this change may also improve performance and file space efficiency for similar scenarios where files are opened with write access.
> ```
None of the other bug fixes use multiple lines. In addition, I don't think the rewordings are improvements - "The I/O code has been reworked" is passive voice and "As a benefit, " is somewhat awkward.

hyoklee:
> > Please rewrite CHANGELOG.md:
> > ```
> > # Fixed error when reading variable-length chunked datasets in read-only mode
> > 
> > When reading from a chunked dataset with a variable-length data type, a non-default fill value, and unwritten chunks, the library would internally attempt to write data to the file. This caused a failure when the file was opened in read-only mode.
> > 
> > The I/O code has been reworked to prevent these unnecessary writes. As a benefit, this change may also improve performance and file space efficiency for similar scenarios where files are opened with write access.
> > ```
> 
> None of the other bug fixes use multiple lines. In addition, I don't think the rewordings are improvements - "The I/O code has been reworked" is passive voice and "As a benefit, " is somewhat awkward.

Please change `read only` to `read-only`.
Add a subject to the "Reworked" sentence for Subject-verb agreement.

mattjala (dismissed):

hyoklee (approved):

mattjala (approved):

PR #6530, opened 2026-07-15
Fix macos-26 workflow failure;  update nvhpc version to 26.5

lrknox:
macos-26-matrix.yml created a symbolic gfortran link to gfortran-15 that began failing when homebrew apparently added gfortran.  The workflow runs without the link.

### DIFF
--- a/.github/workflows/nvhpc.yml
+++ b/.github/workflows/nvhpc.yml
@@ -21,8 +21,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CUDAVERDOT: "13.1"
-      NVERDOT: "26.3"
-      NVERDASH: "26-3"
+      NVERDOT: "26.5"
+      NVERDASH: "26-5"
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/nvhpc.yml
```diff
@@ -21,8 +21,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CUDAVERDOT: "13.1"
```
mattjala: Should this be bumped to 13.2? From [the downloads page](https://developer.nvidia.com/hpc-sdk/releases/26.5) for 26.5 it looks like it only bundles with 12.9 and 13.2

hyoklee (changes requested):
See @mattjala 's comment.

.github/workflows/nvhpc.yml
lrknox: Done.

hyoklee (approved):

PR #6532, opened 2026-07-16
Remove abort on infinite loop during library close

### LINKED ISSUE #6531: abort occurs only if display error message is enabled
**Describe the bug**
After opening a buggy h5 file, app shows when closing:

```
HDF5: infinite loop closing library
      L,G_top,T_top,P,P,Z,FD,VL,VL,PL,E,SL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL
Debug Error!
```

This is fine. In `H5_term_library` (in `H5.c`), there is:

```
    if (pending) {
        /* Only display the error message if the user is interested in them. */
        if (func) {
            fprintf(stderr, "HDF5: infinite loop closing library\n");
            fprintf(stderr, "      %s\n", loop);
#ifndef NDEBUG
            abort();
#endif
        }
    }
```

But if I disabled output with `H5Eset_auto2`, the library doesn't abort anymore.

Is it fine that `abort()` is done only if display is enabled ? Shouldn't abort must be after `if (func)` instead of inside ?

If you confirm that it's a bug, I can easily write a fix.

**Expected behavior**
`abort` should always occurs.

 - HDF5 version : 1.14.6
 - OS and version: Windows 25H2
 - Compiler and version: VS 2022
 - Build system (e.g. CMake version) and generator (e.g. XCode, Ninja)

mattjala:
When the library detects that it cannot make progress closing itself (an "infinite loop closing library"), it now aborts regardless of whether automatic error message display has been enabled via `H5Eset_auto2()`. Previously the abort was skipped whenever error output was turned off, so an application that had disabled error reporting would continue running in this unrecoverable state instead of terminating.

Fixes #6531


jhendersonHDF:
Introduced in https://github.com/HDFGroup/hdf5/commit/339a81a766d4d37e6535f0c8fad7c853b9a849e9, but no particular rationale given. I assume the intention is that if the user asked to disable error reporting, then they don't want any error handling path like this to be triggered. The library is still going to terminate either way, so I don't think skipping the `abort()` carries any danger whatsoever. It's also arguably rude for a library to ever call `abort()`, even if it's during library termination, since it doesn't give any chance for cleanup and can have any number of effects on downstream code.

lrknox (dismissed):

lrknox (dismissed):

lrknox (changes requested):
CHANGELOG.md has 2.2.0 release note entries that have been removed in develop;  they should also be removed here.  Removing abort has been discussed, was removing it settled?

mattjala:
> CHANGELOG.md has 2.2.0 release note entries that have been removed in develop; they should also be removed here. Removing abort has been discussed, was removing it settled?

Yes, we decided that removing the abort makes the most sense.

jhendersonHDF (approved):

lrknox (approved):

PR #6535, opened 2026-07-17
Install the hdf5.tag file to the html directory

### LINKED ISSUE #5593: Create and publish a Doxygen tag file
Doxygen documentation can be interlinked as described in [Linking to external documentation](https://www.doxygen.nl/manual/external.html). By creating (via `GENERATE_TAGFILE`) and deploying a tag file to [https://support.hdfgroup.org/documentation/hdf5/latest/](https://support.hdfgroup.org/documentation/hdf5/latest/), we support external documentation projects that want to reference HDF5 library documentation.

mkitti:
Moves the hdf5.tag file to ${HDF5_INSTALL_DOC_DIR}/html which defaults to ${PREFIX}/share/html. The intention is to deploy the hdf5.tag file to https://support.hdfgroup.org/documentation/hdf5/latest/hdf5.tag

This may fix https://github.com/HDFGroup/hdf5/issues/5593

### DIFF
--- a/docs/doxygen/CMakeLists.txt
+++ b/docs/doxygen/CMakeLists.txt
@@ -71,7 +71,7 @@ if (DOXYGEN_FOUND)
   )
   install (
       FILES ${HDF5_BINARY_DIR}/hdf5.tag
-      DESTINATION ${HDF5_INSTALL_DOC_DIR}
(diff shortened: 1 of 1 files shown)


gheber (approved):
👍

lrknox (approved):

Issue #6391, opened 2026-05-01
Format Specification: Missing shared message encoding in attribute message


vchoi-hdfgroup:
**Describe the bug**
In section IV.A.2.m. The Attribute Message, version 2 and version 3:
When the "flags" field indicates that either datatype or dataspace is shared, the "datatype" or "dataspace"  field will contain a “shared message” encoding instead of the datatype or dataspace encoding. 
The layout of the "shared message" encoding should be added and described.

vchoi-hdfgroup:
Fix is done via PR #6538

PR #6538, opened 2026-07-19
Fixes for issue #6391:

vchoi-hdfgroup:
- The layout of the shared message encoding was actually decribed at the beginning of the section IV.A.2. For clarification, the section was split into two with one section purely for the shared message encoding and another section for the catalog of message types.
- Cross-linked shareable messages to the shared message encoding section.
- Misc. cleanup


gheber (approved):
Looking good.

PR #6539, opened 2026-07-20
Enable error printing for internal data filters

### LINKED ISSUE #6194: Errors in data filters are not printed
When a data filter encounters an error, any error messages posted by that filter are no longer printed to stderr. It still returns an error to the library, and the library prints its internal error messages, but there is no information about what caused the failure in the filter. This can be seen for instance if the flether32 checksum filter is enabled and a checksum doesn't match. This can be seen in both built in filters and external filters that use the default error stack.

This is happening because, in preparation for thread safety work, the library saves the error stack before making a call to a plugin (potentially an application callback), and then restores the stack afterwards, therefore overwriting any changes to the error stack that happened during the call to the plugin.

fortnern:
Prevent pausing the error stack when the filter is mandatory and there is no callback, and prevent saving/restoring the error stack when calling a library internal filter. This currently has two limitations:
1. External filters that push to the default error stack will need to print the error stack before they return
2. If a filter failure callback is used, error messages issued by the filter will not be printed, even if the callback returns failure

I will continue investigating these cases.

Fixes #6194


test/dsets.c
```diff
@@ -19055,6 +19111,246 @@ test_readonly_chunk_vlen_fill(hid_t fapl, bool chunk_cache)
...
+    return SUCCEED;
+
+error:
+    return FAIL;
+} /* end filter_error_msg_stack_handler() */
+
+/*-------------------------------------------------------------------------
+ * Function:    test_filter_error_msg
+ *
+ * Purpose:     Tests that the library properly fails when the buffer size
```
mattjala: Looks like this comment is a leftover copy from `test_bad_buf_size`

test/dsets.c
```diff
@@ -19055,6 +19111,246 @@ test_readonly_chunk_vlen_fill(hid_t fapl, bool chunk_cache)
...
+    H5Z_FILTER_RESERVED - 1, /* Filter id number */
+    1, 1,                    /* Encoding and decoding enabled */
+    "error_msg",             /* Filter name for debugging */
+    NULL,                    /* The "can apply" callback */
+    NULL,                    /* The "set local" callback */
+    filter_error_msg,        /* The actual filter function */
+}};
+static const char *filter_error_msg_msg =
+    "Hello I am a unique error message written for the test_filter_error_msg test in dsets.c";
+bool enable_error_msg;
```
mattjala: Since this is only used in this file, it should probably be declared `static`. Same goes for `error_msg_found`, `dsets_h5zregister_direct`, and `dsets_h5zunregister_direct`.

src/H5Z.c
```diff
@@ -800,12 +800,17 @@ H5Z__prelude_callback(const H5O_pline_t *pline, hid_t dcpl_id, hid_t type_id, hi
...
 
                         /* Prepare & restore library for user callback */
-                        H5_BEFORE_USER_CB(FAIL)
-                            {
-                                /* Make callback to filter's "can apply" function */
-                                status = (fclass->can_apply)(dcpl_id, type_id, space_id);
-                            }
-                        H5_AFTER_USER_CB(FAIL)
+                        /* Don't prepare for a user callback if this is an internal filter */
+                        if (fclass->id < H5Z_FILTER_RESERVED)
```
mattjala, line 803: Is there a reason that the `H5Z_PRELUDE_SET_LOCAL` case below this is excluded from a similar check?

fortnern:
Thanks Matt, I've addressed the comments.

fortnern:
ROS3 and nvhpc failures are unrelated to this PR

lrknox (approved):

mattjala (approved):

PR #6542, opened 2026-07-20
Publish hdf5.tag alongside the Doxygen HTML documentation

mkitti:
## Summary

Follow-up to #6535. That PR moved the **install** destination of `hdf5.tag` to `${HDF5_INSTALL_DOC_DIR}/html`, intending to deploy it to https://support.hdfgroup.org/documentation/hdf5/latest/hdf5.tag. However, the S3-published web docs do not come from the install tree — they come from the `docs-doxygen` GitHub Actions artifact, which is uploaded from the **build tree** (`hdf5lib_docs/html`, see `ctest.yml`). The tag file was generated at the build *root* (`${HDF5_BINARY_DIR}/hdf5.tag`), a sibling of `hdf5lib_docs/`, so it never made it into the artifact or any downstream publish.

This PR generates the tag file **inside** `hdf5lib_docs/html/` so it is captured by the `docs-doxygen` artifact and flows to every destination that consumes it:

- `daily-schedule.yml` → `documentation/hdf5/latest/hdf5.tag` (the primary target URL)
- `release-files.yml` → included in `*.doxygen.zip`
- `publish-release.yml` → `…/documentation/doxygen/hdf5.tag`
- `publish-branch.yml` → included wherever the html dir is synced

## Changes

- **`docs/doxygen/CMakeLists.txt`**: point `DOXYGEN_TAG_FILE` at `${HDF5_BINARY_DIR}/hdf5lib_docs/html/hdf5.tag`. The existing `DIRECTORY` install of `hdf5lib_docs/html` now installs the tag file to `${HDF5_INSTALL_DOC_DIR}/html/hdf5.tag` (preserving #6535's intent), so the redundant standalone `install(FILES … hdf5.tag …)` is removed — it would otherwise reference a file no longer present at the build root.
- **`.github/workflows/ctest.yml`**: verify `hdf5.tag` exists before uploading the `docs-doxygen` artifact.
- **`.github/workflows/publish-release.yml`**: verify `hdf5.tag` is present after extracting the documentation zip.

## Testing

Built docs locally with `-DHDF5_BUILD_DOC=ON` (doxygen 1.9.8):

- `hdf5lib_docs/html/hdf5.tag` is generated (and no longer at the build root)
- `cmake --install --component hdfdocuments` places it at `share/doc/HDF5/html/hdf5.tag` with no missing-file error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6543, opened 2026-07-20
Harden S3 publish workflows against zizmor findings

mkitti:
## Summary

Addresses GitHub Actions static-analysis (zizmor / CodeQL "GitHub Advanced Security") findings in the S3 publishing workflows. These are the two workflows that publish release artifacts and documentation to S3, and they carried the bulk of the `template-injection` findings.

## Changes

**`.github/workflows/publish-release.yml`** (41 `template-injection` + 1 `artipacked`)
- Hoist every `inputs.*` / `secrets.*` / `vars.*` value used inside `run:` blocks into job-level `env:` variables (`USE_TAG`, `FILE_NAME`, `TARGET_DIR`, `DRY_RUN`, `S3_BUCKET`, `TARGET_PATH`) and reference plain shell variables (`$USE_TAG`, …) instead of interpolating `${{ ... }}` directly into the script body. This removes the code-injection vector where a value is expanded into the script before the shell runs.
- Add `persist-credentials: false` to the `actions/checkout` step (`artipacked`). This workflow does not push to git; it only reads sources/scripts and uses separate AWS credentials.

**`.github/workflows/publish-branch.yml`** (1 `artipacked`)
- Add `persist-credentials: false` to the `actions/checkout` step. Its S3 sync step already used `env:` variables.

## No behavioral change

The same values are used — only sourced from the shell environment instead of direct `${{ }}` expansion. `if:` conditions and action `with:` inputs are left as-is (those are not shell-injection contexts).

## Verification

`zizmor` (v1.25.2, the pinned version from `zizmor.yml`) reports **no** findings on either file after the change.

## Scope note

The remaining zizmor findings across other workflows (`maven-*`, `java-*`, additional `artipacked`, and `excessive-permissions`) are intentionally **not** included here. Those require per-workflow analysis — e.g. `persist-credentials: false` would break checkouts used for `git push` / gh-pages deploys, and minimal `permissions:` blocks need to be verified against each job's needs — and are better handled as separate, focused follow-ups.

### DIFF
--- a/.github/workflows/publish-branch.yml
+++ b/.github/workflows/publish-branch.yml
@@ -26,6 +26,7 @@ jobs:
           with:
             fetch-depth: 0
             ref: '${{ github.head_ref || github.ref_name }}'
+            persist-credentials: false
 
         - name: List files for the space
           run: |
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -28,21 +28,33 @@ jobs:
   publish-tag:
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    
+
+    # Expose the templated inputs/secrets/vars as environment variables so the
+    # run steps below reference plain shell variables ($USE_TAG, ...) instead of
+    # interpolating ${{ ... }} directly into the script body, which avoids
+    # template-injection (see zizmor/CodeQL findings).
+    env:
+      USE_TAG: ${{ inputs.use_tag }}
+      FILE_NAME: ${{ inputs.file_name }}
+      TARGET_DIR: ${{ inputs.target_dir }}
+      DRY_RUN: ${{ inputs.dry_run }}
+      S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+      TARGET_PATH: ${{ vars.TARGET_PATH }}
+
     steps:
       - name: Validate Inputs
         run: |
           set -euo pipefail
           echo "Validating inputs..."
-          if [[ ! "${{ inputs.use_tag }}" =~ ^hdf5[_-][0-9]+[._][0-9]+[._][0-9]+([._][0-9]+)?(-.*)?$ ]]; then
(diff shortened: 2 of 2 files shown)


mkitti:
### Deconfliction with #6542 (hdf5.tag)

This PR and #6542 both modify the **`Process documentation`** step of `.github/workflows/publish-release.yml`, so they will conflict:

- **#6542** adds a step-level `env: FILE_NAME` and an `hdf5.tag` presence check.
- **This PR** rewrites the file to use a job-level `env:` block (which already defines `FILE_NAME`) and does not add the tag check.

The clean merged result keeps this PR's job-level `env:` (so #6542's step-level `env: FILE_NAME` becomes redundant and is dropped) and adds the `hdf5.tag` check inside the already-hardened step. Regardless of merge order, resolve the `Process documentation` step to:

```diff
       - name: Process documentation
         run: |
           set -euo pipefail
           DOC_FILE="HDF5/${FILE_NAME}.doxygen.zip"
           if [ -f "$DOC_FILE" ]; then
             echo "📚 Processing documentation..."
             unzip -q "$DOC_FILE"
             if [ -d "${FILE_NAME}.doxygen" ]; then
               echo "✅ Documentation extracted successfully"
             else
               echo "❌ Documentation extraction failed"
               exit 1
             fi
+            # The doxygen tag file is published alongside the HTML docs
+            if [ -f "${FILE_NAME}.doxygen/hdf5.tag" ]; then
+              echo "✅ Found: hdf5.tag"
+            else
+              echo "❌ hdf5.tag not found in documentation"
+              exit 1
+            fi
           else
             echo "⚠️  Documentation file not found, skipping..."
           fi
```

Note: the added check references `${FILE_NAME}` from the job-level `env:` introduced here, so **no step-level `env:` is needed** — that is the only substantive difference from #6542's version of the step. Everything else in #6542 (the `docs/doxygen/CMakeLists.txt` change and the `ctest.yml` guard) is untouched by this PR and merges without conflict.

hyoklee (approved):

PR #6545, opened 2026-07-20
Harden workflow checkout credential hygiene + permissions

mkitti:
Addresses `zizmor` (v1.25.2) static-analysis findings in a set of workflow files. **No behavioral change.**

## artipacked — `persist-credentials: false` added to `actions/checkout`
None of these jobs push to git (verified: no `git push`/`git commit`/`gh-pages`/`peaceiris`/`auto-commit`/`EndBug`), so dropping the persisted checkout credential is safe.

| File | Checkouts fixed |
|------|-----------------|
| codeql.yml | 2 |
| test-binary-installation.yml | 2 |
| clang-format-check.yml | 1 |
| freebsd.yml | 1 |
| h5py.yml | 1 |
| openbsd.yml | 1 |
| signed-plugins.yml | 1 |
| update-progress.yml | 1 |

## excessive-permissions — call-workflows.yml (2 findings)
The top-level `permissions` previously granted `packages: write` and `pull-requests: write` to every job. Restricted the top level to `contents: read`, and moved `packages: write` / `pull-requests: write` to the two jobs (`call-maven-staging`, `call-maven-ffm-staging`) that call the `maven-staging.yml` reusable workflow, which is the only called workflow declaring those needs. All other called workflows already declare `contents: read`.

## Verification
- `zizmor` re-run over all nine files: **No findings to report** (0 findings).
- All nine files pass `yaml.safe_load`.


mkitti:
### Note: interaction with #6546 (Maven/Java hardening)

This PR re-grants `packages: write` + `pull-requests: write` on the two jobs that call `maven-staging.yml` (`call-maven-staging`, `call-maven-ffm-staging`), because on `develop` today `maven-staging.yml` declares those permissions at its top level, and a caller must grant what the reusable workflow declares.

**#6546** reduces `maven-staging.yml` to only `contents: read` (it is a dry-run staging test — no `mvn deploy` or package publish). Once both this PR and #6546 land, the `packages: write` / `pull-requests: write` grants added here become **unused** — a caller granting more than the reusable workflow declares is simply capped to what the callee actually uses, so there is no correctness impact and no new zizmor finding, but it is more than strictly needed.

Both PRs are correct and zizmor-clean independently. The fully least-privilege end state (dropping these two job-level grants to `contents: read`) is intended as a small follow-up once #6546's reduced `maven-staging.yml` permissions are on `develop`.

mkitti:
Checked both Copilot comments:
- Trailing whitespace in `update-progress.yml`: that blank separator line is pre-existing (not touched by this PR) and the same pattern repeats ~18 times throughout the file as an established (if unconventional) formatting style. Left as-is to avoid an inconsistent partial cleanup outside this PR's scope.
- YAML indentation of the new `permissions:` blocks in `call-workflows.yml`: this actually matches the file's existing 4-space-per-level convention exactly (job name at 4, fields at 8, sub-fields at 12) — same depth as the neighboring `with:` blocks. No change needed.

Separately, the one real CI failure (`nvhpc Release / Run Tests`) is a UCX/InfiniBand driver error on the runner (`uct_iface_open(ud_verbs/mana_0:1) failed: Address not valid` → `MPI_INIT` failure), unrelated to this PR's workflow-permission changes. Looks like transient runner infrastructure flakiness — worth a re-run rather than a code change.

lrknox (dismissed):

lrknox (approved):

PR #6546, opened 2026-07-20
Harden Maven/Java workflows against zizmor findings

mkitti:
Hardens the Maven/Java-themed GitHub Actions workflows against all `zizmor` 1.25.2 static-analysis findings. No behavioral change: the same values and logic are preserved, just sourced via `env:` variables and explicit permissions/secrets.

## Findings fixed (per file)

| File | template-injection | artipacked | excessive-permissions | secrets-inherit |
|------|:---:|:---:|:---:|:---:|
| `maven-staging.yml` | 37 | 4 | 2 | - |
| `maven-build-test.yml` | 11 | 3 | 1 | - |
| `test-maven-packages.yml` | 6 | 2 | 0 | - |
| `java-implementation-test.yml` | 0 | 2 | 5 | - |
| `test-maven-deployment.yml` | 1 | 1 | 1 | - |
| `release.yml` | 0 | 0 | 0 | 1 |

Total: 77 findings resolved. Final `zizmor` run reports **zero findings** across all six files.

## What changed

- **template-injection**: attacker-controllable `${{ }}` expressions used inside `run:` bodies (`matrix.*`, `runner.workspace`, `github.actor`, `github.base_ref`, `steps.*.outputs.*`, `needs.*.outputs.*`) moved into step-level `env:` blocks and referenced as `$NAME` (bash) or `$env:NAME` (pwsh). Non-run-body expressions (`if:`, `with:`, `uses:`, `env:` values) left untouched.
- **artipacked**: `persist-credentials: false` added to every `actions/checkout` step. None of these workflows push to git (verified: no `git push`/`git commit`/`gh-pages`/`peaceiris`/`auto-commit`/`EndBug`), so all checkouts were safe to harden.
- **excessive-permissions**: added explicit top-level `permissions: contents: read`. Overly broad workflow-level `packages: write` / `pull-requests: write` reduced; minimal `packages: read`/`packages: write` granted only to the specific jobs that actually consume/deploy packages. `maven-staging.yml` performs only a dry-run deployment (no real `mvn deploy`, PR writes, or package API calls), so its top level drops to `contents: read`.
- **secrets-inherit** (`release.yml`): the call to `test-maven-packages.yml` used `secrets: inherit`, but that reusable workflow declares no `workflow_call` secrets, so the line was removed entirely.


mkitti:
### Note: interaction with #6545 (checkout-hygiene + permissions hardening)

This PR reduces `maven-staging.yml` from top-level `packages: write` + `pull-requests: write` down to `contents: read` (the staging deployment is a dry run — it writes a `settings.xml` with literal Maven `${env.*}` placeholders and never performs a real deploy or package API call; artifact handoff uses `actions/upload-artifact`, which needs no extra permission).

Separately, **#6545** re-grants `packages: write` / `pull-requests: write` on `call-workflows.yml`'s two `maven-staging.yml` caller jobs (required while `develop`'s `maven-staging.yml` still declares them). After both PRs land, those caller grants are harmless but unused; a small follow-up can drop them to `contents: read` for full least-privilege. Flagging so the two are reviewed with the interaction in mind.

mkitti:
Checked the Copilot comment about `actions/checkout@9c091bb2...` being annotated `# v7.0.0`. Verified against the actual `actions/checkout` `v7.0.0` git tag — the SHA does correspond to v7.0.0, and this line is pre-existing (unchanged by this PR, and used the same way in ~a dozen other workflow files in the repo). No fix needed; the review comment was a false positive.

lrknox (dismissed):

lrknox (approved):

PR #6547, opened 2026-07-21
Expose chunk-level HDF5 APIs (H5Dchunk_iter, direct chunk I/O) to Java (JNI + FFM)

mkitti:
## Summary

Exposes HDF5's chunk-level APIs to both Java binding trees (`java/src-jni` JNI and `java/hdf` FFM/Panama), working toward full C API parity in the Java bindings:

- `H5Dchunk_iter`, `H5Dget_num_chunks`, `H5Dget_chunk_info` (by index)
- `H5Dget_chunk_info_by_coord`, `H5Dget_chunk_storage_size`, `H5Dget_chunk_index_type`, `H5Dwrite_chunk`, `H5Dread_chunk` (direct raw chunk I/O, bypassing the filter pipeline/hyperslab), plus the five missing `H5D_CHUNK_IDX_*` constants
- `H5Dchunk_iter_all`, a bulk convenience form of `H5Dchunk_iter` returning every chunk's offset/filter mask/address/size as a single `H5D_chunk_info_t` instead of requiring a callback

Also includes a performance fix: `H5D_chunk_iter_cb` (JNI) was redundantly doing `AttachCurrentThread`/`GetMethodID`/`NewLongArray` on *every* chunk instead of once per `H5Dchunk_iter` call; resolving these once cut per-chunk callback overhead by ~2.8x at large chunk counts.

## Notable design notes

- `H5Dread_chunk` guards against a real silent-failure hazard in the underlying `H5D__chunk_direct_read()`: if the caller's buffer doesn't match the true on-disk chunk size, the C function still returns success but leaves the buffer untouched. Both implementations check this internally and throw `IllegalArgumentException` on mismatch.
- `H5Dchunk_iter_all`'s JNI implementation accumulates chunk info in a callback that is itself pure C (zero JVM crossings per chunk), converting to Java arrays only once at the end — measured ~2x faster than streaming `H5Dchunk_iter` at ~4000 chunks. The FFM implementation cannot replicate this: FFM's callback is necessarily an upcall stub that crosses back into the JVM on every invocation, so `H5Dchunk_iter_all` is offered on FFM purely for a simpler call site (no callback to author), not as a performance win — documented in the method's javadoc.
- A quick benchmark (`TestH5DChunkIterPerf`, both trees) comparing `H5Dchunk_iter`/`H5Dchunk_iter_all` against a by-index `H5Dget_chunk_info` loop confirms the same scaling blowup reported in [HDF5.jl#1031](https://github.com/JuliaIO/HDF5.jl/pull/1031#issuecomment-1407749686): the by-index loop is ~150-1400x slower at ~4000 chunks depending on which chunk-enumeration strategy it's compared against.

## Test plan

- [x] Full `TestH5D` suite passes in both JNI (40 tests) and FFM (39 tests) builds — the count differs because `testH5Dvlen_string_buffer` carries `@Test` + `@Ignore` in the FFM tree (so JUnit reports it as skipped) but only `@Ignore` (no `@Test`) in the JNI tree (so it isn't discovered as a test at all); both trees exercise the same 39 active tests plus this one long-pre-existing disabled case
- [x] Each new test verified individually in isolation (JUnit `Request.method`), not just as part of the full suite
- [x] `H5Dchunk_iter_all`'s bulk result cross-checked against per-chunk `H5Dget_chunk_info()` results, matched by offset (the two APIs aren't guaranteed to visit chunks in the same order)
- [x] `H5Dread_chunk`'s buffer-size-mismatch guard exercised directly (expects `IllegalArgumentException`)
- [x] Benchmarked against a pure-C baseline (temporary, not included) to confirm the by-index blowup is a C-library property, not a binding-layer one, and that the streaming-callback overhead fix and bulk variant close most of the JNI-vs-C gap


mkitti:
## `H5Dchunk_iter_all`: what it is, why it exists, and JNI vs FFM performance

### What it is

`H5Dchunk_iter_all` is a convenience method (added in both the JNI and FFM
trees) that wraps `H5Dchunk_iter` to enumerate **every** chunk of a dataset
in one call and return the results as plain Java arrays, instead of requiring
the caller to write a callback:

```java
H5D_chunk_info_t info = H5.H5Dchunk_iter_all(dataset_id, dxpl_id);
// info.getNumChunks(), info.getOffset(chunkIndex, dim), info.filterMask, info.addr, info.size
```

### Why it's necessary

Before this PR, there were two ways to enumerate a dataset's chunks from
Java:

1. **`H5Dget_chunk_info` in a loop over `0..H5Dget_num_chunks-1`** — simple
   array-based ergonomics, but algorithmically poor: each call does an
   independent chunk-index lookup, so enumerating all chunks this way costs
   **O(n²)** (or worse — see the `indx_ms` column below, which blows up
   super-linearly with a v1 B-tree index). This is exactly the pathology
   [reported for HDF5.jl](https://github.com/JuliaIO/HDF5.jl/pull/1031#issuecomment-1407749686)
   that motivated this PR's benchmark in the first place.
2. **`H5Dchunk_iter` directly** — the C library streams every chunk with a
   single index traversal (O(n) total), but only via a callback, which is
   awkward for the common "just give me all the chunks" case and (for FFM
   specifically) has real per-chunk crossing cost — see below.

`H5Dchunk_iter_all` closes that gap: array-based ergonomics *and* the
library's linear-time traversal, by driving `H5Dchunk_iter` internally and
accumulating into arrays for the caller.

### Performance: methodology

Benchmarked with `TestH5DChunkIterPerf.java` (both trees): a chunked
2-D dataset with a 16×16 chunk shape, swept over `32×32` through
`1024×1024` (4 to 4096 chunks), default library version bounds (v1 B-tree
chunk index — see the `java/CLAUDE.md` note on chunk index defaults), one
untimed warm-up dataset run first to absorb JIT/class-loading/native-linkage
cost, then each approach timed via `System.nanoTime()`. Two independent runs
each; numbers below are the average (they agreed within noise on both
trees). All times in milliseconds.

### Results

| chunks | JNI `iter_ms`<br>(streaming) | JNI `all_ms`<br>(bulk) | FFM `iter_ms`<br>(streaming) | FFM `all_ms`<br>(bulk) | `indx_ms`<br>(by-index loop, either tree)† |
|---:|---:|---:|---:|---:|---:|
| 4    | 0.012 | 0.011 | 0.377 | 0.543  | ~0.02   |
| 16   | 0.022 | 0.016 | 0.488 | 0.728  | ~0.05   |
| 64   | 0.036 | 0.019 | 0.531 | 1.470  | ~0.28   |
| 256  | 0.130 | 0.057 | 1.829 | 4.426  | ~7      |
| 1024 | 0.308 | 0.155 | 2.276 | 8.166  | ~50     |
| 4096 | 1.168 | 0.535 | 4.089 | 11.819 | ~700    |

† `indx_ms` (looping `H5Dget_chunk_info` by index) is shown once since it's
dominated by the C-library algorithm, not the binding layer — both trees
match to within normal run-to-run noise. It grows super-linearly and is
already ~130-600× slower than either chunk-iterate approach by 4096 chunks,
confirming this is a library/index-structure effect, not a JNI/FFM artifact.

### Interpretation

- **JNI: `H5Dchunk_iter_all` is a genuine speedup**, ~1.9-2.3× faster than
  streaming `H5Dchunk_iter` from a few hundred chunks up. This is real and
  architectural: JNI's callback trampoline is plain C, so the bulk variant's
  callback (`H5D_chunk_iter_all_cb`) does zero JNI/JVM crossings per chunk —
  just a native accumulate — with all `n` chunks converted to Java arrays
  once, at the end, instead of once per chunk.
- **FFM: `H5Dchunk_iter_all` is *slower* than streaming `H5Dchunk_iter`**,
  by about 1.4-3.6× depending on size — the opposite direction from JNI.
  This is also architectural, not a missed optimization: in FFM, the
  callback handed to the C library **must** be a Java upcall stub (there is
  no way to give `H5Dchunk_iter` a pure-native callback via
  `java.lang.foreign` alone), so it pays the same per-chunk JVM-crossing
  cost as streaming `H5Dchunk_iter`, *plus* extra per-chunk work: the
  mandatory `.reinterpret()` on the zero-length `offset` `MemorySegment`
  jextract hands the upcall, plus a `MemorySegment.copy` and three
  `setAtIndex` writes to stash each chunk's data before the final bulk
  conversion. There's no callback-side win available to claim back.
- **FFM is slower than JNI in absolute terms** for both variants, as
  expected for a downcall/upcall-per-chunk cost model vs. plain JNI, with
  the gap most pronounced for the bulk variant (~22× at 4096 chunks) since
  that's exactly where FFM's structural overhead compounds and JNI's does
  not.
- Despite FFM's `H5Dchunk_iter_all` not being a perf win there, it's kept in
  both trees for API symmetry/ergonomics — it's documented in the FFM
  javadoc and `java/CLAUDE.md` as ergonomics-only on that side, not a
  performance claim.
- The **actually load-bearing result for the original HDF5.jl-style
  concern** is chunk-iterate (either variant, either tree) vs. by-index
  looping, not JNI vs. FFM — that gap (700ms vs ~1ms at 4096 chunks) is the
  one that matters for any caller enumerating a large number of chunks.

Implementation: `24e9353` (JNI), `96d25ed` (FFM). Benchmark source:
`java/src-jni/test/TestH5DChunkIterPerf.java` /
`java/test/TestH5DChunkIterPerf.java`.

mkitti:
I do not think the one CI failure is related to my changes.

mkitti:
After thinking about `H5Dchunk_iter_all` I wonder if this should be implemented in such a way that other bindings could use it.

mkitti:
@jhendersonHDF any thoughts about this? I am thinking of splititng this this apart and moving `H5Dchunk_iter_all` to its own pull rqeuest.

java/hdf/hdf5lib/HDF5Constants.java
```diff
@@ -127,6 +127,16 @@ public class HDF5Constants {
     /** */
     public static final int H5D_CHUNK_IDX_BTREE = H5D_CHUNK_IDX_BTREE();
     /** */
+    public static final int H5D_CHUNK_IDX_SINGLE = H5D_CHUNK_IDX_SINGLE();
```
jhendersonHDF, line 130: I would consider dropping these unless they're necessary. They're really only useful for the `H5Dget_chunk_index_type()` function and that is meant to be an internal API (if one can even call an API internal). It's not really a function meant to be called or used for anything.

java/src-jni/test/TestH5DChunkIterPerf.java
jhendersonHDF, line 1: This seems useful enough, but since there isn't anything actually being tested for correctness here it should probably be part of some performance testing framework that reports to a dashboard rather than the main Java testing. Otherwise, it's likely that any change in results will just get missed.

jhendersonHDF (commented):
Hi @mkitti,

I haven't really had a chance to look over this and will be out for the next two weeks, but I'll leave a couple initial comments below.

As far as `H5Dchunk_iter_all()`, it's something that might be easier to review in a separate PR; the wrappers over existing C functions are easy enough to review and would make a fairly small PR. My thought on `H5Dchunk_iter_all()` is that it's something that might become unnecessary eventually. I wouldn't mind having it in the Java bindings, but we've had requests in the past to consider switching over to an iterator based approach for functions that currently take callbacks. If we do so, that would likely get the performance benefits of `H5Dchunk_iter_all()`, avoiding the callback overhead, and make it fairly straightforward to implement something close to an `H5Dchunk_iter_all()` but potentially in a more flexible way. For large datasets, I'm thinking something like batched blocks of retrievals instead of allocating all the memory upfront.

PR #6550, opened 2026-07-22
Add Fortran ABI check and auto-detect the release baseline

brtnfld:
## Summary
- Adds a Fortran ABI compatibility check to `abi-report.yml`, mirroring the existing C/HL/C++ `abi-dumper` + `abi-compliance-checker` steps for `libhdf5_fortran.so`, plus a supplementary `nm`-based exported-symbol diff (Fortran name mangling doesn't encode argument lists, so a symbol-name diff catches add/remove even when DWARF-based signature diffing doesn't apply cleanly).
- Fixes `file_ref` (the "previous release" ABI baseline) being hardcoded to `2.0.0` in both `daily-build.yml` and `release.yml` — it had gone stale across two releases (`2.1.0`, `2.1.1`) without being bumped. `abi-report.yml` now auto-detects the latest published `HDFGroup/hdf5` release via the GitHub API when `file_ref` isn't explicitly given, so there's nothing to remember to update after future releases. `file_ref` remains available as an optional override.
- Documents an expected false-positive pattern in the Fortran report: `abi-dumper` misreads gfortran's DWARF encoding of assumed-shape array descriptors as fixed array bounds, producing bulk "Low" severity noise on the KIND/RANK-generated buffer procedures in `H5_gen.F90` (e.g. `integer(kind=16)[189525]` → `[190309]`) that isn't a real interface change. Real signal shows up in Added/Removed Symbols and Medium/High severity findings.

## Validation
Ran the full `daily-build.yml` pipeline end-to-end on a fork (workflow_dispatch) with `libhdf5_fortran.so` from `develop` vs the `2.0.0` baseline (before the auto-detect fix landed) and inspected the real `abi-reports` artifact:
- C/HL/C++ reports: clean (0 problems, only new symbols listed).
- Fortran report: 1,751 symbols / 36 types compared, 0 removed, 7 added (correctly identified real new API: `h5dflush_f`, `h5fstart_swmr_write_f`, `h5pget/set_append_flush_f`, 3 new float-kind constants), 1 genuine Medium finding (a module-level data array's size legitimately grew to hold the new float-kind constants), and 250 Low findings — all confirmed false positives, all confined to `H5_gen.F90` generic array-buffer procedures as described above.
- Separately verified the auto-detect API call live: it correctly resolves to `2.1.1`, the actual latest release.

### DIFF
--- a/.github/workflows/abi-report.yml
+++ b/.github/workflows/abi-report.yml
@@ -147,6 +147,25 @@ jobs:
           abi-compliance-checker -l ${{ inputs.file_base }}_hl_cpp -old ABI-6.dump -new ABI-7.dump
         continue-on-error: true
 
+      - name: Run fortran ABI report
+        run: |
+          abi-dumper ${{ steps.set-hdf5lib-refname.outputs.HDF5R_ROOT }}/lib/libhdf5_fortran.so -o ABI-8.dump
+          abi-dumper ${{ steps.set-hdf5lib-name.outputs.HDF5_ROOT }}/lib/libhdf5_fortran.so -o ABI-9.dump
+          abi-compliance-checker -l ${{ inputs.file_base }}_fortran -old ABI-8.dump -new ABI-9.dump
+        continue-on-error: true
+
+      - name: Run fortran exported symbol diff
+        run: |
+          # gfortran mangled names don't encode argument lists, so abi-compliance-checker
+          # (DWARF-based) is the primary signal above; this is a cheap supplementary check
+          # that still catches added/removed exported symbols even if debug info is stripped.
+          nm -D --defined-only ${{ steps.set-hdf5lib-refname.outputs.HDF5R_ROOT }}/lib/libhdf5_fortran.so | awk '{print $3}' | sort > old_fortran_syms.txt
+          nm -D --defined-only ${{ steps.set-hdf5lib-name.outputs.HDF5_ROOT }}/lib/libhdf5_fortran.so | awk '{print $3}' | sort > new_fortran_syms.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/abi-report.yml
```diff
@@ -147,6 +163,35 @@ jobs:
...
 
+      # NOTE ON READING THE FORTRAN REPORT: expect a large number of "Low" severity
+      # "Problems with Symbols" findings on the generic KIND/RANK-generated buffer
+      # procedures in H5_gen.F90 (h5aread_*, h5awrite_*, h5dread_*, h5dwrite_*), e.g.
+      # "Type of Nth parameter buf has been changed from integer(kind=K)[X] to
+      # integer(kind=K)[Y]". These are false positives: abi-dumper misreads gfortran's
+      # DWARF encoding of assumed-shape array descriptors as if the bound were a fixed
+      # compile-time array size, so the reported X/Y values are debug-info noise, not a
+      # real interface change. Real signal shows up in Added/Removed Symbols and in
+      # Medium/High severity findings (e.g. a module-level data size actually changing) -
```
hyoklee, line 174: `e.g.` -> `e.g.,`

brtnfld: Could we save comments for substantive issues (content, accuracy, argument) rather than micro-style grammar points like comma placement after "e.g."? Those are quick to normalize at the copyediting stage if we decide to do so, so flagging each instance individually ends up costing more time than it saves.

lrknox (approved):
I tried the changes in a branch on my fork.  They automatically picked up 2.1.1 as the last version and added a Fortran compatibility report.

PR #6551, opened 2026-07-22
Update so numbers for HDF5 2.2.0 and set version to 2.2.0-rc1.

lrknox:


lrknox:
This needs to go to hdf5_2_2_0 branch.

PR #6553, opened 2026-07-23
Fix h5ls issue with quoting when displaying integer data as ASCII characters

jhendersonHDF:
When using the `-r` option to have `h5dump` display 1-byte integer datasets and attributes as ASCII characters, the data display logic includes a section that deals with including a closing double-quote character for the data values. This logic wasn't carried over to `h5ls` for its similar `-s` option, causing it to drop the closing double-quote:

```C
#include "hdf5.h"

int
main(int argc, char **argv)
{
    hid_t file_id;
    hid_t dset_id;
    hid_t type_id;
    hid_t space_id;
    hid_t attr_id;
    hsize_t dims[] = { 81 };
    /* 81-byte string, including NUL terminator */
    char *data = "abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz012345678901234567";

    file_id = H5Fcreate("testchar.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);

    type_id = H5Tcopy(H5T_NATIVE_UCHAR);

    space_id = H5Screate_simple(1, dims, NULL);

    dset_id = H5Dcreate2(file_id, "dset", type_id, space_id,
                         H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);

    H5Dwrite(dset_id, type_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, data);

    attr_id = H5Acreate2(dset_id, "attr", type_id, space_id,
                         H5P_DEFAULT, H5P_DEFAULT);

    H5Awrite(attr_id, type_id, data);

    H5Sclose(space_id);
    H5Tclose(type_id);
    H5Aclose(attr_id);
    H5Dclose(dset_id);
    H5Fclose(file_id);

    return 0;
}
```

```bash
h5ls -ds testchar.h5
dset                     Dataset {81}
    Data:
         "abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz012345678901234567\000
```

These changes restore the double-quote for `h5ls` and refactor the logic for `h5dump` to fix some other similar formatting issues when data elements are wrapped to new lines according to the tool's column limit setting.


hyoklee:
What's the issue number for this PR?

hyoklee (approved):

release_docs/CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ The `h5repack` tool now obtains its default low and high library version bounds
 
 ## Tools
 
+### Fixed an issue with quoting of data values in h5ls and h5dump when displaying as ASCII characters
```
hyoklee, line 86: `quoting of data values` -> `data value quoting`

release_docs/CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
+
+   When using the `-s` (h5ls) or `-r` (h5dump) option to display 1-byte integer datasets and
+   attributes as ASCII characters, a closing double-quote character for data values was being
```
hyoklee: `was being dropped` -> `was dropped`

release_docs/CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
+   dropped in some cases. This double-quote character has been restored and similar formatting
```
hyoklee: `has been restored and similar` ->` has been restored, and similar`

release_docs/CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
+   issues have been fixed for cases where elements are wrapped to new lines according to the
```
hyoklee: `are wrapped` -> `wrap`

release_docs/CHANGELOG.md
```diff
@@ -239,6 +239,14 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
+   particular tool's column limit setting.
```
hyoklee: Remove `particular`.

hyoklee (dismissed):

release_docs/CHANGELOG.md
jhendersonHDF: This is at best a preference

release_docs/CHANGELOG.md
jhendersonHDF: Particular here is to distinguish between h5ls and h5dump, as each tool has its own column limit setting

release_docs/CHANGELOG.md
jhendersonHDF: I find the former better to read since the issue was with quoting and not "data value XXX"

jhendersonHDF:
> What's the issue number for this PR?

No issue currently, as the problem was reported external to GitHub

tools/lib/h5tools_dump.c
```diff
@@ -4518,19 +4518,6 @@ h5tools_dump_data(FILE *stream, const h5tool_format_t *info, h5tools_context_t *
                 status = h5tools_dump_mem(stream, &string_dataformat, &datactx, obj_id);
             }
         }
-        if (datactx.display_char && H5Tget_size(f_type) == 1 && H5Tget_class(f_type) == H5T_INTEGER) {
```
jhendersonHDF, line 4521: This logic was added so that h5dump adds the closing quote character when displaying the data as ASCII characters, but it goes against the general tools architecture and complicates the logic. The fix that follows the tools architecture, allowing `h5tools_simple_prefix()` to print the closing quote character, is below.

tools/src/h5ls/h5ls.c
```diff
@@ -1547,6 +1547,21 @@ dump_dataset_values(hid_t dset)
             h5tools_render_element(rawoutstream, info, &ctx, &buffer, &curr_pos, (size_t)info->line_ncols,
                                    (hsize_t)0, (hsize_t)0);
         }
+
+        /* Output closing line suffix character when printing 1-byte integer data as ASCII */
+        if (info->ascii) {
```
jhendersonHDF, line 1552: Since h5ls doesn't use the `h5tools_dump_data()` function from the tools library, it needs its own specific logic to cover this case.

release_docs/CHANGELOG.md
```diff
@@ -242,10 +242,10 @@ The `h5repack` tool now obtains its default low and high library version bounds
 ### Fixed an issue with quoting of data values in h5ls and h5dump when displaying as ASCII characters
 
    When using the `-s` (h5ls) or `-r` (h5dump) option to display 1-byte integer datasets and
-   attributes as ASCII characters, a closing double-quote character for data values was being
-   dropped in some cases. This double-quote character has been restored and similar formatting
-   issues have been fixed for cases where elements are wrapped to new lines according to the
-   particular tool's column limit setting.
+   attributes as ASCII characters, a closing double-quote character for data values was dropped
+   in some cases. This double-quote character has been restored and similar formatting issues
```
hyoklee, line 90: Add comma. `has been restored and similar` ->` has been restored, and similar`

hyoklee (dismissed):

tools/test/h5dump/h5dumpgentest.c
```diff
@@ -15179,3 +15181,37 @@ gent_float4(void)
...
+{
+    hsize_t dims[] = {81}; /* Assume default 80-column width for tools */
+    hid_t   file_id;
+    hid_t   dset_id;
+    hid_t   type_id;
+    hid_t   space_id;
+    hid_t   attr_id;
+
+    /* 81-byte string, including NUL terminator */
+    unsigned char *data = "abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz012345678901234567";
```
mattjala: I think casting this string literal to an unsigned type will trip a warning with `-Wpointer-sign` that can be avoided by using `const char *data` instead.

jhendersonHDF: Interesting that GCC seems to accept this but Clang doesn't. We currently only run the Werror build with GCC though and should add Clang. Will fix since the type doesn't strictly need to be `unsigned char` anyway.

lrknox (dismissed):

mattjala (dismissed):

jhendersonHDF:
Fixed the conflict in CHANGELOG.md

mattjala (approved):

lrknox (approved):
It appears you've considered previous suggestions for the CHANGELOG.md entry.  Most CHANGELOG.md entries are a single line for the description part of the entry;  you might consider changing that, or it can be done later if necessary.

jhendersonHDF:
> It appears you've considered previous suggestions for the CHANGELOG.md entry. Most CHANGELOG.md entries are a single line for the description part of the entry; you might consider changing that, or it can be done later if necessary.

I applied a few of the suggestions, but don't particularly agree with the remaining ones so left them alone.

PR #6557, opened 2026-07-23
One more fix related to issue #6391 - version 1 shared message encoding.

vchoi-hdfgroup:
One field is missing in Layout: Shared Message (Version 1) in section IV.A.2 Disk Format: Level 2A2 - Data Object Header Shared Message Encoding.  This is based on library coding in H5Oshared.c.


gheber (approved):
👍

PR #6561, opened 2026-07-24
Pin Java toolchain to the JDK found by find_package(Java)

### LINKED ISSUE #6559: CMake: Java toolchain is discovered independently of find_package(Java), allowing compiler and runtime to be different JDKs
> **Update.** Two corrections since this was filed, both material:
> the defect is **not macOS-specific**, and the fix suggested below was **wrong**.
> Details in the amended sections. Fixed in
> https://github.com/HDFGroup/hdf5/pull/6561.

## Summary

When `HDF5_BUILD_JAVA=ON`, HDF5 discovers Java **twice**, through two CMake mechanisms
that have opposite search precedence, and never reconciles them:

1. `find_package(Java)` at [`CMakeLists.txt:1098`](https://github.com/HDFGroup/hdf5/blob/develop/CMakeLists.txt#L1098) →
   `Java_JAVA_EXECUTABLE`, `Java_JAVAC_EXECUTABLE`, `Java_JAR_EXECUTABLE`, `Java_VERSION_STRING`
2. `project(HDF5_JAVA C Java)` at [`java/CMakeLists.txt:2`](https://github.com/HDFGroup/hdf5/blob/develop/java/CMakeLists.txt#L2) →
   `CMAKE_Java_COMPILER`, `CMAKE_Java_RUNTIME`, `CMAKE_Java_ARCHIVE`

These can resolve to **different JDKs**. HDF5 then makes decisions based on the JDK from (1)
while compiling and running with the JDK from (2).

## Symptom

On macOS this produces a build that compiles fine and then fails every single Java test.
Example: <https://my.cdash.org/builds/3805691> (macOS 26.5.2 arm64, xcode-27, HDF5 2.2.0,
`HDF5_ENABLE_JNI=ON`). ~95 tests fail — all `JUnit-*` and `H5_jcompat-h5-*`, nothing else —
each with:

```
-- JAVA COMMAND:  /usr/bin/java; -Xmx1024M ...
Exception in thread "main" java.lang.UnsupportedClassVersionError:
  test/TestH5Eparams has been compiled by a more recent version of the Java Runtime
  (class file version 69.0), this version of the Java Runtime only recognizes
  class file versions up to 65.0
```

Class file 69 = JDK 25, 65 = JDK 21. The bytecode was produced by one JDK and executed by
another.

Note `CMAKE_Java_RUNTIME` was recorded as **`/usr/bin/java`** — Apple's JavaVM-framework
stub, not a JDK. The stub selects a JDK at *each invocation* from `$JAVA_HOME`, so the
cache entry pins nothing; which JDK builds and which JDK tests is decided by whatever
environment happens to invoke it.

## Root cause

`FindJava.cmake` puts `JAVA_HOME` in **`HINTS`**, which outrank `$PATH`:

```cmake
if(_JAVA_HOME)
  list(APPEND _JAVA_HINTS ${_JAVA_HOME}/bin)
endif()
```

(`_JAVA_HOME` comes from `CMakeFindJavaCommon.cmake` and honors both `$ENV{JAVA_HOME}` and
`-DJAVA_HOME=`.) So `find_package(Java)` returns a coherent set of absolute paths from one
JDK.

`CMakeDetermineJavaCompiler.cmake` puts `$ENV{JAVA_HOME}/bin` in **`PATHS`**, the
lowest-priority slot, searched *after* `$PATH`:

```cmake
find_program(CMAKE_Java_COMPILER NAMES javac PATHS ${Java_BIN_PATH})
find_program(CMAKE_Java_RUNTIME  NAMES java  PATHS ${Java_BIN_PATH})
```

`/usr/bin` is always on `$PATH`, so on macOS `/usr/bin/javac` and `/usr/bin/java` always
shadow `$JAVA_HOME/bin/*`. It also ignores the CMake variable `JAVA_HOME` entirely.

Net effect: `JAVA_HOME` steers (1) but cannot steer (2).

## Impact beyond the test failures

The version gate at [`java/CMakeLists.txt:28`](https://github.com/HDFGroup/hdf5/blob/develop/java/CMakeLists.txt#L28)
selects the FFM vs JNI implementation from `Java_VERSION_STRING`:

```cmake
if (Java_VERSION_STRING VERSION_GREATER_EQUAL "25.0.0")
```

That version is measured by running `Java_JAVA_EXECUTABLE` — a *different* binary from the
one that compiles and runs the code. So it is possible for `find_package(Java)` to report 25,
HDF5 to switch to FFM and invoke `$ENV{JAVA_HOME}/bin/jextract`, while `CMAKE_Java_COMPILER`
resolves to an older JDK that cannot compile the generated bindings. The implementation
choice, the compiler, and the runtime are three independent decisions today.

This is also why [36ef3cc](https://github.com/HDFGroup/hdf5/commit/36ef3cc8e7d9888c374b1b5fac83cdaae495e5b0)
("Fix macos-26 java workflow", #6151) had to *remove* `JAVA_HOME` handling rather than
correct it — with the current CMake logic there is no way to set `JAVA_HOME` and get a
consistent toolchain. Not setting it works only because the runner image's default
`JAVA_HOME` then applies uniformly to both stubs.

## Not macOS-specific

The original report framed this as a macOS problem. It is not. On Windows with
`PATH` → JDK 11 and `JAVA_HOME` → JDK 20, current `develop` configures as:

```
-- Found Java: .../OpenJDK/jdk-20/bin/java.exe (found version "20.0.0")
-- Building HDF5 Java with JNI implementation (Java 20)

CMAKE_Java_COMPILER ".../Microsoft/jdk-11.0.16.101-hotspot/bin/javac.exe"
CMAKE_Java_RUNTIME  ".../Microsoft/jdk-11.0.16.101-hotspot/bin/java.exe"
TEST_JAVA=.../Microsoft/jdk-11.0.16.101-hotspot/bin/java.exe
```

The FFM/JNI gate ran against JDK 20 while the build compiled and tested with
JDK 11. No `/usr/bin` stubs involved. The stubs are what make the split *lethal*
on macOS — they let `javac` and `java` diverge within a single build — but the
underlying divergence is latent anywhere `JAVA_HOME` and `$PATH` disagree.

## Suggested fix

Propagate the toolchain from `find_package(Java)` before the `Java` language is
enabled, in the existing block in `CMakeLists.txt`, filling each variable
**independently**:

```cmake
if (NOT CMAKE_Java_COMPILER AND Java_JAVAC_EXECUTABLE)
  set (CMAKE_Java_COMPILER "${Java_JAVAC_EXECUTABLE}")
endif ()
if (NOT CMAKE_Java_RUNTIME AND Java_JAVA_EXECUTABLE)
  set (CMAKE_Java_RUNTIME "${Java_JAVA_EXECUTABLE}")
endif ()
if (NOT CMAKE_Java_ARCHIVE AND Java_JAR_EXECUTABLE)
  set (CMAKE_Java_ARCHIVE "${Java_JAR_EXECUTABLE}")
endif ()
```

**The earlier version of this section was wrong** and is worth spelling out,
because it fails in exactly the configuration people reach for. It proposed a
single all-or-nothing guard:

```cmake
if (NOT CMAKE_Java_COMPILER)          # <-- broken
  set (CMAKE_Java_COMPILER ...)
  set (CMAKE_Java_RUNTIME  ...)
  set (CMAKE_Java_ARCHIVE  ...)
endif ()
```

A caller pinning the compiler on the command line —

```bash
cmake -DJAVA_HOME=$JAVA_HOME -DCMAKE_Java_COMPILER=$JAVA_HOME/bin/javac ...
```

— sets `CMAKE_Java_COMPILER` as a cache entry, so that block is skipped
entirely. `CMakeDetermineJavaCompiler.cmake` then skips *its* block for the same
reason, and its guard closes only after the runtime and archive lookups. The
result is an empty `CMAKE_Java_RUNTIME`, an empty `TEST_JAVA` in the generated
CTest files, and ctest invoking no JVM at all — measured, not hypothesized:

```
CMAKE_Java_COMPILER = .../jdk-20/bin/javac.exe
CMAKE_Java_RUNTIME  =
CMAKE_Java_ARCHIVE  =
```

Filling the three in one at a time keeps an explicit override working while
still completing the toolchain.

Also dropped from the earlier suggestion: `find_package(Java COMPONENTS
Development REQUIRED)`. It is defensible, but it promotes a missing `javac`/`jar`
from "silently proceed" to a configure-time error and would turn some
currently-limping configurations into hard failures. That is a separate change.

## Verification

- **Windows**, `PATH` → JDK 11, `JAVA_HOME` → JDK 20: compiler, runtime, archiver
  and `TEST_JAVA` now all resolve to JDK 20, matching the `Java_VERSION_STRING`
  that drove the implementation choice.
- **macOS 26.5.2 arm64**, xcode-27, `JAVA_HOME` → JDK 25: the same build went
  from **95 failing Java tests to 0**
  (<https://my.cdash.org/builds/3805691> → <https://my.cdash.org/builds/3807172>).
- **GitHub `macos-26`**, three configurations — image-default `JAVA_HOME`;
  `JAVA_HOME` → 25 with only `-DCMAKE_Java_COMPILER` passed; `JAVA_HOME` → 25
  with JDK 21 first on `$PATH` — all yield a single-JDK toolchain and 255/255
  Java tests passing, including across the `ctest -D ExperimentalConfigure`
  reconfigure, where CMake is re-run in a shell whose `JAVA_HOME` has reverted to
  the image default.

## Not addressed

If `JAVA_HOME` is unset entirely on macOS, `find_package(Java)` itself falls
through to `$PATH` and finds `/usr/bin/java`, so the toolchain gets pinned to a
stub. It is at least pinned *consistently*, but hardening that would mean
rejecting `/usr/bin` runtimes or resolving through `/usr/libexec/java_home` at
configure time.

## Environment

- macOS 26.5.2 arm64, Xcode 27, CMake/CTest 4.4.0
- Windows 11, CMake 4.2, JDK 11 on `$PATH` / JDK 20 in `JAVA_HOME`
- HDF5 2.2.0 (`develop`), `HDF5_BUILD_JAVA=ON`, `HDF5_ENABLE_JNI=ON`, shared, OpenMPI, Fortran
- GitHub `macos-26` images ship JDK 11/17/21/25 with 21 as the default `JAVA_HOME`;
  `/usr/libexec/java_home` with no `-v` returns 25

hyoklee:
Close https://github.com/HDFGroup/hdf5/issues/6559.

## Problem

When `HDF5_BUILD_JAVA=ON`, HDF5 discovers Java **twice**, through two CMake
mechanisms with opposite search precedence, and never reconciles them:

1. `find_package(Java)` in `CMakeLists.txt` → `Java_JAVA_EXECUTABLE`,
   `Java_JAVAC_EXECUTABLE`, `Java_JAR_EXECUTABLE`, `Java_VERSION_STRING`
2. `project(HDF5_JAVA C Java)` in `java/CMakeLists.txt` → `CMAKE_Java_COMPILER`,
   `CMAKE_Java_RUNTIME`, `CMAKE_Java_ARCHIVE`

`FindJava.cmake` puts `JAVA_HOME` in **`HINTS`**, which outrank `$PATH`, and
honors both `$ENV{JAVA_HOME}` and `-DJAVA_HOME=` via `CMakeFindJavaCommon.cmake`.
`CMakeDetermineJavaCompiler.cmake` puts `$ENV{JAVA_HOME}/bin` in **`PATHS`**, the
lowest-priority slot, searched *after* `$PATH`, and ignores the CMake variable
`JAVA_HOME` entirely.

So when `JAVA_HOME` and `$PATH` name different JDKs, the two diverge. HDF5 then
evaluates the FFM/JNI version gate in `java/CMakeLists.txt`

```cmake
if (Java_VERSION_STRING VERSION_GREATER_EQUAL "25.0.0")
```

against one JDK while compiling and testing with another.

This is not macOS-specific. On Windows with `PATH` → JDK 11 and `JAVA_HOME` →
JDK 20, current `develop` reports:

```
-- Found Java: .../OpenJDK/jdk-20/bin/java.exe (found version "20.0.0")
-- Building HDF5 Java with JNI implementation (Java 20)

CMAKE_Java_COMPILER ".../Microsoft/jdk-11.0.16.101-hotspot/bin/javac.exe"
CMAKE_Java_RUNTIME  ".../Microsoft/jdk-11.0.16.101-hotspot/bin/java.exe"
TEST_JAVA=.../Microsoft/jdk-11.0.16.101-hotspot/bin/java.exe
```

macOS merely makes it lethal rather than latent: both lookups land on the
`/usr/bin` stubs, which are not JDKs but dispatchers that re-select one from
`$JAVA_HOME` at *each invocation*. `javac` and `java` can therefore resolve to
different JDKs inside a single build, and every Java test dies at class load:

```
-- JAVA COMMAND:  /usr/bin/java; -Xmx1024M ...
Exception in thread "main" java.lang.UnsupportedClassVersionError:
  test/TestH5Eparams has been compiled by a more recent version of the Java
  Runtime (class file version 69.0), this version of the Java Runtime only
  recognizes class file versions up to 65.0
```

(69 = JDK 25, 65 = JDK 21.)

## Fix

Propagate the toolchain from `find_package(Java)` before the `Java` language is
enabled, filling each variable **independently**.

Independently matters. The guard in `CMakeDetermineJavaCompiler.cmake` is
`if(NOT CMAKE_Java_COMPILER)` and it closes only *after* the runtime and archive
lookups, so a caller doing

```bash
cmake -DJAVA_HOME=$JAVA_HOME -DCMAKE_Java_COMPILER=$JAVA_HOME/bin/javac ...
```

suppresses the other two lookups entirely and ends up with an empty
`CMAKE_Java_RUNTIME` and an empty `TEST_JAVA` in the generated CTest files — so
ctest invokes no JVM at all. Filling the three in one at a time keeps an explicit
override working while still completing the toolchain.

## Verification

**Windows**, `PATH` → JDK 11, `JAVA_HOME` → JDK 20: compiler, runtime, archiver
and `TEST_JAVA` now all resolve to JDK 20, matching the `Java_VERSION_STRING`
that drove the implementation choice.

**macOS 26.5.2 arm64** (shared / OpenMPI / Fortran / Java), `JAVA_HOME` → JDK 25:
the same build went from **95 failing Java tests to 0**.

**GitHub `macos-26`**, three configurations — image-default `JAVA_HOME`;
`JAVA_HOME` → 25 with only `-DCMAKE_Java_COMPILER` passed; and `JAVA_HOME` → 25
with JDK 21 first on `$PATH` — all produce a single-JDK toolchain and 255/255
Java tests passing. This includes the `ctest -D ExperimentalConfigure` path,
where CMake is re-run in a shell whose `JAVA_HOME` has reverted to the image
default; the pin survives because `-DJAVA_HOME=` is a cache entry and
`CMakeFindJavaCommon.cmake` consults the CMake variable before
`$ENV{JAVA_HOME}`.

## Notes

- Behavior is unchanged when `JAVA_HOME` and `$PATH` already agree, which is the
  common case; the added block only fills variables that are otherwise unset.
- An explicit `-DCMAKE_Java_COMPILER=` / `-DCMAKE_Java_RUNTIME=` still wins.
- Not addressed here: if `JAVA_HOME` is unset entirely on macOS,
  `find_package(Java)` itself falls through to `$PATH` and finds `/usr/bin/java`,
  so the toolchain is pinned to a stub. It is at least pinned *consistently*, but
  hardening that would mean rejecting `/usr/bin` runtimes or resolving through
  `/usr/libexec/java_home` at configure time.
- No `release_docs/RELEASE.txt` entry is included; happy to add one if that is
  wanted for a build-system fix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1181,6 +1181,33 @@ if (HDF5_BUILD_JAVA)
   else ()
     find_package (Java)
     message (VERBOSE "Java version is ${Java_VERSION_STRING}")
+
+    # Pin the Java language toolchain to the JDK that find_package(Java) just
+    # probed. FindJava searches JAVA_HOME via HINTS, which outrank $PATH, but
+    # the toolchain enabled by project(... Java) in java/ searches
+    # $ENV{JAVA_HOME}/bin only as a low-priority PATHS entry, below $PATH. The
+    # two can therefore land on different JDKs - on macOS the toolchain picks
+    # the /usr/bin stubs, which re-resolve through JAVA_HOME on every
+    # invocation, so javac and java can differ and the resulting class files
+    # fail to load. Java_VERSION_STRING also gates the FFM/JNI choice in
+    # java/CMakeLists.txt, so it must describe the JDK that actually builds.
+    #
+    # Each variable is filled in independently. The guard in
+    # CMakeDetermineJavaCompiler.cmake is if(NOT CMAKE_Java_COMPILER) and it
+    # closes only after the runtime and archive lookups, so a caller passing
+    # just -DCMAKE_Java_COMPILER= suppresses the other two entirely and leaves
+    # CMAKE_Java_RUNTIME empty. Filling them one at a time keeps an explicit
+    # override working while still completing the toolchain.
+    if (NOT CMAKE_Java_COMPILER AND Java_JAVAC_EXECUTABLE)
+      set (CMAKE_Java_COMPILER "${Java_JAVAC_EXECUTABLE}")
+    endif ()
+    if (NOT CMAKE_Java_RUNTIME AND Java_JAVA_EXECUTABLE)
+      set (CMAKE_Java_RUNTIME "${Java_JAVA_EXECUTABLE}")
+    endif ()
+    if (NOT CMAKE_Java_ARCHIVE AND Java_JAR_EXECUTABLE)
+      set (CMAKE_Java_ARCHIVE "${Java_JAR_EXECUTABLE}")
+    endif ()
+    message (VERBOSE "Java toolchain: ${CMAKE_Java_COMPILER} / ${CMAKE_Java_RUNTIME} / ${CMAKE_Java_ARCHIVE}")
   endif ()
 endif ()
 


lrknox (approved):

jhendersonHDF (commented):
Rather than override the `CMAKE_Java_XXX` variables, we should probably remove Java from the list of languages specified in the Java-specific CMakeLists files and also push down this find_package call into those files. See the last comment on https://gitlab.kitware.com/cmake/cmake/-/work_items/18174. Also, the CMake documentation doesn't list support for Java as a language: https://cmake.org/cmake/help/latest/command/project.html#options.

PR #6568, opened 2026-07-24
Reviewed 2.0.0 occurrences in workflow files, updated "previous release"

lrknox:
versions to 2.1.0, changed default api tags to v200 and did not change azure/trusted-signing-action versions.

### DIFF
--- a/.github/workflows/main-spc.yml
+++ b/.github/workflows/main-spc.yml
@@ -309,7 +309,7 @@ jobs:
         working-directory: ${{ runner.workspace }}/build
 
   build_v2_0_linux:
-    name: "gcc DBG v2.0.0 default API no deprecated"
+    name: "gcc DBG v200 default API no deprecated"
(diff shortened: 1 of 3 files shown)


.github/workflows/main-spc.yml
```diff
@@ -309,7 +309,7 @@ jobs:
         working-directory: ${{ runner.workspace }}/build
 
   build_v2_0_linux:
-    name: "gcc DBG v2.0.0 default API no deprecated"
+    name: "gcc DBG v200 default API no deprecated"
```
lrknox, line 312: v200 is the current default api, successor to v114, v112, v110, v18, etc.  In general, functions will be versioned in major releases.  If that holds true, there may be future v300, v400, ... default apis.

jhendersonHDF (approved):

bmribler (approved):

PR #6576, opened 2026-07-29
Fix version references left wrong by the 2.2.0 alias update

brtnfld:
Follow-up to #6567, per review comment: https://github.com/HDFGroup/hdf5/pull/6567#issuecomment-5110473563

## Summary
- `sec_rel_spec_2x_migrate` in `hdf5_2x.dox` documents the fixed 1.14->2.0 transition (H5Dread_chunk/H5Tdecode/H5Iregister_type signature changes, the aws-c-s3 ROS3 backend swap), but its three CHANGELOG links used `\PREVIOUS_VER`, which floats to whatever release preceded the current one (now 2.1.1) instead of staying pointed at the 2.0.0 CHANGELOG that actually documents these changes. Pinned to `2.0.0` explicitly.
- `sw_changes_2.x.dox`'s ABI compatibility report entry said "Release 2.2.0 versus Release 2.1.0", but the true predecessor of 2.2.0 is 2.1.1 (the latest published release per `gh release list` and the abi-report.yml auto-detect logic). Converted this entry to use `\CURRENT_VER`/`\PREVIOUS_VER` (matching `hdf5_2x.dox`'s equivalent line) instead of a hardcoded literal — that duplication of the same fact across two files, one aliased and one not, is what let them drift out of sync in the first place.
- Left the `subsubsec_rel_spec_2x_change_22versus21` section heading/API list at 2.1.0 since that section tracks new APIs since the last minor release (2.1.1 was a patch release with no new APIs).

Note for future releases: before bumping `CURRENT_VER`/`PREVIOUS_VER` in `Doxyfile.in`, freeze the `sw_changes_2.x.dox` ABI entry to a literal first (it becomes historical record once superseded), then add a new aliased line above it for the new release.

## Test plan
- [ ] Doxygen build renders both pages without broken links

### DIFF
--- a/docs/doxygen/dox/hdf5_2x.dox
+++ b/docs/doxygen/dox/hdf5_2x.dox
@@ -98,7 +98,7 @@ Since this portion of the HDF5 documentation is now part of the source code, it 
 
 \section sec_rel_spec_2x_migrate Migrating from HDF5 1.14 to HDF5 2.x
 
-This section highlights some changes that might affect migrating to HDF5 2.x, please refer to the \ref sec_rel_spec_2x_change or the <a href="https://\SRCURLALL/hdf5_\PREVIOUS_VER/release_docs/CHANGELOG.md">CHANGELOG.md</a> for detail.
+This section highlights some changes that might affect migrating to HDF5 2.x, please refer to the \ref sec_rel_spec_2x_change or the <a href="https://\SRCURLALL/hdf5_2.0.0/release_docs/CHANGELOG.md">CHANGELOG.md</a> for detail.
 
 \li Default file format is now 1.8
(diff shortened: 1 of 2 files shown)


jhendersonHDF:
As the 2.2.0 tag has been created already, merging this would cause the release branch to drift from the tag which isn't ideal. Deleting the tag at this point is also not great. My thoughts are that the web-facing docs are more likely to be used than the local docs and this needs to be updated in develop too either way.

bmribler:
And I'm working on the PR to develop for the 2.2.0 release page, I will apply the changes here to that if it's ok with you, @brtnfld.

brtnfld:
Sounds good — please go ahead and fold these into the develop PR. Closing this one.

bmribler:
OK, thanks!

PR #6579, opened 2026-07-30
bound scaleoffset and fletcher32 reads to the chunk buffer

naruto-lgtm:
## Describe your changes

Repro: store a chunk with `H5Dwrite_chunk()` that is shorter than the fixed-size block the filter itself writes, then read the dataset back. 16 bytes for scaleoffset, whose parameter block is 21 bytes, or 2 bytes for fletcher32, whose checksum trailer is 4. A second scaleoffset case stores 25 bytes with minbits at full precision: long enough for the parameter block, shorter than the data the element count implies.
Cause: both read paths subtract their fixed block size from the chunk length before checking the chunk is that long, so the `size_t` subtraction wraps and the bound derived from it never trips. `H5Z__filter_scaleoffset()` also copies `d_nelmts * dtype_size` bytes out of the input buffer, with both counts coming from the filter pipeline message rather than from the buffer it was handed.
Fix: reject a chunk shorter than the filter's own header or trailer, and check the buffer holds the region scaleoffset is about to copy.

The chunk length is the trust boundary here rather than the element count. `*buf_size` is what the pipeline actually allocated, so validating against it at the top of the reverse path keeps every later index into that buffer sound without disturbing the filters' own arithmetic, and it is where the existing `H5_IS_BUFFER_OVERFLOW()` calls in scaleoffset already sit.

On the unpatched tree ASan reports `heap-buffer-overflow ... READ of size 16` at `H5Zscaleoffset.c:1248` against a 25 byte region, and a read 2 bytes before a 16 byte region at `H5Zfletcher32.c:75`. Both are clean afterwards. `test_short_filtered_chunk()` in `test/direct_chunk.c` covers all three cases and fails if either fix is reverted.


mattjala:
By and large I think this is a good thing to do, but we should check against stored chunk length rather than the allocated size, and should do similar checks in other filters.  I've built on top of these changes in #6585

I'm pretty sure the CI build failures here are unrelated to the changes in this PR.

naruto-lgtm:
You're right on the buffer size. H5D__chunk_lock() allocates MAX(chunk_disk_size, chunk_size) and passes that as *buf_size while nbytes carries the stored length, so my checks were bounded by the allocation rather than the data, and a truncated chunk could still decompress out of the uninitialized tail. ASan wouldn't have caught that one either, since those reads stay inside the allocation. #6585 is the right shape, so I'm happy for it to supersede this, and I can close this one whenever suits you.

On other filters, the one still open after #6585 looks like nbit. H5Z__filter_nbit() hands *buf to H5Z__nbit_decompress() with no length at all, and the decompress helpers walk it on a running index with d_nelmts coming from cd_values[2], so it's the same class but needs a size threaded through the static one_* helpers rather than a single header check. Happy to do that as a separate PR if it's useful.

One data point for the szip check in #6585: when I probed raw lengths of 1 to 3, libaec rejected the input before over-reading, so that one looked like a latent underflow rather than a reachable OOB. Still worth guarding, there just wasn't a crashing repro behind it.

Agreed the CI failures look unrelated.

naruto-lgtm:
You're right about nbytes. H5D__chunk_lock() sizes a filtered chunk buffer to MAX(stored, unfiltered) and fills only the stored bytes, so `*buf_size` is routinely longer than the data in it, and a read past the stored length stays inside the allocation where ASan never sees it. #6585 looks like the right shape to me. Happy to close this one in favor of it if that's cleaner for you.

On the CI failures, they're a doxygen error at docs/doxygen/dox/hdf5_2x.dox:118 ("Unexpected subsubsection command found inside section"), which #6582 already fixed on develop. This branch just predates it, so nothing to do with the filter changes.

One filter is still open after szip: H5Z__filter_nbit() doesn't pass nbytes to H5Z__nbit_decompress() at all, so the whole decompress chain walks the input at an offset driven by d_nelmts and cd_values out of the pipeline message with no bound on the buffer it was handed. Same class as the scaleoffset full-precision copy, but there's no length in scope to check against, so it means threading one through about six static helpers. That felt too big to fold into #6585. I can put it up separately if you want it.

release_docs/CHANGELOG.md
```diff
@@ -178,6 +178,10 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
 
+### Fixed out-of-bounds reads when a filtered chunk is shorter than its filter header
+
+   The scaleoffset and fletcher32 filters each store a fixed-size block alongside the chunk data: a 21-byte parameter block in front of the compressed stream for scaleoffset, and a 4-byte checksum after the data for fletcher32. On read, each one subtracted that fixed size from the chunk length without first checking the chunk was at least that long. A truncated or corrupted chunk record made the subtraction wrap around, so both filters read far past the end of the chunk buffer. `H5Z__filter_scaleoffset()` also copied `d_nelmts * dtype_size` bytes out of the buffer in its full-precision path, with both counts taken from the filter pipeline message rather than from the buffer it was given, so a small chunk with a large declared element count returned adjacent heap contents as dataset data. Both filters now reject a chunk shorter than their own header or trailer, and scaleoffset checks the buffer holds the region it is about to copy.
```
hyoklee, line 183: "checking the chunk was" → "checking that the chunk was"

"checks the buffer holds" → "checks that the buffer holds"

hyoklee (approved):

mattjala:
Closing since this is deprecated by #6585. The other nbit issue can be addressed separately.

PR #6585, opened 2026-07-31
Bound filter header reads to the stored chunk length

mattjala:
This PR is an expansion of #6579

The scaleoffset bounds checks measured against *buf_size, the size of the
chunk buffer, rather than nbytes, the number of bytes the pipeline read from
the file. H5D__chunk_lock() sizes a filtered chunk buffer to MAX(stored size,
unfiltered size) so filters have room to expand into, and fills only the
stored bytes, so the buffer is routinely longer than the data in it. A
truncated chunk therefore passed the header check and decompressed out of the
uninitialized remainder of the allocation, returning unrelated heap contents
as dataset data. Because the reads stay inside the allocation, the address
sanitizer does not see them.

This PR switches every read-side check in H5Z__filter_scaleoffset()
to bound against nbytes instead. Similar short-chunk checks are added to szip and fletcher32.

The direct chunk test has been extended to cover a chunk whose unfiltered size is
larger than the bytes stored for it, and to cover szip.


release_docs/CHANGELOG.md
```diff
@@ -178,6 +178,10 @@ The `h5repack` tool now obtains its default low and high library version bounds
...
...meter block in front of the compressed stream for scaleoffset, a 4-byte uncompressed-length header for szip, and a 4-byte checksum after the data for fletcher32. On read, each one subtracted that fixed size from the chunk length without first checking the chunk was at least that long. A truncated or corrupted chunk record made the subtraction wrap around, so all three filters read far past the end of the chunk buffer. `H5Z__filter_scaleoffset()` also copied `d_nelmts * dtype_size` bytes out of the buffer in its full-precision path, with both counts taken from the filter pipeline message rather than from the buffer it was given, so a small chunk with a large declared element count returned adjacent heap contents as dataset data. Its bounds checks also measured the chunk buffer's allocated size, which the library deliberately sizes to hold the larger unfiltered chunk, so a truncated chunk could still decompress out of the uninitialized remainder of that allocation. All three filters now reject a chunk shorter than their own header or trailer, and scaleoffset validates every read against the number of bytes actually stored for the chunk rather than the size of the buffer holding them.
```
hyoklee: scaleoffset, fletcher32 and szip-> scaleoffset, fletcher32, and szip

checking the chunk was -> checking that the chunk was

hyoklee (dismissed):

hyoklee (dismissed):

release_docs/CHANGELOG.md
```diff
@@ -77,6 +77,10 @@ We would like to thank the many HDF5 community members who contributed to this r
...
+
```
hyoklee, line 83: `now reject a chunk shorter than` -> `now reject any chunk shorter than`

hyoklee (dismissed):

hyoklee (dismissed):

src/H5Zscaleoffset.c
```diff
@@ -1235,6 +1242,12 @@ H5Z__filter_scaleoffset(unsigned flags, size_t cd_nelmts, const unsigned cd_valu
...
         p.minbits = minbits;
 
+        /* The compressed stream starts after the fixed-size parameter header, so
+         * the chunk has to hold that header before it can be skipped
+         */
+        if (H5_IS_BUFFER_OVERFLOW((unsigned char *)*buf, buf_offset, (unsigned char *)*buf + nbytes - 1))
+            HGOTO_ERROR(H5E_ARGS, H5E_BADVALUE, 0, "buffer too short");
+
         /* calculate size of output buffer after decompression */
         size_out = d_nelmts * (size_t)p.size;
```
jrideout, line 1257: `d_nelmts` is `cd_values[H5Z_SCALEOFFSET_PARM_NELMTS]` (`:1146`) and `p.size` is `cd_values[H5Z_SCALEOFFSET_PARM_SIZE]` (`:1195`). Neither is range-checked, and this product is passed to `H5MM_malloc()` a few lines below, before any of the new bounds run.

The `H5_IS_BUFFER_OVERFLOW(..., size_out, ... nbytes - 1)` added in the full-precision branch does catch an oversized `size_out`, but it runs after the allocation, so it bounds the copy rather than the request. `d_nelmts = 0xFFFFFFFF` with a 4-byte declared datatype requests about 16 GiB.

Over-allocation rather than corruption: on LP64 the product cannot overflow `size_t`, and `H5MM_malloc()` fails cleanly.

Suggested:

```c
if (p.size && d_nelmts > SIZE_MAX / p.size)
    HGOTO_ERROR(H5E_PLINE, H5E_OVERFLOW, 0, "scaleoffset output size overflows");
size_out = d_nelmts * (size_t)p.size;
```

Checking `size_out` against the dataset's chunk size would be tighter than `SIZE_MAX`.

mattjala: That's a real problem, but it's mostly unrelated to this PR, and the general problem of unvalidated `cd_values` shows up in other filter decode paths as well, so I think it should be addressed separately and all-at-once. I've added the SIZE_MAX guard here for now as a minor improvement, but the chunk size itself isn't reachable from inside the filter (`H5Z_func_t` callback only gets cd_values, nbytes, and buf).

hyoklee (approved):

Issue #6401, opened 2026-05-07
Heap-buffer-overflow in H5O__mtime_new_decode() in H5Omtime.c


TYGLS:
**Describe the bug**

A heap buffer overflow occurs in `H5O__mtime_new_decode` when parsing a malformed HDF5 file. 

**To Reproduce**

1. Clone the hdf5 repository and build it refer to oss-fuzz.

```shell
export CC=clang \
    CXX=clang++ \
    CFLAGS='-fsanitize=address -O0 -g' \
    CXXFLAGS='-fsanitize=address -O0 -g' \
    LIB_FUZZING_ENGINE='-fsanitize=fuzzer'
```

2. Run the PoC using h5_extended_fuzzer:

[poc.zip](https://github.com/user-attachments/files/27470497/poc.zip)

   The PoC is provided as a zip archive. After extracting, run:

```shell
./h5_extended_fuzzer ./poc
```

**ASAN Report**

```
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 469515172
./fuzzers/h5_extended_fuzzer: Running 1 inputs 1 time(s) each.
Running: ./crashes/poc
=================================================================
==591009==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x51200003bb57 at pc 0x5e2b67d7d5de bp 0x7ffd220317f0 sp 0x7ffd220317e8
READ of size 4 at 0x51200003bb57 thread T0
    #0 0x5e2b67d7d5dd in H5O__mtime_new_decode /home/hexijie/fuzz/project/hdf5/src/H5Omtime.c:137:5
    #1 0x5e2b67d289a0 in H5O__chunk_deserialize /home/hexijie/fuzz/project/hdf5/src/H5Ocache.c:1475:35
    #2 0x5e2b67d223bc in H5O__cache_deserialize /home/hexijie/fuzz/project/hdf5/src/H5Ocache.c:300:9
    #3 0x5e2b67abd47d in H5C__load_entry /home/hexijie/fuzz/project/hdf5/src/H5Centry.c:1241:26
    #4 0x5e2b67abd47d in H5C_protect /home/hexijie/fuzz/project/hdf5/src/H5Centry.c:3149:30
    #5 0x5e2b67a735bd in H5AC_protect /home/hexijie/fuzz/project/hdf5/src/H5AC.c:1302:26
[...]

PR #6589, opened 2026-08-03
check object header message size after decoding its header

naruto-lgtm:
## Describe your changes

Repro: open a file whose object header holds a message that declares a body reaching the end of its chunk, where the tail of the message header sits in those last bytes. Issue #6401 has a fuzzer PoC for it. By hand it is a version 1 header with a 12 byte chunk 0, one mtime message declaring an 8 byte body, and only 4 bytes left after the 8 byte message header.
Cause: `H5O__chunk_deserialize()` compares `mesg_size` against `p_end` while `chunk_image` still points at the flags byte, so the flags plus reserved bytes (version 1) or flags plus creation index (version 2) that follow are not counted. The body begins up to four bytes further into the chunk than the check assumed, and `mesg->raw` / `mesg->raw_size` then describe a region ending past the chunk image.
Fix: run the same check once the whole message header has been decoded, against the pointer that actually becomes the message body.

Every message class builds its own `p_end` from `raw` and `raw_size`, so this is not specific to mtime. Whichever decoder the type byte selects gets handed a buffer end outside the allocation, which is why the check belongs in the chunk loop rather than in the individual decoders, where the chunk image bound is not known.

On the unpatched library ASan reports a heap-buffer-overflow read at `H5Omtime.c:137` from `H5Ocache.c:1475`, 0 bytes past the chunk image, and is clean afterwards. The trailing image size check rejects the file either way, so the new case in `test/ohdr.c` earns its keep under a sanitizer or valgrind, the same way `test_ohdr_badness()` above it does.

#6401


release_docs/CHANGELOG.md
```diff
@@ -151,6 +151,12 @@ The `h5repack` tool now obtains its default low and high library version bounds
 
 ## Library
 
+### Fixed a heap buffer overflow when decoding object header messages
+
+   The size stored in an object header message header was checked against the chunk before the rest of that message header was decoded, so the check allowed a message body to start up to four bytes further into the chunk than the check accounted for. A corrupted or fuzzed file could declare a size that passed the check and still ran past the end of the chunk image, and the message's decode callback was then handed a buffer end outside the allocation. `H5O__chunk_deserialize()` now checks the message size once the whole message header has been decoded.
```
hyoklee: so the check allowed -> allowing

hyoklee (dismissed):

vchoi-hdfgroup (dismissed):

hyoklee (dismissed):

release_docs/CHANGELOG.md
naruto-lgtm: Done, reworded to "allowing".

release_docs/CHANGELOG.md
```diff
@@ -77,6 +77,12 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## Library
 
+### Fixed a heap buffer overflow when decoding object header messages
+
+   The size stored in an object header message header was checked against the chunk before the rest of that message header was decoded, allowing a message body to start up to four bytes further into the chunk than the check accounted for. A corrupted or fuzzed file could declare a size that passed the check and still ran past the end of the chunk image, and the message's decode callback was then handed a buffer end outside the allocation. `H5O__chunk_deserialize()` now checks the message size once the whole message header has been decoded.
```
hyoklee: `ran` -> `extended`

hyoklee (dismissed):

release_docs/CHANGELOG.md
naruto-lgtm: Done, switched to extended.

hyoklee (dismissed):

vchoi-hdfgroup (dismissed):

lrknox:
@vchoi-hdfgroup @hyoklee  I resolved the conflict in CHANGELOG.md, adding the release note with no other changes.  Your approvals were dismissed; can you check and approve again?

hyoklee (approved):

vchoi-hdfgroup (approved):

PR #6594, opened 2026-08-04
Close datatype IDs derived from memory type in the JNI translation helpers

### LINKED ISSUE #6592: `H5Tget_super()` datatype IDs leaked on the normal path in `h5util.c` translate helpers
There is a datatype-ID leak in the HDF5 Java translation helpers. The `H5T_VLEN`, `H5T_ARRAY`, and `H5T_COMPLEX` branches of all four `translate_*` helpers obtain a base datatype ID with `H5Tget_super()` and never close it. Unlike a JNI local reference, a leaked `hid_t` is not reclaimed when the native method returns, so these IDs accumulate in the library for the lifetime of the process.

`java/src-jni/jni/h5util.c`

Functions:

- `translate_atomic_rbuf`
- `translate_atomic_wbuf`
- `translate_rbuf`
- `translate_wbuf`

### 1. `H5Tget_super()` IDs are never closed

The pattern is the same in each branch: acquire `memb`, use it to size and classify the base type, recurse, then `break` out of the switch:

```c
case H5T_ARRAY: {
    if (!(memb = H5Tget_super(mem_type_id)))
        H5_LIBRARY_ERROR(ENVONLY);
    if ((vlClass = H5Tget_class(memb)) < 0)
        H5_LIBRARY_ERROR(ENVONLY);
    if (!(vlSize = H5Tget_size(memb)))
        H5_LIBRARY_ERROR(ENVONLY);
    ...
    translate_rbuf(ENVONLY, jList, memb, vlClass, (jsize)typeCount, objBuf, typeSize);
    ...
    break;
}
```

and every one of the four helpers ends with a `done:` label that does no
cleanup at all:

```c
done:

    return;
}
```

The affected sites (line numbers against the current `develop` tree):

| Function | `H5T_VLEN` | `H5T_ARRAY` | `H5T_COMPLEX` | bare `done:` |
| --- | --- | --- | --- | --- |
| `translate_atomic_rbuf` | 4300 | 4364 | 4532 | 4566 |
| `translate_atomic_wbuf` | 4612 | 4695 | 4822 | 4865 |
| `translate_rbuf` | 4925 | 5006 | 5074 | 5118 |
| `translate_wbuf` | 5154 | 5260 | 5322 | 5372 |

Note that the `H5T_COMPOUND` branch in each helper *does* get this right — it closes the per-member type inside the loop (lines 4355, 4686, 4991, 5251):

```c
jobj = translate_atomic_rbuf(ENVONLY, memb, memb_vlClass,
                             char_buf + i * typeSize + memb_offset, memb_vlSize);
...
H5Tclose(memb);
```

so the missing `H5Tclose()` in the super-type branches looks accidental rather than intentional.

This leaks on the ordinary success path, not just on errors: every read or write of variable-length, array, or complex data leaks at least one datatype ID, and because these helpers recurse for nested types, one call can leak several.

### 2. `translate_atomic_rbuf` leaks class references per element

`translate_atomic_rbuf()` unconditionally looks up seven classes on entry and
never deletes any of them:

```c
jclass    arrCList      = ENVPTR->FindClass(ENVONLY, "java/util/ArrayList");
...
jclass cByte   = ENVPTR->FindClass(ENVONLY, "java/lang/Byte");
jclass cShort  = ENVPTR->FindClass(ENVONLY, "java/lang/Short");
jclass cInt    = ENVPTR->FindClass(ENVONLY, "java/lang/Integer");
jclass cLong   = ENVPTR->FindClass(ENVONLY, "java/lang/Long");
jclass cFloat  = ENVPTR->FindClass(ENVONLY, "java/lang/Float");
jclass cDouble = ENVPTR->FindClass(ENVONLY, "java/lang/Double");
```

These are JNI local references, so they are reclaimed when the outer native method returns — this is not a leak across calls. The problem is the count within a single call. `translate_rbuf()` invokes `translate_atomic_rbuf()` once per member per row from its `H5T_COMPOUND` branch:

```c
for (i = 0; i < (size_t)count; i++) {
    ...
    for (x = 0; x < (size_t)nmembs; x++) {
        ...
        jobj = translate_atomic_rbuf(ENVONLY, memb, memb_vlClass,
                                     char_buf + i * typeSize + memb_offset, memb_vlSize);
```

so reading a compound dataset with N rows and M members creates roughly `7 * N * M` local references that all stay live until the read returns. For a large compound read that is enough to grow the local reference table substantially, and it will show up under `-Xcheck:jni`.

`translate_wbuf()` has the same unreleased `FindClass("java/util/ArrayList")` on each recursion level, though at much lower multiplicity.

### Suggested fix

For the datatype IDs, close `memb` on every path where `H5Tget_super()` succeeded. Since the helpers already use `goto done` error macros, the natural form is to initialise `memb` to `H5I_INVALID_HID`, reset it after each use, and close it in the `done:` block:

```c
hid_t memb = H5I_INVALID_HID;

...

done:
    if (memb >= 0)
        H5Tclose(memb);

    return;
```

The `H5T_COMPOUND` loops already close their per-member IDs, so those should set `memb = H5I_INVALID_HID` after `H5Tclose()` to avoid a double close.

For the class lookups in `translate_atomic_rbuf()`, the cheapest fix is to hoist them out of the per-element helper — the seven classes are constant, so caching them once (as global references at library init, alongside the other cached IDs) removes the per-element cost entirely. Deleting them before each `return` would also work but is more error-prone given the number of exit paths.

mattjala:
The object-tree read/write helpers in h5util.c derive a base datatype from the memory type with H5Tget_super() for the variable-length, array and complex classes, but never closed it. Because an hid_t is not reclaimed when the native method returns, every read or write of such data leaked at least one datatype ID for the lifetime of the process, and nested types leaked one per level.

This PR updates the helpers to close the derived type in their done: blocks, which covers both the success and the error paths, and to reset the id in the compound loops to avoid the potential for double closes.

It also has the helpers release the class references that the per-element helpers look up on entry. These are local references, so they were reclaimed when the enclosing native method returned, but a compound read calls the helper once per member per element and held one set per call until then. Releasing them at the single exit bounds the count of references to the recursion depth.

Fixes #6592


java/src-jni/test/TestH5D.java
```diff
@@ -1695,6 +1695,229 @@ public void testH5DArray_datatype_ids_stable() throws Throwable
...
+                             " write/read cycles on a variable-length datatype",
+                         before, openDatatypeCount());
+        }
+        finally {
+            if (dset_id >= 0)
+                try {
+                    H5.H5Dclose(dset_id);
+                }
+                catch (Exception ex) {
+                }
```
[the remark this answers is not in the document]
mattjala: This is the same pattern used elsewhere in the file. Close failures aren't directly related to this PR, since this test and fix regard the leaking of transient IDs.

jhendersonHDF (dismissed):

java/src-jni/jni/h5util.c
```diff
@@ -4558,6 +4559,27 @@ translate_atomic_rbuf(JNIEnv *env, jlong mem_type_id, H5T_class_t type_class, vo
     } /* switch(type_class) */
 
 done:
+    /* The base type derived from a VLEN/ARRAY/COMPLEX memory type, or the member type of
+     * a COMPOUND that has not been closed by the loop above, is owned by this call. */
+    if (memb >= 0)
```
bmribler, line 4564: H5Tclose(0) will fail, right?

bmribler (changes requested):

java/src-jni/jni/h5util.c
mattjala: True. `memb` shouldn't be able to be zero exactly here, but we may as well be precise. I'll update this check and some similar ones.

bmribler: Thanks!

jhendersonHDF: I don't believe it's guaranteed anywhere that the ID won't be 0 or that `H5Tclose(0)` will fail. The H5I code called when closing also asserts `id >= 0` and checking for `>= 0` is a fairly common pattern throughout the library.

mattjala: H5Tclose(0) does fail at present, but that's a consequence of how the hash table is set up presently and, I suppose, not something we want to design against. Reverted to `>= 0` for consistency.

jhendersonHDF: A direct test of `H5Tclose(0)` would fail if there is no open datatype object that was assigned an ID of 0. Whether or not an ID can actually ever be 0 is something I don't currently know, but at least the library code seems designed to accept that this could happen.

jhendersonHDF (approved):

lrknox (approved):

bmribler (approved):

PR #6595, opened 2026-08-05
Document the Metadata Cache Image block and header message

vchoi-hdfgroup:
Adds docs/doxygen/dox/H5.format.4.0.dox coverage for:
- Level 1J: the Metadata Cache Image Block (subsec_fmt4_infra_mdci)
- IV.A.3.y: the Metadata Cache Image header message (0x0018)
- TOC/cross-reference entries and the superblock-extension list entry for both

Also clarifies the 'Size of Header Message #n Data' field description for v1 and v2 object header messages to note it includes flags/reserved (v1) or flags/creation-order (v2) fields, not just type and length.

### DIFF
--- a/docs/doxygen/dox/H5.format.4.0.dox
+++ b/docs/doxygen/dox/H5.format.4.0.dox
@@ -35,6 +35,7 @@ Navigate back: \ref index "Main" / \ref SPEC
           <li> @ref subsec_fmt4_infra_fractalheap</li>
           <li> @ref subsec_fmt4_infra_freespaceindex</li>
           <li> @ref subsec_fmt4_infra_sohm</li>
+          <li> @ref subsec_fmt4_infra_mdci</li>
           </ol></li>
       <li> @ref sec_fmt4_dataobject
         <ol type="A">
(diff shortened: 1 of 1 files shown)


brtnfld:
**Should fix**

docs/doxygen/dox/H5.format.4.0.dox (new subsec_fmt4_infra_mdci text, diff ~lines 403–406): the paragraph states the MDCI message "is removed from the superblock extension once the library has read the referenced block back in at the next file open, at which point the file space it occupied is reclaimed." This is only true when the file is opened read-write. Per src/H5Cimage.c:698-706 (comment) and cache_ptr->delete_image = rw at line 729, opening the file read-only deliberately leaves the message and the on-disk image block untouched (dirty entries are instead marked clean in memory to suppress a write attempt on close). As written, the doc will mislead anyone using it to predict on-disk state across multiple opens, or to implement a third-party reader/writer against this spec.
Fix: add a clause, e.g. "...at the next file open when the file is opened for read/write access. If the file is opened read-only, the message and image block are left unchanged."

**Consider**

docs/doxygen/dox/H5.format.4.0.dox (new IV.A.3.y message table, diff ~lines 392–394): the Status row is missing its closing </td> before </tr>:

<td colspan="2"><b>Status:</b> Optional; may not be repeated. 
</tr>

Looks good otherwise

brtnfld (approved):

PR #6597, opened 2026-08-05
Auto merge dependabot

lrknox:
This was the prompt given to Claude:

dependabot creates github PRs with "bump he github-actions group with" in the title.  I would like to add a workflow that reviews the checks for any PR created by dependabot, verifies that all checks are finished and that none of them failed and that there were no errors.  If not, the workflow will approve and commit the PR.  If such workflow can be created it should run daily at an appropriate time after dependabot creates PRs, probably at an early morning time before typical developer activity begins.

### DIFF
--- a/.github/scripts/dependabot-auto-merge.js
+++ b/.github/scripts/dependabot-auto-merge.js
@@ -63,7 +63,7 @@ async function alreadyApproved({ github, owner, repo, pull_number }) {
   return reviews.some((r) => r.user?.login === APPROVER_LOGIN && r.state === "APPROVED");
 }
 
-async function processPR({ github, owner, repo, pr, core }) {
+async function processPullRequest({ github, owner, repo, pr, core }) {
   const { ready, reason } = await allChecksPassed({ github, owner, repo, ref: pr.head.sha });
   if (!ready) {
     core.info(`Skipping: ${reason}`);
@@ -125,7 +125,7 @@ async function run({ github, context, core }) {
   for (const pr of candidates) {
     core.startGroup(`#${pr.number} — ${pr.title}`);
     try {
-      await processPR({ github, owner, repo, pr, core });
+      await processPullRequest({ github, owner, repo, pr, core });
     } catch (e) {
       core.error(`#${pr.number}: ${e.message}`);
     } finally {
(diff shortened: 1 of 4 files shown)


jhendersonHDF (commented):
Just putting this out there as an opinion: auto-merging anything doesn't seem like a particularly good idea, but that also assumes that someone is going to review the updates made by dependabot before merging, which historically hasn't really been the case. https://github.blog/security/supply-chain-security/the-case-for-a-cooldown-why-dependabot-now-waits-before-issuing-version-updates/ tells me that the more friction involved, the better.

hyoklee (approved):

lrknox:
> Just putting this out there as an opinion: auto-merging anything doesn't seem like a particularly good idea, but that also assumes that someone is going to review the updates made by dependabot before merging, which historically hasn't really been the case. https://github.blog/security/supply-chain-security/the-case-for-a-cooldown-why-dependabot-now-waits-before-issuing-version-updates/ tells me that the more friction involved, the better.

Should we add another few days delay, or do you think that merging should just not be automated?

jhendersonHDF:
> Should we add another few days delay, or do you think that merging should just not be automated?

I lean a bit more toward not auto-merging anything ever, but if no one actually reviews all the changes dependabot is making the effect is really the same. I think my main complaint would be that it seems unnecessary to add a couple hundred lines of workflow/js logic to be maintained when someone can just easily use the UI to merge the PRs. That also at least gives the slight chance that someone might see or remember an issue with a particular version of an action.

brtnfld:
To me, "**Depend**-abot" is not living up to its name if we CANNOT depend on what it is doing. Ain't nobody got the time to make sure Depend-a-bot is dependable. I don't see a difference between auto-merge and someone just blindly clicking a button.

jhendersonHDF:
> "Depend-abot" is not living up to its name if we CANNOT depend on what it is doing. Ain't nobody got the time to make sure Depend-a-bot is dependable.

Think "dependencies" bot, not "dependable" bot.

> I don't see a difference between auto-merge and someone just blindly clicking a button.

Arguably, someone _should_ be reviewing the changes in the dependencies. But blindly clicking a button also takes all of a few seconds for someone to do, instead of running a workflow every day that will just check out the library and then do nothing the vast majority of the time.

lrknox:
Decided at engineering standup 8/18/2026 to close this PR and continue human review and click to merge dependabot PRs.

PR #6598, opened 2026-08-06
Fix standalone example build issues

mattjala:
* Fixed the C++ examples failing to compile when built standalone

The standalone examples build used C++98, but `H5public.h` includes
`<cinttypes>`, which requires C++11. This affected any C++ translation unit
including `hdf5.h`, and did not match the HDF5 C++ library itself, which is
built as C++11. The C++ examples did not compile, against either static or
shared HDF5. The examples are now built as C++11.

Only the standalone build was affected. Examples built as part of the HDF5
build inherit the library's own C++ standard.

* Fixed the examples skipping the HL, Fortran and C++ programs in some configurations

When built standalone against an installed HDF5, the examples chose between
the shared and static HL, Fortran and C++ libraries using `BUILD_SHARED_LIBS`,
while the C library used `H5EXAMPLE_USE_SHARED_LIBS`. Since
`H5EXAMPLE_USE_SHARED_LIBS` determines which component is requested from
`find_package`, and therefore which `HDF5_<linkage>_<lang>_FOUND` variables
exist, `BUILD_SHARED_LIBS` could not select a linkage on its own. With
`H5EXAMPLE_USE_SHARED_LIBS` on and `BUILD_SHARED_LIBS` unset, those examples
were disabled with a "libs not found" message even though the libraries were
installed and had been found. The selection now uses
`H5EXAMPLE_USE_SHARED_LIBS`, matching the C library.

Builds driven through `CTestScript.cmake` were not affected, since its cache
file forces `BUILD_SHARED_LIBS` on. This affected cases where the examples
were built directly without that cache file.

BUILD_SHARED_LIBS remains documented as a user option in
config/examples/HDF5_Examples_options.cmake but no longer influences
library selection in this case. Pending discussion, we may want to remove that variable
or rework this mechanism.


release_docs/CHANGELOG.md
```diff
@@ -85,6 +85,34 @@ We would like to thank the many HDF5 community members who contributed to this r
...
+  exist, `BUILD_SHARED_LIBS` could not select a linkage on its own. With
+  `H5EXAMPLE_USE_SHARED_LIBS` on and `BUILD_SHARED_LIBS` unset, those examples
+  were disabled with a "libs not found" message even though the libraries were
+  installed and had been found. The selection now uses
+  `H5EXAMPLE_USE_SHARED_LIBS`, matching the C library.
+
+  Builds driven through `CTestScript.cmake` were not affected, since its cache
+  file forces `BUILD_SHARED_LIBS` on. This affected cases where the examples
+  were built directly without that cache file.
+
```
bmribler, line 121: Just curious, is it possible to combine them so that there are no two different things giving the same effect?

jhendersonHDF: Since the examples project is currently meant to be built as both a standalone project and as part of HDF5, it might complicate the logic a little to deal with this. It could make sense to inherit the default for `H5EXAMPLE_USE_SHARED_LIBS` from `BUILD_SHARED_LIBS` though.

jhendersonHDF (approved):

PR #6599, opened 2026-08-07
Minor correction to H5Tget_super failure checks in JNI helpers

mattjala:
The base type lookups in the object-tree helpers tested the returned hid_t for truth rather than for a negative value. A failed lookup returns H5I_INVALID_HID, so any failure wouldn't be caught until later.

Rework the checks to use the more standard `< 0` comparison for ID-related failures.

### DIFF
--- a/java/src-jni/jni/h5util.c
+++ b/java/src-jni/jni/h5util.c
@@ -4291,7 +4291,7 @@ translate_atomic_rbuf(JNIEnv *env, jlong mem_type_id, H5T_class_t type_class, vo
 
     switch (type_class) {
         case H5T_VLEN: {
-            if (!(memb = H5Tget_super(mem_type_id)))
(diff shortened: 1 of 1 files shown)


bmribler (approved):

java/src-jni/jni/h5util.c
```diff
@@ -4291,7 +4291,7 @@ translate_atomic_rbuf(JNIEnv *env, jlong mem_type_id, H5T_class_t type_class, vo
 
     switch (type_class) {
         case H5T_VLEN: {
-            if (!(memb = H5Tget_super(mem_type_id)))
+            if ((memb = H5Tget_super(mem_type_id)) < 0)
```
jhendersonHDF, line 4294: Good catch; likely a copy-paste from the nearby calls to `H5Tget_size()`

jhendersonHDF (approved):

PR #6603, opened 2026-08-11
Reject find_package calls with non-standard capitalization

mattjala:
HDF5 installs hdf5-config.cmake, and CMake's config-mode search accepts either <Name>Config.cmake or <lowercased-Name>-config.cmake, so find_package() locates HDF5 for any capitalization the caller provides (HDF5, hdf5, HdF5). CMake then records the requested components under the exact capitalization that was used, in <capitalization>_FIND_COMPONENTS.

The config reads that state as ${HDF5_PACKAGE_NAME}_FIND_COMPONENTS, and HDF5_PACKAGE_NAME is the uppercased package name, so it always reads the specific variable HDF5_FIND_COMPONENTS. Any capitalization other than HDF5 has its component list read back as empty. The package is still found, but the config concludes that no components were requested and applies its defaults.

This PR swaps ${HDF5_PACKAGE_NAME} for ${CMAKE_FIND_PACKAGE_NAME} in the names of variables used by CMake's mechanisms in order to correctly retain the components list for any provided capitalization pattern for HDF5.

EDIT: The problem this PR addressed is now resolved by rejecting incorrect capitalizations instead of doing work to make sure they still get the correct configuration information.


jhendersonHDF:
I may need to think about this more, but this is a case where I might actually suggest simply rejecting any package name that isn't exactly `HDF5` in our configuration file instead of trying to work around the problem. Both `find_package (hdf5 MODULE REQUIRED)` and `find_package (HdF5 MODULE REQUIRED)` will fail because the module in CMake is `FindHDF5.cmake` and the convention for variables set by that is `HDF5_XXX`, so we typically would want to always match that convention in our configuration file for variables set for compatibility. Something like `hdf5_XXX` would only confuse the issue.

Issue #6483, opened 2026-06-26
Format Specification: corrections to dimension-related fields in the document


vchoi-hdfgroup:
**Describe the bug**
Problems with dimension-related fields throughout the document:
1. N is often used but undefined and unclear in what that is
2. Inconsistent indexing base (0- or 1-based)
3. Use of "dimension size" to indicate element count conflicts with the document's usage of "size" to indicate byte-count e.g. size of offset, size of lengths

**Expected behavior**
Need to identify and correct areas in the document with the above problems:
1. Define an explicit variable and state what that is
2. Use consistent indexing base throughout the document
3. Suggest to use "dimension extent" which is clear to mean a span of elements not bytes

vchoi-hdfgroup:
Fixed with PR #6604.

PR #6604, opened 2026-08-13
Fix for issue #6483

vchoi-hdfgroup:
Fix undefined and inconsistent dimension indexing in the format spec.:
- Dataspace Message (Versions 1 and 2)
- Array Datatype Property (Datatype Versions 2 and 3)
- Compound Datatype Version 1 
- Data Layout Message (Versions 1/2 and 4/5) and their Chunked Storage
  Property Description tables
- Chunk index computation example in Appendix C
- Hyperslab Selection Info (Versions 1, 2, 3 Regular, 3 Irregular)
- Version 2 B-tree Type 10/11 Record Layout


gheber (approved):
👍

PR #6607, opened 2026-08-13
Fix memory leaks and ID reference counting in H5E code

jhendersonHDF:
When pushing errors to an error stack that is already full, the library skipped the push operation but didn't inform calling code about what happened. This resulted in calling code leaking memory, leaving reference counts on IDs incremented and causing an infinite loop while closing the library.


hyoklee (dismissed):

mattjala (dismissed):

jhendersonHDF:
Fixed merge conflict

mattjala (approved):

PR #6614, opened 2026-08-16
bound nbit decompression reads to the stored chunk length

naruto-lgtm:
## Describe your changes

The reverse nbit filter decompresses a chunk by walking its bit reader through the input buffer using the element count and precision from the on-disk filter pipeline message, but it never gets the number of bytes actually stored for the chunk. `H5D__chunk_lock()` sizes a filtered chunk buffer to `MAX(stored, unfiltered)` and fills only the stored bytes, so the buffer handed to the filter is routinely longer than the data in it. A truncated or corrupted chunk then decompresses out of the uninitialized remainder of that allocation and hands it back as dataset data.

This threads the stored length through `H5Z__nbit_decompress` and its per-type helpers and rejects any read past it, matching the bound the scaleoffset decompressor already carries. Keeping the check next to each `buffer[*j]` read means every element/array/compound path is covered without callers having to pre-validate the chunk. This is the nbit case left over from the filter bounds discussion in #6585.

Follow-up to #6585 (nbit was noted there as a separate fix).

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -54,9 +54,9 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## Library
 
-### Bound nbit decompression reads to the stored chunk length
+### Bound n-bit decompression reads to the stored chunk length
 
-   The reverse nbit filter walked its bit reader through the chunk buffer using only the element count and precision from the filter pipeline message, never the number of bytes actually stored for the chunk. The chunk buffer the filter is handed is sized to hold the larger unfiltered chunk, so a truncated or corrupted chunk decompressed out of the uninitialized remainder of that allocation and returned it as dataset data. The decompression path now carries the stored length and rejects any read past it.
+   The reverse n-bit filter walked its bit reader through the chunk buffer using only the element count and precision from the filter pipeline message, never checking the number of bytes actually stored for the chunk. Because the chunk buffer provided to the filter is sized to hold the larger unfiltered chunk, a truncated or corrupted chunk would decompress out of the uninitialized remainder of that allocation and return it as dataset data. The decompression path now tracks the stored length and rejects any read past it.
 
 ## Parallel Library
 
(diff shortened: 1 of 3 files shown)


nopa12:
there is one gap in the same file it doesn't cover: `H5Z__nbit_decompress_one_array` still divides by a file-controlled size with no zero-check:
```c
n = total_size / p.size;   // also: total_size / base_size
```

naruto-lgtm:
Good catch. All three divisions in H5Z__nbit_decompress_one_array take their divisor from the pipeline message, and the precision/offset check passes when both are zero, so a zero size did reach the division. Pushed a zero-check before each one; the nbit tests in dsets and direct_chunk still pass.

release_docs/CHANGELOG.md
```diff
@@ -54,6 +54,10 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## Library
 
+### Bound nbit decompression reads to the stored chunk length
+
+   The reverse nbit filter walked its bit reader through the chunk buffer using only the element count and precision from the filter pipeline message, never the number of bytes actually stored for the chunk. The chunk buffer the filter is handed is sized to hold the larger unfiltered chunk, so a truncated or corrupted chunk decompressed out of the uninitialized remainder of that allocation and returned it as dataset data. The decompression path now carries the stored length and rejects any read past it.
```
hyoklee: `nbit` -> `n-bit`

Fix grammar:

```
The reverse n-bit filter walked its bit reader through the chunk buffer using only the element count and precision from the filter pipeline message, never checking the number of bytes actually stored for the chunk. Because the chunk buffer provided to the filter is sized to hold the larger unfiltered chunk, a truncated or corrupted chunk would decompress out of the uninitialized remainder of that allocation and return it as dataset data. The decompression path now tracks the stored length and rejects any read past it.
```

hyoklee (dismissed):

release_docs/CHANGELOG.md
naruto-lgtm: Done, took your wording and switched the heading to n-bit as well.

hyoklee (approved):

Issue #6605, opened 2026-08-13
Format Specification: clean up Version 2 B-tree section


vchoi-hdfgroup:
***

**Describe the bug**
N is overloaded in version 2 B-tree internal node layout.

**Expected behavior**
Define symbols clearly for each counter field to be described.
Follow the indexing convention in the spec for concrete-vs-symbol list indices.

vchoi-hdfgroup:
fixed with PR #6615

PR #6615, opened 2026-08-16
Fix issue #6605:

vchoi-hdfgroup:
--Fix overloaded N and inconsistent # indexing in V2 B-tree section 
--Clean up Global Heap section as well.


gheber (approved):
👍

PR #6637, opened 2026-08-23
bound the table title read in H5TBinsert_field/H5TBdelete_field

naruto-lgtm:
## Describe your changes

Repro: open a table whose TITLE attribute is longer than 254 bytes, then call H5TBinsert_field or H5TBdelete_field on it.
Cause: both read the title back with H5TBAget_title(), which passes their fixed char[255] straight to H5Aread(). H5Aread() copies the whole stored attribute, so a longer title runs off the stack buffer (ASan reports a 600-byte write into table_title[255]).
Fix: read the title through a bounded helper that sizes its own buffer to the stored attribute and copies back a length-capped, NUL-terminated string. H5TBAget_title() has no length argument so it cannot be made safe in place; the fix lives at the two callers that pair it with a fixed buffer.

N/A

### DIFF
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -148,12 +148,12 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ### Bounded the table TITLE read in H5TBinsert_field/H5TBdelete_field
 
-   Both functions read a table's TITLE attribute back with H5TBAget_title(),
+   Both functions read a table's TITLE attribute back using H5TBAget_title(),
    which passes the caller's buffer straight to H5Aread(). The buffer is a
    fixed 255-byte stack array, but H5Aread() copies the whole stored
-   attribute, so opening a table whose TITLE is longer than that buffer
-   overran the stack. The title is now read into a buffer sized to the stored
-   attribute and copied back with a length cap.
+   attribute. Consequently, opening a table with a TITLE longer than the
+   buffer overran the stack. The title is now read into a buffer sized to the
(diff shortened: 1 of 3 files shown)


release_docs/CHANGELOG.md
```diff
@@ -146,6 +146,15 @@ We would like to thank the many HDF5 community members who contributed to this r
...
 
+### Bounded the table TITLE read in H5TBinsert_field/H5TBdelete_field
+
+   Both functions read a table's TITLE attribute back with H5TBAget_title(),
+   which passes the caller's buffer straight to H5Aread(). The buffer is a
+   fixed 255-byte stack array, but H5Aread() copies the whole stored
+   attribute, so opening a table whose TITLE is longer than that buffer
+   overran the stack. The title is now read into a buffer sized to the stored
+   attribute and copied back with a length cap.
+
```
hyoklee, line 157: Fix run-on sentences:

```
Both functions read a table's TITLE attribute back using H5TBAget_title(), which passes the caller's buffer straight to H5Aread(). The buffer is a fixed 255-byte stack array, but H5Aread() copies the whole stored attribute. Consequently, opening a table with a TITLE longer than the buffer overran the stack. The title is now read into a buffer sized to the stored attribute and copied back with a length cap.
```

hyoklee (dismissed):

release_docs/CHANGELOG.md
naruto-lgtm: Done, applied your wording.

hyoklee (approved):

PR #6640, opened 2026-08-25
Fix leaked identifiers in H5DSattach_scale()

### LINKED ISSUE #6639: H5DSattach_scale() leaks a dataset ID per existing reference, keeping the file open (pass-through VOL only)
### Summary

`H5DSattach_scale()` leaks one dataset identifier per reference already stored
in the scale's `REFERENCE_LIST` attribute. Each leaked identifier keeps its file
open, so a later `H5Fcreate(..., H5F_ACC_TRUNC, ...)` on the same file fails
with *"unable to truncate a file which is already open"* — an error with nothing
in it to point back at a dimension scale.

Only reachable through a pass-through VOL connector (see "Why it is normally
invisible" below).

### The code

`hl/src/H5DS.c`, in the loop that recreates the references read from the
existing `REFERENCE_LIST`:

```c
for (j = 0; j < nelmts - 1; j++) {
    if (is_new_ref) {
        ndsbuf_w[j].dim_idx = ndsbuf[j].dim_idx;
        tmp_id              = H5Ropen_object(&ndsbuf[j].ref, H5P_DEFAULT, H5P_DEFAULT);
        if (tmp_id < 0)
            goto out;
        if (H5Rcreate_object(tmp_id, ".", H5P_DEFAULT, &ndsbuf_w[j].ref) < 0) {
            H5Dclose(tmp_id);        /* <-- error path closes it */
            goto out;
        }
                                     /* <-- success path does not */
    }
    else {
        dsbuf_w[j] = dsbuf[j];
    }
}
```

`tmp_id` is closed only when `H5Rcreate_object()` fails. On success it is
dropped. The identical loop in `H5DSdetach_scale()` has always had the closing
`H5Dclose(tmp_id)` after the `H5Rcreate_object()` call, so this looks like a
straightforward omission rather than a deliberate difference.

### Why it is normally invisible

`H5DSwith_new_ref()` sets

```c
*with_new_ref = (config_flag || !native);
```

so with the native VOL, and without `H5_DIMENSION_SCALES_WITH_NEW_REF`,
`is_new_ref` is false and the `else` branch runs — which opens nothing. The
leaking branch is reached only by a **pass-through VOL connector**, where
`H5VLobject_is_native()` reports false.

### Reproducer

Build against `develop` and run under any pass-through connector. This uses the
library's own `pass_through`, forced non-native by hand; in the original report
it was a third-party pass-through connector.

```c
#include <hdf5.h>
#include <H5DSpublic.h>
#include <stdio.h>
#define F "rep.h5"

static void rep(const char *t) {
    printf("%-22s file ids = %zd, dataset ids = %zd\n", t,
           (ssize_t)H5Fget_obj_count(H5F_OBJ_ALL, H5F_OBJ_FILE),
           (ssize_t)H5Fget_obj_count(H5F_OBJ_ALL, H5F_OBJ_DATASET));
}

int main(void) {
    hsize_t dsd[1] = {3}, dims[2] = {3, 2};
    hid_t fid = H5Fcreate(F, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
    hid_t gid = H5Gcreate1(fid, "g", 0);
    hid_t dss = H5Screate_simple(1, dsd, dsd);
    hid_t scale = H5Dcreate1(gid, "dimscale", H5T_NATIVE_INT, dss, H5P_DEFAULT);
    H5DSset_scale(scale, "name");

    hid_t s1 = H5Screate_simple(1, dims, dims);
    hid_t v1 = H5Dcreate1(gid, "var1", H5T_NATIVE_INT, s1, H5P_DEFAULT);
    H5DSattach_scale(v1, scale, 0);          /* REFERENCE_LIST is empty here */

    hid_t s3 = H5Screate_simple(2, dims, dims);
    hid_t v3 = H5Dcreate1(gid, "var3", H5T_NATIVE_INT, s3, H5P_DEFAULT);
    H5DSattach_scale(v3, scale, 0);          /* one existing reference -> leak */
    rep("after 2nd attach");

    H5Dclose(scale); H5Dclose(v1); H5Dclose(v3);
    H5Sclose(dss); H5Sclose(s1); H5Sclose(s3); H5Gclose(gid); H5Fclose(fid);
    rep("after H5Fclose");

    fid = H5Fcreate(F, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
    printf("recreate = %lld\n", (long long)fid);
    return fid < 0;
}
```

Native VOL:

```
after 2nd attach       file ids = 1, dataset ids = 4
after H5Fclose         file ids = 0, dataset ids = 0
recreate = 72057594037927937
```

Pass-through connector:

```
after 2nd attach       file ids = 1, dataset ids = 5      <-- one extra
after H5Fclose         file ids = 1, dataset ids = 0      <-- file still open
HDF5-DIAG: Error detected in HDF5 (2.3.0):
  #003: H5Fint.c line 1949 in H5F_open(): unable to truncate a file which is already open
recreate = -1
```

The extra dataset identifier is `tmp_id`; the still-open file identifier is what
it holds.

### How it was found

Running the netCDF-C test suite over a pass-through VOL connector.
`h5_test/tst_h_dimscales` builds a dimension-scale file, reads it back, then
re-creates the same file with `H5F_ACC_TRUNC` and fails there.

### Fix

Close `tmp_id` on the success path too, as `H5DSdetach_scale()` does. PR to
follow.

hyoklee:
## Describe your changes

`H5DSattach_scale()` rewrites the scale's `REFERENCE_LIST` attribute whenever a
dataset is attached to a scale that already has one, and that rewrite acquired
three things it never released:

```c
tmp_id = H5Ropen_object(&ndsbuf[j].ref, H5P_DEFAULT, H5P_DEFAULT);
if (tmp_id < 0)
    goto out;
if (H5Rcreate_object(tmp_id, ".", H5P_DEFAULT, &ndsbuf_w[j].ref) < 0) {
    H5Dclose(tmp_id);      /* error path closed it */
    goto out;
}
                           /* success path did not */
```

1. the reopened dataset identifier above -- one per reference already attached
   to the scale;
2. the references read from the old attribute into `ndsbuf`, which were
   `free()`d without ever being destroyed;
3. the reference appended at index `nelmts - 1` of `ndsbuf_w`, because that
   buffer was reclaimed with `sid`, the *old* dataspace, which is one element
   shorter than the buffer.

Each of these keeps the file open, which is the symptom in the issue: a later
`H5Fcreate()` with `H5F_ACC_TRUNC` on the same file fails with *"unable to
truncate a file which is already open"*, an error with nothing in it to point
back at a dimension scale. `H5DSdetach_scale()` closes the identifier and
reclaims both buffers; attach now does the same.

### Reachability

The rewrite loop runs only on the new-style reference path, which
`H5DSwith_new_ref()` selects when the library is built with
`H5_DIMENSION_SCALES_WITH_NEW_REF` or when the object's **terminal** VOL
connector is not the native one. Stacking a pass-through connector over the
native connector does *not* reach it -- `H5VLobject_is_native()` compares the
terminal connector class, so pass-through-over-native still counts as native.
An earlier revision of this PR said "only a pass-through VOL connector"; the
comments and the `H5DSwith_new_ref()` documentation are corrected accordingly.

### Test

`test_attach_scale_id_leak()` in `hl/test/test_ds.c` attaches one scale to four
datasets, then checks that the scale is the only dataset identifier still open
and that the file can be truncated after being closed. It skips itself on the
old-reference path, where the loop opens nothing. Reverting either fix fails
it:

| reverted | result |
| --- | --- |
| `H5Dclose(tmp_id)` | `7 dataset ID(s) open after attaching, expected 1` |
| the `H5Treclaim()` calls | the truncate fails |
| nothing | passes |

All 133 tests pass in a `-DHDF5_DIMENSION_SCALES_NEW_REF=ON` build; `HL_*` and
the `h5dump`/`h5ls`/`h5repack` tool tests pass in a default build.

### CMake

`-DHDF5_DIMENSION_SCALES_NEW_REF=ON` was a no-op, so the code path above could
not be exercised from a stock build at all. The option was declared in
`hl/CMakeLists.txt`, a child scope that the top-level `H5pubconf.h`
`configure_file()` never sees, so `H5_DIMENSION_SCALES_WITH_NEW_REF` was never
defined and `libhdf5.settings` reported nothing for it. It now lives in
`CMakeBuildOptions.cmake`, which is included before both `src/` and `hl/`.

Fixes #6639

### DIFF
--- a/CMakeBuildOptions.cmake
+++ b/CMakeBuildOptions.cmake
@@ -83,6 +83,17 @@ mark_as_advanced (HDF5_ENABLE_PLUGIN_SUPPORT)
 
 option (HDF5_BUILD_HL_LIB "Build HIGH Level HDF5 Library" ON)
 
+# Must be set here, not in hl/, so that it reaches the H5pubconf.h and
+# build-settings configure_file() calls
+option (HDF5_DIMENSION_SCALES_NEW_REF "Use new-style references with dimension scale APIs" OFF)
+mark_as_advanced (HDF5_DIMENSION_SCALES_NEW_REF)
+if (HDF5_DIMENSION_SCALES_NEW_REF)
+  set (H5_DIMENSION_SCALES_WITH_NEW_REF 1)
+  set (DIMENSION_SCALES_WITH_NEW_REF "yes")
+else ()
+  set (DIMENSION_SCALES_WITH_NEW_REF "no")
+endif ()
+
 option (HDF5_BUILD_FORTRAN "Build FORTRAN support" OFF)
 
 option (HDF5_BUILD_CPP_LIB "Build HDF5 C++ Library" OFF)
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -146,6 +146,23 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## High-Level Library
 
+### Fixed a leaked dataset ID in H5DSattach_scale()
+
+   When attaching a dimension scale that already had one or more datasets
+   attached to it, `H5DSattach_scale()` reopened each reference stored in the
+   scale's `REFERENCE_LIST` attribute with `H5Ropen_object()` but only closed
+   the resulting identifier on the error path. One dataset identifier was
+   leaked per reference already attached, and each leaked identifier kept its
+   file open, so a later `H5Fcreate()` with `H5F_ACC_TRUNC` on the same file
+   failed with "unable to truncate a file which is already open".
+
+   The identifier is now closed on the success path as well, matching the
+   equivalent loop in `H5DSdetach_scale()`. The leak was only reachable through
+   a pass-through VOL connector, since `H5DSwith_new_ref()` selects the
+   old-style reference path -- which opens nothing -- whenever the object is
+   native and the library was not built with
+   `H5_DIMENSION_SCALES_WITH_NEW_REF`.
+
 ## Fortran High-Level APIs
 
 ## Documentation
--- a/release_docs/CHANGELOG.md
+++ b/release_docs/CHANGELOG.md
@@ -146,22 +146,28 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## High-Level Library
 
-### Fixed a leaked dataset ID in H5DSattach_scale()
+### Fixed leaked identifiers in H5DSattach_scale()
 
    When attaching a dimension scale that already had one or more datasets
-   attached to it, `H5DSattach_scale()` reopened each reference stored in the
-   scale's `REFERENCE_LIST` attribute with `H5Ropen_object()` but only closed
-   the resulting identifier on the error path. One dataset identifier was
-   leaked per reference already attached, and each leaked identifier kept its
-   file open, so a later `H5Fcreate()` with `H5F_ACC_TRUNC` on the same file
-   failed with "unable to truncate a file which is already open".
+   attached to it, `H5DSattach_scale()` rewrote the scale's `REFERENCE_LIST`
+   attribute without releasing everything it had acquired to do so. It reopened
+   each reference in the existing list with `H5Ropen_object()` but only closed
+   the resulting identifier on the error path, and it destroyed neither the
+   references it read from the old attribute nor the one it appended to the new
+   one -- the buffer being written was reclaimed with the old, one element
+   shorter, dataspace. Every one of those kept the file open, so a later
+   `H5Fcreate()` with `H5F_ACC_TRUNC` on the same file failed with "unable to
+   truncate a file which is already open", an error with nothing in it to point
+   back at a dimension scale.
 
-   The identifier is now closed on the success path as well, matching the
-   equivalent loop in `H5DSdetach_scale()`. The leak was only reachable through
-   a pass-through VOL connector, since `H5DSwith_new_ref()` selects the
-   old-style reference path -- which opens nothing -- whenever the object is
-   native and the library was not built with
-   `H5_DIMENSION_SCALES_WITH_NEW_REF`.
+   The identifier is now closed on the success path as well, and both reference
+   buffers are reclaimed, matching the equivalent code in
+   `H5DSdetach_scale()`. This code is only reached when `H5DSwith_new_ref()`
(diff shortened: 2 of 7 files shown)


CMakeBuildOptions.cmake
```diff
@@ -83,6 +83,17 @@ mark_as_advanced (HDF5_ENABLE_PLUGIN_SUPPORT)
 
 option (HDF5_BUILD_HL_LIB "Build HIGH Level HDF5 Library" ON)
 
+# Must be set here, not in hl/, so that it reaches the H5pubconf.h and
+# build-settings configure_file() calls
+option (HDF5_DIMENSION_SCALES_NEW_REF "Use new-style references with dimension scale APIs" OFF)
+mark_as_advanced (HDF5_DIMENSION_SCALES_NEW_REF)
+if (HDF5_DIMENSION_SCALES_NEW_REF)
```
jhendersonHDF, line 90: This part and below should be moved back to hl/ so that only the `option ()` and `mark_as_advanced ()` are in this file. The changes in https://github.com/HDFGroup/hdf5/pull/6617 fix the configuration file so that it uses the option value and doesn't need the variables below. In fact, I think setting `DIMENSION_SCALES_WITH_NEW_REF` can just be removed. I just need to update the usage in H5build_settings.cmake.c.in.

release_docs/CHANGELOG.md
```diff
@@ -146,6 +146,29 @@ We would like to thank the many HDF5 community members who contributed to this r
...
+   truncate a file which is already open", an error with nothing in it to point
+   back at a dimension scale.
+
+   The identifier is now closed on the success path as well, and both reference
+   buffers are reclaimed, matching the equivalent code in
+   `H5DSdetach_scale()`. This code is only reached when `H5DSwith_new_ref()`
+   selects the new-style reference path, which happens when the object's
+   terminal VOL connector is not the native one -- a pass-through connector
+   stacked over the native connector does not qualify -- or when the library is
+   built with `H5_DIMENSION_SCALES_WITH_NEW_REF`; the old-style path opens
```
lrknox, line 169: Does old-style refer to references here?  If so I think including reference in "the old-style reference path" will be worth the extra characters.

PR #6645, opened 2026-08-27
Implement internally concurrent multithreading for chunk dataset I/O reads

fortnern:
If the library is compiled with HDF5_HAVE_CONCURRENCY On and a global thread pool is created (H5TSglobal_pool_create()), the library will use threads in the thread pool to concurrently process all chunks in the operation - read from disk, filter, and scatter to memory. Currently all these sub-operations are protected by a mutex so there is no expected performance gain. Currently only supports read operations. There are at least two bugs that remain - one is related to the error code's use of free lists, the other is so far unresolved.


src/H5Dpkg.h
```diff
@@ -267,18 +267,14 @@ typedef struct {
...
 typedef struct {
     void *buf;   /* Buffer for compact dataset */
     bool *dirty; /* Pointer to dirty flag to mark */
 } H5D_compact_storage_t;
 
 typedef union H5D_storage_t {
-    H5D_contig_storage_t  contig;  /* Contiguous information for dataset */
-    H5D_chunk_storage_t   chunk;   /* Chunk information for dataset */
+    H5D_contig_storage_t contig; /* Contiguous information for dataset */
+    /* No global storage information for a chunked dataset for this union */
```
fortnern, line 277: This only stored the scaled coordinates of a chunk, and this union is supposed to be about the entire dataset, so this information is really misplaced here. Eliminated this from the few places that used it and instead passed the scaled coordinates (which were always available separately) directly.

gheber:
- The table entry for `HDF5_ENABLE_CONCURRENCY` in `INSTALL_CMake_options.md` needs updating. It reads *... NOTE: Currently non-functional and experimental.*
- The `H5TS_pool_create` public contract is incomplete: `num_threads` must exceed zero, but that is undocumented; zero is rejected in `H5TSpool.c`. Worker-thread error-stack behavior and the current lack of expected speedup should also appear on the `H5P` pages or in Known Problems.
- The documentation says the APIs exist only with concurrency enabled, but their public declarations are unconditional while their definitions are conditional.
- The module guide is too thin: `H5TSmodule.h` contains only a one-sentence overview. It should explain the create → read/configure → destroy workflow, eligibility, defaults, and limitations.

src/H5Epkg.h
```diff
@@ -52,6 +52,11 @@
 #define H5E__get_my_stack() (H5E_stack_g + 0)
 #endif /* H5_HAVE_THREADSAFE_API */
 
+/* Disable free lists in this package when concurrency is enabled until free lists are threadsafe */
+#ifdef H5_HAVE_CONCURRENCY
```
jhendersonHDF: This seems like it would be relying on the order of including `H5Epkg.h` vs. `H5FLprivate.h` and might be a bit fragile. Since only `H5E.c` and `H5Eint.c` are currently including `H5FLprivate.h`, could this block be moved to those two places next to the `#include` call to make that more explicit? Of course, there could also still be future issues if another separate inclusion in those files indirectly includes `H5FLprivate.h` before the direct inclusion, but that just seems to be a consequence of trying to disable free lists for one package in this way. Maybe it would be better to handle this at the CMake level by setting compile definitions for H5E*.c files when `HDF5_ENABLE_CONCURRENCY` is enabled?

src/H5TSpublic.h
```diff
@@ -0,0 +1,103 @@
...
+ *
+ *          The thread pool must not already exist. If this function is called
+ *          when the thread pool already exists, an error will be returned.
+ *
+ * \note    This function is only present when the library is compiled with HDF5_ENABLE_CONCURRENCY ON.
+ *
+ * \since 2.3.0
+ *
+ */
+H5_DLL herr_t H5TSglobal_pool_create(unsigned num_threads);
```
jhendersonHDF: A couple of questions and observations on this functionality, mostly to go over pros/cons of a global thread pool:

- Are there any future plans to allow users to manipulate this thread pool, such as adding their own user-defined threads to it? With these APIs we're introducing the ability for users to directly manipulate internal global state used by the library, without much information on when it's appropriate to do so or stop doing so and where those windows of time could change as the thread pool is expanded to be used by other concurrency features. That seems like a bit of a programming model headache to me, especially when it involves threading with anything global.
- A global thread pool with a fixed number of worker threads also seems a bit inflexible for cases where different numbers of threads are desired for different tasks, where those tasks could use separate thread pools that don't interfere with each other. These functions could act as just wrappers around H5TS calls and return a handle to be passed in to other functions, though I think that still complicates the programming model a bit much.
- If performance of creating a thread pool during something like file open is poor, it could still be performed internally on the first file open and tracked for later use, then cleaned up by the library as needed. Explicit control over when those resources are freed is of course lost, but could be traded off for a request to free up those resources at a time the library knows it's safe to do so. That at least doesn't place the responsibility on users.

release_docs/CHANGELOG.md
```diff
@@ -54,6 +54,18 @@ We would like to thank the many HDF5 community members who contributed to this r
 
 ## Library
 
+### Added support for internally concurrent multithreaded reads of chunked datasets
+
+   Also added 4 new functions to support this: H5TSglobal_pool_create(),
+   H5TSglobal_pool_destroy(), H5Pset_io_threads(), and H5Pget_io_threads(). When
+   the library is configured with HDF5_ENABLE_CONCURRENCY ON, the global thread
```
hyoklee: HDF5_ENABLE_CONCURRENCY=ON

fortnern:
> * The table entry for `HDF5_ENABLE_CONCURRENCY` in `INSTALL_CMake_options.md` needs updating. It reads _... NOTE: Currently non-functional and experimental._

Will update

>     * The `H5TS_pool_create` public contract is incomplete: `num_threads` must exceed zero, but that is undocumented; zero is rejected in `H5TSpool.c`. Worker-thread error-stack behavior and the current lack of expected speedup should also appear on the `H5P` pages or in Known Problems.

I'm thinking I'll change the function to allow 0 (has no effect), as well as allow use with an already existing pool (in which case it destroys it first), and allow the pool destroy function to not fail if the pool doesn't exists (in which case nothing happens). Does this sound OK?

I'll add a note about the worker thread error stack behaviour, though the lack of expected speedup will be changed before the release so I don't think that needs to be added.

>     * The documentation says the APIs exist only with concurrency enabled, but their public declarations are unconditional while their definitions are conditional.

This follows the precedent of the publicly-accessible functions in H5TSdevelop.h (with respect to H5_HAVE_THREADSAFE_API in that case). I'm open to changing it but we should think about it first.

>     * The module guide is too thin: `H5TSmodule.h` contains only a one-sentence overview. It should explain the create → read/configure → destroy workflow, eligibility, defaults, and limitations.

H5TSpublic.h will likely contain things that aren't related to the thread pool at some point, so I don't think this is the right place to add this information. I can add more information to the RM entry for pool_create. If we want a high level guide to internal multithreading I'd think that should be in the User's Guide.

hyoklee (dismissed):
standard CMake flag representation

gheber (dismissed):
👍

src/H5Dpkg.h
jhendersonHDF: It's unrelated to the changes in this PR so I wouldn't expect it to be addressed here, but doesn't that also imply that the `contig` and `compact` fields should get the same treatment?

src/H5Ppublic.h
```diff
@@ -8800,6 +8800,52 @@ H5_DLL herr_t H5Pset_modify_write_buf(hid_t plist_id, bool modify_write_buf);
...
+ * accelerate I/O. By default this is set to true. However, the global thread pool must be created using
+ * H5TSglobal_pool_create() before the library can use threads to accelerate I/O. H5Pset_io_threads() can be
+ * used to disable this acceleration even if the global thread pool exists.
+ *
+ * \note    This function is only present when the library is compiled with HDF5_ENABLE_CONCURRENCY ON.
+ *
+ * \since 2.3.0
+ *
+ */
+H5_DLL herr_t H5Pset_io_threads(hid_t plist_id, bool io_threads_enabled);
```
jhendersonHDF, line 8825: Related to my comments in `H5TSpublic.h`, to me this also makes a case for setting up the global thread pool automatically and _maybe_ offering a call to explicitly cleanup thread resources at some point like file close (which could probably also be done with a property). It's probably just a difference of opinion, but I would see creating an internal thread pool for file operations as being the default, with a property to allow setting up and using a global thread pool if needed.

src/H5TSprivate.h
```diff
@@ -293,6 +294,12 @@ typedef atomic_flag H5TS_spinlock_t;
 /* Library-private Variables */
 /*****************************/
 
+/* Global thread pool */
+extern H5TS_pool_t *H5TS_pool_g;
+
+/* Whether there are concurrent threads in the library (from internal spawning) */
+extern bool H5TS_currently_concurrent_g;
```
jhendersonHDF, line 302: #ifdef H5_HAVE_CONCURRENCY here

src/H5Z.c
```diff
@@ -1409,6 +1412,17 @@ H5Z_pipeline(const H5O_pline_t *pline, unsigned flags, unsigned *filter_mask /*i
                 continue; /* filter excluded */
             }
 
+#ifdef H5_UNSAFE_CONCURRENCY /* We currently take this lock before entering H5Z_pipeline(), and this is not a recursive lock, so don't take it again here */
+#ifdef H5_HAVE_CONCURRENCY
+            /* If we're using concurrent threads, lock the internal mutex to search for the plugin */
+            if (H5TS_currently_concurrent_g) {
```
jhendersonHDF, line 1418: What about the call to `H5PL_load` that might come from `H5Z_filter_avail()`?

jhendersonHDF, line 1418: It also seems like this could be a point of a lot of locking and might be able to be reworked a little so that locking only has to be done in the case where we're actually going to load a plugin.

src/H5TSprivate.h
```diff
@@ -316,6 +323,13 @@ H5_DLL herr_t H5TS_api_lock(unsigned *dlftt);
 #endif
 H5_DLL herr_t H5TS_api_unlock(void);
 
+/* Internal locking */
+#ifdef H5_HAVE_CONCURRENCY
+H5_DLL herr_t H5TS_internal_lock(void);
+H5_DLL herr_t H5TS_internal_unlock(void);
+H5_DLL bool   H5TS_internal_locked(void);
```
jhendersonHDF: The current usages of `H5TS_internal_lock()` / `H5TS_internal_unlock()` seem to just make use of a `bool` for the locked status. Is there a case for `H5TS_internal_locked()` anywhere?

Maybe a better question is what is the purpose of this lock going forward? I think that should be documented somehow so that it's not misused in the future. Things like should the lock be restricted to a small section of code where a `bool` is fine for the locked/unlocked status or could it be carried across modules and then who is responsible for managing it?

src/H5TSint.c
```diff
@@ -446,6 +453,81 @@ H5TS_api_unlock(void)
...
+herr_t
+H5TS_internal_unlock(void)
+{
+    herr_t ret_value = SUCCEED;
+
+    FUNC_ENTER_NOAPI_NAMECHECK_ONLY
+
+    /* Acquire the library's internal lock */
+    if (H5_UNLIKELY(H5TS_mutex_unlock(&H5TS_api_info_p.internal_mutex) < 0))
+        HGOTO_DONE(FAIL);
```
jhendersonHDF, line 509: I'm probably just missing something obvious, but is there a `H5TS_api_info_p.internal_mutex_locked = false;` needed? If so, maybe `assert()`s on the status of the lock pre-lock and pre-unlock would be useful.

jhendersonHDF (commented):
Leaving some comments early while I review H5Dchunk.c

src/H5TSpublic.h
fortnern: > A couple of questions and observations on this functionality, mostly to go over pros/cons of a global thread pool:
> 
>     * Are there any future plans to allow users to manipulate this thread pool, such as adding their own user-defined threads to it? With these APIs we're introducing the ability for users to directly manipulate internal global state used by the library, without much information on when it's appropriate to do so or stop doing so and where those windows of time could change as the thread pool is expanded to be used by other concurrency features. That seems like a bit of a programming model headache to me, especially when it involves threading with anything global.

No concrete plans to allow further control over the thread pool. We might allow per-operation thread limits, or add a dynamic sizing option at some point. This is meant to be similar to a typical cache size setting, letting the library know how many threads (instead of memory in the analogy) it's allowed to use.

>     * A global thread pool with a fixed number of worker threads also seems a bit inflexible for cases where different numbers of threads are desired for different tasks, where those tasks could use separate thread pools that don't interfere with each other. These functions could act as just wrappers around H5TS calls and return a handle to be passed in to other functions, though I think that still complicates the programming model a bit much.

See above.

>     * If performance of creating a thread pool during something like file open is poor, it could still be performed internally on the first file open and tracked for later use, then cleaned up by the library as needed. Explicit control over when those resources are freed is of course lost, but could be traded off for a request to free up those resources at a time the library knows it's safe to do so. That at least doesn't place the responsibility on users.

We could do that if we want to automatically create a thread pool, or just at library startup.

src/H5Dpkg.h
fortnern: The contig and compact fields store data that's global to the dataset, so they won't cause the same type of problem I saw with chunks. I haven't looked into whether they can be eliminated to streamline the code however.

src/H5Z.c
fortnern: I don't see how that function could be reached inside the threaded section (unless the filter callback does something really weird, in which case it shouldn't be marked as threadsafe once we add that option).

fortnern: I agree I'd like to push it down, but it's disabled for now anyways and we'll work on it once we start properly working on threadsafe H5Z. My original prototype had a fancier setup in H5PL_load() but that didn't work so I just did this for now.

src/H5TSprivate.h
fortnern: I don't recall why I added H5TS_internal_locked(). It may be necessary if at some point code in the library can be reached either from code inside or outside the lock. But then a recursive mutex may make more sense. I'll go ahead and remove it. As you point out in the other comment it's broken anyways,

src/H5Z.c
jhendersonHDF: I may have missed it, but if it doesn't exist it could be useful to have `#undef H5_UNSAFE_CONCURRENCY` or `/* #define H5_UNSAFE_CONCURRENCY */` somewhere with a comment so it can be enabled quickly. That or a CMake option to do that.

jhendersonHDF: I was thinking cases like a filter that could call out to others and might use `H5Zfilter_avail()`. Maybe that's not a problem in practice currently?

src/H5TSprivate.h
fortnern: The purpose of the lock is documented in the function description of H5TS_internal_lock().

src/H5Dchunk.c
```diff
@@ -3187,14 +3360,75 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
...
+                    HGOTO_ERROR(H5E_DATASET, H5E_CANTFREE, FAIL, "can't destroy semaphore");
+                threaded_io_info->sem_init = false;
+
+                /* Check for failure */
+                if (threaded_io_info->failed)
+                    HGOTO_ERROR(H5E_DATASET, H5E_READERROR, FAIL, "threaded read worker(s) failed");
+            }
+
+            /* Free chunk_info array */
+            H5MM_free(threaded_io_info->chunk_info);
```
jhendersonHDF, line 3432: Maybe worth setting this to NULL after free in case other error checking is added in the future

src/H5Dchunk.c
```diff
@@ -4598,7 +5026,12 @@ H5D__chunk_cache_prune(const H5D_t *dset, size_t size)
  */
 static void *
 H5D__chunk_lock(const H5D_io_info_t H5_ATTR_NDEBUG_UNUSED *io_info, const H5D_dset_io_info_t *dset_info,
-                H5D_chunk_ud_t *udata, bool relax, bool prev_unfilt_chunk)
+                H5D_chunk_ud_t *udata, bool relax, bool prev_unfilt_chunk
+#ifdef H5_HAVE_CONCURRENCY
+                ,
+                H5D_threaded_io_info_t *threaded_io_info, bool *threaded_chunk
```
jhendersonHDF, line 5068: Minor nitpick really, but it looks like the `threaded_io_info` structure is only passed in for access to that chunk's `old_pline`, `chunk_nbytes` and `buf_alloc` fields which would all be available types even if `H5_HAVE_CONCURRENCY` isn't defined. If you had a structure for those that you could optionally pass a pointer to you could avoid the awkward `H5D_CHUNK_LOCK_NO_THREADING_PARAMS` approach. I just tend to find it a little hard to read since I might have to keep the macro definition nearby.

release_docs/CHANGELOG.md
```diff
@@ -56,14 +56,13 @@ We would like to thank the many HDF5 community members who contributed to this r
...
-   pool exists (created by H5TSglobal_pool_create()), and I/O threads are not
-   disabled (using H5Pset_io_threads()), the library will use the threads in the
-   thread pool to concurrently read from disk, unfilter, and scatter to memory
-   all chunks in a read operation on a chunked dataset. Currently each of these
-   sub-operations is serialized (protected by a mutex) so there is not yet
+   Also added 3 new functions to support this: H5TSset_internal_threads(),
+   H5Pset_io_threads(), and H5Pget_io_threads(). When the library is configured
+   with HDF5_ENABLE_CONCURRENCY=ON, the number of internal threads is greater
+   than zero (set by H5TSset_internal_threads()), and I/O threads are not
+   disabled (using H concurrently read from disk, unfilter, and scatter to
```
gheber: The text looks garbled, and I can't follow the sentence.

src/H5Ppublic.h
```diff
@@ -8804,17 +8804,18 @@ H5_DLL herr_t H5Pget_modify_write_buf(hid_t plist_id, bool *modify_write_buf);
...
  *
  * \return \herr_t
  *
- * \details H5Pset_io_threads() sets whether the library is allowed to use the global thread pool to
- * accelerate I/O. By default this is set to true. However, the global thread pool must be created using
- * H5TSglobal_pool_create() before the library can use threads to accelerate I/O. H5Pset_io_threads() can be
- * used to disable this acceleration even if the global thread pool exists.
+ * \details H5Pset_io_threads() sets whether the library is allowed to use internal multithreading to
+ * accelerate I/O. By default this is set to true. However, internal threading must be enabled using
+ * H5TSset_internal_threads() before the library can use threads to accelerate I/O.
```
gheber, line 8816: What's the error code to indicate that threading was not enabled?

src/H5Ppublic.h
```diff
@@ -8830,14 +8831,15 @@ H5_DLL herr_t H5Pset_io_threads(hid_t plist_id, bool io_threads_enabled);
...
  *
- * \details H5Pget_io_threads() gets the "I/O threads enabled" property from the dataset transfer property
- * list \p plist_id. This property determines whether the library is allowed to use the global thread pool to
- * accelerate I/O. The default value for io_threads_enabled is true. However, the global thread pool must be
- * created using H5TSglobal_pool_create() before the library can use threads to accelerate I/O.
+ * \details H5Pget_io_threads() gets the "I/O threads enabled" property from the dataset transfer
+ * property list \p plist_id. This property determines whether the library is allowed to use internal
+ * multithreading to accelerate I/O. The default value for io_threads_enabled is true. However,
+ * internal threading must be enabled using H5TSset_internal_threads() before the library can use
+ * threads to accelerate I/O.
```
gheber, line 8842: Ditto. What's the error code to indicate that?

gheber (dismissed):
👍

src/H5Ppublic.h
fortnern: There's no error code

fortnern: Threading is enabled by default on H5Pset_io_threads(), so it wouldn't make sense to have it error if there are no threads to use.

src/H5Ppublic.h
fortnern: See above. This property is intended to be used to *disable* threading for specific operations. To enable threading all you need to do is call H5TSset_internal_threads()

src/H5Ppublic.h
gheber: So what does `herr_t` mean?

src/H5Z.c
fortnern: In terms of race conditions, that sort of thing will be tackled when we make H5Z properly threadsafe, though we may just ban calling back into HDF5 for threadsafe plugins at least initially. However, that type of thing might cause a deadlock even with non-threadsafe plugins, so we may need to change the internal lock to be recursive. Not an issue now though since the lock in H5Z is disabled and the locks in the worker function cannot be reached if H5TS_currently_concurrent_g is true.

fortnern: That's a possibility but I don't think it's necessary right now. We'll be working on opening up H5Z next, and it will be in before the release.

src/H5Dchunk.c
fortnern: We would also need to pass in threaded_chunk. We could do that but I'm trying to avoid adding unused variables and parameters to the code in the case concurrency is off. If the macro is a major problem I can replace it with #ifdef H5_HAVE_CONCURRENCY / #else #endif macros for every call to H5Dchunk_lock()

hyoklee (dismissed):

fortnern:
>     * The documentation says the APIs exist only with concurrency enabled, but their public declarations are unconditional while their definitions are conditional.

Actually Claude pointed out that other public symbols like the MPI functions are gated on #defines created by cmake, so it would be appropriate to add them here. However, doing so breaks doxygen. I am investigating.

src/H5Dchunk.c
```diff
@@ -3187,14 +3366,80 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
...
+                do {
+                    if (H5_UNLIKELY(H5TS_semaphore_wait(&threaded_io_info->sem) < 0))
+                        HGOTO_ERROR(H5E_DATASET, H5E_CANTWAIT, FAIL, "can't wait for worker thread");
+                    threads_in_flight--;
+                } while (threads_in_flight);
+
+                /* Mark that we are no longer concurrent */
+                H5TS_currently_concurrent_g = false;
+
+                /* Unlock all chunks */
```
mattjala, line 3444: The threaded version of H5D__chunk_lock() inserts the raw chunk buffer into the rdcc before the read/unfilter step, since both of those steps are now in H5D__chunk_thread_read.

The unlocking of the chunks (H5Dchunk.c:3407) is done before we check for threaded IO failure (H5Dchunk.c:3427). So on a failure, every chunk is unlocked while holding uninit/partially-unfiltered data which is marked as clean (since it was allocated with H5FL_CALLOC). The top level dataset read for that request will return a failure, but the next H5Dread will read from the cache and return garbage data.

This can't happen in the serial path since serial H5D__chunk_lock() frees and fails before insertion.

Failed chunks should probably be checked earlier and evicted rather than unlocked.

src/H5Dchunk.c
```diff
@@ -3187,14 +3366,80 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
...
         } /* end while */
-    }     /* end else */
 
-done:
-    /* Release chunk lock if we failed while holding it */
-    if (chunk_locked && chunk)
-        if (H5D__chunk_unlock(io_info, dset_info, &udata, false, chunk, src_accessed_bytes) < 0)
-            HDONE_ERROR(H5E_IO, H5E_CANTUNLOCK, FAIL, "unable to unlock raw data chunk");
+#ifdef H5_HAVE_CONCURRENCY
+        /* Handle chunks that were deferred for concurrent processing */
```
mattjala, line 3373: This loop locks every chunk for its full duration. H5D__chunk_cache_prune() skips entries that are locked, which means that nbytes_used can grow unbounded. An N-chunk read could hold up to N * chunk_size bytes in the cache at once, regardless of what the chunk cache is configured for with H5Pset_chunk_cache.

This might be addressible by doing the loop in batches to give some of the chunks a chance to unlock.

src/H5Dchunk.c
```diff
@@ -3211,16 +3456,237 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
...
+    /* Report failure to task invoker (actual return value is ignored by the thread pool) */
+    if (ret_value < 0) {
+        threaded_chunk_info->threaded_io_info->failed = true;
+        if (mutex_held && H5TS_internal_unlock() < 0)
+            HDONE_ERROR(H5E_DATASET, H5E_CANTUNLOCK, FAIL, "can't unlock internal mutex");
+        mutex_held = false;
+        if (api_ctx_pushed && H5CX_pop(false) < 0)
+            HDONE_ERROR(H5E_SYM, H5E_CANTRESET, FAIL, "can't reset API context");
+        api_ctx_pushed = false;
+        (void)H5E_dump_api_stack();
```
mattjala, line 3744: In H5D__chunk_thread_read()'s done block, the mutex is released before H5E_dump_api_stack / H5E_clear_stack are called, and the same pair at 3682-3676 runs after the mutex is asserted to not be held. So both of these cases can execute concurrently across workers.

H5E_dump_api_stack() calls the default handler H5E__print2, which does H5I_object_verify() at H5Eint.c:878 and three more in H5E__walk2_cb (1178, 1179, 1192). H5I has no internal locking, so if two workers fail at once, then they'd have unsafe concurrent access to the global H5I index.

Additionally, H5E_clear_stack() → H5E__clear_entries() calls H5I_dec_ref() (H5Eint.c:1803, 1806, 1809), which writes to the H5I index. These are guarded to skip library-internal error IDs, so they only fire for application-registered ones. This is more narrow than the case above, but still potentially unsafe.

It looks like both of these error-handling calls should be inside the mutex.

src/H5Dchunk.c
fortnern: H5D__chunk_cache_prune() is called after the unlock loop, so the chunk cache should be the correct size after the operation. It's unavoidable that we'll need to have a buffer open for every chunk in the concurrent operation during the operation if we want it to be concurrent. I can add a note about memory usage to the docs for H5TSset_internal_threads().

src/H5Dchunk.c
fortnern: It's not currently possible to set a non-default error handler for the worker threads, but I can add a mutex.

src/H5Dchunk.c
```diff
@@ -3211,16 +3456,237 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
...
+        (void)H5E_clear_stack();
+    }
+
+    assert(!mutex_held);
+    assert(!api_ctx_pushed);
+
+    /* Signal the semaphore to indicate that we've finished processing one chunk */
+    if (H5_UNLIKELY(H5TS_semaphore_signal(&threaded_chunk_info->threaded_io_info->sem) < 0)) {
```
mattjala: H5D__chunk_read waits for N semaphore signals (H5Dchunk.c:3396-3402), then destroys the semaphore at 3422 and frees the parent struct at 3436.

The wait loop returns as soon as the Nth signal is observed, which is slightly distinct from the Nth worker having finished executing H5TS_semaphore_signal(). With [glibc ≥ 2.21](https://sourceware.org/bugzilla/show_bug.cgi?id=12674) this is fine, but on older versions (still used by RHEL/CentOS 6 (2.12) and 7 (2.17)) sem_post touches the semaphore after a waiter has returned from sem_wait, which is a use-after-free.

We might want to set a glibc floor for HDF5_ENABLE_CONCURRENCY, or else wait on worker completion directly via an atomic completion counter.

test/dsets.c
```diff
@@ -19408,6 +19428,222 @@ test_filter_error_msg(hid_t file)
...
+    cdims[1] = THREADED_CHUNKS_CDIM2;
+    if (H5Pset_chunk(dcpl, 2, cdims) < 0)
+        TEST_ERROR;
+
+    /* Disable chunk cache on dapl */
+    if (H5Pset_chunk_cache(dapl, 0, 0, 0.) < 0)
+        TEST_ERROR;
+
+    /* Create the dataset */
+    if ((dataset = H5Dcreate2(file, "dset", H5T_NATIVE_INT, space, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT)) <
```
mattjala: test_threaded_chunks() creates its first dataset with H5P_DEFAULT in the DCPL slot at test/dsets.c:19488, so "dset" is contiguous instead of chunked.

As such, the entire test matrix here never enters H5D__chunk_read. `dcpl` should be passed as the 6th argument at 19488.

src/H5Dchunk.c
fortnern: That's a pretty obscure find, thanks! I'll probably change the semaphore to be a global that's initialized when the global pool is created and destroyed after the global pool is destroyed.

fortnern: The pthread_joins in the pool destroy should prevent this from being an issue (as it does for the semaphore used for the pool)

fortnern: Actually I'll go ahead and implement an atomic uint and cond var, since the global semaphore may be a thread safety liability in the future when more concurrency is added to the library

fortnern:
@mattjala I believe I've addressed all of your comments

mattjala (dismissed):

mattjala (dismissed):

hyoklee (dismissed):

src/H5Ppublic.h
fortnern: It returns a generic error value if something went wrong. This is consistent with most of the other API functions, and I believe the documentation on the return value is consistent with that of other functions as well.

fortnern: It does not return an error if threads are enabled by H5Pset_io_threads but disabled by H5TSset_internal_threads. It simply does not enabled threads.

lrknox:
I put this in a reply to Gerd's comment to CHANGELOG.md, which you've addressed.  I still find the sentence beginning with "When the library" to be confusing, so I asked Claude to break it up and added a couple of small tweaks of my own.  Consider this for the CHANGELOG.md entry:

### Added support for internally concurrent multithreaded reads of chunked datasets

   Added 3 new functions to support this: H5TSset_internal_threads(), H5Pset_io_threads(), and H5Pget_io_threads().
  
   Concurrent multithreaded reads of chunked datasets require three things: the library must be configured with HDF5_ENABLE_CONCURRENCY=ON, the number of internal threads must be set greater than zero via H5TSset_internal_threads(), and I/O threads must not be disabled via H5Pset_io_threads(). When all three conditions are met, the library will concurrently read from disk, unfilter, and scatter to memory all chunks in a read operation on a chunked dataset. Currently each of these sub-operations is serialized (protected by a mutex), so there is not yet likely to be any performance improvement.

lrknox:
I think prompted by Gerd's other comments/questions about error codes from H5Pget_io_threads and H5Pget_io_threads, it may be helpful to the caller if those functions would also output a message about the status of the number of internal threads, or at least a message about needing to set some number of internal threads > 0 if the call was  H5Pset_io_threads(true) when it is already true (the default).  A similar indication when calling  H5Pget_io_threads could be useful.  Alternatively a usage example for these functions, especially since the main usage of H5Pset_io_threads() is to turn it off for a specific operation when the number of internal threads is already >0.  

That's in the doxygen documentation if the user reads all the way to the end and understands that whether setting the library to be allowed to use internal multithreading for concurrent multithreaded reads of chunked datasets is dependent on the other 2 settings.

fortnern:
There are actually more conditions to enable the feature, and I don't think we need to list all of them here. How about:

Added 3 new functions to support this: H5TSset_internal_threads(), H5Pset_io_threads(), and H5Pget_io_threads().

This feature internally parallelizes read operations on chunked datasets. H5TSset_internal_threads() is used to enable the feature globally, while H5Pset_io_threads() can be used to disable the feature on a per-operation basis. These functions are only available when the library is configured with HDF5_ENABLE_CONCURRENCY=ON. When performing an internally threaded read, the library will concurrently read from disk, unfilter, and scatter to memory all chunks in a read operation on a chunked dataset. Currently each of these sub-operations is serialized (protected by a mutex), so there is not yet likely to be any performance improvement.

lrknox:
Yes, I think that is much easier to understand.

lrknox (approved):

mattjala (approved):

PR #6646, opened 2026-08-27
Add parallel example for HDF5 on an MPI_Comm_split sub-communicator

brtnfld:
Adds `HDF5Examples/C/H5PAR/ph5_comm_split.c`, an example of using parallel
HDF5 on a communicator produced by `MPI_Comm_split` rather than on
`MPI_COMM_WORLD`.

Applications often divide `MPI_COMM_WORLD` into groups that do different
work, where only some of those groups touch the HDF5 file. None of the
existing H5PAR examples show that arrangement; they all open the file on
`MPI_COMM_WORLD`.

### DIFF
--- a/HDF5Examples/C/H5PAR/ph5_comm_split.c
+++ b/HDF5Examples/C/H5PAR/ph5_comm_split.c
@@ -1,0 +1,365 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * Copyright by The HDF Group.                                               *
+ * All rights reserved.                                                      *
+ *                                                                           *
+ * This file is part of HDF5.  The full HDF5 copyright notice, including     *
+ * terms governing use, modification, and redistribution, is contained in    *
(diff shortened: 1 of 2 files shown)


HDF5Examples/C/H5PAR/ph5_comm_split.c
```diff
@@ -0,0 +1,365 @@
...
+         * Writer group: reopen read-only on the same
+         * sub-comm and verify what was written
+         * ------------------------------------------------
+         *
+         * Reopening needs a fresh FAPL carrying the same communicator. A
+         * FAPL does not survive the file it was used to open, and the
+         * communicator must match the set of ranks that will make the
+         * collective calls on the reopened file.
+         */
+        fapl_id = make_fapl(sub_comm);
```
jhendersonHDF, line 307: Strictly speaking this shouldn't ever be necessary, but it's also not a terrible idea.

jhendersonHDF (approved):

PR #6649, opened 2026-08-28
Fix h5open_f failing to re-initialize the Fortran interface (#6642)

### LINKED ISSUE #6642: h5open_f silently fails to re-initialize the Fortran interface after h5close_f (prohibited argument aliasing in h5close_f)
## Disclaimer

This is my first bug report for hdf5. I use the library extensively but am not an expert on the code itself. I used Anthrophic Claude and a reproducer provided by a colleague to investigate and write up the issue.

## Summary

Since [PR #1657](https://github.com/HDFGroup/hdf5/pull/1657) ("HDFFV-11306 Fixed", merged 2022-04-19), the sequence `h5open_f` --> `h5close_f` --> `h5open_f` can leave the Fortran interface uninitialized. The second `h5open_f` returns `error = 0` but never calls `h5init_types_c`, so all predefined type handles (`H5T_NATIVE_INTEGER`, etc.) are stale and invalid. Every subsequent Fortran HDF5 call fails or misbehaves, with no error from `h5open_f` itself.

Whether this happens is compiler-dependent, because the underlying defect is prohibited argument aliasing. It reproduces with **Intel oneAPI 2024.2.0 (ifort)**; it does **not** reproduce with gfortran or with Intel oneAPI 2026.0.0 (ifx). Note that the relevant compiler is the one used to build `libhdf5_fortran`, not the one used to build the application.

Present in 1.13.2 through 1.14.6, 2.0.0, 2.1.0, and `develop` (checked at d8ec63966).

## Reproducer

```fortran
program test_hdf5_reopen

   use hdf5
   use h5f, only : H5OPEN_NUM_OBJ

   implicit none

   integer :: ierr
   logical :: valid

   write(*,'(a,x,i0)') 'INITIAL H5OPEN_NUM_OBJ:', H5OPEN_NUM_OBJ

   call h5open_f(ierr)

   write(*,'(13x,3(x,a24))') 'H5OPEN_NUM_OBJ', 'H5T_NATIVE_INTEGER', 'H5IIS_VALID_F'
   call h5iis_valid_f(H5T_NATIVE_INTEGER, valid, ierr)
   write(*,'(a13,2(x,i24),(23x,l))') 'AFTER OPEN:', H5OPEN_NUM_OBJ, H5T_NATIVE_INTEGER, valid

   call h5close_f(ierr)

   call h5iis_valid_f(H5T_NATIVE_INTEGER, valid, ierr)
   write(*,'(a13,2(x,i24),(23x,l))') 'AFTER CLOSE:', H5OPEN_NUM_OBJ, H5T_NATIVE_INTEGER, valid

   call h5open_f(ierr)

   call h5iis_valid_f(H5T_NATIVE_INTEGER, valid, ierr)
   write(*,'(a13,2(x,i24),(23x,l))') 'AFTER REOPEN:', H5OPEN_NUM_OBJ, H5T_NATIVE_INTEGER, valid
end program test_hdf5_reopen
```

Affected build (HDF5 1.14.5 built with Intel oneAPI 2024.2.0 ifort):

```
INITIAL H5OPEN_NUM_OBJ: 0
                        H5OPEN_NUM_OBJ       H5T_NATIVE_INTEGER            H5IIS_VALID_F
  AFTER OPEN:                       51       216172782113783911                        T
 AFTER CLOSE:                      -51       216172782113783911                        F
AFTER REOPEN:                      -51       216172782113783911                        F
```

`AFTER REOPEN` should show a valid type. `H5OPEN_NUM_OBJ` should be `0` after close, not `-51`.

Unaffected build (same 1.14.5 sources, gfortran and ifx):

```
  AFTER OPEN:                       51       216172782113783912                       T
 AFTER CLOSE:                        0       216172782113783912                       F
AFTER REOPEN:                       51       216172782113783963                       T
```

## Cause

`h5close_f` (`fortran/src/H5_ff.F90`) resets the counter like this:

```fortran
! Reset the number of open objects from h5open_f to zero
CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_ALL_F, H5OPEN_NUM_OBJ,  error)
```

`H5OPEN_NUM_OBJ` is a module variable of `H5F`. It is passed as the actual argument for the `INTENT(OUT)` dummy `obj_count`, and `h5fget_obj_count_f` (`fortran/src/H5Fff.F90`) *also* references `H5OPEN_NUM_OBJ` directly by use association:

```fortran
obj_count = H5Fget_obj_count(file_id, INT(obj_type, C_INT))

hdferr = 0
IF(obj_count.LT.0) hdferr = -1

! Don't include objects created by H5open in the H5F_OBJ_ALL_F count
IF(file_id.EQ.INT(H5F_OBJ_ALL_F,HID_T))THEN
   obj_count = obj_count - H5OPEN_NUM_OBJ
ENDIF
```

This is prohibited by F2018 15.5.2.13: a variable accessible by use association must not be redefined through a dummy argument in the same call, and having been redefined, must not be referenced. The result is compiler-dependent:

- If the compiler stores through the alias before reading the right-hand side (gfortran, ifx 2026.0.0), the subtraction evaluates as `0 - 0 = 0` and the counter is reset correctly, by accident.
- If the compiler uses copy-in/copy-out for the dummy, or caches `H5OPEN_NUM_OBJ` in a register as the standard permits, the subtraction evaluates as `0 - 51 = -51`.

(The raw `H5Fget_obj_count(H5F_OBJ_ALL)` after `h5close_types_c` is genuinely `0` — verified with a direct C-interop call — so `-51` can only arise from the stale read.)

Once `H5OPEN_NUM_OBJ` is `-51`, the guard added by #1657 at the top of `h5open_f`:

```fortran
! Check if H5open_f has already been called. If so, skip doing it again.
IF(H5OPEN_NUM_OBJ .NE. 0) RETURN
```

sees a non-zero value and returns immediately with `error = 0`. The interface is never re-initialized, and the failure is completely silent.

## Earlier/related changes in this file

The same aliasing hazard was already found and worked around in `h5open_f` in 2018 (commits `c52d04320` / `fe325fb55`, *"HDFFV-10405: Using h5fget_obj_count_f with a file id of H5F_OBJ_ALL_F does not work properly"*), which introduced a local temporary:

```fortran
INTEGER(SIZE_T) :: H5OPEN_NUM_OBJ_LOC = 0
...
CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_ALL_F, H5OPEN_NUM_OBJ_LOC,  error)
H5OPEN_NUM_OBJ = H5OPEN_NUM_OBJ_LOC
```

PR #1657 added the same call to `h5close_f` without that temporary.

## Proposed fix

In `h5close_f`, the intent stated in the comment is simply to zero the counter, so do that directly instead of routing it through an aliased call? Additionally, make `h5fget_obj_count_f` alias-safe for any caller by computing into a local.

```diff
--- a/fortran/src/H5_ff.F90
+++ b/fortran/src/H5_ff.F90
@@
-    ! Reset the number of open objects from h5open_f to zero
-    CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_ALL_F, H5OPEN_NUM_OBJ,  error)
+    ! Reset the number of open objects from h5open_f to zero.
+    H5OPEN_NUM_OBJ = 0
```

```diff
--- a/fortran/src/H5Fff.F90
+++ b/fortran/src/H5Fff.F90
@@
     INTEGER        , INTENT(OUT) :: hdferr
+    INTEGER(SIZE_T) :: count_loc
@@
-    obj_count = H5Fget_obj_count(file_id, INT(obj_type, C_INT))
+    ! Use a local variable so that the result is correct even if the actual
+    ! argument associated with obj_count is H5OPEN_NUM_OBJ itself.
+    count_loc = H5Fget_obj_count(file_id, INT(obj_type, C_INT))
 
     hdferr = 0
-    IF(obj_count.LT.0) hdferr = -1
+    IF(count_loc.LT.0) hdferr = -1
 
     ! Don't include objects created by H5open in the H5F_OBJ_ALL_F count
     IF(file_id.EQ.INT(H5F_OBJ_ALL_F,HID_T))THEN
-       obj_count = obj_count - H5OPEN_NUM_OBJ
+       count_loc = count_loc - H5OPEN_NUM_OBJ
     ENDIF
+
+    obj_count = count_loc
```

## Environment

- HDF5: 1.14.5 (also present in 1.14.6, 2.0.0, 2.1.0, develop @ d8ec63966)
- Affected: Intel oneAPI 2024.2.0 (ifort)
- Not affected: gfortran (all `-O` levels, with and without LTO); Intel oneAPI 2026.0.0 (ifx)
- Fortran API, serial build

### LINKED ISSUE #6648: Fortran: h5fget_obj_ids_f returns the interface's own identifiers, disagreeing with h5fget_obj_count_f
## Summary

`h5fget_obj_count_f` and `h5fget_obj_ids_f` disagree about the same query. The count
excludes the objects `h5open_f` opens to represent the predefined types; the id list
returns them. An application that asks "what do I have open" gets a consistent answer
from the count and a misleading one from the list.

The disagreement is internal to the Fortran wrapper: the two routines answer the
same question differently.

## Reproducer

```fortran
PROGRAM harm2
  USE HDF5
  IMPLICIT NONE
  INTEGER :: e
  INTEGER(SIZE_T) :: n, nids
  INTEGER(HID_T) :: f, g, ids(500)
  CALL h5open_f(e)
  CALL h5fcreate_f("harm2.h5", H5F_ACC_TRUNC_F, f, e)
  CALL h5gcreate_f(f, "g", g, e)

  CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_ALL_F, n, e)
  CALL h5fget_obj_ids_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_ALL_F, 500_SIZE_T, ids, e, nids)
  WRITE(*,'(A,I0,A,I0)') "ALL      -> count=", n, "   ids=", nids
  CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_DATATYPE_F, n, e)
  CALL h5fget_obj_ids_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_DATATYPE_F, 500_SIZE_T, ids, e, nids)
  WRITE(*,'(A,I0,A,I0)') "DATATYPE -> count=", n, "   ids=", nids

  CALL h5gclose_f(g, e); CALL h5fclose_f(f, e); CALL h5close_f(e)
END PROGRAM harm2
```

Only a file and a group are open, so both lines should report 2 and 0.

```
ALL      -> count=2   ids=62
DATATYPE -> count=0   ids=60
```

The equivalent C program reports:

```
ALL      -> count=2   ids=2
DATATYPE -> count=0   ids=0
```

This is not a direct comparison, and it does not show that the Fortran numbers are
wrong. h5open_f does considerably more than H5open: it opens 60 datatypes to represent
the predefined types, which the C program never creates. A C program that made the same
60 H5Tcopy calls would report 62 and 62. What the C run does show is that the C API is
internally consistent, which the Fortran wrapper is not.

## Cause

`h5fget_obj_count_f` (`fortran/src/H5Fff.F90`) subtracts the objects h5open_f opened
when `file_id` is `H5F_OBJ_ALL_F`, reporting what the application has open rather than
what is actually open. That subtraction dates to HDFFV-10405.

`h5fget_obj_ids_f` passes straight through to `H5Fget_obj_ids` with no equivalent
filtering, so it reports what is actually open.

Both are defensible in isolation. Neither is obviously the bug; the bug is that the
two do not agree.

## Why it matters

A caller cannot reconcile the two. Sizing a buffer from the count and filling it from
the list does not overflow, because `max_objs` bounds the call, but it can silently
truncate: the caller may receive the interface's identifiers instead of its own,
depending on iteration order. Anything that walks the list to close leftover objects
finds 60 datatypes it never opened, and closing them breaks the Fortran interface.

## Which way to reconcile them

Either direction makes the pair consistent, and the choice is a policy question about
what the Fortran binding should report:

1. Filter the identifiers to match the count. Both then report what the application
   has open, extending the existing HDFFV-10405 decision to `h5fget_obj_ids_f`. The
   interface's identifiers become unreachable through the Fortran API.

2. Drop the subtraction from the count. Both then report what is actually open, which
   is the literal truth given that `h5open_f` did open those datatypes. This changes
   long-standing behavior of `h5fget_obj_count_f`, and a caller checking for leaked
   objects has to know to expect the interface's own.

## Notes

Found while fixing #6642; not caused by it. That PR documents the current behavior of
both routines so the difference is at least written down, and leaves the behavior
alone.

brtnfld:
h5close_f reset its count of the objects created by h5open_f with

  CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_ALL_F, &
                          H5OPEN_NUM_OBJ, error)

passing the H5F module variable H5OPEN_NUM_OBJ as the actual argument for the INTENT(OUT) obj_count dummy, while h5fget_obj_count_f also reads H5OPEN_NUM_OBJ by use association. F2018 15.5.2.13 prohibits referencing a variable through use association once it has been redefined through a dummy argument in the same call, so the result depended on how the compiler implemented argument association. Where the actual argument was passed by reference the subtraction collapsed to 0 - 0 and produced the intended zero; where the compiler used copy-in/copy-out it evaluated 0 - H5OPEN_NUM_OBJ and left the count negative.

A negative count then defeats the guard at the top of h5open_f, which returns early when H5OPEN_NUM_OBJ is non-zero. h5open_f reported success without calling h5init_types_c, leaving H5T_NATIVE_INTEGER and the other predefined types holding identifiers that h5close_f had released.

h5close_f now assigns the count directly, which is what the comment there has always described. h5fget_obj_count_f computes into a local variable so that no caller can reintroduce the aliasing; note that this change alone would make the old h5close_f call site produce the negative count on every compiler rather than only on some, so the two belong together. H5OPEN_NUM_OBJ is also given an initial value, since h5open_f tests it before anything assigns to it.

h5fget_obj_count_f subtracted every object created by h5open_f from a count of a single object type, so with the interface open a query such as

  CALL h5fget_obj_count_f(INT(H5F_OBJ_ALL_F,HID_T), H5F_OBJ_FILE_F, n, error)

returned a negative n and hdferr of 0. h5open_f now records what it leaves open per object type, and a count is adjusted by the recorded value for the types being counted, so the adjustment does not depend on which types the initialization creates. The check for a negative count runs both on the value returned by H5Fget_obj_count and after the adjustment.

h5fget_obj_ids_f applied no such adjustment, so it returned the identifiers h5open_f opened alongside the application's own and disagreed with h5fget_obj_count_f about the same query: with only a file and a group open, H5F_OBJ_ALL_F counted 2 objects but listed 62. The C API reports 2 and 2. An application walking the list found datatypes it never opened, and closing them breaks the Fortran interface. h5fget_obj_ids_f now excludes those identifiers, requesting enough from H5Fget_obj_ids that max_objs of the application's own can still be returned when the two are interleaved.

The h5open/h5close test verified its object counts by calling h5fget_obj_count_f after h5close_f, when h5open_f is the only call the Fortran interface permits. Those checks move to after the interface is reopened, where they additionally confirm that the predefined types are valid again, that the preceding h5close_f released the previous h5open_f's types, and that h5fget_obj_ids_f agrees with h5fget_obj_count_f.

The Fortran tests also aborted unrecoverable failures with STOP, which exits with a success status whether the stop code is absent or is a string, so a run that died part way through reported no failure to CTest. They now exit through h5_exit_f(1). The STOPs that end a run normally, in fflush1 and in the async test's skip path for a build without MPI_THREAD_MULTIPLE, are unchanged.

Fixes #6642
Fixes #6648

Reported and diagnosed by Dom Heinzeller.

### DIFF
--- a/fortran/src/H5Fff.F90
+++ b/fortran/src/H5Fff.F90
@@ -41,7 +41,13 @@ MODULE H5F
   IMPLICIT NONE
 
   ! Number of objects opened in H5open_f
-  INTEGER(SIZE_T) :: H5OPEN_NUM_OBJ
+  INTEGER(SIZE_T) :: H5OPEN_NUM_OBJ = 0
+  ! The same count broken down by object type, in the order of the object type
+  ! list in h5fget_obj_count_f, so that a count can be adjusted without assuming
+  ! which types h5open_f creates.
+  INTEGER(SIZE_T), DIMENSION(1:4) :: H5OPEN_NUM_OBJ_BY_TYPE = 0
+
+  PRIVATE :: h5open_num_obj_matching, h5open_owns_id
 
 
 #ifndef H5_DOXYGEN
@@ -863,20 +869,54 @@ CONTAINS
 
   END SUBROUTINE h5fclose_async_f
 
+  ! Number of the objects h5open_f left open that a count or list of obj_type
+  ! would include.
+  FUNCTION h5open_num_obj_matching(obj_type) RESULT(num_obj)
+    IMPLICIT NONE
+    INTEGER, INTENT(IN) :: obj_type
+    INTEGER(SIZE_T) :: num_obj
+
+    IF(obj_type.EQ.H5F_OBJ_ALL_F)THEN
+       num_obj = H5OPEN_NUM_OBJ
+    ELSE
+       num_obj = 0
+       IF(IAND(obj_type,H5F_OBJ_FILE_F    ).NE.0) num_obj = num_obj + H5OPEN_NUM_OBJ_BY_TYPE(1)
+       IF(IAND(obj_type,H5F_OBJ_DATASET_F ).NE.0) num_obj = num_obj + H5OPEN_NUM_OBJ_BY_TYPE(2)
+       IF(IAND(obj_type,H5F_OBJ_GROUP_F   ).NE.0) num_obj = num_obj + H5OPEN_NUM_OBJ_BY_TYPE(3)
+       IF(IAND(obj_type,H5F_OBJ_DATATYPE_F).NE.0) num_obj = num_obj + H5OPEN_NUM_OBJ_BY_TYPE(4)
+    ENDIF
+
+  END FUNCTION h5open_num_obj_matching
+
+  ! Whether id is one of the identifiers h5open_f opened for the Fortran interface.
+  FUNCTION h5open_owns_id(id) RESULT(owns)
+    IMPLICIT NONE
+    INTEGER(HID_T), INTENT(IN) :: id
+    LOGICAL :: owns
+
+    owns = ANY(predef_types.EQ.id) .OR. ANY(floating_types.EQ.id) .OR. ANY(integer_types.EQ.id)
+
+  END FUNCTION h5open_owns_id
+
 !>
 !! \ingroup FH5F
 !!
 !! \brief Gets number of the objects open within a file.
 !!
 !! \param file_id   File identifier
-!! \param obj_type  Type of the object; possible values are:
+!! \param obj_type  Type of the objects to count, either one of the following or
+!!                  several of them combined with IOR():
 !!                  \li H5F_OBJ_FILE_F
 !!                  \li H5F_OBJ_DATASET_F
 !!                  \li H5F_OBJ_GROUP_F
 !!                  \li H5F_OBJ_DATATYPE_F
-!!                  \li H5F_OBJ_ALL_F
+!!                  \li H5F_OBJ_ALL_F, which counts all of the above
 !! \param obj_count Number of open objects
 !! \param hdferr    \fortran_error
+!!
+!! \note When \p file_id is H5F_OBJ_ALL_F the count excludes the objects that
+!!       h5open_f opens for the Fortran interface, so it reports only what the
+!!       application itself has open.
 !!
 !! See C API: @ref H5Fget_obj_count()
 !!
@@ -886,6 +926,8 @@ CONTAINS
     INTEGER        , INTENT(IN)  :: obj_type
     INTEGER(SIZE_T), INTENT(OUT) :: obj_count
     INTEGER        , INTENT(OUT) :: hdferr
+
+    INTEGER(SIZE_T) :: count_loc
 
     INTERFACE
        INTEGER(SIZE_T) FUNCTION H5Fget_obj_count(file_id, obj_type) BIND(C,NAME='H5Fget_obj_count')
@@ -897,15 +939,21 @@ CONTAINS
        END FUNCTION H5Fget_obj_count
     END INTERFACE
 
-    obj_count = H5Fget_obj_count(file_id, INT(obj_type, C_INT))
+    count_loc = H5Fget_obj_count(file_id, INT(obj_type, C_INT))
 
     hdferr = 0
-    IF(obj_count.LT.0) hdferr = -1
 
-    ! Don't include objects created by H5open in the H5F_OBJ_ALL_F count
-    IF(file_id.EQ.INT(H5F_OBJ_ALL_F,HID_T))THEN
-       obj_count = obj_count - H5OPEN_NUM_OBJ
-    ENDIF
+    ! H5Fget_obj_count reports failure as a negative count.
+    IF(count_loc.LT.0) hdferr = -1
+
+    ! Don't include the objects h5open_f left open.
+    IF(file_id.EQ.INT(H5F_OBJ_ALL_F,HID_T)) &
+         count_loc = count_loc - h5open_num_obj_matching(obj_type)
+
+    ! A count is never negative, so neither is one that has been adjusted.
+    IF(count_loc.LT.0) hdferr = -1
+
+    obj_count = count_loc
 
   END SUBROUTINE h5fget_obj_count_f
 
@@ -915,16 +963,21 @@ CONTAINS
 !! \brief Get list of open objects identifiers within a file.
 !!
 !! \param file_id  File identifier
-!! \param obj_type Type of the object; possible values are:
+!! \param obj_type Type of the objects to list, either one of the following or
+!!                 several of them combined with IOR():
 !!                 \li H5F_OBJ_FILE_F
 !!                 \li H5F_OBJ_DATASET_F
 !!                 \li H5F_OBJ_GROUP_F
 !!                 \li H5F_OBJ_DATATYPE_F
-!!                 \li H5F_OBJ_ALL_F
+!!                 \li H5F_OBJ_ALL_F, which lists all of the above
 !! \param max_objs Maximum # of objects to retrieve
 !! \param obj_ids  Array of open object identifiers
 !! \param hdferr   \fortran_error
 !! \param num_objs Number of open objects
+!!
+!! \note When \p file_id is H5F_OBJ_ALL_F the list excludes the identifiers that
+!!       h5open_f opens for the Fortran interface, matching h5fget_obj_count_f, so
+!!       it reports only what the application itself has open.
 !!
 !! See C API: @ref H5Fget_obj_ids()
 !!
@@ -938,6 +991,10 @@ CONTAINS
     INTEGER(SIZE_T), INTENT(OUT), OPTIONAL :: num_objs
 
     INTEGER(SIZE_T) :: c_num_objs ! Number of open objects of the specified type
+    INTEGER(SIZE_T) :: num_skip   ! Objects h5open_f left open that this query would return
+    INTEGER(SIZE_T) :: idx, num_kept
+    INTEGER :: alloc_stat
+    INTEGER(HID_T), ALLOCATABLE :: obj_ids_loc(:)
 
     INTERFACE
        INTEGER(SIZE_T) FUNCTION H5Fget_obj_ids(file_id, obj_type, max_objs, obj_ids) &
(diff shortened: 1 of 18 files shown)


climbfuji:
Thanks for the quick turnaround! I'll give this a test today with oneapi 2024.2 ifort.

hyoklee (approved):

climbfuji:
@brtnfld This are looking good! Here is what I ran for this branch on a Cray system with Intel oneAPI 2024.2.0, ifort.
```
$ mkdir build && cd build
$ cmake -DHDF5_BUILD_FORTRAN=ON -DH5EXAMPLE_BUILD_FORTRAN=ON -DBUILD_TESTING=ON -DHDF5_TEST_FORTRAN=ON .. 2>&1 | tee log.cmake
$ make VERBOSE=1 2>&1 | tee log.make
$ ctest 2>&1 | tee log.ctest

[...]

100% tests passed, 0 tests failed out of 3033

Total Test time (real) = 639.59 sec

The following tests did not run:
        373 - H5LS-tstarfile (Disabled)
        374 - H5LS-tqmarkfile (Disabled)
        585 - H5IMPORT-DUMP-BINARY_F64 (Disabled)
        591 - H5IMPORT-DUMP-BINARY_I8 (Disabled)
        702 - H5REPACK-gzip_individual (Disabled)
        703 - H5REPACK-gzip_all (Disabled)
        704 - H5REPACK-szip_individual (Disabled)
        705 - H5REPACK-szip_all (Disabled)
        722 - H5REPACK-all_filters (Disabled)
        723 - H5REPACK-CMP-gzip_verbose_filters (Disabled)
        724 - H5REPACK-szip_copy (Disabled)
        725 - H5REPACK-szip_remove (Disabled)
        726 - H5REPACK-deflate_copy (Disabled)
        727 - H5REPACK-deflate_remove (Disabled)
        768 - H5REPACK-remove_all (Disabled)
        769 - H5REPACK-deflate_convert (Disabled)
        770 - H5REPACK-szip_convert (Disabled)
        771 - H5REPACK-deflate_limit (Disabled)
        772 - H5REPACK-deflate_file (Disabled)
        942 - H5REPACK-VERIFY_LAYOUT-layout_long_switches (Disabled)
        943 - H5REPACK-VERIFY_LAYOUT-layout_short_switches (Disabled)
        944 - H5REPACK-global_filters (Disabled)
        945 - H5REPACK-old_style_layout_short_switches (Disabled)
        962 - H5REPACK-gt_mallocsize (Disabled)
        999 - H5REPACK_VERIFY_LAYOUT_VDS-vds_dset_chunk20x10x5 (Disabled)
        1000 - H5REPACK_VERIFY_LAYOUT_VDS-vds_chunk2x5x8 (Disabled)
        1001 - H5REPACK_VERIFY_LAYOUT_VDS-vds_chunk3x6x9 (Disabled)
        1002 - H5REPACK_VERIFY_LAYOUT_VDS-vds_compa (Disabled)
        1003 - H5REPACK_VERIFY_LAYOUT_VDS-vds_conti (Disabled)
        1077 - H5REPACK-h5copy_extlinks_src-base (Disabled)
        1110 - H5REPACK-textlinkfar-merge (Disabled)
        1111 - H5REPACK-textlinksrc-merge (Disabled)
        1112 - H5REPACK-textlinktar-merge (Disabled)
        1149 - H5REPACK-textlinkfar-mergeprune (Disabled)
        1150 - H5REPACK-textlinksrc-mergeprune (Disabled)
        1151 - H5REPACK-textlinktar-mergeprune (Disabled)
        2235 - H5DUMP-t128bit_float (Disabled)
```

climbfuji:
Thanks for fixing this so quickly, much appreciated. Looking forward to the next release that has this update.

PR #6652, opened 2026-08-29
H5Pget_driver_info: do not push an error when the driver has no info block

### LINKED ISSUE #6651: H5Pget_driver_info() pushes an error when the driver has no info block, contrary to its documented contract
### Summary

`H5Pget_driver_info()` pushes an error onto the error stack when the driver has
registered no driver-specific properties. Both its own header comment and its entry in
`H5Ppublic.h` say that case returns NULL with **no** error pushed.

`src/H5Pfapl.c`:

```c
/*
 *        Failure:    NULL. Null is also returned if the driver has
 *                not registered any driver-specific properties
 *                although no error is pushed on the stack in
 *                this case.
 */
const void *
H5Pget_driver_info(hid_t plist_id)
{
    ...
    /* Get the driver info */
    if (NULL == (ret_value = (const void *)H5P_peek_driver_info(plist)))
        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, NULL, "can't get driver info");
```

`H5Ppublic.h` states the same contract to users:

> If no driver-specific properties have been registered, H5Pget_driver_info() returns NULL.

`H5P_peek_driver_info()` returns `driver_prop.driver_info`, which is legitimately NULL
whenever the driver was selected without an info block — which is what
`H5Pset_driver_by_name()` and the `HDF5_DRIVER` environment variable both do. The
function cannot distinguish that from a failure, so it reports every one as a failure.

The sibling `H5Pget_driver_config_str()` handles the same situation correctly:

```c
    if ((config_str = H5P_peek_driver_config_str(plist))) {
```

### Effect

Any VFD that fetches its optional FAPL info block sees an HDF5-DIAG stack printed on
every open, describing an error that did not occur:

```
HDF5-DIAG: Error detected in HDF5 (2.3.0):
  #000: src/H5Pfapl.c line 1546 in H5Pget_driver_info(): can't get driver info
    major: Property lists
    minor: Can't get value
```

HDF5's own `src/H5FDmulti.c` calls it in exactly this pattern, in four places, and treats
NULL as "no info block":

```c
        fa = (const H5FD_multi_fapl_t *)H5Pget_driver_info(fapl_id);
```

The noise is harmless to the operation, but it is not harmless to testing: any test that
compares a program's stderr against a reference sees a diff, and anyone reading the
output of a driver under development is told there is an error when there is not.

I hit it running the netCDF-C test suite over an out-of-tree VFD selected with
`HDF5_DRIVER=<name>`; every `H5Fopen` in the run produced one of these blocks.

### Fix

Drop the wrapper `HGOTO_ERROR` and assign the result. The two genuine failures still
report themselves — `H5P_peek_driver_info()` pushes `H5E_BADTYPE` for a list that is not
a file access property list, and `H5E_CANTGET` when the property cannot be read — so
nothing that was diagnosed stops being diagnosed; only the frame that fired on a normal
NULL goes away. That also removes the duplicate second frame stacked on top of the two
real errors.

PR to follow.

### Environment

HDF5 `develop` @ `b7b85e7abf9`, Windows 11, MSVC 19.x. The code is platform-independent
and dates to at least [svn-r15510] (2008), so this is not a recent regression.

hyoklee:
Fixes #6651.

`H5Pget_driver_info()` pushed `H5E_CANTGET` whenever `H5P_peek_driver_info()` returned NULL. NULL is not necessarily a failure there: a driver that registered no driver-specific properties leaves `driver_info` NULL, which is exactly what `H5Pset_driver_by_name()` and the `HDF5_DRIVER` environment variable produce.

Both the function's own header comment and its entry in `H5Ppublic.h` say that case returns NULL with no error pushed:

> Failure: NULL. Null is also returned if the driver has not registered any driver-specific properties although no error is pushed on the stack in this case.

and

> If no driver-specific properties have been registered, H5Pget_driver_info() returns NULL.

The sibling `H5Pget_driver_config_str()` already handles the same situation correctly:

```c
    if ((config_str = H5P_peek_driver_config_str(plist))) {
```

### Effect

Any VFD that fetches its optional FAPL info block — including HDF5's own `H5FDmulti.c`, in four places — prints an HDF5-DIAG stack on every open describing an error that did not occur:

```
HDF5-DIAG: Error detected in HDF5 (2.3.0):
  #000: src/H5Pfapl.c line 1546 in H5Pget_driver_info(): can't get driver info
    major: Property lists
    minor: Can't get value
```

I hit this running the netCDF-C test suite over an out-of-tree VFD selected with `HDF5_DRIVER=<name>`: every `H5Fopen` in the run produced one of these. Harmless to the operation, but not to testing — any test comparing a program's stderr against a reference sees a diff, and someone reading the output of a driver under development is told there is an error when there is not.

### The change

Drop the wrapper `HGOTO_ERROR` and assign the result. Nothing that was diagnosed stops being diagnosed: `H5P_peek_driver_info()` pushes `H5E_BADTYPE` itself for a list that is not a file access property list, and `H5E_CANTGET` when the property cannot be read. The removed frame only ever duplicated one of those or fired on a normal NULL.

Not a recent regression — the code dates to at least [svn-r15510] (2008).

### DIFF
--- a/src/H5Pfapl.c
+++ b/src/H5Pfapl.c
@@ -1541,9 +1541,22 @@ H5Pget_driver_info(hid_t plist_id)
     if (NULL == (plist = (H5P_genplist_t *)H5I_object_verify(plist_id, H5I_GENPROP_LST)))
         HGOTO_ERROR(H5E_ARGS, H5E_BADTYPE, NULL, "not a property list");
 
-    /* Get the driver info */
-    if (NULL == (ret_value = (const void *)H5P_peek_driver_info(plist)))
-        HGOTO_ERROR(H5E_PLIST, H5E_CANTGET, NULL, "can't get driver info");
+    /* Get the driver info.
+     *
+     * NULL here is not necessarily a failure: a driver that registered no
+     * driver-specific properties leaves driver_info NULL, and both this
+     * function's own header above and its entry in H5Ppublic.h say that case
+     * returns NULL with nothing pushed on the error stack.  It used to push
+     * H5E_CANTGET anyway, so every H5Fopen through a driver selected by
+     * H5Pset_driver_by_name() or the HDF5_DRIVER environment variable -- which
+     * set no info block -- printed an error stack that described no error.
+     *
+     * The two real failures still report themselves: H5P_peek_driver_info()
+     * pushes H5E_BADTYPE for a list that is not a file access property list,
+     * and H5E_CANTGET when the property cannot be read.  Adding a second frame
+     * on top of those said nothing the first did not.
+     */
+    ret_value = (const void *)H5P_peek_driver_info(plist);
 
 done:
     FUNC_LEAVE_API(ret_value)


hyoklee:
Verified with a minimal reproducer. `H5Pset_driver_by_name(fapl, "sec2", NULL)` is what `HDF5_DRIVER=sec2` does internally and sets no driver-info block, so `driver_info` is legitimately NULL:

```c
if ((fapl = H5Pcreate(H5P_FILE_ACCESS)) < 0) return 1;
if (H5Pset_driver_by_name(fapl, "sec2", NULL) < 0) return 1;

info  = H5Pget_driver_info(fapl);
nerrs = H5Eget_num(H5E_DEFAULT);
```

Before, on `develop` @ `b7b85e7abf9`, Windows 11 / MSVC 19.x:

```
HDF5-DIAG: Error detected in HDF5 (2.3.0):
  #000: src\H5Pfapl.c line 1546 in H5Pget_driver_info(): can't get driver info
    major: Property lists
    minor: Can't get value
driver_info = NULL (no info block)
errors on the stack after the call = 1
```

After:

```
driver_info = NULL (no info block)
errors on the stack after the call = 0
```

Same return value, and the stack is now clean — which is what the documented contract says should have been happening all along.

PR #6653, opened 2026-08-31
Align LICENSE with the OSI-approved BSD-3-Clause text

brtnfld:
## Motivation

While auditing data collected by our CASS sustainability-metrics tooling ([corsa-center/metrics](https://github.com/corsa-center/metrics)), we found it couldn't confirm HDF5's license as OSI-approved. Digging in: GitHub's own license classifier already can't do so either — it returns `spdx_id: NOASSERTION`, `license_key: "other"` for this repo, because `LICENSE` isn't a verbatim match for BSD-3-Clause. Nothing in it is *wrong*, but the differences are enough that automated tooling can't confirm the match:

- Clause 3 is reworded (names HDF Group/University/Contributor explicitly, requires permission "from" each respectively) instead of the standard generic wording.
- An "Enhancements" contributor-license-grant clause is appended after the disclaimer — legally real and important, just not part of the BSD-3-Clause template.
- Several institutional/funding-agency notices (LBNL, LLNL, NCSA, NSF, DOE contract numbers) are also appended in the same file.

This PR doesn't change what anyone has agreed to — it relocates the non-license material out of `LICENSE` so the file is a clean, verbatim, automatically-recognizable BSD-3-Clause license, while keeping everything else fully intact elsewhere.

## What changed

- **`LICENSE`**: now the verbatim OSI/SPDX BSD-3-Clause text, diffed word-for-word against [opensource.org/license/bsd-3-clause](https://opensource.org/license/bsd-3-clause) to confirm an exact match. Copyright holders (HDF Group / University of Illinois) preserved.
- **`NOTICE`** (new): the LBNL/LLNL/NCSA/NSF/DOE funding acknowledgments and contributor credits, moved here verbatim. This mirrors the standard Apache Software Foundation convention: every Apache-2.0 project keeps `LICENSE` as the pure license grant and pushes attribution/acknowledgment text into a companion `NOTICE` file. We are not introducing a novel pattern here, just adopting an established one.
- **`CONTRIBUTING.md`**: the "Enhancements" contributor-grant clause moved here verbatim, next to the existing contributor-agreement notice — it's a grant *from* contributors, so it fits more naturally in the doc contributors are already pointed to.
- **`release_docs/RELEASE_PROCESS.md`**: added NOTICE to the release-time doc-review checklist.
- **`README.md`**: license badge updated from a generic "BSD" static badge to "BSD-3-Clause" (kept static rather than switching to a live `shields.io/github/license/...` badge, since that would show whatever GitHub currently classifies the *default* branch as — stale until this PR actually merges).
- **`CMakeInstallation.cmake`**: binary-package install rules (RPM/DEB/Windows installer) only referenced `LICENSE` by name, so the LLNL redistribution notice would have silently dropped out of binary distributions once it moved to `NOTICE`. Now installs both files, configures a matching `NOTICE.txt` for the Windows package, and updates `CPACK_RPM_PACKAGE_LICENSE` from the free-text "BSD-style" to the proper SPDX identifier "BSD-3-Clause". (No Autotools or in-repo spec/debian/conda packaging exists in this repo to check — this repo has no top-level `Makefile.am`, and downstream packaging like conda-forge lives elsewhere.)

## Verification

- Every clause in the original `LICENSE` preserved somewhere — checked via grep for each distinguishing phrase (DOE contract numbers, contributor list, Enhancements grant, clause 3 wording).
- New `LICENSE` body diffed against the canonical BSD-3-Clause text: exact match.
- Searched the repo for every other file referencing LICENSE/BSD/Enhancements; only `README.md` (a badge link, unaffected) and `CONTRIBUTING.md` (updated) had substantive ties.

## Note for maintainers

`CONTRIBUTING.md` already says contributor agreement is scoped to "the LICENSE file"; I've updated that sentence to also point at the Enhancements clause's new home. Since this touches the contributor-agreement mechanism, it may be worth a quick look from whoever owns license/contribution policy before merging.


CMakeInstallation.cmake
```diff
@@ -252,18 +252,19 @@ endif ()
 HDF_README_PROPERTIES(HDF5_BUILD_FORTRAN)
 
 #-----------------------------------------------------------------------------
-# Configure the LICENSE.txt file for the windows binary package
+# Configure the LICENSE.txt and NOTICE.txt files for the windows binary package
 #-----------------------------------------------------------------------------
 if (WIN32)
   configure_file (${HDF5_SOURCE_DIR}/LICENSE ${HDF5_BINARY_DIR}/LICENSE.txt @ONLY)
+  configure_file (${HDF5_SOURCE_DIR}/NOTICE ${HDF5_BINARY_DIR}/NOTICE.txt @ONLY)
```
jhendersonHDF: I'm not sure that either of these are needed since there don't appear to be any `configure_file` transformations taking place; it's possible this is only being used to add the .txt extension (which can be done other ways or just not at all). Besides that, if there were transformations taking place, the `install` command below is installing the pre-transformation versions anyway. Suggest reviewing LICENSE vs. LICENSE.txt and NOTICE vs. NOTICE.txt to see if the `configure_file` can be exchanged/removed or if the install command should be fixed. Since the current Windows binary has the file as LICENSE with no extension, I lean toward just removing the `configure_file`.

brtnfld: You're right about the NOTICE.txt one — removed in 112032b, it had no consumer.

The LICENSE.txt one turns out to be real though, not just an extension add: CPACK_RESOURCE_FILE_LICENSE gets overridden further down (in the WIX_EXECUTABLE block) to point at ${HDF5_BINARY_DIR}/LICENSE.txt specifically, for the WiX installer's license-acceptance screen — CMake's docs say that variable needs a .txt or .rtf extension to get converted to RTF by the WiX generator. So it's disconnected from the plain install() step (matching your observation that the shipped binary has the file as LICENSE with no extension), but it does feed that one WiX-specific variable. Left it in place and added a comment pointing at where it's used so the next person doesn't have to trace it again.

jhendersonHDF: Seems like it is just an extension add then, which would be less confusing if done with a `COPY_FILE` or similar. Would also make sense to only do so near the packaging logic, since WiX isn't always available and the file may not even be used. Unrelated to this PR though so no changes needed here.

jhendersonHDF (approved):

brtnfld:
Closing, as more research is needed.

PR #6654, opened 2026-08-31
Fix deadlock in ROS3 VFD on Windows

### LINKED ISSUE #6560: ROS3/S3 VFD deadlocks at process exit on Windows
## Summary

A program that reads an object over the ROS3 (S3) VFD and then exits normally hangs forever during process shutdown on Windows. All application work completes successfully (files open, read, and close fine); the process simply never exits.

A native stack captured from a minidump of the hung process shows the main thread blocked in the HDF5 S3 cleanup that runs at process exit, trying to join AWS worker threads while the Windows loader lock is held.

This does not reproduce on Linux or macOS. This was reported previously in https://github.com/h5py/h5py/issues/2850 and separately caused downstream CI issues in https://github.com/NeurodataWithoutBorders/pynwb/issues/2228 which I maintain.

_Disclaimer:_ This issue was investigated using AI (Claude Opus 4.8) to help narrow down the root cause and suggest fixes. AI was used to analyze the native stack, understand the AWS C SDK behavior, and draft this issue report. I reviewed and edited the final content to the best of my ability, but I do not have a deep understanding of the HDF5 or AWS C SDK internals, so please verify the analysis and proposed fixes.

## Environment

- HDF5 2.1.0 (conda-forge build, ROS3 VFD backed by the AWS C SDK)
- aws-c-s3 0.12.8, aws-c-common 0.14.2, aws-c-io 0.27.3, aws-c-auth 0.10.4 (conda-forge)
- h5py 3.16.0, Python 3.14 (but the deadlock is in native teardown and is not Python-specific)
- Windows Server 2025 (GitHub Actions `windows-latest`)

## Reproduction

Open any S3 object with the `ros3` driver on Windows, then let the process exit:

```python
import h5py

url = "https://dandiarchive.s3.amazonaws.com/ros3test.nwb"
with h5py.File(url, mode="r", driver="ros3", aws_region=b"us-east-2") as f:
    _ = f  # read/close all succeed
# On Windows the process now hangs here, during interpreter/process exit.
```

The read completes and the file closes cleanly; the hang is purely at process teardown.

## Native stack of the hung main thread

Captured with `procdump -ma <pid>` and `cdb -z dump.dmp -c "~*k"`. The `hdf5!...` frame symbols are approximate (nearest exported symbol + offset) because the conda-forge `hdf5.dll` ships without PDBs; the `aws_c_*` and `ntdll!Ldr*` frames are exported and unambiguous.

```
ntdll!NtWaitForAlertByThreadId
ntdll!RtlSleepConditionVariableSRW
KERNELBASE!SleepConditionVariableSRW
aws_c_common!aws_condition_variable_wait
aws_c_common!aws_condition_variable_wait_pred
aws_c_common!aws_thread_join_all_managed        <- blocks joining AWS worker threads
aws_c_s3!aws_s3_library_clean_up                <- AWS-C-S3 library teardown
hdf5!  (ROS3 / H5FD s3comms cleanup)            <- approximate symbol
ucrtbase!execute_onexit_table                   <- hdf5.dll CRT atexit table
ntdll!LdrpCallInitRoutine
ntdll!LdrShutdownProcess                        <- DLL_PROCESS_DETACH, loader lock held
ntdll!RtlExitUserProcess
kernel32!ExitProcessImplementation
ucrtbase!common_exit
python314!Py_Exit
...
python314!Py_RunMain
```

## Claude's hypothesized root cause

HDF5 registers the AWS-C-S3 library teardown (`aws_s3_library_clean_up`) to run at process exit via the C runtime's `onexit`/`atexit` table in `hdf5.dll`. On Windows, that table runs inside `DLL_PROCESS_DETACH`, which the loader executes **while holding the loader lock**. `aws_s3_library_clean_up()` calls `aws_thread_join_all_managed()`, which blocks waiting to join AWS-managed worker threads. Those threads cannot make progress to completion while the loader lock is held, so the join never returns and the process deadlocks.

Joining threads during `DllMain`/`DLL_PROCESS_DETACH` is not supported on Windows; this is a classic loader-lock deadlock.

### Why it is Windows-only

Linux and macOS join threads at exit without a loader-lock-during-detach constraint, so the same cleanup completes normally.

### Why it appeared with HDF5 2.x

HDF5 2.x's ROS3 VFD is backed by the AWS C SDK (`aws-c-s3` / `aws-c-common`), which spawns managed worker threads and registers the join-on-exit cleanup. The earlier libcurl-based ROS3 VFD spawned no such threads, so exit was clean.

### Corroborating observation

In h5py's test suite (`h5py/tests/test_ros3.py`), the tests that call `h5py.File(..., driver="ros3")` to open or attempt to open an S3 location **hang on exit**: `test_ros3` (opens a real object) and `test_ros3_s3_fails` (attempts an open against a nonexistent bucket, which still initializes the S3 VFD). `test_ros3_temp_token`, which only builds and queries a file-access property list (`set_fapl_ros3` / `get_fapl_ros3_token`) without opening a file, does not hang. This split is reported in https://github.com/h5py/h5py/issues/2850 and is consistent with the deadlock being in `aws_s3_library_clean_up` joining threads that exist only once the AWS-C-S3 library has been initialized by an S3 open.

### What does not work as a workaround

`os._exit()` / `ExitProcess` do **not** avoid the hang, because on Windows they still run `DLL_PROCESS_DETACH` and hit the same deadlock. Only terminating the process from an external parent process (which is not itself running the loader teardown) avoids it.

## Claude's suggested fixes

- Do not join threads during process-exit DLL detach on Windows. For example, skip the `aws_thread_join_all_managed()` path when running under `DLL_PROCESS_DETACH`, or
- Do not register `aws_s3_library_clean_up` to run via `atexit`/DLL detach at all; instead tie AWS-C-S3 init/cleanup to explicit `H5FDros3`/S3 VFD driver registration/termination or to `H5close()`, so cleanup happens before process exit rather than during loader teardown.

jhendersonHDF:
Avoid calling aws-c-s3 library cleanup if the process is in the middle of DLL shutdown

### DIFF
--- a/src/H5FDros3_s3comms.c
+++ b/src/H5FDros3_s3comms.c
@@ -46,6 +46,14 @@
 #include <aws/http/request_response.h>
 #include <aws/sdkutils/aws_profile.h>
 
+#if defined(H5_HAVE_WINDOWS) && defined(H5_HAVE_RTLDLLSHUTDOWNINPROGRESS)
+/* Prototype for RtlDllShutdownInProgress has to be declared - see
+ * https://learn.microsoft.com/en-us/windows/win32/devnotes/rtldllshutdowninprogress.
+ */
+#include <windows.h>
+BOOLEAN NTAPI RtlDllShutdownInProgress(VOID);
+#endif
+
 /****************/
 /* Local Macros */
 /****************/
@@ -335,12 +343,16 @@ H5FD__s3comms_init(void)
         }
     }
 
-    /* Work around issue where aws-c-s3 library doesn't shut down
-     * cleanly when called from HDF5's default atexit() handler.
+    /* On non-Windows platforms, work around issue where aws-c-s3 library
+     * doesn't shut down cleanly when called from HDF5's default atexit()
+     * handler. On Windows platforms, this cleanup is performed during VFD
+     * termination.
      */
+#ifndef H5_HAVE_WINDOWS
     if (0 != atexit(H5FD__s3comms_term_func))
         HGOTO_ERROR(H5E_VFL, H5E_CANTINIT, FAIL,
                     "couldn't register function for cleaning up aws-c-s3 library");
+#endif
 
     H5FD_ros3_aws_init_g = true;
 
@@ -368,7 +380,21 @@ H5FD__s3comms_term_func(void)
     if (H5FD_ros3_aws_init_g) {
         aws_host_resolver_release(H5FD_ros3_aws_host_resolver_g);
         aws_event_loop_group_release(H5FD_ros3_aws_event_loop_group_g);
+
+#ifndef H5_HAVE_WINDOWS
         aws_s3_library_clean_up();
+#elif defined(H5_HAVE_RTLDLLSHUTDOWNINPROGRESS)
+        /* On Windows, the loader lock can be held if the library is terminating
+         * as a result of the process terminating. In this case, calling the
+         * aws-c-s3 library cleanup routine might cause a deadlock when trying to
+         * cleanup threads. If DLL shutdown is in progress, just skip cleanup for
+         * now until a better solution is available. Ideally, calling H5close()
+         * would be a requirement so that aws-c-s3 cleanup can be properly handled
+         * at that time rather than during process termination.
+         */
+        if (!RtlDllShutdownInProgress())
+            aws_s3_library_clean_up();
+#endif
 
         /* Clean up logger interface after being sure everything else
          * has finished cleaning up
@@ -397,9 +423,12 @@ H5FD__s3comms_term(void)
 
     FUNC_ENTER_PACKAGE_NOERR
 
-    /* Currently handled by atexit() function above to work around cleanup
-     * ordering issues.
+    /* On non-Windows platforms, cleanup is currently handled by an atexit()
+     * function above to work around cleanup ordering issues.
      */
+#ifdef H5_HAVE_WINDOWS
+    H5FD__s3comms_term_func();
+#endif
 
     FUNC_LEAVE_NOAPI(ret_value)
 } /* end H5FD__s3comms_term() */
--- a/config/ConfigureChecks.cmake
+++ b/config/ConfigureChecks.cmake
@@ -642,6 +642,12 @@ if (HDF5_ENABLE_ROS3_VFD)
       message (FATAL_ERROR "Found aws-c-s3 library but CMake target for library didn't exist")
     endif ()
 
+    if (WINDOWS)
+      # RtlDllShutdownInProgress is used to avoid potential Windows loader lock
+      # deadlocks during library shutdown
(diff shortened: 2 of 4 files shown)


jhendersonHDF:
Testing of the VFD on Windows is still manual as the testing is setup to expect usage of the s3proxy docker image, which is only available for Linux currently. More work is needed to get the tests running on Windows.

src/H5FDros3_s3comms.c
```diff
@@ -368,7 +380,21 @@ H5FD__s3comms_term_func(void)
...
+#elif defined(H5_HAVE_RTLDLLSHUTDOWNINPROGRESS)
+        /* On Windows, the loader lock can be held if the library is terminating
+         * as a result of the process terminating. In this case, calling the
+         * aws-c-s3 library cleanup routine might cause a deadlock when trying to
+         * cleanup threads. If DLL shutdown is in progress, just skip cleanup for
+         * now until a better solution is available. Ideally, calling H5close()
+         * would be a requirement so that aws-c-s3 cleanup can be properly handled
+         * at that time rather than during process termination.
+         */
+        if (!RtlDllShutdownInProgress())
```
jhendersonHDF, line 395: Using `RtlDllShutdownInProgress` is really a hack here, but always performing proper cleanup would require much bigger changes to the library. Some function needs to be available for calling before the process exits, with `H5close()` being the most natural. With the changes in this PR, calling `H5close()` before exiting will properly clean things up, but calling `H5close()` isn't a requirement for HDF5 applications and the cleanup could still be called during process exit.

jhendersonHDF:
@rly @hmaarrfk @takluyver

hmaarrfk:
Interesting.  

The use of atexit has created other similar problems for me. 

I'll try to open an issue on h5py soon about this.

rly:
Thanks @jhendersonHDF. With help from AI, I tested this on Windows and it fixes the deadlock.

The diff applies cleanly to the `hdf5_2.1.0` tag, so I built that tag with and without the patch and swapped the resulting `hdf5.dll` under conda-forge's `h5py` 3.16.0 (same ABI, no h5py rebuild), then ran pynwb's ROS3 suite (which was passing but deadlocked on exit) plus a minimal open-and-exit repro on GitHub Actions' `windows-latest`:

| | conda-forge 2.1.0 | 2.1.0 built here | 2.1.0 built here + this PR |
|---|---|---|---|
| process exit after a ROS3 read | still alive at 180s | still alive at 180s | exited in 1.1s |
| pynwb ROS3 tests | 4 passed | 4 passed | 4 passed |

The middle column is the same build as the right one minus the patch, so the patch is the only variable.

Setup and full logs are in https://github.com/NeurodataWithoutBorders/pynwb/pull/2256 (run: https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/33473222181). That job is throwaway, but if any of it is useful as a starting point for Windows ROS3 CI on your side, please take it.

jhendersonHDF:
Thanks for testing @rly; all the information and caveats listed in that PR match the tradeoffs I had to make considering the current architecture of HDF5, so everything looks good from the side of implementation and testing.

lrknox (approved):

lrknox (approved):

PR #6660, opened 2026-09-03
Fix assertion failure checking number of filters in pipeline

### LINKED ISSUE #6655: `assert()` for maximum limit on number of filters in pipeline is off by 1
The maximum number of filters that can be added to a filter pipeline is documented as 32 (`H5Z_MAX_NFILTERS`). The `assert()` in `H5Z_pipeline()` currently checks that the number of filters in a pipeline is less than OR equal to this number, causing a filter pipeline with 32 filters in it to result in a crash for debug builds. This should be updated to allow the documented number of filters and a test should be added.

jhendersonHDF:


docs/doxygen/dox/H5.format.4.0.dox
```diff
@@ -9146,7 +9146,7 @@ link may have changed. If the link does not work, use the latest version of this
   <td colspan="4"><br />Filter Description List <em>(variable size)</em><br /><br /></td>
 </tr>
 </table>
-u
```
jhendersonHDF, line 9149: A previous PR seems to have unintentionally added this stray character

lrknox (approved):

PR #6663, opened 2026-09-04
✨ RFC-HDFG-2026-001: String-based filter configuration API

### LINKED ISSUE #6153: ✨ [Feature Request] String-Based Configuration Interface for Filters

brtnfld:
## Summary
- Adds a human-readable key=value parameter string API for HDF5 filters, alongside the existing integer `cd_values` arrays: `H5Pappend_filter`, `H5Pmodify_filter_by_idx`, `H5Pget_filter_params_by_idx`, `H5Zconfig_get_int/double/bool/str`, `H5Z_filter_id_by_name`, `H5Zget_filter_class_info`.
- Extends `H5Z_class3_t` with `name` (canonical, validated at registration) and `description`, plus `set_config`/`get_config` callbacks, and threads `dxpl_id`/`scaled[]`/`ndims` through `H5Z_pipeline` so v3 filter callbacks have full context. All six built-in filters (deflate, shuffle, fletcher32, nbit, szip, scaleoffset) implement `set_config`/`get_config`.
- New on-disk pipeline v3 format: adds an `H5F_LIBVER_V300` file-format bound (the new `LATEST`) under which the canonicalized parameter string is persisted verbatim alongside each filter's `cd_values` in the pipeline message, so a file/tool without the original plugin can still show a human-readable configuration. Older bounds are unaffected; read compatibility with pre-3.0 files/libraries is preserved.
- Vendors a TOML subset parser (tomlc17, MIT) in `src/tomlc17/`, built unconditionally into libhdf5 with hidden visibility to avoid symbol collisions; updated to tag R260821 to pick up upstream's own subnormal-handling fix.
- Adds Fortran, C++, and Java (JNI + FFM) bindings, plus `h5dump`/`h5repack` support for the new parameter strings, including a `UD=id,flags,key=value` h5repack CLI form and hex-float annotation for binary-exact floats in h5dump output.

**Resolves the review feedback from the earlier iteration of this PR:** @mattjala noted that the `filter_title`/`cd_values` mechanism described in RFC §4.3/5.2/10.1 didn't match the implementation. RFC-HDFG-2026-001 has since been amended to describe the mechanism actually implemented: `H5Z_class3_t.name` is a canonical, uniqueness-checked identifier validated at `H5Zregister()` time and persisted on disk via the pipeline v3 message (satisfying the "survives without the plugin" goal from a different angle than originally spec'd); `description` remains registry-only and is documented as best-effort. The reserved in-file blob-callback placeholders that had no implementation have been removed from `H5Z_class3_t`.

This branch also went through a full comprehensive-review pass: 3 Critical, 10 High, 13 Medium, and 7 Low findings were identified and all but two have been fixed and verified (including a self-inflicted file-corruption path from an unbounded canonicalized parameter string, and a stack buffer over-read in h5dump). The two exceptions are called out as deliberate, tracked follow-ups rather than being folded in here:
- `H5P_ENCODE_VERS` was not bumped — it's shared by every encoded property list, not just the filter pipeline, and deserves its own focused review.
- The 10 new Java JNI test methods are not yet ported to the FFM test suite (no JDK available in this environment to verify a port).

Fixes #6153.


release_docs/CHANGELOG.md
```diff
@@ -152,66 +353,80 @@ We would like to thank the many HDF5 community members who contributed to this r
...
+     attribute into a fixed `char *` buffer would overwrite it with a heap-allocated `char *` pointer
+     rather than the string content, which is undefined behaviour and could corrupt memory or produce
+     garbage comparison results.
+     All four routines now read VLEN CLASS attributes properly (via `H5Treclaim`) and compare the
+     string content: `H5DSis_scale()`, `H5IMis_image()`, and `H5IMis_palette()` return 1 when the
+     value matches exactly, and `H5DS_is_reserved()` correctly identifies reserved class names stored
+     as VLEN strings.
+   - **NUL-termination hardening.** The read buffer is now allocated one byte larger than the stored
+     attribute size, and a NUL terminator is explicitly written after the attribute data. This protects
+     `strcmp` from over-reading files where the CLASS attribute was written without strictly honouring
```
hyoklee, line 413: `honouring` -> `honoring`

release_docs/CHANGELOG.md
```diff
@@ -152,66 +353,80 @@ We would like to thank the many HDF5 community members who contributed to this r
...
+   guard and were directly exposed.) These
+   comparisons now use `strcmp()` so only an exact class name is accepted.
 
-## C++ APIs
+   Additional fixes applied to all four routines:
 
-## Testing
+   - **VLEN-string CLASS attributes are now handled correctly.** Previously, reading a VLEN-typed
+     attribute into a fixed `char *` buffer would overwrite it with a heap-allocated `char *` pointer
+     rather than the string content, which is undefined behaviour and could corrupt memory or produce
```
hyoklee, line 405: `behaviour` -> `behavior`

hyoklee (changes requested):

docs/doxygen/dox/H5.format.4.0.dox
```diff
@@ -9321,6 +9321,120 @@ u
...
+<table>
+<caption align="top"><strong>Fields: Filter Description - Version 3</strong></caption>
+<tr>
+  <th width="30%">Field Name</th>
+  <th>Description</th>
+</tr>
+<tr valign=top>
+  <td>Filter Identification, Name Length, Flags, Number of Client Data Values, Name, Client Data</td>
+  <td>Identical in encoding and semantics to the version 2 <em>Filter Description</em> fields described
+      above.</td>
```
vchoi-hdfgroup, line 9410: Add a reference link to the "version 2 filter description" field.

docs/doxygen/dox/H5.format.4.0.dox
```diff
@@ -9321,6 +9321,120 @@ u
...
+  <th width="30%">Field Name</th>
+  <th>Description</th>
+</tr>
+<tr valign=top>
+  <td>Version</td>
+  <td>The version number for this message. This table describes version 3. A filter pipeline message is
+      version 3 only when at least one filter in the pipeline carries a stored configuration string (see
+      <em>Config String Length</em> and <em>Config String</em> below); a pipeline with no stored strings is
+      still written as version 2, so version 2 and version 3 files remain byte-identical whenever this
+      feature is unused. Version 3 messages are written only when the file's high library-version bound
```
vchoi-hdfgroup, line 9354: Suggest to change to "when the file's latest format version bound (set with H5Pset_libver_bounds()) is ...."

c++/src/H5FilterParam.h
```diff
@@ -0,0 +1,115 @@
...
+    Wraps H5Zconfig_has_key, H5Zconfig_get_int, H5Zconfig_get_double,
+    H5Zconfig_get_bool, and H5Zconfig_get_str as overloaded
+    \p config_get_param methods.
+*/
+class FilterParam {
+  public:
+    ///\brief Returns true if \p key is present in \p params, false if absent.
+    ///\exception H5::LibraryIException on error.
+    static bool
+    config_has_key(const H5std_string &params, const H5std_string &key)
```
bmribler, line 37: Method's name should conform to existing pattern, e.g., static void clearErrorStack();

c++/src/H5FilterParam.h
```diff
@@ -0,0 +1,115 @@
...
+        return ret > 0;
+    }
+
+    ///\brief Retrieves a boolean parameter from \p params.
+    ///\return true if found, false if absent.
+    ///\exception H5::LibraryIException on error.
+    static bool
+    config_get_param(const H5std_string &params, const H5std_string &key, bool &value)
+    {
+        hbool_t c_val = false;
```
bmribler, line 75: Should be bool
htri_t H5Zconfig_get_bool(const char *params, const char *key, bool *out)

c++/src/H5FilterParam.h
```diff
@@ -0,0 +1,115 @@
...
+        if (ret < 0)
+            throw LibraryIException("FilterParam::config_get_param", "H5Zconfig_get_str failed");
+        if (ret == 0)
+            return false;
+        size_t            capacity = buf_size + 1;
+        std::vector<char> buf(capacity, '\0');
+        ret = H5Zconfig_get_str(params.c_str(), key.c_str(), buf.data(), &capacity);
+        if (ret < 0)
+            throw LibraryIException("FilterParam::config_get_param", "H5Zconfig_get_str failed");
+        value = H5std_string(buf.data());
```
bmribler, line 101: value = buf.data() is adequate, saving the temporary object created by H5std_string()

test/h5test.c
```diff
@@ -111,7 +111,7 @@ const char *LIBVER_NAMES[] = {"earliest", /* H5F_LIBVER_EARLIEST = 0  */
                               "v112",     /* H5F_LIBVER_V112 = 3      */
                               "v114",     /* H5F_LIBVER_V114 = 4      */
                               "v200",     /* H5F_LIBVER_V200 = 5      */
-                              "latest",   /* H5F_LIBVER_LATEST        */
+                              "v300",     /* H5F_LIBVER_V300 = 6      */
```
vchoi-hdfgroup, line 114: Should "v300" be added after "v200" instead of replacing "latest"?

vchoi-hdfgroup (changes requested):

bmribler (changes requested):

ajelenak:
### The proposal

Pass the dataset's unfiltered chunk size to the filter plugin callback.

New `H5Z_class3_t` extends the callback signature:

```c
typedef size_t (*H5Z_func2_t)(unsigned int flags, size_t cd_nelmts,
                              const unsigned int cd_values[],
                              hid_t dxpl_id, const hsize_t *scaled, size_t ndims,
                              size_t nbytes, size_t *buf_size, void **buf);
```

and `H5Z_pipeline` follows. Adding one more `size_t` parameter alongside them
should be a small addition. Especially because the current v3 shape tells the
plugin *where* its chunk sits in the dataset, but not *how big* it is meant to
eventually come out. Now is the right time to consider this need as well.

### How this helps?

1. Sometimes the guess is impossible. This came up in HDFGroup/hdf5_plugins#116 and HDFGroup/hdf5_plugins#280. Zstandard frames normally record their decompressed size in a header, and the HDF5 Zstd plugin reads it from there. But that field is optional, a writer using Zstandard's streaming interface omits it by default. Such chunk output is perfectly valid and HDF5 has a supported route for producing one via `H5Dwrite_chunk`. When the header field is missing, the filter has nothing left to read the size from, so today the plugin either guesses or gives up. The library, meanwhile, knows the number the whole time.

2. Guessing is expensive when it does work. Zstandard offers an upper-bound function, but for an unknown-size frame the bound is "number of blocks x 128 KB". On the test fixture added in hdf5_plugins#280, a 265-byte chunk whose true expanded size is 256 bytes gets a bound of 131072 — a 512x over-allocation. (That function is also declared static-linking-only, so a plugin built against a shared libzstd cannot call it at all.)

3. Every filter re-solves it. Each compression plugin invents its own way to recover this number the library already has, e.g.: read a container header, stash the size in `cd_values` from `set_local`, or prepend a length field to the stored bytes. That is duplicated effort and duplicated bugs, and the `set_local` route only works for files written through that same plugin, not for chunks written by third-party producers.

### What to be aware of

- Multi-filter pipelines. The library knows the size of the *fully* unfiltered chunk, not the intermediate sizes between stacked filters. So the value is exact for the last decode step, and only a hint for anything running before it. I would suggest documenting it plainly as a dataset property — "the final byte size of dataset chunk" — rather than as a per-step promise. In the common arrangements (a compressor alone, or a compressor behind a size-preserving pre-filter such as shuffle) it is exactly the expected output.
- It is a hint, not a mandate. Filters should stay free to ignore it, and the existing post-pipeline check should stay exactly where it is — the value is a starting point for allocation, not a substitute for validation.

fortran/src/H5f90proto.h
```diff
@@ -587,9 +585,13 @@ H5FC_DLL int_f h5dont_atexit_c(void);
...
-H5FC_DLL int_f h5zunregister_c(int_f *filter);
-H5FC_DLL int_f h5zfilter_avail_c(int_f *filter, int_f *flag);
-H5FC_DLL int_f h5zget_filter_info_c(int_f *filter, int_f *flag);
+H5FC_DLL herr_t H5Pappend_filter_str_c(hid_t plist, H5Z_filter_t id, unsigned flags, const char *params);
+H5FC_DLL herr_t H5Pappend_filter_raw_c(hid_t plist, H5Z_filter_t id, unsigned flags, size_t cd_nelmts,
+                                       const unsigned *cd_values);
+H5FC_DLL herr_t H5Pmodify_filter_by_idx_str_c(hid_t plist, unsigned filter_idx, unsigned flags,
+                                              const char *params);
+H5FC_DLL herr_t H5Pmodify_filter_by_idx_raw_c(hid_t plist, unsigned filter_idx, unsigned flags,
+                                              size_t cd_nelmts, const unsigned *cd_values);
```
bmribler, line 594: Why are these H5P and not h5p?

fortnern:
Reminder that we will need to create a new branch for the version 2 line when we merge this, unless it's after 2.3.0 and we're sure there won't be any more releases until 3.0.0.

PR #6674, opened 2026-09-16
Delay allocating chunk buffer until threaded section during threaded chunk reads

fortnern:
This appears to substantially  improve performance. I tried also freeing the chunk in the threaded section to reduce memory footprint, but that greatly hurt performance. It may be worth adding that as an option anyways for memory concerns.


src/H5Dchunk.c
```diff
@@ -3598,11 +3599,35 @@ H5D__chunk_thread_read(void *_threaded_chunk_info)
...
+     * up to *buf_size bytes into *buf.  For incompressible chunks
+     * chunk_disk_size >= chunk_size so buf_alloc stays at chunk_nbytes. */
+    buf_alloc = (threaded_chunk_info->chunk_nbytes < threaded_chunk_info->threaded_io_info->chunk_size)
+                    ? threaded_chunk_info->threaded_io_info->chunk_size
+                    : threaded_chunk_info->chunk_nbytes;
+
+    /* If we're using filters we can allocate the chunk outside of the mutex, otherwise we need to protect it
+     * since the free list package isn't yet threadsafe. This is a hack that should be removed once H5FL is
+     * threadsafe. */
+    if (threaded_chunk_info->old_pline->nused)
```
mattjala: This doesn't guard against cases where `old_pline` is NULL, which it can be for partial edge chunks if `H5O_LAYOUT_CHUNK_DONT_FILTER_PARTIAL_BOUND_CHUNKS` is set.  Could potentially segfault in some cases.

src/H5Dchunk.c
mattjala: For a filtered dataset with `H5D_CHUNK_DONT_FILTER_PARTIAL_CHUNKS` enabled, on a partial edge chunk with the default chunk cache enabled, this can lead to a NULL derefence on a second read.

The sequence is: The first read creates an entry inside `H5D_chunk_lock`, ent->chunk is populated with the default NULL value for a chunk at H5Dchunk.c:5539 (since the allocation of chunk was moved out of the shared block and into the threaded block). The first read gets the correct data from `threaded_chunk_info->chunk`, but because `old_pline` isn't populated, `ent->chunk` stays NULL. Then during the second read, `H5D__chunk_read` will then go down the non-threaded in-cache branch, set cpt_store.compact.buf to NULL, and attempt to memcpy from NULL during the read.

If the `UINT_MAX != threaded_chunk_info->udata.idx_hint` chunk-pointer patch is moved up to outisde of the `if(threaded_chunk_info->old_pline)`, then it would run on every threaded chunk and none of them would end up NULL during cached reads.

mattjala: This failure unlock loop no longer functions properly. When every registered chunk was non_NULL, they would all consistently get unlocked, but now that `chunk_info[i].chunk` is NULL for all of them (since this whole block depends on `chunks_locked`, which can only be true during a failure if it fails before workers are dispatched, which in turn means that no worker had the chance to overwrite/populate the `chunk` field), the condition at 3527 will always resolve false and the entries will stay locked and in the cache indefinitely.

I think we could probably just drop the guard on `chunk` here.

src/H5Dchunk.c
```diff
@@ -3459,7 +3458,8 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
                     /* We must evict all chunks if a worker failed, because the chunk may be in an
                      * inconsistent state in memory */
                     for (size_t i = 0; i < threaded_io_info->num_chunks; i++)
```
mattjala, line 3460: Should this loop iterate on `idx_hint` instead of `i`? `i` isn't used as an index in the loop body.

src/H5Dchunk.c
mattjala: `H5D__chunk_lock` now returns non-negative/negative, and the new `_chunk` out parameter's behavior should be documented.

src/H5Dchunk.c
```diff
@@ -6630,7 +6656,7 @@ static herr_t
 H5D__chunk_prune_fill(H5D_chunk_it_ud1_t *udata, bool new_unfilt_chunk)
 {
     const H5D_io_info_t *io_info         = udata->io_info;          /* Local pointer to I/O info */
-    const H5D_t         *dset            = udata->dset_info->dset;  /* Local pointer to the dataset info */
+    const H5D_t         *dset            = udata->dset_info->dset;  /* Local pointer to qthe dataset info */
```
mattjala: typo: `qthe` -> `the`

mattjala (changes requested):

src/H5Dchunk.c
fortnern: I forgot to point to the udata in the ith chunk_info, thanks

fortnern:
@mattjala I think I've addressed all your comments

src/H5Dchunk.c
```diff
@@ -5111,30 +5136,36 @@ H5D__chunk_cache_prune(const H5D_t *dset, size_t size)
...
- *        by the caller but will be valid until it is unlocked.  The
- *        input value IDX_HINT is used to speed up cache lookups and
- *        it's output value should be given to H5D__chunk_unlock().
- *        IDX_HINT is ignored if it is out of range, and if it points
- *        to the wrong entry then we fall back to the normal search
- *        method.
+ * Purpose:    Return a pointer to a dataset chunk via the _chunk
+ *        parameter.  The pointer points directly into the chunk cache
+ *        and should not be freed by the caller but will be valid
+ *        until it is unlocked.  The value in udata->idx_hunt is used
```
mattjala, line 5142: typo: `idx_hunt` -> `idx_hint`

src/H5Dchunk.c
```diff
@@ -3459,8 +3458,11 @@ H5D__chunk_read(H5D_io_info_t *io_info, H5D_dset_io_info_t *dset_info)
                     /* We must evict all chunks if a worker failed, because the chunk may be in an
                      * inconsistent state in memory */
                     for (size_t i = 0; i < threaded_io_info->num_chunks; i++)
-                        if (H5D__chunk_cache_evict(dset_info->dset,
-                                                   dset_info->dset->shared->cache.chunk.slot[udata.idx_hint],
+                        if (dset_info->dset->shared->cache.chunk
+                                .slot[threaded_io_info->chunk_info[i].udata.idx_hint] &&
```
mattjala, line 3462: It would be a good idea to have a  `idx_hint == UINT_MAX` guard here before indexing, since that could if there's a worker failure.

mattjala (dismissed):

mattjala (approved):

PR #6675, opened 2026-09-18
Separate internal threading from API thread safety

mattjala:
The internal chunk threading introduced in #6645 depends on the `HDF5_ENABLE_CONCURRENCY` CMake option. That option is flagged as incompatible with the high-level library, because it makes `H5is_library_threadsafe()` report true and the high-level library is not, in general, threadsafe.

Internal threading on its own has no problem with the HL library. It adds no application threads and its workers never enter the public API, so a build that only parallelizes the library's internals is safe for the HL library.

This PR splits internal multithreading control out of `HDF5_ENABLE_CONCURRENCY` into its own CMake option, `HDF5_ENABLE_INTERNAL_THREADS`, which is compatible with the HL library.

HDF5_ENABLE_CONCURRENCY is a superset of HDF5_ENABLE_INTERNAL_THREADS, so concurrency builds are unchanged from #6645.

Enabling internal threading without global concurrency required splitting off an additional macro from existing concurrency and threadsafe controls: `H5_HAVE_THREAD_LOCAL_STATE`, which enables per-thread API context and error stacks. Concurrency, Threadsafe, and internal multi-threading builds now all use this shared symbol.

Builds without an API lock have no H5TS_api_lock() to perform the one-time initialization of the thread-safety code, so H5_API_LOCK calls the new H5TS_first_thread_init() in that slot instead.

This is primarily motivated by h5py compatibility. h5py always builds the HL library, but currently has no distributions that enable concurrency at present.


src/H5TSint.c
```diff
@@ -174,9 +178,13 @@ H5TS_term_package(void)
     /* Reset global API lock info */
 #ifdef H5_HAVE_THREADSAFE
     H5TS_mutex_destroy(&H5TS_api_info_p.api_mutex);
-#else /* H5_HAVE_CONCURRENCY */
+#elif defined(H5_HAVE_CONCURRENCY)
     H5TS_rwlock_destroy(&H5TS_api_info_p.api_lock);
 #endif
+
+#ifdef H5_HAVE_INTERNAL_THREADS
+    H5TS_mutex_destroy(&H5TS_api_info_p.internal_mutex);
```
mattjala, line 186: This is a bit of cleanup which was missing, not strictly related to the CMake/macro changes

jhendersonHDF:
As a general comment, I don't see

> This is primarily motivated by h5py compatibility. h5py always builds the HL library, but currently has no distributions that enable concurrency at present.

as being particularly good motivation. Either the internal threads feature is going to be considered production ready and thus h5py enabling concurrency makes sense, or that wont be the case. We should not introduce a new option just because a particular project doesn't choose to enable a particular option. Even with this option, there would still need to be h5py distributions built with it enabled anyway, which to me entirely goes against the motivation of this PR to begin with.

This new option also continues to leave the concurrency option with no purpose, has no relation to the internal threading already being performed by Subfiling, and further divides builds of a thread-enabled library between `H5_HAVE_THREADSAFE_API`, `H5_HAVE_THREAD_LOCAL_STATE`, `H5_HAVE_CONCURRENCY` and `H5_HAVE_INTERNAL_THREADS` which seems very easy to get incorrect.

qkoziol:
> As a general comment, I don't see
> 
> > This is primarily motivated by h5py compatibility. h5py always builds the HL library, but currently has no distributions that enable concurrency at present.
> 
> as being particularly good motivation. Either the internal threads feature is going to be considered production ready and thus h5py enabling concurrency makes sense, or that wont be the case. We should not introduce a new option just because a particular project doesn't choose to enable a particular option. Even with this option, there would still need to be h5py distributions built with it enabled anyway, which to me entirely goes against the motivation of this PR to begin with.
> 
> This new option also continues to leave the concurrency option with no purpose, has no relation to the internal threading already being performed by Subfiling, and further divides builds of a thread-enabled library between `H5_HAVE_THREADSAFE_API`, `H5_HAVE_THREAD_LOCAL_STATE`, `H5_HAVE_CONCURRENCY` and `H5_HAVE_INTERNAL_THREADS` which seems very easy to get incorrect.

In a big-picture way, I agree - this needs some more design thought.   There's two fundamental ideas here: "is the API threadsafe for multithreaded applications?" and "is the library able and allowed to use multithreaded algorithms?".   Those ideas are orthogonal (although both dependent on threading being available to use).  It would be good it users were presented with just those two ideas.

mattjala:
> Either the internal threads feature is going to be considered production ready and thus h5py enabling concurrency makes sense, or that wont be the case. We should not introduce a new option just because a particular project doesn't choose to enable a particular option. Even with this option, there would still need to be h5py distributions built with it enabled anyway, which to me entirely goes against the motivation of this PR to begin with.

h5py currently always builds with the HL library enabled (in order to get dimension scales), which is flagged as incompatible with concurrency. As such, getting h5py to distribute this new feature would require them to turn on not only concurrency but also the 'enable unsupported' flag in one or more of their distributions, which I think they would be reluctant to do.

I spoke with Neil about this, and my understanding is that nothing about the chunk read internal threading substantively interacts with the high-level API, so separating the concerns like this is a reasonable approach.

> This new option also continues to leave the concurrency option with no purpose, has no relation to the internal threading already being performed by Subfiling, and further divides builds of a thread-enabled library between `H5_HAVE_THREADSAFE_API`, `H5_HAVE_THREAD_LOCAL_STATE`, `H5_HAVE_CONCURRENCY` and `H5_HAVE_INTERNAL_THREADS` which seems very easy to get incorrect.

I agree that this contributes to an unfortunate proliferation of parallel options. Though it's not quite as bad as it seems from your description, since `H5_HAVE_THREAD_LOCAL_STATE`/`H5_HAVE_THREADSAFE_API` are purely internal. And I'd point out that this leaves the concurrency option with as much purpose as it had before #6645.

PR #6676, opened 2026-09-18
Increase characters for Priority.

lrknox:
20 characters in line 65 "fieldValues(first: 20) {" was insufficient - increased to 100 which gets the 4 Priority levels.


.github/workflows/update-progress.py
```diff
@@ -62,7 +62,7 @@ def _get_query(self) -> str:
                 pageInfo { hasNextPage, endCursor }
                 nodes {
                   id
-                  fieldValues(first: 20) {
+                  fieldValues(first: 100) {
```
lkurz, line 65: LGTM

lkurz (approved):

